PR #145, opened 2019-08-07
Docker: add initial flow tests script

Galvien:
Add initial flow test script which will be used for testing that the easymesh flows are not broken.
For start, add a test to cover section 5.3.1 in WiFi certified EasyMesh test plan V1.0 - MCUT Supported services discovery test.

Signed-off-by: Gal Wiener <gal.wiener@intel.com>


tomereli:
Again Travis failed us with false positive.

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
+#find prplmesh folder
```
arnout: Missing copyright header (yes I know, a couple of other scripts are also still missing it).

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
...
+}
+
+success() {
+    printf '\033[1;32m'"$@\n"'\033[0m'
+}
+
+run() {
+    dbg "$*"
+    "$@" || exit $?
+}
```
arnout, line 36: Separate issue: we have a bunch of scripts that have the above boilerplate in various forms. Maybe it's time to refactor that...

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
...
+        success "OK $msg"
+    else
+        err "FAIL $msg"
+        error=1
+    fi
+}
+
+start_gw_repeater() {
+    dbg "delete running containers before starting test"
+    docker rm -f $(docker ps -a -q) >/dev/null 2>&1
```
arnout: I have other containers running on my system, so this is not OK...

Can't we just specify 'gateway' and 'repeater' here? Yes, it's possible that this was given a different name when launching the test_gw_repeater script, but we don't need to support that use case I think.

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
...
+    dbg "start gw and repeater"
+    ${scriptdir}/test_gw_repeater.sh >/dev/null 2>&1
+}
+
+send_topology_query() {
+    dbg "Sending toplogy query to al_mac $1"
+    docker exec -it gateway ${installdir}/bin/beerocks_cli -c "bml_trigger_topology_discovery $1" >/dev/null 2>&1
```
arnout: Do we need to redirect output here? It may be useful for debugging.

Perhaps we can define a global variable `redirect=">/dev/null 2>&1"` that gets set to empty if VERBOSE is on.

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
...
+}
+
+check_log_for_query() {
+    dbg "Confirming topology query was received"
+    docker exec -it repeater cat tmp/$USER/beerocks/logs/beerocks_agent.log | grep -i -q "Topology Query"
```
arnout: No need for cat, just grep directly.

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
...
+}
+
+test_topology() {
+    status "test topology query"
+    send_topology_query $1
+    check_log_for_query
```
arnout: I don't think it makes much sense to have functions for send_topology_query and check_for_log_query - they will be used in only one place, and they're very simple.

Conversely, it might be useful to have functions for triggering a BML command:
```
send_bml_command() {
    docker exec -it gateway ${installdir}/bin/beerocks_cli -c "$*"
}
```
(note the `"$*"`, not the usual `"$@"`, because beerock_cli supports only one argument to -c).

The rest can just as well be inlined here directly I think.

(nitpick)

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
...
+}
+
+test_init() {
+    status "test initialization"
+    start_gw_repeater || {
+        err "start GW+Repeater failed, abort"
+        exit 1
+    }
+    AL_MAC=$(docker exec -it gateway ${installdir}/bin/beerocks_cli -c bml_conn_map | grep IRE_BRIDGE | awk '{print $5}' | cut -d ',' -f 1)
```
arnout, line 138: Ouch. After my changes for the non-Intel agent, there won't be a IRE_BRIDGE in the topology any more...

I was thinking that we should use fixed MAC addresses instead of the random ones we have now. We can pass a four-byte base MAC address as an argument to the runner, and extend it with predictable sub-addresses for the various interfaces we create within the container.

But that's something for a later PR.

tools/docker/tests/test_flows.sh
```diff
@@ -0,0 +1,106 @@
...
+}
+
+usage() {
+    echo "usage: $(basename $0) [-hva]"
+    echo "  options:"
+    echo "      -h|--help - show this help menu"
+    echo "      -v|--verbose - verbosity on"
+    echo "      -a|--all - test all flows"
+    echo "      --test-topology - test the topology discovery flow"
```
arnout: I like this, having the option to test all or just one.

However, there probably will be many flows after some time. So maybe it is better to have positional arguments for the flows to run, so we don't have to pass it to getopt. And we should then of course use the name of the function `test_topology` so we don't have to match the string either.

arnout (changes requested):
Although all of my comments are just nitpicks, except for the copyright header, I'd still like the chance to review the second iteration so I don't approve yet.

tools/docker/tests/test_flows.sh
tomereli: Gal - please fix this and in a different PR add the copyright header to the other scripts which are missing it :)

tools/docker/tests/test_flows.sh
tomereli: First task for Coral! :)

tools/docker/tests/test_flows.sh
tomereli:
```suggestion
    docker exec -it repeater grep -i -q "Topology Query" tmp/$USER/beerocks/logs/beerocks_agent.log
```

tomereli: Note - do not commit the suggestion as github offers - it adds a separate commit without DCO, so just overhead. Instead, change it in the file and add a fixup commit according to the CONTRIBUTING.md file.

tools/docker/tests/test_flows.sh
tomereli: I'm failing to understand the difference with having long options for running specific tests compared to having them as positional arguments - can you elaborate?

The use case we had in mind is to always use the --all option while developing, and see all of the flows which are not yet implemented as FAILURES.
Then for each new flow we will add the test and it will pass.

In travis, we will only use short options, so every time a new flow is ready, part of the PR is to update `.travis.yaml` to include that flow's test.

Finally - passing no positional arguments == `running test_gw_repeater --rm`

arnout: Positional arguments -> call it like this:
```
test_flows.sh -v test-topology test-foo test-bar
```
i.e. without the `--`.

In code, it becomes completely different of course. You wouldn't have variables for the different tests, but just something like:
```
for test; do
    ${test}
done
```
(without an `in`, the for loop iterates over the positional arguments).

For handling the `-a` option, you'd do something like
```
set $all_tests
```
where `all_tests` is a variable containing the names of all tests. Obviously.

tools/docker/tests/test_flows.sh
arnout: Actually, when you press "Commit suggestion" you can give the commit title, and there you can set the fixup message, i.e. "fixup! docker: add initial flow tests script".

And of course, you can anyway still move the fixup commit to the right place manually while doing git rebase -i.

tools/docker/tests/test_flows.sh
arnout: I did this and the change was merged, so you can be sure that the gateway's AL_MAC is 00:11:22:33:00:00 and the repeater's is aa:bb:cc:dd:00:00.

Note that you need to refresh your docker images for that (or pull them from dockerhub).

tools/docker/tests/test_flows.sh
tomereli: Oh I get it. makes the code much simpler.

tomereli (changes requested):
Now that full gw + repeater test is working in CI, we can add this to our CI as well.
Please also update .travis.yaml to run this script with the tests that should pass.
Every time a new test is added, .travis.yaml needs to also be updated. Make sure to add this to the CONTRIBUTING.md.

tools/docker/tests/test_flows.sh
Galvien: Will fix in the next review iteration

tools/docker/tests/test_flows.sh
```diff
@@ -35,25 +42,86 @@ report() {
 
 start_gw_repeater() {
     dbg "delete running containers before starting test"
-    docker rm -f $(docker ps -a -q) >/dev/null 2>&1
+    docker rm -f gateway
+    docker rm -f repeater
```
arnout: I didn't know about the -f option - that's convenient! :+1:

tools/docker/tests/test_flows.sh
```diff
@@ -35,25 +42,86 @@ report() {
...
     dbg "start gw and repeater"
-    ${scriptdir}/test_gw_repeater.sh >/dev/null 2>&1
+    ${scriptdir}/test_gw_repeater.sh -d 5 #>/dev/null 2>&1
```
arnout: Please don't leave commented-out code in the commit - or when you do, make sure that you document in the commit message why you have commented-out code.

tools/docker/tests/test_flows.sh
```diff
@@ -35,25 +42,86 @@ report() {
...
+    return 1
 }
 
 test_topology() {
     status "test topology query"
-    send_topology_query $1
-    check_log_for_query
+    send_bml_command "bml_trigger_topology_discovery $AL_MAC"
+    dbg "Confirming topology query was received"
+    docker exec -it repeater grep -i -q "Topology Query" tmp/$USER/beerocks/logs/beerocks_agent.log
```
arnout: Shouldn't this be `/tmp/...` ?

tools/docker/tests/test_flows.sh
```diff
@@ -85,21 +166,20 @@ main() {
...
+    echo $@
     test_init
-    [ -n "$TEST_ALL" -o -n "$TEST_TOPOLOGY" ] && report "test topology query" test_topology $AL_MAC
-    # TODO add more tests here    
+    for test in $@; do
+       report "test_${test}" test_${test}
+    done
 }
 
+ALL_TESTS="topology initial_ap_config ap_config_renew ap_config_bss_tear_down channel_selection ap_capability_info_reporting client_capability_info_reporting link_metric_collection steering_mandate_opportunity client_steering client_association backhaul_optimization layer_data_payload_trigger layer_data_payload"
```
arnout: Two minor nits here:

- I find it more logical to put the defaults that may be overridden by command-line options here, but to put 'global const' variables in the beginning of the script (i.e. before defining functions, i.e. the place where you'd define global variables in C code). But this is just personal preference.
- The long line should be split. Since it's between quotes and since it will be whitespace-expanded where it is used, you can just split it anywhere, so:

```
ALL_TESTS="topology initial_ap_config ap_config_renew ap_config_bss_tear_down
    channel_selection ap_capability_info_reporting client_capability_info_reporting
    ..."
```

(I didn't count the characters here, but we wrap at 100 columns, cfr. .clang-format.)

arnout (approved):
I have a couple of minor nits. Please rebase and fix up while rebasing (or, if you disagree with my comments, don't fix it up). I'm approving already so it can be merged as soon as the checks succeed.

tools/docker/tests/test_flows.sh
arnout: I hadn't noticed that this review comment wasn't taken into account yet. Anyway, there is no need to take into account changes that will be made by another PR. So feel free to leave this particular change for a future PR.

tomereli (approved):

PR #147, opened 2019-08-07
WSC Flow phase 1 - support for none Intel agent

arnout:
The M2 still needs to be filled in with the relevant information, but that's a bigger change because it requires the database to be updated with SSID information (currently SSID updates are only performed from an explicit command).

This PR goes on top of #154.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4397,15 +4397,23 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
         }
     }
 
-    if (cmdu_rx.getNextTlvType() != uint8_t(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
-        LOG(ERROR) << "Not vendor specific TLV (not Intel?)";
+    if (slave_state != STATE_WAIT_FOR_JOINED_RESPONSE) {
+        LOG(ERROR) << "slave_state != STATE_WAIT_FOR_JOINED_RESPONSE";
         return false;
     }
 
-    LOG(INFO) << "Intel controller join response";
-    if (!parse_intel_join_response(sd, cmdu_rx)) {
-        LOG(ERROR) << "Parse join response failed";
-        return false;
+    if (cmdu_rx.getNextTlvType() == uint8_t(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
+        LOG(INFO) << "Intel controller join response";
+        if (!parse_intel_join_response(sd, cmdu_rx)) {
+            LOG(ERROR) << "Parse join response failed";
+            return false;
+        }
+    } else {
+        LOG(INFO) << "Non-Intel controller join response";
+        if (!parse_non_intel_join_response(sd)) {
+            LOG(ERROR) << "Parse join response failed";
+            return false;
+        }
     }
 
     return true;
@@ -4538,6 +4546,54 @@ bool slave_thread::parse_intel_join_response(Socket *sd, ieee1905_1::CmduMessage
 
         slave_state = STATE_UPDATE_MONITOR_SON_CONFIG;
     }
+
+    return true;
+}
+
+bool slave_thread::parse_non_intel_join_response(Socket *sd)
+{
+    // request the current vap list from ap_manager
+    auto request = message_com::create_vs_message<
+        beerocks_message::cACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_REQUEST>(cmdu_tx);
+    if (request == nullptr) {
+        LOG(ERROR) << "Failed building cACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_REQUEST message!";
+        return false;
+    }
+    message_com::send_cmdu(ap_manager_socket, cmdu_tx);
+
+    // send all pending_client_association notifications
+    for (auto notify : pending_client_association_cmdu) {
+        auto notification = message_com::create_vs_message<
+            beerocks_message::cACTION_CONTROL_CLIENT_ASSOCIATED_NOTIFICATION>(cmdu_tx);
+        if (notification == nullptr) {
+            LOG(ERROR) << "Failed building message!";
+            return false;
+        }
+        notification->params() = notify.second;
+        send_cmdu_to_controller(cmdu_tx);
+    }
+    pending_client_association_cmdu.clear();
+
+    // No version checking for non-Intel controller
+
+    // TODO
+    //        auto notification = message_com::create_vs_message<
+    //            beerocks_message::cACTION_PLATFORM_MASTER_SLAVE_VERSIONS_NOTIFICATION>(cmdu_tx);
+    //        if (notification == nullptr) {
+    //            LOG(ERROR) << "Failed building message!";
+    //            return false;
+    //        }
+    //        string_utils::copy_string(notification->versions().master_version, master_version.c_str(),
+    //                                  sizeof(beerocks_message::sVersions::master_version));
+    //        string_utils::copy_string(notification->versions().slave_version, BEEROCKS_VERSION,
+    //                                  sizeof(beerocks_message::sVersions::slave_version));
+    //        message_com::send_cmdu(platform_manager_socket, cmdu_tx);
+    //        LOG(DEBUG) << "send ACTION_PLATFORM_MASTER_SLAVE_VERSIONS_NOTIFICATION";
+
+    // TODO set son_config
+    log_son_config();
+
+    slave_state = STATE_UPDATE_MONITOR_SON_CONFIG;
 
     return true;
 }
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4388,13 +4388,6 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
                    << "     bssid: " << bssid << std::endl
                    << "     authentication_type: " << authentication_type << std::endl
                    << "     encryption_type: " << encryption_type << std::endl;
-
-        if (!manufacturer.compare("Intel")) {
-            // TODO add support for none Intel agents
-            // https://github.com/prplfoundation/prplMesh/issues/134
-            LOG(ERROR) << "None Intel controller " << manufacturer << " , dropping message";
-            return false;
-        }
     }
 
     if (slave_state != STATE_WAIT_FOR_JOINED_RESPONSE) {
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -200,6 +200,25 @@ std::string db::get_node_ipv4(std::string mac)
     return n->ipv4;
 }
 
+bool db::set_node_manufacturer(std::string mac, std::string manufacturer)
+{
+    auto n = get_node(mac);
+    if (!n) {
+        return false;
+    }
+    n->manufacturer = manufacturer;
+    return true;
+}
+
+std::string db::get_node_manufacturer(std::string mac)
+{
+    auto n = get_node(mac);
+    if (!n) {
+        return std::string();
+    }
+    return n->manufacturer;
+}
+
 int db::get_node_channel(std::string mac)
 {
     auto n = get_node(mac);
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -1098,6 +1098,8 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
 
         database.set_node_ipv4(backhaul_mac, bridge_ipv4);
         database.set_node_ipv4(bridge_mac, bridge_ipv4);
+        database.set_node_manufacturer(backhaul_mac, "Intel");
+        database.set_node_manufacturer(bridge_mac, "Intel");
 
         database.set_node_type(backhaul_mac, beerocks::TYPE_IRE_BACKHAUL);
 
@@ -1113,6 +1115,7 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
         database.set_node_state(eth_switch_mac, beerocks::STATE_CONNECTED);
         database.set_node_name(eth_switch_mac, slave_name + "_ETH");
         database.set_node_ipv4(eth_switch_mac, bridge_ipv4);
+        database.set_node_manufacturer(eth_switch_mac, "Intel");
 
         //run locating task on ire
         if (!database.is_node_wireless(backhaul_mac)) {
@@ -1298,6 +1301,7 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
 
     database.set_node_name(radio_mac, slave_name + "_AP");
     database.set_node_ipv4(radio_mac, bridge_ipv4);
+    database.set_node_manufacturer(radio_mac, "Intel");
 
     // sd is assigned to src bridge mac
     sd->setPeerMac(bridge_mac);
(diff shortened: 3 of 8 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -500,10 +500,98 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
...
+
+    if (false && /* TODO use certification mode */
+        cmdu_rx.getNextTlvType() == int(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
+        LOG(INFO) << "Intel radio agent join (al_mac=" << al_mac << " ruid=" << ruid;
+        if (!handle_intel_slave_join(sd, cmdu_rx, cmdu_tx)) {
+            LOG(ERROR) << "Intel radio agent join failed (al_mac=" << al_mac << " ruid=" << ruid << ")";
+            return false;
+        }
+    } else {
+        LOG(INFO) << "Non-Intel radio agent join (al_mac=" << al_mac << " ruid=" << ruid << ")";
```
tomereli, line 617: I think all this should go to a separate function

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -490,14 +490,6 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
         }
     }
 
-    std::string manufacturer =
```
tomereli, line 606: I think it is beneficial to still have the manufacturer at least printed out here, and also add the manufacturer to the node structure in the DB so it will be seen in the network map.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4349,15 +4349,61 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
         }
     }
 
-    if (cmdu_rx.getNextTlvType() != uint8_t(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
-        LOG(ERROR) << "Not vendor specific TLV (not Intel?)";
-        return false;
-    }
+    if (cmdu_rx.getNextTlvType() == uint8_t(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
```
tomereli, line 4398: still - a few lines up the manufacturer check will return false for none intel controller.

controller/src/beerocks/master/son_master_thread.cpp
arnout: There is currently nothing in the DB for this information.

If we consider adding it, then maybe we should first think about refactoring the database in general...

arnout:
I finally understand why I was strugling so much with this PR. The problem is that the Intel slave join message really has nothing to do with 1905.1 autoconfig - it really is a topology message.

Therefore, I'm not so sure what to do with this. In the non-Intel case, we really should build up this information from the 1905.1 Topology Response message, which contains much more of the required information. It contains no information about the radios, but that should be retrieved from AP Capability Report messages, which is a separate flow. So doing all of this in the WSC M1 seems wrong to me...

tomereli:
(nitpick) we should consider adding an option to the agent which will make it start in generic mode. In this mode it will not identify itself in autoconfig as Intel, but as "prplMesh" manufacturer.
We should remove the trigger for tasks which are intel specific in the controller for none Intel agents and add a different print to the agent to mark it as operational, maybe after autoconfig success. This can be done in a different PR of course.

Even better - we can have one radio agent be Intel manufacturer and the other prplMesh, this way we are testing both modes in CI.

arnout:
> (nitpick) we should consider adding an option to the agent which will make it start in generic mode. In this mode it will not identify itself in autoconfig as Intel, but as "prplMesh" manufacturer.

That is in fact an excellent idea! Do you prefer that in this PR or in a new one?

> We should remove the trigger for tasks which are intel specific in the controller for none Intel agents and add a different print to the agent to mark it as operational, maybe after autoconfig success.

I'm not sure how to do that. Could I give that to @morantr ?

tomereli:
> > (nitpick) we should consider adding an option to the agent which will make it start in generic mode. In this mode it will not identify itself in autoconfig as Intel, but as "prplMesh" manufacturer.
> 
> That is in fact an excellent idea! Do you prefer that in this PR or in a new one?

This PR will be complete with this - so I suggest you start as a new commit in this PR, and if you don't finish by EOD we'll move it to a different PR so we can merge this one today.

> 
> > We should remove the trigger for tasks which are intel specific in the controller for none Intel agents and add a different print to the agent to mark it as operational, maybe after autoconfig success.
> 
> I'm not sure how to do that. Could I give that to @morantr ?

Yes - Moran can start working on that after the CAPI stuff. Could you open a task for this?

tomereli:
> (nitpick) we should consider adding an option to the agent which will make it start in generic mode. In this mode it will not identify itself in autoconfig as Intel, but as "prplMesh" manufacturer.
> We should remove the trigger for tasks which are intel specific in the controller for none Intel agents and add a different print to the agent to mark it as operational, maybe after autoconfig success.

One small improvement to the above - we can set one radio agent to be Intel, and the other to be generic, both by default. Then we can test in CI that the 2 modes are working. The option should go to the `beerocks_agent.conf` - we add the manufacturer there and the agent will use this in the WSC autoconfig flow.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: (nitpick) perhaps we can just print the manufacturer here so we'll have it in the log?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1425,6 +1440,92 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+    database.add_node(bridge_mac, backhaul_mac, ire_type);
+    database.set_node_state(bridge_mac, beerocks::STATE_CONNECTED);
+
+    // Controller expects backhaul which contains bridge which contains radio, but we only get radio...
+
+    // Update existing node, or add a new one
+    if (database.has_node(radio_mac)) {
+        if (database.get_node_type(radio_mac) != beerocks::TYPE_SLAVE) {
+            database.set_node_type(radio_mac, beerocks::TYPE_SLAVE);
+            LOG(ERROR) << "Existing mac node is not TYPE_SLAVE";
```
tomereli, line 1472: Shouldn't we return false here?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1425,6 +1440,92 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+        }
+        database.clear_hostap_stats_info(radio_mac);
+    } else {
+        // TODO Intel Slave Join has separate radio MAC and UID; we use radio_mac for both.
+        database.add_node(radio_mac, bridge_mac, beerocks::TYPE_SLAVE, radio_mac);
+    }
+    database.set_hostap_is_acs_enabled(radio_mac, false);
+    // TODO
```
tomereli: There should be no slave reconf in generic mode - we have WSC autoconfig renew for that, so I guess this TODO should be removed.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1425,6 +1440,92 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+    //        if (!notification->is_slave_reconf()) {
+    //            son_actions::set_hostap_active(database, tasks, radio_mac,
+    //                                           false); // make sure AP is marked as not active
+    //        }
+
+    // TODO Assume repeater mode
+    database.set_hostap_repeater_mode_flag(radio_mac, true);
+    // TODO Assume no backhaul manager
+    database.set_hostap_backhaul_manager(radio_mac, false);
```
tomereli, line 1484: This could be problematic since one of the agents in our implementation must be the "backhaul", so in the case of both radio agents being non-Intel the controller behaviour may be unexpected, we should verify that my concerns are invalid once we have the option to start both repeater radios in generic mode.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1425,6 +1440,92 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+    // TODO ipv4 will not be set
+
+    // sd is assigned to src bridge mac
+    sd->setPeerMac(bridge_mac);
+
+    // TODO database.set_hostap_supported_channels from radio_basic_caps->operating_classes_info_list
+    // TODO assume SSIDs are not hidden
+    database.set_hostap_advertise_ssid_flag(radio_mac, true);
+
+    // TODO
```
tomereli, line 1505: WSC autoconfig message contains the capabilities in the WSC M1 RF_Bands attribute:
image
We should use this information here.

controller/src/beerocks/master/son_master_thread.h
```diff
@@ -54,6 +54,9 @@ class master_thread : public beerocks::socket_thread {
     void handle_cmdu_control_ieee1905_1_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
     bool handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx,
                                  ieee1905_1::CmduMessageTx &cmdu_tx);
+    bool handle_non_intel_slave_join(Socket *sd, std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1,
+                                     std::string bridge_mac, std::string radio_mac,
```
tomereli, line 58: Better to use `const std::string` or `std::string &` to avoid copy constructor since we're not changing `bridge_mac` and `radio_mac`.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4397,15 +4397,62 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
+        if (slave_state != STATE_WAIT_FOR_JOINED_RESPONSE) {
+            LOG(ERROR) << "slave_state != STATE_WAIT_FOR_JOINED_RESPONSE";
+            return false;
+        }
 
-    LOG(INFO) << "Intel controller join response";
-    if (!parse_intel_join_response(sd, cmdu_rx)) {
-        LOG(ERROR) << "Parse join response failed";
-        return false;
+        // request the current vap list from ap_manager
```
tomereli: Deserves a separate function IMHO

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4397,15 +4397,62 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
-        return false;
-    }
+    if (cmdu_rx.getNextTlvType() == uint8_t(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
+        LOG(INFO) << "Intel controller join response";
+        if (!parse_intel_join_response(sd, cmdu_rx)) {
+            LOG(ERROR) << "Parse join response failed";
+            return false;
+        }
+    } else {
+        LOG(INFO) << "Non-Intel controller join response";
```
tomereli, line 4405: Could be helpful for future debugging to print the manufacturer string as part of this log print.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4397,15 +4397,62 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
+            if (notification == nullptr) {
+                LOG(ERROR) << "Failed building message!";
+                return false;
+            }
+            notification->params() = notify.second;
+            send_cmdu_to_controller(cmdu_tx);
+        }
+        pending_client_association_cmdu.clear();
+
+        // No version checking for non-Intel controller
```
tomereli: Agree - so why keep the TODO?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4388,13 +4388,6 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
                    << "     bssid: " << bssid << std::endl
                    << "     authentication_type: " << authentication_type << std::endl
                    << "     encryption_type: " << encryption_type << std::endl;
-
-        if (!manufacturer.compare("Intel")) {
```
tomereli, line 4392: Can we add the manufacturer print in the above log? will be useful for future debugging.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -200,6 +200,25 @@ std::string db::get_node_ipv4(std::string mac)
     return n->ipv4;
 }
 
+bool db::set_node_manufacturer(std::string mac, std::string manufacturer)
```
tomereli, line 203: I know the rest of this file doen't follow this - but better to have the parameters as const. This applies to all newly added functions below.

tomereli (changes requested):

arnout:
> > > We should remove the trigger for tasks which are intel specific in the controller for none Intel agents and add a different print to the agent to mark it as operational, maybe after autoconfig success.
> > 
> > I'm not sure how to do that. Could I give that to @morantr ?
> 
> Yes - Moran can start working on that after the CAPI stuff. Could you open a task for this?

#162

controller/src/beerocks/master/son_master_thread.cpp
arnout: > Shouldn't we return false here?

No.

The Intel slave handling doesn't do that.

Also, it makes sense that more recent information takes precedence over whatever happens to be stored in the database - as long as that is done in a consistent way. Setting the type to TYPE_SLAVE does exactly that.

controller/src/beerocks/master/son_master_thread.cpp
arnout: > This could be problematic since one of the agents in our implementation must be the "backhaul", so in the case of both radio agents being non-Intel the controller behaviour may be unexpected, we should verify that my concerns are invalid once we have the option to start both repeater radios in generic mode.

I don't think so. AFAICS, in our current test, neither of the radios in the repeater is a backhaul (it's wired backhaul).

controller/src/beerocks/master/son_master_thread.cpp
arnout: > WSC autoconfig message contains the capabilities in the WSC M1 RF_Bands attribute:
> We should use this information here.

I skipped that because in the current database, this information is not kept separately. Instead, it is constructed out of the channel information. We should get the channel information from the AP Capability Report message.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: I don't understand your question. The TODO is about passing the master/slave versions to the platform manager, where it will eventually be used by bml_get_master_slave_versions. The "No version checking" comment is about not checking the version (i.e. the version is always considered correct).

controller/src/beerocks/master/db/db.cpp
arnout: You made this comment in another review as well, but I don't understand it. Do you really want a signature like

```
void foo(const int x1, const bool yesno, const std::string bar);
```

?

Or do you only want const-ness for `std::string` and not for scalar types?

Or do you actually want a const ref `const std::string&` because you don't trust the copyref optimisation?

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: It's already printed together with the encrypted settings.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: It is already there.

controller/src/beerocks/master/db/db.cpp
tomereli: I prefer a signature with const values only for non-scalar types whenever possible. Also the function should be declared `const` if it a class function and doesn't change any members:

```
void foo(,int x1, bool yesno, const std::string &str) const; //function const for class members when possible
```
If the parameter is inout, then it should be passed without const.
If the function returns scalar type, it should return none const (anyway copied).
If the function returns non-scalar type - then it should return const when possible to save an extra copy constructor call.

Of course all these are really (nitpicks), so don't bother changing anything if you disagree.
Also 
```

tomereli (approved):

Issue #137, opened 2019-07-31
[BUG] tlvf: complex lists are not created correctly


arnout:
A complex list (i.e. a variable length list of structures, not scalars) is not written correctly by the TLVF generated code.

## Steps to reproduce

A commented-out test that exposes this behaviour is added to the tlvf test in PR #136. Uncomment the indicated sections, run the test, and the verification of the second complex list element fails. The generated bytes are also incorrect (visual inspection).

PR #148, opened 2019-08-07
tlvf: Finalize complex lists issues (solves #137)

tomereli:
tlvf: Finalize complex lists issues (solves #137)

Issue left is that on parse of complex list of classes -
Create is done correctly, but parsing fails due to the following:

1. On init(), the complex list index member  is set to the length
of the list (lets call it N).
2. create-add is called N times
2.1 On create(), the src pointer used for creating the cInner member
is not updated based on the index and is always set to the list start.
This breaks  parsing of complex lists with more then one member.
2.2 On add(), the src pointer used for validating the entry added
is the expected one is not set as it is in the create flow.
This as well breaks parsing of complex lists with more then one member.

Solution is to refactor how parsing of complex lists work -
1. On init(), set the index to 0 (doesn't affect create)
2. On complex list member create, use the correct src pointer which
points to the last added member end pointer, as done in create flow.
3. On complex list member add, incrememnt the index also for parse
and set the src pointer as done in create to fix the sanity
test of checking the member start pointer is correct


framework/tlvf/tlvf.py
```diff
@@ -819,7 +819,10 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
                 lines_cpp.append("%s %s = *m_%s;" %(length_type.type_str, param_length, param_length))
                 if length_type.swap_needed:
                     lines_cpp.append("if (m_%s__ && m_%s__) {  %s&%s%s; }" %(self.MEMBER_PARSE, self.MEMBER_SWAP, length_type.swap_prefix, param_length, length_type.swap_suffix))
-                lines_cpp.append("m_%s_idx__ = %s;" % (param_name, param_length))
+                if TypeInfo(param_type).type == TypeInfo.CLASS:
```
arnout, line 821: I think it would be better to merge this condition with the one immediately below. That would make is slightly less non-obvious why we initialise it to 0 for the class and to the length otherwise, because you can see that for the class the create function gets called.

framework/tlvf/tlvf.py
```diff
@@ -1030,15 +1033,15 @@ def addClassVarLenMethods(self, obj_meta, param_type, param_name, param_meta, pa
 
             lines_cpp.append("%suint8_t *src = (uint8_t *)m_%s;" % (self.getIndentation(1), param_meta.name))
             if param_length != None:
-                lines_cpp.append("%sif (!m_parse__) {" %self.getIndentation(1))
-                lines_cpp.append("%sif (m_%s_idx__ > 0) {" % (self.getIndentation(2), param_name))
+                #lines_cpp.append("%sif (!m_parse__) {" %self.getIndentation(1))
```
arnout: Is there any reason to comment the line instead of just removing it?

Same a few lines below.

framework/tlvf/tlvf.py
```diff
@@ -1030,15 +1033,15 @@ def addClassVarLenMethods(self, obj_meta, param_type, param_name, param_meta, pa
...
                 ptr = "(uint8_t *)m_%s_vector[m_%s_idx__ - 1]->getBuffPtr()" %(param_meta.name,param_name)
-                lines_cpp.append("%ssrc = %s;" %(self.getIndentation(3), ptr))                    
-                lines_cpp.append("%s}" %self.getIndentation(2))
+                lines_cpp.append("%ssrc = %s;" %(self.getIndentation(2), ptr))                    
                 lines_cpp.append("%s}" %self.getIndentation(1))
+                #lines_cpp.append("%s}" %self.getIndentation(1))
 
             lines_cpp.append( "%sif (ptr->getStartBuffPtr() != src) {" % (self.getIndentation(1)) )
-            lines_cpp.append( '%sTLVF_LOG(ERROR) << "Received to entry pointer is different than expected (excepting the same pointer returned from add method)";' %  self.getIndentation(2) )
+            lines_cpp.append( '%sTLVF_LOG(ERROR) << "Received entry pointer is different than expected (excepting the same pointer returned from add method)";' %  self.getIndentation(2) )
```
arnout: If you're anyway fixing this grammar error, fix the vocabulary mistake as well: excepting -> expecting.

framework/tlvf/tlvf.py
```diff
@@ -1048,8 +1051,11 @@ def addClassVarLenMethods(self, obj_meta, param_type, param_name, param_meta, pa
             lines_cpp.append( "%s}" % self.getIndentation(1) )
 
             if param_length or is_dynamic_len:
+                if not is_dynamic_len:
```
arnout: This condition is only needed because in the the parsing of a dynamic list in init(), we also increment the `_idx__`. If we remove that increment in init() (which we did in an earlier commit for var lists), then we can do the increment unconditionally here and things are more understandable IMO.

framework/tlvf/test/main.cpp
```diff
@@ -167,7 +169,17 @@ int test_complex_list()
     }
 
     if (tlv4->var1()->var1() != 0xb11b) {
-        MAPF_ERR("Unexpected var1 value" << tlv4->var1()->var1());
+        MAPF_ERR("Unexpected var1 value " << tlv4->var1()->var1());
+        ++errors;
+    }
+
+    if (tlv4->var2() != 0xabababab) {
+        MAPF_ERR("Unexpected var2 value " << tlv4->var2());
```
arnout, line 177: This should have been done in the commit that added var2 - but OK, not worth changing it now.

framework/tlvf/yaml/tlvf/test/tlvVarList.yaml
```diff
@@ -25,16 +25,25 @@ tlvTestVarList:
...
+    _value_const: 1
+  # TODO tlvf currently requires length to be second field
+  length: uint16_t
   # TODO tlvf currently doesn't support list length field of more than 1 byte
   list_length: uint8_t
   list:
     _type: uint8_t
     _length: [ list_length ]
   var1: uint32_t
+  unknown_length_list_inner:
```
arnout, line 47: Since you're not checking the values in the test, maybe it's worth giving a a _value or _value_const so we at least see something in the hexdump.

arnout (approved):
Grmbl, github messed up the order of the commits again...

Only nitpick comments so approving.

The commit messages are really excellent, especially the one of the second commit.

For the final commit's message: in tlvf, it is actually called "dynamic list", not "unkown length list". Although I agree that "unknown length list" is maybe a more descriptive name, perhaps it's better to stick to dynamic list in the commit message.

framework/tlvf/tlvf.py
tomereli: you're right, forgot to remove it - will do

framework/tlvf/tlvf.py
tomereli: Will do

framework/tlvf/tlvf.py
tomereli: done

framework/tlvf/tlvf.py
tomereli: Spot on!

framework/tlvf/yaml/tlvf/test/tlvVarList.yaml
tomereli: Fixed by setting value and verifying it

.travis.yml
```diff
@@ -17,4 +17,6 @@ script:
   - git update-index -q --refresh
   - git diff-index --exit-code HEAD
   # Check that Gw only test passes (full test not supported yet in travis)
-  - sudo ./tests/test_gw_repeater.sh --rm --gateway-only 
+  - sudo ./tests/test_gw_repeater.sh --rm --gateway-only
+  # Check tlvf test
+  - ./run.sh --entrypoint "$PWD/../../../build/install/bin/tests/tlvf_test"
```
arnout, line 22: I apparently missed this commit during my review, because I would have suggested a different approach: add a test target to maptools, and execute it by default when building. The unit tests take just a few seconds, so there is no reason not to run them I think.

PR #149, opened 2019-08-08
Use dockerhub in CI

arnout:
We have a dockerhub organisation for prplfoundation, with repositories for prplmesh-builder and prplmesh-runner.

@tomereli and @arnout have push access to these repositories.

Add a script to pull from there instead of building the images.

Also fix an issue introduced by @arnout when he added the -n/--native option.

The USER and WORKDIR were encoded in the builder image. Because of that, it was not universally usable.
Move those options to the docker run call, so the values are appropriate for the context in which the image is used.

Ubuntu 18.04 images have already been pushed to dockerhub, so Travis should work right away.

@tomereli may want to push Ubuntu 19.04 images as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,8 @@ services:
 
 script:
   - cd tools/docker
-  - sudo ./image-build.sh
+  # Try to pull from dockerhub, build the container images if that fails
+  - sudo ./image-pull.sh || sudo ./image-build.sh
   # Run clang-format and check there are no modified files by checking git status
   - sudo ./clang-format.sh
   - git diff-index --exit-code HEAD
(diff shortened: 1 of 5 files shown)


tomereli (approved):
Only small type in the last commit message - "rebuilding the all the time" --> "rebuilding them all the time"

tomereli:
> We have a dockerhub organisation for prplfoundation, with repositories for prplmesh-builder and prplmesh-runner.
> 
> @tomereli and @arnout have push access to these repositories.
> 
> Add a script to pull from there instead of building the images.
> 
> Also fix an issue introduced by @arnout when he added the -n/--native option.
> 
> The USER and WORKDIR were encoded in the builder image. Because of that, it was not universally usable.
> Move those options to the docker run call, so the values are appropriate for the context in which the image is used.
> 
> Ubuntu 18.04 images have already been pushed to dockerhub, so Travis should work right away.
> 
> @tomereli may want to push Ubuntu 19.04 images as well.

Pushed

PR #151, opened 2019-08-08
docker: fix the MAC address inside the containers

arnout:
A predictable MAC address makes testing and debugging easier.

In addition, if the MAC address of the bridge is not set explicitly, it is taken from one of the interfaces. This may happen to be a wlan interface, and in that
case the controller doesn't accept it as a slave due to an address conflict with the bridge.


arnout:
Note that the docker images must be regenerated after this is merged, otherwise the entrypoint is not compatible with the run script.

tomereli (approved):

PR #152, opened 2019-08-08
Fix gateway+repeater test

arnout:
The gateway+repeater test was sometimes failing in Travis, and it failed more or less consistently on my PC.

It turns out that this is a scheduling issue, because the repeater tries to join before the gateway agent has joined.

An extra delay fixes this, so we can enable the full gateway+repeater test again in Travis.


tomereli:
Wow!!! PR OF THE MONTH, congratulations! 🎉🎉🎉😃😃😃

tomereli (approved):

PR #153, opened 2019-08-11
docker: tests: test_gw_repeater support extra args

tomereli:
Populate remaining agruments comming after `--` to the called run.sh
script.

For example, running the following will pass --tag argument to the
run.sh script which will then use it for finding the docker image to
use:
./test_gw_repeater.sh --rm -- --tag ubuntu19-04

### DIFF
--- a/tools/docker/tests/test_gw_repeater.sh
+++ b/tools/docker/tests/test_gw_repeater.sh
@@ -39,11 +39,11 @@ usage() {
     echo "  options:"
     echo "      -h|--help - show this help menu"
     echo "      -v|--verbose - verbosity on"
-    echo "      -d|--delay - delay between starting the containers 
+    echo "      -d|--delay - delay between starting the containers
     and testing their status (default %DELAY seconds"
     echo "      -g|--gateway - gateway container name"
     echo "      -r|--repeater - repeater container name"
-    echo "      --rm - remove containers after test completes" 
+    echo "      --rm - remove containers after test completes"
 }
 
 main() {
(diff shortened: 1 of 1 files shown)


tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,7 +66,7 @@ main() {
             * ) err "unsupported argument $1"; usage; exit 1 ;;
         esac
     done
-
+
```
arnout: Whitespace error introduced here.

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -76,25 +76,25 @@ main() {
     dbg DELAY=$DELAY
 
     status "Start GW (Controller + local Agent)"
-    ${scriptdir}/../run.sh ${OPT} start-controller-agent -d -n ${GW_NAME} -m 00:11:22:33
+    ${scriptdir}/../run.sh ${OPT} start-controller-agent -d -n ${GW_NAME} -m 00:11:22:33 $@
```
arnout: I think this should be `"$@"`.

In general, i the quotes are not needed, I prefer to use `$*` instead (the two are equivalent when it's not quoted). Since `$*` is less used, it draws attention to the missing quotes.

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -76,25 +76,25 @@ main() {
...
     [ -z $GATEWAY_ONLY ] && {
         sleep ${DELAY}
         status "Start Repeater (Remote Agent)"
-        ${scriptdir}/../run.sh ${OPT} start-agent -d -n ${REPEATER_NAME} -m aa:bb:cc:dd
+        ${scriptdir}/../run.sh ${OPT} start-agent -d -n ${REPEATER_NAME} -m aa:bb:cc:dd $@
```
arnout: Same here.

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -76,25 +76,25 @@ main() {
...
     }
     
     status "Delay ${DELAY} seconds..."
     sleep ${DELAY}
     
     error=0
     report "GW operational" \
-        ${scriptdir}/../test.sh ${OPT} -n ${GW_NAME}
+        ${scriptdir}/../test.sh ${OPT} -n ${GW_NAME} $@
```
arnout: The options after `--` are options understood by run.sh. test.sh only understands the -n and -v options, and both of them are already passed, so I don't think it makes any sense to pass `$@` here. In fact, it will trigger errors from test.sh.

arnout (changes requested):
Typo in commit message: agruments -> arguments

Don't bother with a fixup commit, just rebase right away.

tools/docker/tests/test_gw_repeater.sh
tomereli: Will fix all whitespace errors in this file while doing so, is that ok?

arnout (approved):

Issue #126, opened 2019-07-25
[TASK] Autoconfig phase 2 - encryption support


tomereli:
## Description

Encryption support for autoconfig WSC M1+M2 in prplMesh.
The goal is to add the encryption parameters for M1 and M2 WSC messages, add encryption support to the framework and use it to encrypt the attributes of the encrypted settings in the M2 message which contains the configuration for the radio agent.
On receive, the radio agent will authenticate the message using the authenticator attribute in the M2 message, and then decrypt the encrypted settings.

### Encryption

- [x] Add Diffie-Hellman Key exchange
- [x] Add function for generating nonce
- [x] Add function for sha256
- [x] Add function for AES encryption
- [x] Add function for sha256/AES HMAC
- [x] Compute authentication and key wrap keys according to WSC spec
- [x] Perform HMAC on encrypted settings to generate authenticator and key_wrap_auth
- [x] Apply the above to M2 frame
- [x] Authenticate and decrypt the data in the agent to verify the above

### TLVF support for encryption

The M2 message includes multiple strings, which are implemented in the TLVF using variable-length lists. Support for multiple variable-length lists has recently been added, but there is no support class members or variable-length lists class members.
This is required for the encrypted settings in the M2 TLV since it contains multiple attributes and a length field, so it has to be a TLV class (although not a real TLV).
PR #91 (WIP) adds support for dynamic length list of classes for this purpose and actually is working. It was considered a valid workaround since it was thought that the encrypted settings are the last member of the WSC M2 TLV, but apparently, it is not - since the last attribute is the validator attribute.

The solution has to be adding support for class members - which should involves adding `memmov` support like done in #115 and #116 (the reason we can't use variable length list of classes of size 1 is that it will include the length field which is not part of the M2 frame).

**TODO - open a new task for this**

- [x] Add support for class members

## Exit criteria

- [x] Controller and remote agent pass m1/m2 and get to operational
- [x] Agent decryptes the encrypted settings and prints them. Compare they match what was set in the controller. Remove the prints afterwards
- [x] Autoconfig packets captured in Wireshark and are valid
- [x] Compare to Arris autoconfig Wireshark capture

## Detailed description
Section 5.7.1 in [beerocks_1.4 architecture.pdf](https://github.com/prplfoundation/prplMesh/blob/master/documentation/BeeRocks_v1_4_Architecture.pdf)

- **Problem statement** - Passing easymesh certification
[...]

PR #154, opened 2019-08-11
autoconfig: wsc encryption/decryption support

tomereli:
Issue #126 

Encrypt/decrypt and authenticate the WSC M2 frame.

Steps:
- [x] Add Diffie-Hellman key exchange
- [x] Add function for generating nonce
- [x] Add function for sha256
- [x] Add function for AES encryption
- [x] Add function for sha256/AES HMAC
- [x] Compute authentication and key wrap keys according to WSC spec
- [ ] Perform HMAC on encrypted settings to generate authenticator and key_wrap_auth
- [x] Perform AES on encrypted settings
- [x] Apply the above to M2 frame

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4269,6 +4269,35 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
                                   backhaul_params.bridge_mac);
 }
 
+/**
+ * @brief Diffie-Hellman public key exchange keys calculation
+ *        class member params authkey and keywrapauth are computed
+ *        on success.
+ *
+ * @param[in] m1 WSC M1 TLV received from the radio agent
+ * @param[out] authkey 32 bytes calculated authentication key
+ * @param[out] keywrapkey 16 bytes calculated key wrap key
+ * @return true on success
+ * @return false on failure
+ */
+bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                                 uint8_t authkey[32], uint8_t keywrapkey[16])
+{
+    auto mac = network_utils::mac_from_string(backhaul_params.bridge_mac);
+    if (!dh) {
+        LOG(ERROR) << "diffie hellman member not initialized";
+        return false;
+    }
+    if (!mapf::encryption::wps_calculate_keys(
+            *dh, m2->public_key_attr().data, m2->public_key_attr().data_length, dh->nonce(),
+            mac.oct, m2->registrar_nonce_attr().data, authkey, keywrapkey)) {
+        LOG(ERROR) << "Failed to calculate WPS keys";
+        return false;
+    }
+
+    return true;
+}
+
 bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     // Check if this is a M1 message that we sent to the controller, which was just looped back.
@@ -4323,21 +4352,32 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
     }
 
     for (auto tlvWscM2 : m2_list) {
+        uint8_t authkey[32];
+        uint8_t keywrapkey[16];
+        LOG(DEBUG) << "M2 Parse: calculate keys";
+        if (!autoconfig_wsc_calculate_keys(tlvWscM2, authkey, keywrapkey))
+            return false;
+
         auto encrypted_settings = tlvWscM2->encrypted_settings();
-        auto ssid  = std::string(encrypted_settings->ssid(), encrypted_settings->ssid_length());
-        auto bssid = network_utils::mac_to_string(encrypted_settings->bssid_attr().data);
-        auto authentication_type = encrypted_settings->authentication_type_attr().data;
-        auto encryption_type     = encrypted_settings->encryption_type_attr().data;
-        // TODO - finish with decryption and authentication
-        // https://github.com/prplfoundation/prplMesh/pull/91
-        // char network_key[WSC::WSC_MAX_NETWORK_KEY_LENGTH];
-        // std::copy_n(encrypted_settings.network_key_attr().data,
-        //             encrypted_settings.network_key_attr().data_length,
-        //             network_key);
-        // char key_wrap_auth[WSC::WSC_KEY_WRAP_AUTH_LENGTH];
-        // std::copy_n(encrypted_settings.key_wrap_auth_attr().data,
-        //             encrypted_settings.key_wrap_auth_attr().data_length,
-        //             network_key);
+        uint8_t *iv             = reinterpret_cast<uint8_t *>(encrypted_settings->iv());
+        LOG(DEBUG) << "M2 Parse: aes decrypt";
+        mapf::encryption::aes_decrypt(keywrapkey, iv,
+                                      (uint8_t *)encrypted_settings->encrypted_settings(),
+                                      encrypted_settings->encrypted_settings_length());
+
+        // TODO authenticate encrypted settings
+        // before passing to the config_data constructor
+        LOG(DEBUG) << "M2 Parse: parse config_data, len = "
+                   << sizeof(encrypted_settings->encrypted_settings_length());
+        WSC::cConfigData config_data(
+            reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings()),
+            encrypted_settings->encrypted_settings_length(), true, true);
+
+        auto ssid                = std::string(config_data.ssid(), config_data.ssid_length());
+        auto bssid               = network_utils::mac_to_string(config_data.bssid_attr().data);
+        auto authentication_type = config_data.authentication_type_attr().data;
+        auto encryption_type     = config_data.encryption_type_attr().data;
+
         std::string manufacturer =
             std::string(tlvWscM2->manufacturer(), tlvWscM2->manufacturer_length());
         // ONLY FOR DEBUG!!!
@@ -4766,16 +4806,18 @@ bool slave_thread::autoconfig_wsc_add_m1()
     if (!tlvWscM1->set_device_name("prplMesh-agent"))
         return false;
     std::memset(tlvWscM1->uuid_e_attr().data, 0xff, tlvWscM1->uuid_e_attr().data_length);
-    tlvWscM1->authentication_type_flags_attr().data = WSC::WSC_AUTH_OPEN | WSC::WSC_AUTH_WPA2;
-    tlvWscM1->encryption_type_flags_attr().data     = WSC::WSC_ENCR_NONE;
     tlvWscM1->rf_bands_attr().data =
         hostap_params.iface_is_5ghz ? WSC::WSC_RF_BAND_5GHZ : WSC::WSC_RF_BAND_2GHZ;
     // Simulate that this radio supports both fronthaul and backhaul BSS
     tlvWscM1->vendor_extensions_attr().subelement_value  = WSC::FRONTHAUL_BSS | WSC::BACKHAUL_BSS;
     tlvWscM1->primary_device_type_attr().sub_category_id = WSC::WSC_DEV_NETWORK_INFRA_AP;
-    // TODO: M1 should also have values for:
-    // enrolee_nonce_attr -> to be added by encryption
-    // public_key_attr -> to be added by encryption
+
+    // encryption support - diffie-helman key-exchange
+    dh = std::make_unique<mapf::encryption::diffie_hellman>();
+    std::copy(dh->pubkey(), dh->pubkey() + dh->pubkey_length(), tlvWscM1->public_key_attr().data);
+    std::copy(dh->nonce(), dh->nonce() + dh->nonce_length(), tlvWscM1->enrolee_nonce_attr().data);
+    tlvWscM1->authentication_type_flags_attr().data = WSC::WSC_AUTH_OPEN | WSC::WSC_AUTH_WPA2PSK;
+    tlvWscM1->encryption_type_flags_attr().data     = WSC::WSC_ENCR_AES;
 
     return true;
 }
--- a/agent/src/beerocks/slave/son_slave_thread.h
+++ b/agent/src/beerocks/slave/son_slave_thread.h
@@ -11,10 +11,15 @@
 
 #include "ap_manager_thread.h"
 
+#include <beerocks/bcl/beerocks_backport.h>
 #include <beerocks/bcl/beerocks_logging.h>
 #include <beerocks/bcl/beerocks_socket_thread.h>
 
 #include <beerocks/tlvf/beerocks_message_header.h>
+
+#include <mapf/common/encryption.h>
+#include <tlvf/ieee_1905_1/tlvWscM1.h>
+#include <tlvf/ieee_1905_1/tlvWscM2.h>
 
 namespace son {
 class slave_thread : public beerocks::socket_thread {
@@ -256,6 +261,17 @@ private:
     beerocks::eRadioStatus iface_status_bh_wired_prev = beerocks::eRadioStatus::INVALID;
     bool iface_status_operational_state               = false;
     bool iface_status_operational_state_prev          = false;
+
+    // Encryption support - move to common library
+    bool autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                       uint8_t authkey[32], uint8_t keywrapkey[16]);
+    bool autoconfig_wsc_parse_m2_encrypted_settings(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                                    uint8_t authkey[32], uint8_t keywrapkey[16],
+                                                    std::string &ssid, sMacAddr &bssid,
+                                                    WSC::eWscAuth &auth_type,
+                                                    WSC::eWscEncr &encr_type);
+
+    std::unique_ptr<mapf::encryption::diffie_hellman> dh = nullptr;
 
     bool parse_intel_join_response(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
     bool handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -373,6 +373,97 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
     return son_actions::send_cmdu_to_agent(sd, cmdu_tx);
 }
 
+/**
+ * @brief Encrypt the config data using AES and add to the WSC M2 TLV
+ *        The encrypted data length is the config data length padded to 16 bytes boundary.
+ *
+ * @param[in] m2 WSC M2 TLV
+ * @param[in] config_data config data in network byte order (swapped)
+ * @param[in] authkey 32 bytes calculated authentication key
+ * @param[in] keywrapkey 16 bytes calculated key wrap key
+ * @return true on success
+ * @return false on failure
+ */
+bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
+    std::shared_ptr<ieee1905_1::tlvWscM2> m2, WSC::cConfigData &config_data, uint8_t authkey[32],
+    uint8_t keywrapkey[16])
+{
+    size_t len = (config_data.getLen() + 15) & ~0xFU; // Align to 16 bytes boundary
+
+    auto encrypted_settings = m2->create_encrypted_settings();
+    if (!encrypted_settings)
+        return false;
+    if (!encrypted_settings->alloc_encrypted_settings(len))
+        return false;
+    if (!m2->add_encrypted_settings(encrypted_settings))
+        return false;
+
+    std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(),
+                encrypted_settings->encrypted_settings());
+    uint8_t *iv = reinterpret_cast<uint8_t *>(m2->encrypted_settings()->iv());
+
+    if (!mapf::encryption::create_iv(iv, WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH))
+        return false;
+
+    if (!mapf::encryption::aes_encrypt(
+            keywrapkey, iv, reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings()),
+            encrypted_settings->encrypted_settings_length())) {
+        LOG(DEBUG) << "aes encrypt";
+        return false;
+    }
+
+    return true;
+}
+
+/**
+ * @brief Calculate keys and update M2 attributes.
+ *
+ * @param[in] m1 WSC M1 TLV received from the radio agent
+ * @param[in] m2 WSC M2 TLV to be sent to the radio agent
+ * @param[in] dh diffie helman key exchange class containing the keypair
+ * @param[out] authkey 32 bytes calculated authentication key
+ * @param[out] keywrapkey 16 bytes calculated key wrap key
+ * @return true on success
+ * @return false on failure
+ */
+bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM1> m1,
+                                                  std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                                  const mapf::encryption::diffie_hellman &dh,
+                                                  uint8_t authkey[32], uint8_t keywrapkey[16])
+{
+    m2->authentication_type_flags_attr().data = WSC::WSC_AUTH_OPEN | WSC::WSC_AUTH_WPA2PSK;
+    m2->encryption_type_flags_attr().data     = WSC::WSC_ENCR_NONE | WSC::WSC_ENCR_AES;
+    std::copy_n(m1->enrolee_nonce_attr().data, m1->enrolee_nonce_attr().data_length,
+                m2->enrolee_nonce_attr().data);
+    std::copy_n(dh.nonce(), dh.nonce_length(), m2->registrar_nonce_attr().data);
+    if (!mapf::encryption::wps_calculate_keys(
+            dh, m1->public_key_attr().data, m1->public_key_attr().data_length,
+            m1->enrolee_nonce_attr().data, m1->mac_attr().data.oct, m2->registrar_nonce_attr().data,
+            authkey, keywrapkey)) {
+        LOG(ERROR) << "Failed to calculate WPS keys";
+        return false;
+    }
+    std::copy(dh.pubkey(), dh.pubkey() + dh.pubkey_length(), m2->public_key_attr().data);
+
+    return true;
+}
+
+/**
+ * @brief add WSC M2 TLV to the current CMDU
+ *
+ *        the config_data contains the secret ssid, authentication and encryption types,
+ *        the network key, bssid and the key_wrap_auth attribute.
+ *        It does encryption using the keywrapkey and HMAC with the authkey generated
+ *        in the WSC keys calculation from the M1 and M2 nonce values, the radio agent's
+ *        mac, and a random initialization vector.
+ *        The encrypted config_data blob is copied to the encrypted_data attribute
+ *        in the M2 TLV, which marks the WSC M2 TLV ready to be sent to the agent.
+ *
+ * @param tlvWscM1 WSC M1 TLV received from the radio agent as part of the WSC autoconfiguration
+ *        CMDU
+ * @return true on success
+ * @return false on failure
+ */
 bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1> tlvWscM1)
 {
     if (!tlvWscM1) {
@@ -397,9 +488,6 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1> 
     if (!tlvWscM2->set_serial_number("prpl12345"))
         return false;
     std::memset(tlvWscM2->uuid_r_attr().data, 0xee, tlvWscM2->uuid_r_attr().data_length);
-    tlvWscM2->authentication_type_flags_attr().data =
-        tlvWscM1->authentication_type_flags_attr().data;
-    tlvWscM2->encryption_type_flags_attr().data = tlvWscM1->encryption_type_flags_attr().data;
     tlvWscM2->rf_bands_attr().data = (tlvWscM1->rf_bands_attr().data & WSC::WSC_RF_BAND_5GHZ)
                                          ? WSC::WSC_RF_BAND_5GHZ
                                          : WSC::WSC_RF_BAND_2GHZ;
@@ -408,30 +496,45 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1> 
             ? WSC::FRONTHAUL_BSS
             : (bss_type & WSC::BACKHAUL_BSS ? WSC::BACKHAUL_BSS : WSC::TEARDOWN);
     tlvWscM2->primary_device_type_attr().sub_category_id = WSC::WSC_DEV_NETWORK_INFRA_GATEWAY;
-    // Encrypted settings
-    auto encrypted_settings = tlvWscM2->create_encrypted_settings();
-    if (!encrypted_settings) {
-        LOG(ERROR) << "Failed to create encrypted settings";
-        return false;
-    }
-    // TODO - actually encrypt the data
-    encrypted_settings->set_ssid("prplMesh_ssid");
-    encrypted_settings->authentication_type_attr().data = WSC::WSC_AUTH_WPA2; //DUMMY
-    encrypted_settings->encryption_type_attr().data     = WSC::WSC_ENCR_AES;
-    std::fill(encrypted_settings->network_key_attr().data,
-              encrypted_settings->network_key_attr().data +
-                  encrypted_settings->network_key_attr().data_length,
-              0xaa); //DUMMY
-    encrypted_settings->bssid_attr().data = tlvWscM1->mac_attr().data;
-    std::fill(encrypted_settings->key_wrap_auth_attr().data,
-              encrypted_settings->key_wrap_auth_attr().data +
-                  encrypted_settings->key_wrap_auth_attr().data_length,
-              0xff); //DUMMY
 
-    if (!tlvWscM2->add_encrypted_settings(encrypted_settings)) {
-        LOG(ERROR) << "Failed to add encrypted settings";
+    ///////////////////////////////
+    // @brief encryption support //
+    ///////////////////////////////
+    mapf::encryption::diffie_hellman dh;
+    uint8_t authkey[32];
+    uint8_t keywrapkey[16];
+    if (!autoconfig_wsc_calculate_keys(tlvWscM1, tlvWscM2, dh, authkey, keywrapkey))
         return false;
-    }
+
+    // Encrypted settings
+    // Encrypted settings are the ConfigData + IV. First create the ConfigData,
+    // Then copy it to the encrypted data, add an IV and encrypt.
+    // Finally, add HMAC
+
+    // Create ConfigData
+    uint8_t buf[1024];
+    WSC::cConfigData config_data(buf, sizeof(buf), false, true);
+    config_data.set_ssid("prplMesh-ssid");
+    config_data.authentication_type_attr().data = WSC::WSC_AUTH_WPA2;
+    config_data.encryption_type_attr().data     = WSC::WSC_ENCR_AES;
+    std::fill(config_data.network_key_attr().data,
+              config_data.network_key_attr().data + config_data.network_key_attr().data_length,
+              0xaa); //DUMMY
+
+    LOG(DEBUG) << "WSC config_data:" << std::endl
+               << "     ssid: " << config_data.ssid() << std::endl
+               << "     authentication_type: " << config_data.authentication_type_attr().data
+               << std::endl
+               << "     encryption_type: " << config_data.encryption_type_attr().data << std::endl;
+
+    config_data.class_swap();
+
+    if (!autoconfig_wsc_add_m2_encrypted_settings(tlvWscM2, config_data, authkey, keywrapkey))
+        return false;
+
+    // TODO - add HMAC and authenticator
+    std::fill(tlvWscM2->authenticator().data,
+              tlvWscM2->authenticator().data + tlvWscM2->authenticator().data_length, 0xbb);
 
     return true;
 }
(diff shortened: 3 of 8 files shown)


agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -257,6 +261,16 @@ class slave_thread : public beerocks::socket_thread {
     bool iface_status_operational_state               = false;
     bool iface_status_operational_state_prev          = false;
 
+    // Encryption support - move to common library
+    bool autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM2> m2);
+    bool autoconfig_wsc_parse_m2_encrypted_settings(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                                    std::string &ssid, std::string &bssid,
+                                                    WSC::eWscAuth &auth_type,
+                                                    WSC::eWscEncr &encr_type);
+    mapf::encryption::diffie_hellman dh;
```
arnout: This causes the keypair to be generated implicitly in the slave_thread constructor.

Athough that is not much of a problem per se, it has two implications:

- It unnecessarily slows down start-up of the slave thread - the keypair is only needed much later, so this could be done asynchronously. I'm not sure if this is significant though - it may depend on available entropy (which is high in docker, but may be more problematic on a real router).
- The same key will be reused for any subsequent autoconfig. Key reuse in general is a red flag for security. For this particular use case, it is probably not critical because there is no easy way for an attacker to trigger autoconfig (unless they can already impersonate the controller, but then the game is lost anyway). Still, it's probably better to follow best practices.

There are three ways that these problems can be solved:
- Move the key generation from dh constructor to an explicit init function. This is a bit problematic because we'd have to add checks that the init function has been called. 
- Move the key generation to compute_key(). However, I'm not sure if we need to use the same DH keypair for the controller if we send multiple M2 (and each M2 definitely has its own call to compute_key()).
- Make `dh` a pointer instead of a direct member so we can replace it.

It may be a bit too much work to do this now, so it's OK to create an issue for it instead.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4266,6 +4269,82 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+ *        class member params authkey and keywrapauth are computed
+ *        on success.
+ *
+ * @param m1 WSC M1 TLV received from the radio agent
+ * @return true on success
+ * @return false on failure
+ */
+bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM2> m2)
+{
+    auto mac = network_utils::mac_from_string(backhaul_params.bridge_mac);
```
arnout, line 4286: Not for this PR, but: I think we really should have an explicit field for the AL-MAC instead of relying on it being equal to the bridge MAC (which is BTW almost certainly wrong, because the AL-MAC must be a locally administered address while for the bridge MAC you typically want an OUI-based address).

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -257,6 +261,16 @@ class slave_thread : public beerocks::socket_thread {
...
+    uint8_t authkey[32];
+    uint8_t keywrapkey[16];
```
arnout: These two have a very short valid lifetime: they are the authkey and keywrapkey of the last received M2, and they are only valid for that specific M2.

Therefore, I don't think it makes much sense to have them as members of the slave_thread object. It think it would be more appropriate as local variables in handle_autoconfiguration_wsc() that get passed to autoconfig_wsc_calculate_keys() and autoconfig_wsc_parse_m2_encrypted_settings().

(nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4266,6 +4269,82 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+
+    return true;
+}
+
+bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
+    std::shared_ptr<ieee1905_1::tlvWscM2> m2, std::string &ssid, std::string &bssid,
+    WSC::eWscAuth &auth_type, WSC::eWscEncr &encr_type)
+{
+    auto encrypted_settings = m2->encrypted_settings();
+    uint8_t *iv             = reinterpret_cast<uint8_t *>(encrypted_settings->iv());
```
arnout: Why is iv declared as char in the yaml file? Is it because a limitation of the type for dynamic lists? Does that limitation still apply?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4266,6 +4269,82 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+    LOG(DEBUG) << "M2 Parse: aes decrypt";
+    mapf::encryption::aes_decrypt(keywrapkey, iv,
+                                  (uint8_t *)encrypted_settings->encrypted_settings(),
```
arnout: Why do you use C-style cast here instead of reinterpret_cast like above? (Obviously, this becomes irrelevant if encrypted_settings() return uint8_t like it should).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4266,6 +4269,82 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+    LOG(DEBUG) << "M2 Parse: parse config_data, len = " << sizeof(buf);
+    // Need to convert to host byte order to get config data length,
+    // which is needed for computing the KWA (1st 64 bits of HMAC)
+    // However, the KWA is computed on the finalized version of the config
+    // data, hence the class_swap.
+    // Finally, once authentication succeeds, swap back to host byte order.
+    // Theoretically, this can be avoided by not passing swap_needed=True to
+    // cConfigData constructor, but then parsing will fail since the length
+    // will be calculated wrong. TLVF does not support parsing network byte order
+    // without full swap, so keep this workaround for now (another future TLVF V2 feature)
```
arnout: Excellent comment!

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4266,6 +4269,82 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+    WSC::cConfigData config_data(buf, sizeof(buf), true, true);
+
+    // get length of config_data for KWA authentication
+    size_t len = config_data.getLen();
+
+    // Swap to network byte order for KWA HMAC calculation
+    config_data.class_swap();
+
+    uint8_t kwa[8];
```
arnout: It would be a bit more generic to use WSC_AUTHENTICATOR_LENGTH instead of 8 (nitpick).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4266,6 +4269,82 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+    }
+    if (!std::equal(kwa, kwa + sizeof(kwa), (uint8_t *)m2->authenticator().data)) {
+        LOG(ERROR) << "Authentication failure";
+        return false;
+    }
+
+    // Swap back to host byte order to read and use config_data
+    config_data.class_swap();
+    ssid      = std::string(config_data.ssid(), config_data.ssid_length());
+    bssid     = network_utils::mac_to_string(config_data.bssid_attr().data);
```
arnout: It would make more sense to pass this around as a cMacAddress and only convert where needed (nitpick).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4758,16 +4835,18 @@ bool slave_thread::autoconfig_wsc_add_m1()
...
     tlvWscM1->vendor_extensions_attr().subelement_value  = WSC::FRONTHAUL_BSS | WSC::BACKHAUL_BSS;
     tlvWscM1->primary_device_type_attr().sub_category_id = WSC::WSC_DEV_NETWORK_INFRA_AP;
-    // TODO: M1 should also have values for:
-    // enrolee_nonce_attr -> to be added by encryption
-    // public_key_attr -> to be added by encryption
+
+    // encryption support - diffie-helman key-exchange
+    std::copy(dh.pubkey(), dh.pubkey() + dh.pubkey_length(), tlvWscM1->public_key_attr().data);
+    mapf::encryption::create_nonce(tlvWscM1->enrolee_nonce_attr().data,
+                                   tlvWscM1->enrolee_nonce_attr().data_length);
```
arnout: We *must* also store the nonce locally and compare it when handling M2. Otherwise it leaves the door wide open for replay attacks.

I propose to make the nonce a member of the diffie_hellman class, and pass it at construction/init time. It could be optional: the controller would pass the nonce from M1 at construction time, while the agent would leave it empty to have one generated and copy out the nonce when creating M1.

Anyway, this can be done in a later PR, but an issue should be created for it.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,94 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
     return son_actions::send_cmdu_to_agent(sd, cmdu_tx);
 }
 
+/**
+ * @brief Encrypt the config data using AES and add to the WSC M2 TLV
+ *        The enctypted data length is the config data length aligned to 16 bytes boundary.
```
arnout: typo encrypted

aligned -> padded

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,94 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+ * @param config_data config data in network byte order (swapped)
+ * @return true on success
+ * @return false on failure
+ */
+bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
+    std::shared_ptr<ieee1905_1::tlvWscM2> m2, WSC::cConfigData &config_data)
+{
+    // Calculate length of data to encrypt
+    // (= plaintext length + 32 bits HMAC aligned to 16 bytes boundary)
+    size_t len             = (config_data.getLen() + 8 + 16) & ~0xFU;
```
arnout: This formula does not correspond to the comment above...

Based on the comment, I would expect it to be:
```suggestion
    size_t len             = (config_data.getLen() + sizeof(sWscAttrAuthenticator) + 15) & ~0xFU;
```

I don't know which one is correct though.

Also, I think it's worth creating a function somewhere in common to do round_up_to_multiple(). So this would become:
```suggestion
    size_t len             = mapf::round_up_to_multiple(config_data.getLen() + sizeof(sWscAttrAuthenticator, 16);
```

Finally, it would be better to replace the 16 with a constant that is defined in mapf::encryption.

All of the above are probably (nitpick).

controller/src/beerocks/master/son_master_thread.h
```diff
@@ -58,14 +58,25 @@ class master_thread : public beerocks::socket_thread {
...
+                                                  WSC::cConfigData &config_data);
+    bool autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM1> m1,
+                                       std::shared_ptr<ieee1905_1::tlvWscM2> m2);
+
     db &database;
     task_pool tasks;
+
+    // encryption support - TODO move to common library
+    mapf::encryption::diffie_hellman dh;
+    uint8_t authkey[32];
```
arnout: Like for the slave, I think authkey and keywrapkey should be local variables instead of members.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,94 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+    if (!mapf::encryption::kwa_compute(authkey, encrypted, config_data.getLen(), kwa))
+        return false;
+    std::copy_n(kwa, m2->authenticator().data_length, m2->authenticator().data);
+
+    uint8_t *iv = reinterpret_cast<uint8_t *>(m2->encrypted_settings()->iv());
+
+    if (!mapf::encryption::create_nonce(iv, WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH))
+        return false;
+
+    if (m2->encryption_type_flags_attr().data != WSC::WSC_ENCR_AES) {
```
arnout: This is wrong... First of all, the encryption type flags in M2 indicate the encryption supported by the registrar (i.e. controller), not the enrollee (i.e. agent). Second, the encryption type flags indicate the encryption types supported for the EAP itself, *not* for WSC. For WSC, there are no options. It *must* be AES-128 (KeyWrapKey == 128bits).

In other words, this condition doesn't belong here.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,94 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+ * @param m1 WSC M1 TLV received from the radio agent
+ * @param m2 WSC M2 TLV to be sent to the radio agent
+ * @return true on success
+ * @return false on failure
+ */
+bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM1> m1,
+                                                  std::shared_ptr<ieee1905_1::tlvWscM2> m2)
+{
+    m2->authentication_type_flags_attr().data = m1->authentication_type_flags_attr().data;
+    m2->encryption_type_flags_attr().data     = m1->encryption_type_flags_attr().data;
```
arnout: As mentioned above, these two are not entirely correct: the M2 values should be the authentication and encryption types supported by the registrar. I don't really know what it's supposed to be used for though, because the registrar would already have filtered the actual settings based on the flags in the M1.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,94 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+    std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(), encrypted);
+    uint8_t *kwa = &encrypted[config_data.getLen()];
+    // Add KWA which is the 1st 64 bits of HMAC of config_data using AuthKey
+    if (!mapf::encryption::kwa_compute(authkey, encrypted, config_data.getLen(), kwa))
+        return false;
+    std::copy_n(kwa, m2->authenticator().data_length, m2->authenticator().data);
+
+    uint8_t *iv = reinterpret_cast<uint8_t *>(m2->encrypted_settings()->iv());
+
+    if (!mapf::encryption::create_nonce(iv, WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH))
```
arnout: This is a bit confusing, because an IV is not exactly a nonce... But it doesn't matter much so OK.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,94 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+        LOG(ERROR) << "Failed to calculate WPS keys";
+        return false;
+    }
+    std::copy(dh.pubkey(), dh.pubkey() + dh.pubkey_length(), m2->public_key_attr().data);
+
+    return true;
+}
+
+/**
+ * @brief TODO
```
arnout: Can you fill this in?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,94 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+    auto encrypted_settings = m2->create_encrypted_settings();
+    if (!encrypted_settings)
+        return false;
+    if (!encrypted_settings->alloc_encrypted_settings(len))
+        return false;
+    if (!m2->add_encrypted_settings(encrypted_settings))
+        return false;
+
+    std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(), encrypted);
+    uint8_t *kwa = &encrypted[config_data.getLen()];
```
arnout: This should actually be the key_wrap_auth member of config_data.

I think the best approach is to remove key_wrap_auth from config_data and add it directly to the buffer here. However, you do need the type and length fields of the attribute as well.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -408,30 +493,41 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
...
+    config_data.encryption_type_attr().data     = WSC::WSC_ENCR_AES;
+    std::fill(config_data.network_key_attr().data,
+              config_data.network_key_attr().data + config_data.network_key_attr().data_length,
               0xaa); //DUMMY
-    encrypted_settings->bssid_attr().data = tlvWscM1->mac_attr().data;
-    std::fill(encrypted_settings->key_wrap_auth_attr().data,
-              encrypted_settings->key_wrap_auth_attr().data +
-                  encrypted_settings->key_wrap_auth_attr().data_length,
+    std::fill(config_data.key_wrap_auth_attr().data,
+              config_data.key_wrap_auth_attr().data + config_data.key_wrap_auth_attr().data_length,
```
arnout: This is obviously wrong, key_wrap_auth should contain correct data when encryption is done.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -408,30 +493,41 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
...
               0xff); //DUMMY
 
-    if (!tlvWscM2->add_encrypted_settings(encrypted_settings)) {
-        LOG(ERROR) << "Failed to add encrypted settings";
+    // TODO - remove this print before merging!
```
arnout: Don't forget :-)

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -370,3 +363,19 @@ cWscAttrEncryptedSettings:
...
+  _type: class
+  _is_tlv_class: True
+  type:
+    _type: eWscAttributes
+    _value_const: ATTR_ENCR_SETTINGS
+  length:
+    _type: uint16_t
+    _length_var: True
+  iv:
+    _type: char
```
arnout, line 376: Why is this not uint8_t?

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -189,6 +190,88 @@ int test_complex_list()
     return errors;
 }
 
+bool add_encrypted_settings(std::shared_ptr<tlvWscM2> m2, uint8_t *keywrapkey, uint8_t *iv)
```
arnout, line 193: This function contains a copy of son::master_thread::autoconfig_wsc_add_m2_encrypted_settings(). That doesn't make much sense, because we actually want to test that function. So to make it testable, it should be moved out somehow.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -276,34 +359,34 @@ int test_all()
...
+    m2.compute_key(key2, key2_length, m1.pubkey(), m1.pubkey_length());
+    key1_length = sizeof(key1);
+    uint8_t m1_nonce[192];
+    uint8_t m2_nonce[192];
+    mapf::encryption::create_nonce(m1_nonce, sizeof(m1_nonce));
+    mapf::encryption::create_nonce(m2_nonce, sizeof(m2_nonce));
+    uint8_t mac[6] = {0x11, 0x22, 0x33, 0x44, 0x55, 0x66};
+    uint8_t authkey[32];
+    uint8_t keywrapkey[16];
+    uint8_t iv[16];
```
arnout, line 379: Again, there's a lot of duplication here from the actual implementation.

I think we really would want an extra class that does all this functionality, i.e. move it out of the master_thread and slave_thread. This would also solve part of the lifetime issue of the diffie_hellman class, and it could also store the nonce.

It would necessarily need to have access to the M1 and M2 definitions, so encryption.h/cpp is not the appropriate place. But then I'm not sure what is the appropriate place...

Note that this can be done in a future refactoring PR. In other words: (nitpick).

arnout (changes requested):
The commit messages have some weird wrapping. Standard wrapping is at 72 columns.

AFAICS the first three, maybe four commits are OK as they are. Could you open a separate PR for them so they can already be merged?

tomereli:
> The commit messages have some weird wrapping. Standard wrapping is at 72 columns.
> 
> AFAICS the first three, maybe four commits are OK as they are. Could you open a separate PR for them so they can already be merged?

github as we know messes up the commits order - so I'm not sure which commits you're referring to, so just to be sure - these are the first four commits - did you mean these?
> 71de33a - Arnout Vandecappelle (Essensium/Mind), 5 weeks ago : framework: common: encryption impl & unit test
8730832 - Tomer Eliyahu, 11 hours ago : autconfig: fix radio agent ruid comparison bug
2b7b2d2 - Tomer Eliyahu, 11 hours ago : autoconfig: set max vaps to 1 in radio agent

**edit**
It was a silly question, the other commits contain loads of (good) comments, so I know you meant these four:)
c5c3eae - Tomer Eliyahu, 10 hours ago : debug: add controller gdb configuration

PR opened - see #157

agent/src/beerocks/slave/son_slave_thread.h
tomereli: Creating a pointer is fine - not that much work, shall I go with this approach?

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Do you mean an explicit field in the DB node structure? If so, I agree - and I guess it can be done as part of the handling of none intel agents, correct?

agent/src/beerocks/slave/son_slave_thread.h
tomereli: Will fix.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: It is declared as char since the uint8_t doesn't have a convinient getter function and instead returns that weird tupple. Does it matter?

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: We should update dynamic list of uint8_t getter function to return the pointer to the buffer instead of a tuple with the first byte, but that's for a different PR. Opened a task for it - #156

tomereli: Will change the cast to `reinterpret_cast`

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: I'll do that as part of this PR

controller/src/beerocks/master/son_master_thread.cpp
tomereli: We can add another function which does the same thing in encryption.cpp, then move the implementation to be private and have create_nonce and create_iv call it.
So:
cpp file:
```
static bool create_random(uint8_t *buf, size_t buf_len)
{
    // Implementation...
}
bool create_nonce(uint8_t *buf, size_t buf_len) { create_random(buf, buf_len); }
bool create_iv(uint8_t *buf, size_t buf_len) { create_random(buf, buf_len); }
```
h file:
```
bool create_nonce(uint8_t *buf, size_t buf_len);
bool create_iv(uint8_t *buf, size_t buf_len);
```

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Yep, missed that somehow

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
tomereli: See #156

framework/tlvf/test/tlvf_test.cpp
tomereli: Perhaps we can move it to they tlvf none generated files?

framework/tlvf/test/tlvf_test.cpp
tomereli: We can move it to the TLVF as none generated files, like BaseClass.cpp.
Why not add autoconfig class in the framework as part of mapfcommon.so? It means that mapfcommon.so will need to link to tlvf.so, is that an issue?

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Fixed the encryption type, not sure about the authentication type - should it just be WPA2? we do support OPEN but shouldn't this attribute say which authentication type was used by the registrar (controller) in the M2?

controller/src/beerocks/master/son_master_thread.cpp
tomereli: wrapping in `#ifdef DEBUG` so we have it in debug builds, ok?

framework/tlvf/test/tlvf_test.cpp
tomereli: Anyway I will do that in a separate commit, after we finish the first review iteration.

framework/tlvf/test/tlvf_test.cpp
tomereli: will do that in a separate PR / commit, keeping discussion open so I don't forget.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: I would make a separate PR of that change. But I can take it, yes.

agent/src/beerocks/slave/son_slave_thread.h
arnout: > Creating a pointer is fine - not that much work, shall I go with this approach?

Go for it!

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: > It is declared as char since the uint8_t doesn't have a convinient getter function and instead returns that weird tupple. Does it matter?

OK, I see you created a task for that.

controller/src/beerocks/master/son_master_thread.cpp
arnout: > Fixed the encryption type, not sure about the authentication type - should it just be WPA2? we do support OPEN but shouldn't this attribute say which authentication type was used by the registrar (controller) in the M2?

The WSC specification says "Bit field of authentication types supported by the Registrar." I suppose we only support Open and WPA2 (not WEP or WPA1, and Enterprise would require an extension to how we build M2), so those should be put in there.

controller/src/beerocks/master/son_master_thread.cpp
arnout: Note sure if that is so much better, but it's definitely not worse :-)

controller/src/beerocks/master/son_master_thread.cpp
arnout: > wrapping in `#ifdef DEBUG` so we have it in debug builds, ok?

Then you can just leave it in regardless. It's not as if any secrets are printed here. And anyway, if there were, they would be known in the controller already.

framework/tlvf/test/tlvf_test.cpp
arnout: > We can move it to the TLVF as none generated files, like BaseClass.cpp.
> Why not add autoconfig class in the framework as part of mapfcommon.so? It means that mapfcommon.so will need to link to tlvf.so, is that an issue?

I thought tlvf.so already depended on mapfcommon, but apparently it doesn't so indeed we can put it in mapfcommon.

arnout:
> github as we know messes up the commits order - so I'm not sure which commits you're referring to, 

Indeed, for this reason I stopped looking at the commits on github, I just look at them directly with git and I do the review on all commits together.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -436,8 +431,8 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
 bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM1> m1,
                                                   std::shared_ptr<ieee1905_1::tlvWscM2> m2)
 {
-    m2->authentication_type_flags_attr().data = m1->authentication_type_flags_attr().data;
-    m2->encryption_type_flags_attr().data     = m1->encryption_type_flags_attr().data;
+    m2->authentication_type_flags_attr().data = WSC::WSC_AUTH_WPA2;
```
arnout: Should be `WSC_AUTH_OPEN | WSC_AUTH_WPA2PSK`. WPA2 is Enterprise.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -436,8 +431,8 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
+    m2->encryption_type_flags_attr().data     = WSC::WSC_ENCR_AES;
```
arnout: Should be `WSC_ENCR_NONE | WSC_ENCR_AES`.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -455,11 +450,19 @@ bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tl
 }
 
 /**
- * @brief TODO
+ * @brief add WSC M2 TLV to the current CMDU.
+ *        the config_data contains the secret ssid, authentication and encryption types,
```
arnout, line 454: Leave an empty line here, else the entire paragraph is part of the brief...

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -455,11 +450,19 @@ bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tl
...
+ *        the network key, bssid and the key_wrap_auth attribute.
+ *        It goes through AES encryption using the authkey and keywrapkey generated
```
arnout: It does encryption using the keywrapkey and HMAC with the authkey.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,105 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+        return false;
+    if (!m2->add_encrypted_settings(encrypted_settings))
+        return false;
+
+    std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(), encrypted);
+    uint8_t *kwa = &encrypted[config_data.getLen()];
+    // Add KWA which is the 1st 64 bits of HMAC of config_data using AuthKey
+    if (!mapf::encryption::kwa_compute(authkey, encrypted, config_data.getLen(), kwa))
+        return false;
+    std::copy_n(kwa, m2->authenticator().data_length, m2->authenticator().data);
```
arnout: I only notice now...

The authenticator that you need to fill in here is the key_wrap_authenticator, not the global authenticator. The global authenticator follows the same algorithm to compute it, but over the entire M1 + M2 (excluding the authenticator itself, obviously) instead of only the encrypted settings. Cfr. page 48 of the WSC spec: "M2 = Version || N1 || N2 || Description || PKR [ || ConfigData ] || HMACAuthKey(M1 || M2*)" (M2* == M2 excluding the authenticator itself).

So, obviously this shouldn't be done here, but in a separate step after M2 is completely filled in.

Also note that on the slave side, this means you need to keep a copy of M1 around to be able to calculate the hmac (you can't keep a partially calculated hmac because you don't have the shared secret yet at the time M1 is sent, and the authkey is mixed in with the message to get an HMAC).

arnout (changes requested):
The fixup! commit mixes fixups for several of the previous commits. E.g. the part in encryption.h/cpp should be squashed into the first commit of #157.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: I accidently did this as part of the previous commit fixup, i'll fix in during rebase, but Done.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4842,7 +4849,8 @@ bool slave_thread::autoconfig_wsc_add_m1()
     tlvWscM1->primary_device_type_attr().sub_category_id = WSC::WSC_DEV_NETWORK_INFRA_AP;
 
     // encryption support - diffie-helman key-exchange
-    std::copy(dh.pubkey(), dh.pubkey() + dh.pubkey_length(), tlvWscM1->public_key_attr().data);
+    dh = std::make_shared<mapf::encryption::diffie_hellman>();
```
arnout: make_unique would be more appropriate here IMO. dh is only used within this object, so it is not shared. Of course, the behaviour is the same, but unique_ptr is more explicit IMO.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Right - opened #158 for it, will fix in a separate PR.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4266,6 +4269,84 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+    // Swap to network byte order for KWA HMAC calculation
+    config_data.class_swap();
+
+    uint8_t kwa[8];
+    // Compute KWA based on decrypted settings
+    if (!mapf::encryption::kwa_compute(authkey, buf, len, kwa)) {
+        LOG(ERROR) << "kwa compute";
+        return false;
+    }
+    if (!std::equal(kwa, kwa + sizeof(kwa), (uint8_t *)m2->authenticator().data)) {
```
arnout: This has to be keywrapauth from config_data.

arnout (changes requested):
Authenticator still has to be fixed.

tomereli:
I think we can just merge this PR and close #154 once review is done, don't you?

controller/src/beerocks/master/son_master_thread.cpp
tomereli: will be fixed as part of #158

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4323,21 +4348,31 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
     }
 
     for (auto tlvWscM2 : m2_list) {
+        uint8_t authkey[32];
+        uint8_t keywrapkey[16];
+        if (!autoconfig_wsc_calculate_keys(tlvWscM2, authkey, keywrapkey))
+            return false;
+        LOG(DEBUG) << "M2 Parse: calculate keys";
```
arnout: It is weird to output this message *after* they've been calculated...

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4323,21 +4348,31 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
+        mapf::encryption::aes_decrypt(keywrapkey, iv,
+                                      (uint8_t *)encrypted_settings->encrypted_settings(),
+                                      encrypted_settings->encrypted_settings_length());
+
+        // TODO authenticate encrypted settings
+        // before passing to the config_data constructor
+        uint8_t buf[encrypted_settings->encrypted_settings_length()];
+        std::copy_n(encrypted_settings->encrypted_settings(),
+                    encrypted_settings->encrypted_settings_length(), buf);
+        LOG(DEBUG) << "M2 Parse: parse config_data, len = " << sizeof(buf);
```
arnout: Why do we need to make a copy here? We could just construct config_data with encrypted_settings() as buffer, no?

If we do need to make a copy, then actually it would be better to change aes_decrypt so it *doesn't* work in-place.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4323,21 +4348,31 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
     }
 
     for (auto tlvWscM2 : m2_list) {
+        uint8_t authkey[32];
+        uint8_t keywrapkey[16];
+        if (!autoconfig_wsc_calculate_keys(tlvWscM2, authkey, keywrapkey))
```
arnout, line 4358: Do we need a check/assert/ something that dh != nullptr?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4323,21 +4348,31 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
+        std::copy_n(encrypted_settings->encrypted_settings(),
+                    encrypted_settings->encrypted_settings_length(), buf);
+        LOG(DEBUG) << "M2 Parse: parse config_data, len = " << sizeof(buf);
+        WSC::cConfigData config_data(buf, sizeof(buf), true, true);
+
+        auto ssid                = std::string(config_data.ssid(), config_data.ssid_length());
+        auto bssid               = network_utils::mac_to_string(config_data.bssid_attr().data);
+        auto authentication_type = config_data.authentication_type_attr().data;
+        auto encryption_type     = config_data.encryption_type_attr().data;
+
```
arnout, line 4380: It's a great idea to split off the authentication in a separate commit :-)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,97 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+ * @param[in] authkey 32 bytes calculated authentication key
+ * @param[in] keywrapkey 16 bytes calculated key wrap key
+ * @return true on success
+ * @return false on failure
+ */
+bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
+    std::shared_ptr<ieee1905_1::tlvWscM2> m2, WSC::cConfigData &config_data, uint8_t authkey[32],
+    uint8_t keywrapkey[16])
+{
+    size_t len             = (config_data.getLen() + 16) & ~0xFU; // Align to 16 bytes boundary
```
arnout: I may have a problem with my math-fu, but align to 16 bytes should add just 15 to it, no? Otherwise, if it already is a multiple of 16, you add redundant padding.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -373,6 +373,97 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+    if (!m2->add_encrypted_settings(encrypted_settings))
+        return false;
+
+    std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(), encrypted);
+    uint8_t *iv = reinterpret_cast<uint8_t *>(m2->encrypted_settings()->iv());
+
+    if (!mapf::encryption::create_iv(iv, WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH))
+        return false;
+
+    if (!mapf::encryption::aes_encrypt(keywrapkey, iv, encrypted, sizeof(encrypted))) {
```
arnout: Same here: can't we use encrypted_settings->encrypted_settings() directly instead of this intermediate buffer?

arnout (approved):
nitpicks only. Even the wrong calculation of the padding is a nitpick because it will still work with redundant padding.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Will change it, you're right.

PR #155, opened 2019-08-12
tools: docker: move scripts boilerplate

tomereli:
We have now a bunch of scripts all using similar infrastructure such as
functions for printing debug, error, info, etc.
Move this to a new script - functions.sh which is sourced by all other
scripts.

### DIFF
--- a/tools/docker/build.sh
+++ b/tools/docker/build.sh
@@ -6,13 +6,10 @@
 # See LICENSE file for more details.
 ###############################################################
 
-run() {
-    echo "$*"
-    "$@" || exit $?
-}
-
 scriptdir="$(cd "${0%/*}"; pwd)"
 topdir="${scriptdir%/*/*/*}"
+
+. ${topdir}/prplMesh/tools/docker/functions.sh
 
 main() {
     docker image inspect prplmesh-builder >/dev/null 2>&1 || {
--- a/tools/docker/functions.sh
+++ b/tools/docker/functions.sh
@@ -1,0 +1,44 @@
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2019 Tomer Eliyahu (Intel)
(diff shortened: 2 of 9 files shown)


tools/docker/build.sh
```diff
@@ -6,14 +6,11 @@
...
 
-run() {
-    echo "$*"
-    "$@" || exit $?
-}
-
 scriptdir="$(cd "${0%/*}"; pwd)"
 topdir="${scriptdir%/*/*/*}"
 
+. ${topdir}/prplMesh/tools/docker/functions.sh
```
arnout, line 12: `${scriptdir}/functions.sh` would be more appropriate I think. (nitpick)

Same for all other scripts obviously.

tools/docker/functions.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
```
arnout: Since this is not an executable, a shebang is not appropriate. (nitpick)

tools/docker/functions.sh
```diff
@@ -0,0 +1,45 @@
...
+info() {
+	printf '\033[1;35m'"$(basename $0): $@"'\033[0m\n'
+}
+
+success() {
+    printf '\033[1;36m'"$@\n"'\033[0m'
+}
+
+run() {
+    echo "$*"
```
arnout, line 32: Some of the existing scripts use `dbg` here instead of `echo`. I'm not entirely sure which one is more appropriate. Echo is probably good.

arnout (approved):
Only nitpick comments, which you may not even agree with.

tools/docker/build.sh
tomereli: not all scripts are in the same path so it will become `${scriptdir}/../functions.sh` for some which I find less appealing. I think we can keep it as it is now, since `test_flows.sh` and `test_gw_repeater.sh` can be unified to a single test script which can move to the same directory as all others.

PR #157, opened 2019-08-13
framework: common: encryption implementation & unit test

tomereli:
framework: common: encryption impl & unit test

Add encryption implementation mapfcommon.so for supporting EasyMesh AP
Autoconfiguration per WiFi Simple Configuration (WSC) specification v2.0.6:

- Add openssl to build.
- Create a new source and include for encryption in the mapfcommon library.
- Create diffie_hellman class for diffie hellman key exchange
- Create sha256 and hmac classes used for encryption & authentication
- API for creating nonce values (used for IV, M1 & M2 nonces)
- API for KWA - Key Wrap Attribute generation (1st 64Bits of HMAC)
- API for WPS keys calculation which computes secret KeyWrapAuth and
AuthKey keys based on M1 & M2 nonce values and local and remote public
keys
- API for AES encryption / decryption
- Create a class diffie_hellman to do diffie_hellman exchange, as well as
  the corresponding test program. The test program is built even if MSGLIB
  is None - as opposed to the other tests.

Add encryption unit test which performs diffie hellman key exchange,
create nonce values, calculates WPS keys, adds HMAC authentication and
encrypts the plaintext.
Then the test decrypts the cyphertext, verifies HMAC authentication and
that the cyphertext is equal to the plaintext.

Notes:

The authkey is used to generate/verify the authenticator attribute (by
taking HMAC of the config_data).

It is also used to generate/verify the key_wrap_auth attribute that
comes at the end of the encrypted settings, by taking HMAC of the
plaintext of encrypted settings (excluding key_wrap_auth).

keywrap key will be used to encrypt/decrypt the encrypted settings
(config_data) block in the M2 message (including key_wrap_auth).

In addition:
- add controller gdb configuration
- set max vaps to 1 in radio agent (dummy mode)
- Fix radio agent ruid comparison bug

Signed-off-by: Arnout Vandecappelle (Essensium/Mind) <arnout@mind.be>


arnout (changes requested):
As noted in #154, part of the fixup should be applied here.

arnout (approved):

PR #159, opened 2019-08-13
autoconfig: add authentication

tomereli:
Add WSC KWA (Key wrap auth) and global M1 and M2 authentication.

The KWA is the key wrap authenticator which comes after the
config_data in the encrypted settgins attribute.
It is calculated as the 1st 64 bits of HMAC of the config_data using the
authkey computed in wps_calculate_keys().

The global authenticator is carried in the M2 message in
the authenticator attribute, and is calculated with the same algorithm
but on the whole M1 || M2* messages (where M2* is the M2 message without
the authenticator attribute).

The radio agent handles each M2 TLV in the autoconfiguration WSC MultiAP
message by calculating the secret authkey and keywrapkey using
wps_calculate_keys() API, decrypts the encrypted settings buffer using keywrapkey,
then calculates the KWA using the authkey and authenticates the config_data by comparing
it to it to the KWA which comes after the config data.

For global authentication, it saves the M1 buffer each time, then on M2
reception the radio agent compares the authenticator attribute to its
computed one.

Note that in order to authenticate the config_data, and since parsing a
TLVF class automatically converts the whole class to host byte order,
We are required to convert back to network byte order before the authentication process.
Then we convert back to host byte order to use the decrypted config_data
settings. The first conversion to host byte order is required for getting the
config_data block length.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -465,7 +465,13 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
 
     switch (cmdu_message_type) {
     case ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_WSC_MESSAGE:
-        return handle_autoconfiguration_wsc(sd, cmdu_rx);
+        if (!handle_autoconfiguration_wsc(sd, cmdu_rx)) {
+            LOG(ERROR) << "Autoconfiguration WSC failure, slave reset";
+            stop_on_failure_attempts--;
+            slave_reset();
+            return false;
+        }
+        return true;
     case ieee1905_1::eMessageType::CHANNEL_PREFERENCE_QUERY_MESSAGE:
         return handle_channel_preference_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE:
@@ -4298,6 +4304,115 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
     return true;
 }
 
+/**
+ * @brief autoconfig global authenticator attribute calculation
+ * 
+ * Calculate authentication on the Full M1 || M2* whereas M2* = M2 without the authenticator
+ * attribute. M1 is a saved buffer of the swapped M1 sent in the WSC autoconfig sent by the agent.
+ * 
+ * @param m2 WSC M2 TLV
+ * @param authkey authentication key
+ * @return true on success
+ * @return false on failure
+ */
+bool slave_thread::autoconfig_wsc_authenticate(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                               uint8_t authkey[32])
+{
+    if (!m1_auth_buf) {
+        LOG(ERROR) << "Invalid M1";
+        return false;
+    }
+
+    uint8_t buf[m1_auth_buf_len + m2->getLen() - sizeof(WSC::sWscAttrKeyWrapAuthenticator)];
+    auto next = std::copy_n(m1_auth_buf, m1_auth_buf_len, buf);
+    m2->class_swap(); //swap to get network byte order
+    std::copy_n(m2->getStartBuffPtr(), m2->getLen() - sizeof(WSC::sWscAttrKeyWrapAuthenticator),
+                next);
+    m2->class_swap(); //swap back
+
+    uint8_t kwa[WSC::WSC_AUTHENTICATOR_LENGTH];
+    // Add KWA which is the 1st 64 bits of HMAC of config_data using AuthKey
+    if (!mapf::encryption::kwa_compute(authkey, buf, sizeof(buf), kwa)) {
+        LOG(ERROR) << "kwa_compute failure";
+        return false;
+    }
+
+    if (!std::equal(kwa, kwa + sizeof(kwa),
+                    reinterpret_cast<uint8_t *>(m2->authenticator().data))) {
+        return false;
+    }
+
+    LOG(DEBUG) << "WSC Global authentication success";
+    return true;
+}
+
+bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
+    std::shared_ptr<ieee1905_1::tlvWscM2> m2, uint8_t authkey[32], uint8_t keywrapkey[16],
+    std::string &ssid, sMacAddr &bssid, WSC::eWscAuth &auth_type, WSC::eWscEncr &encr_type)
+{
+    auto encrypted_settings = m2->encrypted_settings();
+    uint8_t *iv             = reinterpret_cast<uint8_t *>(encrypted_settings->iv());
+    auto buf                = reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings());
+
+    LOG(DEBUG) << "M2 Parse: aes decrypt";
+    mapf::encryption::aes_decrypt(keywrapkey, iv, buf,
+                                  encrypted_settings->encrypted_settings_length());
+
+    LOG(DEBUG) << "M2 Parse: parse config_data, len = "
+               << encrypted_settings->encrypted_settings_length();
+    // Need to convert to host byte order to get config data length,
+    // which is needed for computing the KWA (1st 64 bits of HMAC)
+    // However, the KWA is computed on the finalized version of the config
+    // data, hence the class_swap.
+    // Finally, once authentication succeeds, swap back to host byte order.
+    // Theoretically, this can be avoided by not passing swap_needed=True to
+    // cConfigData constructor, but then parsing will fail since the length
+    // will be calculated wrong. TLVF does not support parsing network byte order
+    // without full swap, so keep this workaround for now (another future TLVF V2 feature)
+    WSC::cConfigData config_data(buf, encrypted_settings->encrypted_settings_length(), true, true);
+
+    // get length of config_data for KWA authentication
+    size_t len = config_data.getLen();
+
+    // Swap to network byte order for KWA HMAC calculation
+    config_data.class_swap();
+
+    uint8_t kwa[WSC::WSC_AUTHENTICATOR_LENGTH];
+    // Compute KWA based on decrypted settings
+    if (!mapf::encryption::kwa_compute(authkey, buf, len, kwa)) {
+        LOG(ERROR) << "kwa compute";
+        return false;
+    }
+
+    // Basically, the keywrap authenticator is part of the config_data (last member of the
+    // config_data to be precise).
+    // However, since we need to calculate it over the part of config_data without the keywrap
+    // authenticator, and since it is anyway going to be encrypted so config_data is not a
+    // substructure of encrypted_settings, it is easier to define it separately and just append to
+    // config_data.
+    WSC::sWscAttrKeyWrapAuthenticator *keywrapauth =
+        reinterpret_cast<WSC::sWscAttrKeyWrapAuthenticator *>(&buf[len]);
+    keywrapauth->struct_swap();
+    if ((keywrapauth->attribute_type != WSC::ATTR_KEY_WRAP_AUTH) ||
+        (keywrapauth->data_length != WSC::WSC_KEY_WRAP_AUTH_LENGTH) ||
+        !std::equal(kwa, kwa + sizeof(kwa), reinterpret_cast<uint8_t *>(keywrapauth->data))) {
+        LOG(ERROR) << "WSC KWA (Key Wrap Auth) failure" << std::endl
+                   << "type: " << std::hex << int(keywrapauth->attribute_type)
+                   << "length: " << std::hex << int(keywrapauth->data_length);
+        return false;
+    }
+
+    // Swap back to host byte order to read and use config_data
+    config_data.class_swap();
+    ssid      = std::string(config_data.ssid(), config_data.ssid_length());
+    bssid     = config_data.bssid_attr().data;
+    auth_type = config_data.authentication_type_attr().data;
+    encr_type = config_data.encryption_type_attr().data;
+    LOG(DEBUG) << "KWA (Key Wrap Auth) success";
+
+    return true;
+}
+
 bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     // Check if this is a M1 message that we sent to the controller, which was just looped back.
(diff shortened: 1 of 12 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4298,6 +4303,115 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
...
+    uint8_t kwa[WSC::WSC_AUTHENTICATOR_LENGTH];
+    // Add KWA which is the 1st 64 bits of HMAC of config_data using AuthKey
+    if (!mapf::encryption::kwa_compute(authkey, buf, sizeof(buf), kwa)) {
+        LOG(ERROR) << "kwa_compute failure";
+        return false;
+    }
+
+    if (!std::equal(kwa, kwa + sizeof(kwa),
+                    reinterpret_cast<uint8_t *>(m2->authenticator().data))) {
+        LOG(ERROR) << "Global authentication failure";
```
tomereli: All these debug prints will be removed once review is completed

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4298,6 +4303,115 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
...
+        LOG(DEBUG) << "buf:" << std::endl << utils::dump_buffer(buf, sizeof(buf));
+        LOG(DEBUG) << "calculated authenticator: " << std::endl
+                   << utils::dump_buffer(kwa, sizeof(kwa));
+        LOG(DEBUG) << "received authenticator: " << std::endl
+                   << utils::dump_buffer((uint8_t *)m2->authenticator().data,
+                                         sizeof(m2->authenticator().data));
+        return false;
+    }
+
+    LOG(DEBUG) << "TOMER GLOBAL AUTHENTICATION SUCCESS";
```
tomereli: This will change to a different print that will be captured by the flow test

common/beerocks/bcl/include/beerocks/bcl/beerocks_utils.h
```diff
@@ -76,6 +76,7 @@ class utils {
     static void hex_dump(const std::string &description, uint8_t *addr, int len,
                          const char *calling_file = __builtin_FILE(),
                          int calling_line         = __builtin_LINE());
+    static std::string dump_buffer(uint8_t *buffer, size_t len);
```
arnout, line 79: Two small nits:

- I don't really like that we have two functions (hex_dump and dump_buffer) that do essentially the same thing but have completely separate implementations.
- I prefer if the order of things in the header and in the implementation is the same.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4298,6 +4304,104 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
...
+        LOG(ERROR) << "Invalid M1";
+        return false;
+    }
+
+    uint8_t buf[m1_auth_buf_len + m2->getLen() - sizeof(WSC::sWscAttrKeyWrapAuthenticator)];
+    auto next = std::copy_n(m1_auth_buf, m1_auth_buf_len, buf);
+    m2->class_swap(); //swap to get network byte order
+    std::copy_n(m2->getStartBuffPtr(), m2->getLen() - sizeof(WSC::sWscAttrKeyWrapAuthenticator),
+                next);
+    m2->class_swap(); //swap back
```
arnout, line 4331: Instead of making copies, you could extend kwa_compute to either:

- take a second buffer argument; or
- return an intermediate state so it can be called a second time.

(nitpick and actually I'm not so sure either of these is really better)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4298,6 +4304,104 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
...
+    // Swap to network byte order for KWA HMAC calculation
+    config_data.class_swap();
+
+    uint8_t kwa[WSC::WSC_AUTHENTICATOR_LENGTH];
+    // Compute KWA based on decrypted settings
+    if (!mapf::encryption::kwa_compute(authkey, buf, len, kwa)) {
+        LOG(ERROR) << "kwa compute";
+        return false;
+    }
+    // key_wrap_auth comes after the config data
```
arnout: This is a bit awkward so it deserves a bigger comment... Though maybe better in the yaml file and not here.

Basically, the keywrap authenticator is part of the config_data. However, since we need to calculate it over the part of config_data without the keywarp authenticator, and since it is anyway going to be encrypted so config_data is not a substructure of encrypted_settings, it is easier to define it separately and just append to config_data.

Because of this, we can't use addClass below. (well, also because it's not a CMDU so it doesn't have addClass)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4298,6 +4304,104 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
...
+    WSC::sWscAttrAuthenticator *authenticator =
+        reinterpret_cast<WSC::sWscAttrAuthenticator *>(&buf[len]);
```
arnout: This should be sWscAttrKeyWrapAuthenticator.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4298,6 +4304,104 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
...
+    if (!std::equal(kwa, kwa + sizeof(kwa), reinterpret_cast<uint8_t *>(authenticator->data))) {
+        LOG(ERROR) << "WSC KWA (Key Wrap Auth) failure";
+        return false;
+    }
```
arnout, line 4403: We should probably also verify that it really is the KeyWrapAuthenticator, i.e. if type and length are correct. And that len + sizeof(sWscAttrKeyWrapAuthenticator) <= encrypted_settings_length

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4298,6 +4304,104 @@ bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlv
...
+ */
+bool slave_thread::autoconfig_wsc_authenticate(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                               uint8_t authkey[32])
+{
+    if (!m1_auth_buf) {
+        LOG(ERROR) << "Invalid M1";
+        return false;
+    }
+
+    uint8_t buf[m1_auth_buf_len + m2->getLen() - sizeof(WSC::sWscAttrKeyWrapAuthenticator)];
```
arnout, line 4326: This should be sWscAttrAuthenticator. They're identical except for their type, obviously, but correct is correct :-)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4358,36 +4462,28 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
+        if (!autoconfig_wsc_parse_m2_encrypted_settings(tlvWscM2, authkey, keywrapkey, ssid, bssid,
+                                                        authtype, enctype))
+            return false;
 
         std::string manufacturer =
             std::string(tlvWscM2->manufacturer(), tlvWscM2->manufacturer_length());
-        // ONLY FOR DEBUG!!!
-        // TODO - remove this print, it is showing the encrypted settings in the logs!
+
+#ifdef DEBUG // TODO can we print that in production?
```
arnout: Why not?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -400,14 +402,22 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
 
     std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(),
                 encrypted_settings->encrypted_settings());
+    WSC::sWscAttrAuthenticator authenticator;
+    auto buf     = reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings());
```
arnout: It would be better to declare this a bit earlier and use it in the copy_n above as well.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -388,7 +388,9 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
     std::shared_ptr<ieee1905_1::tlvWscM2> m2, WSC::cConfigData &config_data, uint8_t authkey[32],
     uint8_t keywrapkey[16])
 {
-    size_t len = (config_data.getLen() + 15) & ~0xFU; // Align to 16 bytes boundary
+    // Calculate length of data to encrypt
+    // (= plaintext length + 32 bits HMAC aligned to 16 bytes boundary)
```
arnout: 32 bits? It is 64 bits HMAC + 32 bit attribute header = 96 bits = 12 bytes... Which is BTW mentioned explicitly somewhere in the spec: " The Key Wrap Authenticator is 96 bits long"

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -388,7 +388,9 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
+    size_t len = (config_data.getLen() + sizeof(WSC::sWscAttrAuthenticator) + 15) & ~0xFU;
```
arnout: sWscAttrKeyWrapAuthenticator

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -400,14 +402,22 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
 
     std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(),
                 encrypted_settings->encrypted_settings());
+    WSC::sWscAttrAuthenticator authenticator;
```
arnout: sWscAttrKeyWrapAuthenticator.

Also, this struct doesn't have a constructor, you need to call struct_init explicitly to set the type and length fields.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -400,14 +402,22 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
+    auto buf     = reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings());
+    uint8_t *kwa = reinterpret_cast<uint8_t *>(authenticator.data);
+    // Add KWA which is the 1st 64 bits of HMAC of config_data using AuthKey
+    if (!mapf::encryption::kwa_compute(authkey, buf, config_data.getLen(), kwa))
+        return false;
+
+    std::copy_n(reinterpret_cast<uint8_t *>(&authenticator), sizeof(WSC::sWscAttrAuthenticator),
```
arnout: To make sure you really have the right type, I prefer to use `sizeof(authenticator)` in a situation like this.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -400,14 +402,22 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
     uint8_t *iv = reinterpret_cast<uint8_t *>(m2->encrypted_settings()->iv());
 
     if (!mapf::encryption::create_iv(iv, WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH))
         return false;
 
-    if (!mapf::encryption::aes_encrypt(
-            keywrapkey, iv, reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings()),
-            encrypted_settings->encrypted_settings_length())) {
+    if (!mapf::encryption::aes_encrypt(keywrapkey, iv, buf,
+                                       encrypted_settings->encrypted_settings_length())) {
```
arnout: This could already have been done before, but: you can just use `len` here.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -448,6 +458,46 @@ bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tl
...
+ * 
+ * Calculate authentication on the Full M1 || M2* whereas M2* = M2 without the authenticator
+ * attribute.
+ * 
+ * @param m1 WSC M1 TLV
+ * @param m2 WSC M2 TLV
+ * @param authkey authentication key
+ * @return true on success
+ * @return false on failure
+ */
```
arnout, line 473: Do we put doxygen comments in the header or in the source? Or maybe in the header for external (i.e.public) function and in the source for private functions?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -448,6 +458,46 @@ bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tl
...
+ * @brief autoconfig global authenticator attribute calculation
+ * 
+ * Calculate authentication on the Full M1 || M2* whereas M2* = M2 without the authenticator
+ * attribute.
+ * 
+ * @param m1 WSC M1 TLV
+ * @param m2 WSC M2 TLV
+ * @param authkey authentication key
+ * @return true on success
+ * @return false on failure
```
arnout, line 472: Does doxygen allow two @return tags?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -448,6 +458,46 @@ bool master_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tl
...
+bool master_thread::autoconfig_wsc_authentication(std::shared_ptr<ieee1905_1::tlvWscM1> m1,
+                                                  std::shared_ptr<ieee1905_1::tlvWscM2> m2,
+                                                  uint8_t authkey[32])
+{
+    // Authentication on Full M1 || M2* (without the authenticator attribute)
+    // authentication is done on swapped data, so first swap the m1 and m2, calculate authenticator, then swap back
+    // since finalize will do the swapping.
+    m1->class_swap();
+    m2->class_swap();
+    uint8_t buf[m1->getLen() + m2->getLen() - sizeof(WSC::sWscAttrKeyWrapAuthenticator)];
```
arnout: sWscAttrAuthenticator

arnout (changes requested):
2nd commit message: "config_data in the encrypted sttgins attribute" -> settings.

All comments are nitpicks or simple to fix. However, I'd like to double-check the correctness of the two authenticator types, so changes requested.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Correct👌

controller/src/beerocks/master/son_master_thread.cpp
tomereli: This AutoGenerated by vscode doxygen extension so I'd assume it does, but I never actually generated doxygen so not sure.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: I tend to follow the file I'm editing so in this case, the cpp file. I prefer them in the header so what we did in the encryption impl shouldn't be changed, but we can open a refractor task for moving all doxygen comments to the headers.

common/beerocks/bcl/include/beerocks/bcl/beerocks_utils.h
tomereli: OK

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: I liked the idea of intermediate state so implemented that :)

common/beerocks/bcl/source/beerocks_utils.cpp
```diff
@@ -310,53 +310,3 @@ std::string utils::dump_buffer(uint8_t *buffer, size_t len)
...
-    std::stringstream print_stream;
-
-    // Process every byte in the data.
-    for (i = 0; i < len; i++) {
-        // Multiple of 16 means new line (with line offset).
-
-        if ((i % 16) == 0) {
-            // Just don't print ASCII for the zeroth line.
-            if (i != 0) {
-                print_stream << "   " << std::string(ascii_chars) << std::endl;
```
arnout: We're really loosing a lot of features here: the ASCII representation, the indentation, the offset... It might make sense to keep the implementation of hex_dump and just remove the caller part.

framework/common/encryption.cpp
```diff
@@ -347,16 +347,30 @@ bool wps_calculate_keys(const diffie_hellman &dh, const uint8_t *remote_pubkey,
...
-bool kwa_compute(const uint8_t *authkey, uint8_t *data, uint32_t data_len, uint8_t *kwa)
+bool kwa_compute(const uint8_t *authkey, uint8_t *data, uint32_t data_len, uint8_t *kwa, bool more)
 {
     uint8_t hmac_[32];
-    hmac hmac_kwa(authkey, 32);
-    if (!hmac_kwa.update(data, data_len))
-        return false;
-    if (!hmac_kwa.digest(hmac_))
-        return false;
+    static hmac *hmac_kwa = nullptr;
```
arnout: Hm, I really don't like it when functions are made non-reentrant... I much prefer the copy in that case.

Basically, we should use the hmac class directly instead of kwa_compute.

Or keep it as it was :-)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -402,15 +402,17 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
 
     std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(),
                 encrypted_settings->encrypted_settings());
-    WSC::sWscAttrAuthenticator authenticator;
+    WSC::sWscAttrKeyWrapAuthenticator keywrapauth;
+    keywrapauth.struct_init();
+    keywrapauth.struct_swap();
```
arnout, line 408: Right, I forgot that that was needed!

arnout (approved):
OK to merge if you kill the fixup that reworks kwa_compute.

If you do want to still do that differently, I'd like to re-review.

framework/common/encryption.cpp
tomereli: we'll kill the fixup

PR #161, opened 2019-08-14
Add ap capability flow to controller, agent, and CI

Galvien:
This pull request adds support for passing Controller Certification AP Capability Query test as defined in section 5.6.1 in the EasyMesh Test Plan v1.0 [1].

First, a shell test flow as described in [1][5.6.1] for the AP Capability Query sent by the controller to the agent is added to `test_flows.sh`.

Then, the controller is updated to support sending the query message, it was created in the tlvf;
bml was updated in order to support sending the message to the specified al_mac, and the CLI was updated to support the command for sending the message.

Lastly, the agent backhaul manager was updated to subscribe and handle the received message, thus passing the test.

Implements #164 👌 

Note that the current handling of the message by the agent is stub, and will be handled later during agent certification.


tomereli:
Looks like the topology test failed, please try to test it locally

tomereli (approved):
I will keep it open for a few hours so @arnout can also review he wants to / have the time, but LGTM.

tools/docker/tests/test_flows.sh
```diff
@@ -49,12 +49,13 @@ test_channel_selection() {
     return 1
 }
 
-test_ap_capability_info_reporting() {
-    #TODO: Implement
-    return 1
+test_ap_capability_query() {
+    dbg "sending ap capability query message"
+    eval send_bml_command "bml_trigger_ap_capability" $AL_MAC $redirect
```
arnout, line 54: I hate it that I come so late with this feedback, but I think this is completely the wrong approach to implement the flow tests.

I think we should stay as close as possible to how the certification test itself operates. In other words, I think we should *not* insert new BML commands here, but instead we should use `bml_wfa_ca_controller,DEV_SEND_1905,...`

I'm not going to block this PR because of that however, but for the future it's something to consider. To be discussed with @tomereli

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1576,12 +1579,26 @@ bool main_thread::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
  */
 bool main_thread::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cmdu_rx)
 {
-    LOG(DEBUG) << "Received Topology Query message";
+    LOG(DEBUG) << "Received Topology Discovery Query message";
```
arnout, line 1582: The message is called "Topology Query" so the original log was correct.

arnout (approved):
Even though I propose a major design change for this PR, I'm not going to block this PR because of it.

There are some small mistakes in the commit messages.

    controller: add bml ap capability query handling
    
    When receiving ap capability query trigger in the controller from bml
    sends an ap capabillity query message to the respecive agent.

That sentence doesn't make sense. -> When receiving ap capability query trigger in the controller from bml, send an ap capabillity query message to the respecive agent.'

tools:docker: change font colors for status prints -> missing space after colon.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -126,8 +126,8 @@ bool main_thread::init()
             ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_WSC_MESSAGE,
             ieee1905_1::eMessageType::CHANNEL_PREFERENCE_QUERY_MESSAGE,
             ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE,
-            ieee1905_1::eMessageType::ACK_MESSAGE,
-            ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE})) {
+            ieee1905_1::eMessageType::ACK_MESSAGE, ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE,
```
morantr, line 129: Why ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE moved to the same line with ieee1905_1::eMessageType::ACK_MESSAGE? Did the clang move it?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1560,6 +1560,9 @@ bool main_thread::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
         return handle_1905_discovery_query(cmdu_rx);
         //LOG(INFO) << "I got the Topology Query message!";
     }
+    case ieee1905_1::eMessageType::AP_CAPABILITY_QUERY_MESSAGE: {
+        return handle_1905_ap_capability_query(cmdu_rx);
+    }
```
morantr, line 1565: Why did you implement the stub here? shouldn't this message handler be on son_slave?

morantr (changes requested):

tools/docker/tests/test_flows.sh
morantr: I agree with you Arnout.
And it is a really small change too. Just dropping all the commits related to bml, move the creation of the CMDU to wfa-ca, and update the test flow to send "bml_wfa_ca_controller <command>".

And on the way, do the same to topology query 😈

tomereli: @arnout I agree - but I asked Gal to do this since its better than nothing and we're not in the phase that we can use `bml_wfa_ca_controller,DEV_SEND_1905,...` now, are we?
Once it is ready, we'll refactor all tests and remove the BMLs which are not needed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: Apparently not - this is a CMDU without TLVs so the message gets sent to the main agent, which in turn would send VS messages to query all radio agents.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: agree - @Galvien please add a fixup commit for this.

tools/docker/tests/test_flows.sh
tomereli: @Galvien lets do that, we're not in a hurry to merge this, so if there's a better way lets go for it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: clang moved it because the `))) {` are no longer on the same line, just a comma. Therefore, the TOPOLOGY_QUERY_MESSAGE now fits.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: So what it doesn't have TLVs? Channel preference query is also a message without TLVs and being handled in the son_slave which is responsible for the radio.
The backhaul manager automatically forwards messages to the son_slaves if not handled in the backhaul manager itself.
I don't see the point of handling it here and then create new CMDU to send it to the son_slave.

arnout:
As discussed, a different approach should be taken, sending wfa_ca commands directly instead of creating new BML commands.

Issue #121, opened 2019-07-22
[TASK]  CAPI Commands support via bml_wfaca for  CTT Test Bed


kantera800:
Description
Currently, the BML command uses local dummy function instead of drilling the message to the controller.
HLD #132 

changes have to be done:

- [x] remove dummy function and drill down WFA_CA raw string command to the controller with response mechanism.  (3 SP) 

- [x] Add certification  flag to controller configuration (1 SP) 

- [x] Create a handler for "DEV_RESET_DEFAULT" that performs prpl_mesh_utils stop/start and  return "OK"(1.5 SP)

- [x] Create a handler for "DEV_SET_CONFIG" that only return dummy "OK" without actually set the device to a specific configuration, since the channel selection test does not actually check that. (0.5 SP)

- [x] Create a handler for "DEV_SEND_1905" that parses the command by converting the stream of the text into a stream of data, and send it to the required destination.
Implementation will be parsing the wfa-ca string and do a specific implementation for each CMDU type.
analyze with a sniffer and estimate implementation (0.5sp) implementation (TBD)

- [x] when certification flag is set, hold in channel selection request on controller DB, and don't send it automatically. (0.5 SP) 

- [x] Internal unitest validation. generate dummy message via listener script ucc.sh  run wfa_ca-proxy.sh and make sure its parsed correctly to reach bml_wfaca . (2 SP) 

Exit criteria:
 following commands  sent via bml_wfaca ,  expected result according to the description or/and  Wi-Fi CERTIFIED EasyMesh™ Test Plan.

- DEV_RESET_DEFAULT  
- DEV_SET_CONFIG with parameters:name, program, bss_infoX, backhaul
- DEV_SEND_1905 with parameters: DestALid, MessageTypeValue, tlv_type, tlv_length, tlv_value

Assumptions:
-  we can test CAPI commands via  Wifi test suit or  ucc.sh script.
-  in case DEV_SEND_1905 handler  implementation estimation is more than 2 sp it will be treated in separate task

PR #163, opened 2019-08-14
Feature/wfa ca ucc support

morantr:
Add UCC Listener mode to beerocks_cli that listen WFA-CA commands from
the network, and send the replies back to the sender.

In order to use the new UCC listener mode, execute the cli with
argument '-u <port>', for example:
./beerocks_cli -u 7001
The UCC listener will listen to incoming TCP connections on this port
and reply to WFA-CA commands.

As part of this PR, fixed small bugs and remove old WFA-CA old test files.

#121


controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -237,14 +240,17 @@ void wfa_ca::handle_wfa_ca_message(
     switch (command_type) {
     case eWfaCaCommand::CA_GET_VERSION: {
         // TODO
```
tomereli: You can just implement it and return "prplMesh-v1.0" or something..

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -814,7 +814,21 @@ void cli_bml::events_update_to_socket_cb(const struct BML_EVENT *event)
...
+    cli_bml *pThis = reinterpret_cast<cli_bml *>(bml_get_user_data(ctx));
+    if (!pThis) {
+        std::cout << "ERROR: Internal error - invalid context!" << std::endl;
+        return;
+    }
+
+    static const std::string ref("status,RUNNING");
+    std::string reply(buffer);
+    pThis->pending_response =
+        ((reply.size() >= ref.size()) && (reply.substr(0, ref.size()) == ref)) ? true : false;
```
tomereli, line 829: The logic to check for pending response is SO complex and made me read this 5 times till I understood what you're doing here. Its a good sign that this should be done differently.
Why not change the callback to pass `pending_response` instead of this hack style parsing of the reply buffer? it makes no sense to parse it here, this function should be a simple pipe.

controller/src/beerocks/cli/beerocks_cli_main.cpp
```diff
@@ -239,8 +241,8 @@ static int cli_non_interactive(std::string path, std::string tmp_path, std::stri
 
     UTILS_SLEEP_MSEC(1000 * g_wait_time);
 
-    // This mechanism (cli_bml.is_pending_response() ) is only implemented for "bml_nw_map_query"
-    auto timeout = std::chrono::steady_clock::now() + std::chrono::seconds(3);
+    // Max timeout according to CAPI specifications is 120 seconds
```
tomereli, line 243: This should be done in a separate commit OR at least add this to the commit message.

tomereli (dismissed):
Now this is a real problem - and why we need design reviews before implementations - when we'll get to agent certification, there will be no controller, hence no BML - so the whole CLI solution you did here can't be used. Please correct me if i'm wrong.

Also I don't think we should remove the script since it might still be useful in the future.

controller/src/beerocks/master/wfa_ca.cpp
morantr: Not in the scope it the task, but ok i'll do it.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
morantr: Or maybe it is because you review it late at night 😉 
Even if I'll pass it in the callback, this piece of code will just move to somewhere else.
At first, I passed it on the BML WFA-CA response message but you guys asked me to remove it.
Basically what I did here is to check if the response message stat with "status,RUNNING"
Maby I will add a comment that explains what I did here, what do think?

controller/src/beerocks/cli/beerocks_cli_main.cpp
morantr: I did mention it in the commit message, look at the three last lines in the commit message.

morantr:
> Now this is a real problem - and why we need design reviews before implementations - when we'll get to agent certification, there will be no controller, hence no BML - so the whole CLI solution you did here can't be used. Please correct me if i'm wrong.
> 
> Also I don't think we should remove the script since it might still be useful in the future.

The BML supports sending messages to the agent. It doesn't have many commands to send (I think the onbarding query is only one). But when we will get to this point we could add '-r' for a role, or maybe do some command processing to figure out if it is controller or agent.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
morantr: Added a comment

tomereli (commented):
Can we not remove the old wfa_ca code just yet? it has been used previously and worked, so its good to keep it as reference.
I'm not approving yet so @arnout has a chance to also review.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -178,10 +178,13 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
     err_string = "missing mandatory params: ";
     for (const std::string &param_name : mandatory_params) {
         if ((params.find(param_name))->second.empty()) {
-            err_string += "'" + param_name + "', ";
+            err_string += param_name + ", ";
         }
     }
 
+    // Remove last comma
+    err_string.pop_back();
```
arnout, line 187: Doesn't this remove the last *space* (because above, we append `", "`)?

controller/src/beerocks/cli/beerocks_cli_main.cpp
```diff
@@ -215,6 +216,8 @@ static int cli_non_interactive(std::string path, std::string tmp_path, std::stri
     beerocks::cli *cli_ptr = &cli_soc;
     std::string bml_prefix(BML_PREFIX);
 
+    beerocks::string_utils::trim(command_string, "\"\'");
```
arnout: This will break the handling of quoted arguments in parsedCommandArgumentsLinux, won't it?

In addition, I don't see why this would be needed... According to the commit message:

> The non-interactive mode was able to receive only command with no
>     arguments (such as 'bml_conn_map'), 

Actually, I use commands with arguments in non-interactive mode all the time, e.g.  `beerocks_cli  -c "dump_node_info 00:11:22:33:00:20"`. It works for me...

> since the arguments string were
>     cutoff by getopt(). To force getopt() to take the argument one must
>     wrap the command and arguments with single or double-quotes.
>     But then it caused the command to not be found since it has quotes
>     note on its beginning.

Err, no it doesn't... The quotes are interpreted by the shell, and the shell passes the unquoted argument (containing a space) in argv.

>     The solution is to trim these quotes on non-interactive mode.

In other words, I don't see how/why this is needed.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
```diff
@@ -0,0 +1,169 @@
...
+cli_ucc_listener::cli_ucc_listener(const std::string &beerocks_conf_path, uint16_t port)
+    : socket_thread()
+{
+    thread_name          = "ucc_listener";
+    m_beerocks_conf_path = beerocks_conf_path;
+    m_port               = port;
+    set_select_timeout(SELECT_TIMEOUT_MSC);
+
+    // Greetings message
+    std::cout << "Welcome to beerocks TestSuit UCC Listener!" << std::endl
```
arnout: Fixed a spelling mistake here.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
```diff
@@ -0,0 +1,169 @@
...
+ * See LICENSE file for more details.
+ */
+
+#include "beerocks_cli_ucc_listener.h"
+#include "beerocks_cli_bml.h"
+
+#include <beerocks/bcl/beerocks_string_utils.h>
+
+#define SELECT_TIMEOUT_MSC 100
+#define NLF std::endl << std::flush // New Line Flush
```
arnout, line 15: NLF usually means Newline - Line Feed, but OK.

However, why is the flushing needed? This is used only on std::cout, as far as I can see, and cout should already be flushed automatically on endl...

tomereli:
@arnout can we merge this PR? we're waiting for your review.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+void cli_ucc_listener::disconnect()
+{
+    if (m_ctx != nullptr) {
+        bml_disconnect(m_ctx);
+        m_ctx = nullptr;
+    }
+}
+
+bool cli_ucc_listener::custom_message_handler(Socket *sd, uint8_t *rx_buffer, size_t rx_buffer_size)
```
arnout, line 106: TBH I think this approach of a custom message handler is a hack. Basically what you're trying to do is to reuse the listen/accept logic from socket_thread, but without the CMDU handling. Which is essentially a SocketServer. So I think it's better to use SocketServer directly instead of adding this custom_message_handler to socket_thread.

Ideally, there shouldn't even be a need to create a thread for this. You should just be able to add this SocketServer to BML's polling loop. Unfortunatley, the BML interface doesn't allow that. So a new thread is indeed required.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
```diff
@@ -0,0 +1,169 @@
...
+        available_bytes = sd->readBytes(rx_buffer, sizeof(rx_buffer), false, rx_buffer_size);
+        if (available_bytes <= 0) {
+            break;
+        }
+        rx_buffer[available_bytes] = '\0';
+        std::string part(reinterpret_cast<char *>(rx_buffer));
+        command_string += part;
+    } while (available_bytes > 0);
+
+    beerocks::string_utils::trim(command_string, "\"\'");
```
arnout: Does UCC ever put quotes around commands? I don't think this is needed or even appropriate.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
```diff
@@ -0,0 +1,169 @@
...
+    // Socket thread will remove socket from Select and delete it
+    return true;
+}
+
+bool cli_ucc_listener::connect()
+{
+    disconnect(); // Disconnect if already connected
+    const char *conf_path = m_beerocks_conf_path.c_str();
+    int ret               = bml_connect(&m_ctx, conf_path, this);
+    return (ret == BML_RET_OK ? true : false);
```
arnout: `? true : false` makes no sense. If you need an explicit boolean conversion, use `bool (ret == BML_RET_OK)` but that's really not needed here, so just `return (ret == BML_RET_OK)`.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -236,15 +239,45 @@ void wfa_ca::handle_wfa_ca_message(
 
     switch (command_type) {
     case eWfaCaCommand::CA_GET_VERSION: {
-        // TODO
+
+        /*
```
arnout: This and many of the following lines had a spurious space at the end of the line. Please configure your editor to avoid that.

End-of-line space is bad because it can easily lead to spurious differences in later edits.

I've fixed this up while rebasing.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -236,15 +239,45 @@ void wfa_ca::handle_wfa_ca_message(
 
     switch (command_type) {
     case eWfaCaCommand::CA_GET_VERSION: {
-        // TODO
+
+        /* 
+        * This command is used to obtain the version of the Control Agent software. The command does
```
arnout: This should be indented one space more to align the asterisks.

I've fixed this up while rebasing.

arnout:
> Can we not remove the old wfa_ca code just yet? it has been used previously and worked, so its good to keep it as reference.

I disagree. Dead code is not a good reference.

arnout (changes requested):
The use of socket_thread for something which is not a CMDU communication is a major deficiency IMO. I therefore won't do what I proposed before (just apply and fix up in follow-up), but keep it as changes requested.

However, I did rebase with some minor fixups (spelling mistakes, whitespace errors).

I also properly split the first two commits in 5 commits, like they should have been. It took me 10 minutes, but I think it's worth the time investment. Note that I believe the commit "fix wfa-ca command on non interactive mode" is incorrect and should not be included, but I leave it up to @morantr to confirm that.

controller/src/beerocks/master/wfa_ca.cpp
morantr: yes, need to pop 2 time. will fix

controller/src/beerocks/cli/beerocks_cli_main.cpp
morantr: It won't break quoted arguments because it only trims on the sides.
> This will break the handling of quoted arguments in parsedCommandArgumentsLinux, won't it?
> 
> In addition, I don't see why this would be needed... According to the commit message:
> 
> > The non-interactive mode was able to receive only command with no
> > arguments (such as 'bml_conn_map'),
>

Exactly, because you used quotes. If you  weren't, then optarg was equal to 'dump_node_info', without the rest of the command.
> Actually, I use commands with arguments in non-interactive mode all the time, e.g. `beerocks_cli -c "dump_node_info 00:11:22:33:00:20"`. It works for me...
> 
> > since the arguments string were
> > cutoff by getopt(). To force getopt() to take the argument one must
> > wrap the command and arguments with single or double-quotes.
> > But then it caused the command to not be found since it has quotes
> > note on its beginning.
> 
AFAIR, when I tested it there were quote characters on optarg.
> Err, no it doesn't... The quotes are interpreted by the shell, and the shell passes the unquoted argument (containing a space) in argv.
> 
> > ```
> > The solution is to trim these quotes on non-interactive mode.
> > ```
> 
> In other words, I don't see how/why this is needed.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
morantr: For debug reasons. Since it is a multithreaded file, I wanted to make sure that the prints will be flushed.

I looked here and it doesn't say explicitly that std::endl flushes the output.
https://en.cppreference.com/w/cpp/io/manip/endl

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
morantr: Agreed on the daily meeting to split it socket thread to cmdu_socket_thread which will be derived from socket thread.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
morantr: I think, because I didn't trim it at first, and Tomer asked me to add it when it worked with UCC.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
morantr: I have no idea how I wrote such an ugly as that. I will fix.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I don't have such a setting in my editor

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
arnout: Turns out I was wrong - it is std::cerr that is flushed automatically, and not just on endl but on any operation: https://en.cppreference.com/w/cpp/io/manip/unitbuf

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
tomereli: I don't recall that I asked that...

morantr: So what did you asked me to change? I don't remember either.
Maybe \r\n? because it also trim these characters as well.
So maybe we could change it to:
`beerocks::string_utils::trim(command_string);`

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
arnout: Created #178 to do this in the future.

controller/src/beerocks/cli/beerocks_cli_main.cpp
arnout: I will remove this.

controller/src/beerocks/cli/beerocks_cli_ucc_listener.cpp
arnout: I will remove this.

morantr: but not completely, right? only the second argument because trimming the \r\n is required.

arnout: Thanks for reminding me!

arnout (approved):

PR #168, opened 2019-08-15
tools: docker: Add channel selection to test flow

morantr:
Add the channel selection test flow to the test_flows.sh script.
Sending channel preference query  and channel selection request commands
and grep on the agent log to these messages + 1905.1 ACK message.


tomereli (changes requested):
LGTM but please also add "channel_selection" to the test_flows command in .travis.yaml so that CI will also run this

tomereli (approved):

PR #169, opened 2019-08-18
controller:wfa_ca: fix bug with consecutive sends

CoralMalachi:
The MCUT (MultiAP Controller Under Test) gets a string command from
WTS UCC (Wifi Test Suite Unified CAPI Console).

The wfa_cp.cpp file is responsible of parsing the string to a command
and handle its execution.

The current bug is that the WTS successfuly sends the command on the
first time,but on the second time it fails.

The reason for that is that the unordered_map params was defined as a
static variable, preventing its value from resetting, causimg it to
increase in size.

This caused the size check on line 434 to be false in every iteration
after the first.Thus causing it not to create a CMDU.

The solution is to make the unordered_map an instance variable on
relevant locations.

### DIFF
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -277,7 +277,7 @@ void wfa_ca::handle_wfa_ca_message(
 
         // NOTE: Note sure this parameters are actually needed. There is a controversial
         // between TestPlan and CAPI specification regarding if these params are required.
-        static std::unordered_map<std::string, std::string> params{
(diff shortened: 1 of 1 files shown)


morantr:
Looks good. Run ./clang_format so Travis will pass and amend the change.
Then make it a regular PR.

tomereli (commented):
LGTM, fix wrong line wrappings in commit message, fix typos, run clang-format.sh like Moran said and make this a regular PR.

Very good commit message!

tomereli (approved):

Issue #172, opened 2019-08-19
[TASK] Add sniffer capability to virtual certification testbed


arnout:
## Description

We added a virtual certification testbed that is able to run the MCUT in a container together with several instances of a stub Python script that returns success to all UCC commands (cfr. #170). However, the tests currently stop when UCC tries to get results from the sniffer. To cover more situations with the virtual certification testbed, implement a virtual sniffer by running pcap on `br-lan`.

## Exit criteria

UCC is able to receive the 1905.1 packets that go over `br-lan` from the virtual sniffer.

arnout:
This has been done a long time ago.

PR #175, opened 2019-08-20
Add initial virtual certification testbed support

tomereli:
This PR adds the capability to run the WiFi Alliance UCC application for EasyMesh certification virtually using a single Windows PC. See (Virtual certification testbed on Windows WiKi)[https://github.com/prplfoundation/prplMesh/wiki/Virtual-certification-testbed-on-Windows] for full instructions.

Once the WiFi-TestSuite is installed, the `AllInitConfig_MAP.txt` is updated so that all testbed devices are set to localhost on running port numbers.
A dummy UCC server python script is also included which listens on those ports and replies to the testbed UCC application instead of real devices, thus allowing the tests to proceed with only a single DUT which is prplMesh controller & agent.

We are running prplMesh from a container and publish port 5000 which is used to receive commands from the UCC application.

This PR includes the following commits:
- Fix symlink issue - on Windows, symlinks are converted to text files which breaks native build as our build system mistakingly thinks that we are cross-compiling
- Add  Docker image for for complete containers which allow building and running from the same container.
- Add UCC AllInitConfig_MAP.txt reference and ucc dummy server 

Issue #172

### DIFF
--- a/framework/wfa/ucc-dummy.py
+++ b/framework/wfa/ucc-dummy.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2019 Tomer Eliyahu (Intel)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+import socket
+import argparse
+import threading
+
+def handle(host, port, respDict):
+    id = threading.get_ident()
+    print("{}: Listening on {}:{}".format(id, host, port))
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind((host, port))
+        s.listen()
+        with s.accept() as (conn, addr):
+            print("{}: {}@{} Connected by {}".format(id, host, port, addr))
+            while True:
+                try:
+                    data = conn.recv(1024)
+                except Exception as e:
+                    print("Exception {}".format(e))
+                    break
+                print("{}@{} Received: {}".format(host, port, data))
+                resp = [b'Status,RUNNING', b'Status,COMPLETE'] # default response
+                for key in respDict:
+                    if key in str(data):
+                        resp = respDict[key] # specific resonse
+                for r in resp:
+                    print("{}@{} Sending: {}".format(host, port, r))
+                    conn.sendall(r)
+    print("{}: {}@{} Exiting!".format(id, host, port))
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--ports", nargs='+', type=int,
+                    default=[7000, 7001, 7002, 7003, 7004, 7005, 7006,
+                             7007, 7008, 7009, 7010, 7011, 7012, 8888])
+parser.add_argument("--host", default='127.0.0.1')
+parser.add_argument("--dst-alid", default="11:22:33:44:55:66")
+args = parser.parse_args()
+
+respDict = {
+    'dev_reset_default' : [b'Status,RUNNING', b'Status,COMPLETE'],
+    'ca_get_version' : [b'Status,RUNNING', b'Status,COMPLETE,vendor,dummy,model,dummy,version,1.0'],
+    'dev_get_parameter' : [b'Status,RUNNING', 'Status,COMPLETE,aLid,{}'.format(args.dst_alid).encode()],
+}
+
+print("Starting threads on host {}, ports {}".format(args.host, args.ports))
+threads = list()
+
+for port in args.ports:
+    print("Creating thread {}@{}".format(args.host, port))
+    t = threading.Thread(target=handle, args=(args.host, port, respDict,))
+    threads.append(t)
+    print("Starting thread {}@{}".format(args.host, port))
+    t.start()
+
+for t in threads:
+    print("Before joining thread {}@{}".format(args.host, port))
+    t.join()
+    print("After joining thread {}@{}".format(args.host, port))
--- a/tools/docker/all/Dockerfile
+++ b/tools/docker/all/Dockerfile
@@ -1,0 +1,42 @@
+ARG image=ubuntu:18.04
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2019 Tomer Eliyahu (Intel)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+FROM $image
+
+# Update Software repository
+RUN apt-get update && apt-get install -y \
+	curl \
+	gcc \
+	cmake \
+	ninja-build \
+	git \
+	autoconf \
+	autogen \
+	libtool \
+	pkg-config \
+	gdb \
+	binutils \
+	libreadline-dev \
+	libncurses-dev \
+	libssl-dev \
+	libjson-c-dev \
+	libnl-genl-3-dev \
+	libzmq3-dev \
+	python \
+	python-yaml \
+	vim \
+	net-tools \
+	bridge-utils \
+	iputils-ping \
+	iproute2 \
+	psmisc \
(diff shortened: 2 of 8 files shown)


framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
```
arnout, line 8: License header is missing

tools/docker/all/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+	python-yaml \
+	vim \
+	net-tools \
+	bridge-utils \
+	iputils-ping \
+	iproute2 \
+	psmisc \
+	clang-format \
+	tmux \
+    && rm -rf /var/lib/apt/lists/*
```
arnout, line 40: I would add /bin/bash as the entrypoint.

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+import socket
+import argparse
+import sys
```
arnout: Not used.

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+
+import socket
+import argparse
+import sys
+import threading
+import binascii
```
arnout: Not used.

I suggest to use flake8 to detect such issues.

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+
+def handle(host, port, respDict):
+    id = threading.get_ident()
+    print("{}: Listening on {}:{}".format(id, host, port))
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind((host, port))
+        s.listen()
+        conn, addr = s.accept()
+        with conn:
```
arnout: Any reason not to do

```
with s.accept() as (conn, addr):
```

?

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+            print("{}: {}@{} Connected by {}".format(id, host, port, addr))
```
arnout, line 20: Can we assume python 3.6 and write

```
print(f"{id}: {host}@{port} Connected by {addr}")
```

?

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+            while True:
+                try:
+                    data = conn.recv(1024)
+                except Exception as e:
+                    print("Exception {}".format(e))
+                    break
+                print("{}@{} Received: {}".format(host, port, data))
```
arnout, line 27: Ideally, you should buffer incoming data and split on `\r\n` because the TCP stream has no packet boundaries. However, since UCC will not send a new command until it received a response from the previous one, it will work as is. So OK.

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+                resp = [b'Status,RUNNING', b'Status,COMPLETE'] # default response
+                for key in respDict:
+                    if key in str(data):
```
arnout, line 30: That is a very roundabout way of working... I would do something more proper, like:
```
command_argv = str(data).strip().split(',')
command = command_argv[0].lower()
resp = respDict.get(command, resp)
```

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+                        resp = respDict[key] # specific resonse
+                for r in resp:
+                    print("{}@{} Sending: {}".format(host, port, r))
+                    conn.sendall(r)
+    print("{}: {}@{} Exiting!".format(id, host, port))
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--ports", nargs='+', default=[7000, 7001, 7002, 7003, 7004, 7005, 7006, 7007, 7008, 7009, 7010, 7011, 7012, 8888])
```
arnout: If you add `type=int` to the parameters of add_argument, you don't need the conversion below.

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+parser.add_argument("--host", default='127.0.0.1')
+parser.add_argument("--dst-alid", default="11:22:33:44:55:66")
+args = parser.parse_args()
+
+host = args.host
+ports = [int(p) for p in args.ports]
+alid = args.dst_alid
```
arnout: IMO there's no need to create local variables for these, just use args.host etc. directly.

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+
+respDict = {
```
arnout, line 45: IMHO it's more appropriate to define a global constant like this before the first function. It's also what I would do in C++.

framework/wfa/ucc-dummy.py
```diff
@@ -0,0 +1,63 @@
...
+    'ca_get_version' : [b'Status,RUNNING', b'Status,COMPLETE,vendor,dummy,model,dummy,version,1.0'],
+    'dev_get_parameter' : [b'Status,RUNNING', 'Status,COMPLETE,aLid,{}'.format(alid).encode()],
+}
+
+print("Starting threads on host {}, ports {}".format(host, ports))
+threads = list()
+
+for port in args.ports:
+    print("Creating thread {}@{}".format(host, port))
+    t = threading.Thread(target=handle, args=(host, port, respDict,))
```
arnout: Personally I'd have used an asyncio server instead of threads, but OK.

arnout (approved):
Lots of comments but all nitpicks or just general education.

The only must-fix thing is the license header in the python file.

framework/wfa/ucc-dummy.py
tomereli: Will it solve the fact that ctrl+c doesn't work?

framework/wfa/ucc-dummy.py
tomereli: I tried to add flake8 and it complains on things that aren't errors, probably bad integration with vscode.

framework/wfa/ucc-dummy.py
tomereli: I think its better to keep it as is, so it supports even python2.7. But good to know this is possible with new versions of python.

framework/wfa/ucc-dummy.py
tomereli: Not sure I understand to which lines you refer - can you add this with "insert suggestion"?

framework/wfa/ucc-dummy.py
tomereli: Problematic since respDict depends on args.

framework/wfa/ucc-dummy.py
tomereli: Feel free to change that at any time then

framework/wfa/ucc-dummy.py
arnout: I've never known flake8 to complain on things that aren't PEP8 coding style errors... Can you give a concrete example?

framework/wfa/ucc-dummy.py
arnout: Yes, f-strings are great.

Things will be even better in Python 3.8, where you can write:
```
while data := conn.recv(1024):
```
instead of the `while True`. Python is still adding interesting features.

framework/wfa/ucc-dummy.py
arnout: I guess it would. But I haven't actually used asyncio myself. At least not directly, I've used it through twisted but that hides all the gory details of coroutines and continuations. But basically, with asyncio you still have a single thread and you don't loose your tty association.

Issue #162, opened 2019-08-14
[TASK] Controller shouldn't trigger tasks for non-Intel agents


arnout:
## Description

We should remove the trigger for tasks which are intel specific in the controller for non-Intel agents.

We need to distinguish between Intel and non-Intel agents in the topology database - either with a new type, or with a flag. The conn_map should show this difference.

## Exit criteria

* Non-Intel agents are visible in conn_map.
* Non-Intel agents are marked differently in conn_map.
* No tasks are started for non-Intel agents, unless they can be done with the already-implemented EasyMesh flows.

## Detailed description

The current controller ("master") implementation assumes the agents are Intel agents, and follow the Intel protocol with Vendor-Specific messages. We want to reuse the existing implementation but still support non-Intel agents. Some of the tasks may be able to function without the Intel Vendor-Specific messages, but until they do, it makes no sense to start these tasks for non-Intel agents.

We want a simple way to have an operational controller with as much functionality as possible even if non-Intel agents exist in the network.

Therefore, non-Intel agents should be indicated as such in the database, and tasks should skip over them if they don't work standard Multi-AP messages.

ghost:
@arnout @morantr Is this issue still unfinished? If it's finished (or no longer needed), please close it.
If it's not yet fully completed, but the missing pieces are not needed for wired certification, please change the milestone.
If any work remains for this milestone, we need to know ASAP

arnout:
It is not done, but I think the current state is good enough - at least no VS messages are sent to non-Intel agents. So closing.

PR #176, opened 2019-08-20
controller: disable statistics polling task

tomereli:
The statistics polling task is not needed for controller certification
and only clatters the controller log with none useful log prints since
there are no responses from the slaves (since they are running in dummy
mode). Therefore, disable it by default.

Issue #162 

### DIFF
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -122,7 +122,23 @@ public:
         bool client_optimal_path_roaming_prefer_signal_strength = false;
     } sDbMasterSettings;
 
-    db(sDbMasterConfig &config_, beerocks::logging &logger_) : config(config_), logger(logger_) {}
+    db(sDbMasterConfig &config_, beerocks::logging &logger_) : config(config_), logger(logger_)
+    {
+        settings.enable_dfs_reentry &= config_.load_dfs_reentry;
+        settings.client_band_steering &= config_.load_client_band_steering;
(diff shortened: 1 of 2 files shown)


morantr:
Why didn't you disabled it from the controller config file instead?

tomereli:
> Why didn't you disabled it from the controller config file instead?

You are right - done.

controller/src/beerocks/master/db/db.h
```diff
@@ -122,7 +122,23 @@ class db {
         bool client_optimal_path_roaming_prefer_signal_strength = false;
     } sDbMasterSettings;
 
-    db(sDbMasterConfig &config_, beerocks::logging &logger_) : config(config_), logger(logger_) {}
+    db(sDbMasterConfig &config_, beerocks::logging &logger_) : config(config_), logger(logger_)
+    {
+        settings.enable_dfs_reentry &= config_.load_dfs_reentry;
```
arnout, line 127: This should be `&&=`. (nitpick)

arnout (approved):
One single nitpick.

Also small grammar mistake in the second commit's log: "does sets" -> sets.

controller/src/beerocks/master/db/db.h
tomereli: &&= is not a valid expression it appears
>home/tbeliyah/work/dev1/prplMesh/controller/src/beerocks/master/db/db.h:127:39: error: expected primary-expression before '=' token                                                                        
         settings.enable_dfs_reentry &&= config_.load_dfs_reentry;

PR #177, opened 2019-08-20
controller: wfa_ca: implement initial testbed commands handlers

tomereli:
This PR adds implementation to DEV_RESET_DEFAULT, DEV_GET_INFO, DEV_GET_PARAMETER and adds a topology query handler. All are needed by the first MCUT test case as defined in EasyMesh test plan 1.0 section 5.3.1.

### DIFF
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -87,6 +87,8 @@ wfa_ca::eWfaCaCommand wfa_ca::wfa_ca_command_from_string(std::string command)
         return eWfaCaCommand::DEV_SEND_1905;
     } else if (command == "DEV_SET_CONFIG") {
         return eWfaCaCommand::DEV_SET_CONFIG;
+    } else if (command == "DEV_GET_PARAMETER") {
+        return eWfaCaCommand::DEV_GET_PARAMETER;
     }
 
     return eWfaCaCommand::WFA_CA_COMMAND_MAX;
@@ -286,8 +288,19 @@ void wfa_ca::handle_wfa_ca_message(
         break;
     }
     case eWfaCaCommand::DEV_GET_PARAMETER: {
-        // TODO
-        reply(sd, cmdu_tx, eWfaCaStatus::INVALID, "unimplemented command");
+
+        // For controller certification, the following command is received to get the ALid:
+        // DUT (127.0.0.1:5000) ---> dev_get_parameter,program,map,parameter,ALid
+
+        if (!reply(sd, cmdu_tx, eWfaCaStatus::RUNNING)) {
+            LOG(ERROR) << "failed to send reply";
+            break;
+        }
+
+        std::string alid;
+        if (!net::network_utils::linux_iface_get_mac("br-lan", alid))
+            reply(sd, cmdu_tx, eWfaCaStatus::INVALID, "FAIL");
+        reply(sd, cmdu_tx, eWfaCaStatus::COMPLETE, std::string("aLid,") + alid);
         break;
     }
     case eWfaCaCommand::DEV_RESET_DEFAULT: {
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -272,8 +272,17 @@ void wfa_ca::handle_wfa_ca_message(
         break;
     }
     case eWfaCaCommand::DEVICE_GET_INFO: {
-        // TODO
-        reply(sd, cmdu_tx, eWfaCaStatus::INVALID, "unimplemented command");
+
+        // send back first reply
+        if (!reply(sd, cmdu_tx, eWfaCaStatus::RUNNING)) {
+            LOG(ERROR) << "failed to send reply";
+            break;
+        }
+
+        // Send back second reply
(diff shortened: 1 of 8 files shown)


controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -313,7 +313,7 @@ void wfa_ca::handle_wfa_ca_message(
         }
 
         if (params.find("type") != params.end()) {
-            if (params["type"] != "test bed" && params["type"] != "dut") {
+            if (params["type"] != "test bed" && params["type"] != "DUT") {
```
morantr, line 372: Basically, the CAPI commands should be **not** case sensitive with some exceptions like SSID, so maybe it was better to downcase the content of params["type"]. But on the other hand, it will be ugly and if the UCC sends it all the time up-case, then I guess it is ok.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -240,7 +240,43 @@ void wfa_ca::handle_wfa_ca_message(
...
+        */
+        
+        // send back first reply
+        if (!reply(sd, cmdu_tx, eWfaCaStatus::RUNNING)) {
+            LOG(ERROR) << "failed to send reply";
+            break;
+        }
+
+        // Send back second reply
+        reply(sd, cmdu_tx, eWfaCaStatus::COMPLETE, "vendor,Intel,model,prplMesh,version,1.4.0");
```
morantr: I would replace the hard codded version to the macro BEEROCKS_VERSION.
see the implementation of CA_GET_VERSION.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -280,7 +282,42 @@ void wfa_ca::handle_wfa_ca_message(
...
+        * 
+        * Param Name    | Values                        | Description
+        * ----------------------------------------------------------------------------------------
+        * "program"     | string (MAP)                  | Program name
+        * "ruid"        | string                        | A radio Unique ID
+        * "SSID"        | string                        | The SSID that fronthaul AP is configured. When SSID is not specified for the parameter
+                                                          "macaddr", it implies that device should return backhaul STA's MAC address. SSID is also
+                                                          not required for the parameter "ALid"
+        * 
+        * Return Values:
```
morantr: Missing param name "parameter" on the table.
As well the description exceeds 100 notes. Please fix this on other tables you added. I look better when open the code in github.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -280,7 +282,42 @@ void wfa_ca::handle_wfa_ca_message(
...
+        // DUT (127.0.0.1:5000) ---> dev_get_parameter,program,map,parameter,ALid
+
+        if (!reply(sd, cmdu_tx, eWfaCaStatus::RUNNING)) {
+            LOG(ERROR) << "failed to send reply";
+            break;
+        }
+    
+        std::string alid;
+        if (!net::network_utils::linux_iface_get_mac("br-lan", alid))
+            reply(sd, cmdu_tx, eWfaCaStatus::INVALID, "FAIL");
```
morantr, line 302: Since this failure is not related to command validity and occurs after you send status RUNNING, I should be ERROR instead of INVALID.
Also please add debug log, and break.

morantr (dismissed):

controller/src/beerocks/master/wfa_ca.cpp
tomereli: It sends it in uppercase

arnout: OK for this one, but if we're going to add more of this kind of comparisons, I think we should implement a helper function `get_param(std::string type, std::string default = "")` that handles the condition and converts to lowercase.

controller/src/beerocks/master/wfa_ca.cpp
arnout: Not entirely related, but: we can't keep vendor as Intel. We will have to make it configurable (also in the WSC attributes etc.), either with a define or from the config file.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -280,7 +282,42 @@ void wfa_ca::handle_wfa_ca_message(
...
+        // For controller certification, the following command is received to get the ALid:
+        // DUT (127.0.0.1:5000) ---> dev_get_parameter,program,map,parameter,ALid
+
+        if (!reply(sd, cmdu_tx, eWfaCaStatus::RUNNING)) {
+            LOG(ERROR) << "failed to send reply";
+            break;
+        }
+    
+        std::string alid;
+        if (!net::network_utils::linux_iface_get_mac("br-lan", alid))
```
arnout, line 301: We *really* need a better way to get the AL-MAC address... But that's not for this PR.

arnout (approved):
Typo in last commit message: topolohy -> topology

Moran made all the relevant change requests, so I'll approve already.

arnout:
I've made the requested changes and rebased. @morantr please re-review.

controller/src/beerocks/master/wfa_ca.cpp
tomereli: removed all comments since they violate license

PR #179, opened 2019-08-20
controller: wfa_ca: add validation checks for input

CoralMalachi:
The MCUT (MultiAP Controller Under Test) gets a string command from
WTS UCC (Wifi Test Suite Unified CAPI Console).

The `tlv_value` fields' input should be inside curly brackets
and in hexadecimal format.

The following changes consist of two validation checks:
1.Check for balanced parentheses input (only curly brackets).
2.Check if value string conforms to hex notation.

### DIFF
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -73,6 +73,53 @@ static bool is_hex_notation(const std::string &s)
            s.find_first_not_of("0123456789abcdefABCDEF", 2) == std::string::npos;
 }
 
+static bool input_validation_check(std::string &command_str, std::string &param_name,
+                                   std::string &err_string)
+{
+    int len_command_str     = command_str.length();
+    bool prev_space         = false;
+    bool prev_curly_bracket = false;
+    int count_open_brackets = 0;
+    for (int i = 0; i < len_command_str; i++) {
+        auto c = command_str[i];
+        if (c == '{') {
+            prev_curly_bracket = true;
+            prev_space         = false;
+            count_open_brackets++;
+            continue;
+        }
+        if (count_open_brackets == 0) {
+            err_string = "Unbalanced curly brackets in value of param " + param_name + "'";
+            return false;
+        } else if (c == '}') {
+            count_open_brackets--;
+            if (count_open_brackets < 0) {
+                err_string = "Unbalanced curly brackets in value of param " + param_name + "'";
+                return false;
+            }
+            prev_space         = false;
+            prev_curly_bracket = false;
+        } else if (c == ' ') {
+            prev_space         = true;
+            prev_curly_bracket = false;
+        } else {
+            if (prev_space || prev_curly_bracket) {
+                auto token = command_str.substr(i, command_str.find_first_of(" }", i) - i);
+                token.pop_back();
+                //Check if a given string conforms to hex notation.
+                if (is_hex_notation(token) == false) {
+                    err_string = "hex notation test failed in value of param " + param_name + "'";
+                    return false;
+                }
+                i                  = i + token.length() - 1;
+                prev_space         = false;
+                prev_curly_bracket = false;
+            }
+        }
+    }
+    return true;
+}
+
 /**
  * @brief Convert string of WFA-CA command into enum
  * 
@@ -173,49 +220,11 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
             return false;
         }
         if ((*std::next(it))[0] == '{') {
-            int len_command_str     = (*std::next(it)).length();
-            bool prev_space         = false;
-            bool prev_curly_bracket = false;
-            int count_open_brackets = 0;
+            //check if mac address in curly brackets are in hex notation and
+            //make sure that the curly brackets are balanced.
             std::string command_str = *std::next(it);
-            for (int i = 0; i < len_command_str; i++) {
-                auto c = command_str[i];
-                if (c == '{') {
-                    prev_curly_bracket = true;
-                    prev_space         = false;
-                    count_open_brackets++;
-                    continue;
-                }
-                if (count_open_brackets == 0) {
-                    err_string = "Unbalanced curly brackets in value of param " + param_name + "'";
-                    return false;
-                } else if (c == '}') {
-                    count_open_brackets--;
-                    if (count_open_brackets < 0) {
-                        err_string =
-                            "Unbalanced curly brackets in value of param " + param_name + "'";
-                        return false;
-                    }
-                    prev_space         = false;
-                    prev_curly_bracket = false;
-                } else if (c == ' ') {
-                    prev_space         = true;
-                    prev_curly_bracket = false;
-                } else {
-                    if (prev_space || prev_curly_bracket) {
-                        auto token = command_str.substr(i, command_str.find_first_of(" }", i) - i);
-                        token.pop_back();
-                        //Check if a given string conforms to hex notation.
-                        if (is_hex_notation(token) == false) {
-                            err_string =
-                                "hex notation test failed in value of param " + param_name + "'";
-                            return false;
-                        }
-                        i                  = i + token.length() - 1;
-                        prev_space         = false;
-                        prev_curly_bracket = false;
-                    }
-                }
+            if (!input_validation_check(command_str, param_name, err_string)) {
+                return false;
             }
         }
         params[param_name] = *std::next(it);
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -67,6 +67,12 @@ const std::string wfa_ca::wfa_ca_status_to_string(eWfaCaStatus status)
     return std::string();
 }
 
+static bool is_hex_notation(const std::string &s)
+{
+    return s.compare(0, 2, "0x") == 0 && s.size() > 2 &&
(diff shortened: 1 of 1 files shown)


controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -67,6 +67,12 @@ const std::string wfa_ca::wfa_ca_status_to_string(eWfaCaStatus status)
     return std::string();
 }
 
+static bool is_hex_notation(std::string const &s)
```
morantr: I would remove the "is" prefix.
Also, I think te convention in our code is to put the const before the type (nitpick)

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -166,6 +175,60 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
...
+            int count_open_brackets = 0;
+            std::string command_str = *std::next(it);
+            for (int i = 0; i < len_command_str; i++) {
+                auto c = command_str[i];
+                LOG(ERROR) << "char: " << c;
+                if (c == '{') {
+                    prev_curly_bracket = true;
+                    prev_space         = false;
+                    count_open_brackets++;
+                } else {
```
morantr: You could avoid nesting by using "continue":
```
if (c == '{') {
    // code
    continue;
}

// rest of the code
```

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -166,6 +175,60 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
...
+
+                    if (count_open_brackets == 0) {
+                        err_string =
+                            "The following string is not valid (curly brackets test failed) : '" +
```
morantr: I would change it to something like "Unbalanced curly brackets in value of param '<param_name>'"
No need to mention that the string is invalid because the return code is "INVALID".

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -166,6 +175,60 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
...
+                            *std::next(it);
+                        return false;
+                    }
+                    if (c == '}') {
+                        count_open_brackets--;
+                        if (count_open_brackets < 0) {
+
+                            err_string = "The following string is not valid (curly brackets test "
```
morantr: same here

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -166,6 +175,60 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
...
+                                         "failed) : '" +
+                                         *std::next(it);
+                            return false;
+                        }
+                        prev_space = false;
+                    } else if (c == ' ') {
+                        prev_space = true;
+
+                    } else {
+                        if (prev_space == true || prev_curly_bracket == true) {
```
morantr: Here you could also avoid nesting:
```
if (!condition) {
    // continue;
}
// rest of the code
```

morantr: IMO the comparison to true/false of booleans is not necessary.
if (VAR == true) -> if (VAR)
if (VAR == false) -> if (!VAR)

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -166,6 +175,60 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
...
+                            auto token =
+                                command_str.substr(i, command_str.find_first_of(" }", i) - i);
+                            token.pop_back();
+                            //Check if a givven string conforms to hex notation.
```
morantr: givven -> given

morantr (changes requested):
Regarding the brackets balance check, I see what you did there, but I think it is over complicated.
I would instead split the string by spaces and get a vector for example:
"{0x00904C2A11C2", "0x14", "{0x51", "{0x00", "0x00}}"
Check that the first element starts with '{' and the last one end s with '}'
Then, use std::count() to count how many '{' and '}' to create a balance of the brackets,
if the balance if negative return error.
Then check the hex notation of the value.
I think it will be more easy to read.

Second, I think this should be in a static file function.

Also, I know that I thought that it should be checked in the parse function, but yesterday I pushed a commit that handles specifically dev_send_1905 command, and I think it will be more fit to check it the function get_send_1905_1_tlv_hex_list() of the followup commit:
https://github.com/prplfoundation/prplMesh/pull/180/commits/f8bd822b5189616984d97d1e3c6a5746e2d9b9c9

So meanwhile fix the PR, and split it brackets balance check to a different function, and when my commit will be merged, use it in the correct places.

controller/src/beerocks/master/wfa_ca.cpp
tomereli: Agree on the const, disagree about the is

controller/src/beerocks/master/wfa_ca.cpp
CoralMalachi: Ok, I'll change this .

CoralMalachi:
> Regarding the brackets balance check, I see what you did there, but I think it is over complicated.
> I would instead split the string by spaces and get a vector for example:
> "{0x00904C2A11C2", "0x14", "{0x51", "{0x00", "0x00}}"
> Check that the first element starts with '{' and the last one end s with '}'
> Then, use std::count() to count how many '{' and '}' to create a balance of the brackets,
> if the balance if negative return error.
> Then check the hex notation of the value.
> I think it will be more easy to read.
> 
> Second, I think this should be in a static file function.
> 
> Also, I know that I thought that it should be checked in the parse function, but yesterday I pushed a commit that handles specifically dev_send_1905 command, and I think it will be more fit to check it the function get_send_1905_1_tlv_hex_list() of the followup commit:
> [f8bd822](https://github.com/prplfoundation/prplMesh/commit/f8bd822b5189616984d97d1e3c6a5746e2d9b9c9)
> 
> So meanwhile fix the PR, and split it brackets balance check to a different function, and when my commit will be merged, use it in the correct places.

Do you mean to do the brackets balance check first and just after that do the hex notation check?

arnout:
When looking at #180, I actually wonder if it really makes sense to have a separate check. The conversion from hex to uint8_t implicitly verifies that it really is a hex number.

As for the curly braces: we don't really care about them. It is not a requirement that we fail if the curly braces matching is done incorrectly on input. For us, it is sufficient to just remove them, like is done in #180.

@tomereli @morantr  please comment.

morantr:
> When looking at #180, I actually wonder if it really makes sense to have a separate check. The conversion from hex to uint8_t implicitly verifies that it really is a hex number.
> 
> As for the curly braces: we don't really care about them. It is not a requirement that we fail if the curly braces matching is done incorrectly on input. For us, it is sufficient to just remove them, like is done in #180.
> 
> @tomereli @morantr please comment.

The hex notation check is definitely necessary since in #180 I use std::stoi() that if it will fail, it will throw an exception, so it is a good idea to do add hex notation check before converting it to an integer.
Regarding, the curly bracket balance check, yes it is a "nice to have check", and the question is if we want to validate that input.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -147,6 +153,7 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
                           std::unordered_map<std::string, std::string> &params,
                           std::string &err_string)
 {
+
```
tomereli: Extra line break

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -157,7 +164,9 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
     for (auto it = std::next(command_tokens.begin()); it != command_tokens.end(); it++) {
         std::string param_name = *it;
         std::transform(param_name.begin(), param_name.end(), param_name.begin(), ::tolower);
+
         if (params.find(param_name) != params.end()) {
+
```
tomereli: Here as well

tomereli (approved):

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -166,6 +175,60 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
...
+                    prev_space         = false;
+                    count_open_brackets++;
+                    continue;
+                }
+
+                if (count_open_brackets == 0) {
+                    err_string = "Unbalanced curly brackets in value of param " + param_name + "'";
+                    return false;
+                }
+                if (c == '}') {
```
simonbaren: else if

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -166,6 +175,60 @@ bool wfa_ca::parse_params(const std::vector<std::string> &command_tokens,
...
+
+                        err_string =
+                            "Unbalanced curly brackets in value of param " + param_name + "'";
+                        return false;
+                    }
+                    prev_space         = false;
+                    prev_curly_bracket = false;
+                    continue;
+                }
+                if (c == ' ') {
```
simonbaren: else if

morantr (changes requested):
I'm not approving it now. There are still changes need to be done.
Also, the validation check could be in a more proper place after #180 will be merged.
@tomereli don't rush to merge it.

arnout:
As discussed, these checks got included in #180.

PR #180, opened 2019-08-20
Feature/add support capi command dev send 1905 with tlvs

morantr:
Add support on CAPI command DEV_SEND_1905 with TLV.
Parse the command with hex values, create CMDU of it and send it.

### DIFF
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -44,6 +44,32 @@ static std::string check_forbidden_chars(const std::string &str)
 }
 
 /**
+ * @brief Checks that the given string is in a hex notation
+ * 
+ * @param[in] str String to check hex notation.
+ * @return true if string is valid hex notation, else false.
+ */
+static bool validate_hex_notation(const std::string &str)
+{
+    // Each value must start with 0x
+    if (str.substr(0, 2) != "0x") {
+        return false;
+    }
+
+    // The number of digits must be even
+    if (str.size() % 2 != 0) {
+        return false;
+    }
+
+    // Check if string has non-hex character
+    if (str.find_first_not_of("0123456789abcdefABCDEF", 2) != std::string::npos) {
+        return false;
+    }
+
+    return true;
+}
+
+/**
  * @brief Convert enum of WFA-CA status to string.
  * 
  * @param[in] status Enum of WFA-CA status. 
@@ -383,6 +409,14 @@ void wfa_ca::handle_wfa_ca_message(
             break;
         }
 
+        // TLV check
+        std::list<tlv_hex_t> tlv_hex_list;
+        if (!get_send_1905_1_tlv_hex_list(tlv_hex_list, params, err_string)) {
+            LOG(ERROR) << err_string;
+            reply(sd, cmdu_tx, eWfaCaStatus::INVALID, err_string);
+            break;
+        }
+
         // Send back first reply
         if (!reply(sd, cmdu_tx, eWfaCaStatus::RUNNING)) {
             LOG(ERROR) << "failed to send reply";
@@ -392,16 +426,37 @@ void wfa_ca::handle_wfa_ca_message(
         // Check if additional tlvs were added to command. If yes, then needs to parse the tlv and
         // send it as it.
         if (mandatory_params_num < params.size()) {
-
             if (!cmdu_tx.create(g_mid, ieee1905_1::eMessageType(message_type))) {
                 LOG(ERROR) << "cmdu creation of type 0x" << message_type_str << ", has failed";
                 reply(sd, cmdu_tx, eWfaCaStatus::ERROR, err_internal);
                 break;
             }
 
-            // TODO: Add the TLVs from the command
+            // We need to fill in CmduMessage's class vector. However, we really don't care about
+            // the classes, and we don't want any swapping to be done. Still, we need something
+            // to make sure the length of the tlvs is taken into account.
+            // Therefore, use the tlvPrefilledData class, which can update the end pointer.
+            auto tlvs = cmdu_tx.addClass<tlvPrefilledData>();
+            if (!tlvs) {
+                LOG(ERROR) << "addClass tlvPrefilledData has failed";
+                reply(sd, cmdu_tx, eWfaCaStatus::ERROR, err_internal);
+                break;
+            }
 
-            son_actions::send_cmdu_to_agent(agent_sd, cmdu_tx);
+            if (!tlvs->add_tlvs_from_list(tlv_hex_list, err_string)) {
+                LOG(ERROR) << err_string;
+                reply(sd, cmdu_tx, eWfaCaStatus::ERROR, err_string);
+                break;
+            }
+
+            LOG(INFO) << "Send TLV CMDU " << message_type << " with " << tlv_hex_list.size()
+                      << " TLVs, length " << cmdu_tx.getMessageLength();
+
+            if (!son_actions::send_cmdu_to_agent(agent_sd, cmdu_tx)) {
+                LOG(ERROR) << "Failed to send CMDU";
+                reply(sd, cmdu_tx, eWfaCaStatus::ERROR, err_internal);
+                break;
+            }
 
             // If not using TLVs from the command, create CMDU from existing buffer
         } else {
@@ -409,17 +464,17 @@ void wfa_ca::handle_wfa_ca_message(
             if (!certification_tx_buffer) {
                 LOG(ERROR) << "certification_tx_buffer is not allocated";
             }
-            ieee1905_1::CmduMessageTx certification_cmdu_tx(
+            ieee1905_1::CmduMessageTx external_cmdu_tx(
                 certification_tx_buffer.get() + sizeof(beerocks::message::sUdsHeader),
                 message::MESSAGE_BUFFER_LENGTH - sizeof(beerocks::message::sUdsHeader));
 
-            if (!create_cmdu(certification_cmdu_tx, ieee1905_1::eMessageType(message_type))) {
+            if (!create_cmdu(external_cmdu_tx, ieee1905_1::eMessageType(message_type))) {
                 LOG(ERROR) << "Failed to create CMDU";
                 reply(sd, cmdu_tx, eWfaCaStatus::ERROR, err_internal);
                 break;
             }
 
-            if (!son_actions::send_cmdu_to_agent(agent_sd, certification_cmdu_tx)) {
+            if (!son_actions::send_cmdu_to_agent(agent_sd, external_cmdu_tx)) {
                 LOG(ERROR) << "Failed to send CMDU";
                 reply(sd, cmdu_tx, eWfaCaStatus::ERROR, err_internal);
                 break;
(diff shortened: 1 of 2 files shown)


arnout:
I thought: there has to be a better way to do this, so I tried to implement a better way. I've pushed this as a fixup commit. It's not finished though, it needs to be cleaned up a bit. Also, the first commit is no longer needed with my change, so that has to be reverted.

As I said, it's not finished yet, but I wanted to push it before the weekend.

morantr:
@arnout, I liked how you have changed it, wish I would have thought about it myself.
I allowed myself to squash your changes and remove the first commit, and add a fixup that change the code to how I think it should be, based on your change.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,32 @@ static std::string check_forbidden_chars(const std::string &str)
     return forbidden_chars_found;
 }
 
+/**
+ * @brief Checks that the given string is in a hex notation
+ * 
+ * @param[in] str String to check hex notation.
+ * @return true if string is valid hex notation, else false.
+ */
+static bool validate_hex_notation(const std::string &str)
```
arnout, line 52: This is an excellent improvement over the initial implementation by Coral in #179.

While merging, I'll add Coral's Sob for completeness.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+                    break;
+                } else {
+                    err_string = "missing param name '" + lookup_str + "'";
+                    return false;
+                }
+            }
+
+            // Type
+            if (tlv_member_idx == 0) {
+                tlv_hex.type = &it->second;
```
arnout, line 659: We could actually do the conversion to `uint8_t` here, i.e. have uint8_t and uint16_t as the first two members of `tlv_hex_t`. Actually, we could even do the same for the value member and do the conversion immediately. Hm, but that would require creating a buffer in which to write first, so maybe not such a great idea...

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+{
+    static const int max_tlv                          = 256; // Magic number
+    size_t tlv_idx                                    = 0;
+    bool skip                                         = false;
+    bool finish                                       = false;
+    static const std::vector<std::string> tlv_members = {"tlv_type", "tlv_length", "tlv_value"};
+
+    do {
+        tlv_hex_t tlv_hex;
+        for (size_t tlv_member_idx = 0; tlv_member_idx < tlv_members.size(); tlv_member_idx++) {
```
arnout, line 638: There are so many conditions on `tlv_member_idx` inside this loop that I wonder if it's really worth to make a loop for it.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+        tlv_idx++;
+
+        if (skip) {
+            skip = false;
+            continue;
+        }
+
+        tlv_hex_list.push_back(tlv_hex);
+
+    } while (tlv_idx < max_tlv);
```
arnout, line 715: Is it worth to have this explicit stop condition? We can just make it an endless loop, no? There is no way that it can really be an endless loop because we stop as soon as one of the parameters is missing.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+            for (auto value : values) {
+                if (!validate_hex_notation(value)) {
+                    err_string = "param name '" + lookup_str + "' has invalid hex notation value";
+                    return false;
+                }
+
+                // Type and Length fields must have fixed size:
+
+                // Type
+                if (tlv_member_idx == 0 && value.size() != 4) {
```
arnout, line 690: AFAIU, the 0x really is optional... So size==4 is not a correct check. In practice, though, UCC really always sends these with the 0x so it should be fine.

Maybe we can change the validate_hex_notation() function to return the number of bytes (since it already deals with the 0x).

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,32 @@ static std::string check_forbidden_chars(const std::string &str)
...
+ * @brief Checks that the given string is in a hex notation
+ * 
+ * @param[in] str String to check hex notation.
+ * @return true if string is valid hex notation, else false.
+ */
+static bool validate_hex_notation(const std::string &str)
+{
+    // Each value must start with 0x
+    if (str.substr(0, 2) != "0x") {
+        return false;
```
arnout, line 56: I believe the 0x is optional. If you look at the examples (in general, not specifically for send 1905), some of them do have a 0x and some of them don't.

We could make the argument non-const and remove the 0x here. That would simplify the rest of the code. It's a bit ugly though.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+            } else if (tlv_member_idx == 1) {
+                tlv_hex.length = &it->second;
+
+                // Value
+            } else if (tlv_member_idx == 2) {
+                // The hex values contains '{' and '}' delimiters, but we don't care about them.
+                it->second.erase(std::remove(it->second.begin(), it->second.end(), '{'),
+                                 it->second.end());
+                it->second.erase(std::remove(it->second.begin(), it->second.end(), '}'),
+                                 it->second.end());
```
arnout, line 671: We also need to eliminate the `:`, because a MAC address may be passed in colon notation.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+        *m_buff_ptr__ = uint8_t(length >> 8);
+        m_buff_ptr__++;
+        *m_buff_ptr__ = uint8_t(length);
+        m_buff_ptr__++;
+
+        auto values = string_utils::str_split(*tlv.value, ' ');
+        for (auto value : values) {
+            // iterate every to chars (1 octet) and write it to buffer
+            // start with char_idx = 2 to skip "0x"
+            for (size_t char_idx = 2; char_idx < value.size(); char_idx += 2) {
```
arnout, line 755: Again, I believe the 0x is optional.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+
+                // Type and Length fields must have fixed size:
+
+                // Type
+                if (tlv_member_idx == 0 && value.size() != 4) {
+                    err_string = "type field is too large: " + value;
+                    return false;
+
+                    // Length
+                } else if (tlv_member_idx == 1 && value.size() != 6) {
```
arnout, line 695: I'm not sure if this will always be 4 digits. It's possible that UCC will just print the significant digits. Although the examples indeed always have 4 digits.

arnout (approved):
All my comments are nitpicks. The comments about the 0x may be important, but we can handle that when it turns out that UCC indeed does create hex numbers without 0x.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -669,3 +724,160 @@ bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessag
...
+                    err_string = "length field is too large: " + value;
+                    return false;
+                }
+            }
+        }
```
arnout, line 700: One thing I forgot to mention: it would be a good idea to also validate that the length corresponds to the number of bytes in the value. During my testing, I accidentally made mistakes with that, and it causes ieee1905_transport to barf.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I think it is better this way, instead of comparing it to string every time, which will create a larger code.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I prefer that this function will focus on parsing, and another one will do the copy to the buffer.

controller/src/beerocks/master/wfa_ca.cpp
arnout: What I meant is to remove the loop. There is not that much that is in common for the three members. In terms of lines of code, just removing the conditions on tlv_member_idx may already reduce the number of lines.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I think it could happen. I didn't see any limitations on TLVs number.
Besides, it is harmless, and generally good practice to make a stop condition, and also prevent from static code analysis tools to complain about potential endless loop like klocwork which we use on Intel.

controller/src/beerocks/master/wfa_ca.cpp
morantr: "0x" prefix is not optional. It is mentioned on CAPI specifications that "0x" prefix is an example to hex string.
Did you see any DEV_SEND_1905 with TLVs that show otherwise?

controller/src/beerocks/master/wfa_ca.cpp
morantr: I saw it also on CAPI specifications, but I believe it is a mistake because clearly aa::bb:cc:dd:ee:ff is not hex format.
If in the end, we will see that I've been mistaken, I will fix it.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I don't think it is optional. If we will see that you have been right, I will fix it.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I think it must have 4 digits because how the data is presented in the command is the way we should write it to the buffer. Again if I'm wrong, I will fix in the future.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I believe I did it in the follow-up commit.

controller/src/beerocks/master/wfa_ca.cpp
arnout: > static code analysis tools to complain about potential endless loop

Indeed, that's a good reason to do this.

controller/src/beerocks/master/wfa_ca.cpp
arnout: The "TLV Value" example has `Hex value: Ex.: {11:22:33:44:55:66 aa:bb:cc:dd:ee:f0}`.

The "Data types" section says "String using the 0-9 and A-F characters, e.g 1B234A3467"

PR #184, opened 2019-08-22
autoconfig: change WSC M1 CMDU TLVs order

tomereli:
While testing with reference CTT agent in the testbed, it was seen that
the agent sends the WSC autoconfig CMDU in a different order than what
the prplMesh controller expects - radio identifier TLV followed by WSC
M1 TLV, which caused the prplMesh controller to fail on parsing.

This commit changes the order of the created M1 to align to the golden
unit.

In addition, this commit changes the detection of looped back WSC
messages in the controller and agent WSC handlers as follows:

Autoconfig WSC message sent by an agent contains one AP Radio Basic
Capabilities TLV and one WSC TLV containing M1.
Autoconfig WSC message sent by a controller contains one AP Radio
Identifier TLV and one or more WSC TLVs containing M2.

Based on this, the controller can detect that a message is looped back
if it contains an AP Radio Identifier TLV. This is done using
the hasTlv() API from CmduMessage class.

The radio agent can detect that a message is either looped back or sent
by another agent (only for WSC CMDUs) if the message contains an AP
Radio Basic Capabilities. This is done as well using the hasTlv() API.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -3697,6 +3697,13 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             return false;
         }
 
+        // All attributes which are not explicitely set below are set to
+        // default by the TLV factory, see WSC_Attributes.yml
+        if (!autoconfig_wsc_add_m1()) {
+            LOG(ERROR) << "Failed adding WSC M1 TLV";
+            return false;
+        }
+
         auto radio_basic_caps = cmdu_tx.addClass<wfa_map::tlvApRadioBasicCapabilities>();
         if (!radio_basic_caps) {
             LOG(ERROR) << "Error creating TLV_AP_RADIO_BASIC_CAPABILITIES";
@@ -3732,13 +3739,6 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
                 LOG(ERROR) << "add_operating_classes_info_list failed";
                 return false;
             }
-        }
-
-        // All attributes which are not explicitely set below are set to
-        // default by the TLV factory, see WSC_Attributes.yml
-        if (!autoconfig_wsc_add_m1()) {
-            LOG(ERROR) << "Failed adding WSC M1 TLV";
-            return false;
         }
 
         auto vs = cmdu_tx.add_vs_tlv(ieee1905_1::tlvVendorSpecific::eVendorOUI::OUI_INTEL);
@@ -4424,8 +4424,8 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
     // way to distinguish them: the M1 message has the AP_Radio_Basic_Capabilities TLV,
     // while the M2 has the AP_Radio_Identifier TLV.
     // If this is a looped back M1 CMDU, we can treat is as handled successfully.
-    if (cmdu_rx.getNextTlvType() == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES)) {
-        LOG(DEBUG) << "looped back M1 CMDU";
+    if (cmdu_rx.hasTlv(uint8_t(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES))) {
+        LOG(DEBUG) << "agent M1 CMDU, ignore";
         return true;
     }
 
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -601,8 +601,10 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
     // way to distinguish them: the M1 message has the AP_Radio_Basic_Capabilities TLV,
     // while the M2 has the AP_Radio_Identifier TLV.
     // If this is a looped back M2 CMDU, we can treat is as handled successfully.
-    if (cmdu_rx.getNextTlvType() == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER))
+    if (cmdu_rx.hasTlv(uint8_t(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER))) {
+        LOG(DEBUG) << "CMDU contains radio identifier TLV, ignore (looped back WSC M2 CMDU)";
         return true;
+    }
 
     LOG(DEBUG) << "Received AP_AUTOCONFIGURATION_WSC_MESSAGE";
     /**
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -31,6 +31,7 @@
 
 #include <beerocks/tlvf/beerocks_message_control.h>
 #include <tlvf/ieee_1905_1/eMessageType.h>
+#include <tlvf/ieee_1905_1/eTlvType.h>
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
@@ -591,7 +592,7 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1> 
 bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
                                                            ieee1905_1::CmduMessageRx &cmdu_rx)
 {
-    // Check if this is a M2 message that we sent to the agent, which was just looped back.
+    // Check if this is a M2 message that we sent to the agent, which was just looped back,
     // The M1 and M2 messages are both of CMDU type AP_Autoconfiguration_WSC. Thus,
     // when we send the M2 to the local agent, it will be published back on the local bus because
     // the destination is our AL-MAC, and the controller does listen to this CMDU.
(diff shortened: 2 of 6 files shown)


framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
```diff
@@ -69,6 +69,7 @@ class CmduMessage {
     uint8_t *getMessageBuff() const;
     bool getNextTlvType(eTlvType &tlvType) const;
     int getNextTlvType() const;
+    bool hasTlv(uint8_t tlvType) const;
```
arnout, line 72: Why isn't this `eTlvType`?

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
```diff
@@ -78,6 +79,21 @@ class CmduMessage {
 
 protected:
     void reset();
+    // packed TLV struct to help in the parsing of the packet
+#pragma pack(push, 1)
```
arnout, line 83: tlvf and most other places use `__attribute__((packed))`. Only ieee1905_transport.h uses the pragma.

I don't care much either way, but I'd like some consistency :-)

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
```diff
@@ -78,6 +79,21 @@ class CmduMessage {
...
+        uint8_t tlvType;
+        uint16_t tlvLength;
+
+    public:
+        uint8_t &type() { return tlvType; }
+
+        Tlv *next() { return (Tlv *)((uint8_t *)this + size()); }
+
+        size_t size() { return sizeof(Tlv) + ntohs(tlvLength); }
+    };
```
arnout, line 95: I'm surprised that such a class didn't exist already... Oh, actually it does - you just copied it from ieee1905_transport.h... I'm not a big fan of duplicating code. Any chance of moving it to a common header file instead?

I would call it TlvHeader, though. It's only the TL part, not the V. Also, it's not just any TLV, it's a 1905 TLV (e.g. WSC TLVs have two bytes for the type). So the 1905 aspect should be somewhere in the name as well.

framework/tlvf/AutoGenerated/src/CmduMessage.cpp
```diff
@@ -28,6 +28,36 @@ std::shared_ptr<BaseClass> CmduMessage::getClass(size_t idx) const
     }
 }
 
+bool CmduMessage::hasTlv(uint8_t tlvType) const
+{
+    uint8_t type = 0;
+    Tlv *tlvhdr;
+
+    // for already parsed TLVs, check the class vector (no need for endianess conversion)
+    for (size_t idx = 0; idx < m_class_vector.size(); idx++) {
```
arnout, line 37: Sorry for my lack of C++ skills, but why can't you use `for (auto it : m_class_vector)` here?

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
```diff
@@ -78,6 +79,21 @@ class CmduMessage {
...
+    protected:
+        uint8_t tlvType;
+        uint16_t tlvLength;
+
+    public:
+        uint8_t &type() { return tlvType; }
+
+        Tlv *next() { return (Tlv *)((uint8_t *)this + size()); }
+
+        size_t size() { return sizeof(Tlv) + ntohs(tlvLength); }
```
arnout, line 94: It's called length, not size.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -591,7 +592,7 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
 bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
                                                            ieee1905_1::CmduMessageRx &cmdu_rx)
 {
-    // Check if this is a M2 message that we sent to the agent, which was just looped back.
+    // Check if this is a M2 message that we sent to the agent, which was just looped back,
```
arnout, line 595: Why this change?

The capitalization of the following line should be changed as well then.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -608,16 +609,30 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
...
+    for (int i = 0; i < num_tlvs; i++) {
+        int type = cmdu_rx.getNextTlvType();
+        if (type == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES)) {
+            LOG(DEBUG) << "Found TLV_AP_RADIO_BASIC_CAPABILITIES TLV";
+            radio_basic_caps = cmdu_rx.addClass<wfa_map::tlvApRadioBasicCapabilities>();
+        } else if (type == int(ieee1905_1::eTlvType::TLV_WSC)) {
+            LOG(DEBUG) << "Found TLV_WSC TLV (assuming M1)";
+            tlvwscM1 = cmdu_rx.addClass<ieee1905_1::tlvWscM1>();
+        } else {
+            LOG(ERROR) << "Unexpected TLV " << type;
```
arnout, line 626: To be correct, it should always be allowed to have a vendor-specific TLV.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -608,16 +609,30 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
     * @brief Parse AP-Autoconfiguration WSC which should include one AP Radio Basic Capabilities
     * TLV and one WSC TLV containing M1
     */
-    auto radio_basic_caps = cmdu_rx.addClass<wfa_map::tlvApRadioBasicCapabilities>();
+    static const int num_tlvs                                              = 2;
+    std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_basic_caps = nullptr;
+    std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1                         = nullptr;
+    for (int i = 0; i < num_tlvs; i++) {
```
arnout, line 617: There's no reason to have a fixed limit. Just go through all TLVs.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3697,6 +3697,13 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             return false;
         }
 
+        // All attributes which are not explicitely set below are set to
+        // default by the TLV factory, see WSC_Attributes.yml
+        if (!autoconfig_wsc_add_m1()) {
+            LOG(ERROR) << "Failed adding WSC M1 TLV";
+            return false;
+        }
```
arnout, line 3705: I don't see the need for this change. The controller now supports any order of these two TLVs, so we don't need to match the order of the golden unit.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -601,8 +601,10 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
     // way to distinguish them: the M1 message has the AP_Radio_Basic_Capabilities TLV,
     // while the M2 has the AP_Radio_Identifier TLV.
     // If this is a looped back M2 CMDU, we can treat is as handled successfully.
-    if (cmdu_rx.getNextTlvType() == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER))
+    if (cmdu_rx.hasTlv(uint8_t(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER))) {
```
arnout, line 604: Since we can now accept TLVs in any order, this check is no longer needed.

Instead, we should add a case for TLV_AP_RADIO_IDENTIFIER in the swtich below, and exit there.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4424,8 +4424,8 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
     // way to distinguish them: the M1 message has the AP_Radio_Basic_Capabilities TLV,
     // while the M2 has the AP_Radio_Identifier TLV.
     // If this is a looped back M1 CMDU, we can treat is as handled successfully.
-    if (cmdu_rx.getNextTlvType() == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES)) {
-        LOG(DEBUG) << "looped back M1 CMDU";
+    if (cmdu_rx.hasTlv(uint8_t(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES))) {
+        LOG(DEBUG) << "agent M1 CMDU, ignore";
```
arnout, line 4428: I would feel more comfortable if we did the same in the agent as in the controller. It's not that complicated, right?

I realize that it's a bit of wasted effort, since with the full TLV list parsing that @Galvien is working on, it will no longer be needed. But we only need to handle for types of TLVs so it's just about 20 lines of code...

Note that with that, there is no need for hasTlv at all any more. Which means that all my comments on it no longer apply :-)

arnout (changes requested):
All of my comments are nitpick-style, so I'm actually OK with merging this as is. However, it anyway needs to be rebased, and I'd like to see the responses to my comments before merging. Therefore: Changes Requested.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: Was in a hurry and meant to change that, will do.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: I knew this comment will come - and yes, i'll move it to a common place:)

framework/tlvf/AutoGenerated/src/CmduMessage.cpp
tomereli: Apparently it is I who is lacking C++ skills since I didn't do this with range based loop to begin with, will change it.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: mistake, will revert

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Right, but since we don't support skip TLV yet, and are looking for 2 consecutive TLVs, it is considered unexpected.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: ok

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: IMHO hasTlv might still needed, at least in the agent, if first TLV is WSC  (if sometime in the future we decide to switch the order for some reason). In this case we will assume M2 and fail since it is M1 from the other agent or ourselves. Nothing serious but error log while its not really an error, hasTlv solves that. It is a real corner case but no real harm with using hasTlv to be 100% sure this is the CMDU we think it is, don't you think? I have no problem reverting this all together, not so sure what I prefer myself..

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Changed this a bit - if we see a vendor specific TLV, we assume it is Intel agent (same logic as today)

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: If hasTlv isn't needed in this PR in the end, just push it to a new branch without PR to save it for later, and remove it from this PR.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: OK

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: Removed hasTlv

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: removed `hasTlv()`

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: removed `hasTlv()`

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: OK - done, see #232

tomereli:
Superseded by #232 . Not deleting the branch so if in the future we will need `hasTlv()`, we can use it from here.

PR #226, opened 2019-08-22
ieee1905: transport: pad short cmdus to 64 bytes

tomereli:
Add padding to 64 bytes for short CMDUs, since CMDUs are ethernet frames
and must conform to the minimum size requirement of 64 bytes.


arnout (approved):
Are you sure this is needed? AFAIU, the padding should be added automatically by the NIC. Have you done wireshark on the copy port to verify that there are indeed short packets on the wire?

It's true that the padding will not be added automatically when you're not traveling over Ethernet. For example, if it goes over wireless, it could go into an AMPDU and end up without padding. However, AFAICS, there is no minimal length specified at all in 1905.1. It is only in 802.1 that there's a minimal length.

arnout (commented):
Also, are you sure about the 64 bytes? AFAIK, the minimal length on the wire is 64 octets, including all overhead. You do subtract the Ethernet header length (though in reality, the Ethernet header may be larger than that due to VLAN tag, but OK), but you don't subtract the 4 bytes CRC.

tomereli:
do not merge yet, will test again without padding just to be sure this indeed solves the issue.

tomereli:
Indeed not needed, just retested and it is working fine without the padding.

PR #228, opened 2019-08-22
framework: tlvf: fix association state length

tomereli:
Association state length should be 2 bytes, and was defined as 1 byte in
the yaml, which caused M1 received from a reference (none Intel) agent
to be parsed wrong, and the next TLV to fail on paring since the type
was wrong, so fix that.


arnout (approved):
Small grammar mistake in commit message: it is "non-foo", not "none foo" (none is an adverb, not an adjective).

arnout:
> Small grammar mistake in commit message: it is "non-foo", not "none foo" (none is an adverb, not an adjective).

I fixed, rebased and pushed.

PR #231, opened 2019-08-22
controller: autoconfig WSC fix none-intel agent

tomereli:
Fix a big in autoconfig WSC M1 - there was a check that if we are in
certification mode then the agent is assumed to be Intel, which is
obviousely not the case, so remove that.

### DIFF
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -655,8 +655,7 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
         }
     }
 
-    if (database.setting_certification_mode() ||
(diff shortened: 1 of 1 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -655,8 +655,7 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
         }
     }
 
-    if (database.setting_certification_mode() ||
```
arnout, line 658: It should actually have been
```
    if (!database.setting_certification_mode() &&
```

arnout (approved):
I'm OK with just deleting this certification_mode condition. But the original intention was to not handle Intel VS messages in certification mode.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: We have to handle VS messages also in certification mode, for the local agent

PR #232, opened 2019-08-26
Feature/autoconfig any order tlvs

tomereli:
While testing with a reference CTT agent in the real testbed, it has been observed that the agent sends the WSC Autoconfiguration with the M1 TLV before the radio capabilities TLVs, which caused prplMesh controller to faile in parsing.
This is a legal case, TLVs can be ordered in any way in the CMDU, so fix that in the controller by adding support for any order in M1 messages.
While we're at it - do the same in the agent.

A full fix for all kinds of TLVs will be done in a seperate task.

Note - this PR supersedes #184

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4416,27 +4416,62 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
     return true;
 }
 
+/**
+ * @brief Parse AP-Autoconfiguration WSC which should include one AP Radio Identifier
+ *        TLV and one or more WSC TLV containing M2
+ * 
+ * @param sd socket descriptor
+ * @param cmdu_rx received CMDU containing M2
+ * @return true on success
+ * @return false on failure
+ */
 bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
-    // Check if this is a M1 message that we sent to the controller, which was just looped back.
-    // The M1 and M2 messages are both of CMDU type AP_Autoconfiguration_WSC. Thus,
-    // when we send the M2 to the local agent, it will be published back on the local bus because
-    // the destination is our AL-MAC, and the controller does listen to this CMDU.
-    // Ideally, we should use the message type attribute from the WSC payload to distinguish.
-    // Unfortunately, that is a bit complicated with the current tlv parser. However, there is another
-    // way to distinguish them: the M1 message has the AP_Radio_Basic_Capabilities TLV,
-    // while the M2 has the AP_Radio_Identifier TLV.
-    // If this is a looped back M1 CMDU, we can treat is as handled successfully.
-    if (cmdu_rx.getNextTlvType() == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES)) {
-        LOG(DEBUG) << "looped back M1 CMDU";
-        return true;
+    LOG(DEBUG) << "Received AP_AUTOCONFIGURATION_WSC_MESSAGE";
+
+    std::shared_ptr<wfa_map::tlvApRadioIdentifier> ruid = nullptr;
+    std::vector<std::shared_ptr<ieee1905_1::tlvWscM2>> m2_list;
+    bool intel_agent = false;
+    int type         = cmdu_rx.getNextTlvType();
+
+    while ((type = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
+        if (type == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER)) {
+            LOG(DEBUG) << "Found TLV_AP_RADIO_IDENTIFIER TLV";
+            ruid = cmdu_rx.addClass<wfa_map::tlvApRadioIdentifier>();
+        } else if (type == int(ieee1905_1::eTlvType::TLV_WSC)) {
+            // parse all M2 TLVs
+            LOG(DEBUG) << "Found TLV_WSC TLV (assuming M2)";
+            auto tlvWscM2 = cmdu_rx.addClass<ieee1905_1::tlvWscM2>();
+            if (!tlvWscM2) {
+                LOG(ERROR) << "Not an WSC M2 TLV!";
+                return false;
+            }
+            m2_list.push_back(tlvWscM2);
+        } else if (type == int(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
+            // If this is an Intel Agent, it will have VS TLV as the last TLV.
+            // Currently, we don't support skipping TLVs, so if we see a VS TLV, we assume
+            // It is an Intel agent, and will add the class in the Intel agent handling below.
+            intel_agent = true;
+            break;
+        } else if (type == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES)) {
+            // Check if this is a M1 message that we sent to the controller, which was just looped back.
+            // The M1 and M2 messages are both of CMDU type AP_Autoconfiguration_WSC. Thus,
+            // when we send the M2 to the local agent, it will be published back on the local bus because
+            // the destination is our AL-MAC, and the controller does listen to this CMDU.
+            // Ideally, we should use the message type attribute from the WSC payload to distinguish.
+            // Unfortunately, that is a bit complicated with the current tlv parser. However, there is another
+            // way to distinguish them: the M1 message has the AP_Radio_Basic_Capabilities TLV,
+            // while the M2 has the AP_Radio_Identifier TLV.
+            // If this is a looped back M1 CMDU, we can treat is as handled successfully.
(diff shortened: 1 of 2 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -589,36 +590,64 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
...
-    /**
-    * @brief Parse AP-Autoconfiguration WSC which should include one AP Radio Basic Capabilities
-    * TLV and one WSC TLV containing M1
-    */
-    auto radio_basic_caps = cmdu_rx.addClass<wfa_map::tlvApRadioBasicCapabilities>();
+
+    std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_basic_caps = nullptr;
+    std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1                         = nullptr;
+    bool intel_agent                                                       = false;
+    while (1) {
```
morantr: while(1) is bad. Static analysis tools like klocwork will complain about this because they think it is endless loop.
Stop condition must be in while body.
Do it how I did it in channel selection:
```
int tlvType;
while ((tlvType = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
    if (tlvType < 0) {
        LOG(ERROR) << "getNextTlvType has failed";
        return false;
    }

    if (tlvType == int(wfa_map::eTlvTypeMap::SOME_TLV)) {
        //code
    }
}
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -589,36 +590,64 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
...
+        } else if (type == int(ieee1905_1::eTlvType::TLV_WSC)) {
+            LOG(DEBUG) << "Found TLV_WSC TLV (assuming M1)";
+            tlvwscM1 = cmdu_rx.addClass<ieee1905_1::tlvWscM1>();
+        } else if (type == int(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
+            // If this is an Intel Agent, it will have VS TLV as the last TLV.
+            // Currently, we don't support skipping TLVs, so if we see a VS TLV, we assume
+            // It is an Intel agent, and will add the class in the Intel agent handling below.
+            intel_agent = true;
+            break;
+        } else if (type == int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
```
morantr: if you will do the while loop as I said before, you could remove this 'else if' block.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4413,27 +4413,61 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
...
-    // If this is a looped back M1 CMDU, we can treat is as handled successfully.
-    if (cmdu_rx.getNextTlvType() == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES)) {
-        LOG(DEBUG) << "looped back M1 CMDU";
-        return true;
+    LOG(DEBUG) << "Received AP_AUTOCONFIGURATION_WSC_MESSAGE";
+
+    std::shared_ptr<wfa_map::tlvApRadioIdentifier> ruid = nullptr;
+    std::vector<std::shared_ptr<ieee1905_1::tlvWscM2>> m2_list;
+    bool intel_agent = false;
+    while (1) {
```
morantr: Here you could also replace the while(1)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4413,27 +4413,61 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
...
+                return false;
+            }
+            m2_list.push_back(tlvWscM2);
+        } else if (type == int(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC)) {
+            // If this is an Intel Agent, it will have VS TLV as the last TLV.
+            // Currently, we don't support skipping TLVs, so if we see a VS TLV, we assume
+            // It is an Intel agent, and will add the class in the Intel agent handling below.
+            intel_agent = true;
+            break;
+        } else if (type == int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
```
morantr: and remove this 'else if' block as well

morantr (changes requested):

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -607,8 +607,8 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
     std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_basic_caps = nullptr;
     std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1                         = nullptr;
     bool intel_agent                                                       = false;
-    while (1) {
-        int type = cmdu_rx.getNextTlvType();
+    int type = cmdu_rx.getNextTlvType();
+    do {
```
arnout: Do we want to avoid this pattern?
```
int type;
while ((type = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE))
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4416,27 +4416,60 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
     return true;
 }
 
+/**
+ * @brief Parse AP-Autoconfiguration WSC which should include one AP Radio Identifier
+ *        TLV and one or more WSC TLV containing M2
+ * @param sd socket descriptor
```
arnout, line 4423: In general, there should be an empty line after the brief to signal the end of the paragraph. In this case, the param handles that, but it's cleaner to always have an empty line.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4416,27 +4416,60 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
...
-        return true;
-    }
+    LOG(DEBUG) << "Received AP_AUTOCONFIGURATION_WSC_MESSAGE";
+
+    std::shared_ptr<wfa_map::tlvApRadioIdentifier> ruid = nullptr;
+    std::vector<std::shared_ptr<ieee1905_1::tlvWscM2>> m2_list;
+    bool intel_agent = false;
+    int type         = cmdu_rx.getNextTlvType();
+    do {
+        if (type == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER)) {
```
arnout, line 4438: I would prefer a switch - which is difficult here because of the (IMO unnecessary) distinction between ieee1905_1::eTlvType and wfa_map::eTlvTypeMap.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -589,36 +590,63 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
...
+            // the destination is our AL-MAC, and the controller does listen to this CMDU.
+            // Ideally, we should use the message type attribute from the WSC payload to distinguish.
+            // Unfortunately, that is a bit complicated with the current tlv parser. However, there is another
+            // way to distinguish them: the M1 message has the AP_Radio_Basic_Capabilities TLV,
+            // while the M2 has the AP_Radio_Identifier TLV.
+            // If this is a looped back M2 CMDU, we can treat is as handled successfully.
+            LOG(DEBUG) << "Loopbed back M2 CMDU";
+            return true;
+        } else {
+            LOG(ERROR) << "Uknown TLV, type " << std::hex << type;
```
arnout:
```suggestion
            LOG(ERROR) << "Unknown TLV, type " << std::hex << type;
```

arnout (approved):
Nitpicks only.

controller/src/beerocks/master/son_master_thread.cpp
morantr: Why did you choose to use do-while instead of regular while?
In do-while, you assume that the CMDU must have TLVs. But hypothetically, what will happen if the first TLV is EndOfMessage (CMDU has no TLVs) ?

morantr (approved):

controller/src/beerocks/master/son_master_thread.cpp
tomereli: we don't, personal taste. But since both you and Moran like this one better, I'll align :)

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: I also prefer switch - but indeed it is more difficult, so lets stick with if / else.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Can't happen - Autoconfig CMDU contains at least 2 TLVs. But since both you and Arnout prefer the `while` pattern, I changed it to your liking.

Issue #229, opened 2019-08-22
[Feature] Handling WFA-CA command "DEV_SET_CONFIG"


morantr:
### **Introduction**
To pass EasyMesh certification, the system is required to be able to receive various commands from TestSuit(UCC) application.
Current design focus on the command "DEV_SET_CONFIG" which is used to configure a device (Agent) with its configuration parameters.

### **Sequence Diagram**

![dev_set_config](https://www.plantuml.com/plantuml/png/0/dLF9Ri8m4Br7oZzONDfAB89Z54M442fKbef0UqGc66AHsf7j2FZxZhD025tKvZHvC_FytepTbmeBbNx4fc4Q2OGqf0bc2gqyxnRmE5E2np4Ht-73mMt6tHEcDAzf01jgEPLQ2mtzj_N23_pbUkYFtDLaQHhNj2w67Wk1F9NAfZUVlaxyfVy3bsRovhFH-BcvaN9DsOwlkvUmrsots_13yyJ9D1XN1CLa1meHV8rBqwc1Gk36b4cqEICkjd14NcHdbjoORfq2c2T4O4NPteknb2KQ1EDX6RkfYg29Xb36sVASa3BZir2Sapew8-h8nU65d4jaWCD3bE9uS3rW941XXDC28soBj1yzdjiXLhTZeTBZWzGMTtHVcMTcxSkrt1TKPwt5WP1cBXEYmenfP0trLTcwmJf6L177VMt1T7Wk9HNPW510i0WZD2LIuZsnExhUgTigjFnAjY0ouKoIExh_gdiFl98BJRiLkkgWwwpvKNPsRx8TQCwwfT9JusyQUiWEUSo59SZMUIp9d-nEXZeLXNsu9V3GFm40 "dev_set_config")

### **Command Syntax**

Name | Description | Value
-- | -- | --
Backhaul | Set a single backhual interface to Ethernet or Wi-Fi. For Wi-Fi, a RUID associated with its frequency band (2.4GHz, 5GHz Low, 5GHz High) is provided. For Ethernet, the device should return status,COMPLETE until it completes Ethernet onboarding.</br><span style="color:red">NOTE: This argument is only relevant, when UCC send the command to the agent.</span>| String: EthRUID
[...]

Issue #62, opened 2019-06-26
Channel selection full flow (agent + controller)


arnout:
## Description (Full channel selection flow)
Channel selection is a Multiap core module responsible for selecting AP channels dynamically, using Beerocks and Wi-Fi (WAVE) algorithms, controlling the DFS protocol flows and updating the Beerocks database with relevant channel selection data.

This task will focus on the initial implementation of channel selection task on the Controller side, with no DFS channel support, as described on Multiap standard R1, and include the following sub-tasks:

- [x] 1. Add the operating class table and channel/operating-class converting functions.
- [ ] 2. Disable the old channel selection task from working.
- [ ] 3. Creating a channel selection task skeleton which will include:
    - Create task event types.
    - Create a stub state machine.
    - Create a stub event/response handler (will also include handlers on son_master_thread).
    - Create Task setters/getter on Controller DB.
- [ ] 4. Add an Initiate of a new channel selection task on son_master_thread.
- [x] 5. Add a triggering of Agent channel selection on channel selection task, via BML.
- [x] 6. Add a triggering of Agent channel selection on channel selection task, upon AutoConfig M2 message.

Channel Selection Task Core can be split into the following sub-tasks:
- [ ] 1. Send preference query CMDU.
- [ ] 2. Parse preference query report CMDU.
- [ ] 3. Send channel selection request CMDU.
- [ ] 4. Handle channel selection response CMDU and wait for operating channel report.
- [ ] 5. Parse operation channel report CMDU.
- [ ] 6. Activate Agent on Controller DB.

Channel Selection Task implementation will also include:
- [ ] 1. Increment task state machine.
- [ ] 2. Implement task logic.
- [ ] 3. May require to add setters/getters to Controller DB.

### Exit criteria
To be defined for each sub-task with SPs

### Detailed description
[MultiAP_Channel_Selection_v17.docx](https://github.com/prplfoundation/prplMesh/files/3342757/MultiAP_Channel_Selection_v17.docx) - internal Intel design document:
- [ ] upload the design document to prplMesh github under documentation

- **Problem statement** - Passing easymesh controller certification.
- **Possible solutions** - In order to verify that Channel Selection on Controller works properly, it is needed to test it with a certified Agent. If a certified Agent is not available, implementation on the Agent side may be required. An alternative solution can simply be implementing only the controller side and testing with the testbed setup - this is the preferred approach.
- **Components affected** - controller/common/agent(optional)
- **Build considerations** - None.
[...]

PR #236, opened 2019-08-27
Feature/wfa-ca DEV_SET_CONFIG implementation

morantr:
Until now, "DEV_SET_CONFIG" had a dummy handler.
Add correct handling and parsing of CAPI command "DEV_SET_CONFIG". 
After the bss info string is being parsed, it is saved on the controller
the database.
In addition, clear bss info configuration when receiving CAPI command "DEV_RESET_DEFAULT".
DR #229

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4517,8 +4517,8 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
                    << "     Manufacturer: " << manufacturer << std::endl
                    << "     ssid: " << ssid << std::endl
                    << "     bssid: " << network_utils::mac_to_string(bssid) << std::endl
-                   << "     authentication_type: " << authtype << std::endl
-                   << "     encryption_type: " << enctype << std::endl;
+                   << "     authentication_type: " << std::hex << int(authtype) << std::endl
+                   << "     encryption_type: " << int(enctype) << std::dec << std::endl;
     }
 
     if (slave_state != STATE_WAIT_FOR_JOINED_RESPONSE) {
@@ -5012,8 +5012,9 @@ bool slave_thread::autoconfig_wsc_add_m1()
     dh = std::make_unique<mapf::encryption::diffie_hellman>();
     std::copy(dh->pubkey(), dh->pubkey() + dh->pubkey_length(), tlvWscM1->public_key_attr().data);
     std::copy(dh->nonce(), dh->nonce() + dh->nonce_length(), tlvWscM1->enrolee_nonce_attr().data);
-    tlvWscM1->authentication_type_flags_attr().data = WSC::WSC_AUTH_OPEN | WSC::WSC_AUTH_WPA2PSK;
-    tlvWscM1->encryption_type_flags_attr().data     = WSC::WSC_ENCR_AES;
+    tlvWscM1->authentication_type_flags_attr().data =
+        uint16_t(WSC::eWscAuth::WSC_AUTH_OPEN) | uint16_t(WSC::eWscAuth::WSC_AUTH_WPA2PSK);
+    tlvWscM1->encryption_type_flags_attr().data = uint16_t(WSC::eWscEncr::WSC_ENCR_AES);
 
     // Authentication support - store swapped M1 for later M1 || M2* authentication
     // This is the content of M1, without the type and length.
--- a/common/beerocks/bcl/source/beerocks_string_utils.cpp
+++ b/common/beerocks/bcl/source/beerocks_string_utils.cpp
@@ -93,8 +93,16 @@ std::vector<std::string> string_utils::str_split(const std::string &s, char deli
     return elems;
 }
 
-int64_t string_utils::stoi(std::string str, const char *calling_file, int calling_line)
+int64_t string_utils::stoi(const std::string &str, const char *calling_file, int calling_line)
 {
+    if (str.find_first_not_of("0123456789") != std::string::npos) {
+        auto calling_file_str       = std::string(calling_file);
+        const auto caller_file_name = calling_file_str.substr(calling_file_str.rfind('/') + 1);
+        LOG(WARNING) << "string_utils::stoi(), string \"" << str
+                     << "\" has illegal characters, caller: " << caller_file_name << "["
+                     << calling_line << "]";
+        return 0;
+    }
     std::stringstream val_s(str);
     int64_t val;
     val_s >> val;
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -3082,6 +3082,19 @@ void db::setting_certification_mode(bool en)
     config.certification_mode = true;
 }
 
+void db::add_bss_info_configuration(const std::string &al_mac, const bss_info_conf_t &bss_info)
+{
+    bss_infos[al_mac].push_back(bss_info);
+}
+
+std::list<db::bss_info_conf_t> &db::get_bss_info_configuration(const std::string &al_mac)
+{
+    // If al_mac not exist, it will be added, and return empty list
+    return bss_infos[al_mac];
+}
+
+void db::clear_bss_info_configuration() { bss_infos.clear(); }
+
 //
 // PRIVATE FUNCTIONS
 //   must be used from a thread safe context
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -78,10 +78,110 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
 }
 
 /**
+ * @brief Parse bss_info string into bss_info_conf_struct.
+ * 
+ * @param[in] bss_info_str String containing bss info configuration.
+ * @param[out] bss_info_conf Controller database struct filled with configuration.
+ * @param[out] err_string Contains an error description if the function fails.
+ * @return al_mac on success, empty string if not.
+ */
+static std::string parse_bss_info(const std::string &bss_info_str,
+                                  db::bss_info_conf_t &bss_info_conf, std::string &err_string)
+{
+    auto confs = string_utils::str_split(bss_info_str, ' ');
+    if (confs.size() != 8) {
+        err_string = "missing configuration";
+        return std::string();
+    }
+
+    // Alid
+    std::string al_mac(confs[0]);
+    if (!net::network_utils::is_valid_mac(al_mac)) {
+        err_string = "invalid al_mac";
+        return std::string();
+    }
+
+    bss_info_conf = {};
+
+    // Operating class
+    uint operating_class = string_utils::stoi(confs[1]);
+    if (operating_class > UINT8_MAX || operating_class == 0) {
+        err_string = "invalid operating class";
+        return std::string();
+    }
+    bss_info_conf.operating_class = operating_class;
+
+    // SSID
+    bss_info_conf.ssid = confs[2];
+    if (bss_info_conf.ssid.size() > WSC::eWscLengths::WSC_MAX_SSID_LENGTH) {
+        err_string = "ssid is too long";
+        return std::string();
+    }
+
+    // Authentication type
+    auto &authentication_type_str = confs[3];
+    if (!validate_hex_notation(authentication_type_str, 2)) {
+        err_string = "invalid authentication type format";
+        return std::string();
+    }
+
+    uint16_t authentication_type = std::stoi(authentication_type_str, nullptr, 16);
+
+    if (!WSC::eWscAuthValidate::check(authentication_type)) {
+        err_string = "invalid authentication type value";
+        return std::string();
+    }
+    bss_info_conf.authentication_type = static_cast<WSC::eWscAuth>(authentication_type);
+
+    // Encryption type
+    auto &encryption_type_str = confs[4];
+    if (!validate_hex_notation(encryption_type_str, 2)) {
+        err_string = "invalid encryption type format";
+        return std::string();
+    }
+
+    uint16_t encryption_type = std::stoi(encryption_type_str, nullptr, 16);
+
+    if (!WSC::eWscEncrValidate::check(encryption_type)) {
+        err_string = "invalid encryption type value";
+        return std::string();
+    }
+    bss_info_conf.encryption_type = static_cast<WSC::eWscEncr>(encryption_type);
+
+    // Network key
+    bss_info_conf.network_key = confs[5];
+    if (bss_info_conf.network_key.size() > WSC::eWscLengths::WSC_MAX_NETWORK_KEY_LENGTH) {
+        err_string = "network key is too long";
+        return std::string();
+    }
+
+    // Bit 6 of Multi-AP IE's extention attribute, aka "Backhaul BSS"
+    auto &bit_6_str = confs[6];
+    if (bit_6_str != "0" && bit_6_str != "1") {
+        err_string = "invalid bit 6 of Multi-AP IE's extention attribute";
+        return std::string();
+    }
+    uint8_t bss_type = (bit_6_str == "1" ? WSC::eWscVendorExtSubelementBssType::BACKHAUL_BSS : 0);
+
+    // Bit 5 of Multi-AP IE's extention attribute, aka "Fronthaul BSS"
+    auto &bit_5_str = confs[7];
+    if (bit_5_str != "0" && bit_5_str != "1") {
+        err_string = "invalid bit 5 of Multi-AP IE's extention attribute";
+        return std::string();
+    }
+    bss_type |= (bit_5_str == "1" ? WSC::eWscVendorExtSubelementBssType::FRONTHAUL_BSS : 0);
+
+    // Update Bits 6 and 5
+    bss_info_conf.bss_type = static_cast<WSC::eWscVendorExtSubelementBssType>(bss_type);
+
+    return al_mac;
+}
+
+/**
  * @brief Convert enum of WFA-CA status to string.
  * 
  * @param[in] status Enum of WFA-CA status. 
- * @return Status as string. 
+ * @return Status as string.
  */
 const std::string wfa_ca::wfa_ca_status_to_string(eWfaCaStatus status)
 {
@@ -373,9 +473,7 @@ void wfa_ca::handle_wfa_ca_message(
             break;
         }
 
-        // TODO: Perfrom reset to "name". Currently not needed.
-        // NOTE: This action is an asynchronous procedure and will be required to be executed
-        //       from task infrastructure.
+        database.clear_bss_info_configuration();
 
         // Send back second reply
         reply(sd, cmdu_tx, eWfaCaStatus::COMPLETE);
@@ -503,8 +601,6 @@ void wfa_ca::handle_wfa_ca_message(
             // {"program", std::string()},
         };
 
-        // NOTE: Uncertainty if param names: "backhaul", "bss_info{1..N}" are mandatory
-
         if (!parse_params(cmd_tokens_vec, params, err_string)) {
             LOG(ERROR) << err_string;
             reply(sd, cmdu_tx, eWfaCaStatus::INVALID, err_string);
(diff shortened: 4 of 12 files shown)


common/beerocks/bcl/source/beerocks_string_utils.cpp
```diff
@@ -93,8 +93,16 @@ std::vector<std::string> string_utils::str_split(const std::string &s, char deli
     return elems;
 }
 
-int64_t string_utils::stoi(std::string str, const char *calling_file, int calling_line)
+int64_t string_utils::stoi(const std::string &str, const char *calling_file, int calling_line)
 {
+    if (str.find_first_not_of("0123456789") != std::string::npos) {
```
arnout, line 98: This will break the (currently accepted, but I don't know if it's actually used) case of `stoi("0xff")`.

I think a better solution would be to check that `val_s` reached EOF, i.e. change the condition below to `if (val_s.fail() || !val_s.eof())`.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -676,24 +688,18 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
                 return false;
             }
 
-            // Validate hex notation on value or list of values separated by space
+                // Validate hex notation on list of values separated by space
             auto values = string_utils::str_split(it->second, ' ');
-            for (auto value : values) {
+                for (const auto &value : values) {
```
arnout, line 823: What happened to indentation here?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4494,8 +4494,8 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
                    << "     Manufacturer: " << manufacturer << std::endl
                    << "     ssid: " << ssid << std::endl
                    << "     bssid: " << network_utils::mac_to_string(bssid) << std::endl
-                   << "     authentication_type: " << authtype << std::endl
-                   << "     encryption_type: " << enctype << std::endl;
+                   << "     authentication_type: " << std::hex << int(authtype) << std::endl
```
arnout, line 4520: It would be nice if tlvf would generate a pretty-printer function `operator<<(el::base::Writer&, foo)` for every class and struct `foo` it generates. Then this would print a string instead of a hex value. Let me create a task for that: #237

controller/src/beerocks/master/db/db.cpp
```diff
@@ -3082,6 +3082,19 @@ void db::setting_certification_mode(bool en)
     config.certification_mode = true;
 }
 
+void db::set_bss_info_configuration(const std::string &al_mac, const bss_info_conf_t &bss_info)
```
arnout: There can be multiple bss_info per al_mac, and you're doing push_back, so this should be called `add_bss_info_configuration`.

Also, to be consistent with `bss_info_conf_t`, I would call it `add_bss_info_conf` (or vice versa, of course).

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -518,17 +618,45 @@ void wfa_ca::handle_wfa_ca_message(
...
-        // TODO: Find out what to do with value of params: "backhaul", bss_info{1..N}, "name".
+        if (params.find("backhaul") != params.end()) {
+            err_string = "parameter 'backhaul' is relevant only on an agent";
+            LOG(ERROR) << err_string;
+            reply(sd, cmdu_tx, eWfaCaStatus::INVALID, err_string);
+            break;
+        }
+
+        db::bss_info_conf_t bss_info_conf;
+        std::string al_mac;
```
arnout: These should move inside of the loop, below. Which will make it clear that in the current implementation, only one bss_info_conf gets saved in the database instead of all of them... So saving it in the database should also move inside the loop, or maybe even into the parse_bss_info function.

That does mean that if it turns out one of the parameters is invalid, we get a half-configured system. But I don't think it's important to handle that case. In practice, UCC would stop the test and eventually do a device reset, so it's not important what is in the database. And the CAPI specification doesn't say anything about what should happen in case of INVALID or ERROR.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -73,6 +73,106 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
...
+/**
+ * @brief Parse bss_info string into bss_info_conf_struct.
+ * 
+ * @param[in] bss_info_str String conataining bss info configuration
+ * @param[out] bss_info_conf Controller database struct filled with configuration.
+ * @param[out] err_string Contains an error description if the function fails.
+ * @return true if successful, false if not.
+ */
+static std::string parse_bss_info(const std::string &bss_info_str,
+                                  db::bss_info_conf_t &bss_info_conf, std::string &err_string)
```
arnout, line 89: Maybe it's better to pass al_mac as an `[out]` parameter and return bool. The doxygen claims that it return bool, BTW :-).

It may be a good idea if we would converge on the pattern to use for any function. Now, we have some functions that return a result which may have a sentinel value to indicate error (like in this case, the empty string), and other functions that return bool and take arguments for their outputs.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -73,6 +73,106 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
...
+{
+    auto confs = string_utils::str_split(bss_info_str, ' ');
+    if (confs.size() != 8) {
+        err_string = "missing configuration";
+        return std::string();
+    }
+
+    // Alid
+    std::string al_mac(confs[0]);
+    if (!net::network_utils::is_valid_mac(al_mac)) {
```
arnout, line 99: Actually, IMO it would be better to use sMacAddr in the database (*), and therefore also use mac_from_string here.

But now I notice that that function doesn't even have proper error handling :-(

(*) Yes, I realize that the database currently uses strings all over the place to store MAC addresses, but I think that that is wrong and that we should (gradually) change it.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -73,6 +73,106 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
...
+ * @param[in] bss_info_str String conataining bss info configuration
+ * @param[out] bss_info_conf Controller database struct filled with configuration.
+ * @param[out] err_string Contains an error description if the function fails.
+ * @return true if successful, false if not.
+ */
+static std::string parse_bss_info(const std::string &bss_info_str,
+                                  db::bss_info_conf_t &bss_info_conf, std::string &err_string)
+{
+    auto confs = string_utils::str_split(bss_info_str, ' ');
+    if (confs.size() != 8) {
```
arnout, line 92: It is also valid to have just two parameters: ALID and operating class. In that case, the entries for that ALID/operatingclass combination should be removed. Cfr. the second example for BSS_Info1.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -73,6 +73,106 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
...
+    std::string al_mac(confs[0]);
+    if (!net::network_utils::is_valid_mac(al_mac)) {
+        err_string = "invalid al_mac";
+        return std::string();
+    }
+
+    bss_info_conf = {};
+
+    // Operating class
+    uint operating_class = string_utils::stoi(confs[1]);
```
arnout, line 107: If you look at Table 11, you see that operating class can be "8x", "11x" or "12x". At the end of section 2.6.1, these strings are explained: 8x stands for 81 or 83 or 84, 11x stands for 115 or 116, and 12x stands for 124, 125 or 126. I haven't checked what these map on exactly, but I suppose 8x == 2.4GHz, 11x == 5GHz low, and 12x == 5GHz high.

To be fully flexible, the database should keep a list of operating classes in bss_info_conf. However, I think we can probably suffice with just putting 2.4 or 5GHz. Do we already have an enum for this?

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -73,6 +73,106 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
...
+    bss_info_conf.ssid = confs[2];
+    if (bss_info_conf.ssid.size() > WSC::eWscLengths::WSC_MAX_SSID_LENGTH) {
+        err_string = "ssid is too long";
+        return std::string();
+    }
+
+    // Authentication type
+    auto &authentication_type_str = confs[3];
+    if (!validate_hex_notation(authentication_type_str, 2)) {
+        err_string = "invalid authentication type format";
```
arnout, line 124: Wouldn't it be useful to also include the value in the error string? I.e. `err_string = "..." + authentication_type_str`

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -73,6 +73,106 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
...
+
+    // SSID
+    bss_info_conf.ssid = confs[2];
+    if (bss_info_conf.ssid.size() > WSC::eWscLengths::WSC_MAX_SSID_LENGTH) {
+        err_string = "ssid is too long";
+        return std::string();
+    }
+
+    // Authentication type
+    auto &authentication_type_str = confs[3];
```
arnout, line 122: Should be const.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -73,6 +73,106 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
...
+        return std::string();
+    }
+
+    // Bit 6 of Multi-AP IE's extention attribute, aka "Backhaul BSS"
+    auto &bit_6_str = confs[6];
+    if (bit_6_str != "0" && bit_6_str != "1") {
+        err_string = "invalid bit 6 of Multi-AP IE's extention attribute";
+        return std::string();
+    }
+    uint8_t bss_type = (bit_6_str == "1" ? WSC::eWscVendorExtSubelementBssType::BACKHAUL_BSS : 0);
```
arnout, line 164: For symmetry, I think it's nicer to set `bss_info_conf.bss_type = 0;` before, and then use `bss_info_conf.bss_type |= ...` both here and for the fronthaul.

Argh, that's not possible because eWscVendorExtSubelementBssType doesn't have an `operator|` :-)

arnout (changes requested):

common/beerocks/bcl/source/beerocks_string_utils.cpp
morantr: when we expect to convert hex value, we use std::stoi() instead of string_utils::stoi(), and explicit mention the base 16.
But I agree that maybe it will be best to refactor the use of stoi. Not in this PR of course.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I have no idea, but it was fixed in follow-up commit because I don't see it in my code.
This fix requires to rebase with edit, so I will fix it when I'll rebase.
Keeping it unresolved to remember...

controller/src/beerocks/master/db/db.cpp
morantr: will fix

controller/src/beerocks/master/wfa_ca.cpp
morantr: Yes, you are right, now that you are saying that I see that I save only the last bss_info instead of all of them. I tried to separate the input validation from the saving.
I think to change the database setter to get a list instead of one config.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I would prefer to avoid as much as I can from getting the result from a function argument.
I hate when I have to declare a return variable before calling to a function.
But I'll change the doxygen return.

controller/src/beerocks/master/wfa_ca.cpp
morantr: I guess it is easier to work with string...
But maybe you are right, anyway, maybe we will do it as part of database refactoring.
(There a lot of things to refactor there...)

controller/src/beerocks/master/wfa_ca.cpp
morantr: I believe they had a copy/paste error there. The description says otherwise.
If It appears to be that I wrong, I'll fix it.

controller/src/beerocks/master/wfa_ca.cpp
morantr: Adding an operating class table will take place as part of channel selection full flow #62.
It is not true how you have divided the operating classes. 
For example OpClass 128 has channels: {42,58,106,122,138,155}.
Since I'm not sure what exact bss_info will send us, I prefer to keep it that way to have flexibility. What you proposed will reduce flexibility.

controller/src/beerocks/master/wfa_ca.cpp
morantr: Why? I don't care if it will change.

controller/src/beerocks/master/wfa_ca.cpp
morantr: Maybe, but from the outside, I add to which bss_info this failed, so the user can easily locate its problem.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -61,6 +61,10 @@ static bool validate_hex_notation(const std::string &str)
         return false;
     }
 
+    if (expected_octets != 0 && (str.size() - 2) / 2 != expected_octets) {
```
tomereli, line 68: This deserves a comment explaining how this is calculated ( `str.size() -2 / 2`)

controller/src/beerocks/master/wfa_ca.cpp
morantr: been there 😆

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -49,7 +49,7 @@ static std::string check_forbidden_chars(const std::string &str)
  * @param[in] str String to check hex notation.
  * @return true if string is valid hex notation, else false.
  */
-static bool validate_hex_notation(const std::string &str)
+static bool validate_hex_notation(const std::string &str, uint8_t expected_octets = 0)
```
tomereli, line 53: Need to update this function's params.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -666,10 +670,18 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
             // Type
             if (tlv_member_idx == 0) {
                 tlv_hex.type = &it->second;
+                if (!validate_hex_notation(*tlv_hex.type, 1)) {
```
tomereli, line 799: What will happen if the testbed will send `tlv_type,0x02` or `0x0002`? Is it invalid? Why do we want to make things harder for the end user?

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -685,24 +697,18 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
             }
 
-            // Validate hex notation on value or list of values separated by space
+                // Validate hex notation on list of values separated by space
             auto values = string_utils::str_split(it->second, ' ');
-            for (auto value : values) {
+                for (const auto &value : values) {
                 if (!validate_hex_notation(value)) {
-                    err_string = "param name '" + lookup_str + "' has invalid hex notation value";
+                        err_string =
```
tomereli, line 825: wrong indentation as well

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -685,24 +697,18 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
-                    return false;
-
-                    // Length
-                } else if (tlv_member_idx == 1 && value.size() != 6) {
-                    err_string = "length field is too large: " + value;
+                }
+            } else {
+                LOG(ERROR) << "Illegal tlv_member_idx value: " << int(tlv_member_idx);
+                err_string = err_internal;
                     return false;
```
tomereli: also wrong indentation

controller/src/beerocks/master/wfa_ca.cpp
morantr: Argh, didn't find an elegant way to do it as I wanted. I'll do it as you suggested.

controller/src/beerocks/master/wfa_ca.cpp
tomereli: agree with Moran on this one, I also hate it when I have to declare an extra variable so if it can be avoided like here when we return an empty string, then I prefer that.

controller/src/beerocks/master/wfa_ca.cpp
tomereli: So can you open a task for this refactoring?

tomereli (changes requested):

controller/src/beerocks/master/wfa_ca.cpp
morantr: Yes, it is invalid because the UCC should send the hex notation string with the number of octets it wants to be written to the buffer.
for two bytes: 0x0002,
for one byte: 0x02

controller/src/beerocks/master/wfa_ca.cpp
morantr: will add

controller/src/beerocks/master/wfa_ca.cpp
morantr: I have no idea, but it was fixed in follow-up commit because I don't see it in my code.
This fix requires to rebase with edit, so I will fix it when I'll rebase.
Keeping it unresolved to remember...

controller/src/beerocks/master/wfa_ca.cpp
morantr: I have no idea, but it was fixed in follow-up commit because I don't see it in my code.
This fix requires to rebase with edit, so I will fix it when I'll rebase.
Keeping it unresolved to remember...

controller/src/beerocks/master/wfa_ca.cpp
morantr: Well, it will require DR first.

controller/src/beerocks/master/wfa_ca.cpp
tomereli: Now I understand but in that case, why not use `sizeof("0x")`?
Regarding the division by 2 - still not clear to me. One octet == one byte, so why 2?

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -79,7 +83,7 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
  * @param[in] bss_info_str String conataining bss info configuration
  * @param[out] bss_info_conf Controller database struct filled with configuration.
  * @param[out] err_string Contains an error description if the function fails.
- * @return true if successful, false if not.
+ * @return al_mac on success, empty string if not.
```
tomereli, line 86:
```
@return al_mac on success, empty string otherwise
```

tomereli (changes requested):

controller/src/beerocks/master/wfa_ca.cpp
morantr: sizeof("0x") is 3 because it add '\0'.
I can create const string do that:
```
const std::string prefix("0x");
(str.size() - prefix.size()) /2
```

> Regarding the division by 2 - still not clear to me. One octet == one byte, so why 2?

Because one byte is represented by to characters. 
For example, "FF" is stored in 1 byte, but has two chars 'F' and 'F'.

controller/src/beerocks/master/wfa_ca.cpp
arnout: We're deciding on a lot of this kind of coding rules in our reviews, so I've updated the CODINGSTYLE.md with the ones I remember: #242 

It would be nice if others would also update CODINGSTYLE.md like that.

controller/src/beerocks/master/wfa_ca.cpp
arnout: No need for a full DR to open a refactoring task: #243

controller/src/beerocks/master/wfa_ca.cpp
arnout: No, the description says it too (though the wording is a bit confusing): "To clear the BSS info stored for a specific operating class, only the device ALID and operating class are retained."

controller/src/beerocks/master/wfa_ca.cpp
arnout: It is very clearly stated in 3 different places that this can be 8x, 11x or 12x. So string_utils::stoi is certainly wrong.

For reference: the three examples in section 10.10 of the CAPI spec have 8x; section 2.6.1 of the EasyMesh test plan explains the meaning of the three possible values; and section 3.3.1, Table 11 gives all possible bss_info strings, and it's always one of these three.

controller/src/beerocks/master/wfa_ca.cpp
arnout: > Yes, it is invalid because the UCC should send the hex notation string with the number of octets it wants to be written to the buffer.

Neither the CAPI specification nor the EasyMesh test plan really say that for type and value the number of hex characters must exactly correspond to the size of these two. However, all the examples have it with the exact size, and UCC does generate it with the exact size, and it makes our implementation a bit simpler if we can rely on the exact size. So, I agree with Moran that it's good like this. If it turns out that UCC does generate something else, we can adapt.

BTW this check was already there before. It's just a refactoring.

controller/src/beerocks/master/wfa_ca.cpp
arnout: I don't think the std::string("0x").size() makes things much clearer, I prefer the comment. It was anyway already clear to me without comment :-)

arnout (approved):
What is clearly still missing is an update to test_flows.sh that calls DEV_SET_CONFIG with one of the strings from the test plan (Table 11).

Also, the parsing of the operating class is clearly wrong.

However, I need this to be able to proceed with the M2, so I'll merge it as is. So, please add a test and fix the 8x/11x/12x parsing in a follow-up PR.

tomereli (approved):

Issue #227, opened 2019-08-22
[TASK] Autconfig WSC - fill in M2 information from database


arnout:
## Description

The Intel vendor-specific architecture doesn't really do autoconfiguration. Currently, the M2 of the autoconfiguration is just filled in with fixed data. For certification, it needs to be filled in with real data.

## Exit criteria

- [ ] Controller database contains a list of ALIDs, operating classes, SSIDs and credentials to configure.
- [ ] This table is initialised with the static data currently used (for testing).
- [ ] Autoconfig M1 handling matches the band to the table in the database and constructs M2s based on that.
- [ ] Agent reports the details of the received M2 (for testing).
- [ ] DEV_RESET_DEFAULT resets the credentials in the database.
- [ ] DEV_SET_CONFIG adds credentials to the database.
- [ ] test_flows.sh includes a test for both of the above.

PR #239, opened 2019-08-27
Hotfix/fix non intel agent con map

tomereli:
This goes on top of #246 

The current design of the network map in the prplMesh controller is not
aligned to EasyMesh, and therefore when a none-Intel agent onboards, we
don't have its backhaul MAC (which is not really used in case of
ethernet backhaul which is the only type we support in the controller
for now).

Due to that, the agent node is added in the wrong DB hierarchy and not
under the GW ETHERNET node as it should.

The solution is to port the relevant code section from the Intel agent
handling and define the backhaul mac as a transformation of the AL MAC - last octet incremented by 1.

The non-Intel agent is then added correctly and can also be seen with
bml_conn_map under the GW ETHERNET interface.

### DIFF
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -1080,8 +1080,8 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
               << "    low_pass_filter_on = " << int(notification->low_pass_filter_on()) << std::endl
               << "    radio_identifier = " << radio_identifier << std::endl
               << "    radio_mac = " << radio_mac << std::endl
-              << "    acs_enabled = " << int(notification->wlan_settings().acs_enabled)
-              << std::endl;
+              << "    acs_enabled = " << int(notification->wlan_settings().acs_enabled) << std::endl
+              << "    is_gw_slave = " << int(is_gw_slave) << std::endl;
 
     if (!is_gw_slave) {
 
@@ -1532,9 +1532,47 @@ bool master_thread::handle_non_intel_slave_join(Socket *sd,
                                                 std::string bridge_mac, std::string radio_mac,
                                                 ieee1905_1::CmduMessageTx &cmdu_tx)
 {
-    // TODO backhaul handling.
-    // Multi-AP Agent doesn't say anything about the backhaul, so we have to rely on Intel Slave Join for that.
-    std::string backhaul_mac = "";
+
+    // Multi-AP Agent doesn't say anything about the backhaul, so simulate ethernet backhaul to satisfy
+    // network map. MAC address is the bridge MAC with the last octet incremented by 1.
+    // The mac address for the backhaul is the same since it is ethernet backhaul.
+    sMacAddr mac = network_utils::mac_from_string(bridge_mac);
+    mac.oct[5]++;
+    std::string backhaul_mac = network_utils::mac_to_string(mac);
+    mac.oct[5]++;
+    std::string eth_switch_mac   = network_utils::mac_to_string(mac);
+    std::string parent_bssid_mac = network_utils::ZERO_MAC_STRING;
+    std::string manufacturer(tlvwscM1->manufacturer(), tlvwscM1->manufacturer_length());
+
+    LOG(INFO) << "IRE generic Slave joined, sd=" << intptr_t(sd) << std::endl
+              << "    manufacturer=" << manufacturer << std::endl
+              << "    parent_bssid_mac=" << parent_bssid_mac << std::endl
+              << "    al_mac=" << bridge_mac << std::endl
+              << "    eth_switch_mac=" << eth_switch_mac << std::endl
+              << "    backhaul_mac=" << backhaul_mac << std::endl
+              << "    radio_identifier = " << radio_mac << std::endl;
+
+    LOG(DEBUG) << "simulate backhaul connected to the GW's LAN switch ";
+    auto gw_container = database.get_nodes_from_hierarchy(0, beerocks::TYPE_GW);
+    if (gw_container.empty()) {
+        LOG(ERROR) << "can't get GW node!";
+        return false;
+    }
+
+    auto gw_mac          = *gw_container.begin();
+    auto gw_lan_switches = database.get_node_children(gw_mac, beerocks::TYPE_ETH_SWITCH);
+
+    if (gw_lan_switches.empty()) {
+        LOG(ERROR) << "GW has no LAN SWITCH node!";
+        return false;
+    }
+
+    auto gw_lan_switch = *gw_lan_switches.begin();
+
+    LOG(DEBUG) << "add a placeholder backhaul_mac = " << backhaul_mac
+               << " gw_lan_switch = " << gw_lan_switch << " TYPE_IRE_BACKHAUL , STATE_CONNECTED";
+    database.add_node(backhaul_mac, gw_lan_switch, beerocks::TYPE_IRE_BACKHAUL);
+    database.set_node_state(backhaul_mac, beerocks::STATE_CONNECTED);
 
     // TODO bridge handling.
     // Assume repeater
@@ -1550,9 +1588,20 @@ bool master_thread::handle_non_intel_slave_join(Socket *sd,
                << ire_type;
     database.add_node(bridge_mac, backhaul_mac, ire_type);
     database.set_node_state(bridge_mac, beerocks::STATE_CONNECTED);
-    database.set_node_manufacturer(bridge_mac, tlvwscM1->manufacturer());
-
-    // Controller expects backhaul which contains bridge which contains radio, but we only get radio...
+    database.set_node_socket(bridge_mac, sd);
+    database.set_node_platform(backhaul_mac, beerocks::ePlatform::PLATFORM_LINUX);
+    database.set_node_platform(bridge_mac, beerocks::ePlatform::PLATFORM_LINUX);
+    database.set_node_backhaul_iface_type(backhaul_mac, beerocks::eIfaceType::IFACE_TYPE_ETHERNET);
+    database.set_node_backhaul_iface_type(bridge_mac, beerocks::IFACE_TYPE_BRIDGE);
+    database.set_node_manufacturer(backhaul_mac, manufacturer);
+    database.set_node_manufacturer(bridge_mac, manufacturer);
+    database.set_node_type(backhaul_mac, beerocks::TYPE_IRE_BACKHAUL);
+    database.set_node_name(backhaul_mac, manufacturer + "_BH");
+    database.set_node_name(bridge_mac, manufacturer);
+    database.add_node(eth_switch_mac, bridge_mac, beerocks::TYPE_ETH_SWITCH);
+    database.set_node_state(eth_switch_mac, beerocks::STATE_CONNECTED);
+    database.set_node_name(eth_switch_mac, manufacturer + "_ETH");
+    database.set_node_manufacturer(eth_switch_mac, eth_switch_mac);
 
     // Update existing node, or add a new one
     if (database.has_node(radio_mac)) {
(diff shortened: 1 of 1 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1527,7 +1527,46 @@ bool master_thread::handle_non_intel_slave_join(Socket *sd,
...
+    LOG(INFO) << "IRE generic Slave joined, sd=" << intptr_t(sd) << std::endl
+              << "    manufacturer=" << manufacturer << std::endl
+              << "    parent_bssid_mac=" << parent_bssid_mac << std::endl
+              << "    backhaul_mac=" << backhaul_mac << std::endl
+              << "    al_mac=" << bridge_mac << std::endl
+              << "    radio_identifier = " << radio_mac << std::endl;
+
+    //TODO might need to handle bssids of VAP nodes as well in this case
+    if (database.get_node_state(backhaul_mac) != beerocks::STATE_CONNECTED &&
+        database.get_node_state(backhaul_mac) != beerocks::STATE_CONNECTED_IP_UNKNOWN) {
```
morantr: If you just generated the backhaul mac, it will not be in the DB, so why do you write this condition?
`database.get_node_state(backhaul_mac)` will return `"beerocks::STATE_MAX"` so the condition will always be true.

morantr (changes requested):
It seems ok but one thing bothers me. You generate a random mac address for the backhaul agent.
What will happen if the same agent will disconnect and connect multiple times? Every time it will generate another random mac address which will be added to the database as a new node and will end up by a lot of irrelevant nodes in the database.
I think it will be better if you generate the backhaul mac from another mac you have similar to how we do on ethernet node (+1 to the last octet of the bridge).

tomereli:
> It seems ok but one thing bothers me. You generate a random mac address for the backhaul agent.
> What will happen if the same agent will disconnect and connect multiple times? Every time it will generate another random mac address which will be added to the database as a new node and will end up by a lot of irrelevant nodes in the database.
> I think it will be better if you generate the backhaul mac from another mac you have similar to how we do on ethernet node (+1 to the last octet of the bridge).

When an agent disconnects, doesn't it also remove the nodes from the DB?

tomereli:
> > It seems ok but one thing bothers me. You generate a random mac address for the backhaul agent.
> > What will happen if the same agent will disconnect and connect multiple times? Every time it will generate another random mac address which will be added to the database as a new node and will end up by a lot of irrelevant nodes in the database.
> > I think it will be better if you generate the backhaul mac from another mac you have similar to how we do on ethernet node (+1 to the last octet of the bridge).
> 
> When an agent disconnects, doesn't it also remove the nodes from the DB?

OK you are absolutely right, I fixed it to be static MAC with the last octet being +1 of the AL_MAC. I also removed the commit which added get_random_mac() since it is not needed anymore.

morantr (approved):
I see you removed the adding of the ethernet node of the non-intel agent.
It means that if we will want to connect another agent to this agent via ethernet, we will have to add an ethernet node.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -527,7 +527,6 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
         return false;
     }
 
-    int bss_type  = tlvWscM1->vendor_extensions_attr().subelement_value;
```
arnout: This makes me realize that our definition of the vendor extension in M1 is wrong. M1 has a vendor extension element to contain the WSC version (the interpretation of the Version attribute was so badly specified in the first version of WSC that it was not possible to put "version 2" in it, so they had to add a vendor-specific element to carry the "version 2" information). It happens to work correctly because the length of the version2 element is the same as the length of the MultiAP vendor extension in M2, and because the tlvf doesn't check the values. But to be really correct, we need to have separate definitions.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -576,6 +571,7 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
     std::fill(config_data.network_key_attr().data,
               config_data.network_key_attr().data + config_data.network_key_attr().data_length,
               0xaa); //DUMMY
+    config_data.vendor_extensions_attr().subelement_value = WSC::TEARDOWN;
```
arnout: That's already the default value, and it will have to be changed anyway as part of #227, so don't bother setting it here.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -681,11 +677,11 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
...
-    for (int i = 0; i < radio_basic_caps->maximum_number_of_bsss_supported(); i++) {
-        if (!autoconfig_wsc_add_m2(tlvwscM1)) {
-            LOG(ERROR) << "Failed setting M2 attributes";
-            return false;
-        }
+    // TODO add M2 for each VAP to be configured based on DB configuration set by certification
+    // testbed CAPI command dev_set_config
+    if (!autoconfig_wsc_add_m2(tlvwscM1)) {
+        LOG(ERROR) << "Failed setting M2 attributes";
+        return false;
```
arnout: I don't see a reason for this change. It will be updated as part of #227, but the loop will most likely stay. We will basically have to iterate over the number of supported BSSes in parallel with the BSSes configured i the database.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1525,9 +1521,43 @@ bool master_thread::handle_non_intel_slave_join(Socket *sd,
...
+
+    LOG(INFO) << "IRE generic Slave joined, sd=" << intptr_t(sd) << std::endl
+              << "    manufacturer=" << manufacturer << std::endl
+              << "    parent_bssid_mac=" << parent_bssid_mac << std::endl
+              << "    al_mac=" << bridge_mac << std::endl
+              << "    backhaul_mac=" << backhaul_mac << std::endl
+              << "    radio_identifier = " << radio_mac << std::endl;
+
+    LOG(DEBUG) << "simulate backhaul connected to the GW's LAN switch ";
+    auto gw_container = database.get_nodes_from_hierarchy(0, beerocks::TYPE_GW);
```
arnout, line 1556: The first commit does some whitespace mess-up which is fixed by the second commit. Please do it correctly in the first commit while rebasing.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1563,18 +1602,22 @@ bool master_thread::handle_non_intel_slave_join(Socket *sd,
     // TODO Assume repeater mode
     database.set_hostap_repeater_mode_flag(radio_mac, true);
     // TODO Assume no backhaul manager
-    database.set_hostap_backhaul_manager(radio_mac, false);
+    database.set_hostap_backhaul_manager(radio_mac, true);
```
arnout: Why this change? And shouldn't the comment above be adapted as well?

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvWscM2.yaml
```diff
@@ -76,6 +76,5 @@ tlvWscM2:
   configuration_error_attr: WSC::sWscAttrConfigurationError
   device_password_id_attr: WSC::sWscAttrDevicePasswordID
   os_version_attr: WSC::sWscAttrOsVersion
-  vendor_extensions_attr: WSC::sWscAttrVendorExtension
```
arnout: The Version2 attribute must be present here, according to the WSC spec. The BBF code doesn't check for it, but maybe other implementations do.

BTW, in prplMesh-old, I added the WFA extension directly in this vendor extension (so there was one vendor extension 2 IEs: version2 and MultiAP). I had missed the part specifying that it was to be in the encrypted section.

arnout (changes requested):
Version2 attribute really should stay, both in M1 and M2. Even though it works as is, it's not according to spec :-)

Also, the commit message of the first commit still says that it's random, which is no longer true. Also, it should mention the other changes made in that commit (e.g. trigger BML update).

Don't forget to remove the part about teardown from the second commit message if you revert to the old behaviour (which already had the teardown bit set).

controller/src/beerocks/master/son_master_thread.cpp
tomereli: The controller requires at least one slave to be the backhaul manager slave, so it is better to set it by default *I think*.
@morantr can you comment?
I will update the //TODO

arnout:
The second commit is superseded by #246

controller/src/beerocks/master/son_master_thread.cpp
tomereli: When a testbed testcase starts, the agent does onboarding to the controller. At this point `dev_set_config` was not yet requested from the UCC application, so the correct operation the controller needs to take is to tear down all radios, and for that we don't need more than one M2.
Whoever will implement #227 will update this code section to check if there is a configuration in the DB, and if so - use it instead of tear down both radios.
Also - this is how the reference controller behaves during the tests, so it is easier at least for now to stick to the same behavior so we can easily compare sniffer logs of prplMesh controller VS reference controller.

So bottom line - I think we should keep this change.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Fixed

controller/src/beerocks/master/son_master_thread.cpp
tomereli: I assume this is handled in #246 ?

controller/src/beerocks/master/son_master_thread.cpp
morantr: Well setting the son_slave to backhaul manager is only matter on flows which we use beerocks proprietary messages, and as far as I know, Multi-Ap standard does not mention which radio is the backhaul manager, so when we will fully work with the Multi-Ap flows, I think we could remove it in anyway.
For the non-intel agent I think you can remove this line completely.

tomereli: Ok, will do.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: rebased on top of #246 and removed the second obsolete commit

controller/src/beerocks/master/son_master_thread.cpp
tomereli: I will do that in a separate commit or even a separate PR, so marking as resolved for now as I removed this commit from the PR.

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvWscM2.yaml
tomereli: marking as resolved, this commit is removed and the PR is rebased on top of #246

tomereli:
> I see you removed the adding of the ethernet node of the non-intel agent.
> It means that if we will want to connect another agent to this agent via ethernet, we will have to add an ethernet node.

The reason I removed it is because I didn't understand why the MAC address of the backhaul and the MAC address of the ethernet is always the same in BML connection map. I now understand that it is because of the fact we are connecting the backhaul via ethernet, but for some reason if I set both the backhaul map and the eth_switch_mac to the same address, the whole agent is not seen in `bml_conn_map`. Any idea why?

morantr:
> > I see you removed the adding of the ethernet node of the non-intel agent.
> > It means that if we will want to connect another agent to this agent via ethernet, we will have to add an ethernet node.
> 
> The reason I removed it is because I didn't understand why the MAC address of the backhaul and the MAC address of the ethernet is always the same in BML connection map. I now understand that it is because of the fact we are connecting the backhaul via ethernet, so I changed the MAC address of both to be the same and added ethernet node as well.

I don't know why you have seen they have the same mac, this seems to be a bug to me.
The backhaul mac is the mac of the backhaul interface used to connect to the controller which is set on beerocks_agent.conf.
When the son_slave sends to the backhaul manager "BACKHAULE_ENABLE" it gives him the backhaul radios and ethernet interfaces names if it enabled (platform manager is responsible to read ethernet enable flag from platform DB) if it is not enabled, he sends ZERO_MAC.
When the backhaul manager sends "ACTION_BACKHAUL_CONNECTED_NOTIFICATION" he put there the mac address of the interface it connected with on "backhaul_mac", and that is what eventually the controller gets on slave join.

tomereli:
> > > I see you removed the adding of the ethernet node of the non-intel agent.
> > > It means that if we will want to connect another agent to this agent via ethernet, we will have to add an ethernet node.
> > 
> > 
> > The reason I removed it is because I didn't understand why the MAC address of the backhaul and the MAC address of the ethernet is always the same in BML connection map. I now understand that it is because of the fact we are connecting the backhaul via ethernet, so I changed the MAC address of both to be the same and added ethernet node as well.
> 
> I don't know why you have seen they have the same mac, this seems to be a bug to me.
> The backhaul mac is the mac of the backhaul interface used to connect to the controller which is set on beerocks_agent.conf.
> When the son_slave sends to the backhaul manager "BACKHAULE_ENABLE" it gives him the backhaul radios and ethernet interfaces names if it enabled (platform manager is responsible to read ethernet enable flag from platform DB) if it is not enabled, he sends ZERO_MAC.
> When the backhaul manager sends "ACTION_BACKHAUL_CONNECTED_NOTIFICATION" he put there the mac address of the interface it connected with on "backhaul_mac", and that is what eventually the controller gets on slave join.

You are talking about beerocks agent, but this PR is handling non-beerocks agents - so no beerocks_agent.conf.. We generate the MAC for both the ethernet and the backhaul with different MAC addresses, and they are seen with `dump_node_info` correctly, but in bml_conn_map the backhaul and ethernet interface have the same MAC, which makes sense - if the backhaul is ETHERNE, then the ETHERNET backhaul should have the same MAC. Anyway if there is a bug hiding as a result it is probably in the BML conn map so I wouldn't block merging of this PR on that.

arnout (approved):

PR #242, opened 2019-08-28
CODINGSTYLE.md: add coding patterns section

arnout:
Through recent reviews, we converged on a number of C++ coding patterns
that we use in our code. Document these patterns in a new Coding
patterns section in CODINGSTYLE.md.

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -22,6 +22,86 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
 * Getter functions use the name without "m": `data()`.
 * Setter functions use the "set" prefix: `set_data(value)`.
 
+### Coding patterns
+
+This section describes some of the coding patterns we prefer to see in our code.
+Using the same kind of coding patterns consistently makes it easier for others to understand the code at first glance.
+Conversely, code that breaks the usual patterns takes more time to understand and may put the reader on the wrong foot.
+
+Exceptions are not used at all in prplMesh code.
+If a standard library function is called that may throw an exception, make sure to do the necessary checks before, to make sure an exception cannot be thrown.
+
+#### Loops
+
+Avoid endless loops like `while (true)`.
+Some static analysis tools give an error or have a hard time analysing it.
+They are usually able to analyse some forms of endless loops, but it's better to be consistent and never have them.
+Example:
+
+```cpp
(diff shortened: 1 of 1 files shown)


CODINGSTYLE.md
```diff
@@ -22,6 +22,68 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
+#### Loops
+
+Avoid endless loops - some static analysis tools give an error or have a hard time analysing it.
+They are usually able to analyse some forms of endless loops, but it's better to be consistent and never have them.
+Example:
+
+```cpp
+    static const size_t max_tlv = 256;
+    size_t tlv_idx;
+    for (tlv_idx = 0; tlv_idx < max_tlv; tlv_idx++) {
```
morantr, line 44: I would give a bad example like `while (true)`

CODINGSTYLE.md
```diff
@@ -22,6 +22,68 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
+       ...
+    }
+    if (tlv_idx == max_tlv) {
+        LOG(ERROR) << "Reached stop condition";
+        return false;
+    }
+```
+
+Prefer to use C++-style iterators if at all possible.
```
morantr: Actually, if possible it is better to use ranged for loop:
```
for (const auto& element : elements) {
    // code
}
```
If there are some limitations. use iterators. 

Also, I think that @tomereli would have added to avoid `std::for_each()` 😉

tomereli (approved):

CODINGSTYLE.md
arnout: In general I prefer to avoid bad examples unless you can make it visually *very* clear that it's bad.

I will add "like `while (true)`" in the text, that would work.

CODINGSTYLE.md
arnout: > Actually, if possible it is better to use ranged for loop:

That's what I meant actually :-). Reworded, and added your example.

I've also added an example of when to use a traditional loop instead of a ranged loop.

> Also, I think that @tomereli would have added to avoid `std::for_each()` wink

Hey, I love for_each and friends!

morantr (approved):

tomereli:
Can't re-run the failed build. @arnout can you rebase and force push? This will force CI rebuild.

PR #245, opened 2019-08-28
Feature/autoconfig parse radio caps

tomereli:
Add a new API to the database to set the supported operating class as received from the Radio Basic Capabilities TLV as part of the autoconfig WSC message from the agent.
Currently, prplMesh stores channels and not operating classes. Since EasyMesh specification is using the operating classes and not directly channels, it makes sense to change our database
and channel selection code to use operating classes.
However, this will be a big change and should be done as part of the full channel selection flow.
In order to put things in the right direction, this commit adds a new API for storing the supported operating class in the DB, which performs the conversion to sWifiChannel which is used across the code. This will be used in a followup commit to storing the operating classes in
the DB which will be then updating the radio capabilities, which eventually allow bml_conn_map to show the correct band based on the operating class.

### DIFF
--- a/common/beerocks/bcl/source/son/son_wireless_utils.cpp
+++ b/common/beerocks/bcl/source/son/son_wireless_utils.cpp
@@ -455,3 +455,56 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
 
     return channels;
 }
+
+/**
+ * @brief convert operating class to channel set based on Table 4-E in the ieee 802.11 specification
+ *
+ * @param operating_class operating class
+ * @return std::set<uint8_t> set of supported channels by the operating class or empty if failure
+ */
+std::set<uint8_t> wireless_utils::operating_class_to_channel_set(uint8_t operating_class)
+{
+    static const std::map<uint8_t, std::set<uint8_t>> operating_class_to_channel_set = {
+        {81, {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13}},
+        {82, {14}},
+        {83, {1, 2, 3, 4, 5, 6, 7, 8, 9}},
+        // TODO channels for operating classes 85,96,87 should be taken from the regulatory domain
+        {94, {133, 137}},
+        {95, {132, 134, 136, 138}},
+        {96, {131, 132, 133, 134, 135, 136, 137, 138}},
+        {101, {21, 25}},
+        {102, {11, 13, 15, 17, 19}},
+        {103, {1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+        {104, {184, 192}},
+        {105, {188, 196}},
+        {106, {191, 195}},
+        {107, {189, 191, 193, 195, 197}},
+        {108, {188, 189, 190, 191, 192, 193, 194, 195, 196, 197}},
+        {109, {184, 188, 192, 196}},
+        {110, {183, 184, 185, 186, 187, 189}},
+        {111, {182, 183, 184, 185, 186, 187, 189}},
+        {112, {8, 12, 16}},
+        {113, {7, 8, 9, 10, 11}},
+        {114, {6, 7, 8, 9, 10, 11}},
+        {115, {36, 40, 44, 48}},
+        {116, {36, 44}},
+        {117, {40, 48}},
+        {118, {52, 56, 60, 64}},
+        {119, {52, 60}},
+        {120, {56, 64}},
+        {121, {100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144}},
+        {122, {100, 108, 116, 124, 132, 140}},
+        {123, {104, 112, 120, 128, 136, 144}},
+        {124, {149, 153, 157, 161}},
+        {125, {149, 153, 157, 161, 165, 169}},
+        {126, {149, 157}},
+        {127, {153, 161}},
+        {180, {1, 2, 3, 4, 5, 6}}};
+
+    auto it = operating_class_to_channel_set.find(operating_class);
+    if (it == operating_class_to_channel_set.end()) {
+        LOG(ERROR) << "reserved operating class " << int(operating_class);
+        return std::set<uint8_t>();
+    }
+    return it->second;
+}
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -1232,6 +1232,21 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
     return n->hostap->supported_channels;
 }
 
+std::string db::get_hostap_supported_channels_string(const std::string &radio_mac)
+{
+    std::ostringstream os;
+    auto supported_channels = get_hostap_supported_channels(radio_mac);
+    for (const auto &val : supported_channels) {
+        if (val.channel > 0) {
+            os << " ch = " << int(val.channel) << " | dfs = " << int(val.is_dfs_channel)
+               << " | tx_pow = " << int(val.tx_pow) << " | noise = " << int(val.noise)
+               << " [dbm] | bss_overlap = " << int(val.bss_overlap) << std::endl;
+        }
+    }
+
+    return os.str();
+}
+
 bool db::set_hostap_band_capability(std::string mac, beerocks::eRadioBandCapability capability)
 {
     auto n = get_node(mac);
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -1247,6 +1247,60 @@ std::string db::get_hostap_supported_channels_string(const std::string &radio_ma
     return os.str();
 }
 
+/**
+ * @brief Add supported operating class to the database.
+ * Currently this function is a wrapper which converts the operating
+ * class to a set of supported channels and updates the list of currently
+ * supported channels.
+ * 
+ * @param mac radio mac
+ * @param operating class operating class to add
+ * @tx_power transmit power
+ * @non_operable_channels list of statically non-operable channels
+ * @return true on success
+ * @return false on failure
+ */
+bool db::add_hostap_supported_operating_class(const std::string &radio_mac, uint8_t operating_class,
+                                              uint8_t tx_power,
+                                              const std::vector<uint8_t> &non_operable_channels)
+{
+    auto supported_channels = get_hostap_supported_channels(radio_mac);
+    auto channel_set        = wireless_utils::operating_class_to_channel_set(operating_class);
+
+    // Update current channels
+    for (auto c : channel_set) {
+        auto channel = std::find_if(
+            supported_channels.begin(), supported_channels.end(),
+            [&c](const beerocks_message::sWifiChannel &ch) { return ch.channel == c; });
+        if (channel != supported_channels.end()) {
+            channel->tx_pow = tx_power;
+            //TODO fill other channel parameters
+        } else {
+            beerocks_message::sWifiChannel ch = {0};
+            ch.channel                        = c;
+            ch.tx_pow                         = tx_power;
+            supported_channels.push_back(ch);
+        }
+    }
+
+    // Delete non-operable channels
+    for (auto c : non_operable_channels) {
+        auto channel = std::find_if(
+            supported_channels.begin(), supported_channels.end(),
+            [&c](const beerocks_message::sWifiChannel &ch) { return ch.channel == c; });
+        if (channel != supported_channels.end())
+            supported_channels.erase(channel);
+    }
+
+    set_hostap_supported_channels(radio_mac, &supported_channels[0], supported_channels.size());
+    // dump new supported channels state
+    LOG(DEBUG) << "New supported channels for hostap" << radio_mac << " operating class "
+               << int(operating_class) << std::endl
+               << get_hostap_supported_channels_string(radio_mac);
+
+    return true;
+}
+
 bool db::set_hostap_band_capability(std::string mac, beerocks::eRadioBandCapability capability)
 {
     auto n = get_node(mac);
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -38,9 +38,6 @@
 #include <tlvf/ieee_1905_1/tlvSearchedRole.h>
 #include <tlvf/ieee_1905_1/tlvSupportedFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvSupportedRole.h>
-#include <tlvf/ieee_1905_1/tlvWscM1.h>
-#include <tlvf/ieee_1905_1/tlvWscM2.h>
-#include <tlvf/wfa_map/tlvApRadioBasicCapabilities.h>
 #include <tlvf/wfa_map/tlvApRadioIdentifier.h>
 #include <tlvf/wfa_map/tlvChannelPreference.h>
 #include <tlvf/wfa_map/tlvChannelSelectionResponse.h>
@@ -697,7 +694,7 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
 
     if (intel_agent) {
         LOG(INFO) << "Intel radio agent join (al_mac=" << al_mac << " ruid=" << ruid;
-        if (!handle_intel_slave_join(sd, cmdu_rx, cmdu_tx)) {
+        if (!handle_intel_slave_join(sd, radio_basic_caps, cmdu_rx, cmdu_tx)) {
             LOG(ERROR) << "Intel radio agent join failed (al_mac=" << al_mac << " ruid=" << ruid
                        << ")";
             return false;
@@ -707,7 +704,7 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
         // Multi-AP Agent doesn't say anything about the bridge, so we have to rely on Intel Slave Join for that.
         // We'll use AL-MAC as the bridge
         // TODO convert source address into AL-MAC address
-        if (!handle_non_intel_slave_join(sd, tlvwscM1, al_mac, ruid, cmdu_tx)) {
+        if (!handle_non_intel_slave_join(sd, radio_basic_caps, tlvwscM1, al_mac, ruid, cmdu_tx)) {
             LOG(ERROR) << "Non-Intel radio agent join failed (al_mac=" << al_mac << " ruid=" << ruid
                        << ")";
             return false;
@@ -998,8 +995,9 @@ bool master_thread::handle_cmdu_1905_operating_channel_report(Socket *sd,
     return son_actions::send_cmdu_to_agent(sd, cmdu_tx);
 }
 
-bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx,
-                                            ieee1905_1::CmduMessageTx &cmdu_tx)
+bool master_thread::handle_intel_slave_join(
+    Socket *sd, std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_caps,
+    ieee1905_1::CmduMessageRx &cmdu_rx, ieee1905_1::CmduMessageTx &cmdu_tx)
 {
     auto tlv_vs = cmdu_tx.add_vs_tlv(ieee1905_1::tlvVendorSpecific::eVendorOUI::OUI_INTEL);
     if (!tlv_vs) {
@@ -1412,6 +1410,7 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
     } else {
         database.set_hostap_band_capability(radio_mac, beerocks::SUBBAND_CAPABILITY_UNKNOWN);
     }
+    autoconfig_wsc_parse_radio_caps(radio_mac, radio_caps);
 
     // send JOINED_RESPONSE with son config
     {
@@ -1527,10 +1526,67 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
     return true;
 }
 
-bool master_thread::handle_non_intel_slave_join(Socket *sd,
-                                                std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1,
-                                                std::string bridge_mac, std::string radio_mac,
-                                                ieee1905_1::CmduMessageTx &cmdu_tx)
+/**
+ * @brief Parse the radio basic capabilities TLV and store the operating class
+ * in the database as supported channels.
+ * 
+ * @param radio_mac radio mac address (RUID in non-Intel agent case)
+ * @param radio_caps radio basic capabilities TLV received from the remote agent
+ * @return true on success
+ * @return false on failure
+ */
+bool master_thread::autoconfig_wsc_parse_radio_caps(
+    std::string radio_mac, std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_caps)
+{
+    // read all operating class list
+    auto operating_classes_list_length = radio_caps->operating_classes_info_list_length();
+    if (operating_classes_list_length > beerocks::message::SUPPORTED_CHANNELS_LENGTH) {
+        LOG(WARNING) << "operating class info list larger then maximum supported channels";
+        operating_classes_list_length = beerocks::message::SUPPORTED_CHANNELS_LENGTH;
+    }
+    for (int oc_idx = 0; oc_idx < operating_classes_list_length; oc_idx++) {
+        std::stringstream ss;
+        auto operating_class_tuple = radio_caps->operating_classes_info_list(oc_idx);
+        if (!std::get<0>(operating_class_tuple)) {
+            LOG(ERROR) << "getting operating class entry has failed!";
+            return false;
+        }
+        auto &op_class                  = std::get<1>(operating_class_tuple);
+        auto operating_class            = op_class.operating_class();
+        auto maximum_transmit_power_dbm = op_class.maximum_transmit_power_dbm();
+        ss << "operating_class=" << int(operating_class) << std::endl;
+        ss << "maximum_transmit_power_dbm=" << int(maximum_transmit_power_dbm) << std::endl;
+        ss << "channel list={ ";
+        auto channel_list = son::wireless_utils::operating_class_to_channel_set(operating_class);
+        for (auto channel : channel_list) {
+            ss << int(channel) << " ";
+        }
+        ss << "}" << std::endl;
+        ss << "statically_non_operable_channel_list={ ";
+
+        auto non_oper_channels_list_length =
+            op_class.statically_non_operable_channels_list_length();
+        std::vector<uint8_t> non_operable_channels;
+        for (int ch_idx = 0; ch_idx < non_oper_channels_list_length; ch_idx++) {
+            auto ch_tuple = op_class.statically_non_operable_channels_list(ch_idx);
+            auto channel  = std::get<1>(ch_tuple);
+            ss << int(channel) << " ";
+            non_operable_channels.push_back(channel);
+        }
+        ss << " }" << std::endl;
+        LOG(DEBUG) << ss.str();
+        // store operating class in the DB for this hostap
+        database.add_hostap_supported_operating_class(
+            radio_mac, operating_class, maximum_transmit_power_dbm, non_operable_channels);
+    }
+
+    return true;
+}
+
+bool master_thread::handle_non_intel_slave_join(
+    Socket *sd, std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_caps,
+    std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1, std::string bridge_mac, std::string radio_mac,
+    ieee1905_1::CmduMessageTx &cmdu_tx)
 {
 
     // Multi-AP Agent doesn't say anything about the backhaul, so simulate ethernet backhaul to satisfy
@@ -1640,7 +1696,7 @@ bool master_thread::handle_non_intel_slave_join(Socket *sd,
     // sd is assigned to src bridge mac
     sd->setPeerMac(bridge_mac);
 
-    // TODO database.set_hostap_supported_channels from radio_basic_caps->operating_classes_info_list
+    autoconfig_wsc_parse_radio_caps(radio_mac, radio_caps);
     // TODO assume SSIDs are not hidden
     database.set_hostap_advertise_ssid_flag(radio_mac, true);
 
--- a/controller/src/beerocks/master/son_master_thread.h
+++ b/controller/src/beerocks/master/son_master_thread.h
@@ -22,6 +22,7 @@
 #include <mapf/common/encryption.h>
 #include <tlvf/ieee_1905_1/tlvWscM1.h>
 #include <tlvf/ieee_1905_1/tlvWscM2.h>
+#include <tlvf/wfa_map/tlvApRadioBasicCapabilities.h>
 
 #include <cstddef>
 #include <ctime>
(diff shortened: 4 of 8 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1516,10 +1513,160 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
     return true;
 }
 
-bool master_thread::handle_non_intel_slave_join(Socket *sd,
-                                                std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1,
-                                                std::string bridge_mac, std::string radio_mac,
-                                                ieee1905_1::CmduMessageTx &cmdu_tx)
+static inline std::vector<uint8_t> operating_class_to_channel_set(uint8_t operating_class)
```
morantr: Let's assume we would have wanted this function. It certainly wouldn't be here. it will be on son_wireless_utils.
Second, operating class has also bw property, which you ignored. Nitpickers would add band and dfs.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1516,10 +1513,160 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+            LOG(ERROR) << "getting operating class entry has failed!";
+            return supported_channels;
+        }
+        auto &op_class_channels_rx = std::get<1>(operating_class_tuple);
+        auto operating_class       = op_class_channels_rx.operating_class();
+        auto channel_set           = operating_class_to_channel_set(operating_class);
+        for (auto c : channel_set) {
+            beerocks_message::sWifiChannel ch;
+            ch.channel = c;
+            supported_channels.push_back(ch);
```
morantr: This part of this flow should be implemented as part of the channel selection full flow.
Besides, the fact that we have used channels list until now, does not say we have to continue working that way.
AFAIU, the Multi-AP messages use operating class instead of a channel, so maybe it will be more convenient to save it as it, instead of converting it back and forth.
IMO, print only what you get from the CMDU, and don't store it in the database. When we will implement channel selection we will know better.

morantr (changes requested):

controller/src/beerocks/master/son_master_thread.cpp
tomereli: The only reason I did that is for the Conn map to show if the radio is 5ghz or 2.4ghz. I agree that we should store this as is without converting but we will do this refactoring as part of channel selection.
In the meanwhile it won't hurt to have this conversion even if it is only for the network map, since it's a nice thing to show Ina demo.
In fact, it will also allow us to use our steering algorithm before implementing full channel selection which is also nice to have and we plan to try to get this working for the demo.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: You are right, I will move it to son_wireless_utils. What do you mean by "let's assume we would want this function"? Its not matter of wanting it, I needed it for getting the right channel and understanding weather it is a 5g or 2.4g radio.

Let me remind you that this is still draft, I didn't explain why I did everything I did in this commit so try to keep an open mind:)

I ignored everything other than channel on purpose - all we need at the moment is the channel.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1520,14 +1513,198 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+    }
+    for (int oc_idx = 0; oc_idx < operating_classes_list_length; oc_idx++) {
+        std::stringstream ss;
+        auto operating_class_tuple = radio_caps->operating_classes_info_list(oc_idx);
+        if (!std::get<0>(operating_class_tuple)) {
+            LOG(ERROR) << "getting operating class entry has failed!";
+            return supported_channels;
+        }
+        auto &op_class_channels_rx = std::get<1>(operating_class_tuple);
+        auto operating_class       = op_class_channels_rx.operating_class();
```
arnout: I agree with Moran that it's better to store the operating classes in the database. However, I also agree with Tomer that this is too big a change. So as middle ground, I propose to already add the function to the database to set operating classes, and do the conversion of operating class to channel inside the database.

For this new function, I think it's better to use `add_hostap_supported_operating_class`, not `set_hostap_supported_operating_classes` like is done for the channels (`set_hostap_supported_channels`). That will probably make it easier to separate the parsing from the database function and the database function from how it's internally stored.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1516,10 +1513,160 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
     return true;
 }
 
-bool master_thread::handle_non_intel_slave_join(Socket *sd,
-                                                std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1,
-                                                std::string bridge_mac, std::string radio_mac,
-                                                ieee1905_1::CmduMessageTx &cmdu_tx)
+static inline std::vector<uint8_t> operating_class_to_channel_set(uint8_t operating_class)
+{
+    switch (operating_class) {
```
arnout: Please add a comment (either here or in the doxygen header, but probably here) that explains where you get this table.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1516,10 +1513,160 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+    case 81:
+        return std::vector<uint8_t>{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13};
```
arnout: Instead of this big switch, isn't it better to use a map?
```
std::map<uint8_t, std::vector<uint8_t>> operating_class_to_channel_list = {
    {81, {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13}},
...
};
```

tomereli: Done

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
     return n->hostap->supported_channels;
 }
 
+/**
+ * @brief Add supported operating class to the database.
+ * Currently this function is a wrapper which converts the operating
```
arnout, line 1252: There should be an empty line after the @brief, otherwise the whole paragraph is considered the "brief". (nitpick)

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+ * class to a set of supported channels and updates the list of currently
+ * supported channels.
+ * 
+ * @param mac radio mac
```
arnout, line 1256: In EasyMesh terminology, it would be Radio UID. (nitpcik)

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+ * @param info operating class info from the tlvApRadioBasicCapabilities TLV
+ * @return true on success
+ * @return false on failure
+ */
+bool db::add_hostap_supported_operating_class(std::string mac, wfa_map::cOperatingClassesInfo &info)
```
arnout: So, this is *not* what I meant :-/

I would prefer to not expose the details about how things happen to be structured in the MAP TLVs to the database. So I would expand the structure into an operating class and transmit power argument and a vector of statically non-operable channels.

(nitpick)

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+        auto old =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&](const beerocks_message::sWifiChannel ch) { return ch.channel == c; });
+        if (old != supported_channels.end())
+            supported_channels.erase(old);
+        // Now add the new channel's settings
+        beerocks_message::sWifiChannel ch;
+        ch.channel        = c;
+        ch.tx_pow         = info.maximum_transmit_power_dbm();
+        ch.radar_affected = affected_channels.find(c) != affected_channels.end();
```
arnout: ??? What do *statically* non-operable channels have to do with radar?

I think statically non-operable channels should just be removed from channel_set.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+        affected_channels.insert(ch);
+    }
+
+    for (auto c : channel_set) {
+        // First, delete old channel if exists
+        auto old =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&](const beerocks_message::sWifiChannel ch) { return ch.channel == c; });
+        if (old != supported_channels.end())
+            supported_channels.erase(old);
```
arnout: Well, no... we should keep the old one and update it with the new information - basically, setting tx_pow to the maximum of the old one and the new one.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+        auto ch_tuple = info.statically_non_operable_channels_list(ch_idx);
+        auto ch       = std::get<1>(ch_tuple);
+        affected_channels.insert(ch);
+    }
+
+    for (auto c : channel_set) {
+        // First, delete old channel if exists
+        auto old =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&](const beerocks_message::sWifiChannel ch) { return ch.channel == c; });
```
arnout: Might be worth to make channel_set a map instead of a set.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+            supported_channels.erase(old);
+        // Now add the new channel's settings
+        beerocks_message::sWifiChannel ch;
+        ch.channel        = c;
+        ch.tx_pow         = info.maximum_transmit_power_dbm();
+        ch.radar_affected = affected_channels.find(c) != affected_channels.end();
+        supported_channels.push_back(ch);
+    }
+
+    set_hostap_supported_channels(mac, &supported_channels[0], supported_channels.size());
```
arnout: Instead of getting the channels, adding one, and then setting it, I think it's worthwhile to add an update function to the database, `add_hostap_supported_channels`. This would also do the merging in case it's needed.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -454,4 +454,59 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+ * @param operating_class operating class
+ * @return std::unordered_set<uint8_t> set of supported channels by the operating class or empty if failure
+ */
+std::unordered_set<uint8_t> wireless_utils::operating_class_to_channel_set(uint8_t operating_class)
+{
+    static const std::map<uint8_t, std::unordered_set<uint8_t>> operating_class_to_channel_set = {
+        {81, {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13}},
+        {82, {14}},
+        {83, {1, 2, 3, 4, 5, 6, 7, 8, 9}},
+        {85, {}},
```
arnout: Do you know what this even means? Why do they define operating classes without channels?

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -454,4 +454,59 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+
+/**
+ * @brief convert operating class to channel set based on Table 4-E in the ieee 802.11 specification
+ *
+ * @param operating_class operating class
+ * @return std::unordered_set<uint8_t> set of supported channels by the operating class or empty if failure
+ */
+std::unordered_set<uint8_t> wireless_utils::operating_class_to_channel_set(uint8_t operating_class)
+{
+    static const std::map<uint8_t, std::unordered_set<uint8_t>> operating_class_to_channel_set = {
```
arnout: I should have thought of this before, of course, but: an operating class is more than just a channel set, it also puts limits, including bandwidth. So maybe it would be more appropriate to use sWifiChannel directly?

But that can be done in a later commit.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1653,6 +1699,7 @@ bool master_thread::handle_non_intel_slave_join(Socket *sd,
     //            }
     //        } else {
     database.set_hostap_band_capability(radio_mac, beerocks::SUBBAND_CAPABILITY_UNKNOWN);
+    autoconfig_wsc_parse_radio_caps(radio_mac, radio_caps);
```
arnout: This call should be a little higher, replacing the "TODO database.set_hostap_supported_channels"

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1527,10 +1524,59 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+ * @brief parse the radio basic capabilities TLV and store the operating class
+ * in the database as supported channels.
+ * 
+ * @param radio_mac radio mac address
+ * @param radio_caps radio basic capabilities TLV received from the remote agent
+ * @return true on success
+ * @return false on failure
+ */
+bool master_thread::autoconfig_wsc_parse_radio_caps(
+    std::string radio_mac, std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_caps)
```
arnout, line 1539: In one of my PRs, I passed this stuff by const ref instead of passing the shared_ptr, i.e. dereferencing it in the caller. There was no discussion about that, but maybe this is something we should also agree on: prefer to pass by ref or by pointer (in case the pointer is known to be non-null, obviously).

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1527,10 +1524,59 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+            op_class.statically_non_operable_channels_list_length();
+        for (int ch_idx = 0; ch_idx < non_oper_channels_list_length; ch_idx++) {
+            auto ch_tuple   = op_class.statically_non_operable_channels_list(ch_idx);
+            auto channel_rx = std::get<1>(ch_tuple);
+            ss << int(channel_rx);
+
+            // add comma if not last channel in the list, else close list by add curl brackets
+            ss << (((ch_idx + 1) != non_oper_channels_list_length) ? "," : "}");
+        }
+        LOG(DEBUG) << ss.str();
```
arnout, line 1577: Is it really worth adding all this code for a single debug message?

arnout (approved):
As usual, a lot of comments, but all of them are either nitpicks or something that can be changed later (when we have a better idea of what is really needed).

tomereli:
Do not merge yet. Review however is welcome

controller/src/beerocks/master/db/db.cpp
morantr: radio_mac and radio_uid are two different things and not necessarily equal. The current database convention is to use mac. However, I don't know how could we support it, since I don't know if we are getting the radio mac from somewhere.
Maybe you need to use here the conversion function that convert the al_mac and ruid to database key.

controller/src/beerocks/master/db/db.cpp
morantr: I think both the input argument should be a const reference.

controller/src/beerocks/master/son_master_thread.cpp
morantr: I think i is better to pass is a reference, that way you don't have to do pointer validation (which you didn't do).
Also, radio_mac should be const reference.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1527,10 +1524,59 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
     return true;
 }
 
-bool master_thread::handle_non_intel_slave_join(Socket *sd,
-                                                std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1,
-                                                std::string bridge_mac, std::string radio_mac,
-                                                ieee1905_1::CmduMessageTx &cmdu_tx)
+/**
+ * @brief parse the radio basic capabilities TLV and store the operating class
```
morantr: You forgot to use a capital letter in the sentence first letter.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1527,10 +1524,59 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+        auto operating_class_tuple = radio_caps->operating_classes_info_list(oc_idx);
+        if (!std::get<0>(operating_class_tuple)) {
+            LOG(ERROR) << "getting operating class entry has failed!";
+            return false;
+        }
+        auto &op_class                  = std::get<1>(operating_class_tuple);
+        auto operating_class            = op_class.operating_class();
+        auto maximum_transmit_power_dbm = op_class.maximum_transmit_power_dbm();
+        ss << "operating_class=" << int(operating_class);
+        ss << "maximum_transmit_power_dbm=" << int(maximum_transmit_power_dbm);
```
morantr: You forgot to add ", " before "maximum_transmit_power_dbm".

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1527,10 +1524,59 @@ bool master_thread::handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageR
...
+{
+    // read all operating class list
+    auto operating_classes_list_length = radio_caps->operating_classes_info_list_length();
+    if (operating_classes_list_length > beerocks::message::SUPPORTED_CHANNELS_LENGTH) {
+        LOG(WARNING) << "operating class info list larger then maximum supported channels";
+        operating_classes_list_length = beerocks::message::SUPPORTED_CHANNELS_LENGTH;
+    }
+    for (int oc_idx = 0; oc_idx < operating_classes_list_length; oc_idx++) {
+        std::stringstream ss;
+        auto operating_class_tuple = radio_caps->operating_classes_info_list(oc_idx);
```
morantr, line 1549: Maybe you can to get the pointer to the list once from the tuple, and then tread it as array (access with square brackets).
BTW, we should really get rid of this tuple, and return pointer instead.

controller/src/beerocks/master/son_master_thread.h
```diff
@@ -54,9 +55,10 @@ class master_thread : public beerocks::socket_thread {
     void handle_cmdu_control_ieee1905_1_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
     bool handle_intel_slave_join(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx,
                                  ieee1905_1::CmduMessageTx &cmdu_tx);
-    bool handle_non_intel_slave_join(Socket *sd, std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1,
-                                     std::string bridge_mac, std::string radio_mac,
-                                     ieee1905_1::CmduMessageTx &cmdu_tx);
+    bool handle_non_intel_slave_join(
+        Socket *sd, std::shared_ptr<wfa_map::tlvApRadioBasicCapabilities> radio_caps,
+        std::shared_ptr<ieee1905_1::tlvWscM1> tlvwscM1, std::string bridge_mac,
+        std::string radio_mac, ieee1905_1::CmduMessageTx &cmdu_tx);
```
morantr, line 63: Here as well need to change strings arguments to const reference, and maybe as Arnout has suggested, to pass reference instead of shared_ptr.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
morantr: I wonder why did you prefer unordered_set instead of set. Since there is no intention to add new values, I think there is only gain by having the channels ordered.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -454,4 +454,59 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
\ No newline at end of file
+}
+
+/**
+ * @brief convert operating class to channel set based on Table 4-E in the ieee 802.11 specification
+ *
+ * @param operating_class operating class
+ * @return std::unordered_set<uint8_t> set of supported channels by the operating class or empty if failure
+ */
+std::unordered_set<uint8_t> wireless_utils::operating_class_to_channel_set(uint8_t operating_class)
```
morantr: const reference return

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -454,4 +454,59 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+        {121, {100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144}},
+        {122, {100, 108, 116, 124, 132, 140}},
+        {123, {104, 112, 120, 128, 136, 144}},
+        {124, {149, 153, 157, 161}},
+        {125, {149, 153, 157, 161, 165, 169}},
+        {126, {149, 157}},
+        {127, {153, 161}},
+        {180, {1, 2, 3, 4, 5, 6}}};
+
+    auto channel_set = operating_class_to_channel_set.find(operating_class);
```
morantr: It is not channel_set what you get. It is an iterator to map.
I would change the name to "op_class_it" or "just "it".
```cpp
    auto it = operating_class_to_channel_set.find(operating_class);
    if (it == operating_class_to_channel_set.end()) {
       // code
    }
    auto&  channel_set = it->second;
    return channel_set;
```
nitpick

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -454,4 +454,59 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+    if (channel_set == operating_class_to_channel_set.end()) {
+        LOG(ERROR) << "reserved operating class " << operating_class;
```
morantr: since operating_class is uint8_t, then it needs to be casted int

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+ * 
+ * @param mac radio mac
+ * @param info operating class info from the tlvApRadioBasicCapabilities TLV
+ * @return true on success
+ * @return false on failure
+ */
+bool db::add_hostap_supported_operating_class(std::string mac, wfa_map::cOperatingClassesInfo &info)
+{
+    auto supported_channels = get_hostap_supported_channels(mac);
+    auto channel_set = wireless_utils::operating_class_to_channel_set(info.operating_class());
```
morantr: here also `auto&`

controller/src/beerocks/master/db/db.cpp
morantr: It is better to remove the from the lambda the reference for all, and explicitly supply what you need, and as well use the input argument as referencfe too:
```cpp
std::find_if(supported_channels.begin(), supported_channels.end(),
                         [&c](const beerocks_message::sWifiChannel& ch) { return ch.channel == c; });
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
 }
 
+/**
+ * @brief Add supported operating class to the database.
+ * Currently this function is a wrapper which converts the operating
+ * class to a set of supported channels and updates the list of currently
+ * supported channels.
+ * 
+ * @param mac radio mac
+ * @param info operating class info from the tlvApRadioBasicCapabilities TLV
```
morantr: Capital letter on the first letter of description.

controller/src/beerocks/master/db/db.cpp
morantr: I think it is better to explicitly call the argument "radio_mac" instead of "mac".

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,48 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+ * @return true on success
+ * @return false on failure
+ */
+bool db::add_hostap_supported_operating_class(std::string mac, wfa_map::cOperatingClassesInfo &info)
+{
+    auto supported_channels = get_hostap_supported_channels(mac);
```
morantr: what if "mac" does not exist and the function return empty container?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1552,24 +1555,29 @@ bool master_thread::autoconfig_wsc_parse_radio_caps(
         auto &op_class                  = std::get<1>(operating_class_tuple);
         auto operating_class            = op_class.operating_class();
         auto maximum_transmit_power_dbm = op_class.maximum_transmit_power_dbm();
-        ss << "operating_class=" << int(operating_class);
-        ss << "maximum_transmit_power_dbm=" << int(maximum_transmit_power_dbm);
-        ss << ", statically_non_operable_channel_list={";
+        ss << "operating_class=" << int(operating_class) << std::endl;
+        ss << "maximum_transmit_power_dbm=" << int(maximum_transmit_power_dbm) << std::endl;
+        ss << "channel list={ ";
+        auto channel_list = son::wireless_utils::operating_class_to_channel_set(operating_class);
```
morantr, line 1560: auto&

morantr (changes requested):

controller/src/beerocks/master/db/db.cpp
tomereli: See `handle_non_intel_slave_join` - this is the RUID. We keep the same convention of calling it radio mac only for alignment with the current code base, but @arnout 's comment is correct so I'll add it to the description.

controller/src/beerocks/master/db/db.cpp
tomereli: Why the maximum of the old and new? I changed this to set the tx_pow to the new one.

controller/src/beerocks/master/db/db.cpp
tomereli: I don't understand why - Do you mean it will be simpler if instead of set we use a 1 to 1 mapping - something like a lookup table? so instead of `find_if` we can simply access `channel_map[channel]` to see if it exists?

controller/src/beerocks/master/db/db.cpp
tomereli: Perhaps in a future refactor. I already spent too much time on this side quest so I rather keep this one as is (assuming nitpick - correct me if I'm wrong)

common/beerocks/bcl/source/son/son_wireless_utils.cpp
tomereli: These operating classes are defined as "GeoDB" - they should be used based on the regulatory domain. I will set these as invalid operating class and add a TODO comment. This will be part of the channel selection full flow.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
tomereli: Part of the channel selection full flow then.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Now it is, that I changed the API to receive statically non operable channels.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Fixed (I added newline instead)

controller/src/beerocks/master/db/db.cpp
tomereli: What is the problem in that? supported channels will be empty and it will be filled with the channels in we get according to the operating class.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: auto& where?

morantr: auto& channel_list

tomereli: `operating_class_to_channel_set` doesn't return reference (and it shouldn't)

controller/src/beerocks/master/son_master_thread.cpp
tomereli: I'm not sure I understood your proposal. In any case, I guess this should have been a (nitpick) - you're welcome to change this on a followup PR.

controller/src/beerocks/master/son_master_thread.h
tomereli: This should be in a separate commit / PR and should be a (nitpick)

common/beerocks/bcl/source/son/son_wireless_utils.cpp
tomereli: good point. There was no real reason to prefer unordered set on top of a normal set.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
tomereli: On error we return an empty set, and I don't want to define a static const empty set just for returning const reference. So I'll skip this suggestion (and should have been (nitpick))

common/beerocks/bcl/source/son/son_wireless_utils.cpp
tomereli: changed name to `it`

controller/src/beerocks/master/db/db.cpp
tomereli: same answer as before, we don't return const reference on purpose.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1247,6 +1247,60 @@ std::string db::get_hostap_supported_channels_string(std::string mac)
...
+            ch.tx_pow                         = tx_power;
+            supported_channels.push_back(ch);
+        }
+    }
+
+    // Delete non-operable channels
+    for (auto c : non_operable_channels) {
+        auto channel =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&c](const beerocks_message::sWifiChannel ch) { return ch.channel == c; });
```
morantr: Need reference beerocks_message::sWifiChannel as well

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1247,6 +1247,60 @@ std::string db::get_hostap_supported_channels_string(std::string mac)
...
+                                              const std::vector<uint8_t> &non_operable_channels)
+{
+    auto supported_channels = get_hostap_supported_channels(radio_mac);
+    auto channel_set        = wireless_utils::operating_class_to_channel_set(operating_class);
+
+    // Update current channels
+    for (auto c : channel_set) {
+        auto channel =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&c](const beerocks_message::sWifiChannel ch) { return ch.channel == c; });
```
morantr: Need reference beerocks_message::sWifiChannel as well

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,21 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
     return n->hostap->supported_channels;
 }
 
+std::string db::get_hostap_supported_channels_string(std::string mac)
```
morantr: const std::string &radio_mac

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1232,6 +1232,21 @@ const std::vector<beerocks_message::sWifiChannel> db::get_hostap_supported_chann
...
+{
+    std::ostringstream os;
+    auto supported_channels = get_hostap_supported_channels(mac);
+    for (auto val : supported_channels) {
```
morantr: const auto val&

morantr (changes requested):

morantr (approved):

arnout:
> radio_mac and radio_uid are two different things and not necessarily equal.

Is it useful to make that distinction? AFAIU the multi-AP specification requires the radio UID to be globally unique, which implies it really is the radio MAC address. I think the standard uses UID rather than MAC address because an AP doesn't really have a single MAC address; it has a number of addresses that can be used as BSSID.

That said, for the time being it's probably better to make the distinction, until the controller code has been cleaned up to use radio UID everywhere.

controller/src/beerocks/master/son_master_thread.cpp
arnout: > I'm not sure I understood your proposal. In any case, I guess this should have been a (nitpick) - you're welcome to change this on a followup PR.

Get rid of this tuple -> tlvf should generate code that returns a pointer instead of a tuple, so you don't need all this crazy std::get<1> stuff.

The part about "get the pointer to the list once from the tuple" is something I don't understand either.

PR #246, opened 2019-08-28
Hotfix/correct wsc attributes

arnout:
Rework the handling of the WSC M1/M2 attributes.

This supersedes the second commit of #239


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -544,11 +550,14 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
         return false;
     if (!tlvWscM2->set_serial_number("prpl12345"))
         return false;
-    std::memset(tlvWscM2->uuid_r_attr().data, 0xee, tlvWscM2->uuid_r_attr().data_length);
+    tlvWscM2->primary_device_type_attr().sub_category_id = WSC::WSC_DEV_NETWORK_INFRA_GATEWAY;
+
+    // TODO Maybe the band should be taken from bss_info_conf.operating_class instead?
```
tomereli, line 555: According to the specification - the paragraph above Table 44 in section 12 of the WSC technical specification - the RF bands attribute "can be used to indicate a specific (or group) of RF Bands to which a setting applies".

So my understanding is that we should set it to `0x7` when bss_info_conf is not set (after dev_reset_default or first boot), and according to bss_info_conf otherwise.

tomereli (changes requested):

tomereli (approved):

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -256,13 +256,13 @@ sWscAttrVendorExtension:
...
     _type: uint8_t
-    _value: 0x6
+    _value: 0x0
   subelement_length:
     _type: uint8_t
     _value: 0x1
   subelement_value:
     _type: uint8_t
-    _value: TEARDOWN
+    _value: 0x20
```
morantr: Why did you replace the enum with value?
And why this value is defined as the default value?

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -264,6 +264,33 @@ sWscAttrVersion2:
...
+    _value: WSC_VENDOR_ID_WFA_3
+  subelement_id:
+    _type: uint8_t
+    _value: 0x6
+  subelement_length:
+    _type: uint8_t
+    _value: 0x1
+  subelement_value:
+    _type: uint8_t
+    _value: TEARDOWN
```
morantr: Why here you used enum and on sWscAttrVersion2 you used value?

arnout: > Why here you used enum and on sWscAttrVersion2 you used value?

This is copied from how it was before this PR, that's why it uses an enum.

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
arnout: > Why did you replace the enum with value?

 Good point... will fix.

> And why this value is defined as the default value?

It is the only allowed value for the WFA-WSC Version IE. But maybe you mean it should be `_const_value`? Then you're probably right, but this entire yaml file should use `_const_value`.

morantr: _value_const: TEARDOWN
https://github.com/prplfoundation/prplMesh/tree/master/framework/tlvf#classes

controller/src/beerocks/master/son_master_thread.cpp
arnout: > According to the specification - the paragraph above Table 44 in section 12 of the WSC technical specification - the RF bands attribute "can be used to indicate a specific (or group) of RF Bands to which a setting applies".

No, that sentence only applies to the optional RF Bands attribute in a Credentials/EncryptedData attribute. This is the global (required) RF Bands attribute in the M1/M2. So the relevant sentence is "This attribute is used to indicate a specific RF band that is utilized during message exchange to permit end points and proxies to communicate over a consistent radio interface."


> So my understanding is that we should set it to `0x7` when bss_info_conf is not set (after dev_reset_default or first boot), and according to bss_info_conf otherwise.

Based on the sentence above, it should be set to 0x0 (because no specific RF band is utilized during message exchange). Or maybe, just set it to the same value as M1. 

The BBF implementation sets it to the radio's band in M1, and according to bss_info_conf in M2, and ignores it when receiving M2.

hostapd/wpa_supplicant sets it according to the channel over which the M1 was received. Not much we can do with that...

I'm going to leave it as a TODO :-)

PR #248, opened 2019-08-28
controller:beerocks:cli:fix bml_conn_map command

CoralMalachi:
This pull request fixes the` bml_conn_map` command in the
beerocks CLI.

The following changes were made :

- Move the action of set to false pThis->pending_response to the right place

- Remove the "-w" option from in the beerocks CLI.

### DIFF
--- a/controller/src/beerocks/cli/beerocks_cli_main.cpp
+++ b/controller/src/beerocks/cli/beerocks_cli_main.cpp
@@ -34,7 +34,6 @@ BEEROCKS_INIT_BEEROCKS_VERSION
 
 static bool g_running       = true;
 static bool g_loop_cmd_exec = false;
-static long g_wait_time     = 0;
 
 #ifdef HAVE_READLINE
 static std::vector<std::string> g_cli_cmds;
@@ -237,8 +236,6 @@ static int cli_non_interactive(std::string path, std::string tmp_path, std::stri
     if (-1 == cli_ptr->multiple_commands(command_string)) {
         status_code = 1;
     }
-
-    UTILS_SLEEP_MSEC(1000 * g_wait_time);
 
     // Max timeout according to CAPI specifications is 120 seconds
     auto timeout = std::chrono::steady_clock::now() + std::chrono::seconds(120);
@@ -471,7 +468,7 @@ int main(int argc, char *argv[])
     std::string command_string;
     uint16_t port = 0;
 
-    while ((opt = getopt(argc, argv, "pi:c:w:a:u:")) != -1) {
+    while ((opt = getopt(argc, argv, "pi:c:a:u:")) != -1) {
         switch (opt) {
         case 'p': {
             cli_role = CLI_PROXY;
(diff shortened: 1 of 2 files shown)


controller/src/beerocks/cli/beerocks_cli_main.cpp
```diff
@@ -486,10 +483,6 @@ int main(int argc, char *argv[])
             cli_role       = CLI_NON_INTERACTIVE;
             break;
         }
-        case 'w': {
```
morantr, line 489: Please add to the commit message why did you remove the wait option.

morantr (changes requested):

controller/src/beerocks/cli/beerocks_cli_main.cpp
arnout: > Please add to the commit message why did you remove the wait option.

Whatever the reason: it should probably be a separate commit anyway.

arnout (commented):
Couple of comments about the commit message:

> The following commit moves the action of set to false the

"The following commit" refers to the next commit, not to this one. So it should be "This commit". However, saying "This commit" is usually redundant. As explained in the text about writing a good commit message, you should first explain the problem, then describe the change in the imperative. So the sentence should be "Move the action of set to false `pThis->pending_response` to the right place.", and it should come after the problem description.


> `pThis->pending_response` to the right place.
> 
> In cases the connection map needs to be split to more than one fragment

In case

> nothing is printed, causing a few problems, for example, making the
> AL_MAC be empty.
> 
> The solution is to set  `pThis->pending_response` to true only after we

set to *false*

Also, use the imperative.

> have read the last fragment in the `connection_map_cb` method. then when

then -> Then

> we check if we still need to wait for more fragments at the
> `cli_non_interactive` method we get the right result, prevents us from
> disconnecting at the wrong time.
>

controller/src/beerocks/cli/beerocks_cli_main.cpp
CoralMalachi: Yes, I meant to do this yesterday, I'll do this now.

CoralMalachi:
> Couple of comments about the commit message:
> 
> > The following commit moves the action of set to false the
> 
> "The following commit" refers to the next commit, not to this one. So it should be "This commit". However, saying "This commit" is usually redundant. As explained in the text about writing a good commit message, you should first explain the problem, then describe the change in the imperative. So the sentence should be "Move the action of set to false `pThis->pending_response` to the right place.", and it should come after the problem description.
> 
> > `pThis->pending_response` to the right place.
> > In cases the connection map needs to be split to more than one fragment
> 
> In case
> 
> > nothing is printed, causing a few problems, for example, making the
> > AL_MAC be empty.
> > The solution is to set  `pThis->pending_response` to true only after we
> 
> set to _false_
> 
> Also, use the imperative.
> 
> > have read the last fragment in the `connection_map_cb` method. then when
> 
> then -> Then
> 
> > we check if we still need to wait for more fragments at the
> > `cli_non_interactive` method we get the right result, prevents us from
> > disconnecting at the wrong time.

Ok

controller/src/beerocks/cli/beerocks_cli_main.cpp
CoralMalachi: Fixed.

morantr (approved):

arnout (approved):
Very good commit messages, thanks.

tomereli (approved):

PR #251, opened 2019-08-29
controller: Fix saving of operating class from DEV_SET_CONFIG

morantr:
When receiving bss info configuration as part of CAPI command
"DEV-SET_COMMAND", one of the fields is operating class.
On the documentation example this field appeared to be in the form:
"8x","11x" or "112x".
At first implementation of DEV_SET_CONFIG handler, we assumed that
instead of the 'x' the UCC will be an actual number so we tried to
convert it to integer and save it in the database, but it was wrong
assumption.

Instead of converting to an integer, check which operating class group
we received and save it in the database as a list of relevant operating
classes of this group, according to the EasyMesh Test Plan document.

### DIFF
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -508,7 +508,7 @@ public:
     };
 
     struct bss_info_conf_t {
-        uint8_t operating_class;
+        std::list<uint8_t> operating_class;
         std::string ssid;
         WSC::eWscAuth authentication_type;
         WSC::eWscEncr encryption_type;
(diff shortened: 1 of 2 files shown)


controller/src/beerocks/master/db/db.h
```diff
@@ -508,7 +508,7 @@ class db {
     };
 
     struct bss_info_conf_t {
-        uint8_t operating_class;
+        std::list<uint8_t> operating_class;
```
arnout, line 511: As usual, I'm trying to learn: why use std::list instead of std::vector (and vice versa)?

arnout (approved):

PR #253, opened 2019-08-29
tools:docker:make use of start_gw_repeater method

CoralMalachi:
The `start_gw_repeater` method in `test_flows.sh` is unused.
Make use of this method by calling it in `test_init` method.

### DIFF
--- a/tools/docker/tests/test_flows.sh
+++ b/tools/docker/tests/test_flows.sh
@@ -21,9 +21,13 @@ error=0
 start_gw_repeater() {
     dbg "delete running containers before starting test"
     eval docker rm -f gateway $redirect
-    eval docker rm -f repeater $redirect
-    dbg "start gw and repeater"
-    eval ${scriptdir}/test_gw_repeater.sh -d 5 $redirect
+    eval docker rm -f repeater $redirect    
+    dbg "start gw and repeater"  
+    eval ${scriptdir}/test_gw_repeater.sh -f -d 5 $redirect|| 
+    {        
+        err "start GW+Repeater failed, abort"
+        exit 1
+    }
 }
 
 send_bml_command() {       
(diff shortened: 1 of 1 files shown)


tools/docker/tests/test_flows.sh
```diff
@@ -21,9 +21,13 @@ error=0
 start_gw_repeater() {
     dbg "delete running containers before starting test"
     eval docker rm -f gateway $redirect
-    eval docker rm -f repeater $redirect
-    dbg "start gw and repeater"
-    eval ${scriptdir}/test_gw_repeater.sh -d 5 $redirect
+    eval docker rm -f repeater $redirect    
+    dbg "start gw and repeater"  
+    eval ${scriptdir}/test_gw_repeater.sh -f -d 5 $redirect||
```
tomereli, line 26: please add a space before `||`

tomereli (changes requested):
I really don't understand why we need this commit - the start_gw_repeater function is not doing anything really other than call the script. I prefer to simply remove it and keep the usage we have now. Unless you mean to change that to start more than one repeater?

tools/docker/tests/test_flows.sh
```diff
@@ -21,9 +21,13 @@ error=0
 start_gw_repeater() {
     dbg "delete running containers before starting test"
     eval docker rm -f gateway $redirect
-    eval docker rm -f repeater $redirect
-    dbg "start gw and repeater"
-    eval ${scriptdir}/test_gw_repeater.sh -d 5 $redirect
+    eval docker rm -f repeater $redirect
```
tomereli, line 24: This is not needed anymore since the `-f` argument to `test_gw_repeater.sh` will remove the repeater.

tools/docker/tests/test_flows.sh
```diff
@@ -144,12 +148,10 @@ test_topology() {
 }
 test_init() {
     status "test initialization"
-
-    eval ${scriptdir}/test_gw_repeater.sh -f -d 5 $redirect || {
-        err "start GW+Repeater failed, abort"
-        exit 1
-    }
-    AL_MAC=$(docker exec -it gateway ${installdir}/bin/beerocks_cli -c bml_conn_map | grep IRE_BRIDGE | awk '{print $5}' | cut -d ',' -f 1)
+    eval start_gw_repeater $redirect
```
arnout, line 151: There should be no redirect here.

arnout (changes requested):
I agree with Tomer, I don't see a reason to use the start_gw_repeater function since it's anyway used in only one place.

So instead, I'd just remove the definition of the function.

tools/docker/tests/test_flows.sh
CoralMalachi: Yes, in my branch the start_gw_repeater deletes 3 containers since I have 2 repeaters.
maybe in the future, the start_gw_repeater method would be more necessary.

tools/docker/tests/test_flows.sh
CoralMalachi: Ok.

tools/docker/tests/test_flows.sh
CoralMalachi: Ok

tomereli:
@CoralMalachi Can we close this PR? Or do you still intend to update it and merge it?

CoralMalachi:
> @CoralMalachi Can we close this PR? Or do you still intend to update it and merge it?

I'll update this today

Issue #214, opened 2019-08-22
5.8.1 Client Steering for Steering Mandate and Steering Opportunity


c0dingn0mad:
## Description

The prplMesh controller needs to pass the test case as defined in 5.8.1 

## Exit criteria

The flow is completed once the controller passes this test in the scenario and testbed described in the upstream definition.

- [x] #212 
- [x] #213

Issue #212, opened 2019-08-22
[TASK] complete development needed for flow 5.8.1


c0dingn0mad:
## Description

Implement all code required to complete perform flow 5.8.1

- [ ] Controller handles Steering BTM Report by writing a log message and sending an Ack.
- [ ] Controller handles Steering Completed by writing a log message and sending an Ack.
- [ ] test_flows.sh is updated with the 5.8.1 flow.
- [ ] Agent handles Client Steering Request Mandate by writing a log message, sending an Ack and sending a dummy Steering BTM Report.
- [ ] Agent handles Client Steering Request Opportunity by writing a log message, sending an Ack and sending a dummy Steering Completed.

## Exit criteria

The task will be complete when the test script has been properly written, and necessary components have been implemented, running the corresponding test in test_flows.sh returns an OK, and .travis.yml has been updated accordingly.

PR #257, opened 2019-09-01
Feature/add client steering flow

CoralMalachi:
According to section 5.8.1 in the EasyMesh Test Plan v1.0 [1], 
client steering for Steering Mandate and Steering Opportunity test.

According to Multi-AP Specification v1.0, the client steering control 
message enable efficient steering of STAs between BSSs in a Multi-AP 
network. There are two message types:

1.  steering mandate : 
Request Mode bit set to one indicating a Steering Mandate.
The controller uses the Steering Mandate mechanism to
mandate an agent to attempt steering of one or more associated STAs. 

2. steering opportunity :
Request Mode bit set to zero indicating a Steering Opportunity.
The controller uses the Steering Opportunity mechanism to provide a time
window for an agent to steer one or more associated STAs.
The agent may attempt to steer the STA(s) identified in the request to any
other BSS in the Multi-AP network that the Multi-AP Agent has identified as
suitable target BSS(s) 

#212 #214

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -489,6 +489,8 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         return handle_channel_selection_request(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CLIENT_STEERING_REQUEST_MESSAGE:
         return handle_client_steering_request(sd, cmdu_rx);
+    case ieee1905_1::eMessageType::ACK_MESSAGE:
+        return handle_ack_message(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE:
         return handle_client_capability_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE:
@@ -4817,6 +4819,15 @@ bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::Cmd
     }
     LOG(DEBUG) << "sending ACK message back to controller";
     return send_cmdu_to_controller(cmdu_tx);
+}
+
+bool slave_thread::handle_ack_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    //TODO - this is a stub handler for the purpose of controller certification testing,
+    //       will be implemented later on agent certification
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received ACK_MESSAGE, mid=" << std::dec << int(mid);
+    return true;
 }
 
 bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -32,6 +32,8 @@
 #include <tlvf/wfa_map/tlvChannelPreference.h>
 #include <tlvf/wfa_map/tlvChannelSelectionResponse.h>
 #include <tlvf/wfa_map/tlvOperatingChannelReport.h>
+#include <tlvf/wfa_map/tlvSteeringBTMReport.h>
+#include <tlvf/wfa_map/tlvSteeringRequest.h>
 #include <tlvf/wfa_map/tlvTransmitPowerLimit.h>
 
 // BPL Error Codes
@@ -485,6 +487,8 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         return handle_channel_preference_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE:
         return handle_channel_selection_request(sd, cmdu_rx);
+    case ieee1905_1::eMessageType::CLIENT_STEERING_REQUEST_MESSAGE:
+        return handle_client_steering_request(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE:
         return handle_client_capability_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE:
@@ -4812,6 +4816,66 @@ bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::Cmd
         return false;
     }
     LOG(DEBUG) << "sending ACK message back to controller";
+    return send_cmdu_to_controller(cmdu_tx);
+}
+
+bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received CLIENT_STEERING_REQUEST_MESSAGE , mid=" << std::hex << int(mid);
+    auto steering_request_tlv = cmdu_rx.addClass<wfa_map::tlvSteeringRequest>();
+    if (!steering_request_tlv) {
+        LOG(ERROR) << "addClass wfa_map::tlvSteeringRequest failed";
+        return false;
+    }
+
+    auto request_mode = steering_request_tlv->request_flags().request_mode;
+    LOG(DEBUG) << "request_mode: " << std::hex << int(request_mode);
+
+    //TODO: validate the BTM request before sending the ack.
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+
+    LOG(DEBUG) << "sending ACK message back to controller";
+    if (!send_cmdu_to_controller(cmdu_tx)) {
+        LOG(ERROR) << "failed to send ACK_MESSAGE";
+        return false;
+    }
+    //TODO - Send BTM Request frame to the STA including a Neighbor Report element specifying
+    // the BSSID, Operating Class and Channel Number of the identified target BSS
+
+    //check if the Request Mode bit is set
+    if (request_mode) {
+        //Steering Mandate
+        LOG(DEBUG) << "Request Mode bit is set - Steering Mandate";
+        // build and send Client Steering BTM report message
+        if (!cmdu_tx.create(0, ieee1905_1::eMessageType::CLIENT_STEERING_BTM_REPORT_MESSAGE)) {
+            LOG(ERROR) << "cmdu creation of type CLIENT_STEERING_BTM_REPORT_MESSAGE, has failed";
+            return false;
+        }
+        auto steering_btm_report_tlv = cmdu_tx.addClass<wfa_map::tlvSteeringBTMReport>();
+        if (!steering_btm_report_tlv) {
+            LOG(ERROR) << "addClass wfa_map::tlvSteeringBTMReport failed";
+            return false;
+        }
+        LOG(DEBUG) << "sending CLIENT_STEERING_BTM_REPORT_MESSAGE back to controller";
+    } else {
+        LOG(DEBUG) << "Request Mode bit is not set - Steering Opportunity";
+        // build and send steering completed message
+        auto cmdu_tx_header =
+            cmdu_tx.create(0, ieee1905_1::eMessageType::STEERING_COMPLETED_MESSAGE);
+
+        if (!cmdu_tx_header) {
+            LOG(ERROR) << "cmdu creation of type STEERING_COMPLETED_MESSAGE, has failed";
+            return false;
+        }
+        LOG(DEBUG) << "sending STEERING_COMPLETED_MESSAGE back to controller";
+    }
     return send_cmdu_to_controller(cmdu_tx);
 }
 
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -436,8 +436,6 @@ bool slave_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
         }
         }
     } else { // IEEE 1905.1 message
-        LOG(DEBUG) << "handle_cmdu_control_ieee1905_1_message " << std::hex
-                   << int(cmdu_rx.getMessageType());
         return handle_cmdu_control_ieee1905_1_message(sd, cmdu_rx);
     }
     return true;
--- a/controller/src/beerocks/master/beerocks_master_mrouter.cpp
+++ b/controller/src/beerocks/master/beerocks_master_mrouter.cpp
@@ -59,6 +59,7 @@ bool master_mrouter::init()
         ieee1905_1::eMessageType::OPERATING_CHANNEL_REPORT_MESSAGE,
         ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE,
         ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE,
+        ieee1905_1::eMessageType::STEERING_COMPLETED_MESSAGE,
         ieee1905_1::eMessageType::ACK_MESSAGE,
     }));
 }
--- a/controller/src/beerocks/master/beerocks_master_mrouter.cpp
+++ b/controller/src/beerocks/master/beerocks_master_mrouter.cpp
@@ -55,6 +55,7 @@ bool master_mrouter::init()
         ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_WSC_MESSAGE,
         ieee1905_1::eMessageType::CHANNEL_PREFERENCE_REPORT_MESSAGE,
         ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE,
+        ieee1905_1::eMessageType::CLIENT_STEERING_BTM_REPORT_MESSAGE,
         ieee1905_1::eMessageType::OPERATING_CHANNEL_REPORT_MESSAGE,
         ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE,
         ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE,
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -224,6 +224,8 @@ bool master_thread::handle_cmdu_1905_1_message(Socket *sd, ieee1905_1::CmduMessa
         return handle_cmdu_1905_channel_preference_report(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE:
         return handle_cmdu_1905_channel_selection_response(sd, cmdu_rx);
+    case ieee1905_1::eMessageType::STEERING_COMPLETED_MESSAGE:
+        return handle_cmdu_1905_steering_completed_message(sd, cmdu_rx);
     case ieee1905_1::eMessageType::ACK_MESSAGE:
         return handle_cmdu_1905_ack_message(sd, cmdu_rx);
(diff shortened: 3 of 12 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4729,6 +4731,14 @@ bool slave_thread::handle_client_capability_query(Socket *sd, ieee1905_1::CmduMe
     return true;
 }
 
+bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received CLIENT_STEERING_REQUEST_MESSAGE , mid=" << std::dec << int(mid);
```
morantr: I would print the mid in hex format (std::hex). It is easier to read it that way.
(nitpick)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -902,6 +904,13 @@ bool master_thread::handle_cmdu_1905_channel_preference_report(Socket *sd,
     return son_actions::send_cmdu_to_agent(sd, cmdu_tx);
 }
 
+bool master_thread::handle_cmdu_1905_ack_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received ACK_MESSAGE, mid=" << std::dec << int(mid);
```
morantr: std::hex (nitpick)

controller/src/beerocks/master/beerocks_master_mrouter.cpp
```diff
@@ -55,6 +55,7 @@ bool master_mrouter::init()
         ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_WSC_MESSAGE,
         ieee1905_1::eMessageType::CHANNEL_PREFERENCE_REPORT_MESSAGE,
         ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE,
+        ieee1905_1::eMessageType::ACK_MESSAGE,
```
morantr: Probably it will be best if you add it last, I can see how clang will make it ugly that way...

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4748,6 +4741,27 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
...
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+    auto error_code_tlv = cmdu_tx.addClass<wfa_map::tlvErrorCode>();
```
morantr: AFAIU, we need to add error_code_tlv only if the station received in the command is not connected to the AP.
I don't think we want to simulate error scenarios.
What actually does missing is sending back CMDU of BTM REPORT (0x9c).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4759,7 +4760,28 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+    // the BSSID, Operating Class and Channel Number of the identified target BSS
+
+    // build and send Client Steering BTM report message
+    if (!cmdu_tx.create(0, ieee1905_1::eMessageType::CLIENT_STEERING_BTM_REPORT_MESSAGE)) {
+        LOG(ERROR) << "cmdu creation of type CLIENT_STEERING_BTM_REPORT_MESSAGE, has failed";
+        return false;
+    }
+
+    auto seering_btm_report_tlv = cmdu_tx.addClass<wfa_map::tlvSteeringBTMReport>();
+    if (!error_code_tlv) {
```
morantr: you check the wrong tlv (guess it is a copy-paste error)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4759,7 +4760,28 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
-    return send_cmdu_to_controller(cmdu_tx);
+    if (!send_cmdu_to_controller(cmdu_tx)) {
+        LOG(ERROR) << "failed to send ACK_MESSAGE";
+        return false;
+    }
+    //TODO - Send BTM Request frame to the STA including a Neighbor Report element specifying
+    // the BSSID, Operating Class and Channel Number of the identified target BSS
+
+    // build and send Client Steering BTM report message
+    if (!cmdu_tx.create(0, ieee1905_1::eMessageType::CLIENT_STEERING_BTM_REPORT_MESSAGE)) {
```
morantr: Oh I see you added the btm response here

morantr (dismissed):

morantr:
Please do all the changes related to the handling of the steer request and sending the response to a single commit.

BTW, I assigned myself to this flow before the weekend and was surprised to see you have taken it.

agent/src/beerocks/slave/son_slave_thread.cpp
CoralMalachi: you are right, thanks!

tools/docker/tests/test_flows.sh
```diff
@@ -149,7 +149,7 @@ test_init() {
         err "start GW+Repeater failed, abort"
         exit 1
     }
-    AL_MAC=$(docker exec -it gateway ${installdir}/bin/beerocks_cli -c bml_conn_map | grep IRE_BRIDGE | awk '{print $5}' | cut -d ',' -f 1)
+    AL_MAC=$(docker exec -it gateway ${installdir}/bin/beerocks_cli -c bml_conn_map | grep IRE_BRIDGE| head -1 | awk '{print $5}' | cut -d ',' -f 1)
```
arnout: I would replace this with an explicit `aa:bb:cc:dd:00:30` (or whatever the correct MAC address is). It may feel a little fragile, but this entire script heavily relies on the exact way that we set up containers that I don't think relying on a specific MAC address is too much to ask.

(nitpick)

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -60,6 +60,9 @@ main() {
         sleep ${DELAY}
         status "Start Repeater (Remote Agent)"
         ${scriptdir}/../run.sh ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${REPEATER_NAME} -m aa:bb:cc:dd "$@"
+        
+        status "Start Repeater2 (Remote Agent2)"
+        ${scriptdir}/../run.sh ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${REPEATER2_NAME} -m aa:bb:cc:ee "$@"
```
arnout: I would prefer that there would be a command-line option that specifies the number of agents, and that they're started in a loop. The MAC address can be set to `aa:bb:cc:$n$n` where n is the iterator.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -478,6 +480,10 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         return handle_channel_preference_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE:
         return handle_channel_selection_request(sd, cmdu_rx);
+    case ieee1905_1::eMessageType::CLIENT_STEERING_REQUEST_MESSAGE:
+        return handle_client_steering_request(sd, cmdu_rx);
+    case ieee1905_1::eMessageType::ACK_MESSAGE:
```
arnout, line 490: I may get in conflict with @morantr here :-) but I prefer a separate commit for the handling of the ACK message.

However, at least on the longer term, I think it's more appropriate if the ACK messages are handled in the transport. I believe that retransmissions are currently done there as well.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4738,6 +4744,50 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
...
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received CLIENT_STEERING_REQUEST_MESSAGE , mid=" << std::hex << int(mid);
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+
```
arnout, line 4852: There should be a TODO here that we have to validate the BTM request before sending the ack.

controller/src/beerocks/master/beerocks_master_mrouter.cpp
```diff
@@ -55,7 +55,9 @@ bool master_mrouter::init()
         ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_WSC_MESSAGE,
         ieee1905_1::eMessageType::CHANNEL_PREFERENCE_REPORT_MESSAGE,
         ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE,
+        ieee1905_1::eMessageType::CLIENT_STEERING_BTM_REPORT_MESSAGE,
```
arnout, line 58: Again, contrary to @morantr, I think the controller and the agent part should be in separate commits.

framework/tlvf/tlvf_conf.yaml
```diff
@@ -18,6 +19,8 @@ include_yaml_path: {
   "tlvf/wfa_map/tlvRadioOperationRestriction.yaml",
   "tlvf/wfa_map/tlvChannelSelectionResponse.yaml",
   "tlvf/wfa_map/tlvTransmitPowerLimit.yaml",
+  "tlvf/wfa_map/tlvSteeringBTMReport.yaml",
```
arnout: This should definitely be a separate commit. Also, the AutoGenerated seems to be missing.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -902,6 +906,32 @@ bool master_thread::handle_cmdu_1905_channel_preference_report(Socket *sd,
     return son_actions::send_cmdu_to_agent(sd, cmdu_tx);
 }
 
+bool master_thread::handle_cmdu_1905_ack_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received ACK_MESSAGE, mid=" << std::hex << int(mid);
```
arnout: For debugging it will be useful to also print the error code, if any.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: Where did you get this idea? I have the same opinion as you. :)

CoralMalachi: I'll update my changes soon.

tools/docker/tests/test_gw_repeater.sh
tomereli: Agree with Arnout, we can update the `-r|--repeater` argument as follows:
```
while true; do
    case "$1" in
        ...
        -r | --repeater) REPEATER_NAME="$REPEATER_NAME $2"; shift; shift ;;
        ...
    esac
done
```

And then create multiple repeaters by iterating through `REPEATER_NAME` as follows:
```
for repeater in $REPEATER_NAME
do
    # Start repeater with $repeater name
done
```

tomereli (dismissed):

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4741,6 +4744,42 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
     return true;
 }
 
+bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received CLIENT_STEERING_REQUEST_MESSAGE , mid=" << std::hex << int(mid);
+    //TODO: validate the BTM request before sending the ack.
```
arnout, line 4844: I think the minimum validation we should to is to check the BSSID. Indeed, both slaves are handling this request, so both of them will send an Ack and a BTM Report. This is clearly wrong.

However, it doesn't matter for certification, and this will anyway be rewritten extensively when we complete the agent implementation, so (nitpick).

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -222,13 +222,17 @@ bool master_thread::handle_cmdu_1905_1_message(Socket *sd, ieee1905_1::CmduMessa
...
+    case ieee1905_1::eMessageType::ACK_MESSAGE:
+        return handle_cmdu_1905_ack_message(sd, cmdu_rx);
     case ieee1905_1::eMessageType::OPERATING_CHANNEL_REPORT_MESSAGE:
         return handle_cmdu_1905_operating_channel_report(sd, cmdu_rx);
     default:
         break;
     }
 
-    LOG(WARNING) << "Unknown 1905 message received. Ignoring";
+    LOG(WARNING) << "Unknown 1905 message received. Ignoring, Message mid=" << std::hex
```
arnout: Good change, but probably should have been a separate commit. (nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4744,6 +4746,15 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
     return true;
 }
 
+bool slave_thread::handle_ack_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
```
arnout: I'm ambivalent about this one... As per #273 there are two options for handling acks: either in the transport, or in each component. However, if acks are handled in the component, they will need to be handled both in the backhaul manager (for messages that are emitted from there) and in the slaves - and the backhaul manager should only forward the acks that do not correspond to a message it sent itself.

I guess for now, however, it's OK to handle the acks in both slaves.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4759,6 +4759,10 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
 {
     const auto mid = cmdu_rx.getMessageId();
     LOG(DEBUG) << "Received CLIENT_STEERING_REQUEST_MESSAGE , mid=" << std::hex << int(mid);
+    auto rx_buffer        = cmdu_rx.getMessageBuff();
+    uint16_t request_mode = (uint16_t)(rx_buffer[17]);
```
arnout: I really don't like this... The TLV has been defined already, so why not do the proper addClass and get the request_mode?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4776,18 +4780,33 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+        }
+        auto steering_btm_report_tlv = cmdu_tx.addClass<wfa_map::tlvSteeringBTMReport>();
+        if (!steering_btm_report_tlv) {
+            LOG(ERROR) << "addClass wfa_map::tlvSteeringBTMReport failed";
+            return false;
+        }
+        LOG(DEBUG) << "sending CLIENT_STEERING_BTM_REPORT_MESSAGE back to controller";
+    } else {
+        LOG(DEBUG) << "Request Mode bit is not set - Steering Opportunity";
+        // build and send steering complited message
```
arnout: complited -> completed

arnout (changes requested):
There's a typo in the "remove handle_cmdu" commit: neseccary -> necessary

I do want the parsing of the steering request TLV to be fixed before merging. The rest are nitpicks or general remarks.

CoralMalachi:
@tomereli @arnout 
Do you want me to do rebase and autosquash all the fixup commits?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -961,10 +963,26 @@ bool master_thread::handle_cmdu_1905_channel_preference_report(Socket *sd,
...
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received STEERING_COMPLETED_MESSAGE, mid=" << std::hex << int(mid);
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+    LOG(DEBUG) << "sending ACK message back to agent";
```
tomereli: Add MID to the log

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4759,8 +4760,14 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
 {
     const auto mid = cmdu_rx.getMessageId();
     LOG(DEBUG) << "Received CLIENT_STEERING_REQUEST_MESSAGE , mid=" << std::hex << int(mid);
-    auto rx_buffer        = cmdu_rx.getMessageBuff();
-    uint16_t request_mode = (uint16_t)(rx_buffer[17]);
+    auto steering_btm_report_tlv = cmdu_rx.addClass<wfa_map::tlvSteeringRequest>();
```
arnout: It's a steering request, not a BTM report.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4759,8 +4760,14 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+    if (!steering_btm_report_tlv) {
+        LOG(ERROR) << "addClass wfa_map::tlvSteeringRequest failed";
+        return false;
+    }
+
+    // auto x = steering_btm_report_tlv->request_flags()
```
arnout: What does this comment mean?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4741,6 +4745,76 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
...
+    LOG(DEBUG) << "sending ACK message back to controller";
+    if (!send_cmdu_to_controller(cmdu_tx)) {
+        LOG(ERROR) << "failed to send ACK_MESSAGE";
+        return false;
+    }
+    //TODO - Send BTM Request frame to the STA including a Neighbor Report element specifying
+    // the BSSID, Operating Class and Channel Number of the identified target BSS
+
+    //check if the Request Mode bit is set
+    if ((request_mode >> (7)) & 1) {
```
arnout: Is this correct? I thought that request_mode was a one-bit value already. IOW, I'd expect
```suggestion
    if (request_mode) {
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -222,13 +220,21 @@ bool master_thread::handle_cmdu_1905_1_message(Socket *sd, ieee1905_1::CmduMessa
...
+        return handle_cmdu_1905_client_steering_btm_report_message(sd, cmdu_rx);
     case ieee1905_1::eMessageType::OPERATING_CHANNEL_REPORT_MESSAGE:
         return handle_cmdu_1905_operating_channel_report(sd, cmdu_rx);
     default:
         break;
     }
 
-    LOG(WARNING) << "Unknown 1905 message received. Ignoring";
+    LOG(WARNING) << "Unknown 1905 message received. Ignoring, Message mid=" << std::hex
+                 << int(cmdu_rx.getMessageType());
```
arnout: I didn't notice before: you're not printing the MID here, but the type. You probably want both.

arnout (changes requested):
Please rebase -i on master for a final review (and fixup the small remaining issues in the process).

Make sure the tlv_conf.yaml + AutoGenerated update are in a separate commit.

agent/src/beerocks/slave/son_slave_thread.cpp
CoralMalachi: Yes, you are right, that wad before I used the addClass, I'll change it now.

arnout (approved):

PR #258, opened 2019-09-01
controller: Refactor WFA-CA DEV_SEND_1905 handling

morantr:
When DEV_SEND_1905 command was received with no TLVs, we had to try to
create the CMDU from scratch because there was no way to know if it was
really CMDU with no TLVs or it just CMDU which we need to fill the TLVs
manually from a preprepared buffer.

Refactor the handler so it will try automatically to create the CMDU
from a preprepared buffer, and if there is no preprepared buffer from that
message type it will skip it, and add the TLV's from the command.
This saves confusion to if we need to create a handler for a specific
message type or not, and shorten the code since it is no more necessary
to create CMDU, for command with no TLVs.

### DIFF
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -489,8 +489,6 @@ void wfa_ca::handle_wfa_ca_message(
         std::unordered_map<std::string, std::string> params{{"destalid", std::string()},
                                                             {"messagetypevalue", std::string()}};
 
-        const auto mandatory_params_num = params.size();
-
         if (!parse_params(cmd_tokens_vec, params, err_string)) {
             LOG(ERROR) << err_string;
             reply(sd, cmdu_tx, eWfaCaStatus::INVALID, err_string);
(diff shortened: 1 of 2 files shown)


controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -534,9 +532,32 @@ void wfa_ca::handle_wfa_ca_message(
...
-        // Check if additional tlvs were added to command. If yes, then needs to parse the tlv and
-        // send it as it.
-        if (mandatory_params_num < params.size()) {
+        // Check if CMDU message type has preprepared CMDU which can be loaded
+        auto certification_tx_buffer = database.get_certification_tx_buffer();
+        if (!certification_tx_buffer) {
+            LOG(ERROR) << "certification_tx_buffer is not allocated";
+        }
+        ieee1905_1::CmduMessageTx external_cmdu_tx(
+            certification_tx_buffer.get() + sizeof(beerocks::message::sUdsHeader),
```
simonbaren, line 543: possible segfault. 
you check for null before but didn't return. so possibly certification_tx_buffer is nullptr.

tomereli (commented):
PR / commit message typo

> " and if the is no preprepared " 

--> " and if there is no preprepared"

morantr:
> PR / commit message typo
> 
> > " and if the is no preprepared "
> 
> --> " and if there is no preprepared"

I don't see the typo (btw, I have put the commit message in Grammarly and it said it's ok).

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -676,48 +678,26 @@ void wfa_ca::handle_wfa_ca_message(
...
- * @return true if successful, false if not.
+ * @param[in] message_type CMDU message type to create.
+ * @param[in] database Controllers database.
+ * @param[out] err_string Contains an error description if the function fails.
+ * @return true if CMDU was successfully loaded, false with empty err_string if CMDU was not loaded,
+ * but didn't fail, false with non-empty err_string on failure.
  */
-bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessageType message_type)
+bool wfa_ca::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx, ieee1905_1::eMessageType message_type,
+                         db &database, std::string &err_string)
```
arnout, line 694: You don't actually use the db argument...

But I guess this is a cherry-pick issue due to some other change you're making, so I'll mark that as (nitpick).

arnout (approved):
I think it would be better to have the following logic:

1. Create a CMDU of the correct type
2. Handle TLV arguments
3. call create_cmdu
4. send tx_cmdu

I.e., always call create_cmdu, regardless of whether there were TLV arguments or not.

However, that approach does not work with how we're doing this certification buffer at the moment. I thought about it before and I couldn't think of a way to make it work with the certification buffer. Maybe copy the contents of certification buffer to tx_cmdu's buffer? Feels like even more than a hack...

Therefore, I'm approving as is.

arnout:
> > PR / commit message typo
> > > " and if the is no preprepared "
> > 
> > 
> > --> " and if there is no preprepared"
> 
> I don't see the typo (btw, I have put the commit message in Grammarly and it said it's ok).

As Tomer wrote: the -> there. But since there are several "the" in the commit message, he had to give some context, which kind of hides the fix. Grammarly obviously doesn't find all grammar mistakes.

controller/src/beerocks/master/wfa_ca.cpp
morantr: It is preparation for taking data from the database as we talked today.
I mention it on the @breif.

morantr:
> > > PR / commit message typo
> > > > " and if the is no preprepared "
> > > 
> > > 
> > > --> " and if there is no preprepared"
> > 
> > 
> > I don't see the typo (btw, I have put the commit message in Grammarly and it said it's ok).
> 
> As Tomer wrote: the -> there. But since there are several "the" in the commit message, he had to give some context, which kind of hides the fix. Grammarly obviously doesn't find all grammar mistakes.

Thanks, for some reason, when I read Tomer's comment, both sentences looked identical.

PR #259, opened 2019-09-02
bugfix: prevent segmentation faults when receiving mismatched TLVs

### LINKED ISSUE #234: [BUG] tlvf: check struct size for dynamic-length members
tlvf generates an `init()` function for classes that parses the class's substructures. This function has an initial check for the minimum length:

```
    if (getBuffRemainingBytes() < kMinimumLength) {
        TLVF_LOG(ERROR) << "Not enough available space on buffer. Class init failed";
        return false;
    }
```

However, there is no such check for subsequent dynamic/variable length substructures. Therefore, parsing an invalid packet may lead to an out-of-bounds read.

## Steps to reproduce

This can easily be reproduced with the M1 and M2 WSC TLVs, since they have the same type but different internal structure. Thus, if an M2 message is interpreted as an M1 or vice versa, havoc will result.

1. Run controller & agent for the gateway
2. Add a non-Intel agent that puts the M1 WSC TLV before the AP Radio Basic Capabilities TLV.
3. Start AP Autoconfig from the non-Intel agent
4. With a bit of luck, the beerocks agent on the gateway crashes

## Proposed solution

Instead of checking getBuffRemainingBytes only at the beginning of init(), do it before each substructure that is parsed, including the fixed-length substructures. It is also needed for fixed-length substructures because a fixed-length structure coming after a dynamic-length substructure may be the one that triggers the overflow.

Note that the reproducer would also be fixed by checking the correctness of const_value substructures, because we'd immediately recognise that it's in fact not an M2 message. However, we still want to fix the fundamental issue to protect against intentionally misconstructed packets.

Galvien:
Currently, a known bug causes the radio agent threads to segfault and
die as a result of receiving an M1 message when expecting an M2 message,
the agents attempt to parse an M2 TLV but if they end up receiving an M1
message.
Upon inspection, this is caused when `m_buff_ptr__` is incremented by
incorrect amounts as a result of the mismatch, causing a segmentation
fault.
**Of course, this issue could come up with other message types so it has to be handled on a larger scale.**
This pull requests implements changes that verify the message parsing process does not cause segmentation faults.
The method of action is the creation of the new `buffPtrIncrementSafe()` method and it's implementation in the relevant files generated by TLVF .

Closes #234

### DIFF
--- a/framework/tlvf/src/src/BaseClass.cpp
+++ b/framework/tlvf/src/src/BaseClass.cpp
@@ -21,6 +21,23 @@ size_t BaseClass::getBuffRemainingBytes(void *start)
 {
     return (m_buff_len__ - (size_t)(((uint8_t *)(start ? start : m_buff_ptr__)) - ((uint8_t *)m_buff__)));
 }
+
 size_t BaseClass::getLen() { return (size_t)((uint8_t *)m_buff_ptr__ - (uint8_t *)m_buff__); }
 
 bool BaseClass::isInitialized() { return m_init_succeeded; }
+
+/**
+ * @brief Increments m_buff_ptr__ pointer while avoiding out-of-bounds accesses.
+
+ * @param length size to increment by.
+ * @return true if incrementation is possible.
+ * @return false if incrementation is not possible and the process should be halted to avoid a segfault.
+ */
+bool BaseClass::buffPtrIncrementSafe(size_t length)
+{
+    if (getBuffRemainingBytes() <= length) {
+        return false;
+    }
+    m_buff_ptr__ += length;
+    return true;
+}
--- a/framework/tlvf/tlvf.py
+++ b/framework/tlvf/tlvf.py
@@ -714,8 +714,9 @@ class TlvF:
                             param_val_const = param_type + "::" + param_val_const
                         lines_cpp.append("if (!m_%s__) *m_%s = %s;" % ( self.MEMBER_PARSE, param_name, param_val_const) )
                     elif param_val != None: lines_cpp.append("if (!m_%s__) *m_%s = %s;" % ( self.MEMBER_PARSE, param_name, param_val) )
-                    elif param_length_var:  lines_cpp.append("if (!m_%s__) *m_%s = 0;" % ( self.MEMBER_PARSE, param_name) )
-                    lines_cpp.append( "m_%s__ += sizeof(%s) * 1;" % ( self.MEMBER_BUFF_PTR, param_type) )
+                    elif param_length_var: lines_cpp.append("if (!m_%s__) *m_%s = 0;" % ( self.MEMBER_PARSE, param_name) )
+                    lines_cpp.append("%sif (!buffPtrIncrementSafe(sizeof(%s))) { return false; }" % (self.getIndentation(0), param_type) )
+
                     if obj_meta.is_tlv_class and param_name != MetaData.TLV_TYPE_TYPE and param_name != MetaData.TLV_TYPE_LENGTH:
                         lines_cpp.append( "if(m_length && !m_%s__){ (*m_length) += sizeof(%s); }" % ( self.MEMBER_PARSE, param_type) )
 
@@ -811,7 +812,7 @@ class TlvF:
                     lines_cpp.append("%s}" %(self.getIndentation(1)))
                 else:
                     lines_cpp.append("%sm_%s_idx__ = len/sizeof(%s);" % (self.getIndentation(1), param_name, param_type))
-                    lines_cpp.append("%sm_%s__ += len;" % (self.getIndentation(1), self.MEMBER_BUFF_PTR))
+                    lines_cpp.append("%sif (!buffPtrIncrementSafe(len)) { return false; }" % (self.getIndentation(1)))
                 lines_cpp.append("}")
             if is_var_len or is_dynamic_len:
                 param_meta.list_index = obj_meta.list_index
@@ -840,9 +841,9 @@ class TlvF:
                     lines_cpp.append("}")
                 else:
                     lines_cpp.append("m_%s_idx__ = %s;" % (param_name, param_length))
-                    lines_cpp.append("m_%s__ += sizeof(%s)*(%s);" % (self.MEMBER_BUFF_PTR, param_type, param_length))
+                    lines_cpp.append("%sif (!buffPtrIncrementSafe(sizeof(%s)*(%s))) { return false; }" % (self.getIndentation(0), param_type,param_length) )
             if is_int_len or is_const_len:
-                lines_cpp.append( "m_%s__ += (sizeof(%s) * %s);" % (self.MEMBER_BUFF_PTR, param_type, param_length) )
+                lines_cpp.append("%sif (!buffPtrIncrementSafe(sizeof(%s)*(%s))) { return false; }" % (self.getIndentation(0), param_type,param_length) )
                 lines_cpp.append("m_%s_idx__  = %s;" % (param_name, param_length))
                 if obj_meta.is_tlv_class or TypeInfo(param_type).type == TypeInfo.STRUCT:
                     lines_cpp.append("if (!m_parse__) {")
@@ -1078,7 +1079,7 @@ class TlvF:
                 lines_cpp.append( "%sm_%s_vector.push_back(ptr);" % (self.getIndentation(1), param_name ))
             else:
                 lines_cpp.append( "%sm_%s_ptr = ptr;" % (self.getIndentation(1), param_name ))
-            lines_cpp.append( "%sm_%s__ += len;" % (self.getIndentation(1), self.MEMBER_BUFF_PTR) )
+            lines_cpp.append("%sif (!buffPtrIncrementSafe(len)) { return false; }" % (self.getIndentation(1)) )
             if obj_meta.is_tlv_class:
                 lines_cpp.append( "%sif(!m_parse__ && m_length){ (*m_length) += len; }" % (self.getIndentation(1)))
             lines_cpp.append( "%sm_%s__ = false;" % (self.getIndentation(1), self.MEMBER_LOCK_ALLOCATION))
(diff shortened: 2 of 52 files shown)


framework/tlvf/tlvf.py
```diff
@@ -710,8 +710,9 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
                             param_val_const = param_type + "::" + param_val_const
                         lines_cpp.append("if (!m_%s__) *m_%s = %s;" % ( self.MEMBER_PARSE, param_name, param_val_const) )
                     elif param_val != None: lines_cpp.append("if (!m_%s__) *m_%s = %s;" % ( self.MEMBER_PARSE, param_name, param_val) )
-                    elif param_length_var:  lines_cpp.append("if (!m_%s__) *m_%s = 0;" % ( self.MEMBER_PARSE, param_name) )
-                    lines_cpp.append( "m_%s__ += sizeof(%s) * 1;" % ( self.MEMBER_BUFF_PTR, param_type) )
+                    elif param_length_var: lines_cpp.append("if (!m_%s__) *m_%s = 0;" % ( self.MEMBER_PARSE, param_name) )
+                    lines_cpp.append("%sif (!buffPtrIncrementSafe(sizeof(%s))) { return false; }" % (self.getIndentation(0), param_type) )
```
arnout, line 718: There's another increment on line 810. I'll admit that there, you can statically determine that it will not overflow because the offset is exactly equal to the remaining bytes, but still it's better IMO to do it consistently.

There's another increment on line 839 and on line 841, and those *are* essential I think. In fact, I think that that is the one that triggered the issue to begin with.

There are increments on line 1053 and 1081, but again those are not overflowing because of conditions that come before. There, however, I think it would make sense to use a condition on buffPtrIncrementSafe instead.

And finally, I think the condition on line 1434 can be removed, as it is essentially checking the same thing, only too late because the out-of-bounds accesses have already happened.

framework/tlvf/src/src/BaseClass.cpp
```diff
@@ -21,6 +21,23 @@ size_t BaseClass::getBuffRemainingBytes(void *start)
 {
     return (m_buff_len__ - (size_t)(((uint8_t *)(start ? start : m_buff_ptr__)) - ((uint8_t *)m_buff__)));
 }
+
 size_t BaseClass::getLen() { return (size_t)((uint8_t *)m_buff_ptr__ - (uint8_t *)m_buff__); }
 
 bool BaseClass::isInitialized() { return m_init_succeeded; }
+
+/**
+ * @brief Increments m_buff_ptr__ pointer while avoiding a segfault caused by accessing unreserved memory.
```
arnout:
```
 * @brief Increments m_buff_ptr__ pointer while avoiding out-of-bounds accesses.
```

arnout (changes requested):
Please add a line like

```
Closes: #234 
```

to the commit message.

I personally would have split up the addition of buffPtrIncrementSafe to BaseClass in a separate commit from the update of tlvf.py, and I would have put the AutoGenerated files of tlvf.py in the same commit as the update of tlvf.py, but that's just a nitpick and maybe others disagree.

There's a typo in the second commit message: safeer -> safer

morantr:
It seems odd to me that it is actually solving it. Since on every init function, we are initializing the TLV members to places on the buffer. Incrementing the buffer pointer clearly will not cause to segfault. Only accessing it will. But this will be prevented because at the end of the init function there is the following check which will fail the init before letting any members access:
```cpp
if (m_buff_ptr__ - m_buff__ > ssize_t(m_buff_len__)) {
        TLVF_LOG(ERROR) << "Not enough available space on buffer. Class init failed";	        
        return false;
}
```
Now that I look at the code, I don't understand why sWscAttrMessageType is not checked to validate if it is m1 or m2.

Please show me where exactly the code crash.

tomereli:
> It seems odd to me that it is actually solving it. Since on every init function, we are initializing the TLV members to places on the buffer. Incrementing the buffer pointer clearly will not cause to segfault. Only accessing it will. But this will be prevented because at the end of the init function there is the following check which will fail the init before letting any members access:
> 
> ```c++
> if (m_buff_ptr__ - m_buff__ > ssize_t(m_buff_len__)) {
>         TLVF_LOG(ERROR) << "Not enough available space on buffer. Class init failed";	        
>         return false;
> }
> ```
> 

The problem is with variable length lists. When we have a variable-length list, on init() we are dereferencing `m_<list>_length`, so if `m_<list>_length` points to invalid address as with the case where we parse M1 where in fact it is M2, this causes out of bounds access.

Indeed - the last check is now irrelevent with Gal's change - @Galvien please remove that.

> Now that I look at the code, I don't understand why sWscAttrMessageType is not checked to validate if it is m1 or m2.
> 
> Please show me where exactly the code crash.

Because the `sWscAttrMessageType` is a member of a TLV, and we don't support this in TLVF. Adding support for these kinds of TLV inside TLV structures will be done as part of TLVF V2 development.

morantr:
> > It seems odd to me that it is actually solving it. Since on every init function, we are initializing the TLV members to places on the buffer. Incrementing the buffer pointer clearly will not cause to segfault. Only accessing it will. But this will be prevented because at the end of the init function there is the following check which will fail the init before letting any members access:
> > ```c++
> > if (m_buff_ptr__ - m_buff__ > ssize_t(m_buff_len__)) {
> >         TLVF_LOG(ERROR) << "Not enough available space on buffer. Class init failed";	        
> >         return false;
> > }
> > ```
> 
> The problem is with variable length lists. When we have a variable-length list, on init() we are dereferencing `m_<list>_length`, so if `m_<list>_length` points to invalid address as with the case where we parse M1 where in fact it is M2, this causes out of bounds access.
> 
> Indeed - the last check is now irrelevent with Gal's change - @Galvien please remove that.
> 

IMO this solution is clearly an overkill solution. Doing this check on every pointer increment is bad idea because it is not needed in most cases. Why not doing it only when we actually dereference the pointer?

> > Now that I look at the code, I don't understand why sWscAttrMessageType is not checked to validate if it is m1 or m2.
> > Please show me where exactly the code crash.
> 
> Because the `sWscAttrMessageType` is a member of a TLV, and we don't support this in TLVF. Adding support for these kinds of TLV inside TLV structures will be done as part of TLVF V2 development.

I don't understand what we don't support. It's a member of M1/M2 which we use. So what is the problem?
I think that maybe we can decide that if TLV (class) member is defined as const value, then on the init function on parsing, comparing the value in the buffer to that const (will happen on tlv types for sure). That way if we do addClass X on class Y, we will fail instead of crash.
So in the case of TLV_WSC we can define sWscAttrMessageType members in the TLV_WSC class itself instead of a different struct on the class, and since the data member (which hold eMwssageType) will be defined const with different unique values for M1 and M2, addClass will fail if adding the wrong TLV. 

tlvWscM2.yaml:
```cpp
tlvWscM2:
  _type: class
  _is_tlv_class : True
  type:
    _type: eTlvType
    _value_const: TLV_WSC
  length: uint16_t
  version_attr: WSC::sWscAttrVersion
  /* replace this:
  message_type_attr: WSC::sWscAttrMessageType
  */
  // with this:
  sWscAttrMessageType_attribute_type:
    _type: eWscAttributes
    _value: ATTR_MSG_TYPE
  sWscAttrMessageType_data_length:
    _type: uint16_t
    _value: 1
  sWscAttrMessageType_data:
    _type: eWscMessageType
    _value_const: WSC_MSG_TYPE_M2

  // rest of the yaml
```
tlvWscM2.cpp:
```cpp
bool tlvWscM2::init()
{
    if (getBuffRemainingBytes() < kMinimumLength) {
        TLVF_LOG(ERROR) << "Not enough available space on buffer. Class init failed";
        return false;
    }
    m_type = (eTlvType*)m_buff_ptr__;
    // replace this:
    // if (!m_parse__) *m_type = eTlvType::TLV_WSC;
    // with this: //
    if (m_parse__ && *m_type != eTlvType::TLV_WSC ) { // because m_type is const
        return false;
    } else {
        *m_type = eTlvType::TLV_WSC;
    }
    ///////////////////
    m_buff_ptr__ += sizeof(eTlvType) * 1;
    
    // init of sWscAttrVersion
    // same init as m_type, on sWscAttrMessageType_data because it is defined as const
    // rest of the code 
}
```

tomereli:
> > > It seems odd to me that it is actually solving it. Since on every init function, we are initializing the TLV members to places on the buffer. Incrementing the buffer pointer clearly will not cause to segfault. Only accessing it will. But this will be prevented because at the end of the init function there is the following check which will fail the init before letting any members access:
> > > ```c++
> > > if (m_buff_ptr__ - m_buff__ > ssize_t(m_buff_len__)) {
> > >         TLVF_LOG(ERROR) << "Not enough available space on buffer. Class init failed";	        
> > >         return false;
> > > }
> > > ```
> > 
> > 
> > The problem is with variable length lists. When we have a variable-length list, on init() we are dereferencing `m_<list>_length`, so if `m_<list>_length` points to invalid address as with the case where we parse M1 where in fact it is M2, this causes out of bounds access.
> > Indeed - the last check is now irrelevent with Gal's change - @Galvien please remove that.
> 
> IMO this solution is clearly an overkill solution. Doing this check on every pointer increment is bad idea because it is not needed in most cases. Why not doing it only when we actually dereference the pointer?

This was the solution agreed on by me, Gal and Arnout also after alignment with Lior. Why do you think that it is an overkill? I remind you that this is autogenerated code so checking each buffer increment is perfectly fine and not overkill in terms of performance. Also - this is the control path, not datapath, so we don't try to save CPU clocks. This approach is the root of the problem that got us to the current TLVF implementation to begin with - trying to optimize what doesn't need to be optimized.

Doing the check at every increment is the correct approach anyway IMOH since we want to catch this cases as soon as we can and not relay on the fact that there is some dereference pointer cause sometimes there isn't (no variable length lists for example).

What you're proposing is a dedicated solution only to the M1/M2 issue, which is simply not what this PR is aiming to solve. We are providing a defence mechanism for security breaches.

arnout:
> > Because the `sWscAttrMessageType` is a member of a TLV, and we don't support this in TLVF. Adding support for these kinds of TLV inside TLV structures will be done as part of TLVF V2 development.
> 
> I don't understand what we don't support. It's a member of M1/M2 which we use. So what is the problem?
> I think that maybe we can decide that if TLV (class) member is defined as const value, then on the init function on parsing, comparing the value in the buffer to that const (will happen on tlv types for sure). That way if we do addClass X on class Y, we will fail instead of crash.

That's exactly the missing feature I was talking about :-)

> So in the case of TLV_WSC we can define sWscAttrMessageType members in the TLV_WSC class itself

Actually, specifically for the WSC, we will have to solve it differently because the order of the attributes is not defined - the MessageType attribute is *not* necessarily the first one... So we need to treat the content of the WSC TLV as a list of attribute instead of as a struct like we do now. That's IIUC what Gal is working on at the moment. If we treat it as a list of attributes, there will be simply no way to distinguish the M1 from the M2 directly from the tlvf parsing.

tomereli (approved):

tomereli (approved):

tomereli:
Why isn't this merged?

arnout (approved):
The commented-out lines in tlvf.py should have been removed instead. Commented-out code is usually useless - git history should be used to find back how it was originally done.

So I've removed those commented out lines, and added a paragraph to the commit message to explain why that code can be removed.

PR #260, opened 2019-09-02
Feature/add client steering policy test flow

morantr:
Add client steering policy flow for controller certification test flow, and add a tester for it on test_flows.sh

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -485,9 +485,10 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         return handle_channel_preference_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE:
         return handle_channel_selection_request(sd, cmdu_rx);
-    case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE: {
+    case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE:
         return handle_client_capability_query(sd, cmdu_rx);
-    }
+    case ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE:
+        return handle_multi_ap_policy_config_request(sd, cmdu_rx);
     default:
         LOG(ERROR) << "Unknown CMDU message type: " << std::hex << int(cmdu_message_type);
         return false;
@@ -4779,6 +4780,23 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
     const auto mid = cmdu_rx.getMessageId();
     LOG(DEBUG) << "Received AP_CAPABILITY_QUERY_MESSAGE, mid=" << std::dec << int(mid);
     return true;
+}
+
+bool slave_thread::handle_multi_ap_policy_config_request(Socket *sd,
+                                                         ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    // TODO: This is a stub handler for the purpose of controller certification testing, will be
+    // implemented later on agent certification.
+    auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE, mid=" << std::hex << int(mid);
+
+    // send ACK_MESSAGE back to the controller
+    if (!cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE)) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+
+    return send_cmdu_to_controller(cmdu_tx);
 }
 
 bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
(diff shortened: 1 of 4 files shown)


arnout:
> Shouldn't we also check mid is correct?

That's a bigger issue, actually. => #280

arnout (approved):
The second commit was already done by @CoralMalachi so I removed that one.

The test_flows.sh required some updates due to changes from other PRs.

With that, LGTM.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -485,9 +485,10 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         return handle_channel_preference_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE:
         return handle_channel_selection_request(sd, cmdu_rx);
-    case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE: {
+    case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE:
         return handle_client_capability_query(sd, cmdu_rx);
-    }
+    case ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE:
+        return handle_multi_ap_policy_config_request(sd, cmdu_rx);
```
tomereli, line 491: why add "multi_ap" prefix? we don't do that for the rest of the APIs, so I guess we shouldn't here as well (nitpick)
```suggestion
        return handle_ap_policy_config_request(sd, cmdu_rx);
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4781,6 +4782,23 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
     return true;
 }
 
+bool slave_thread::handle_multi_ap_policy_config_request(Socket *sd,
```
tomereli, line 4785:
```suggestion
bool slave_thread::handle_ap_policy_config_request(Socket *sd,
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4781,6 +4782,23 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
...
+                                                         ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    // TODO: This is a stub handler for the purpose of controller certification testing, will be
+    // implemented later on agent certification.
+    auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE, mid=" << std::hex << int(mid);
+
+    // send ACK_MESSAGE back to the controller
+    if (!cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE)) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
```
tomereli, line 4795: Add MID for the error log so we know which ACK was lost.

tools/docker/tests/test_flows.sh
```diff
@@ -127,8 +127,15 @@ test_client_steering_mandate() {
 }
 
 test_client_steering_policy() {
-    #TODO: Implement
-    return 1
+    status "test client steering policy"
+    eval send_bml_command '"bml_wfa_ca_controller \"DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8003,tlv_type,0x89,tlv_length\
```
tomereli, line 134: A comment describing the fields used in this command can be helpful

tools/docker/tests/test_flows.sh
```diff
@@ -127,8 +127,15 @@ test_client_steering_mandate() {
...
+,0x000C,tlv_value,{0x00 0x00 0x01 {0x112233445566 0x01 0xFF 0x14}}\""' $redirect
+    sleep 1
+    dbg "Confirming client steering policy has been received on agent"
+    docker exec -it repeater1 sh -c 'grep -i -q "MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
```
tomereli: As mentioned in #280 - we should parse the MID here so we can match for it in the gateway log.

tools/docker/tests/test_flows.sh
```diff
@@ -127,8 +127,15 @@ test_client_steering_mandate() {
...
+    sleep 1
+    dbg "Confirming client steering policy ack message has been received on the controller"
+    docker exec -it gateway sh -c 'grep -i -q "ACK_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_controller.log'
```
tomereli: Should match the MID of the ACK message to one we sent.

tomereli (changes requested):

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: "MULTI_AP" prefix is unique to this specific message. We ignore "_MESSAGE" suffix which is common to all message types.
I think it is helpful to have this message since it is easier to find that function when searching for this message type string.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: see answer above

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: The MID is being printed one line before when we receive the policy config request. I don't see the point to print it again.

tomereli (approved):

PR #261, opened 2019-09-04
tools:docker:add option to specify agents number

CoralMalachi:
Add a command-line option that lets the user specify the number of
agents. for example, the following command will create two repeaters and
one default gateway:

	./test_gw_repeater.sh -r "repeater1" -r "repeater2"

If any repeater name is not specified the script creates one default
repeater.

### DIFF
--- a/tools/docker/tests/test_flows.sh
+++ b/tools/docker/tests/test_flows.sh
@@ -18,14 +18,6 @@ topdir="${scriptdir%/*/*/*/*}"
 redirect="> /dev/null 2>&1"
 error=0
 
-start_gw_repeater() {
-    dbg "delete running containers before starting test"
-    eval docker rm -f gateway $redirect
-    eval docker rm -f repeater $redirect
-    dbg "start gw and repeater"
-    eval ${scriptdir}/test_gw_repeater.sh -d 5 $redirect
-}
-
 send_bml_command() {       
     docker exec -it gateway ${installdir}/bin/beerocks_cli -c "$*"
 }
@@ -34,17 +26,17 @@ test_initial_ap_config() {
     status "test initial autoconfig"
 
     check_error=0
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "WSC Global authentication success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "WSC Global authentication success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "KWA (Key Wrap Auth) success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "KWA (Key Wrap Auth) success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "Controller configuration (WSC M2 Encrypted Settings)" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "Controller configuration (WSC M2 Encrypted Settings)" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
 
     return $check_error
@@ -65,28 +57,28 @@ test_channel_selection() {
     
     check_error=0
     dbg "Send channel preference query"
-    eval send_bml_command "bml_wfa_ca_controller \"DEV_SEND_1905,DestALid,$AL_MAC,MessageTypeValue,0x8004\"" $redirect
+    eval send_bml_command "bml_wfa_ca_controller \"DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8004\"" $redirect
     sleep 1
     dbg "Confirming channel preference query has been received on agent"
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "CHANNEL_PREFERENCE_QUERY_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "CHANNEL_PREFERENCE_QUERY_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
     
     dbg "Send channel selection request"
-    eval send_bml_command "bml_wfa_ca_controller \"DEV_SEND_1905,DestALid,$AL_MAC,MessageTypeValue,0x8006\"" $redirect
+    eval send_bml_command "bml_wfa_ca_controller \"DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8006\"" $redirect
     sleep 1
     dbg "Confirming channel selection request has been received on agent"
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "CHANNEL_SELECTION_REQUEST_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "CHANNEL_SELECTION_REQUEST_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
     
     dbg "Confirming 1905.1 Ack Message request was received on agent"
     # TODO: When creating handler for the ACK message on the agent, replace lookup of this string
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "handle_cmdu_control_ieee1905_1_message 8000" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec -it repeater sh -c \
+    check docker exec -it repeater1 sh -c \
         'grep -i -q "handle_cmdu_control_ieee1905_1_message 8000" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
 
     return $check_error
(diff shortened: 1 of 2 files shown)


tomereli:
> Add a command-line option that lets the user specify the number of
> agents. for example, the following command will create two repeaters and
> one default gateway:
> 
> ```
> `./test_gw_repeater.sh -r "repeater1" -r "repeater2"`
> ```
> 
> If any repeater name is not specified the script creates one default
> repeater.
> 
> Signed-off-by: Coral Malachi [coral.malachi@intel.com](mailto:coral.malachi@intel.com)

No need for quotation marks:
```
./test_gw_repeater.sh -r repeater1 -r repeater2
```

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
     return $error
 }
 
 VERBOSE=false
 REMOVE=false
-GW_NAME=gateway
-REPEATER_NAME=repeater
 START_GATEWAY=true
 START_REPEATER=true
+REPEATER_NAME=repeater
```
tomereli: Why do you need to define the default repeater name here? please remove and use directly above, since you do the same for the gatway name

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -51,12 +51,16 @@ main() {
 
     status "Starting GW+Repeater test"
 
+    # default values for gateway and repeater[s] names
+    REPEATER_NAMES=${REPEATER_NAMES-${REPEATER_NAME}}
```
tomereli:
```suggestion
    REPEATER_NAMES=${REPEATER_NAMES-repeater}
```

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -51,12 +51,16 @@ main() {
...
+    GW_NAME=${GW_NAME-gateway}
+
     dbg REMOVE=$REMOVE
     dbg GW_NAME=$GW_NAME
-    dbg REPEATER_NAME=$REPEATER_NAME
+    dbg REPEATER_NAMES=$REPEATER_NAMES
     dbg START_GATEWAY=$START_GATEWAY
     dbg START_REPEATER=$START_REPEATER
-    dbg DELAY=$DELAY
+    dbg DELAY=$DELAY
```
tomereli: extra whitespaces

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
-        status "Start Repeater (Remote Agent)"
-        ${scriptdir}/../run.sh ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${REPEATER_NAME} -m aa:bb:cc:dd "$@"
-    }
+    }   
+
+    INDEX=0
+    for r in $REPEATER_NAMES
+    do
+        
+        [ "$START_REPEATER" = "true" ] && {
```
tomereli: This condition should be outside the loop

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
+            status "Start Repeater (Remote Agent)"
+            ${scriptdir}/../run.sh ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${r} -m aa:bb:cc:$INDEX$INDEX "$@"
+        }        
+        echo $r
+        INDEX=$((INDEX+1))
+        status "Delay ${DELAY} seconds..."
```
tomereli: The delay is needed only between starting the gateway and the repeaters, not for each repeater. It is used for making sure the GW's local agents start first, so I would remove this delay.

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
+        sleep ${DELAY}
+    done
 
-    status "Delay ${DELAY} seconds..."
-    sleep ${DELAY}
+    # exit 0
```
tomereli: Remove comment

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
     error=0
     [ "$START_GATEWAY" = "true" ] && report "GW operational" \
-        ${scriptdir}/../test.sh ${VERBOSE_OPT} -n ${GW_NAME}
+        ${scriptdir}/../test.sh ${VERBOSE_OPT} -n ${GW_NAME}   
     
-    [ "$START_REPEATER" = "true" ] && report "Repeater operational" \
-        ${scriptdir}/../test.sh ${VERBOSE_OPT} -n ${REPEATER_NAME}
+    for r in $REPEATER_NAMES
+    do
+        [ "$START_REPEATER" = "true" ] && {
```
tomereli: Testing START_REPEATER should be outside the for loop.

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
+        report "GW operational" \
```
tomereli: Typo `GW operational` --> `Repeater $r operational`

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
     [ "$REMOVE" = "true" ] && {
-        status "Deleting containers ${GW_NAME}, ${REPEATER_NAME}"
-        docker rm -f ${GW_NAME} ${REPEATER_NAME} >/dev/null 2>&1
+        status "Deleting gateway container ${GW_NAME}"
+        docker rm -f ${GW_NAME} >/dev/null 2>&1
     }
-
+    for r in $REPEATER_NAMES
+    do
+        [ "$REMOVE" = "true" ] && {
```
tomereli: condition should be outside the for loop

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -66,37 +70,55 @@ main() {
...
+        status "Deleting repeaters containers ${r}"
```
tomereli:
```suggestion
        status "Deleting repeater ${r} container"
```

tomereli (changes requested):

tomereli:
LGTM

tools/docker/tests/test_flows.sh
```diff
@@ -138,23 +130,32 @@ test_higher_layer_data_payload() {
...
 
-    eval ${scriptdir}/test_gw_repeater.sh -f -d 5 $redirect || {
+    eval ${scriptdir}/test_gw_repeater.sh -f -r "repeater1" -r "repeater2" -d 5 $redirect || {
         err "start GW+Repeater failed, abort"
         exit 1
     }
-    AL_MAC=$(docker exec -it gateway ${installdir}/bin/beerocks_cli -c bml_conn_map | grep IRE_BRIDGE | awk '{print $5}' | cut -d ',' -f 1)
-    RADIO_WLAN0_MAC=$(docker exec -it gateway ${installdir}/bin/beerocks_cli -c bml_conn_map | grep "RADIO: wlan0" | head -1 | awk '{print $4}' | cut -d ',' -f 1 | tr --delete :)    
-    RADIO_WLAN0_MAC="0x${RADIO_WLAN0_MAC}"     
+    AL_MAC_agent1=$(docker exec -it gateway ${installdir}/bin/beerocks_cli -c bml_conn_map | grep IRE_BRIDGE | head -1 | awk '{print $5}' | cut -d ',' -f 1)
```
arnout: Since there are so many addresses to retrieve, it would be better to do the bml_conn_map once, dump it in a file, and then parse it to different variables.

Also, the grep/head/awk can be done in a single simple awk script.

Both comments are nitpicks. I'm planning to do some improvements of these scripts anyway, so I can do it then.

tools/docker/tests/test_flows.sh
```diff
@@ -138,23 +130,32 @@ test_higher_layer_data_payload() {
...
+    echo "AL_MAC_agent1 = ${AL_MAC_agent1}"
```
arnout: This should be a dbg I think.

tools/docker/tests/test_flows.sh
```diff
@@ -210,8 +211,11 @@ main() {
 }
 
 VERBOSE=false
-AL_MAC=
-RADIO_WLAN0_MAC=
-RADIO_WLAN2_MAC=
+AL_MAC_agent1=
```
arnout: These don't need to be initialised. We only initialise default values for variables that can be overridden by options, but these variables are internal variables that will be set later.

tools/docker/tests/test_flows.sh
```diff
@@ -210,8 +211,11 @@ main() {
...
+AL_MAC_agent2=
+RADIO_WLAN0_MAC_agent1=
+RADIO_WLAN2_MAC_agent1=
+RADIO_WLAN0_MAC_agent2=
+RADIO_WLAN2_MAC_agent2=
```
arnout: Nitpick: I don't like the naming convention too much. I definitely don't like the mixing of uppercase with lowercase. Since you're changing all these names in this commit anyway, it's a good time to fix them :-)

I propose to keep them lowercase, and like this: mac_agent1, mac_agent1_wlan0

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -40,7 +40,7 @@ main() {
             -d | --delay)         DELAY="$2"; shift; shift ;;
             -f | --force)         FORCE_OPT="-f"; shift ;;
             -g | --gateway)       GW_NAME="$2"; shift; shift ;;
-            -r | --repeater)      REPEATER_NAME="$2"; shift; shift ;;
+            -r | --repeater)      REPEATER_NAMES="$2 $REPEATER_NAMES"; shift; shift ;;
```
arnout: If you set this to 
```suggestion
            -r | --repeater)      REPEATER_NAMES="$REPEATER_NAMES $2"; shift; shift ;;
```
you don't need to reverse them.

arnout (changes requested):
Is there any reason to keep this PR as draft?

tools/docker/tests/test_flows.sh
CoralMalachi: OK :)

tools/docker/tests/test_flows.sh
CoralMalachi: Oops, I thought I have removed the printing, I'll fix that now.

tools/docker/tests/test_flows.sh
CoralMalachi: I agree with you, I'll try this now.

arnout (approved):
There were still a large number of small things to fix in this PR... I've just made the required changes, rebased and pushed again, so it's now ready to merge.

There were many whitespace errors. Please configure your editor to clean up whitespace when saving. Also, when you do `git diff` (which you should always do before committing), whitespace errors will be marked in red.

tomereli (approved):

PR #262, opened 2019-09-04
tools:docker: fix condition check

CoralMalachi:
The current bug is that when we check if a specific container is already
running, in some cases it returns the wrong answer.

For example, if "repeater1" is already running, the condition returns
true for "repeater".

For that reason, the condition check was changed so now it returns the
right answer.

### DIFF
--- a/tools/docker/run.sh
+++ b/tools/docker/run.sh
@@ -105,8 +105,7 @@ main() {
     else
         DOCKEROPTS="$DOCKEROPTS -d"
     fi
-
-    if [ -n "$(docker ps -q -l -f name="${NAME}")" ]; then
+    if [ -n "$(docker ps -f name="${NAME}" | grep -w "${NAME}")" ]; then
         info "Container ${NAME} is already running"
         if [ "$FORCE" = "true" ]; then
             run docker container rm -f "$NAME"


tools/docker/run.sh
```diff
@@ -106,7 +106,9 @@ main() {
         DOCKEROPTS="$DOCKEROPTS -d"
     fi
 
-    if [ -n "$(docker ps -q -l -f name="${NAME}")" ]; then
```
tomereli, line 109:
```
# docker ps -f name=foo returns all containers whose name contains foo, so make sure that the exact named container exists
if [ -n "$(docker ps -f name="${NAME}") | grep -w "${NAME}"" ]; then

```

tomereli (changes requested):

tools/docker/run.sh
```diff
@@ -105,8 +105,8 @@ main() {
     else
         DOCKEROPTS="$DOCKEROPTS -d"
     fi
-
-    if [ -n "$(docker ps -q -l -f name="${NAME}")" ]; then
+
```
tomereli: extra whitespace at start of line

tomereli (changes requested):

tomereli (approved):
please rebase + autosquash before merging

PR #263, opened 2019-09-04
Add a test for filling M2 from the database

arnout:
To be able to do this test, we need the autoconfig renew message to be handled in the agent, so add that as well.

On the way, some other small issues were found in the agent. These are fixed in separate commits.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -3702,6 +3702,7 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             return false;
         }
 
+        // TODO Move APRadioBasicCapabilities creation to a separate function
         auto radio_basic_caps = cmdu_tx.addClass<wfa_map::tlvApRadioBasicCapabilities>();
         if (!radio_basic_caps) {
             LOG(ERROR) << "Error creating TLV_AP_RADIO_BASIC_CAPABILITIES";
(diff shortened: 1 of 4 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3702,44 +3702,45 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             return false;
         }
 
+        // TODO Move APRadioBasicCapabilities creation to a separate function
```
tomereli: Can we do that as part of this PR? It can even be a separate commit which moves all the big logic out of the case. At the first step I'll even be happy with a single function `start_autoconfiguration` and move all the logic there.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3702,44 +3702,45 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
...
+        }
+
+        // TODO: Currently sending dummy values, just one operating class based on the band. Need
+        // to read them from DWPAL and use the correct WiFi Parameters based on the regulatory
+        // domain
+        uint8_t operating_class                            = hostap_params.iface_is_5ghz ? 110 : 81;
+        operationClassesInfo->operating_class()            = operating_class;
+        operationClassesInfo->maximum_transmit_power_dbm() = 0;
+
+        // TODO - the number of statically non operable channels can be 0 - meaning it is
```
tomereli: Issue #8 is closed - so this TODO comment can be removed - TLVF has support for variable-length lists with zero-length.
IMO the only reason to keep a non-operable channel is to test the controller code which sets the supported channels based on the operating class, but it has been tested so I wouldn't bother.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4510,19 +4520,23 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
         sMacAddr bssid;
         WSC::eWscAuth authtype;
         WSC::eWscEncr enctype;
+        bool fronthaul;
```
tomereli: (nitpick) I prefer to keep multiple variables definitions of the same type in the same line:
```
bool fronthaul, backhaul, teardown;
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4415,6 +4415,20 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
 
     // Swap back to host byte order to read and use config_data
     config_data.class_swap();
+
+    // TODO tlvf should check this
+    if (config_data.ssid_length() > WSC::WSC_MAX_SSID_LENGTH) {
+        LOG(INFO) << "SSID too long: " << config_data.ssid_length();
+        return false;
+    }
+    // TODO tlvf should do conversion to std::string
```
tomereli, line 4391: See #265 - you can take over if needed or we can merge it and you can rebase.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4415,6 +4415,20 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
 
     // Swap back to host byte order to read and use config_data
     config_data.class_swap();
+
+    // TODO tlvf should check this
```
tomereli, line 4386: Correct - we should add a `_length_max` meta variable to TLVF and add this support. Can you open a task for it and refer to it in the comment?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4443,6 +4448,53 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
     return true;
 }
 
+/**
```
tomereli, line 4413: Copy paste bug

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4443,6 +4448,53 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
...
+ * @return false on failure
+ */
+bool slave_thread::handle_autoconfiguration_renew(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    LOG(INFO) << "received autoconfig renew message";
+
+    auto tlvAlMac = cmdu_rx.addClass<ieee1905_1::tlvAlMacAddressType>();
+    if (tlvAlMac) {
+        LOG(DEBUG) << "tlvAlMac=" << network_utils::mac_to_string(tlvAlMac->mac());
+        // TODO register/update mapping of AL-MAC to interface
```
tomereli: Do we have a task for this? if so, it would be useful to add the link here.

tomereli (changes requested):

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: > Can we do that as part of this PR? It can even be a separate commit which moves all the big logic out of the case.

I felt that this PR was already big enough as it is :-) Note that this comment was already there, I just moved it to the appropriate place.

>  At the first step I'll even be happy with a single function `start_autoconfiguration` and move all the logic there.

I'll stick with creating a function just for this tlv, similar to `autoconfig_wsc_add_m1()`

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: I'll see after #265 has been merged.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: NVM - I'm adding support for this as well in #265

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -5047,19 +5047,6 @@ bool slave_thread::add_radio_basic_capabilities()
     operationClassesInfo->operating_class()            = operating_class;
     operationClassesInfo->maximum_transmit_power_dbm() = 0;
```
tomereli: This should be a fixup commit, shouldn't it?

tomereli (approved):

PR #264, opened 2019-09-04
topology discovery minimal support

tomereli:
Add minimal topology discovery support in both the agent and controller.
The agent (backhaul manager thread) sends topology discovery after connecting to the controller, and after that every 60 seconds as defined in the 1905.1 specification section 8.2.1.1:
> A 1905.1 management entity shall transmit a topology discovery message
every 60 seconds or if an implementation-specific event occurs (e.g.,
device initialized or an interface is connected).m

The controller is updated to subscribe to the topology discovery message, and a handler is added to it which adds a node to the database in case it does not exist yet. This is done in order to save the socket descriptor of that agent in case autoconfiguration was not yet completed so the controller is unaware of it.
The need for adding the socket to the database comes from the controller certification for testcase 5.3.1 - topology query, where the CTT agent may or may not do autoconfiguration as part of the test  - it was seen that some reference agents do autoconfig and some do not.

Once the node have been added to the database, the controller can send messages to it when instructed to by the UCC application via 1905_send_cmdu CAPI command.

Tested with Broadcomm CTT agent on the certification testbed acting as the Marvell CTT agent (setting the Broadcomm agent in AllInitConfig_MAP.txt and MasterTestInfo.xml causes the test to fail for reasons we don't yet understand, but setting Agent1 to the IP Address of broadcomm works just fine and the test passes.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -26,6 +26,7 @@
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
+#include <tlvf/ieee_1905_1/tlvMacAddress.h>
 #include <tlvf/ieee_1905_1/tlvSearchedRole.h>
 #include <tlvf/ieee_1905_1/tlvSupportedFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvSupportedRole.h>
@@ -312,6 +313,17 @@ void main_thread::after_select(bool timeout)
             soc->sta_wlan_hal->process_int_events();
             clear_ready(soc->sta_hal_int_events);
         }
+    }
+
+    // Send topology discovery every 60 seconds according to IEEE_Std_1905.1-2013 specification
+    static std::chrono::steady_clock::time_point discovery_timestamp =
+        std::chrono::steady_clock::now();
+    auto now     = std::chrono::steady_clock::now();
+    auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(now - discovery_timestamp);
+    if ((m_eFSMState == EState::CONNECTED) ||
+        (m_eFSMState == EState::OPERATIONAL && elapsed.count() == 60)) {
+        discovery_timestamp = now;
+        send_1905_topology_discovery_message();
     }
 }
 
@@ -802,6 +814,30 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
     }
 
     return (true);
+}
+
+bool main_thread::send_1905_topology_discovery_message()
+{
+    auto cmdu_hdr = cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE);
+    if (!cmdu_hdr) {
+        LOG(ERROR) << "Failed to create TOPOLOGY_DISCOVERY_MESSAGE cmdu";
+        return false;
+    }
+    auto tlvAlMac = cmdu_tx.addClass<ieee1905_1::tlvAlMacAddressType>();
+    if (!tlvAlMac) {
+        LOG(ERROR) << "Failed to create tlvAlMacAddressType tlv";
+        return false;
+    }
+    tlvAlMac->mac() = network_utils::mac_from_string(bridge_info.mac);
+    auto tlvMac     = cmdu_tx.addClass<ieee1905_1::tlvMacAddress>();
+    if (!tlvMac) {
+        LOG(ERROR) << "Failed to create tlvMacAddress tlv";
+        return false;
+    }
+    tlvMac->mac() = network_utils::mac_from_string(bridge_info.mac);
+
+    LOG(DEBUG) << "send_1905_topology_discovery_message, bridge_mac=" << bridge_info.mac;
+    return send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
 }
 
 bool main_thread::send_autoconfig_search_message(std::shared_ptr<SSlaveSockets> soc)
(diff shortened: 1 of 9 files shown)


arnout (changes requested):
I don't think I agree with the architectural direction in which this is going...

I think the proper way to handle things is to have:

- Topology manager that keeps track of the topology information. It handles most of the 1905.1 CMDUs and more importantly it sends them out. It also keeps track of local topology and notices when e.g. an interface changes.
- Controller that uses topology information in its decision making (it's really relevant mostly for backhaul optimisation, and maybe also for client steering to get away from a saturated backhaul link).
- Transport that uses topology information to know on which interface to send out unicast messages destined for a specific AL-MAC address.
- High-level API that exports topology information to interested users.

The micro-services architectures complicates this picture, because there are multiple microservices that need to know the same information. There are a couple of ways in which this can be solved:

1. The topology database is accessed over IPC. However, the pub/sub IPC mechanism that is currently used is not really appropriate for that.
2. A new IPC mechanism is added to access the topology database. This needs to be a transactional IPC to avoid deadlocking with the pub/sub IPC. The most obvious one would be an actual database, e.g. sqlite. Lots of other software does that.
3. Every microservice keeps its own copy of a (reduced) topology database and the topology sends updates to subscribers. This is more or less how the agent works now (backhaul manager and slaves).
4. Every microservice keeps its own copy of a (reduced) topology database and subscribes to the messages that update it. This is the approach taken by the third commit in this PR.

I would actually prefer approach 2 or 3, but 4 is an acceptable solution. I believe it's not ideal because it makes the microservices significantly more heavyweight, 

For solution 4, I think the handling of these messages should move to common code. This makes sure that when new messages are added, they are handled everywhere. Still, that is something that can be refactored later.

So in the end, the only big problem that I have with this PR is the second commit. It basically puts the functionality of the topology manager in the agent. AFAIU, this functionality already exists in the topology manager, but we just don't run that in our tests. Putting the topology manager functionality in the agent would make it very difficult to run a controller-only prplMesh.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -313,6 +314,17 @@ void main_thread::after_select(bool timeout)
...
         }
     }
+
+    // Send topology discovery every 60 seconds according to IEEE_Std_1905.1-2013 specification
+    static std::chrono::steady_clock::time_point discovery_timestamp =
+        std::chrono::steady_clock::now();
+    auto now     = std::chrono::steady_clock::now();
+    auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(now - discovery_timestamp);
+    if ((m_eFSMState == EState::CONNECTED) ||
+        (m_eFSMState == EState::OPERATIONAL && elapsed.count() == 60)) {
```
arnout, line 324: This should be `>= 60` I think.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -313,6 +314,17 @@ void main_thread::after_select(bool timeout)
             clear_ready(soc->sta_hal_int_events);
         }
     }
+
+    // Send topology discovery every 60 seconds according to IEEE_Std_1905.1-2013 specification
+    static std::chrono::steady_clock::time_point discovery_timestamp =
```
arnout, line 319: Instead of making this a static local variable I think it should be a member of the main_thread class.

Or, well, actually, I think socket_thread should be refactored to be a proper event loop that can handle timers as well as sockets. But then again, I think all processes should become single threaded... So that's leading us a bit too far :-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -313,6 +314,17 @@ void main_thread::after_select(bool timeout)
...
+        std::chrono::steady_clock::now();
+    auto now     = std::chrono::steady_clock::now();
+    auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(now - discovery_timestamp);
+    if ((m_eFSMState == EState::CONNECTED) ||
+        (m_eFSMState == EState::OPERATIONAL && elapsed.count() == 60)) {
+        discovery_timestamp = now;
```
arnout, line 325: As usual with repeating timers, the question is: should we accumulate the jitter (like you do here) or should we approximate the rate as much as possible (by doing `discovery_timestamp += 60` instead). But in the latter case, you also have to make sure that you handle multiple deadline misses (e.g. with `if (now - discovery_timestamp >= 60) discovery_timestamp = now`). But again, all of that would lead us a bit too far. Maybe add a TODO.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -804,6 +816,30 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
...
+    if (!cmdu_hdr) {
+        LOG(ERROR) << "Failed to create TOPOLOGY_DISCOVERY_MESSAGE cmdu";
+        return false;
+    }
+    auto tlvAlMac = cmdu_tx.addClass<ieee1905_1::tlvAlMacAddressType>();
+    if (!tlvAlMac) {
+        LOG(ERROR) << "Failed to create tlvAlMacAddressType tlv";
+        return false;
+    }
+    tlvAlMac->mac() = network_utils::mac_from_string(bridge_info.mac);
```
arnout, line 831: Could you refactor this with the exact same code that is in send_autoconfig_search_message?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -804,6 +816,30 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
...
+    auto tlvMac     = cmdu_tx.addClass<ieee1905_1::tlvMacAddress>();
+    if (!tlvMac) {
+        LOG(ERROR) << "Failed to create tlvMacAddress tlv";
+        return false;
+    }
+    tlvMac->mac() = network_utils::mac_from_string(bridge_info.mac);
```
arnout, line 837: Meh. This is wrong. The MAC should be the MAC address of the interface on which you send the message: "MAC address of the interface on which the message is transmitted".

Also it's a neighbour multicast message, which I don't think the transport supports at the moment.

Oh well, minimal implementation I guess...

arnout (approved):
I don't like this PR at all, I guess I made that very clear :-)

I really would prefer to revive the topology manager.

However, it gets the job done, so let's merge it as is...

PR #265, opened 2019-09-05
tlvf: add std::string getter function for CHAR lists

tomereli:
Add an std::string <param_nam>_str() getter function for variable length
lists of CHAR type which returns a string, thus saving the conversion
from being done by the user.


framework/tlvf/tlvf.py
```diff
@@ -903,11 +903,17 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
                 else:
                     lines_h.append("size_t %s_length() { return m_%s_idx__ * sizeof(%s); }" % (param_name, param_name, param_type_full))
 
-            #add function to get char pointer
+            #add function to get std::string
```
arnout: You add both functions, so this comment is incorrect.

framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
```diff
@@ -473,6 +473,10 @@ uint32_t& cInner::var1() {
     return (uint32_t&)(*m_var1);
 }
 
+std::string cInner::unknown_length_list_inner_str() {
+    return std::string(m_unknown_length_list_inner, m_unknown_length_list_inner_idx__);
```
arnout: I believe that if the length is zero, m_unknown_length_list_inner will be nullptr. std::string constructor segfaults on that.

Would it be possible to add a tlvf_test to check this?

The fix is simply to check the length and return `std::string()` if it's empty.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4513,11 +4513,8 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
                                                         authtype, enctype))
             return false;
 
-        std::string manufacturer =
-            std::string(tlvWscM2->manufacturer(), tlvWscM2->manufacturer_length());
-
         LOG(DEBUG) << "Controller configuration (WSC M2 Encrypted Settings)" << std::endl
-                   << "     Manufacturer: " << manufacturer << std::endl
+                   << "     Manufacturer: " << tlvWscM2->manufacturer_str() << std::endl
```
arnout: It's really not nice to mix this change together with the big Autogenerated stuff in a single commit. i.e., please make a separate commit of this.

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -305,18 +305,6 @@ sWscAttrKeyWrapAuthenticator:
...
-  _type: struct
-  attribute_type:
-    _type: eWscAttributes
-    _value: ATTR_SSID
-  data_length:
-    _type: uint16_t
-    _value: WSC_MAX_SSID_LENGTH
-  data:
-    _type: char
-    _length: [ WSC_MAX_SSID_LENGTH ]
```
arnout, line 318: Actually, I think it would be better to change cConfigData to make use of this attribute. The attribute definition is wrong, though, length should be either `[]` or `[ data_length ]`. Oh, and that means it would have to become a class instead of a struct...

Note also that a lot of other attributes suffer from the same shortcoming (they're declared as fixed length instead of dynamic length).

BTW the commit message is wrong, it talks about M1 and M2 but it's actually used in cConfigData.

Obviously, this requires the callers to change as well, and *that* must be done in the same commit as the yaml change.

arnout (changes requested):

framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
tomereli: Fixed, will update the test in tlvf in a separate commit after rebasing

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: ok, but this requires force pushing. So I will also rebase in the process.

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
tomereli: You are right but unfortunately, if my memory serves me right, we don't support variable length lists inside structs inside classes, I can't find an issue for some reason, we must have closed it.
This is also the reason that all the string attributes (such as manufacturer, device name, etc.) are put as is in the M2 and M1 TLVs and not using the attributes.
This should be fixed obviousely since it make the YAML much longer - but not a high priority.

framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
tomereli: Done

arnout (approved):

Issue #220, opened 2019-08-22
5.8.3 Client Association Control Message


c0dingn0mad:
## Description

The prplMesh controller needs to pass the test case as defined in 5.8.3 

## Exit criteria

The flow is completed once the controller passes this test in the scenario and testbed described in the upstream definition.

- [x] #218 
- [x] #219

PR #270, opened 2019-09-08
Feature/add client association control message flow

CoralMalachi:
According to section 5.8.3 in the EasyMesh Test Plan v1.0 [1],
Client Association Control Message test flow.

#218 
#220

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -479,6 +479,8 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         return handle_autoconfiguration_renew(sd, cmdu_rx);
     case ieee1905_1::eMessageType::AP_CAPABILITY_QUERY_MESSAGE:
         return handle_ap_capability_query(sd, cmdu_rx);
+    case ieee1905_1::eMessageType::CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE:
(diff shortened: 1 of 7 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4780,6 +4782,22 @@ bool slave_thread::handle_ap_capability_query(Socket *sd, ieee1905_1::CmduMessag
...
 }
 
+bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    //TODO - this is a stub handler for the purpose of controller certification testing,
+    //       will be implemented later on agent certification
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE, mid=" << std::dec
+               << int(mid);
+
```
arnout, line 4791: Same comment as in #271: ideally, you should parse out the Client Association Control Request TLV, match the BSSID, and only ack if it corresponds to this slave. Still: nitpick.

arnout (approved):
In general, the same comments as in #271 apply - including the compliments.

Ideally, the ack response should be sent to the sender instead of the controller. However, in the slave, we no longer have information of the sender. Also, the sender in that case *must* be the controller (in fact, the backhaul manager should filter out any request which doesn't come from the controller). So it's OK as it is.

Issue #166, opened 2019-08-14
[TASK] Implement Higher layer data payload over 1905 trigger test flow


Galvien:
## Description

Add a test in `test_flows.sh` to trigger the controller to send a Higher Layer Data message to the agent as described in section 5.10.1 of the _EasyMesh Test Plan v1.0_.
Then, the script will check to assure that the message has been received in the respective agent.
As stated in the test plan:

> This test verifies that a MCUT includes appropriate TLVs and field values in a Higher Layer Data message

## Exit criteria

The task will be complete when the test script has been properly written, and necessary components have been implemented, running the `layer_data_payload_trigger` test in `test_flows.sh` returns an OK, and `.travis.yml` has been updated accordingly.

## Detailed description
First the test script needs to be updated to command the controller to send the message to the agent, and then check the agent's logs for the entry stating the message was handled, the respective message needs to be configured in tlvf.
The next step is to update the controller to support the new message, and to update bml and the CLI to support sending the message.
Finally, the agent backhaul manager needs to subscribe to and handle the received message,
And `.travis.yml` needs to be updated to run the respective test.

PR #271, opened 2019-09-08
Feature/add higher layer data payload flow

CoralMalachi:
According to section 5.10.1 in the EasyMesh Test Plan v1.0 [1],
Higher layer data payload over 1905 trigger test flow.

Therefore, add a test in test_flows.sh to trigger the controller to
send a Higher Layer Data message to the agent and make sure 
that the message has been received in the respective agent.

#166

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -30,6 +30,7 @@
 #include <tlvf/ieee_1905_1/tlvSearchedRole.h>
 #include <tlvf/ieee_1905_1/tlvSupportedFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvSupportedRole.h>
+#include <tlvf/wfa_map/tlvHigherLayerData.h>
 #include <tlvf/wfa_map/tlvSearchedService.h>
 #include <tlvf/wfa_map/tlvSupportedService.h>
 
@@ -131,6 +132,7 @@ bool main_thread::init()
             ieee1905_1::eMessageType::ACK_MESSAGE, ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE,
             ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE,
             ieee1905_1::eMessageType::CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE,
+            ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE,
             ieee1905_1::eMessageType::AP_CAPABILITY_QUERY_MESSAGE})) {
 
         LOG(ERROR) << "Failed to init mapf_bus";
@@ -1612,6 +1614,9 @@ bool main_thread::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
         return handle_1905_discovery_query(cmdu_rx);
         //LOG(INFO) << "I got the Topology Query message!";
     }
+    case ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE: {
+        return handle_1905_higher_layer_data_message(cmdu_rx, src_mac);
+    }
 
     default: {
         // TODO add a warning once all vendor specific flows are replaced with EasyMesh
@@ -1634,6 +1639,33 @@ bool main_thread::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cmdu_rx
     //TODO - this should be part of the discovery agent, will be done as part of
     //       agent certification
     return true;
+}
+
+bool main_thread::handle_1905_higher_layer_data_message(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received HIGHER_LAYER_DATA_MESSAGE , mid=" << std::hex << int(mid);
+
+    auto tlvHigherLayerData = cmdu_tx.addClass<wfa_map::tlvHigherLayerData>();
+    if (!tlvHigherLayerData) {
+        LOG(ERROR) << "addClass wfa_map::tlvHigherLayerData failed";
+        return false;
+    }
+
+    auto protocol     = tlvHigherLayerData->protocol();
+    const auto length = tlvHigherLayerData->length();
+    LOG(DEBUG) << "protocol: " << std::hex << protocol;
+    LOG(DEBUG) << "length: " << int(length);
+
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+    LOG(DEBUG) << "sending ACK message to the originator, mid=" << std::hex << int(mid);
+    return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
 }
 
 bool main_thread::handle_1905_autoconfiguration_response(ieee1905_1::CmduMessageRx &cmdu_rx,
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -963,7 +963,8 @@ bool master_thread::handle_cmdu_1905_channel_preference_report(Socket *sd,
 bool master_thread::handle_cmdu_1905_ack_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     auto mid = cmdu_rx.getMessageId();
-    LOG(INFO) << "Received ACK_MESSAGE, mid=" << std::hex << int(mid);
+    //TODO: the ACK should be sent to the correct task and will be done as part of agent certification
+    LOG(DEBUG) << "Received ACK_MESSAGE, mid=" << std::hex << int(mid);
     return true;
 }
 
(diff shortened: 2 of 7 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1599,6 +1603,21 @@ bool main_thread::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cmdu_rx
...
+    LOG(DEBUG) << "Received HIGHER_LAYER_DATA_MESSAGE , mid=" << std::hex << int(mid);
+
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+    LOG(DEBUG) << "sending ACK message back to controller";
+    return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
```
arnout: The message doesn't necessarily come from the controller, so you have to use the source address to identify where to send the ACK.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1599,6 +1603,21 @@ bool main_thread::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cmdu_rx
     return true;
 }
 
+bool main_thread::handle_1905_higher_layer_data_message(ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received HIGHER_LAYER_DATA_MESSAGE , mid=" << std::hex << int(mid);
```
arnout, line 1648: It would be nice to also parse out the higher layer data TLV and print a bit more information about it, e.g. the protocol and the length. But not very important. (nitpick)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -222,6 +222,8 @@ bool master_thread::handle_cmdu_1905_1_message(Socket *sd, ieee1905_1::CmduMessa
         return handle_cmdu_1905_channel_preference_report(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE:
         return handle_cmdu_1905_channel_selection_response(sd, cmdu_rx);
+    case ieee1905_1::eMessageType::ACK_MESSAGE:
+        return handle_cmdu_1905_ack_message(sd, cmdu_rx);
```
arnout: For now this is OK, but at some point we should really handle the acks (or lack thereof) and do retransmissions. => #273

tools/docker/tests/test_flows.sh
```diff
@@ -137,8 +137,26 @@ test_client_association() {
 }
 
 test_higher_layer_data_payload_trigger() {
-    #TODO: Implement
-    return 1
+    status "test higher layer data payload"
+    
+    mac_gateway_hex=`echo $mac_gateway | tr --delete :`
+    mac_gateway_hex="0x${mac_gateway_hex}"
```
arnout, line 159: We're doing this (converting from colon-MAC to 0x) in several places now, it would make sense to make a function (in functions.sh) for it. (nitpick)

arnout (changes requested):
First of all, let me commend you on the quality of this PR. It has all the features we like to see:

- Perfectly split into commits that do one thing only.
- Each commit has an extensive and correct commit message.
- The PR includes a test of the new feature.
- Everything passes CI.

There is one important fix, the destination of the ACK message. The rest of the comments are nitpicks.

Note also that this PR really isn't needed to pass 5.10.1: the only change in the controller is the handling of the ACK, and if that isn't done, the ACK just gets ignored - but this PR also just ignores it, only a bit later.

tools/docker/tests/test_flows.sh
CoralMalachi: you are right, I'll do it :)

CoralMalachi:
> First of all, let me commend you on the quality of this PR. It has all the features we like to see:
> 
> * Perfectly split into commits that do one thing only.
> * Each commit has an extensive and correct commit message.
> * The PR includes a test of the new feature.
> * Everything passes CI.
> 
> There is one important fix, the destination of the ACK message. The rest of the comments are nitpicks.
> 
> Note also that this PR really isn't needed to pass 5.10.1: the only change in the controller is the handling of the ACK, and if that isn't done, the ACK just gets ignored - but this PR also just ignores it, only a bit later.

Thanks! I'll do a fixup.

arnout (approved):
You fixed up the wrong commit: it should be "controller:add ack handler" instead of "tools:docker:add higher layer data payload flow".

But the fix itself looks OK, so please rebase on master (making sure to fix up the right commit) so we can merge.

arnout (commented):
Adding tlvHigherLevelData to the yaml + Autogenerated should be in a separate commit.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1599,6 +1603,21 @@ bool main_thread::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cmdu_rx
...
+
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+    LOG(DEBUG) << "sending ACK message back to controller";
```
tomereli: Please also add the MID value to the debug log

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -955,6 +957,13 @@ bool master_thread::handle_cmdu_1905_channel_preference_report(Socket *sd,
     return son_actions::send_cmdu_to_agent(sd, cmdu_tx);
 }
 
+bool master_thread::handle_cmdu_1905_ack_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received ACK_MESSAGE, mid=" << std::hex << int(mid);
```
tomereli: This should be `LOG(DEBUG)`

tomereli: Should add a `//TODO` comment which mentions that the ACK should be sent to the correct task and will be done as part of agent certification (nitpick)

tools/docker/tests/test_flows.sh
```diff
@@ -137,8 +137,26 @@ test_client_association() {
...
+    copies=200
+    
+    for i in `seq 1 $copies`
+    do
+        payload="$payload $mac_gateway_hex"
+    done
+
+    dbg "Send Higher Layer Data message"
+    eval send_bml_command '"bml_wfa_ca_controller \"DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8018,tlv_type,0xA0,tlv_length,\
```
tomereli, line 171: It will be useful if you can add a comment which explains what is being sent here as 1905 message, describing the TLVs that are part of the command.

tools/docker/tests/test_flows.sh
```diff
@@ -137,8 +137,26 @@ test_client_association() {
...
+0x04b1,tlv_value,{0x00 $payload}\""' $redirect
+
+    dbg "Confirming higher layer data message was received"
```
tomereli: --> dbg "Confirming higher layer data message was received in the agent"

tomereli (approved):
nitpick comments, but please fix them while rebasing

CoralMalachi:
> You fixed up the wrong commit: it should be "controller:add ack handler" instead of "tools:docker:add higher layer data payload flow".
> 
> But the fix itself looks OK, so please rebase on master (making sure to fix up the right commit) so we can merge.

Ok, I'll check it now.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: @tomereli  Why do we need to add the MID value to the debug log after sending the ACK? It's still the same MID.
I understood that only when we set the value of mid to zero in the `CmduMessageTx::create()` method , the value of MID is increasing by one.
is it correct?

tomereli: When we set MID=0, the `transport service` allocates a new MID value which is incremented by one from the last one is used.
I think that for each ACK message we should add the message type to which we are "ack'ing" and the MID value, so it will be easy to see in the log of the agent that ACK was sent, and in the controller that ACK for the same message was received.

This should have been a (nitpick) anyway, we can refactor that later on.

controller/src/beerocks/master/beerocks_master_mrouter.cpp
```diff
@@ -56,7 +56,6 @@ bool master_mrouter::init()
         ieee1905_1::eMessageType::CHANNEL_PREFERENCE_REPORT_MESSAGE,
         ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE,
         ieee1905_1::eMessageType::OPERATING_CHANNEL_REPORT_MESSAGE,
-        ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE,
```
tomereli: Why is this removed??

tomereli (changes requested):

tomereli (approved):

tools/docker/tests/test_flows.sh
CoralMalachi: #282

PR #272, opened 2019-09-09
Controller:master:db use sMacaddr instead of std::string

maksym-bielan:
The master database son::dbuses std::string types to represent MAC addresses. This leads to a lot of conversions to sMacAddr structures, and carries a risk that a string sneaks in somewhere that is not a valid MAC address.

So all MAC addresses currently are represented by sMacAddr structure

### DIFF
--- a/common/beerocks/bcl/include/beerocks/bcl/network/network_utils.h
+++ b/common/beerocks/bcl/include/beerocks/bcl/network/network_utils.h
@@ -36,6 +36,7 @@ class network_utils {
 public:
     static const std::string ZERO_IP_STRING;
     static const std::string ZERO_MAC_STRING;
+    static const sMacAddr ZERO_MAC;
     static const std::string WILD_MAC_STRING;
 
     typedef struct {
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -152,9 +152,9 @@ public:
     void set_log_level_state(const beerocks::eLogLevel &log_level, const bool &new_state);
 
     // General set/get
-    bool has_node(std::string mac);
+    bool has_node(sMacAddr mac);
 
-    bool add_virtual_node(std::string mac, std::string real_node_mac);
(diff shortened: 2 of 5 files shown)


common/beerocks/bcl/include/beerocks/bcl/network/network_utils.h
```diff
@@ -37,6 +37,7 @@ class network_utils {
     static const std::string ZERO_IP_STRING;
     static const std::string ZERO_MAC_STRING;
     static const std::string WILD_MAC_STRING;
+    static const sMacAddr ZERO_MAC;
```
arnout, line 39: :+1:

controller/src/beerocks/master/db/db.h
```diff
@@ -513,18 +508,6 @@ class db {
                     certification_tx_buffer.get());
     };
 
-    struct bss_info_conf_t {
```
arnout: I think this got removed accidentally while rebasing...

arnout (commented):
Although it looks OK too far, I think it's way too ambitious to try to convert everything in a single commit.

Instead, handle just a few functions and their callers at a time.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1106,7 +1106,7 @@ bool master_thread::handle_cmdu_1905_topology_discovery(Socket *sd,
         return false;
     }
     auto al_mac = network_utils::mac_to_string(tlvAlMac->mac());
-    if (database.has_node(al_mac)) {
+    if (database.has_node(network_utils::mac_from_string(al_mac))) {
```
arnout: `al_mac` was just converted to string in the line above. Therefore, it would be better to use `tlvAlMac->mac()` directly here.

arnout (approved):
Looks very good already, I would like to merge it as is. You can still make a follow-up in a new PR.

The commit message should be wrapped at 72 columns - this is the case for the second paragraph but not the first.

The one comment I have is just a nitpick.

tomereli (approved):

Issue #274, opened 2019-09-09
[BUG] controller: autoconfig search parse fail with reference agent


CoralMalachi:
**Describe the bug** -
While testing with reference CTT agent in the testbed, it turned out that
one of the golden unit agents sends the autoconfig search CMDU in a different order than what the prplMesh controller expects - autoconfig freq band TLV  comes last when the controller expects it to be third, which caused the prplMesh controller to fail on parsing.

Fix that in the controller by adding support for any order of TLVs in autoconfig search message.

1. in `main_thread::send_autoconfig_search_message` change the current order of addClass TLV, for example, move  the `addClass<ieee1905_1::tlvAlMacAddressType>()` action to be last.
2. Build
3. Run controller only (./test_gw_repeater.sh -f --gateway-only) and expect failure.
4. See in the controller's logs : "TLV type mismatch".

## Expected behavior
test_gw_repeater failed. 

## Screenshots
![image](https://user-images.githubusercontent.com/38204874/64528374-fb443700-d310-11e9-9ba4-fd603a5b56c4.png)

tomereli:
What is the "planned" label?

tomereli:
Fixed in #277

PR #277, opened 2019-09-09
controller:autoconfig search:support any order TLVs

CoralMalachi:
While testing with reference CTT agent in the testbed, it turned out that
one of the golden unit agents sends the autoconfig search CMDU in a
different order than what the prplMesh controller expects - autoconfig
freq band TLV comes last when the controller expects it to be third,
which caused the prplMesh controller to fail on parsing.

Fix the bug by adding a while loop which checks the type of each TLV the
in the receiving CMDU and add the right class to it, so now the
controller will handle this type of CMDU with any TLVs order inside it.

Implements #274 

### DIFF
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -244,7 +244,42 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
 
     LOG(DEBUG) << "Received AP_AUTOCONFIGURATION_SEARCH_MESSAGE";
 
-    auto tlvAlMacAddressType = cmdu_rx.addClass<ieee1905_1::tlvAlMacAddressType>();
+    std::shared_ptr<ieee1905_1::tlvAlMacAddressType> tlvAlMacAddressType     = nullptr;
+    std::shared_ptr<ieee1905_1::tlvSearchedRole> tlvSearchedRole             = nullptr;
+    std::shared_ptr<ieee1905_1::tlvAutoconfigFreqBand> tlvAutoconfigFreqBand = nullptr;
+    std::shared_ptr<wfa_map::tlvSupportedService> tlvSupportedServiceIn      = nullptr;
+    std::shared_ptr<wfa_map::tlvSearchedService> tlvSearchedService          = nullptr;
+    int type = cmdu_rx.getNextTlvType();
+
+    while ((type = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
+        switch (type) {
+        case int(wfa_map::eTlvTypeMap::TLV_SUPPORTED_SERVICE):
+            LOG(DEBUG) << "Found TLV_SUPPORTED_SERVICE TLV";
+            tlvSupportedServiceIn = cmdu_rx.addClass<wfa_map::tlvSupportedService>();
+            break;
+        case int(ieee1905_1::eTlvType::TLV_AL_MAC_ADDRESS_TYPE):
+            LOG(DEBUG) << "Found TLV_AL_MAC_ADDRESS_TYPE TLV";
+            tlvAlMacAddressType = cmdu_rx.addClass<ieee1905_1::tlvAlMacAddressType>();
+            break;
+        case int(ieee1905_1::eTlvType::TLV_SEARCHED_ROLE):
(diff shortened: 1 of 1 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -239,7 +239,36 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
 
     LOG(DEBUG) << "Received AP_AUTOCONFIGURATION_SEARCH_MESSAGE";
 
-    auto tlvAlMacAddressType = cmdu_rx.addClass<ieee1905_1::tlvAlMacAddressType>();
+    std::shared_ptr<ieee1905_1::tlvAlMacAddressType> tlvAlMacAddressType     = nullptr;
+    std::shared_ptr<ieee1905_1::tlvSearchedRole> tlvSearchedRole             = nullptr;
+    std::shared_ptr<ieee1905_1::tlvAutoconfigFreqBand> tlvAutoconfigFreqBand = nullptr;
+    std::shared_ptr<wfa_map::tlvSupportedService> tlvSupportedServiceIn      = nullptr;
+    std::shared_ptr<wfa_map::tlvSearchedService> tlvSearchedService          = nullptr;
+    int type = cmdu_rx.getNextTlvType();
```
tomereli, line 252: In this case, I think that `switch-case` is more appropriate:
```suggestion
    while ((int type = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
        switch (type) {
            case int(wfa_map::eTlvTypeMap::TLV_SUPPORTED_SERVICE):
                LOG(DEBUG) << "Found TLV_SUPPORTED_SERVICE TLV";
                tlvSupportedServiceIn = cmdu_rx.addClass<wfa_map::tlvSupportedService>();
                break;
            case int(ieee1905_1::eTlvType::TLV_AL_MAC_ADDRESS_TYPE):
                ...
                break;
            default:
                LOG(ERROR) << ...
                return false;
        }
    }
```

Also note that I defined `int type = cmdu_rx.getNextTlvType()` in the while condition, so it is not necessary to do the same outside the `while` loop. Same goes to the assignment at the end of the while.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -239,7 +239,36 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+            LOG(DEBUG) << "Found TLV_SEARCHED_ROLE TLV";
+            tlvSearchedRole = cmdu_rx.addClass<ieee1905_1::tlvSearchedRole>();
+        } else if (type == int(ieee1905_1::eTlvType::TLV_AUTOCONFIG_FREQ_BAND)) {
+            LOG(DEBUG) << "Found TLV_AUTOCONFIG_FREQ_BAND TLV";
+            tlvAutoconfigFreqBand = cmdu_rx.addClass<ieee1905_1::tlvAutoconfigFreqBand>();
+        } else if (type == int(wfa_map::eTlvTypeMap::TLV_SEARCHED_SERVICE)) {
+            LOG(DEBUG) << "Found TLV_SEARCHED_SERVICE TLV";
+            tlvSearchedService = cmdu_rx.addClass<wfa_map::tlvSearchedService>();
+        } else {
+            LOG(ERROR) << "Unknown TLV, type " << std::hex << type;
```
tomereli: Please update the error log to indicate that this autoconfig search message is **dropped**
```suggestion
            LOG(ERROR) << "Unknown TLV, type " << std::hex << type << ", dropping...";
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -239,7 +239,36 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_search(Socket *sd,
...
+            return false;
+        }
+        type = cmdu_rx.getNextTlvType();
```
tomereli: not needed.

controller/src/beerocks/master/son_master_thread.cpp
tomereli, line 252: you already assign `type` in each while loop below, so this is redundant, but keep reading the next comment.

tomereli (changes requested):

tomereli (approved):

PR #282, opened 2019-09-11
Feature/add hex convert method

CoralMalachi:
- Add a method in `functions.sh` for converting colon-MAC address to 0x format.

- Using the new method from the previous commit everywhere is needed.

### DIFF
--- a/tools/docker/functions.sh
+++ b/tools/docker/functions.sh
@@ -16,9 +16,8 @@ status() {
 	printf '\033[1;35m'"$@\n"'\033[0m'
 }
 
-convert_to_hex(){
-   mac=`echo $@ | tr --delete :`
-   mac="0x${mac}"
+mac_to_hex(){
+   echo "0x$1" | tr --delete :
 }
 
 err() {
--- a/tools/docker/functions.sh
+++ b/tools/docker/functions.sh
@@ -16,6 +16,11 @@ status() {
 	printf '\033[1;35m'"$@\n"'\033[0m'
 }
 
+convert_to_hex(){
+   mac=`echo $@ | tr --delete :`
+   mac="0x${mac}"
+}
+
 err() {
     printf '\033[1;31m'"$@\n"'\033[0m'
 }
--- a/tools/docker/tests/test_flows.sh
+++ b/tools/docker/tests/test_flows.sh
@@ -155,8 +155,8 @@ test_client_association() {
 test_higher_layer_data_payload_trigger() {
     status "test higher layer data payload"
     
-    mac_gateway_hex=`echo $mac_gateway | tr --delete :`
-    mac_gateway_hex="0x${mac_gateway_hex}"
+    mac_gateway_hex=$(mac_to_hex $mac_gateway)
+    dbg "mac_gateway_hex = ${mac_gateway_hex}"
     copies=200
     
     for i in `seq 1 $copies`
@@ -188,8 +188,7 @@ test_topology() {
     docker exec -it repeater1 sh -c 'grep -i -q "TOPOLOGY_QUERY_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent.log'
 }
 test_init() {
-    status "test initialization"
-
+    status "test initialization"    
     eval ${scriptdir}/test_gw_repeater.sh -f -r "repeater1" -r "repeater2" -d 7 $redirect || {
         err "start GW+Repeater failed, abort"
         exit 1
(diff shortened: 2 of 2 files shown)


tools/docker/functions.sh
```diff
@@ -16,6 +16,11 @@ status() {
 	printf '\033[1;35m'"$@\n"'\033[0m'
 }
 
+convert_to_hex(){
```
tomereli: Better name would be `mac_to_hex()`

tools/docker/functions.sh
```diff
@@ -16,6 +16,11 @@ status() {
...
+   mac=`echo $@ | tr --delete :`
```
tomereli: mac should be passed as a single variable, so it would be $1
so -
```
mac=`echo $1 | tr --delete :`
```

tools/docker/functions.sh
```diff
@@ -16,6 +16,11 @@ status() {
...
+   mac="0x${mac}"
```
tomereli: So to summarize, you can remove this and do it in a single assignment:
```
mac="0x$(echo $1 | tr --delete :)"
```

tools/docker/tests/test_flows.sh
```diff
@@ -155,8 +155,10 @@ test_client_association() {
 test_higher_layer_data_payload_trigger() {
     status "test higher layer data payload"
     
-    mac_gateway_hex=`echo $mac_gateway | tr --delete :`
-    mac_gateway_hex="0x${mac_gateway_hex}"
+    mac=
```
tomereli: Why is there a new line after assignment?
This should be
`mac_gateway=$(mac_to_hex $mac_gateway)`

tools/docker/tests/test_flows.sh
```diff
@@ -189,7 +191,8 @@ test_topology() {
 }
 test_init() {
     status "test initialization"
-
+    mac=
```
tomereli: not needed

tools/docker/tests/test_flows.sh
```diff
@@ -208,20 +211,20 @@ test_init() {
     mac_agent2=$(grep "IRE_BRIDGE" "$connmap" | sed -n 2p | awk '{print $5}' | cut -d ',' -f 1)
     dbg "mac_agent2 = ${mac_agent2}"
 
-    mac_agent1_wlan0=$(grep "RADIO: wlan0" "$connmap" | head -1 | awk '{print $4}' | cut -d ',' -f 1 | tr --delete :)
-    mac_agent1_wlan0="0x${mac_agent1_wlan0}"
+    convert_to_hex $(grep "RADIO: wlan0" "$connmap" | head -1 | awk '{print $4}' | cut -d ',' -f 1)
```
tomereli: single line assignment:
```
mac = $(mac_to_hex $(grep ...))

File: tools/docker/functions.sh [hunk 7]
```diff
@@ -16,6 +16,11 @@ status() {
...
+}
```
tomereli, line 21: Now after looking at the second commit, I think it would be much simpler to have this function echo the output instead of putting it in `mac` global variable which is pretty ugly.

So the function as I would prefer it would simply be:
```
mac_to_hex() {
    echo "0x$(echo $1 | tr --delete :)"
}
```

So it can be used as follows:
```
mac=$(mac_to_hex 00:11:22:33:44:55)
```
```

**Review by tomereli — DISMISSED** [review]

File: tools/docker/functions.sh (hunk 3 above)
Comment (arnout) [reply tools/docker/functions.sh]: I like the smiley :)

File: tools/docker/functions.sh [hunk 8]
```diff
@@ -16,6 +16,10 @@ status() {
 	printf '\033[1;35m'"$@\n"'\033[0m'
 }
 
+mac_to_hex(){
+   echo "0x$(echo $1 | tr --delete :)"
```
arnout: Sorry, I still have a slight improvement:
```suggestion
   echo "0x$1" | tr --delete :
```

It does mean the smiley is gone :-(

tools/docker/tests/test_flows.sh
```diff
@@ -189,7 +189,7 @@ test_topology() {
 }
 test_init() {
     status "test initialization"
-
+
```
arnout: Spurious whitespace

arnout (approved):
Nitpicks, so approving already.

Please rebase -i --autosquash on master.

PR #283, opened 2019-09-11
tlvf: missing skip tlv function

YuriyMasechko:
Where ever parsing multiap cmdu, there is need to check what is the next tlv that is needed to be parse, and then parse it, and repeat it untill the next tlv is EndOfMessage.
It may happen that cmdu handler function will receive unexpected tlv.
If such thing happens, temporary implementation can be one of:
    1.The parser function will iterate over it infinite times causing the process to stuck. This since the tlv is not added to the cmdu.
    2.The handler function may stop handling the cmdu by return statment. This means that the cmdu will be thrown.

New class tlvUnknown was added to skip unexpected tlv. Use
msg.addClass<tlvUnknown>() to skip tlv.

### DIFF
--- a/framework/tlvf/test/tlvf_test.cpp
+++ b/framework/tlvf/test/tlvf_test.cpp
@@ -16,6 +16,7 @@
 #include "tlvf/ieee_1905_1/tlvLinkMetricQuery.h"
 #include "tlvf/ieee_1905_1/tlvMacAddress.h"
 #include "tlvf/ieee_1905_1/tlvNon1905neighborDeviceList.h"
+#include "tlvf/ieee_1905_1/tlvUnknown.h"
 #include "tlvf/ieee_1905_1/tlvVendorSpecific.h"
 #include "tlvf/ieee_1905_1/tlvWscM2.h"
 #include "tlvf/wfa_map/tlvApCapability.h"
@@ -402,6 +403,12 @@ int test_all()
     LOG(DEBUG) << "Done (WSC M2)";
     MAPF_DBG("TLV LENGTH WSC M2: " << thirdTlv->length());
 
+    auto mactlv                     = msg.addClass<tlvMacAddress>();
+    const uint8_t gTlvMacAddress[6] = {0x01, 0x02, 0x03, 0x04, 0x05, 0x06};
+    std::copy_n(gTlvMacAddress, 6, mactlv->mac().oct);
+
+    MAPF_DBG("TLV LENGHT MAC: " << mactlv->length());
+
     auto fourthTlv     = msg.addClass<tlvTestVarList>();
     fourthTlv->var0()  = 0xa0;
     allocation_succeed = fourthTlv->alloc_simple_list(2);
@@ -542,6 +549,18 @@ int test_all()
         errors++;
     }
 
+    auto tlvunknown = received_message.addClass<tlvUnknown>();
+    if (tlvunknown != nullptr) {
+        LOG(DEBUG) << "TLVUnknown Type: " << int(tlvunknown->type());
+
+        LOG(DEBUG) << "TLVUnknown data:";
+        for (uint8_t data_idx = 0; data_idx < tlvunknown->data_length(); data_idx++) {
+            LOG(DEBUG) << " " << int(tlvunknown->data()[data_idx]);
+        }
+    } else {
+        MAPF_ERR("TLVUnknown is NULL");
+        errors++;
+    }
     auto tlv4 = received_message.addClass<tlvTestVarList>();
     if (tlv4 == nullptr) {
         MAPF_ERR("TLV4 is NULL");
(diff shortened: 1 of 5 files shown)


tomereli:
Please don't use merges, we do rebase instead. See the CONTRIBUTING.md for more information on our git workflow.

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvUnknown.yaml
```diff
@@ -0,0 +1,14 @@
+#
+---
+_namespace: ieee1905_1
+
+tlvUnknown:
+  _type: class
+  _is_tlv_class : True
+  type:
```
tomereli: Please put this in the same line - `type: uint8_t`

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvUnknown.yaml
```diff
@@ -0,0 +1,14 @@
...
+    _type: uint8_t
+  length: uint16_t
+  data_length: uint8_t
```
tomereli: (nitpick) You can also use "unknown length" lists for this:
```
type: uint8_t 
length: uint16_t
data: []
```

tomereli (changes requested):
1. we don't use merges in prplMesh, so please rebase -i on origin/master to remove the merge commit
2. You are missing a commit which updates the autogenerated files

tomereli (approved):

tomereli:
Let's wait with the merge so @arnout can also review

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvUnknown.yaml
```diff
@@ -0,0 +1,13 @@
+#
+---
+_namespace: ieee1905_1
+
+tlvUnknown:
+  _type: class
+  _is_tlv_class : True
+  type: uint8_t
+  length: uint16_t
+  data_length: uint8_t
```
arnout: This is not correct. The only fields we know about are the type and the length fields. There is no data_length field that we can rely on.

So instead, it should be:
```
  type: uint8_t
  length: uint16_t
  data:
    _type: uint8_t
   _length: []
```

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvUnknown.yaml
```diff
@@ -0,0 +1,13 @@
...
+  data:
+    _type: uint16_t
```
arnout: This has to be `uint8_t`, otherwise it would only support an even number of bytes.

framework/tlvf/tlvf.py
```diff
@@ -725,7 +725,7 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
                     
                     # Add type checking
                     lines_cpp = []
-                    if obj_meta.is_tlv_class and param_name == MetaData.TLV_TYPE_TYPE:
+                    if obj_meta.is_tlv_class and param_name == MetaData.TLV_TYPE_TYPE and param_val_const:
```
arnout, line 728: It would be better to make this change in a separate commit, which an explanation in the commit message why it is needed.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -402,6 +403,21 @@ int test_all()
     LOG(DEBUG) << "Done (WSC M2)";
     MAPF_DBG("TLV LENGTH WSC M2: " << thirdTlv->length());
 
+    auto unknownTlv = msg.addClass<tlvUnknown>();
```
arnout: It would be more appropriate to construct a valid TLV here, because the goal is to be able to parse an actual valid TLV but without knowing which class it corresponds to. Take a simple TLV, e.g. tlvMacAddress.

arnout (changes requested):
The defintion of tlvUnknown is not OK. That really has to be fixed.

The other comments are less important.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -542,6 +549,20 @@ int test_all()
...
+    } else {
+        MAPF_ERR("TLVUnknown is NULL");
+        errors++;
+    }
+    LOG(DEBUG) << "TLVUnknown Type: " << int(tlvunknown->type());
+
+    LOG(DEBUG) << "TLVUnknown data:";
+    for (uint8_t data_idx = 0; data_idx < tlvunknown->length(); data_idx++) {
+        LOG(DEBUG) << " " << (int)std::get<1>(tlvunknown->data(data_idx));
+    }
```
arnout: All this should also be in the `tlvunknown != nullptr` branch.

tomereli (dismissed):
@arnout made valid points, so disapproving to give a chance to fix those.

arnout (approved):
I still needed to adapt to the changed tlvf (now returning uint8_t* instead of a tuple). But I've done that and rebased so it can be merged.

PR #286, opened 2019-09-12
agent:backhaul_manager:really parse tlvHigherLayerData

CoralMalachi:
Fix a bug that has been added in commit d491828.

The problem is addClass is performed not on the right CMDU, but on an
empty CMDU, which causes the payload length to be zero.

Therefore, perform the addClass method on the right CMDU.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1653,9 +1653,10 @@ bool main_thread::handle_1905_higher_layer_data_message(ieee1905_1::CmduMessageR
         return false;
     }
 
+    const auto protocol       = tlvHigherLayerData->protocol();
     const auto payload_length = tlvHigherLayerData->payload_length();
-    LOG(DEBUG) << "protocol: " << std::hex << protocol;
-    LOG(DEBUG) << "payload_length: " << int(payload_length);
+    LOG(DEBUG) << "protocol: " << std::hex << int(protocol);
+    LOG(DEBUG) << "payload_length: " << std::hex << int(payload_length);
 
     // build ACK message CMDU
     auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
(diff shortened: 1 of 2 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1647,16 +1647,16 @@ bool main_thread::handle_1905_higher_layer_data_message(ieee1905_1::CmduMessageR
...
-    auto tlvHigherLayerData = cmdu_tx.addClass<wfa_map::tlvHigherLayerData>();
+    auto tlvHigherLayerData = cmdu_rx.addClass<wfa_map::tlvHigherLayerData>();
     if (!tlvHigherLayerData) {
         LOG(ERROR) << "addClass wfa_map::tlvHigherLayerData failed";
         return false;
     }
 
-    auto protocol     = tlvHigherLayerData->protocol();
-    const auto length = tlvHigherLayerData->length();
+    const auto protocol       = tlvHigherLayerData->protocol();
```
tomereli, line 1656: Is this related to this fix?

tomereli (dismissed):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: I wanted it to be uniform, Do you want me to split it to another commit?

tomereli: yes please

arnout (changes requested):
Could you also extend the test so it catches this issue? I guess that matching the protocol and length in the received message should be enough. For that, it's probably also easier to put everything in one line.

Also, the commit 97e8d8e is not in master. In master, it's d491828 (due to rebase when merging, github is a bit stupid in that respect).

Finally, a commit summary that just says "Fix a bug that has been added" is not very useful. Better would be: `agent: backhaul_manager: really parse tlvHigherLayerData`, followed by the rest of you message (i.e. "Fix a bug ...".

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: Ok :)

arnout (approved):

PR #287, opened 2019-09-13
bwl: dummy: add infrastructure for simulating events

tomereli:
Currently, we are using a dummy eventfd for external events which does
nothing but being added to the user's select.
Replace that with inotify, still unconfigured, as the base grounds for
adding external events simulation via files.

The second commit adds basic infrastructure for events simulation by writing
the event string to the EVENT file in <BEEROCKS_TMP_PATH>/<ifname>/EVENT.

The way this is implemented is by using inotify to monitor changes in the EVENT file directory, but not actually processing the inotify event but only using it to be notified that *something* changed.
The reason for that is to simplify the implementation - only the first line in the EVENT file will be parsed.

For example, simulating a STA with MAC 11:22:33:44:55:66 is connected to wlan0 AP is done as follows:

```bash
echo "STA_CONNECTED,11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan0/EVENT
```
Note that currently, no handling is done, this commit provides
infrastructure only. Events handling will be done in separate PRs.

### DIFF
--- a/common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
@@ -14,6 +14,7 @@
 #include <beerocks/bcl/son/son_wireless_utils.h>
 
 #include <easylogging++.h>
+#include <limits.h>
 #include <sys/inotify.h>
 
 #define UNHANDLED_EVENTS_LOGS 20
@@ -89,6 +90,8 @@ base_wlan_hal_dummy::base_wlan_hal_dummy(HALType type, std::string iface_name, b
     // inotify_init1 not available with older kernels, consequently inotify reads block.
     // inotify_init1 allows directory events to complete immediately, avoiding buffering delays. In practice,
     // this significantly improves monotiring of newly created subdirectories.
+    // Note that IN_NONBLOCK is not defined in old kernels so we get to the #else and call inotify_init()
+    // in this case.
 #ifdef IN_NONBLOCK
     m_fd_ext_events = inotify_init1(IN_NONBLOCK);
 #else
@@ -98,6 +101,10 @@ base_wlan_hal_dummy::base_wlan_hal_dummy(HALType type, std::string iface_name, b
     if (m_fd_ext_events < 0) {
         LOG(FATAL) << "Failed creating m_fd_ext_events: " << strerror(errno);
     }
+
+    std::string path = std::string(BEEROCKS_TMP_PATH) + "/" + get_iface_name();
+    mkdir(path.c_str(), S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
+    inotify_add_watch(m_fd_ext_events, path.c_str(), (IN_CREATE | IN_DELETE | IN_MODIFY));
 
     // Initialize the FSM
     fsm_setup();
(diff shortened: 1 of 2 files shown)


common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -83,8 +83,20 @@ base_wlan_hal_dummy::base_wlan_hal_dummy(HALType type, std::string iface_name, b
     }
 
     // Set up dummy external events fd
-    if ((m_fd_ext_events = eventfd(0, EFD_SEMAPHORE)) < 0) {
-        LOG(FATAL) << "Failed creating eventfd: " << strerror(errno);
+    // dummy implementation is based on monitoring text file changes which
+    // are used to simulate clients activity - STA connected, disconnected, etc.
+    // For that we are creating the INOTIFY instance
+    // inotify_init1 not available with older kernels, consequently inotify reads block.
```
morantr, line 90: "consequently read system call on inotify file descriptor will block" is more clear IMO.
Second, do we need to support old kernel here? I mean this is a dummy module after all, and run on the latest Linux version I guess, no?

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -14,9 +14,12 @@
 #include <beerocks/bcl/son/son_wireless_utils.h>
 
 #include <easylogging++.h>
+#include <limits.h>
 #include <sys/inotify.h>
 
 #define UNHANDLED_EVENTS_LOGS 20
+#define EVENT_SIZE (sizeof(struct inotify_event))
+#define EVENT_BUF_LEN (1024 * (EVENT_SIZE + NAME_MAX + 1))
```
morantr: Could you please explain the calculation?
I see in the man page the reason for `sizeof(struct inotify_event) + NAME_MAX + 1` but couldn't understand why did you multiply it by 1024.
It is better to add a comment explaining it, even what said on the man page.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -98,6 +101,10 @@ base_wlan_hal_dummy::base_wlan_hal_dummy(HALType type, std::string iface_name, b
     if (m_fd_ext_events < 0) {
         LOG(FATAL) << "Failed creating m_fd_ext_events: " << strerror(errno);
     }
+    // mkdir(BEEROCKS_TMP_PATH,  S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
```
morantr: This comment should be removed.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -83,8 +83,20 @@ base_wlan_hal_dummy::base_wlan_hal_dummy(HALType type, std::string iface_name, b
...
     // Set up dummy external events fd
-    if ((m_fd_ext_events = eventfd(0, EFD_SEMAPHORE)) < 0) {
-        LOG(FATAL) << "Failed creating eventfd: " << strerror(errno);
+    // dummy implementation is based on monitoring text file changes which
+    // are used to simulate clients activity - STA connected, disconnected, etc.
+    // For that we are creating the INOTIFY instance
+    // inotify_init1 not available with older kernels, consequently inotify reads block.
+    // inotify_init1 allows directory events to complete immediately, avoiding buffering delays. In practice,
+    // this significantly improves monotiring of newly created subdirectories.
+#ifdef IN_NONBLOCK
```
morantr, line 95: And if you explained to use one instead of the other according to the kernel version, then why didn't you do it?
Edit: Ok I think I got it! The `IN_NONBLOCK` is not defined in an old kernel version. I think it is worth mention.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -133,7 +140,44 @@ bool base_wlan_hal_dummy::refresh_vap_info(int vap_id) { return true; }
...
+    if (length < 0) {
+        LOG(ERROR) << "Failed to read from m_fd_ext_events";
+        return false;
+    }
+
+    std::string event_file = std::string(BEEROCKS_TMP_PATH) + "/" + get_iface_name() + "/EVENT";
+    std::ifstream stream;
+    stream.open(event_file);
+    if (!stream.is_open()) {
+        LOG(DEBUG) << "Invalid event, missing " << event_file << " file";
```
morantr, line 181: Isn't it an ERROR?

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -133,7 +140,44 @@ bool base_wlan_hal_dummy::refresh_vap_info(int vap_id) { return true; }
...
+        return true;
+    }
+    std::string event;
+    std::getline(stream, event);
+    LOG(DEBUG) << "Received event " << event;
```
morantr, line 186: I've shown an example that the event is processed, but couldn't see it here. The event is being printed and that it.
I guess you have pending changes you haven't committed yet, am I right?

morantr (changes requested):

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
tomereli: It doesn't hurt to support older kernels. Regarding the comment, I ported this from another project so I prefer to keep the comment as is.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
arnout: > I've shown an example that the event is processed, but couldn't see it here. The event is being printed and that it.
> I guess you have pending changes you haven't committed yet, am I right?

It's mentioned very explicitly in the commit message that this commit/PR only adds the generic handling, specific events will be added later (e.r. #289).

arnout (approved):
I can accept this PR as is (if the commented-out line is removed as mentioned by @morantr). However, I have two "architectural" comments that I think would improve this.

1. I don't like the extra directory very much. I would also encode (dummy) bwl in the name. So I would suggest `$BEEROCKS_TMP_PATH/bwl-event-wlan0`.
2. IMO using inotify for this is pretty kludgy. Why not use a FIFO? This has the advantage of being pretty close to the hostapd socket (just a "file" instead of a UDS, and it's unidirectional).  You do have to take care to open it in non-blocking mode though.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
tomereli: ok

tomereli:
> I can accept this PR as is (if the commented-out line is removed as mentioned by @morantr). However, I have two "architectural" comments that I think would improve this.
> 
> 1. I don't like the extra directory very much. I would also encode (dummy) bwl in the name. So I would suggest `$BEEROCKS_TMP_PATH/bwl-event-wlan0`.
> 2. IMO using inotify for this is pretty kludgy. Why not use a FIFO? This has the advantage of being pretty close to the hostapd socket (just a "file" instead of a UDS, and it's unidirectional).  You do have to take care to open it in non-blocking mode though.

So I tried this in #306 but for some reason I am losing events - I couldn't find a way to overcome this. Maybe you can have a try?

morantr (approved):

PR #291, opened 2019-09-16
Controller:master:db use sMacaddr instead of std::string

maksym-bielan:
The master database son::db uses std::string types to represent MAC
addresses. This leads to a lot of conversions to sMacAddr structures,
and carries a risk that a string sneaks in somewhere that is not a
valid MAC address.

Changed std::string to sMacAddr for son::db::add_node and remove_node
function.

### DIFF
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -9,7 +9,6 @@
 #include "db.h"
 
 #include <beerocks/bcl/beerocks_utils.h>
-#include <beerocks/bcl/network/network_utils.h>
 #include <beerocks/bcl/son/son_wireless_utils.h>
 #include <easylogging++.h>
 
@@ -59,10 +58,10 @@ bool db::add_virtual_node(sMacAddr mac, sMacAddr real_node_mac)
     return true;
 }
 
-bool db::add_node(std::string mac, std::string parent_mac, beerocks::eType type,
-                  std::string radio_identifier)
+bool db::add_node(const sMacAddr &mac, const sMacAddr &parent_mac, beerocks::eType type,
+                  const sMacAddr &radio_identifier)
 {
-    if (mac.empty()) {
+    if (mac == network_utils::ZERO_MAC) {
         LOG(ERROR) << "can't insert node with empty mac";
         return false;
     }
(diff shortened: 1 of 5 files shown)


controller/src/beerocks/master/db/db.cpp
```diff
@@ -59,10 +58,10 @@ bool db::add_virtual_node(sMacAddr mac, sMacAddr real_node_mac)
     return true;
 }
 
-bool db::add_node(std::string mac, std::string parent_mac, beerocks::eType type,
-                  std::string radio_identifier)
+bool db::add_node(sMacAddr mac, sMacAddr parent_mac, beerocks::eType type,
```
morantr: All sMacAddr arguments should be const reference.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -118,31 +120,31 @@ bool db::add_node(std::string mac, std::string parent_mac, beerocks::eType type,
     return true;
 }
 
-bool db::remove_node(std::string mac)
+bool db::remove_node(sMacAddr mac)
```
morantr: const reference here too

morantr (changes requested):

morantr (approved):

controller/src/beerocks/master/db/db.h
```diff
@@ -155,10 +156,11 @@ class db {
     bool has_node(sMacAddr mac);
 
     bool add_virtual_node(sMacAddr mac, sMacAddr real_node_mac);
-    bool add_node(std::string mac, std::string parent_mac = std::string(),
-                  beerocks::eType type         = beerocks::TYPE_CLIENT,
-                  std::string radio_identifier = std::string());
-    bool remove_node(std::string mac);
+    bool add_node(const sMacAddr &mac,
```
arnout, line 159: According to [CppCoreGuideline F.16](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-in), a small object should be past by value instead of const-ref, and they say that 'small' is two-three words. Since sMacAddr is 6 bytes, it is less than two words, so I'd say it's small. And indeed, I can imagine that passing it by reference creates overhead due to additional pointer dereferences.

arnout (approved):
Small nitpick which I'll modify and force-push and then merge.

controller/src/beerocks/master/db/db.h
arnout: I accidentally pushed and merged before committing those changes, so I've created #331 instead.

PR #292, opened 2019-09-16
controller: fix endless loop on addClass failure

tomereli:
When addClass fails, it doesn't increment the buffer pointers so the
next time addClass will fail exactly the same.
This can cause the controller to hang on autoconfig parsing, for example
if the WSC TLV contains anything other than M1.
Fix that by checking the return code of addClass and returning false on
error.

While we're at it - do the same for the agent - there should be NO
addClass without an explicit return code checking.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4479,8 +4479,10 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
 
     while ((type = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
         if (type == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER)) {
-            LOG(DEBUG) << "Found TLV_AP_RADIO_IDENTIFIER TLV";
             ruid = cmdu_rx.addClass<wfa_map::tlvApRadioIdentifier>();
+            if (!ruid)
+                return false;
+            LOG(DEBUG) << "Found TLV_AP_RADIO_IDENTIFIER TLV";
         } else if (type == int(ieee1905_1::eTlvType::TLV_WSC)) {
             // parse all M2 TLVs
             LOG(DEBUG) << "Found TLV_WSC TLV (assuming M2)";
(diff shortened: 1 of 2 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4479,8 +4479,9 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
 
     while ((type = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
         if (type == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_IDENTIFIER)) {
+            if ((ruid = cmdu_rx.addClass<wfa_map::tlvApRadioIdentifier>()) == nullptr)
```
morantr: smart pointers should be checked with '!' and not be compared to nullptr. Comparing to nullptr will cause klocwork issues.
use `if (!ruid) `instead.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -680,11 +680,14 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
 
     while ((type = cmdu_rx.getNextTlvType()) != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
         if (type == int(wfa_map::eTlvTypeMap::TLV_AP_RADIO_BASIC_CAPABILITIES)) {
+            if ((radio_basic_caps = cmdu_rx.addClass<wfa_map::tlvApRadioBasicCapabilities>()) ==
+                nullptr)
```
morantr: same here

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -680,11 +680,14 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
...
+                return false;
             LOG(DEBUG) << "Found TLV_AP_RADIO_BASIC_CAPABILITIES TLV";
-            radio_basic_caps = cmdu_rx.addClass<wfa_map::tlvApRadioBasicCapabilities>();
         } else if (type == int(ieee1905_1::eTlvType::TLV_WSC)) {
+            if ((tlvwscM1 = cmdu_rx.addClass<ieee1905_1::tlvWscM1>()) == nullptr)
```
morantr: same here

morantr (approved):

PR #304, opened 2019-09-17
tools/docker/runner/start-controller: fix typo.

rmelotte:


arnout (approved):
One small comment: the summary line should not end with a period. But this is not important enough to warrant a respin, so I'll commit as is.

PR #306, opened 2019-09-18
bwl: dummy: add infra for simulating events

tomereli:
This commit adds basic infrastructure for events simulation by writing
the event string to the event pipe at <BEEROCKS_TMP_PATH>/bwl-event-<ifname>.

For that, a named pipe is created at init for each bwl radio instance,
and is opened for both reading and writing which provides a way to
listen on the pipe in a non blocking manner.

The file descriptor is saved in m_fd_ext_events which is used in the
select loop by the parent class.

Simulating an event is done by simply writing into the pipe -
For example, simulating a STA with MAC 11:22:33:44:55:66 is connected to
wlan0 AP is done as follows:
echo "STA_CONNECTED,11:22:33:44:55:66" >
/tmp/$USER/beerocks/bwl-event-wlan0

Note that currently, no handling is done, this commit provides
infrastructure only. Events handling will be done in separate PRs.

### DIFF
--- a/common/beerocks/bwl/CMakeLists.txt
+++ b/common/beerocks/bwl/CMakeLists.txt
@@ -36,6 +36,8 @@ elseif(TARGET_PLATFORM STREQUAL "rdkb")
 # Linux
 elseif(TARGET_PLATFORM STREQUAL "linux")
     set(BWL_TYPE_DEFAULT "DUMMY")
+    set(BEEROCKS_TMP_PATH "/tmp/$ENV{USER}/beerocks")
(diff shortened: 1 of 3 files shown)


tomereli:
> For that, a named pipe is created at init for each bwl radio instance,
> and is opened for both reading and writing which provides a way to
> listen on the pipe in a non blocking manner.

After thinking about this a bit more, I'm not sure that using a named pipe is a better approach than using inotify due to the fact that using named pipe we can't receive the same event in both the monitor and ap manager, which is the case when receiving an event from hostapd.

arnout:
Since the inotify approach has already been merged, let's stick with that until it breaks.

I'm keeping the branch for reference, but closing this PR.

PR #308, opened 2019-09-19
Revive beerocks analyzer

rmelotte:
For the moment, the focus is on the connectivity map part


morantr (dismissed):
ed2f731e8c68b06d90147d9d7d5d59114ee77f9e commit message headline length is greater than 72 chars, and wrapped.
f6ca385f5c15acc40a3e56260c74f1d411a4d865 commit message headline typo: usr -> use
23c2d9e48d308d98715b94f4c0382a2a41f75e13 commit message headline length is greater than 72 chars, and wrapped.

Beside that,
1. The component part of the commit message headline is not supposed to be a file path.
Please change it to `tools: beerocks_analyzer: ` on all commits.

2. Most of the commits are not descriptives enough, and missing:
    - the context (current situation);
    - the problem statement (why the current situation is not good);
    - possible solutions with their pros and cons;
    - the chosen solution and why it was chosen;

    (https://github.com/prplfoundation/prplMesh/blob/master/CONTRIBUTING.md#commits)
    This would be more easy to review and understand, especially for a python beginner like me.

3. I see that on beerocks analyzer folder there is a reuierments.txt file. It does not have updated versions anymore after you updated some of the python packages.
I think it would be a great idea to replace this file with a markdown file, which says which packages are needed, and explains how to use the beerocks analyzer.

rmelotte:
Thanks for the comments! I'll improve the commits messages.

> I see that on beerocks analyzer folder there is a reuierments.txt file. It does not have updated versions anymore after you updated some of the python packages.
I think it would be a great idea to replace this file with a markdown file, which says which packages are needed, and explains how to use the beerocks analyzer.

I don't really understand the second sentence. Could you rephrase it ?
The requirements.txt file is needed for pip to be able to install the required dependencies.
I will add a README or even a bash script to run the tool once it's working.

morantr:
1. https://github.com/prplfoundation/prplMesh/pull/308/commits/7069fe0dd59f5b64202d7cc8685228da40d3c88c and https://github.com/prplfoundation/prplMesh/pull/308/commits/e366b8f890fd19570c627d644b00e73d689e4069 have the same headline (no description tough). Maybe can you unify them?

2. Same thing for https://github.com/prplfoundation/prplMesh/pull/308/commits/0f6dc748a5a64660a4a497b292495dc8a1b896a1 and https://github.com/prplfoundation/prplMesh/pull/308/commits/b501ecabc5646fbc6568735ef81521ef7af224c9. Both of them have a headline which is pretty similar which is very generic. Please be more specific about what you change. If there isn't much difference squash them.

3. https://github.com/prplfoundation/prplMesh/pull/308/commits/1a84e2000ca54cac25feccb18134fb930f4e7049 has too generic headline (no description). Please write what you fix explicitly.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -40,29 +40,26 @@
...
+    logger.info("                                                          - can execute with several -conf=<conf_file_name1> -conf=<conf_file_name2> ...")
+    logger.info("    -gw_ip=<IP of GW>                                     - GW ip to get updates from, default=192.168.1.1")
+    logger.info("    -my_ip=[IP of PC]                                     - PC ip to send updates, if not set will automaticly set")
+    logger.info("    -ssh_port=[ssh port of GW]                            - SSH port used to connect to GW")
 
 def signal_handler(signal, frame):
     global t_list, g_run_flag
 
-    PRINTF("")
+    logger.info("Signal received: " + str(signal))
```
arnout: It is better to make a big change like this purely mechanical (i.e. replace all occurrences of PRINTF with logger.info), and do other changes (change info to something else, add more info to the log, add a TODO comment) in separate commits.

If you really do it mechanically (e.g. with a sed program), then include the scriptlet you used in the commit message, and explicitly mention any manual fixups you needed to do afterwards.

However, it's not worth spending effort now to try to split up this commit. Just something to remember for the future.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -176,15 +173,15 @@ def openLogFiles(self, log_start_pos=0):
             try: # open log file
                 self.fdLog = open(self.fname,"rt")
                 if log_start_pos > 0:
-                    # PRINTF("openLogFiles: seek to %d\n" % log_start_pos)
+                    logger.debug("openLogFiles: seek to {:d}", log_start_pos)
```
arnout, line 175: About the "purely mechanical": this is one which would make sense to still do in the same commit, but then the commit message should mention this explicitly, e.g. "Replace commented-out PRINTF calls with logger.debug. Replace `%`-formatted strings with the `{}`-formatting used by logger."

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -685,8 +685,8 @@ def socket_server(log_file):
                 break
             while g_run_flag:
                 try:
-                    i=data.index('\n')
                 except:
+                    i=data.index(r'\n')
```
arnout: This makes no sense - a try-block without body is meaningless...

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -706,13 +710,42 @@ def socket_server(log_file):
...
+        The return status of beerocks_cli.
+    """
+    if not beerocks_cli_path.is_file():
+        raise ValueError("Path to beerocks_cli not found: {}".format(beerocks_cli_path))
+    if not target_ip:
+        target_ip = ip_address("127.0.0.1")
+
+    command = str(beerocks_cli_path) + " -a " + str(target_ip)
+    logger.debug("Starting beerock_cli -a")
+    exit_status = subprocess.call(command.split())
```
arnout: There shouldn't be a reason to build up the path as a string and then to split it. I.e., use:

```
command = [str(beerocks_cli_path), "-a", str(target_ip)]
exit_status = subprocess.call(command)
```

And in this case, I don't even think the explicit conversions to str are needed.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -706,13 +710,42 @@ def socket_server(log_file):
 
     file.close()
 
+
+def local_start_beerocks_cli(beerocks_cli_path: Path, target_ip: IPv4Address) -> int:
```
arnout: Type info, nice!

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -676,11 +676,13 @@ def socket_server(log_file):
                 g_marker_update = False
                 sent_bytes = connection.send("marker")
             try:
-                data += str(connection.recv(256))
+                logger.debug("Trying to get data from the socket")
+                data += connection.recv(256).decode("utf-8")
             except socket.timeout:
-                logger.info("Socket timed out")
+                logger.debug("Socket timed out")
```
arnout: This change seems to be unrelated/unneeded.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -340,43 +340,47 @@ def readSampleThread(self, update_start_time = False):
...
-                        self.timeUpdateSig.sig.emit(param_t1)
-                        if update_widgets and not self.wait_for_mark:
-                            if "BML_EVENT" in line:
-                                self.log_widget.readSampleAndUpdateLogger(line, param_t1, is_start, is_stop)
-                            else:
-                                if self.isMap:
-                                    self.cm_widget.readSample(line, param_t1, is_start, is_stop)
-                                if self.isGraphs:
-                                    self.wa_widget.readSampleAndUpdateGraphs(line, param_t1, is_start, is_stop)
+                if len(line) <= 0:
```
arnout: `len(line)` can never be smaller than 0, so this should be `len(line) == 0`, but then it's better to just do `if not line:`

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -340,43 +340,47 @@ def readSampleThread(self, update_start_time = False):
...
+                    time.sleep(1)
+                    continue
+                [param_t1, is_start, is_stop, is_map_update, sta_ap_ids] = self.readSample(line)
+                if len(sta_ap_ids) > 0:
+                    line += ", " + sta_ap_ids
+                if is_stop:
+                    logger.info("Read stop flag, stopping")
```
arnout, line 337: It is ***really bad*** to hide additional changes in a big indentation change...

tools/beerocks_analyzer/beerocks_analyzer_widget.py
```diff
@@ -19,8 +19,7 @@
 from PySide2.QtCore import QObject, Signal
 from PySide2.QtWidgets import QWidget, QVBoxLayout, QTabWidget, QHBoxLayout
 
-matplotlib.use('Qt4Agg')
-matplotlib.rcParams['backend.qt4']='PySide'
+matplotlib.use('Qt5Agg')
```
arnout, line 22: The same change was made in a different file two commits before. It's better to do those two changes together.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -734,9 +738,11 @@ def local_start_beerocks_cli(beerocks_cli_path: Path, target_ip: IPv4Address) ->
         raise ValueError("Path to beerocks_cli not found: {}".format(beerocks_cli_path))
     if not target_ip:
         target_ip = ip_address("127.0.0.1")
-
-    command = str(beerocks_cli_path) + " -a " + str(target_ip)
-    logger.debug("Starting beerock_cli -a")
+    command = ""
+    if docker_container_name:
+        command += "docker exec " + docker_container_name + " "
```
arnout: It makes more sense to me to first build the basic command and then prepend the rest. With the list, that becomes:

```python
command = [str(beerocks_cli_path), "-a", str(target_ip)]
if docker_container_name:
    command = ["docker", "exec", docker_container_name] + command
```

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -57,18 +57,15 @@ def printUsage():
 def signal_handler(signal, frame):
     global t_list, g_run_flag
 
-    logger.info("Signal received: " + str(signal))
+    logger.info("Signal {} received \n terminating".format(str(signal)))
```
arnout: No need for `.format`, logger.info already implies it.

arnout (approved):
I have a bunch of comments on the commit messages.

- Summary line should never end with a period (cfr. CONTRIBUTING.md).
- "fix the python version check" is not a good commit message; you can make it more specific, like "set minimum python version to 3.5". The commit message should also explain why you need this. The next commit message ("import QtWidgets") is a good example of the right way to do it.
- Similar comment for "fix xrange usage".
- Also "fix mac conversion" warrants a better explanation of why it is wrong.
- Also "fix old Python2 usage" is not a good summary line.
- In that same commit message, you use "directory" to mean "dictionary".
- A few commits later, "fix Python2 usage" again...
- "implement threads termination" is a bit too big for easy review. It should have been split into a commit that introduces the SocketServerThread class (which is largely an indentation change), one that introduces BeerocksCliThread, etc. and finally one that implements terminate on each of them.

Since all of the comments are of the nitpick kind, I'm approving already. So please fix as much as you can/like/have time for, rebase on master and force-push. And set this PR to ready of course.

rmelotte:
Thanks for all the comments, I still need to do that quickly before marking it ready for review.
I just pushed what I have now anyway so that anybody wanting to use it with SSH can still do it already.

rmelotte:
I believe I addressed most of the issues, except for the "implement threads termination" commit, which would require a bit more time to be properly split in separate commits.
I'm marking the PR to ready, feel free to comment if you think that commit should be split anyway.

tools/beerocks_analyzer/connectivity_map_widget.py
```diff
@@ -195,11 +195,11 @@ def add_delta_x_from_node(self, n, delta):
     
     def update_x_positions_from_node(self, n):
         #check if n has clients connected
-        num_of_successors =len(self.graph.successors(n))
+        num_of_successors =len(list(self.graph.successors(n)))
```
arnout: The proper way to do this is actually `num_of_successors = self.graph.out_degree(n)`.

tools/beerocks_analyzer/beerocks_analyzer_widget.py
```diff
@@ -19,8 +19,7 @@
...
 import matplotlib.pyplot as plt
 from matplotlib.backends.backend_qt4agg import FigureCanvasQTAgg as FigureCanvas
```
arnout, line 24: Shouldn't this be changed into qt5 as well? It does work because the qt4 backend just forwards to the qt5 backend, but we can expect that to be removed from matplotlib at some point...

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -734,9 +738,11 @@ def local_start_beerocks_cli(beerocks_cli_path: Path, target_ip: IPv4Address) ->
...
     if not target_ip:
         target_ip = ip_address("127.0.0.1")
-
-    command = str(beerocks_cli_path) + " -a " + str(target_ip)
-    logger.debug("Starting beerock_cli -a")
+    command = ""
+    if docker_container_name:
+        command += "docker exec " + docker_container_name + " "
+    command += str(beerocks_cli_path) + " -a " + str(target_ip)
+    logger.debug(command)
```
arnout: The original debug message was a bit better, so I'd make this `logger.debug("Starting '%s'", ' '.join(command))` (taking into account that I think command should be a list instead of a string).

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -870,8 +898,7 @@ def main(argv):
         g_run_flag=False
 
         for t in t_list:
-            logger.debug("Joining thread {}" + str(t))
-            t.join()
+            t.terminate()
```
arnout: This could be factored with the KeyboardInterrupt, i.e. just do the terminate after the except clause.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -632,147 +629,176 @@ def show():
...
-                             "Stopping".format(str(e)))
-                run_local=False
-                break
-            while g_run_flag:
+class SocketServerThread(threading.Thread):
+
+    def __init__(self, log_file):
+        super().__init__()
+        self.log_file = log_file
+        self.run_local = True
```
arnout: Since all paths that set g_run_flag also call this thread's terminate function, you can in fact eliminate g_run_flag.

Also, I think run_flag or continue_running or something like that is more appropriate than run_local. run_local doesn't tell me what it's supposed to do.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -632,147 +629,176 @@ def show():
...
-            except Exception as e:
-                logger.error("Error when handling data from the socket_server:\n"
-                             "{}\n"
-                             "Stopping".format(str(e)))
-                run_local=False
-                break
-            while g_run_flag:
+class SocketServerThread(threading.Thread):
+
+    def __init__(self, log_file):
```
arnout: In other commits you added typeinfo to new functions. Would be nice to do that here as well.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -624,129 +629,203 @@ def show():
...
+        command = ""
+        if self.docker_container_name:
+            command += "docker exec " + self.docker_container_name + " "
+        command += str(self.beerocks_cli_path) + " -a " + str(self.target_ip)
+        self.logger.debug(command)
+        self.process = subprocess.Popen(command.split())
+
+    def terminate(self):
+        self.logger.debug("{} terminating".format(self.__class__.__name__))
+        self.process.kill()
```
arnout: It would make sense to add a `if self.process` here, in case terminate gets called before the subprocess is even started. But that brings us to the annoying limitation in Python that there's still a race between the creation of the Popen object and its assignment to self.process...

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -624,129 +629,203 @@ def show():
...
+        super().__init__()
+        self.logger = logging.getLogger(__name__)
+        self.host = host
+        self.my_ip = my_ip
+        self.ssh_port = ssh_port
+        self.user = user
+        self.ssh_client = paramiko.SSHClient()
+
+    def run(self):
+        self.ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
arnout: Since paramiko is not that well maintained, maybe it would be a good idea to switch to pxssh in pexpect. But that's not really necessary for this series.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -624,129 +629,203 @@ def show():
...
+        command = 'LD_LIBRARY_PATH=/opt/beerocks/lib /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)
+
+        #create ssh shell
+        ssh_shell = self.ssh_client.invoke_shell()
+
+        #establish connection
+        in_buff = ''
+        while not in_buff.endswith(':~# '):
+            in_buff += ssh_shell.recv(9999).decode("utf-8")
+            if in_buff.endswith('password: '):
```
arnout: If support for passwords is removed, this condition can be removed as well.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -624,129 +629,203 @@ def show():
...
+                time.sleep(3)
+                in_buff = ''
+            elif 'are you sure you want to continue connecting' in in_buff:
```
arnout: With pxssh, this is handled automatically.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -624,129 +629,203 @@ def show():
...
+        self.ssh_port = ssh_port
+        self.user = user
+        self.ssh_client = paramiko.SSHClient()
+
+    def run(self):
+        self.ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+        self.ssh_client.connect(self.host, port=self.ssh_port, username=self.user)
+
+        if not self.my_ip:
+            self.my_ip = self.ssh_client.get_transport().sock.getsockname()[0]
```
arnout, line 815: Hm, pxssh unfortunately doesn't have an equivalent for this... But you could use the ssh options to create a tunnel, so that you'd always tell beerocks_cli to connect to a local port.

But then, it would probably be better to interact directly over the ssh connection instead of creating an additional connection.

tools/beerocks_analyzer/.gitignore
```diff
@@ -0,0 +1,272 @@
+
+# Created by https://www.gitignore.io/api/vim,emacs,python,pycharm
+# Edit at https://www.gitignore.io/?templates=vim,emacs,python,pycharm
```
arnout: That website is *evil*. Or you're not using it correctly. Quoting the gitignore man page:

> Which file to place a pattern in depends on how the pattern is meant to be used.
> 
> - Patterns which should be version-controlled and distributed to other repositories via clone (i.e., files that all developers will want to ignore)  should go into a .gitignore file.
> - Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the $GIT_DIR/info/exclude file.
>  - Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

I think the only ignore pattern that is relevant here is `__pycache__/`. And that pattern should go to tools/.gitignore, which already has `*.pyc`. Well, actually, most of the gitignore files should be hoisted up to the top-level, but that's a different issue :-)

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
+Warning: there still is an enormous list of issues with `beerocks_analyzer`, and only the connection map currently works. This README will walk you through installing dependencies and displaying the connection map.
```
arnout: Please also update the top-level README.md to direct to this specific documentation.

Also, please follow the coding style for markdown files: one line per sentence.

Finally, please add a H1 header to act as a title:

```markdown
# beerocks analyzer
```

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+# Prerequisites
```
arnout: Use H2 (`##`) for sectioning.

Hm, maybe we should add that to the coding style...

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+You need to install a few python packages before using it. This readme will walk you through installing them in a `virtualenv`. Using a virtualenv allows to install the dependencies in a local directory, which avoids messing with the local Python installation.
+
+You only need to create the `venv` once. Later on you just need to activate the `venv` to be able to use the application.
+
+
+## Create a virtualenv to use the application
+
+```sh
+python3 -m venv venv
```
arnout: Ah, this explains why the venv stuff was added to .gitignore. But you may want to call venv from the top-level directory, so it should be added to the top-level .gitignore.

However, wouldn't it be better to give the venv a more specific name, e.g. `analyzer`?

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+The analyzer currently connects to a running controller, then from there opens a socket back to the analyzer. This means that you need to have access to the controller (locally, via docker, or via ssh) but that the controller also needs to be able to access the device on which the analyzer is running.
+
+
+## Using a controller running locally
+
+You need to give the path to beerock's binaries (change `BIN_PATH` according to your installation):
+
+```sh
+BIN_PATH="/home/$USER/prplmesh_root/build/install/bin/beerocks_cli"
```
arnout: Actually, we assume in dozens of other places that the install directory is `../build/install` compared to the top-level of prplMesh, so we could do the same in beerocks_analyzer.py. I.d., calculate the default bin_path based on the location of the script:
```python
script_dir = os.path.dirname(os.path.abspath(sys.argv[0]))
top_dir = os.path.abspath(os.path.join(script_dir, '..', '..'))
bin_path = os.path.abspath(os.path.join(top_dir, '..', 'build', 'install', 'bin', 'beerocks_cli'))
```

And maybe the abspath and intermediate top_dir are not even needed.

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+python beerocks_analyzer.py -map -bin_path="$BIN_PATH" -docker_container=gateway
```
arnout: `-docker_container` is wrong here.

python is wrong, it should be python3. But if we fix the script's shebang, it's not even needed (i.e. use `/usr/bin/env python3` instead of `/usr/bin/python`).

With the above suggestion, this could be as simple as

```sh
./beerocks_analyzer.py -map`
```

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+
+## Using a controller in a local docker container
+
+You need to give the path to beerock's binaries (change `BIN_PATH` according to your installation), the gateway IP from the container point of view (change `MY_IP`, it probably starts with 172), and the container name:
+
+```sh
+BIN_PATH="/home/$USER/prplmesh_root/build/install/bin/beerocks_cli"
+MY_IP=172.19.0.1
```
arnout: You can get this address with `docker network`, like we do in some of the test scripts.

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+DOCKER_CONTAINER="gateway"
+python beerocks_analyzer.py -map -bin_path="$BIN_PATH" -docker_container="$DOCKER_CONTAINER" -my_ip="$MY_IP"
+```
+
+
+## Using a remote controller over SSH
+
+You first need to have a private key (one that can connect to the remote device) registered in your local agent, so that you don't get asked for a password.
+
+If you don't have one running, you can run an agent by doing `` eval `ssh-agent -s` ``. To register your key, just do `ssh-add`.
```
arnout: ssh-agent is only needed if the private key is password protected. I would assume that people who have a password-protected ssh key will also have an agent running already.

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+BIN_PATH="/home/$USER/prplmesh_root/build/install/bin/beerocks_cli"
+MY_IP=172.19.0.1
+DOCKER_CONTAINER="gateway"
+python beerocks_analyzer.py -map -bin_path="$BIN_PATH" -docker_container="$DOCKER_CONTAINER" -my_ip="$MY_IP"
+```
+
+
+## Using a remote controller over SSH
+
+You first need to have a private key (one that can connect to the remote device) registered in your local agent, so that you don't get asked for a password.
```
arnout: Could you update the Turris Omnia wiki page with information on how to add your public key to the device?

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+If you don't have one running, you can run an agent by doing `` eval `ssh-agent -s` ``. To register your key, just do `ssh-add`.
+
+To run the analyzer, you then need to give the path to beerock's binaries (change `BIN_PATH` according to your installation) and the target IP:
+
+```sh
+BIN_PATH="/home/$USER/prplmesh_root/build/install/bin/beerocks_cli"
```
arnout: That doesn't seem right...

arnout (changes requested):
I still have a bunch of comments, and some of these are actually important.

I don't want to rush the inclusion of beerocks_analyzer, people can always use the branch if they need to.

The switch to pxssh is clearly something that doesn't need to be done in this PR. Could you create a feature request issue for it? You can also mark it as "good first issue".

tools/beerocks_analyzer/beerocks_analyzer.py
rmelotte: Do you have a link explaining that limitation ?
I added the `if self.process` condition, but I would like to add a note about the race condition.

tools/beerocks_analyzer/beerocks_analyzer.py
rmelotte: We're using a temporary fork of paramiko that is not maintained, but once [this PR](https://github.com/paramiko/paramiko/pull/1343) is merged, we will be able to switch to upstream paramiko again (which is still maintained as of today).

tools/beerocks_analyzer/beerocks_analyzer.py
rmelotte: After looking at it, there is actually a way to handle this in paramiko by either auto-adding the key or raising a warning. I went for the warning.

tools/beerocks_analyzer/beerocks_analyzer.py
rmelotte: Even without using pxssh, avoiding the additional connection would be an improvement, I'll open an issue about that.

tools/beerocks_analyzer/beerocks_analyzer_widget.py
```diff
@@ -488,7 +504,7 @@ def addRemoveNewFigSubplots(self, ar_t, ar_v, target_lable, is_add, entity, enti
                         else:
                             color = self.getLineColor(entity, entity_num)
                             if color == None:
-                                PRINTF("Error, color == None, param_n=%s\n", param_n)
+                                self.logger.error("Error, color == None, param_n={}".format(param_n))
                                 return
                             marker = self.marker[0]
                             if self.fig_subplot_nums2marker.has_key((f,p)):
```
arnout, line 510: There are still some instances of has_key let in beerocks_analyzer_widget.py.

(nitpick)

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -624,129 +628,238 @@ def show():
...
+
+        self.logger.info("Controller will send update to IP {}".format(self.my_ip))
+        # TODO: add a cli option for the path
+        command = 'LD_LIBRARY_PATH=/opt/beerocks/lib /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)
+
+        #create ssh shell
+        ssh_shell = self.ssh_client.invoke_shell()
```
arnout, line 822: It's probably possible to tell paramiko to start beerocks_cli directly instead of starting the shell. But then the LD_LIBRARY_PATH issue has to be solved first. (nitpick)

tools/beerocks_analyzer/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Note that currently, the binary path is hard-coded to `/opt/beerocks/bin/beerocks_cli`.
+
+
+## Troubleshooting
+
+
+### `Path to beerocks_cli not found`
+
+If you get `ValueError: Path to beerocks_cli not found`, it means the analyzer couldn't find the path to beerocks<sub>cli</sub>'s binary.
```
arnout: `beerocks<sub>cli</sub>`: I guess the org -> md conversion is not perfect :-)

tools/beerocks_analyzer/doc/README.org
```diff
@@ -0,0 +1,71 @@
+#+OPTIONS: toc:nil \n:t
```
arnout: I don't think it makes a lot of sense to include a .org in the repository, since a .md works just as well and that's what we standardised on. So I removed it. Feel free to put it back if you feel it adds value (but then defend why :-).

arnout (approved):
Just nitpicks, and I fixed up a few things myself. I'm going to merge now. If anything more needs to be changed, it will have to be in follow-up commits.

tools/beerocks_analyzer/beerocks_analyzer_widget.py
rmelotte: This could have been fixed at the same time indeed.
Just a note however, beerocks_analyzer_widget.py is not used for the connection map and still needs a its own lot of fixes.

tools/beerocks_analyzer/README.md
rmelotte: The conversion was correct, but the org file wasn't ('_' is used as subscripts like with LaTeX).

PR #312, opened 2019-09-23
tools/docker: add a route to the gateway

rmelotte:
Add a route to the gateway so that beerocks_cli is able to
communicate with a local beerocks_analyzer.

Signed-off-by: Raphaël Mélotte <raphael.melotte@essensium.com>

### DIFF
--- a/tools/docker/run.sh
+++ b/tools/docker/run.sh
@@ -36,9 +36,12 @@ iprand(){
     echo "$ip_rand"
 }
 
+gateway_ip() {
+    echo $(docker network inspect "$1" | jq -r '.[0].IPAM.Config'[0].Gateway)
+}
+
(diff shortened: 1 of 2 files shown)


rmelotte:
Note that the dockerhub images needs to be updated when this is merged or the run script
won't be compatible with the docker image.

morantr (approved):

morantr:
I approved it, but I prefer that someone else will review it as well.

tomereli (approved):

rmelotte:
As Arnout pointed out, docker containers currently use '/32' IPs. By giving them a '/16' instead, they would, by default, get a route for the whole subnet on which the gateway is. 
I'm closing this PR in favor of Arnout's solution.

Issue #309, opened 2019-09-19
[BUG] wf_ca: non hex TLV value not handled in DEV_SEND_1905


kantera800:
**Describe the bug** -
Currently in code when  handling DEV_SEND_1905 capi command  we expect TLV_Value only from the  following  format "0x001E231A67DF".
 But , while running test 5.4.3 via UCC we can see that DEV_SEND_1905 contains TLV_Value  in different  format "00:1E:23:1A:67:DF".

need to add support  for all t "hex" formats.
note:
Its not clear what format is supported , in "Wi-Fi_Test_Suite_Control_API_Specification_v10.7.0" spec written to support only hex format.
But, we can see also examples with other formats (dec, binary, hex.....) in the wifi test plan "Wi-Fi_CERTIFIED_EasyMesh_Test_Plan_v1.0"  .

PR #313

## Found 
while debugging  UCC test 5.4.3 

## Steps to reproduce

build prplmesh
.../prplmesh_utils.sh restart
../beerocks_cli -u 5000

from other terminal:
sudo nc <bridge_ip> 5000

dev_send_1905,DestALid,62:99:12:5f:b1:9e,MessageTypeValue,0x000A,tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,{46:55:66:77:00:00},tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{0x00},tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{0x00}

## current  behavior
from beerocks_controller.log

DEBUG 20:14:37:203 <140515358661504> wfa_ca.cpp[364] --> Received ACTION_BML_WFA_CA_CONTROLLER_REQUEST, command="dev_send_1905,DestALid,62:99:12:5f:b1:9e,MessageTypeValue,0x000A,tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,{46:55:66:77:00:00},tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{0x00},tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{0x00}"

ERROR 20:14:37:203 <140515358661504> wfa_ca.cpp[531] --> param name 'tlv_value1' has invalid hex notation value

## expected  behavior
--> received 1905.1 cmdu message

arnout:
Reassigned to Moran, he has more experience with this.

CoralMalachi:
#313

PR #313, opened 2019-09-23
controller:wfa_ca: handle non hex tlv_value

CoralMalachi:
Currently, the DEV_SEND_1905 CAPI command handler supports tlv_value only in
hex notation. But after checking via UCC,  it turned out that the
DEV_SEND_1905 contains tlv_value in different format "00:1E:23:1A:67:DF".

For that reason, add support to the required notations: binary, decimal and mac address.

#309 

### DIFF
--- a/controller/src/beerocks/master/wfa_ca.cpp
+++ b/controller/src/beerocks/master/wfa_ca.cpp
@@ -16,6 +16,7 @@
 #include <easylogging++.h>
 
 #include <algorithm>
+#include <iomanip>
 
 #define FORBIDDEN_CHARS "$#&*()[];/\\<>?|`~=+"
 
@@ -74,6 +75,50 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
         return false;
     }
 
+    return true;
+}
+
+/**
+ * @brief Checks that the given string is in binary notation
+ * 
+ * @param[in] str String to check binary notation.
+ * @return true if string is in valid binary notation, else false.
+ */
+static bool validate_binary_notation(const std::string &str)
+{
+    // Each value must start with 0b
+    if (str.substr(0, 2) != "0b") {
+        return false;
+    }
+
+    if ((str.size() - 2) % 8 != 0) {
+        return false;
+    }
+
+    // Check if string contains not only 0 and 1
+    if (str.find_first_not_of("01", 2) != std::string::npos) {
+        return false;
+    }
+    return true;
+}
+
+/**
+ * @brief Checks that the given string is in decimal notation
+ * 
+ * @param[in] str String to check decimal notation.
+ * @return true if string is in valid decimal notation, else false.
+ */
+static bool validate_decimal_notation(const std::string &str)
+{
+    // Each value must start with 0b
+    if (str.substr(0, 2) != "0d") {
+        return false;
+    }
+
+    // Check if string has non-decimal character
+    if (str.find_first_not_of("0123456789", 2) != std::string::npos) {
+        return false;
+    }
     return true;
 }
 
@@ -820,10 +865,11 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
 
                 // Validate hex notation on list of values separated by space
                 auto values = string_utils::str_split(it->second, ' ');
-                for (const auto &value : values) {
-                    if (!validate_hex_notation(value)) {
+                for (auto &value : values) {
+                    if (!(validate_hex_notation(value) || net::network_utils::is_valid_mac(value) ||
+                          validate_binary_notation(value) || validate_decimal_notation(value))) {
                         err_string =
-                            "param name '" + lookup_str + "' has invalid hex notation value";
+                            "param name '" + lookup_str + "' has value with invalid format";
                         return false;
                     }
                 }
@@ -852,6 +898,91 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
     LOG(ERROR) << "Reached stop condition";
     err_string = err_internal;
     return false;
+}
+
+/**
+ * @brief Writes hex string into the class buffer.
+ * 
+ * @param[in] value string in hex notation.
+ * @param[out] length Length of current tlv.
+ * @return false if length smaller than data, else true.
+ */
+bool tlvPrefilledData::add_tlv_value_hex_string(const std::string &value, uint16_t &length)
+{
+    // iterate every two chars (1 octet) and write it to buffer
+    // start with char_idx = 2 to skip "0x"
+    for (size_t char_idx = 2; char_idx < value.size(); char_idx += 2) {
+        if (length == 0) {
+            return false;
+        }
+        *m_buff_ptr__ = std::stoi(value.substr(char_idx, 2), nullptr, 16);
+        m_buff_ptr__++;
+        length--;
+    }
+    return true;
+}
+
+/**
+ * @brief Writes a decimal string into the class buffer.
+ * 
+ * @param[in] value Decimal string.
+ * @param[out] length Length of current tlv.
+ * @return false if length smaller than data, else true.
+ */
+bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
+{
+    // Convert to hex string
+    std::stringstream ss;
+    size_t num_of_bytes = 1;
+    // We don't really know how to identify if it is one or two (or more) bytes.
+    // Assume 2 bytes if it has more than 3 digits
+    if (value.size() > 5) {
+        num_of_bytes++;
+    }
+    ss << "0x" << std::setw(num_of_bytes * 2) << std::setfill('0') << std::hex
+       << std::stoi(value.substr(2, value.length()));
+
+    return add_tlv_value_hex_string(ss.str(), length);
+}
+
+/**
+ * @brief Writes a binary string into the class buffer.
+ * 
+ * @param[in] value Binary string.
+ * @param[out] length Length of current tlv.
+ * @return false if length smaller than data, else true.
+ */
+bool tlvPrefilledData::add_tlv_value_binary_string(const std::string &value, uint16_t &length)
+{
+    for (size_t char_idx = 2; char_idx < value.size(); char_idx += 8) {
+        if (length == 0) {
+            return false;
+        }
+        *m_buff_ptr__ = std::stoi(value.substr(char_idx, 8), nullptr, 2);
+        m_buff_ptr__++;
+        length--;
+    }
+    return true;
+}
+
+/**
+ * @brief Writes a mac string into the class buffer.
+ * 
+ * @param[in] value mac address.
+ * @param[out] length of current tlv.
+ * @return false if length smaller than data, else true.
+ */
+bool tlvPrefilledData::add_tlv_value_mac(const std::string &value, uint16_t &length)
+{
+    for (size_t char_idx = 0; char_idx < value.size(); char_idx += 3) {
+        if (length == 0) {
+            return false;
+        }
+        *m_buff_ptr__ = std::stoi(value.substr(char_idx, 2), nullptr, 16);
+        m_buff_ptr__++;
+        length--;
+    }
+    return true;
 }
 
 /**
@@ -885,16 +1016,27 @@ bool tlvPrefilledData::add_tlvs_from_list(std::list<wfa_ca::tlv_hex_t> &tlv_hex_
 
         auto values = string_utils::str_split(*tlv.value, ' ');
         for (auto value : values) {
-            // iterate every to chars (1 octet) and write it to buffer
-            // start with char_idx = 2 to skip "0x"
-            for (size_t char_idx = 2; char_idx < value.size(); char_idx += 2) {
-                if (length == 0) {
+            //Do conversion if needed
+            if (value[1] == 'x') {
+                if (!add_tlv_value_hex_string(value, length)) {
                     err_string = "length smaller than data";
                     return false;
                 }
-                *m_buff_ptr__ = std::stoi(value.substr(char_idx, 2), nullptr, 16);
-                m_buff_ptr__++;
-                length--;
+            } else if (value[2] == ':') {
+                if (!add_tlv_value_mac(value, length)) {
+                    err_string = "length smaller than data";
+                    return false;
+                }
+            } else if (value[1] == 'd') {
+                if (!add_tlv_value_decimal_string(value, length)) {
+                    err_string = "length smaller than data";
+                    return false;
+                }
+            } else if (value[1] == 'b') {
+                if (!add_tlv_value_binary_string(value, length)) {
+                    err_string = "length smaller than data";
+                    return false;
+                }
             }
         }
 
--- a/controller/src/beerocks/master/wfa_ca.h
+++ b/controller/src/beerocks/master/wfa_ca.h
@@ -80,6 +80,10 @@ public:
     void class_swap(){};
     static size_t get_initial_size() { return 0; };
 
+    bool add_tlv_value_hex_string(const std::string &value, uint16_t &length);
+    bool add_tlv_value_decimal_string(const std::string &value, uint16_t &length);
+    bool add_tlv_value_binary_string(const std::string &value, uint16_t &length);
+    bool add_tlv_value_mac(const std::string &value, uint16_t &length);
     bool add_tlvs_from_list(std::list<wfa_ca::tlv_hex_t> &tlv_hex_list, std::string &err_string);
 };
 


controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,44 @@ static std::string check_forbidden_chars(const std::string &str)
     return forbidden_chars_found;
 }
 
+/*
+* @brief convert a mac address to hex notation 
+* 
+* @param[in]  str String mac address
```
morantr: Unnecessary space between `param[in]` to `String mac address`

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,44 @@ static std::string check_forbidden_chars(const std::string &str)
...
+* @return - none
```
morantr: No need hyphen in Doxygen, and the return description doesn't start with capital letter.
`@return - none` -> `@return None`

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,44 @@ static std::string check_forbidden_chars(const std::string &str)
...
+*/
+static void convert_mac_address_to_hex_notation(std::string &str)
```
morantr: I would change it to` void mac_to_hex_notation(std::string &mac_string)` to make things short

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,44 @@ static std::string check_forbidden_chars(const std::string &str)
...
+    str.erase(std::remove(str.begin(), str.end(), ':'), str.end());
+    str = "0x" + str;
+}
+
+/*
+* @brief Checks that the given string represents a mac address 
+* 
+* @param[in]  str String mac address
+* @return - true if string represents a mac address, else false.
+*/
```
morantr: See comments on Doxygen from above. Same here.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,44 @@ static std::string check_forbidden_chars(const std::string &str)
...
+static bool validate_mac_address(const std::string &str)
+{
+    int i                = 0;
```
morantr: Actually you already have function that do the same on network utils:
network_utils::is_valid_mac()

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -810,12 +849,27 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
-                for (const auto &value : values) {
+                for (auto &value : values) {
                     if (!validate_hex_notation(value)) {
-                        err_string =
-                            "param name '" + lookup_str + "' has invalid hex notation value";
-                        return false;
+                        if (validate_mac_address(value)) {
+                            convert_mac_address_to_hex_notation(value);
+                            join = true;
+                        } else {
```
morantr: What you did is not efficient. This solution replaces all cmdu string content just because a single value is not in hex.
Please keep here only the validation with logic AND between them.
As well, more validation are needed:
```cpp
if (!(validate_hex_notation() || validate_mac_notation() || validate_dec_notation() || validate_bin_notation())) {
    //error
}
```
On `add_tlvs_from_list()` where we iterate each value again to put it on the buffer, validate for each of the notations, and do the right conversion.

morantr (changes requested):

kantera800:
please don't merge this PR until  verified  with  UCC testbed TC 5.4.3

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -115,6 +130,50 @@ static bool validate_hex_notation(const std::string &str, uint8_t expected_octet
     return true;
 }
 
+/**
+ * @brief Checks that the given string is in binary format
+ * 
+ * @param[in] str String to check binary format.
+ * @return true if string is in valid binary format, else false.
+ */
+static bool validate_binary_format(const std::string &str)
```
morantr: Since we use the term "notation" previously (validate_hex_notation), I think it is better to keep using it, instead of mixing the similar word "format".
Please replace the word "format" to "notation" in this function, and decimal validation function, and its doxygen description.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -44,41 +44,56 @@ static std::string check_forbidden_chars(const std::string &str)
...
+            n <<= 1;
+            if (binary_string[j] == '1')
+                n |= 1;
+        }
+        if (n <= 9)
+            out.push_back('0' + n);
+        else
+            out.push_back('A' + n - 10);
+    }
+    binary_string = "0x" + out;
```
morantr: Using loops like that is very unreadable, and takes a lot of effort to understand what you did here, also, it looks likes a C code instead of CPP.
I suggest to replace it with the following code:
```cpp
static void binary_to_hex(const std::string &binary_string)
  std::string output ("0x");
  uint8_t idx = 2;

  while (idx < str.length()) {
      // Read 8 bits
      auto binary_byte = binary_string.substr (idx, 8);
      
      // Convert binary byte to integer
      auto integer_byte = std::stoi(binary_byte, nullptr, 2);
      
      // Convert integer to hex string with 2 characters and add it to output
      output += string_utils::int_to_hex_string(integer_byte, 2);
      idx += 8;
  }

  binary_string = output;
}
```
It is more clear, short, and readable, and most important, CPP style.
Also, now the string argument is const reference.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -44,41 +44,56 @@ static std::string check_forbidden_chars(const std::string &str)
...
+* @return String in hex format
+*/
+static void decimal_to_hex(std::string &decimal_string)
+{
+    decimal_string.erase(0, 2);
+    int i_dec = std::stoi(decimal_string);
+    std::stringstream ss;
+    ss << std::hex << i_dec;
+    std::string res(ss.str());
+    decimal_string = "0x" + res;
```
morantr: The problem with this implementation is that it assumes that the decimal value is the only 1-byte size.
I think you need to use modulo by 255 to understand how many bytes the value has, and then convert it again with string_utils::int_to_hex_string(i_dec , 2 * num_of_bytes) to fill zeros from the left. You must do it because std::stoi() does not fill zeros from the left.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -929,6 +973,14 @@ bool tlvPrefilledData::add_tlvs_from_list(std::list<wfa_ca::tlv_hex_t> &tlv_hex_
 
         auto values = string_utils::str_split(*tlv.value, ' ');
         for (auto value : values) {
+            //Do conversion if needed
+            if (net::network_utils::is_valid_mac(value)) {
+                mac_to_hex_notation(value);
```
morantr: If the value is already hex, it is waste to do all these checks. Please do the check only if the value does not start with "0x".
Also, when thinking about it, there is no point of doing full validation again, because it has been validated before, so at this point, we know that the value is valid, so we only need to figure out which of them is the correct one.

```cpp
if (!value.find_first_of("x", 1, 1)) {
        if (value.find_first_of(":", 2, 1)) {
                mac_to_hex_notation(value);
        } else if  (value.find_first_of("b", 1, 1)) {
                binary_to_hex(value);
        } else if  (value.find_first_of("d", 1, 1)) {
                decimal_to_hex(value);
        }
}
```

morantr (changes requested):

controller/src/beerocks/master/wfa_ca.cpp
CoralMalachi: I liked your implementation,
but how can you do an assignment of const reference?

arnout: I think Moran meant to `return output` instead of setting `binary_string = output`.

arnout: Actually, it's better if all the conversion functions have a similar signature. So if this function is changed to return the output rather than taking an in-out argument, it's better to do that everywhere. Or if you prefer not to do that, it's better to keep it as an in-out argument here as well.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -875,6 +973,14 @@ bool tlvPrefilledData::add_tlvs_from_list(std::list<wfa_ca::tlv_hex_t> &tlv_hex_
...
+            } else if (validate_binary_format(value)) {
+                binary_to_hex(value);
+            } else if (validate_decimal_format(value)) {
+                decimal_to_hex(value);
+            }
```
arnout: I think it's pretty pointless to convert the string to another string and then convert it to uint8_t, while it is easier to directly convert it to uint8_t... So instead of mac_to_hex_notation, I think it's better to add a member function `tlvPrefilledData::add_tlv_value_mac(value)` that directly writes to `m_buff_ptr__`. The update of `length` can be done either by passing it as in-out, or by returning the size (i.e. calling it as `length -= add_tlv_value_mac(value)`).

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,59 @@ static std::string check_forbidden_chars(const std::string &str)
     return forbidden_chars_found;
 }
 
+/*
+* @brief Convert a mac address to hex notation 
+* 
+* @param[in] mac_string String mac address
+* @return Mac string in hex format
+*/
+static void mac_to_hex_notation(std::string &mac_string)
```
arnout: The other conversion functions don't have `_notation` at the end.

arnout (changes requested):

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -844,6 +937,27 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
+bool tlvPrefilledData::add_tlv_value_mac(std::string &value, uint16_t &length)
+{
+    for (size_t char_idx = 0; char_idx < value.size(); char_idx += 2) {
+        if (length == 0) {
+            return false;
+        }
+        *m_buff_ptr__ = std::stoi(value.substr(char_idx, 2), nullptr, 16);
+        m_buff_ptr__++;
+        length--;
+        char_idx++;
```
arnout: It is really weird to update char_idx here *and* in the loop update. Is there a reason for that?

controller/src/beerocks/master/wfa_ca.cpp
arnout: I may not have been sufficiently clear in my earlier remark... I only mentioned `add_tlv_value_mac`, but I meant the same thing for the decimal and binary cases.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -43,6 +43,54 @@ static std::string check_forbidden_chars(const std::string &str)
...
+    decimal_string.erase(0, 2);
+
+    int i_dec        = std::stoi(decimal_string);
+    int rem          = i_dec % 255;
+    int num_of_bytes = i_dec / 255;
+    if (rem != 0) {
+        num_of_bytes++;
+    }
+
+    output += string_utils::int_to_hex_string(i_dec, num_of_bytes);
```
arnout: The decimal case is tricky... For hex and binary, you can know the length from the 0-padded number of characters. For decimal, this doesn't really work though. Therefore, I assume that the decimal case is only used in practice for 1-byte values. Is that possible?

arnout (changes requested):

controller/src/beerocks/master/wfa_ca.cpp
CoralMalachi: You're right it was a temporary change, I'll change that

controller/src/beerocks/master/wfa_ca.cpp
CoralMalachi: I thought so but I wasn't sure, I'll add that.

controller/src/beerocks/master/wfa_ca.cpp
CoralMalachi: I have seen at least one case of a decimal number that requires more than one byte for storage, e.g., 0d0600  in test 4.8.5 in Test Plan.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -844,6 +905,70 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
+ * @brief Writes a decimal string into the class buffer.
+ * 
+ * @param[in] value Decimal string.
+ * @param[out] length Length of current tlv.
+ * @return false if length smaller than data, else true.
+ */
+bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
+{
+    int i_dec        = std::stoi(value.substr(2, value.length()));
+    int num_of_bytes = number_of_bytes(i_dec);
```
morantr: As I have said [here](https://github.com/prplfoundation/prplMesh/pull/313#discussion_r328571565), why didn't you use divide and modulo? 
```cpp
num_of_bytes  = i_dec / UINT8_MAX + ( i_dec % UINT8_MAX != 0 ? 0 : 1);
```
This is one simple line, and you can remove `number_of_bytes()`.

morantr: As I have said here, why didn't you use divide and modulo?

num_of_bytes  = i_dec / UINT8_MAX + (i_dec % UINT8_MAX != 0 ? 1 : 0);
This is one simple line, and you can remove number_of_bytes().

morantr (changes requested):
I'm not sure that I like what @arnout has suggested [here](https://github.com/prplfoundation/prplMesh/pull/313#discussion_r329157184) because it makes some code duplication of the writing to the buffer.
Oh well, I guess it is ok. But if doing it like that it will be a good idea IMO to do similar function to the regular hex string notation and handle it like other notation.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -844,6 +905,70 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
+ * @param[out] length of current tlv.
+ * @return false if length smaller than data, else true.
+ */
+bool tlvPrefilledData::add_tlv_value_mac(const std::string &value, uint16_t &length)
+{
+    for (size_t char_idx = 0; char_idx < value.size(); char_idx += 3) {
+        if (length == 0) {
+            return false;
+        }
+        *m_buff_ptr__ = std::stoi(value.substr(char_idx, 2), nullptr, 16);
```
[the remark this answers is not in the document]
CoralMalachi: why? I want to skip the colons.

CoralMalachi:
> I'm not sure that I like what @arnout has suggested [here](https://github.com/prplfoundation/prplMesh/pull/313#discussion_r329157184) because it makes some code duplication of the writing to the buffer.
> Oh well, I guess it is ok. But if doing it like that it will be a good idea IMO to do similar function to the regular hex string notation and handle it like other notation.

I agree with you. 
I'll add a function to handle that case.

controller/src/beerocks/master/wfa_ca.cpp
arnout: > I have seen at least one case of a decimal number that requires more than one byte for storage, e.g., 0d0600 in test 4.8.5 in Test Plan.

Darn. So it means that this code has to stay. But it's insane... for 0d0600 it happens to work out OK with 2 bytes, but what would you do with 0d0200? It would end up as one byte... Whoever thought up this CAPI spec wasn't very awake...

morantr (approved):

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -844,6 +889,92 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
+ * @brief Writes a decimal string into the class buffer.
+ * 
+ * @param[in] value Decimal string.
+ * @param[out] length Length of current tlv.
+ * @return false if length smaller than data, else true.
+ */
+bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
+{
+    int i_dec        = std::stoi(value.substr(2, value.length()));
+    int num_of_bytes = i_dec / UINT8_MAX + (i_dec % UINT8_MAX != 0 ? 1 : 0);
```
arnout: Sorry, @morantr, but this is incorrect... First of all, UINT8_MAX is 255, not 256. More importantly, though, we need log<sub>256</sub>(i_dec) but you're calculating i_dec/256 rounded up...

That said, as I mentioned in previous reviews, this `0d` stuff is utterly broken anyway. So I think we could make the assumption that it's either 1 or 2 bytes. Also, I think we should do it based on the number of digits. So:
```suggestion
    // Assume 2 bytes if it has more than 3 digits
    if (value.size() > 5) {
        *m_buff_ptr__ = i_dec / 256;
        m_buff_ptr__++;
        length--;
    }
    *m_buff_ptr__ = i_dec % 256;
    m_buff_ptr__++;
    length--;
```

This is based on the example that @CoralMalachi found of a `0d0600` value in one of the test cases. Note that I don't really like relying on the i_dec value anyway, because it is possible that they use something like `0d0005` string somewhere for a two-byte value...

But that means, of course, that if we ever find any sample where `0d` is used for 3- or 4-byte values, we have to change this again... So to be a bit more prepared for that, we could still set `num_of_bytes` (to one or two, depending on `value.size()`) and keep the loop.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -844,6 +889,92 @@ bool wfa_ca::get_send_1905_1_tlv_hex_list(std::list<tlv_hex_t> &tlv_hex_list,
...
+    std::string s    = string_utils::int_to_hex_string(i_dec, num_of_bytes);
```
arnout: The idea was to avoid converting back to string...

If you do convert back to string, at least you should reuse the `add_tlv_value_hex_string` function (so prepend `0x`).

But I'd rather to see it done explicitly here (which is a bit more difficult, I admit, because it's big-endian).

```suggestion
    for (size_t idx = 0; idx < num_of_bytes; idx++) {
        m_buff_ptr__[num_of_bytes - 1 - idx] = i_dec % 256;
        i_dec /= 256;
    }
    m_buff_ptr__ += num_of_bytes;
    length -= num_of_bytes;
```

Or, if you want to avoid the inductive approach

```suggestion
    for (size_t idx = 0; idx < num_of_bytes; idx++) {
        *m_buff_ptr__ = i_dec / (1<<(8*idx));
        m_buff_ptr__++;
        length--;
    }
```

arnout (changes requested):

controller/src/beerocks/master/wfa_ca.cpp
morantr: @arnout, yes I can see now that what I suggested will work only on decimals
between 0 - 509
:(

morantr:
Regarding the conversion from decimal to hex string.
What do you think about this:
```cpp
bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
{
    // Convert to hex string
    std::stringstream ss;
    ss<< std::hex << std::stoi(value.substr(2, value.length()));
    std::string hex_string(ss.str());

    // Previous conversion to hex could miss one zero at the begining, so fill it if needed.
    if (hex_string.size() % 2 != 0) {
         hex_string.insert(0, "0");
    }

    hex_string.insert(0, "0x");
    
    return add_tlv_value_hex_string(hex_string, length);
}
```

morantr (changes requested):
Regarding the conversion from decimal to hex string.
What do you think about this:
```cpp
bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
{
    // Convert to hex string
    std::stringstream ss;
    ss<< std::hex << std::stoi(value.substr(2, value.length()));
    std::string hex_string(ss.str());

    // Previous conversion to hex could miss one zero at the begining, so fill it if needed.
    if (hex_string.size() % 2 != 0) {
         hex_string.insert(0, "0");
    }

    hex_string.insert(0, "0x");
    
    return add_tlv_value_hex_string(hex_string, length);
}
```

CoralMalachi:
> Regarding the conversion from decimal to hex string.
> What do you think about this:
> 
> ```c++
> bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
> {
>     // Convert to hex string
>     std::stringstream ss;
>     ss<< std::hex << std::stoi(value.substr(2, value.length()));
>     std::string hex_string(ss.str());
> 
>     // Previous conversion to hex could miss one zero at the begining, so fill it if needed.
>     if (hex_string.size() % 2 != 0) {
>          hex_string.insert(0, "0");
>     }
> 
>     hex_string.insert(0, "0x");
>     
>     return add_tlv_value_hex_string(hex_string, length);
> }
> ```
I like your implementation @morantr , what do you think ? @arnout

arnout:
> > Regarding the conversion from decimal to hex string.
> > What do you think about this:
> > ```c++
> > bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
> > {
> >     // Convert to hex string
> >     std::stringstream ss;
> >     ss<< std::hex << std::stoi(value.substr(2, value.length()));
> >     std::string hex_string(ss.str());
> > 
> >     // Previous conversion to hex could miss one zero at the begining, so fill it if needed.
> >     if (hex_string.size() % 2 != 0) {
> >          hex_string.insert(0, "0");
> >     }
> > 
> >     hex_string.insert(0, "0x");
> >     
> >     return add_tlv_value_hex_string(hex_string, length);
> > }
> > ```
> 
> I like your implementation @morantr , what do you think ? @arnout

I think it's weird to convert to int then to string and then to bytes again, but OK.

However, I still believe it's important to take into account the leading zeroes in the decimal string. It is possible that we get a `0d0040` which is meant to be two bytes. So I still think a `num_of_bytes` variable will be needed. And if we anyway have one, then I think there's not much benefit of converting to string again - it's easier to understand what's going on if it is a loop.

Bottom line: I'm not strictly opposed to the conversion to hex string and reusing add_tlv_value_hex_string(), but we need to have a number of bytes based on the length of the decimal string.

arnout (approved):

arnout:
> > > Regarding the conversion from decimal to hex string.
> > > What do you think about this:
> > > ```c++
> > > bool tlvPrefilledData::add_tlv_value_decimal_string(const std::string &value, uint16_t &length)
> > > {
> > >     // Convert to hex string
> > >     std::stringstream ss;
> > >     ss<< std::hex << std::stoi(value.substr(2, value.length()));
> > >     std::string hex_string(ss.str());
> > > 
> > >     // Previous conversion to hex could miss one zero at the begining, so fill it if needed.
> > >     if (hex_string.size() % 2 != 0) {
> > >          hex_string.insert(0, "0");
> > >     }
> > > 
> > >     hex_string.insert(0, "0x");
> > >     
> > >     return add_tlv_value_hex_string(hex_string, length);
> > > }
> > > ```
> > 
> > 
> > I like your implementation @morantr , what do you think ? @arnout
> 
> I think it's weird to convert to int then to string and then to bytes again, but OK.
> 
> However, I still believe it's important to take into account the leading zeroes in the decimal string. It is possible that we get a `0d0040` which is meant to be two bytes. So I still think a `num_of_bytes` variable will be needed. And if we anyway have one, then I think there's not much benefit of converting to string again - it's easier to understand what's going on if it is a loop.
> 
> Bottom line: I'm not strictly opposed to the conversion to hex string and reusing add_tlv_value_hex_string(), but we need to have a number of bytes based on the length of the decimal string.

By the way, if you use the conversion back to string, then you can do it in one shot within the stringstream, by using `std::showbase` for the `0x` and `std::setw` and `std::setfill` for 0-filling.

arnout:
One more thing, @CoralMalachi please add a comment to the decimal conversion function that we don't really know how to identify if it is one or two (or more) bytes. If we ever encounter this problem in the future, we should have a hint in the source code as to what is going on.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -920,23 +920,25 @@ bool tlvPrefilledData::add_tlv_value_hex_string(const std::string &value, uint16
...
-        num_of_bytes++;
+    /*
+    We don't really know how to identify if it is one or two (or more) bytes.
+    If you encounter this problem now, go back to the source code as a hint to
+    what is going on.
+    */
+
+    // Convert to hex string
+    std::stringstream ss;
+    ss << std::hex << std::stoi(value.substr(2, value.length()));
```
arnout: Again: we *know* this is wrong, because there are cases where there's a 0d with leading zeroes. For example, in the test specification 5.8.3, we have:

> Trigger (DEV_SEND_1905, DestALid,
> WTS_REPLACE_CTTAGENT1_ALID,
> MessageTypeValue,0x8016,
> tlv_type,0x9D,tlv_length,0xXXXX,tlv_value,{
> WTS_REPLACE_CTTAGENT1_BSSID
> 0x00 ***0d0030*** 0x01
> {WTS_REPLACE_CTT_STA_MAC} })
>
> MCUT to send Client Association Control
> Request message to CTT Agent1 with:
>
> - [...]
> - Validity Period field set to 30 seconds
> - [...]

(my emphasis). The validity period `0d0030` has to be two bytes in the TLV, but the above conversion to string will turn it into `0x1e`.

So, if you do want to convert it to string, it has to be:

```
    size_t num_of_bytes = 1;
    // We don't really know how to identify if it is one or two (or more) bytes.
    // Assume 2 bytes if it has more than 3 digits
    if (value.size() > 5) {
        num_of_bytes++;
    }
    ss << "0x" << std::setw(num_of_bytes * 2) << std::setfill('0') << std::hex 
       << std::stoi(value.substr(2, value.length()));
```

(This includes my earlier suggestion of including the additional formatting in the strstream output. So all the code below can be removed.)

arnout (changes requested):

arnout (approved):

morantr (approved):

PR #314, opened 2019-09-24
controller: client steering task

YuriyMasechko:
Client steering task should use the appropriate EasyMesh CMDU.

Change vendor speciefic message ACTION_CONTROL_CLIENT_BSS_STEER_REQUEST
to EasyMesh CLIENT_STEERING_REQUEST_MESSAGE.

### DIFF
--- a/controller/src/beerocks/master/tasks/client_steering_task.cpp
+++ b/controller/src/beerocks/master/tasks/client_steering_task.cpp
@@ -13,6 +13,7 @@
 
 #include <beerocks/bcl/network/network_utils.h>
 #include <easylogging++.h>
+#include <tlvf/wfa_map/tlvSteeringRequest.h>
 
 using namespace beerocks;
 using namespace net;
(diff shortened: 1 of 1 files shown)


controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -174,42 +175,51 @@ void client_steering_task::steer_sta()
...
+    if (!steering_request_tlv) {
+        LOG(ERROR) << "addClass wfa_map::tlvSteeringRequest failed";
+        return;
     }
 
     auto current_ap_mac = database.get_node_parent(sta_mac);
 
+    steering_request_tlv->request_flags().request_mode =
+        wfa_map::tlvSteeringRequest::REQUEST_IS_A_STEERING_MANDATE_TO_TRIGGER_STEERING;
+    steering_request_tlv->request_flags().btm_disassociation_imminent_bit = !!disassoc_imminent;
```
arnout: Excellent idea to use `!!` to make sure it really is 0 or 1. However, in C++ this is not needed: conversion to bool will do that already.

(nitpick)

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -174,42 +175,51 @@ void client_steering_task::steer_sta()
...
+
+    steering_request_tlv->btm_disassociation_timer() = disassoc_timer;
```
arnout, line 197: This is not correct. The disassoc_timer given to this function is in units of 100ms, while the btm_disassociation_timer is in units of 1TU (1024µs).

I think we should convert the uses of the disassociation timer to use TUs everywhere since that's what goes in the BTM request anyway. The bml and cli don't need to be changed. #323

arnout (dismissed):
Excellent PR. Small typo in commit message which I'll fix up while rebasing.

arnout (approved):

Issue #317, opened 2019-09-24
[TASK] agent: handle Client Access Control message


arnout:
## Description

After #316, the controller will send Client Association Control messages instead of vendor specific. The agent should properly handle these messages.

## Exit criteria

* Agent fully validates the Client Association Control TLV.
* Agent emits the appropriate bwl command(s) according to Cient Association Control.
* If a vendor-specific TLV was added in #316, it is handled as well.
* Dummy bwl is updated to handle these commands (by putting something in the log).
* test_flows.sh is updated to check for this dummy bwl output instead of agent output.
* cACTION_CONTROL_CLIENT_ALLOW_REQUEST and cACTION_CONTROL_CLIENT_DISALLOW_REQUEST are removed completely.

arnout:
Fixed by #319

PR #319, opened 2019-09-25
Controller: use EasyMesh messages for access control

### LINKED ISSUE #316: [TASK] controller: use EasyMesh messages for access control
## Description

The beerocks client steering task uses a combination of client access control and client steering. Client steering itself is handled in #297. This task is to make sure the client steering task uses Client Association Control Request messages instead of vendor specific messages.

## Exit criteria

* Controller client steering task uses standard Client Association Control Request message.
* If any information is generated by client steering task that is not included in the standard Client Association Control Request message, a new vendor-specific TLV is defined to include this information.

CoralMalachi:
The client steering task should use the appropriate EasyMesh CMDU.

Change client steering task to use Client Association Control request
messages instead of vendor-specific messages.

Fixes #316, #317 


controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -92,20 +93,36 @@ void client_steering_task::steer_sta()
...
-    request->mac()  = network_utils::mac_from_string(sta_mac);
-    request->ipv4() = network_utils::ipv4_from_string(database.get_node_ipv4(sta_mac));
-    Socket *sd      = database.get_node_socket(radio_mac);
-    TASK_LOG(DEBUG) << "sending allow request for " << sta_mac << " to " << radio_mac
-                    << " id=" << int(id);
+
+    association_control_request_tlv->alloc_sta_list();
+    auto sta_list         = association_control_request_tlv->sta_list(0);
+    std::get<1>(sta_list) = network_utils::mac_from_string(sta_mac);
+    association_control_request_tlv->bssid_to_block_client() =
```
arnout, line 114: Please keep the same order as in the TLV, so first set bssid, then association control, then validity period, and only at the end allocate the station list.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -92,20 +93,36 @@ void client_steering_task::steer_sta()
...
+        network_utils::mac_from_string(current_ap_mac);
+    association_control_request_tlv->validity_period_sec() = disassoc_timer;
+    association_control_request_tlv->association_control() =
+        wfa_map::tlvClientAssociationControlRequest::BLOCK;
```
arnout: This  should be an UNBLOCK. The blocking is done below (cACTION_CONTROL_CLIENT_DISALLOW_REQUEST). That one should also be updated.

arnout (changes requested):
Typo in the second commit message: Associan -> Association.

As advised by @morantr, we should only merge the PR when it is complete, i.e.:

- controller uses the message;
- agent handles the message;
- dummy bwl reports it;
- a test is added for this flow, by issuing a bml command that triggers it, and verifying the bwl log message on the agent.
- VS handling is removed from agent;
- action is removed.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_1905_vs.yaml
```diff
@@ -35,3 +35,9 @@ tlvVsClientAssociationEvent:
   disconnect_type: 
     _type: uint8_t
     _comment: relevant only on disconnect event
+
+
+tlvVsClientAssociationControlRequest:
+  _type: class
+  reject_sta: uint8_t
```
arnout, line 42: reject_sta is only used by the wav and dwpal BWL implementations. However, upstream hostapd doesn't have anything like that, and I don't really understand what its purpose could be.

That said, the Multi-AP specification does say what the reject reason should be:

>  it shall set the Status Code in the Authentication frame or (Re-)Association Response frame to a value that
> does not indicate capabilities mismatch or negotiation failure, and should set the Status Code to value 33 (denied due to
> insufficient bandwidth) or 34 (denied due to poor channel conditions).

So the 33 *does* make sense. However:

- it should always be 33, not 0;
- it's the agent's responsibility to take care of this.

So I think we should remove reject_sta here, and instead handle it inside BWL itself.

Of course, that's not for this PR... -> #430

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_1905_vs.yaml
```diff
@@ -35,3 +35,9 @@ tlvVsClientAssociationEvent:
...
+  ipv4: beerocks::net::sIpv4Addr
```
arnout, line 43: ipv4 is not used for anything except a log message, so I don't think we need to add it. nitpick, can be solved later.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4780,6 +4733,44 @@ bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::Cmd
...
+        return false;
+    }
+    auto sta_mac = std::get<1>(association_control_request_tlv->sta_list(0));
+
+    //Add VS tlv
+    auto vs_tlv =
+        message_com::add_vs_tlv<beerocks_message::tlvVsClientAssociationControlRequest>(cmdu_tx);
+    if (!vs_tlv) {
+        LOG(ERROR) << "add_vs_tlv tlvVsClientAssociationControlRequest failed";
+        return false;
```
arnout: A missing vs_tlv should *not* be fatal!

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4780,6 +4733,44 @@ bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::Cmd
...
+
+    auto block = association_control_request_tlv->association_control();
+    if (block == wfa_map::tlvClientAssociationControlRequest::UNBLOCK) {
+        auto request_out = message_com::create_vs_message<
+            beerocks_message::cACTION_APMANAGER_CLIENT_ALLOW_REQUEST>(cmdu_tx, mid);
+        if (!request_out) {
+            LOG(ERROR) << "Failed building ACTION_APMANAGER_CLIENT_ALLOW_REQUEST message!";
+            return false;
+        }
+        request_out->ipv4() = vs_tlv->ipv4();
```
arnout: Note that vs_tlv may be nullptr.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4780,6 +4733,44 @@ bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::Cmd
...
+        request_out->mac()  = sta_mac;
+    } else if (block == wfa_map::tlvClientAssociationControlRequest::BLOCK) {
+        auto request_out = message_com::create_vs_message<
+            beerocks_message::cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST>(cmdu_tx, mid);
+        if (!request_out) {
+            LOG(ERROR) << "Failed building ACTION_APMANAGER_CLIENT_DISALLOW_REQUEST message!";
+            return false;
+        }
+        request_out->reject_sta() = vs_tlv->reject_sta();
```
arnout: Note that vs_tlv may be nullptr.

controller/src/beerocks/master/son_actions.cpp
```diff
@@ -154,14 +156,27 @@ void son_actions::unblock_sta(db &database, ieee1905_1::CmduMessageTx &cmdu_tx,
             continue;
         }
         Socket *sd = database.get_node_socket(hostap);
-        auto request =
-            message_com::create_vs_message<beerocks_message::cACTION_CONTROL_CLIENT_ALLOW_REQUEST>(
-                cmdu_tx);
-        if (request == nullptr) {
-            LOG(ERROR) << "Failed building ACTION_CONTROL_CLIENT_ALLOW_REQUEST message!";
+        if (!cmdu_tx.create(0,
+                            ieee1905_1::eMessageType::CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE)) {
```
arnout, line 160: There are three places in the controller where very similar code is used to build the Client Association Control Request message. It might be worth moving that to a single function. Unfortunately, we don't have a good place for it...

(nitpick)

controller/src/beerocks/master/son_actions.cpp
```diff
@@ -154,14 +156,27 @@ void son_actions::unblock_sta(db &database, ieee1905_1::CmduMessageTx &cmdu_tx,
...
+            cmdu_tx.addClass<wfa_map::tlvClientAssociationControlRequest>();
+        if (!association_control_request_tlv) {
+            LOG(ERROR) << "addClass wfa_map::tlvClientAssociationControlRequest failed";
+            break;
+        }
+        association_control_request_tlv->bssid_to_block_client() =
+            network_utils::mac_from_string(current_ap_mac);
+        association_control_request_tlv->association_control() =
+            wfa_map::tlvClientAssociationControlRequest::UNBLOCK;
+        //TODO: Get real validity_period_sec
```
arnout: I think this is only valid for blocking, not for unblocking, so it could be set to 0. But I don't find the relevant part of the MultiAP spec.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -148,24 +172,47 @@ void client_steering_task::steer_sta()
...
+            LOG(ERROR) << "addClass wfa_map::tlvClientAssociationControlRequest failed";
             return;
         }
-        request->mac() = network_utils::mac_from_string(sta_mac);
+
+        association_control_block_request_tlv->bssid_to_block_client() =
+            network_utils::mac_from_string(current_ap_mac);
+        association_control_block_request_tlv->association_control() =
+            wfa_map::tlvClientAssociationControlRequest::BLOCK;
+        association_control_block_request_tlv->validity_period_sec() = disassoc_timer;
```
arnout: I don't think this is correct... First of all, of course, disassoc_timer is expressed in TU while validity period is expressed in seconds, so that's a factor 1000 different. But also, the block should definitely still be in place immediately after the disassociation took place. So a more reasonable validity period, I think, would be the timeout used by the client steering task, i.e. steering_wait_time_ms/1000. In fact, that would remove the need to do unblocking in the FINALIZE state.

But again, that's for a different day. Could you make an issue for it?

arnout (approved):
The commit "agent: Replace control client request to easymesh" is actually for the controller. And there's a separate commit "controller:use EasyMesh messages for access control" that *also* changes the controller. It's not so clear based on what it is split like that.

Also, it would be better to have the controller change come *after* the agent change. Indeed, access control no longer works until the agent is updated, while vice versa it didn't work anyway with EasyMesh messages so adding support for them can't break anything. But that's a small thing.

Except for these remarks, the splitting and ordering of commits and the commit messages are excellent.

The rest of the review is mostly nitpicks. Even not failing on a missing VS message is something that can be fixed later, so I will approve already.

controller/src/beerocks/master/son_actions.cpp
CoralMalachi: I think you are right :
`The Validity Period field in a Client Association Control Request message with Association Control field set to 0x01 (Client Unblocking) is ignored`  
So I'll change its value to zero.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_1905_vs.yaml
CoralMalachi: Ok thanks 👍

controller/src/beerocks/master/tasks/client_steering_task.cpp
CoralMalachi: Yes 👍 @arnout

agent/src/beerocks/slave/son_slave_thread.cpp
CoralMalachi: But I'm checking for this here :
`
if (!vs_tlv) {
        LOG(ERROR) << "add_vs_tlv tlvVsClientAssociationControlRequest failed";
        return false;
    }
`
You mean I should not return false and use this vs_tlv only if it valid ? @arnout

arnout: > But I'm checking for this here :
> `if (!vs_tlv) { LOG(ERROR) << "add_vs_tlv tlvVsClientAssociationControlRequest failed"; return false; }`
> You mean I should not return false and use this vs_tlv only if it valid ? @arnout

Yes, that's what I meant with "a missing TLV should not be fatal".

controller/src/beerocks/master/tasks/client_steering_task.cpp
CoralMalachi: #432

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_1905_vs.yaml
CoralMalachi: Why it should always be 33? 
In case of denied due to poor channel conditions, it should not be 34? @arnout

PR #320, opened 2019-09-26
Controller: database

maksym-bielan:
Operating class parameter should be stored in database along with
channel.

Added operating class in database, also implemented setter and getter
functions.

### DIFF
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -232,6 +232,20 @@ int db::get_node_channel(std::string mac)
     return n->channel;
 }
 
+int db::get_hostap_operating_class(const sMacAddr &mac)
+{
+    auto mac_str = network_utils::mac_to_string(mac);
+    auto n       = get_node(mac_str);
+    if (!n) {
+        LOG(WARNING) << "node " << mac_str << " does not exist!";
+        return 0;
+    } else if (n->get_type() != beerocks::TYPE_SLAVE || !n->hostap) {
+        LOG(WARNING) << "node " << mac_str << " is not a valid hostap!";
+        return 0;
+    }
+    return n->hostap->operating_class;
+}
+
 bool db::set_node_vap_id(std::string mac, int8_t vap_id)
 {
     auto n = get_node(mac);
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -2878,6 +2878,12 @@ bool db::set_node_channel_bw(const std::string &mac, int channel, beerocks::eWiF
             n->hostap->vht_center_frequency      = vht_center_frequency;
             auto is_dfs                          = wireless_utils::is_dfs_channel(channel);
             set_hostap_is_dfs(mac, is_dfs);
+            if (channel <= 13)
+                n->hostap->operating_class = 81;
+            else if (channel <= 48)
+                n->hostap->operating_class = 115;
+            else
+                n->hostap->operating_class = 121;
         } else {
             LOG(ERROR) << __FUNCTION__ << " - node " << mac << " is null!";
             return false;
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -176,6 +176,8 @@ public:
 
     int get_node_channel(std::string mac);
 
+    int get_hostap_operating_class(const sMacAddr &mac);
+
     bool set_node_vap_id(std::string mac, int8_t vap_id);
     int8_t get_node_vap_id(std::string mac);
 
--- a/controller/src/beerocks/master/db/node.h
+++ b/controller/src/beerocks/master/db/node.h
@@ -142,6 +142,7 @@ public:
         beerocks::eIfaceType iface_type;
         std::string driver_version;
         std::vector<beerocks_message::sWifiChannel> supported_channels;
+        uint8_t operating_class    = 0;
         int ant_gain               = 0;
         int conducted_power        = 0;
         bool exclude_from_steering = false;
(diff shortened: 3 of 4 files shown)


controller/src/beerocks/master/db/node.h
```diff
@@ -51,7 +51,8 @@ class node {
 
     std::string ipv4;
     std::string manufacturer;
-    int channel = 0;
+    int channel             = 0;
+    uint8_t operating_class = 0;
```
morantr: I'm not sure it is the right place to put it. Because it is radio related data, and where you put it is a place for a common parameters (channel, for example, is common to radio and station nodes)

morantr (changes requested):

morantr:
Please use a more descriptive PR headline.

controller/src/beerocks/master/db/node.h
arnout: > I'm not sure it is the right place to put it. Because it is radio related data, and where you put it is a place for a common parameters (channel, for example, is common to radio and station nodes)

It probably isn't the right place to put it, but at the moment it is where the channel is, and it makes sense to put the operating class in the same place I think.

morantr: > > I'm not sure it is the right place to put it. Because it is radio related data, and where you put it is a place for a common parameters (channel, for example, is common to radio and station nodes)
> 
> It probably isn't the right place to put it, but at the moment it is where the channel is, and it makes sense to put the operating class in the same place I think.

I disagree. I don't understand why it doesn't make sense to put it at the right place, under the radio node.

YuriyMasechko: Moved under radio node. Set\get changed accordingly.

controller/src/beerocks/master/db/db.h
```diff
@@ -174,6 +174,9 @@ class db {
 
     int get_node_channel(std::string mac);
 
+    bool set_hostap_operating_class(std::string mac, uint8_t operating_class);
+    int get_hostap_operating_class(std::string mac);
```
morantr: If you use a string argument, it probably should be a const reference.
@arnout maybe we should use sMacAdde? What do you think?
Second, I would really like to get rid of "hostap" terminology and use "radio" instead.

morantr (changes requested):

controller/src/beerocks/master/db/db.h
arnout: > If you use a string argument, it probably should be a const reference.
> @arnout maybe we should use sMacAdde? What do you think?

Yes please, for any new functionality, use only sMacAddress.

Also, any non-scalar parameter (which std::string definitely is) should use const-ref (cfr. [Coding style, functions, 2nd paragraph](https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md#functions)).

> Second, I would really like to get rid of "hostap" terminology and use "radio" instead.

I agree on the principle. However, currently "radio" is used (almost) nowhere, so using it for this would just create more confusion. So I created a new task #436.

YuriyMasechko: Done.

morantr (approved):

controller/src/beerocks/master/db/db.cpp
```diff
@@ -232,6 +232,35 @@ int db::get_node_channel(std::string mac)
...
 }
 
+int db::get_hostap_operating_class(const sMacAddr &mac)
+{
+    auto mac_str = network_utils::mac_to_string(mac);
+    auto n       = get_node(mac_str);
+    if (!n) {
+        LOG(WARNING) << __FUNCTION__ << " - node " << mac_str << " does not exist!";
+        return 0;
+    } else if (n->get_type() != beerocks::TYPE_SLAVE || n->hostap == nullptr) {
```
arnout:
```suggestion
    } else if (n->get_type() != beerocks::TYPE_SLAVE || !n->hostap) {
```

smartptrs shouldn't be compared to nullptr, instead operator! should be used.

Same comment below.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -232,6 +232,35 @@ int db::get_node_channel(std::string mac)
...
+        LOG(WARNING) << __FUNCTION__ << "node " << mac_str << " is not a valid hostap!";
```
arnout:
```suggestion
        LOG(WARNING) << "node " << mac_str << " is not a valid hostap!";
```

`LOG()` already prints the function if so configured. The default config doesn't print the function name, but since it prints file and line number it's not needed either.

Same comment below as well.

arnout (changes requested):
Just adding the functions isn't very helpful. The PR should also have some code that calls those functions.

Since we don't have the controller code yet to get the current operclass and channel, I would just make a stub implementation, i.e. in `db::set_node_channel_bw` set the operating class to 81, 115 or 124 depending on the channel number. With of course a big TODO that the operclass should be an argument of that function.

In fact, I don't think we need `db::set_hostap_operating_class` at all. It is only ever going to be set together with the other channel parameters, so only from `db::set_node_channel_bw` or its replacement. So, for this first commit I would move the setting to `db::set_node_channel_bw`. In a second commit, the setter should be used for the steering request. If there's any other place where we need the current operating class (which I don't think is the case at the moment), subsequent commits should use operclass there as well.

controller/src/beerocks/master/db/db.cpp
YuriyMasechko: Done.

controller/src/beerocks/master/db/db.cpp
YuriyMasechko: Set function was removed as not needed.

arnout (approved):

Issue #297, opened 2019-09-16
[TASK] Controller client steering task


arnout:
## Description

The controller has a client steering task that manages how the steering of the client is performed.

Make sure that the appropriate EasyMesh CMDUs are sent out, and the responses are handled.

Also make a test using dummy BWL for all the features of this task.

arnout:
Done in #314

arnout:
Merged (again) in 174069a

Issue #293, opened 2019-09-16
[TASK] Create a band steering test


arnout:
## Description

Add a new test script that runs the band steering flow.

## Exit criteria

- Using dummy BWL register some clients on the agent.
- Send band steering command through beerocks_cli.
- Verify that the steered agent sends bwl command to perform steering.
- Dummy BWL confirms steering result.
- Agent sends steering result to controller.

arnout:
Merged in 1155857

PR #322, opened 2019-09-26
agent: client steering request

### LINKED ISSUE #296: [TASK] [agent] EasyMesh client steering request
## Description

Agent handles the Client Steering Request message for legacy steering by doing actual steering through BWL.

## Exit criteria

- When receiving a Client Steering Request, the appropriate bwl call is made.
- A dummy implementation of this call exists to allow testing.
- An ACK is returned to the controller, but nothing else.

YuriyMasechko:
Agent handles the Client Steering Request message for legacy steering by doing
actual steering through BWL.

Call BWL for Client Steering Request.

Fixes #296 #297 #327 #328 #293 #323 


common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -139,6 +139,7 @@ bool ap_wlan_hal_dummy::sta_deauth(int8_t vap_id, const std::string &mac, uint32
 bool ap_wlan_hal_dummy::sta_bss_steer(const std::string &mac, const std::string &bssid, int chan,
                                       int disassoc_timer, int valid_int)
 {
+    LOG(DEBUG) << "Got steer request for " << mac << " steer to " << bssid;
```
arnout, line 142: Please add a second commit that updates test_flows.sh to use this message instead of checking for MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE (which is BTW a bug...).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4866,6 +4867,23 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
 
     //check if the Request Mode bit is set
     if (request_mode) {
+        //TODO Handle 0 or more then 1 sta in list, currenlty cli steers only 1 client
```
arnout, line 4838: I created #324 and #325 for this.

arnout (changes requested):
The commit message is wrong, this is not a legacy request. That's my mistake, I corrected #269 . BTW, please make sure that the pull request description refers to the issue - I'll do that now for this one as an example.

Please also update test_flows.sh to use the bwl log message instead of the one from receiving the message.

There's still a problem that the message will be handled by *all* slaves. I'll create a new issue for that.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
YuriyMasechko: If you are about to change test_client_steering_mandate(), then it's incorrect, there is handling of both options - mandate and opportunity. I added handling only for mandate, so for opportunity bwl   will not be called. I could change message check for only one option.

arnout: I was indeed wrong about the MULTI_AP_POLICY_CONFIG_REQUEST, I don't know how that happened.

But what I meant was to replace the following with a check of bwl log instead of the log in the agent itself.

```sh
    dbg "Confirming Client Steering Request message was received - mandate"
    check docker exec -it repeater1 sh -c \
        'grep -i -q "CLIENT_STEERING_REQUEST_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
    check docker exec -it repeater1 sh -c \
        'grep -i -q "CLIENT_STEERING_REQUEST_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
```

YuriyMasechko: Update test scrip to check for bwl message. Updated it to check only for required agent as check for BSSID was added.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4837,14 +4837,19 @@ bool slave_thread::handle_ack_message(Socket *sd, ieee1905_1::CmduMessageRx &cmd
...
     const auto mid = cmdu_rx.getMessageId();
-    LOG(DEBUG) << "Received CLIENT_STEERING_REQUEST_MESSAGE , mid=" << std::hex << int(mid);
 
     auto steering_request_tlv = cmdu_rx.addClass<wfa_map::tlvSteeringRequest>();
     if (!steering_request_tlv) {
         LOG(ERROR) << "addClass wfa_map::tlvSteeringRequest failed";
         return false;
     }
 
+    if (steering_request_tlv->bssid() != hostap_params.iface_mac) {
```
arnout, line 4810: @morantr Could you double-check if this is the correct parameter to use as bssid?

To be honest, I'm a bit confused about all the different terminology used in beerocks...

tools/docker/tests/test_flows.sh
```diff
@@ -148,9 +148,7 @@ test_client_steering_mandate() {
     sleep 1
     dbg "Confirming Client Steering Request message was received - mandate"
     check docker exec -it repeater1 sh -c \
-        'grep -i -q "CLIENT_STEERING_REQUEST_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec -it repeater1 sh -c \
-        'grep -i -q "CLIENT_STEERING_REQUEST_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+        'grep -i -q "Got steer request" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
```
arnout, line 151: This change would ideally have been in a separate commit, together with adding this log message to bwl. But it's not important enough to still do it now. (nitpick)

tools/docker/tests/test_flows.sh
```diff
@@ -178,6 +178,39 @@ test_client_steering_mandate() {
...
+    dbg "Confirming ACK message was received"
+    check docker exec -it repeater1 sh -c \
+        'grep -i -q "ACK_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+
+    dbg "Disconnect dummy STA from wlan0"
+    check docker exec -it repeater1 sh -c \
+        'echo "AP-STA-DISCONNECTED 11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan0/EVENT'
+    dbg "Connect dummy STA to wlan2"
+    check docker exec -it repeater1 sh -c \
+        'echo "AP-STA-CONNECTED 11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan2/EVENT'
```
arnout, line 213: The test should also verify that the corresponding messages are received and handled by the client task, i.e. verify log messages ["sta ... disconnected after successful steering"](https://github.com/prplfoundation/prplMesh/blob/9c196ee8789bd954857aec9e61650a1e91c74864/controller/src/beerocks/master/tasks/client_steering_task.cpp#L223) and ["steering successful for sta ..."](https://github.com/prplfoundation/prplMesh/blob/9c196ee8789bd954857aec9e61650a1e91c74864/controller/src/beerocks/master/tasks/client_steering_task.cpp#L220).

arnout (approved):
I'm already going to approve, but:

- the finalisation of the client_steering task should also be tested;
- I want @morantr to give feedback as well.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: slave_thread::handle_cmdu_control_message
...
    // to me or not to me, this is the question...
    if (beerocks_header->radio_mac() != hostap_params.iface_mac) {
        return true;
    }

tools/docker/tests/test_flows.sh
YuriyMasechko: Only "steering successful for sta " could be added now.
STA_DISCONNECTED is not sent by ACTION_CONTROL_CLIENT_DISCONNECTED_NOTIFICATION.
Added TODO.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4866,6 +4867,23 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
 
     //check if the Request Mode bit is set
     if (request_mode) {
+        //TODO Handle 0 or more then 1 sta in list, currenlty cli steers only 1 client
+        auto request_out = message_com::create_vs_message<
+            beerocks_message::cACTION_APMANAGER_CLIENT_BSS_STEER_REQUEST>(cmdu_tx, mid);
+        if (request_out == nullptr) {
```
morantr, line 4841: For smart_ptr use '!' instead of comparing to nullptr.
```if (!request_out)```

File: controller/src/beerocks/master/tasks/client_steering_task.cpp [hunk 7]
```diff
@@ -174,42 +175,51 @@ void client_steering_task::steer_sta()
...
+        return;
     }
 
     auto current_ap_mac = database.get_node_parent(sta_mac);
 
+    steering_request_tlv->request_flags().request_mode =
+        wfa_map::tlvSteeringRequest::REQUEST_IS_A_STEERING_MANDATE_TO_TRIGGER_STEERING;
+    steering_request_tlv->request_flags().btm_disassociation_imminent_bit = disassoc_imminent;
+
+    steering_request_tlv->btm_disassociation_timer() = disassoc_timer;
```
morantr, line 197: This line is the main reason why I asked to revert this.
It is a pitfall.
`disassoc_timer` is defined by intel hardware in units of beacon intervals which is 100ms.
The standard, on the other hand, defines it in a TU unit which is approximately 1ms.
I think that the right thing to do is add "_ms" prefix to all the "disassoc_timer" references from controller to the ap_manager, including yaml files in common and framework, and convert to 100ms beacon interval in bwl function.

morantr (changes requested):
I reviewed until 7ec8bb29b477e3bf5f65dae27fb1455d8d21289f.

controller/src/beerocks/master/tasks/client_steering_task.cpp
arnout: > This line is the main reason why I asked to revert this.
> It is a pitfall.
> `disassoc_timer` is defined by intel hardware in units of beacon intervals which is 100ms.
> The standard, on the other hand, defines it in a TU unit which is approximately 1ms.

Yes, I created #323 for this.

Note that it is already wrong, at least on Turris on the dev/turris-omnia-support branch, because the number gets passed as-is to bwl, which passes it directly to the hostapd BSS_TM_REQ command to hostapd. hostapd expects a value in TU there (I checked in he hostapd source code). Note that AFAIU, the practical effect of the disassociation timer value in the BTM request is none, because the client will just steer immediately anyway.

> I think that the right thing to do is add "_ms" prefix to all the "disassoc_timer" references from controller to the ap_manager, including yaml files in common and framework, and convert to 100ms beacon interval in bwl function.

It should be a `_tu` suffix, to be precise. (For reference, a TU is 1024µs) Also, I would express everything in TU, because that's what most of the WFA standards do.

morantr: > Note that it is already wrong, at least on Turris on the dev/turris-omnia-support branch, because the number gets passed as-is to bwl, which passes it directly to the hostapd BSS_TM_REQ command to hostapd. hostapd expects a value in TU there (I checked in he hostapd source code). 

This strengthens my point. The code in the controller and agent should be aligned to the standard units, and be converted on bwl, which is vendor-specific.

> It should be a _tu suffix, to be precise. (For reference, a TU is 1024µs) Also, I would express everything in TU, because that's what most of the WFA standards do.

That's fine by me, as long as the unit is explicitly mentioned in the variable name.
I prefered "ms" because everyone knows its meaning, but will google TU. 
But TU is also fine.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: According to code through the file, every result of **create_vs_message** is checked via nullptr. Should I change only this one?

controller/src/beerocks/master/tasks/client_steering_task.cpp
YuriyMasechko: Done. Please review.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: yes change this one only, and we will need to refactor the rest of it someday.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -443,7 +443,7 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
 
     if (disassoc_timer) {
         cmd += std::string() + " disassoc_imminent=" + "1" +
-               " disassoc_timer=" + std::to_string(disassoc_timer);
+               " disassoc_timer=" + std::to_string(disassoc_timer / 100);
```
morantr: Please add some comment that explains that we divide it by 100, because the hostapd expect to get this value in numbers of beacon interval which is 100ms.
Please do it as well on fapi and wav.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: This is not correct.
The BSSID used in the message is actually a VAP (Virtual AP) mac which could be different from the radio mac.
Under the radio which transmits and lets stations to connect, could be VAPs which will operate on the same radio and will have the same channel and bandwidth, but have different SSID and credentials and MAC, of course.

You will have to pass it to ap_manager, which will check if it has the BSSID as one of its VAPS.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4866,6 +4867,23 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+        auto bssid_list = steering_request_tlv->target_bssid_list(0);
+
+        request_out->params().mac            = std::get<1>(steering_request_tlv->sta_list(0));
+        request_out->params().disassoc_timer = steering_request_tlv->btm_disassociation_timer();
+        request_out->params().bssid.mac      = std::get<1>(bssid_list).target_bssid;
+        request_out->params().bssid.channel  = std::get<1>(bssid_list).target_bss_channel_number;
+        request_out->params().disassoc_imminent =
+            steering_request_tlv->request_flags().btm_disassociation_imminent_bit;
+
+        message_com::send_cmdu(ap_manager_socket, cmdu_tx);
```
morantr, line 4854: From the standard:

> If a Multi-AP Agent receives a Client Steering Request message with Request Mode bit set to one, then it shall respond within one second with a 1905 Ack message. If a STA specified in the Client Steering Request message is not associated with the source BSSID specified in the same message (an error scenario), for each of those unassociated STAs, the Multi-AP Agent shall include an Error Code TLV with the reason code field set to 0x02 and the STA MAC address field included per section 17.2.36 in the 1905 Ack message.

You will have to query the AP manager if the station is connected, and return 1905 ACK message accordingly.
So what @CoralMalachi did in lines 4860-4864 needs to be replaced.

morantr, line 4854: The standard allows the tlv to contain multiple stations and target bssids. It is ok that the current implementation fits to one station and target bssid, but it needs to be documented.
Please a TODO comment that explains that it is a temporary implementation, and what has to be done in the future.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr, line 4838: It is better to compare `request_mode` to the enum. It will be much more clearer that way.

morantr (dismissed):
I don't understand why did you do merge and not rebase. Now when gitlog your branch I don't see the commits before the merge. @arnout how can we fix it?
Second, what about Client Steering BTM Report, isn't it under this PR scope as well?

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: There is such comment above:
//TODO Handle 0 or more then 1 sta in list, currenlty cli steers only 1 client

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
YuriyMasechko: Done

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Done in separate commit.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Done.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Moved to handle in manager. As there is no way to check is STA is connected, added TODO.

rmelotte (approved):
I did not have the time to review all the code, but I tested it on the Turris Omnia, and manually steering clients is working.
If anyone is willing to test it as well, I created a new branch which has the changes from the turris omnia dev branch on top of this branch: dev/turris-omnia-support-easymesh

arnout:
The first commits in this PR (which were the ones originally reviewed) are OK now.

The new commits since the last review need some more work.

Since those first commits (when applied as a whole) don't break anything that was working before, I'm going to force-push this branch with only those, and merge it. Then I'll make a new PR for the rest, and add my review comments to it.

PR #331, opened 2019-09-30
controller: db: pass sMacAddr by value instead of by reference

arnout:
According to CppCoreGuideline F.16, a small object should be passed by
value instead of const-ref. "Small" is not objectively defined, but it
would be two-three words. Since sMacAddr is 6 bytes, it is less than two
words even on a 32-bit architecture, so definitely small.

Therefore, change `add_node` and `remove_node` to take their arguments by
value instead of by const-ref.

This should have been part of #291 but I forgot to commit my changes before pushing/merging.

### DIFF
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -58,8 +58,8 @@ bool db::add_virtual_node(sMacAddr mac, sMacAddr real_node_mac)
     return true;
 }
 
-bool db::add_node(const sMacAddr &mac, const sMacAddr &parent_mac, beerocks::eType type,
-                  const sMacAddr &radio_identifier)
+bool db::add_node(sMacAddr mac, sMacAddr parent_mac, beerocks::eType type,
+                  sMacAddr radio_identifier)
 {
     if (mac == network_utils::ZERO_MAC) {
         LOG(ERROR) << "can't insert node with empty mac";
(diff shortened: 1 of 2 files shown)


tomereli (approved):
From Zion NP - LGTM 👌

morantr (changes requested):
I think it is better to remove this commit.
Small is a relative term. Where is the limit between a memory size to use a reference or to make a copy?
When passing an argument as a reference, we actually passing a pointer, which its size is 4 byte in 32bit architecture.
So if we want to pass an argument that its size is greater than sizeof(int), it is better to use reference.
If its size is lesser than sizeof(int), the reference is redundant since the compiler will pass sizeof(int) anyway, so we can use a copy.

arnout:
> I think it is better to remove this commit.
> Small is a relative term. Where is the limit between a memory size to use a reference or to make a copy?
> When passing an argument as a reference, we actually passing a pointer, which its size is 4 byte in 32bit architecture.
> So if we want to pass an argument that its size is greater than sizeof(int), it is better to use reference.
> If its size is lesser than sizeof(int), the reference is redundant since the compiler will pass sizeof(int) anyway, so we can use a copy.

So, to lay this to rest once and for all, I made a few tests with GCC 8 for ARMv7 (i.e. what gets used on Turris Omnia), and at least for the sMacAddress structure, the code generated with const ref arguments is slightly better (i.e. smaller and equal or less load/store instructions) than the code generated with pass by value. This is true both for `-Os` and `-O3` compilation. I did the same test with two arguments and the effect was the same. And finally, I repeated the test on x86_64 (with GCC 9), and even there the const ref generated slightly better code. The only situation where the code was slightly better for the pass-by-value case was in the call site if the value is zero.

Therefore, we can conclude that @morantr is right and we should pass by reference if the structure is larger than a single word.

So I've created #333 to clarify this in CODINGSTYLE.

Thanks Moran for blocking this!

PR #335, opened 2019-10-07
Controller:change bss_infos to use sMacAddress as key

### LINKED ISSUE #334: [BUG] <controller>: teardown sent due to wrong al-mac address in DB
Describe the bug -

It looks like there is in fact no config in the database for that al-mac address.
the ruid is empty .
controller is sending a teardown to the Broadcom agent.

Found
while debugging UCC test 5.4.1 24G

logs:
https://ftp.essensium.com/owncloud/index.php/apps/files?dir=/prplmesh/certification/prpl%20controller%20tests/FAIL/MAP-5.4.1_ETH_FH24G_Broadcom_Marvell_teardown

link to passing test on master from 2 weeks ago:
https://ftp.essensium.com/owncloud/index.php/apps/files/?dir=/prplmesh/certification/prpl%20controller%20tests/PASS/MAP-5.4.1_ETH_FH24G_setup_prplmesh_broadcom_marvell&fileid=423625

Steps to reproduce
build prplmesh
.../prplmesh_utils.sh restart
../beerocks_cli -u 5000

from UCC:
run test 5.4.1 24G

current behavior  (controller.log):  
INFO 16:09:52:785 <140154884115328> son_master_thread.cpp[734] --> AP_AUTOCONFIGURATION_WSC M1 al_mac=46:55:66:77:00:00 ruid=00:00:00:00:00:00
DEBUG 16:09:52:785 <140154884115328> son_master_thread.cpp[735] -->    device Intel Ubuntu prplMesh-agent prpl12345
DEBUG 16:09:52:785 <140154884115328> encryption.cpp[32] --> Generating DH keypair
DEBUG 16:09:52:787 <140154884115328> encryption.cpp[99] --> Computing DH shared key
DEBUG 16:09:52:789 <140154884115328> son_master_thread.cpp[636] --> WSC config_data: tear down

expected behavior
test pass

CoralMalachi:
Currently, test 5.4.1 24G keeps failing.

The problem is that the controller sends unnecessary teardown request to
the agent which causing to tear down all BSSes on that radio.

The reason for this is that bss_info is being stored with a string key in the database. 
The UCC sends the MAC address with uppercase hex digits, but the
network_utils::mac_to_string method uses lowercase.

As a result, the value of bss_info_confs is empty, which causes the num_bsss's value 
be zero and that leads to tearing down action : 

`

    if (num_bsss == 0) {
        if (!autoconfig_wsc_add_m2(tlvwscM1, nullptr)) {
            LOG(ERROR) << "Failed setting M2 attributes";
            return false;
        }
    }
`

The solution is to use sMacAddress as the database key instead of a string.

Fixes #334 

### DIFF
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -523,8 +523,8 @@ public:
         std::string network_key;
         WSC::eWscVendorExtSubelementBssType bss_type;
     };
-    void add_bss_info_configuration(const std::string &al_mac, const bss_info_conf_t &bss_info);
-    std::list<db::bss_info_conf_t> &get_bss_info_configuration(const std::string &al_mac);
+    void add_bss_info_configuration(const sMacAddr &al_mac, const bss_info_conf_t &bss_info);
+    std::list<db::bss_info_conf_t> &get_bss_info_configuration(const sMacAddr &al_mac);
     void clear_bss_info_configuration();
 
     //
@@ -719,7 +719,7 @@ private:
 
     // certification
     std::shared_ptr<uint8_t> certification_tx_buffer;
-    std::unordered_map<std::string, std::list<bss_info_conf_t>> bss_infos; // key=al_mac
+    std::unordered_map<sMacAddr, std::list<bss_info_conf_t>> bss_infos; // key=al_mac
 };
 
 } // namespace son
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -760,7 +760,7 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
 
     tlvRuid->radio_uid() = network_utils::mac_from_string(ruid);
 
-    const auto &bss_info_confs = database.get_bss_info_configuration(al_mac);
+    const auto &bss_info_confs = database.get_bss_info_configuration(tlvwscM1->mac_attr().data);
     uint8_t num_bsss           = 0;
 
     for (const auto &bss_info_conf : bss_info_confs) {
--- a/common/beerocks/bcl/include/beerocks/bcl/network/net_struct.h
+++ b/common/beerocks/bcl/include/beerocks/bcl/network/net_struct.h
@@ -22,6 +22,18 @@ inline bool operator==(sMacAddr const &lhs, sMacAddr const &rhs)
 
 inline bool operator!=(sMacAddr const &lhs, sMacAddr const &rhs) { return !(rhs == lhs); }
 
+template <> struct std::hash<sMacAddr> {
+    size_t operator()(const sMacAddr &m) const
+    {
+        uint64_t value_to_hash = 0;
+        for (size_t byte = 0; byte < sizeof(m.oct); byte++) {
+            value_to_hash <<= 8;
+            value_to_hash += m.oct[byte];
+        }
+        return std::hash<std::uint64_t>()(value_to_hash);
+    }
+};
+
 namespace beerocks {
 namespace net {
 
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -3157,12 +3157,12 @@ void db::setting_certification_mode(bool en)
     config.certification_mode = true;
 }
 
-void db::add_bss_info_configuration(const std::string &al_mac, const bss_info_conf_t &bss_info)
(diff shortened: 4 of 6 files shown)


arnout (commented):
Good that you pushed this :-)

I think you're making this way too complicated. Isn't it sufficient to define `operator ==` in struct sMacAdress?

arnout:
> I think you're making this way too complicated. Isn't it sufficient to define `operator ==` in struct sMacAdress?

My bad, that operator exists already. What we need is a specialisation `template <> struct std::hash<sMacAddress>` which is exactly what you do here. Only, it should be defined together with the comparison operator (or better yet, in sMacAddr.h) instead of being passed explicitly to the unordered_map.

controller/src/beerocks/master/db/db.h
```diff
@@ -717,9 +717,61 @@ class db {
...
+            return (line1.oct[0] ^ line2.oct[0]);
+        }
+    };
+
+    struct hash_fn {
+        size_t operator()(const sMacAddr &m) const
+        {
+            std::string s1 = beerocks::net::network_utils::mac_to_string(m.oct);
+            s1.erase(std::remove(s1.begin(), s1.end(), ':'), s1.end());
+            size_t h1 = std::hash<std::string>()(s1);
```
arnout: I would create the hash based on a uint64_t:

```cpp
    uint64_t value_to_hash = 0;
    for (size_t byte = 0; byte < sizeof(sMacAddr.oct); byte++) {
        value_to_hash <<= 8;
        value_to_hash += sMacAddr.oct[byte];
    }
    return std::hash(value_to_hash);
}
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -760,8 +760,9 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(Socket *sd,
 
     tlvRuid->radio_uid() = network_utils::mac_from_string(ruid);
 
-    const auto &bss_info_confs = database.get_bss_info_configuration(al_mac);
-    uint8_t num_bsss           = 0;
+    const auto &bss_info_confs =
+        database.get_bss_info_configuration(network_utils::mac_from_string(al_mac));
```
arnout: Use `tlvwscM1->mac_attr().data` directly instead of converting back from string.

CoralMalachi: yes, I have noticed it later but didn't push yet, I'll upload soon :)

controller/src/beerocks/master/db/db.h
```diff
@@ -717,9 +717,21 @@ class db {
 
     std::shared_ptr<vaps_list_t> m_vap_list;
 
+    struct hash_fn {
```
arnout: This definition should move to sMacAddr.h (and the name changed of course). (Note: also the operator== and operator!= definitions should move from net_struct.h to sMacAddr.h - in a separate commit of course).

Also, I think it could be defined as the standard hash function for sMacAddr.

```
template <> struct std::hash<sMacAddress> {
    size_t operator()(const sMacAddr &m) const
    ...
};
```

You might need to play around with the namespaces to get this right.

arnout (approved):
Looks good to me as it is - I'd prefer to move the hash function to the definition of sMacAddr but that's more a nitpick.

This solves the issue in certification, right?

Is there any reason to keep this in draft?

controller/src/beerocks/master/db/db.h
CoralMalachi: I actually thought like you but sMacAddr.h is an auto-generated file, when I do dist-clean all my changes there will be gone, so how to do this right? edit the YAML file? @arnout

arnout: Ah, you're right, I thought sMacAddr.h was just copied from the src directory.

Indeed, then it is not possible to move it to sMacAddr.h. So instead it should move to net_struct.h.

And the rename to `std::hash` should also be possible of course.

CoralMalachi:
> 
> 
> Looks good to me as it is - I'd prefer to move the hash function to the definition of sMacAddr but that's more a nitpick.
> 
> This solves the issue in certification, right?
> 
> Is there any reason to keep this in draft?

Yes, it solves the issue :) (checked with UCC).

controller/src/beerocks/master/db/db.h
CoralMalachi: I planed to move it to net_struct.h but I see now that you already merged this PR.
if I want to add this do you want me to open another PR or revert this one? @arnout

CoralMalachi: Nevermind I can see you have done this already, thanks 👍

PR #426, opened 2019-10-10
tools: docker: add python3 to the images

arnout:
We will switch our Python scripts to use Python3 (only). Therefore, the
docker images should have Python3 available.

For backward compatibility, Python2 is kept for the time being. Once
everything is converted it can be removed again.


arnout:
I have already pushed the updated builder and runner images to dockerhub, because @Galvien depends on them.

morantr (approved):

Issue #323, opened 2019-09-26
[TASK] Convert BTM disassociation timer value to TU


arnout:
## Description

The disassoc_timer used in beerocks is in units of 100ms, while the btm_disassociation_timer in the EasyMesh SteeringRequest TLV is in units of 1TU (1024µs).

We should convert the uses of the disassociation timer to use TUs everywhere since that's what goes in the BTM request anyway.

## Exit criteria

- All uses of disassociation timer are in units of 1TU.
  - `BSS_STEER_DISASSOC_TIMER` in `beerocks_defines.h`
  - `channel_selection_task::DISASSOC_STEER_TIMER_SEC`
  - `disassoc_timer` in `optimal_path_task::work()`
  - bml
  - cli
- bwl implementations are adapted as needed.
- Documentation in `son_actions::steer_sta and client_steering_task` specifies this.
- Documentation of `ap_wlan_hal::sta_bss_steer()` specifies this.

Issue #326, opened 2019-09-26
[TASK] agent: validate client steering parameters


arnout:
## Description

When a Client Steering Request message is received, the STA may not be associated with that particular BSSID. In this case, the ACK message should contain an Error Code TLV with error code 0x02.

## Exit criteria

- If the STA is not associated with that particular BSSID, the ACK message contains an Error Code TLV.
- If multiple STA are specified and some of them do and some of them don't belong to that BSSID, the ACK message only contains error codes for the "wrong" STAs.
- The STA MAC address is included in the error code TLV.
- The steering is still performed for STA which do belong to the BSSID.
- A test is added that validates all of the above scenarios.

PR #427, opened 2019-10-10
EasyMesh BSS Steer Response and use TU

arnout:
This PR is the continuation of #322 and deals with two issues:

- #323 Convert BTM disassociation timer to units of 1TU
- Replace intel proprietary steer response with tlvSteeringBTMReport


agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -711,15 +711,37 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        auto it = std::find_if(vap_unordered_map.begin(), vap_unordered_map.end(),
+                               [&](const std::pair<int, bwl::VAPElement> &element) {
+                                   return element.second.mac == bssid;
+                               });
+
+        if (it == vap_unordered_map.end()) {
+            LOG(DEBUG) << "Steer request for unknown vap";
+            return false;
+        }
+        //TODO Check for STA errors, if error ACK with ErrorCodeTLV
```
arnout, line 726:
```suggestion
        //TODO Check if STA is really connected to this BSS, if not ACK with ErrorCodeTLV
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1019,6 +1041,7 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
             return false;
         }
 
+        //TODO EasyMesh SteeringBTMReport should contain source BSSID and target BSSID
```
arnout, line 1044: Could you create an issue (task) for this?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4807,65 +4832,37 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
         return false;
     }
 
-    if (steering_request_tlv->bssid() != hostap_params.iface_mac) {
```
arnout, line 4810: Darn, now I realise that the partial PR I merged actually *does* break things, because this condition was wrong...

So, I'd like to merge this PR today still.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -316,9 +316,10 @@ sNodeBssSteerTarget:
 sNodeBssSteerRequest:
   _type: struct
   mac: sMacAddr
-  disassoc_timer: uint16_t
+  bssid: sMacAddr
```
arnout: To avoid confusion about this, I would rename this to `source_bssid` or `cur_bssid`.

common/beerocks/bwl/common/ap_wlan_hal.h
```diff
@@ -113,7 +113,7 @@ class ap_wlan_hal : public virtual base_wlan_hal {
      * @param [in] mac The MAC address of the station.
      * @param [in] bssid The MAC address of the target AP.
      * @param [in] chan The channel of the target AP.
-     * @param [in] disassoc_timer Time in milliseconds before the AP should 
+     * @param [in] disassoc_timer Time in TU (time units, 1 TU = 1.024ms) before the AP should
```
arnout, line 116: To complete the transition, we could change the argument name (here and in all implementations) into `disassoc_timer_ms` as well. (nitpick)

arnout (commented):
There is something wrong with the first three commits: when applying the first one only, the build will fail because not all callers have been updated. The proper way to deal with this is to change the function definitions in one place (e.g. tlvf) and then update all callers as well in the same commit.

However, reworking it in that way is going to be way too much work, so it's not necessary for me. Instead, just squash the first three commits in one, except for the AutoGenerated changes which should be done in a separate commit.

The next three commits (agent: move client steering handling to manager) have a similar problem: the first commit is using tlvf objects that are added in the two following commits. There, also, I think the easiest solution now is to squash them and split off the AutoGenerated again.

agent/src/beerocks/slave/ap_manager_thread.cpp
YuriyMasechko: There is more then 1 error for STA

agent/src/beerocks/slave/ap_manager_thread.cpp
YuriyMasechko: #429 created

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -139,6 +139,9 @@ cACTION_APMANAGER_CLIENT_IRE_CONNECTED_NOTIFICATION:
   _type: class
   mac: sMacAddr
 
+cACTION_APMANAGER_ACK:
+  _type: class
```
morantr, line 143: The 1905 Ack message should have a "reason" and "sta_mac" fields, to notify if the station is not connected to the requested bssid.

From Multi-Ap specifications:
> If a STA specified in the Client Steering Request message is not associated
with the source BSSID specified in the same message (an error scenario), for each of those unassociated STAs, the
Multi-AP Agent shall include an Error Code TLV with the reason code field set to 0x02 and the STA MAC address field
included per section 17.2.36 in the 1905 Ack message.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -711,6 +711,28 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        auto bssid                    = network_utils::mac_to_string(request->params().cur_bssid);
+        const auto &vap_unordered_map = ap_wlan_hal->get_radio_info().available_vaps;
+        auto it = std::find_if(vap_unordered_map.begin(), vap_unordered_map.end(),
+                               [&](const std::pair<int, bwl::VAPElement> &element) {
+                                   return element.second.mac == bssid;
+                               });
+
+        if (it == vap_unordered_map.end()) {
+            LOG(DEBUG) << "Steer request for unknown vap";
+            return false;
```
morantr, line 724: There is no need to print it. Just add a comment that says that AP does not have the requested vap, and probably will be handled on the other AP, and return true, since this is not an error.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -711,6 +711,28 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        }
+        //TODO Check for STA errors, if error ACK with ErrorCodeTLV
+        auto response = message_com::create_vs_message<beerocks_message::cACTION_APMANAGER_ACK>(
+            cmdu_tx, beerocks_header->id());
+
+        if (response == nullptr) {
```
morantr, line 730: if (!response)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2403,6 +2403,25 @@ bool slave_thread::handle_cmdu_ap_manager_message(
 
         break;
     }
+    case beerocks_message::ACTION_APMANAGER_ACK: {
+        auto response_in = cmdu_rx.addClass<beerocks_message::cACTION_APMANAGER_ACK>();
+        if (response_in == nullptr) {
```
morantr, line 2408: if (!response_in)

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
YuriyMasechko: I think this should be added together with ErrorCodeTLV and possibility to check if STA is connected, for now there is no way even to fill those fields.

arnout: Indeed, the STA check is the only thing that can possibly fail and that is not implemented yet (because the slave / ap manager doesn't have the station list).

#326 covers this.

But on the other hand, @morantr is right that to be able to implement that, we will need to add reason and sta_mac fields (or ErrorCodeTLV) to cACTION_APMANAGER_ACK.

But of course, if #435 gets implemented first, the point is moot because all the cACTION_APMANAGER can be removed.

YuriyMasechko: Done in #437

agent/src/beerocks/slave/ap_manager_thread.cpp
YuriyMasechko: Done in #437

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Done in #437

PR #428, opened 2019-10-11
master: WSC: ignore authentication and encryption flags

arnout:
The WSC specification seems to claim that the authentication and
encryption flags in the M1 message should indicate which authentication
and encryption types are supported by the enrollee. However, actual
implementations don't seem to follow this: we've seen agents that don't
report support for WPA2PSK (even though MultiAP mandates WPA2PSK), and
we've seen agent that don't set any flags at all.

Therefore, ignore the autentication and encryption flags. An
informational message with all details is still sent to the logs.


CoralMalachi (approved):
So after Qualcomm was turned off and the Marvell STA was commented in the AllInitConfig-MAP.txt, test 5.4.1_ETH_FH5GL passed.

morantr (approved):

PR #431, opened 2019-10-14
Non-Intel agent test

### LINKED ISSUE #171: [TASK] Add non-Intel agent in test
## Description

The controller is supposed to support non-Intel agents (#134). To verify this in CI, let one of the agents in CI act as a non-Intel agent.

## Exit criteria

* Repeater agent in CI does not send Intel VS messages
* Repeater agent in CI errors if it receives Intel VS messages
* Agent config files specifies if it acts as Intel or non-Intel

arnout:
Since we want to be EasyMesh compliant, at least some of the tests we do should be done with a non-Intel agent.

This is done with an option in the agent that makes it behave as non-Intel.
For now, this option only affects the Intel VS TLV used in the M1.
It should be extended to all VS TLVs added to EasyMesh CMDUs.

This option is set for both repeaters, but not for the gateway (otherwise the gateway will not properly start up).

Closes #171.

### DIFF
--- a/agent/src/beerocks/slave/beerocks_slave_main.cpp
+++ b/agent/src/beerocks/slave/beerocks_slave_main.cpp
@@ -174,6 +174,7 @@ static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_
     son_slave_conf.backhaul_wireless_iface_filter_low =
         beerocks::string_utils::stoi(beerocks_slave_conf.sta_iface_filter_low[slave_num]);
     son_slave_conf.backhaul_wireless_iface_type = son_slave_conf.hostap_iface_type;
+    son_slave_conf.no_vendor_specific           = beerocks_slave_conf.no_vendor_specific == "1";
 
     // disable stopping on failure initially. Later on, it will be read from BPL as part of
     // cACTION_PLATFORM_SON_SLAVE_REGISTER_RESPONSE
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -110,7 +110,6 @@ bool slave_thread::init()
     LOG(INFO) << "Slave Info:";
     LOG(INFO) << "hostap_iface=" << config.hostap_iface;
     LOG(INFO) << "hostap_iface_type=" << config.hostap_iface_type;
-    LOG(INFO) << "ruid=" << config.radio_identifier;
 
     if (config.hostap_iface_type == beerocks::IFACE_TYPE_UNSUPPORTED) {
         LOG(ERROR) << "hostap_iface_type '" << config.hostap_iface_type << "' UNSUPPORTED!";
@@ -1791,7 +1790,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         }
 
         notification_out2->params() = notification_in->params();
-        notification_out2->ruid()   = network_utils::mac_from_string(config.radio_identifier);
+        notification_out2->ruid()   = hostap_params.iface_mac;
         LOG(TRACE) << "send ACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION";
         message_com::send_cmdu(backhaul_manager_socket, cmdu_tx);
 
@@ -2241,8 +2240,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             return false;
         }
 
-        channel_preference_tlv->radio_uid() =
-            network_utils::mac_from_string(config.radio_identifier);
+        channel_preference_tlv->radio_uid() = hostap_params.iface_mac;
 
         for (auto preference : preferences) {
             // Create operating class object
@@ -2280,7 +2278,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             }
         }
 
-        LOG(DEBUG) << "sending channel preference report for ruid=" << config.radio_identifier;
+        LOG(DEBUG) << "sending channel preference report for ruid=" << hostap_params.iface_mac;
 
         send_cmdu_to_controller(cmdu_tx);
 
@@ -2918,7 +2916,7 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
         request->local_gw()             = platform_settings.local_gw;
         request->sta_iface_filter_low() = config.backhaul_wireless_iface_filter_low;
         request->onboarding()           = platform_settings.onboarding;
-        request->ruid()                 = network_utils::mac_from_string(config.radio_identifier);
+        request->ruid()                 = hostap_params.iface_mac;
 
         LOG(INFO) << "ACTION_BACKHAUL_REGISTER_REQUEST local_master="
                   << int(platform_settings.local_master)
@@ -3237,7 +3235,7 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
         // Platform Configuration
         notification->low_pass_filter_on()   = config.backhaul_wireless_iface_filter_low;
         notification->enable_repeater_mode() = config.enable_repeater_mode;
-        notification->radio_identifier() = network_utils::mac_from_string(config.radio_identifier);
+        notification->radio_identifier()     = hostap_params.iface_mac;
 
         // Backhaul Params
         notification->backhaul_params().gw_ipv4 =
@@ -3863,8 +3861,8 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
         return false;
     }
     // Check if the message is for this radio agent by comparing the ruid
-    if (network_utils::mac_from_string(config.radio_identifier) != ruid->radio_uid()) {
-        LOG(DEBUG) << "not to me - ruid " << config.radio_identifier << " != " << ruid->radio_uid();
+    if (hostap_params.iface_mac != ruid->radio_uid()) {
+        LOG(DEBUG) << "not to me - ruid " << hostap_params.iface_mac << " != " << ruid->radio_uid();
         return true;
     }
 
@@ -4311,8 +4309,8 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
     for (auto channel_preference_tlv : cmdu_rx.getClassList<wfa_map::tlvChannelPreference>()) {
 
         const auto &ruid = channel_preference_tlv->radio_uid();
-        if (network_utils::mac_to_string(ruid) != config.radio_identifier) {
-            LOG(DEBUG) << "ruid_rx=" << ruid << ", son_slave_ruid=" << config.radio_identifier;
+        if (ruid != hostap_params.iface_mac) {
+            LOG(DEBUG) << "ruid_rx=" << ruid << ", son_slave_ruid=" << hostap_params.iface_mac;
             continue;
         }
 
@@ -4393,8 +4391,7 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
         return false;
     }
 
-    channel_selection_response_tlv->radio_uid() =
-        network_utils::mac_from_string(config.radio_identifier);
+    channel_selection_response_tlv->radio_uid() = hostap_params.iface_mac;
     channel_selection_response_tlv->response_code() =
         wfa_map::tlvChannelSelectionResponse::eResponseCode::ACCEPT;
 
@@ -4415,8 +4412,7 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
         return false;
     }
 
-    operating_channel_report_tlv->radio_uid() =
-        network_utils::mac_from_string(config.radio_identifier);
+    operating_channel_report_tlv->radio_uid() = hostap_params.iface_mac;
 
     auto op_classes_list = operating_channel_report_tlv->alloc_operating_classes_list();
     if (!op_classes_list) {
@@ -4453,7 +4449,7 @@ bool slave_thread::add_radio_basic_capabilities()
         LOG(ERROR) << "Error creating TLV_AP_RADIO_BASIC_CAPABILITIES";
         return false;
     }
-    radio_basic_caps->radio_uid() = network_utils::mac_from_string(config.radio_identifier);
+    radio_basic_caps->radio_uid() = hostap_params.iface_mac;
     //TODO get maximum supported VAPs from DWPAL
     radio_basic_caps->maximum_number_of_bsss_supported() = 2;
 
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -1946,21 +1946,25 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         client_association_event_tlv->association_event() =
             wfa_map::tlvClientAssociationEvent::CLIENT_HAS_LEFT_THE_BSS;
 
-        // Add vendor specific tlv
-        auto vs_tlv =
-            message_com::add_vs_tlv<beerocks_message::tlvVsClientAssociationEvent>(cmdu_tx);
+        if (config.no_vendor_specific) {
+            LOG(DEBUG) << "non-Intel, not adding ClientAssociationEvent VS TLV";
+        } else {
+            // Add vendor specific tlv
+            auto vs_tlv =
+                message_com::add_vs_tlv<beerocks_message::tlvVsClientAssociationEvent>(cmdu_tx);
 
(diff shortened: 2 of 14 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2383,21 +2383,25 @@ bool slave_thread::handle_cmdu_ap_manager_message(
         client_association_event_tlv->association_event() =
             wfa_map::tlvClientAssociationEvent::CLIENT_HAS_LEFT_THE_BSS;
 
-        // Add vendor specific tlv
-        auto vs_tlv =
-            message_com::add_vs_tlv<beerocks_message::tlvVsClientAssociationEvent>(cmdu_tx);
+        if (config.non_intel) {
+            LOG(DEBUG) << "non-Intel, skipping ClientAssociationEvent TLV";
```
rmelotte: It seems like you're not skipping the whole ClientAssociationEven TLV here, but only the VS part which is added afterwards.
If it's the case, the current debug message is confusing and should rather say something like "non-Intel, not adding the VS clientAssociationEvent TLV", or at least mention that it's the VS part that is being skipped.

arnout:
Turns out that this is actually *not* working at the moment. In Travis it works because the docker images haven't been updated, but once the docker images are updated, the repeater slaves are not detected as running.

Fix follows.

morantr:
I don't like this implementation. Adding a checker at all the places we want to add send VS CMDU for an internal test, seems wrong to me.
When running beerocks_slave, it is by definition an Intel agent.
I think that a more correct solution would be to use a compilation flag on the framework tlvf which will cause to the creation of VS CMDU/adding VS TLVs to not really occur (create shared ptr that is not allocated on the CMDU buffer), and run it as a different process.

arnout:
> I don't like this implementation. Adding a checker at all the places we want to add send VS CMDU for an internal test, seems wrong to me.
> When running beerocks_slave, it is by definition an Intel agent.
> I think that a more correct solution would be to use a compilation flag on the framework tlvf which will cause to the creation of VS CMDU/adding VS TLVs to not really occur (create shared ptr that is not allocated on the CMDU buffer), and run it as a different process.

I expect it's a little bit more complicated than it seems at first sight, but it's not a bad idea. So we would build map_agent in addition to beerocks_agent, and map_agent returns a shared ptr to a dummy object in `message_com::add_vs_tlv()`.

Note that it's not possible to do this for VS CMDUs, since those are used for internal messaging between the different threads. But that's not a problem since it turns out that with the non-Intel option (as it's implemented now), no VS CMDUs are sent over the network.

tools/docker/tests/test_gw_repeater.sh
```diff
@@ -76,7 +76,7 @@ main() {
         index=0
         for repeater in $REPEATER_NAMES; do
             status "Start Repeater (Remote Agent): $repeater"
-            ${scriptdir}/../run.sh ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${repeater} -m aa:bb:cc:$index$index "$@"
+            ${scriptdir}/../run.sh ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${repeater} -m aa:bb:cc:$index$index -- -n "$@"
```
tomereli: This leaves the Intel repeater mode vulnerable since it is not tested *at all*.
I think that a good compromise can be that the first agent is always Intel, and all the other agents are none-Intel.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3701,90 +3701,97 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
...
-        notification->backhaul_params().backhaul_is_wireless = backhaul_params.backhaul_is_wireless;
-
-        if (!config.bridge_iface.empty()) {
-            notification->backhaul_params().bridge_mac =
-                network_utils::mac_from_string(backhaul_params.bridge_mac);
-            notification->backhaul_params().bridge_ipv4 =
-                network_utils::ipv4_from_string(backhaul_params.bridge_ipv4);
-            notification->backhaul_params().backhaul_ipv4 =
-                network_utils::ipv4_from_string(backhaul_params.bridge_ipv4);
+        if (config.non_intel) {
```
tomereli: The idea behind supporting Intel controller is that each company can differentiate itself with extra features which are not supported in the standard. For example, DCS and DFS support can be negotiated during autoconfiguration:
The agent includes a VS TLV in the autoconfiguration CMDU which marks its extra capabilities.
prplMesh controller understands that, and adds its own VS TLV to the M2, just like we do today.

Coming to think about it, we have the manufacturer string in the M1 and M2 messages, so instead of using a global flag I suggest to read the manufacturer from the configuration file, and only if its Intel send the VS stuff, in both the controller and the agent.

So the flag you added would just replace the manufacturer in the configuration file from Intel (which should become the default) to something else.

While we're at it - lets make the manufacturer a CMAKE parameter, default being Intel.

tomereli (approved):
Mainly nitpicks, so approving now. I disagree with Moran's comment that he doesn't like the direction in which this is going but I'll answer that separately.

tomereli:
> I don't like this implementation. Adding a checker at all the places we want to add send VS CMDU for an internal test, seems wrong to me.

And yet it is not wrong - and we'll need it sooner than you might think, now that we're about to start agent certification and the controller in the testbed is definitely not Intel.

> When running beerocks_slave, it is by definition an Intel agent.

This is simply not true. It is an open source implementation based on Intel source code, but see for yourself - plenty of companies are now integrating prplMesh to use it as their EasyMesh agent.

> I think that a more correct solution would be to use a compilation flag on the framework tlvf which will cause to the creation of VS CMDU/adding VS TLVs to not really occur (create shared ptr that is not allocated on the CMDU buffer), and run it as a different process.

Again, I disagree - using a compilation flag for the agent is much less generic IMHO. Also, the creation of VS CMDUs should still happen even for non-Intel agents if other implementations decide to use the Intel algorithms.

Bottom line, apart for my comments I think this is going in the right direction and i'd like to see it merged soon.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: > For example, DCS and DFS support can be negotiated during autoconfiguration:

DFS is an R2 feature so not a good example :-)

> The agent includes a VS TLV in the autoconfiguration CMDU which marks its extra capabilities.
> prplMesh controller understands that, and adds its own VS TLV to the M2, just like we do today.

I think we'll want to emulate a non-Intel controller as well, i.e. controller would ignore the VS TLV.
Conversely, there is no actual need for the controller to remove the VS TLVs when talking to a non-Intel agent, because the agent should ignore unknown VS TLVs.


> Coming to think about it, we have the manufacturer string in the M1 and M2 messages, so instead of using a global flag I suggest to read the manufacturer from the configuration file, and only if its Intel send the VS stuff, in both the controller and the agent.

As you note elsewhere: integrators might want to change the manufacturer string but still use the extra features.


> So the flag you added would just replace the manufacturer in the configuration file from Intel (which should become the default) to something else.

I think the config file should be removed entirely, and instead bpl (or something similar) should be used to tie in with the system's config store. Well, config file *may* still be useful for debugging/experimentation options, but for those I'd actually prefer them to come through bml.

> While we're at it - lets make the manufacturer a CMAKE parameter, default being Intel.

I think the default should be prpl :-)

morantr:
> > When running beerocks_slave, it is by definition an Intel agent.
> 
> This is simply not true. It is an open source implementation based on Intel source code, but see for yourself - plenty of companies are now integrating prplMesh to use it as their EasyMesh agent.

I meant that if it runs beerocks slave, it will use Intel VS messages.

tomereli:
> > > When running beerocks_slave, it is by definition an Intel agent.
> > 
> > 
> > This is simply not true. It is an open source implementation based on Intel source code, but see for yourself - plenty of companies are now integrating prplMesh to use it as their EasyMesh agent.
> 
> I meant that if it runs beerocks slave, it will use Intel VS messages.

That's not true, IMO the agent in certification mode should not send any VS messages. It should know how to work with a non-Intel controller so why send the VS messages at all? Bottom line, this should be done based on the manufacturer or some other method which during autoconfig identifies both agent and controller as Intel or non-Intel.

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -505,6 +514,9 @@ int main(int argc, char *argv[])
         }
     }
 
+    //Handle -n option (non-intel agent)
+    beerocks_slave_conf.non_intel = g_non_intel;
```
morantr: I find it weird to save the execution flag where the configurations from the configuration file are.
IMO, adding it to the constructor of the slave is the right way to go.
Maybe even add a struct of execution flags to the constructor.
Another option is to add this field on the config file instead of using flag, and whoever runs the agent, will change the config file field value, before running the agent.

morantr (changes requested):

tomereli:
[Non-Intel controller support](https://trello.com/c/7kJiE6W0/17-non-intel-controller-support)

tomereli:
[Revive non Intel mode for certification](https://trello.com/c/uF5h9akO/69-revive-non-intel-mode-for-certification)

agent/src/beerocks/slave/beerocks_slave_main.cpp
tomereli: Moran - please remove your review. We will merge this as is since it blocks our progress in passing 4.2.1 testcase, the rest of the comments will be discussed in architecture calls and fixed if needed in a different PR.

arnout:
I made the following changes:

- rebased on master (that was a bit of a challenge :-)
- instead of command-line option, make it an option in the config file;
- the -n option to prplmesh-utils.sh now creates a custom config file in /tmp to set that option;
- I called it no_vendor_specific instead of non_intel.

The tests fail now, because the slaves are not properly tracked in the bml_conn_map. This is because of the (IMO incorrect) distinction between RUID and radio MAC, and the RUID is used as Radio ID in the EasyMesh messages. However, I believe @CoralMalachi and @tomereli encountered this same issue in the 4.2.1 certification test, so was this solved?

tomereli:
> The tests fail now, because the slaves are not properly tracked in the bml_conn_map. This is because of the (IMO incorrect) distinction between RUID and radio MAC, and the RUID is used as Radio ID in the EasyMesh messages. However, I believe @CoralMalachi and @tomereli encountered this same issue in the 4.2.1 certification test, so was this solved?

The issue was not related to the RUID - we need to fix `bml_conn_map` if it is now not showing the correct status.

arnout:
It is working now, but still needs some cleanups. That's for tomorrow.

arnout:
> We should still use the connmap IMO, since in real platform we won't have any other way to get this from.

Certification setup gets this information from its config file. So we probably should use the same approach - eventually.

> I think we should extend dummy mode so it will show the repeaters AL Mac and radios with some easy to grep prefixes.
> So for the 5g interface we can add 5ghz to mark it as 5ghz.

I don't know what you mean with this. The only per-radio information the controller receives (without VS TLV) is:

- radio ID;
- band (though some golden devices don't seem to set this correctly);
- other M1 information over which we have no control (most golden devices seem to use the same for all radios).

Am I missing something?

arnout:
> This means that the radio ID now has to match the radio Mac,

Why does it mean that? Things seems to work perfectly like this. Radio ID is still used internally in the agent, but communication with controller now uses radio MAC everywhere (except in the VS message where the radio ID is still communicated).

> so either remove config.radio_identifier completely so we won't need to manually set the RUID in the configuration file every time we flash a new version, or keep using the radio identifier in the code.

It is still used internally in the code. Removing it completely is a bigger job.

The radio ID is not needed for anything. We can remove it. It's just a bit of work.

morantr (approved):

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2241,8 +2241,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             return false;
         }
 
-        channel_preference_tlv->radio_uid() =
-            network_utils::mac_from_string(config.radio_identifier);
+        channel_preference_tlv->radio_uid() = hostap_params.iface_mac;
```
tomereli, line 2251: This will cause failures in the testbed unless we set the RUID in the UCC configuration file and the beerocks_agent.conf to be the interface MAC address, since the UCC will look for the RUID configured in the file in the channel preference TLV.

Bottom line - if we use the radio MAC as RUID, we need to do this everywhere. Otherwise things will break. Please do not merge this commit until we verify it in the testbed.

tomereli (changes requested):
Requesting changes only to make sure this doesn't get accidentally merged - since I honestly think it will break in the certification testbed, see comment inline.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Indeed, you also have to change the config file in the testbed.

I understood that this all-zeroes RUID was in fact an issue in the testbed, so I thought this would fix something. It is not strictly needed for the non-Intel agent test (I had an issue that I thought was caused by this, but it turned out to be something else), so I could also leave out this commit from the PR.

arnout: Turns out that the correct RUID change is needed after all to make this PR work. So please adapt the certification setup config file so we can merge this PR.

tomereli:
In this case I think we should completely remove the RUID from the config file and use the radio Mac automatically everywhere. This will require only a one time change to the UCC laptop config file but we won't need to manually set the beerocks_agent.conf on each version update.
Not mandatory for this PR so I'll approve, but it would be much better to do this soon.

tomereli (approved):

arnout:
> In this case I think we should completely remove the RUID from the config file and use the radio Mac automatically everywhere.

Agreed -> #691 

> This will require only a one time change to the UCC laptop config file

Yes, that needs to be changed as soon as this gets merged (and back again if you want to use an older version/branch). @CoralMalachi @Mila-Sh please take note of this.

> but we won't need to manually set the beerocks_agent.conf on each version update.

For now, no change to beerocks_agent.conf is needed. Only when the RUID is removed completely, it also has to be removed from the conf file (IIRC the agent terminates if it finds an unkown config option).

> Not mandatory for this PR so I'll approve, but it would be much better to do this soon.

True, but I'll first work on the controller certification issues...

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: This will still fail in `dev_get_parameter` since in son_slave_thread `ACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION` we are still using the `config.radio_identifier` - so the VAPs list ruid will be wrong, which will be propagated to the agent ucc listener which eventually won't find the real RUID (the one from the ucc AllInitConfig.txt file) in the VAP list.

```
case beerocks_message::ACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION: {
        auto notification_in = beerocks_header->addClass<
            beerocks_message::cACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION>();
        if (notification_in == nullptr) {
            LOG(ERROR) << "addClass cACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION failed";
            return false;
        }

        LOG(INFO) << "received ACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION";

        auto notification_out = message_com::create_vs_message<
            beerocks_message::cACTION_CONTROL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION>(cmdu_tx);
        if (notification_out == nullptr) {
            LOG(ERROR) << "Failed building message!";
            return false;
        }

        notification_out->params() = notification_in->params();
        LOG(TRACE) << "send ACTION_CONTROL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION";
        send_cmdu_to_controller(cmdu_tx);

        auto notification_out2 = message_com::create_vs_message<
            beerocks_message::cACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION>(cmdu_tx);
        if (notification_out2 == nullptr) {
            LOG(ERROR) << "Failed building message!";
            return false;
        }

        notification_out2->params() = notification_in->params();
        notification_out2->ruid()   = network_utils::mac_from_string(config.radio_identifier);
        LOG(TRACE) << "send ACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION";
        message_com::send_cmdu(backhaul_manager_socket, cmdu_tx);

        break;
    }
```

arnout: OK fair enough, I'll remove it completely in this PR then.

PR #438, opened 2019-10-16
Fix a documentation and logger issue with beerocks_analyzer

rmelotte:
The documentation was missing a dollar sign for one of the Bash variables.

There was also an issue with the usage of the new str.format style in the logger messages.

### DIFF
--- a/tools/beerocks_analyzer/beerocks_analyzer.py
+++ b/tools/beerocks_analyzer/beerocks_analyzer.py
@@ -774,7 +774,7 @@ class BeerocksCliThread(threading.Thread):
         command = [str(self.beerocks_cli_path), "-a", str(self.target_ip)]
         if self.docker_container_name:
             command = ["docker", "exec", self.docker_container_name] + command
-        self.logger.debug("Starting '%s'", ' '.join(command))
+        self.logger.debug("Starting '{}'", ' '.join(command))
         self.process = subprocess.Popen(command)
         outs = None
         errs = None
--- a/tools/beerocks_analyzer/logger_setup.py
+++ b/tools/beerocks_analyzer/logger_setup.py
@@ -15,3 +15,19 @@ def setup_logger():
     with open('logger_config.yaml', 'rt') as file:
         config = yaml.safe_load(file.read())
         logging.config.dictConfig(config)
+        logging.setLogRecordFactory(FormatRecordFactory)
+
+
+class FormatRecordFactory(logging.LogRecord):
+    """A factory which formats messages with str.format."""
+    def getMessage(self):
+        """
+        Return the message for this LogRecord, formatted with str.format.
+
+        Return the message for this LogRecord after merging any user-supplied
+        arguments with the message by using str.format.
+        """
+        msg = str(self.msg)
+        if self.args:
+            msg = msg.format(self.args)
+        return msg
--- a/tools/beerocks_analyzer/README.md
+++ b/tools/beerocks_analyzer/README.md
@@ -12,6 +12,7 @@ Using a virtualenv allows to install the dependencies in a local directory, whic
 
 You only need to create the `venv` once. Later on you just need to activate the `venv` to be able to use the application.
 
+Note for *Windows* users: you may need to run the command-line as an administrator depending on your setup.
 
 ### Create a virtualenv to use the application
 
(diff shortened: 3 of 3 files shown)


morantr:
While I tried to work on the analyzer from windows I got the following insights that can be added to the readme.md file:
1. run command line as an administrator (in order to pip install would work).
2. when creating python environment, use `python` instead of `python3`
3. when activating python enviorment on windows, `source` will no work, and need to do instead:
`analyzer-venv\Scripts\activate.bat`

as well the line:
 `command = 'LD_LIBRARY_PATH=/opt/beerocks/lib /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)` on beerocks_analyzer.py 
should be changed to : 
`command = 'export LD_LIBRARY_PATH=/opt/beerocks/lib && /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)`

Could you add it, please?

arnout:
> `command = 'LD_LIBRARY_PATH=/opt/beerocks/lib /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)` on beerocks_analyzer.py
> should be changed to :
> `command = 'export LD_LIBRARY_PATH=/opt/beerocks/lib && /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)`

Why is this change needed? Both commands have exactly the same effect. If they don't, there's something fishy going on.

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -27,6 +27,7 @@
 import struct
 
 import logger_setup
+from logger_setup import setup_logger, getLogger
```
arnout: It usually doesn't make sense to have both `import foo` and `from foo import bar`.

morantr:
> > `command = 'LD_LIBRARY_PATH=/opt/beerocks/lib /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)` on beerocks_analyzer.py
> > should be changed to :
> > `command = 'export LD_LIBRARY_PATH=/opt/beerocks/lib && /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)`
> 
> Why is this change needed? Both commands have exactly the same effect. If they don't, there's something fishy going on.

@arnout, I agree that "&&" will lead to the same result, but I'm not sure about the "export".
But I'm not a linux expert so...

tools/beerocks_analyzer/logger_setup.py
```diff
@@ -15,3 +15,39 @@ def setup_logger():
...
+
+
+class StyleAdapter(logging.LoggerAdapter):
+    def __init__(self, logger, extra=None):
+        super(StyleAdapter, self).__init__(logger, extra or {})
+
+    def log(self, level, msg, *args, **kwargs):
+        if self.isEnabledFor(level):
+            msg, kwargs = self.process(msg, kwargs)
+            self.logger._log(level, Message(msg, args), (), **kwargs)
```
arnout: The [Python documentation](https://docs.python.org/3/howto/logging-cookbook.html#using-logrecord-factories) suggests two approaches for allowing `{}`-formatted strings for log messages, and this isn't one of them. The second option suggested by Python docs is not really applicable since it requires all calls to be updated. But the first option, using the logrecord factory and overriding getMessage, seems far simpler than this approach.

That said, I'm not sure if we need to spend time "optimising" that.

arnout (approved):
Nitpicks only so I'll approve already. @morantr's comment about the windows-specific documentation is relevant so I'll wait for that before merging.

arnout:
> > > `command = 'LD_LIBRARY_PATH=/opt/beerocks/lib /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)` on beerocks_analyzer.py
> > > should be changed to :
> > > `command = 'export LD_LIBRARY_PATH=/opt/beerocks/lib && /opt/beerocks/bin/beerocks_cli -a {}'.format(self.my_ip)`
> > 
> > 
> > Why is this change needed? Both commands have exactly the same effect. If they don't, there's something fishy going on.
> 
> @arnout, I agree that "&&" will lead to the same result, but I'm not sure about the "export".
> But I'm not a linux expert so...

I am :-)

Putting a variable assignment before a command is equivalent to exporting it in the environment for just that command. Try `FOO=foo env | grep FOO`.

tools/beerocks_analyzer/logger_setup.py
rmelotte: Weird, there is another section (from which I took the implementation) about the same topic [here](https://docs.python.org/3/howto/logging-cookbook.html#use-of-alternative-formatting-styles).
Anyway setting another LogRecordFactory is indeed much simpler, so I switched to that.

morantr (approved):

Issue #430, opened 2019-10-11
[TASK] BWL sta_deny should set proper reason


arnout:
## Description

The Intel VS message to deny a client contained a 'reject_sta' field with the reason code that should be given to the client. However, EasyMesh specifies no such field. Also, it is something that should be handled by the agent according to the specification.

>  it shall set the Status Code in the Authentication frame or (Re-)Association Response frame to a value that
> does not indicate capabilities mismatch or negotiation failure, and should set the Status Code to value 33 (denied due to
> insufficient bandwidth) or 34 (denied due to poor channel conditions).

Therefore, the reject_sta field should be removed, and it should be handled inside the BWL implementation.

## Exit criteria

- reject_sta field is removed.
- AP still uses proper reject code.

tomereli:
@CoralMalachi What's the status with this issue?

CoralMalachi:
Done @tomereli

PR #439, opened 2019-10-16
Feature/handle reject status by agent

CoralMalachi:
The Intel VS message to deny a client contained a 'reject_sta' field with the reason code that should be given to the client. However, EasyMesh specifies no such field. Also, it is something that should be handled by the agent according to the specification.

`
it shall set the Status Code in the Authentication frame or (Re-)Association Response frame to a value that
does not indicate capabilities mismatch or negotiation failure, and should set the Status Code to value 33 (denied due to
insufficient bandwidth) or 34 (denied due to poor channel conditions).
`

Therefore, the following changes were made:
the reject_sta field is removed
set proper reject code inside the BWL implementation of sta_deny method.

#430


common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -370,14 +370,10 @@ bool ap_wlan_hal_dwpal::sta_allow(const std::string &mac)
...
-bool ap_wlan_hal_dwpal::sta_deny(const std::string &mac, int reject_sta)
+bool ap_wlan_hal_dwpal::sta_deny(const std::string &mac)
 {
     // Build command string
-    std::string cmd = "DENY_MAC " + mac;
-
-    if (reject_sta) {
-        cmd += " reject_sta=" + std::to_string(reject_sta);
-    }
+    std::string cmd = "DENY_MAC " + mac + " reject_sta=33";
```
arnout, line 376: @morantr Can you confirm that it's OK to always use 33 here?

arnout (approved):
I think we can actually remove the vs tlv entirely, because the ipv4 field isn't used either (only for a log message). But we can already merge this PR as is.

Before merging, though, I'd like to be really sure that on RDK-B it's also OK to always set the reject reason to 33. I hope @morantr can confirm that, or @kantera800.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
morantr: On the client steering task, you removed the following.
```cpp
        /*
        * Only for the original hostap, add error code "33" to signal hostap that we want to continue 
        * to hear BTM response from client.
        */
        vs_tlv->reject_sta() = (hostap == original_radio_mac) ? 33 : 0;
```
So previous we set the reject code to 33 only on the radio the client was connected to, and now we set it to all hostaps. I gotta say I quite don't understand the comment nither the what the standard says about it. 😕

arnout: Well, status code 0 makes no sense: it means "success", so it would mean to NOT reject the client. However, in the bwl implementation, reject reason 0 means "let hostapd invent its own reject reason". I don't know what reject reason hostapd will invent in that case, but basically, it will be either 33 (in which case we could just as well have passed it explicitly) or something else, in which case it is wrong.

So I *think* passing reject code 33 is always the right thing to do.

However, I hope that someone at Intel can give a definitive answer on this. In other words: why did the Intel code not give a reject reason for any AP except the one it's currently connected to? Maybe @vitalybu knows?

If we want to preserve current behaviour, we could move this condition into bwl itself - but that would require bwl to keep track of the currently associated STAs.

vitalybu: Not 100% sure, but I vaguely remember that disassociating a station with a non 33 reason code, in some cases caused an Authentication Error on the STA (Android? Samsung?).
When this happened, the station marked the network (SSID) as invalid and didn't try reconnecting to it at all.
Perhaps it's fine to always use 33 as the reject code.

morantr (approved):

PR #440, opened 2019-10-17
Feature/add client association dummy flow test

CoralMalachi:
Add a test that triggers the sending of a Client Association Control Request
message using dummy BWL.

### DIFF
--- a/tools/docker/tests/test_flows.sh
+++ b/tools/docker/tests/test_flows.sh
@@ -350,16 +350,16 @@ test_init() {
     mac_agent2=$(grep "IRE_BRIDGE" "$connmap" | sed -n 2p | awk '{print $5}' | cut -d ',' -f 1)
     dbg "mac_agent2 = ${mac_agent2}"
 
-    mac_agent1_wlan0=$(mac_to_hex $(grep "RADIO: wlan0" "$connmap" | head -1 | awk '{print $4}' | cut -d ',' -f 1))
+    mac_agent1_wlan0=$(grep "RADIO: wlan0" "$connmap" | head -1 | awk '{print $4}' | cut -d ',' -f 1)
     dbg "mac_agent1_wlan0 = ${mac_agent1_wlan0}"
 
-    mac_agent2_wlan0=$(mac_to_hex $(grep "RADIO: wlan0" "$connmap" | sed -n 2p | awk '{print $4}' | cut -d ',' -f 1))
+    mac_agent2_wlan0=$(grep "RADIO: wlan0" "$connmap" | sed -n 2p | awk '{print $4}' | cut -d ',' -f 1)
     dbg "mac_agent2_wlan0 = ${mac_agent2_wlan0}"
 
-    mac_agent1_wlan2=$(mac_to_hex $(grep "RADIO: wlan2" "$connmap" | head -1 | awk '{print $4}' | cut -d ',' -f 1))
+    mac_agent1_wlan2=$(grep "RADIO: wlan2" "$connmap" | head -1 | awk '{print $4}' | cut -d ',' -f 1)
     dbg "mac_agent1_wlan2 = ${mac_agent1_wlan2}"
 
-    mac_agent2_wlan2=$(mac_to_hex $(grep "RADIO: wlan2" "$connmap" | sed -n 2p | awk '{print $4}' | cut -d ',' -f 1))
+    mac_agent2_wlan2=$(grep "RADIO: wlan2" "$connmap" | sed -n 2p | awk '{print $4}' | cut -d ',' -f 1)
     dbg "mac_agent2_wlan2 = ${mac_agent2_wlan2}"
 
 }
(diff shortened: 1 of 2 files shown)


tools/docker/tests/test_flows.sh
```diff
@@ -180,6 +180,48 @@ test_client_steering_mandate() {
...
+    check docker exec -it repeater1 sh -c \
+        'echo "AP-STA-CONNECTED 11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan0/EVENT'
+    
+    dbg "Send client association control request to the chosen BSSID to steer the client (UNBLOCK) "
+    eval send_bml_command "client_allow \"11:22:33:44:55:66 aa:bb:cc:00:00:20\"" $redirect
+    sleep 1
+
+    dbg "Confirming Client Association Control Request message was received (UNBLOCK)"
+    check docker exec -it repeater1 sh -c \
+        'grep -i -q "Got client allow request" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
```
arnout: Please also check the MAC address.

tools/docker/tests/test_flows.sh
```diff
@@ -180,6 +180,48 @@ test_client_steering_mandate() {
...
+
+    dbg "Send client association control request to all other (BLOCK) "
+    eval send_bml_command "client_disallow \"11:22:33:44:55:66 aa:bb:cc:00:00:10\"" $redirect
```
arnout: I would prefer to use `$mac_agent1_wlan0` here instead of an explicit MAC address. However, at the moment `$mac_agent1_wlan0` is in hex notation instead of MAC notation. But I think that since commit 1cd4619d62c44 it is no longer needed to use the hex notation. If that is true, I would like a first commit that removes the mac_to_hex for the radio addresses, and then use `$mac_agent1_wlan0` here.

tools/docker/tests/test_flows.sh
```diff
@@ -180,6 +180,48 @@ test_client_steering_mandate() {
         'grep -i -q "ACK_MESSAGE" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
 }
 
+test_client_association_dummy(){
+    status "test client association dummy"
+    check_error=0
+
+    dbg "Connect dummy STA to wlan0"
+    check docker exec -it repeater1 sh -c \
+        'echo "AP-STA-CONNECTED 11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan0/EVENT'
```
arnout: Please make a variable for the STA address and use it below as well. Also, use a different address than in the other tests, to make sure we can distinguish the different tests in the log file.

tools/docker/tests/test_flows.sh
```diff
@@ -180,6 +180,48 @@ test_client_steering_mandate() {
...
+    dbg "Send client association control request to all other (BLOCK) "
+    eval send_bml_command "client_disallow \"11:22:33:44:55:66 aa:bb:cc:11:00:10\"" $redirect
+    sleep 1
+
+    dbg "Confirming Client Association Control Request message was received (BLOCK)"
+    check docker exec -it repeater2 sh -c \
+        'grep -i -q "Got client disallow request" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+
+    dbg "Send client association control request to all other (BLOCK) "
+    eval send_bml_command "client_disallow \"11:22:33:44:55:66 aa:bb:cc:11:00:20\"" $redirect
```
arnout: I don't think the third test adds a lot of value. It is already covered by the previous one.

arnout (changes requested):
Nitpick stuff only, but still I'd like them to be fixed.

tools/docker/tests/test_flows.sh
CoralMalachi: I agree with you. Do you think we need to remove the mac_to_hex method entirely ? from functions.sh also? @arnout

arnout: > Do you think we need to remove the mac_to_hex method entirely ?

If it is no longer used, yes we can remove it. However, it might still be needed in some other context (i.e. outside of a tlv).

arnout: > > Do you think we need to remove the mac_to_hex method entirely ?
> 
> If it is no longer used, yes we can remove it. However, it might still be needed in some other context (i.e. outside of a tlv).

It is still used in `test_higher_layer_data_payload_trigger` and I think it's appropriate to keep that use because the UCC generates it like that (i.e. hex notation, not colon notation). So no, don't remove it from `functions.sh`.

arnout (approved):
Two small nits:

- I think it's more appropriate to mention `test_flows.sh` in the summary line than docker.
- It would be better to split off the change in `test_client_steering_dummy` in a separate commit.

Neither of them is particularly important though. I might do those fixups while rebasing on master. Or not :-)

CoralMalachi:
> Two small nits:
> 
> * I think it's more appropriate to mention `test_flows.sh` in the summary line than docker.
> * It would be better to split off the change in `test_client_steering_dummy` in a separate commit.
> 
> Neither of them is particularly important though. I might do those fixups while rebasing on master. Or not :-)

I'll do it while rebasing :)

Issue #432, opened 2019-10-14
[TASK] Change access control validity period value


CoralMalachi:
## Description
Currently, the client steering task sets the access control validity period to disassoc_timer in the client blocking request. However, the disassoc_timer is expressed in TU while according to specification the validity period is expressed in seconds.
So there is a need to fix the following line which of course is not correct :

![image](https://user-images.githubusercontent.com/38204874/66756021-d3149e80-eea1-11e9-80c2-6f820594583b.png)

In addition, according to section 11.6 in the specification, the  Multi-AP agent shall reject 
a first attempt by a STA specified in the request to associate to the specified BSS operated by the Multi-AP Agent, only if the following condition is true : 

`
The time delta since the message was received is less than the value of the validity period field in the Client Association Control Request message.
`

For that reason, the validity period can be set to the timeout used by the client steering task ( i.e. steering_wait_time_ms/1000)

tomereli:
@CoralMalachi / @arnout - Can this issue be closed?

CoralMalachi:
Yes @tomereli

PR #441, opened 2019-10-17
controller: change disassociation timer value

CoralMalachi:
Currently, the client steering task sets the access control validity period to disassoc_timer in the client blocking request. However, the disassoc_timer is expressed in TU while according to specification the validity period is expressed in seconds.

In addition, according to section 11.6 in the specification, the Multi-AP agent shall reject
a first attempt by a STA specified in the request to associate to the specified BSS operated by the Multi-AP Agent, only if the following condition is true :

`
The time delta since the message was received is less than the value of the validity period field in the Client Association Control Request message.
`

For that reason, set the validity period to the timeout used by the client steering task
(i.e. steering_wait_time_ms/1000).

#432 

### DIFF
--- a/controller/src/beerocks/master/tasks/client_steering_task.cpp
+++ b/controller/src/beerocks/master/tasks/client_steering_task.cpp
@@ -191,7 +191,7 @@ void client_steering_task::steer_sta()
             network_utils::mac_from_string(current_ap_mac);
         association_control_block_request_tlv->association_control() =
             wfa_map::tlvClientAssociationControlRequest::BLOCK;
-        association_control_block_request_tlv->validity_period_sec() = disassoc_timer_ms;
+        association_control_block_request_tlv->validity_period_sec() = steering_wait_time_ms / 1000;
         association_control_block_request_tlv->alloc_sta_list();
         auto sta_list_block         = association_control_block_request_tlv->sta_list(0);
         std::get<1>(sta_list_block) = network_utils::mac_from_string(sta_mac);


controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -75,8 +75,6 @@ void client_steering_task::work()
                 database.update_node_failed_24ghz_steer_attempt(sta_mac);
             }
         }
-
-        finish();
```
morantr: Why did you remove it? This line has ended the task.

morantr (changes requested):

controller/src/beerocks/master/tasks/client_steering_task.cpp
CoralMalachi: you are right I'll fix that.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -297,7 +297,4 @@ void client_steering_task::handle_task_end()
         database.update_node_11v_responsiveness(sta_mac, false);
     }
     database.set_node_handoff_flag(sta_mac, false);
-    if (!steer_restricted) {
-        son_actions::unblock_sta(database, cmdu_tx, sta_mac);
```
morantr: Why did you remove it? please explain.

morantr (changes requested):

controller/src/beerocks/master/tasks/client_steering_task.cpp
CoralMalachi: In the FINALIZE state, the controller sent an unblock message to all agents, but it's not really needed due to the validity period, so I removed that part. @morantr

arnout: > In the FINALIZE state, the controller sent an unblock message to all agents, but it's not really needed due to the validity period, so I removed that part. @morantr

The current bwl implementations don't use the validity period option, so that will have to be implemented first.

I think it's best if you un-remove this code for now, and create a new task to actually use the validity period in bwl. I shouldn't have included that bit in #432 - my bad.

CoralMalachi: Ok 👍 @arnout

arnout (approved):
Small nit about the commit message: the backquote that you used to indicate the text quoted from the specification is not very visually pleasing when reading it as text. It would be better to block-indent the quoted text, or maybe even add `> ` quotation markers like in e-mail.

But I can fix that up while I rebase on master before the merge.

CoralMalachi:
> Small nit about the commit message: the backquote that you used to indicate the text quoted from the specification is not very visually pleasing when reading it as text. It would be better to block-indent the quoted text, or maybe even add `> ` quotation markers like in e-mail.
> 
> But I can fix that up while I rebase on master before the merge.

I'll do rebase now and try what you suggested 👍

morantr (approved):

controller/src/beerocks/master/tasks/client_steering_task.cpp
CoralMalachi: #445

PR #450, opened 2019-10-27
autoconfig: WSC: Change config data network key to be dynamic

### LINKED ISSUE #449: [BUG] <tlvf>: <M2 is using a fixed-length password>
**Describe the bug** -
Currently, test 5.6.2 keeps failing when using Marvell as CTT Agent.

The reason for this is that we send the M2 with a fixed-length attribute
for the password. In cases that the actual password is shorter than the maximum, 
the agent may use zeroes at the end.

The solution is to make the sWscAttrNetworkKey attribute to be a dynamic size, to prevent an agent from adding zeroes at the end.

CoralMalachi:
Currently, test 5.6.2 keeps failing when using Marvell as CTT Agent.

The reason for this is that we send the M2 with a fixed-length attribute
for the password. In cases that the actual password is shorter than the maximum,
the agent may use zeroes at the end.

In order to solve this, the network_key attribute is added manually to cConfigData in the WSC_Attributes YAML file.
Also, sWscAttrNetworkKey is removed since it is not needed anymore.

Fixes #449 

### DIFF
--- a/framework/tlvf/test/tlvf_test.cpp
+++ b/framework/tlvf/test/tlvf_test.cpp
@@ -216,14 +216,16 @@ bool add_encrypted_settings(std::shared_ptr<tlvWscM2> m2, uint8_t *keywrapkey, u
     config_data.set_ssid("test_ssid");
     config_data.authentication_type_attr().data = WSC::eWscAuth::WSC_AUTH_WPA2; //DUMMY
     config_data.encryption_type_attr().data     = WSC::eWscEncr::WSC_ENCR_AES;
-    std::fill(config_data.network_key_attr().data,
-              config_data.network_key_attr().data + config_data.network_key_attr().data_length,
-              0xaa); //DUMMY
+    config_data.set_network_key("test1234");
 
     LOG(DEBUG) << "WSC config_data:" << std::endl
                << "     ssid: " << config_data.ssid() << std::endl
                << "     authentication_type: " << int(config_data.authentication_type_attr().data)
                << std::endl
+               << "     network_key: "
+               << std::string(config_data.network_key(), config_data.network_key_length())
+               << std::endl
+               << "     network_key length: " << int(config_data.network_key_length()) << std::endl
                << "     encryption_type: " << int(config_data.encryption_type_attr().data)
                << std::endl;
     config_data.class_swap();
--- a/framework/tlvf/test/tlvf_test.cpp
+++ b/framework/tlvf/test/tlvf_test.cpp
@@ -168,7 +168,7 @@ int test_complex_list()
         MAPF_ERR("TLV4 is NULL");
         return ++errors;
     }
-    if (!tlv4->test_string_str().compare("1234567")) {
+    if (tlv4->test_string_str().compare("1234567")) {
         MAPF_ERR("FAIL, expected  \"1234567\", received " << tlv4->test_string_str());
         errors++;
     }
@@ -196,7 +196,7 @@ int test_complex_list()
 
     auto str = std::string(tlv4->var1()->unknown_length_list_inner(),
                            tlv4->var1()->unknown_length_list_inner_length());
-    if (!str.compare("prplMesh")) {
+    if (str.compare("prplMesh")) {
         MAPF_ERR("unknown length list failure - expected \"prplMesh\", received " << str);
         errors++;
     }
--- a/framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
+++ b/framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
@@ -330,18 +330,6 @@ sWscAttrEncryptionType:
     _type: eWscEncr
     _value: eWscEncr::WSC_ENCR_AES
 
-sWscAttrNetworkKey:
-  _type: struct
-  attribute_type:
-    _type: eWscAttributes
-    _value: ATTR_NETWORK_KEY
-  data_length:
-    _type: uint16_t
-    _value: WSC_MAX_NETWORK_KEY_LENGTH
-  data:
-    _type: char
-    _length: [ WSC_MAX_NETWORK_KEY_LENGTH ]
-
 sWscAttrBssid:
   _type: struct
   attribute_type:
@@ -379,7 +367,16 @@ cConfigData:
     _length_max: WSC_MAX_SSID_LENGTH
   authentication_type_attr: sWscAttrAuthenticationType
   encryption_type_attr: sWscAttrEncryptionType
-  network_key_attr: sWscAttrNetworkKey
+  network_key_type:
+    _type: eWscAttributes
+    _value: ATTR_NETWORK_KEY
+  network_key_length:
+    _type: uint16_t
+    _length_var: True
+  network_key:
+    _type: char
+    _length: [ network_key_length ]
+    _length_max: WSC_MAX_NETWORK_KEY_LENGTH
   bssid_attr: sWscAttrBssid
   multiap_attr: sWscAttrVendorExtMultiAp
 
--- a/common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_backhaul.cpp
+++ b/common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_backhaul.cpp
@@ -39,30 +39,17 @@ char* cACTION_BACKHAUL_REGISTER_REQUEST::sta_iface(size_t length) {
     return ((char*)m_sta_iface);
 }
 
-bool cACTION_BACKHAUL_REGISTER_REQUEST::set_sta_iface(const std::string& str) {
-    size_t str_size = str.size();
-    if (str_size == 0) {
-        TLVF_LOG(WARNING) << "set_sta_iface received an empty string.";
-        return false;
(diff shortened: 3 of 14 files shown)


framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -365,6 +354,7 @@ sWscAttrAuthenticator:
     _type: char
     _length: [ WSC_AUTHENTICATOR_LENGTH ]
 
+
```
tomereli: extra line break

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -379,10 +369,20 @@ cConfigData:
...
+    _type: uint16_t
+    _length_var: True
+  network_key:
+    _type: char
+    _length: [ network_key_length ]
+    _length_max: WSC_MAX_NETWORK_KEY_LENGTH
   bssid_attr: sWscAttrBssid
   multiap_attr: sWscAttrVendorExtMultiAp
 
+
```
tomereli: extra line break

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -216,9 +216,7 @@ bool add_encrypted_settings(std::shared_ptr<tlvWscM2> m2, uint8_t *keywrapkey, u
     config_data.set_ssid("test_ssid");
     config_data.authentication_type_attr().data = WSC::eWscAuth::WSC_AUTH_WPA2; //DUMMY
     config_data.encryption_type_attr().data     = WSC::eWscEncr::WSC_ENCR_AES;
-    std::fill(config_data.network_key_attr().data,
-              config_data.network_key_attr().data + config_data.network_key_attr().data_length,
-              0xaa); //DUMMY
+    config_data.set_network_key("test1234");
```
tomereli, line 219: This should be done in a separate commit which sets a short password and compares it in the parse, making the test fail.

tomereli (approved):

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -274,16 +273,26 @@ bool parse_encrypted_settings(std::shared_ptr<tlvWscM2> m2, uint8_t *keywrapkey,
...
+
     LOG(DEBUG) << "authenticator type:" << m2->authenticator().attribute_type;
     LOG(DEBUG) << "authenticator length:" << m2->authenticator().data_length;
     LOG(DEBUG) << "authenticator buffer: " << std::endl
                << dump_buffer((uint8_t *)m2->authenticator().data, m2->authenticator().data_length);
+
+    if (network_key != std::string("test1234")) {
+        LOG(ERROR) << "Network key validation failed";
+        return false;
+    }
```
morantr: This whole commit is basically wrong. On afb6e383789d21e0d63e6cb6b60e40a1deb1f9c1 you changed the yaml so the string will need to be dynamically allocated, and you copy it as if it was constant array size. This will end up with a segmentation fault, so you cannot push it (according to the commit guidelines: https://github.com/prplfoundation/prplMesh/blob/master/CONTRIBUTING.md#commits)

But even if weren't change the yaml file to use dynamic list, it was still redundant to convert the char* to string and compare the strings at the end, since it will always be identical strings.

EDIT: ok I understand now that you change the way you set the string on the next commit so it will not break after all. But still it has redundant parts, and I don't see the point of doing it if in the next commit to change it.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -273,23 +276,24 @@ bool parse_encrypted_settings(std::shared_ptr<tlvWscM2> m2, uint8_t *keywrapkey,
...
+               << "     network_key length: " << int(config_data.network_key_length()) << std::endl;
 
     LOG(DEBUG) << "authenticator type:" << m2->authenticator().attribute_type;
     LOG(DEBUG) << "authenticator length:" << m2->authenticator().data_length;
     LOG(DEBUG) << "authenticator buffer: " << std::endl
                << dump_buffer((uint8_t *)m2->authenticator().data, m2->authenticator().data_length);
 
-    if (network_key != std::string("test1234")) {
-        LOG(ERROR) << "Network key validation failed";
+    if (config_data.network_key_str().compare("test1234")) {
```
morantr: This test is basically testing the tlv_factory simple list and is already tested on a simple list test, which makes this check unnecessary.

morantr (changes requested):

morantr:
Regarding commit 2bcfebe2b9884d00a50daa009889fe61659687f1, I think it might cause a problem since now the getter of the string data that return char* will return an un-null-terminated string and this could cause some issues. I think you should modify this function as well to call to the function that returns the string object that will be successfully created, and return its "c_str()" return value.

framework/tlvf/test/tlvf_test.cpp
tomereli: Actually this commit is based on @arnout 's suggestion to include one commit which reproduces the issue, then add a commit which fixes it. It came with the context of adding tests which fails and only then implement the fix, it should help for better review. Since @arnout is OOO this week, we can't wait for him to provide feedback, so your call - if you still think its better to remove this commit, I will.

framework/tlvf/test/tlvf_test.cpp
tomereli: ok, will be removed.

tomereli:
> Regarding commit [2bcfebe](https://github.com/prplfoundation/prplMesh/commit/2bcfebe2b9884d00a50daa009889fe61659687f1), I think it might cause a problem since now the getter of the string data that return char* will return an un-null-terminated string and this could cause some issues. I think you should modify this function as well to call to the function that returns the string object that will be successfully created, and return its "c_str()" return value.

It is the wrong usage by the caller, the function returns `char *` so the caller should not use it for string operations. This is why we have another API which returns `std::string` - see `tlvWscM2.h: std::string manufacturer_str()` for example.

framework/tlvf/test/tlvf_test.cpp
tomereli: Edit - I removed it

morantr (approved):

morantr:
> > Regarding commit [2bcfebe](https://github.com/prplfoundation/prplMesh/commit/2bcfebe2b9884d00a50daa009889fe61659687f1), I think it might cause a problem since now the getter of the string data that return char* will return an un-null-terminated string and this could cause some issues. I think you should modify this function as well to call to the function that returns the string object that will be successfully created, and return its "c_str()" return value.
> 
> It is the wrong usage by the caller, the function returns `char *` so the caller should not use it for string operations. This is why we have another API which returns `std::string` - see `tlvWscM2.h: std::string manufacturer_str()` for example.

I approved it but I still think that leaving it as it is dangerous.

tomereli:
> > > Regarding commit [2bcfebe](https://github.com/prplfoundation/prplMesh/commit/2bcfebe2b9884d00a50daa009889fe61659687f1), I think it might cause a problem since now the getter of the string data that return char* will return an un-null-terminated string and this could cause some issues. I think you should modify this function as well to call to the function that returns the string object that will be successfully created, and return its "c_str()" return value.
> > 
> > 
> > It is the wrong usage by the caller, the function returns `char *` so the caller should not use it for string operations. This is why we have another API which returns `std::string` - see `tlvWscM2.h: std::string manufacturer_str()` for example.
> 
> I approved it but I still think that leaving it as it is dangerous.

It is dangerous but nothing we can do about it - it is the only way to allow users to modify list members, in this case the type is `char` - if someone tries to use this in a print function it is a BUG, this is why we have the `_str()` version.

morantr:
> > > > Regarding commit [2bcfebe](https://github.com/prplfoundation/prplMesh/commit/2bcfebe2b9884d00a50daa009889fe61659687f1), I think it might cause a problem since now the getter of the string data that return char* will return an un-null-terminated string and this could cause some issues. I think you should modify this function as well to call to the function that returns the string object that will be successfully created, and return its "c_str()" return value.
> > > 
> > > 
> > > It is the wrong usage by the caller, the function returns `char *` so the caller should not use it for string operations. This is why we have another API which returns `std::string` - see `tlvWscM2.h: std::string manufacturer_str()` for example.
> > 
> > 
> > I approved it but I still think that leaving it as it is dangerous.
> 
> It is dangerous but nothing we can do about it - it is the only way to allow users to modify list members, in this case the type is `char` - if someone tries to use this in a print function it is a BUG, this is why we have the `_str()` version.

You used it in print function in this PR. I have suggested a solution.

tomereli:
> You used it in print function in this PR. I have suggested a solution.

Where?

morantr:
>I think you should modify this function as well to call to the function that returns the string object that will be successfully created, and return its "c_str()" return value.

morantr:
> > You used it in print function in this PR. I have suggested a solution.
> 
> Where?

If you asked about my solution then:
>I think you should modify this function as well to call to the function that returns the string object that will be successfully created, and return its "c_str()" return value.

If you asked about where is the use in this PR then:
d8df0f9208bcaaa7c50c3b667c9325d87f000981 framework/tlvf/test/tlvf_test.cpp line 228

Issue #318, opened 2019-09-25
[POC] Automatically parsing CMDUs into a collection of TLVs (blindly parsing attributes)


Galvien:
## Abstract

Currently, we have no way to "blindly" parse TLVs (or other structures) from the buffer into objects. I'm suggesting a proof of concept in hope to address this issue: a function that generates switch clauses for parsing structures by their ID, given the files describing each ID's relevant structure in our codebase.
The function is planned to be generic, not only for parsing CMDUs into TLVs but also for later tasks such as parsing WSC attributes, but during my description I could refer specifically to the process of parsing CMDUs into TLVs, for easier understanding.

## Description
The function will be added to TLVf, and will take as an input a few parameters, but mainly:

1.  A collection of tuples of  configuration YAML file names, and the converter functions for each yaml file: 
    - Each tuple in the collection consists of:
      + A YAML file describes an enum of the relevant structures, i.e. an enum of multiAP TLVs.
      + A converter function, specifying TLVf the method of converting the enum values into class names, e.g. [eTlvType.yaml] `TLV_AL_MAC_ADDRESS_TYPE: 0x01` &rightarrow; [Class]`ieee1905_1::tlvAlMacAddressType`
this could be as simple as removing underscores and applying camel case or as complicated as going through each class' YAML file to determine the class-enum value relation.
2. The parameter to run the switch clause on, i.e. the enum index of the class. e.g the result of `getNextTlvType()`

Finally , the function will generate and return a list of lines, representing c++ code of switch clause with the relevant `addClass<T>()` (or whatever relevant function description, this is also a parameter) calls for each value in the enums.

This should solve our current issues with parsing unknown structures, with minimal manual code writing. All the function needs to "know" is the general "family" of the input structures, e.g whether it's a CMDU containing TLVs, in order to determine which YAML file to reference.

arnout:
I don't think it makes sense to have a configurable converter function. We only have two TLV enums at the moment: one for ieee1905 + multiAP (which is in fact spread over two enums, but I think that unnecessarily complicates things), and one for WSC attributes. But the WSC attributes aren't even specified as TLV types yet, so for they anyway need to be changed, so if they have a different converter, then we can/should change the class or enum names.

arnout:
[...]

Issue #160, opened 2019-08-14
More flexible to parse TLVs in autoconfiguration WSC message.


steventing:
Current implementation, the TLVs parsing order when receiving autoconfiguration WSC message is 
- AP Radio Basic Capabilities TLV followed by one WSC M1 **in Controller**
- AP Radio Identifier TLV followed by one or more WSC M2 TLV **in Agent** 

But in the Multi-AP specification v1.0, there seems to be no TLVs order constraint like this. 
For example, if the MAP Controller sends the autoconfiguration WSC message contains 2 WSC M2 TLVs followed by one AP Radio Identifier TLV to the beerocks agent, this cmdu failed be parsed.

**Describe the solution you'd like**
Check whether the mandatory TLVs exist in the certain message format instead of assuming the TLVs order.

**Describe alternatives you've considered**
No

**Additional context**
Just suggest and many thanks

arnout:
This would probably be part of a "tlvf v2" refactoring. Currently, TLVs are supposed to be processed in the order that they appear in the CMDU. "tlvf v2" should allow to search for TLVs of a specific type in the TLV list.

tomereli:
TLVF V2 will certainly aim to solve this so that users will just do a single `parse` which in turn will parse all TLVs it finds. Then the user would call `getClass()` to actually set / read the contents of each TLV.

With the current implementation, user code can just be modified to use `getNextTlvType()` and call the correct `addClass()` each time, till it reaches EOM - so there is no real gap here, just "ugly" code to support all combinations :)

tomereli:
![](https://github.trello.services/images/mini-trello-icon.png) [tlvList refactoring](https://trello.com/c/JHrwp6IN/19-tlvlist-refactoring)

PR #451, opened 2019-10-27
TLVF/POC: automatic parsing of TLV-like structures

Galvien:
This pull request suggests a solution approach that addresses #160 and #318 as per the design document that will follow these notes I thought were worth mentioning:
* currently the generated class, `TlvParser` allows the automatic parsing of a single TLV for each call of the function.
* the build should fail, as the changes to `tlvf.py` are written in python 3, and although as of #426 it **_is_** supported in the docker builder image, I did not get to implementing it. (this is still a draft PR)
* this was pretty much a proof of concept feature, spanning over several weeks (and a holiday period), therefore it involves a lot of overwriting and a hefty amount of commits.
* the next steps are to:
    - create a second function, along the lines of `generateFullParseFunction()`, which will iterate over all tlvs in a cmdu and add them to a multimap, using the id as the key.
    - update the docker builder image to build using python 3

### Issue

Based on enum files such as `eTlvType`, `eTlvTypeMap` and `eWscAttributes` which describe structures sent on the network, we would like to have the ability to automatically generate a parsing mechanism with a switch statement based on the type of structure we received.
This is my suggested solution.

### General Idea

Firstly, a tlvf method, `generateParseSwitch()`  generates a switch satement based on a few parameters, mainly:

* The list of enum values and keys, i.e. `TLV_AL_MAC_ADDRESS_TYPE: 0x01`
* A conversion function that fits the enum key to the appropriate class/struct name i.e.`TLV_AL_MAC_ADDRESS_TYPE` &longleftrightarrow; `tlvAlMacAddressType`

There's a whole bunch of metadata being inputted as well.

Then, a second method, `generateParseFunction()` wraps the switch statement in a function that iterates over all of the available structures and parses them, probably storing them all in a data collection of sort.

Lastly, a method generates a class (technically a static one) that will contain the parse function, which can be called by the parsing structure, i.e. `cmdu_rx`.

To access the parsed classes, a function similar to `addClass<T>` can be added to `cmdu_rx` that finds the first TLV of the requested type and casts it to the correct class.

### Additional notes

* It's possible to have a single class with many parsing mechanisms as well, the main thing to note is that `generateParseSwitch()` and `generateParseFunction()` are completely type, namespace and class structure agnostic, and in order to implement them on non-tlv structure types the only things needed to be written are conversion functions and some syntax modifications over the current `tlvTypeDefaultConverter()` (converts `UPPER_SNAKE_CASE` to `lowerCamelCase`) and `tlvDefaultAddClass` (returns a formatted string of the form `addClass<`***yourClassHere***>`()`)

* The last method mentioned, currently being `generateTlvParser()` for the TLV parsing file, needs to be written specifically for each structure enum being parsed, in order to include the relevant headers in the output file and to tailor each of the input parameters for `generateParseSwitch()`.

### DIFF
--- a/framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/tlvParser.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/tlvParser.h
@@ -16,9 +16,6 @@
 #include <cstddef>
 #include <stdint.h>
 #include <tlvf/swap.h>
-#include <string.h>
-#include <memory>
-#include <tlvf/BaseClass.h>
 #include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvMacAddress.h>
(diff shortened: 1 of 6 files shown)


arnout:
Replaced by a new PR that will come soon.

Issue #435, opened 2019-10-15
[Refactor] Agent: single executable per radio


arnout:
## Description

The agent architecture is currently a bit weird. It has an agent process with a main thread that is mostly idle and a backhaul manager thread and a platform manager thread. Then for each radio, there is a separately forked slave, again with a main thread that is mostly idle, and a slave thread and an ap manager thread. Each slave again forks a separate monitor process.

This is too complicated.

In a first step, we want to merge the monitor into the slave, still as a separate thread.

In a second step, we want to connect the slaves directly to the bus. Currently, the backhaul manager is responsible for forwarding some 1905/map messages to the slaves, but they could just as well receive them directly.

In a third step, we want to reduce the number of threads. Since all these threads are listening to sockets, we can have just a single select that monitors all sockets. Only the platform manager may need a separate thread since it has to interact with the platform bus mainloop.

morantr:
I think the first step should moving out the ap_manager thread out to monitor the process, and call it radio manager, which will have to classes inside it: ap_manager and monitor (I really consider to change its name to "measurment_manager").

second, make the son_slave be derived from btl instead of the backhaul_manager, and move out all parts of backhaul_manager that are not related to physical link establishment to the son_slave.
as part of it, change the son_slave name to agent, and run only one instance of it.

third, make the master thread to be derived from btl instead of mrouter, and remove mrouter.
bonus: change thread name to controller :)

I think that the backhaul manager should have a thread of its own since we don't want the agent to be blocked while waiting for some radio operation (like scanning).

tomereli:
![](https://github.trello.services/images/mini-trello-icon.png) [mrouter refactoring](https://trello.com/c/ubEitQl9/3-mrouter-refactoring)

tomereli:
![](https://github.trello.services/images/mini-trello-icon.png) [mrouter refactoring](https://trello.com/c/6CLMwoNb/6-mrouter-refactoring)

PR #452, opened 2019-10-28
Feature/ucc listener run internally on controller and agent

morantr:
Replace cli_ucc_listener and wfa_ca module on the controller with controller and agent ucc listener threads.


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3568,10 +3568,12 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
     case STATE_BACKHAUL_MANAGER_CONNECTED: {
         LOG(TRACE) << "MASTER_CONNECTED";
 
+        master_socket = backhaul_manager_socket;
+        master_socket->setPeerMac(backhaul_params.controller_bridge_mac);
```
tomereli, line 3597: Are you sure this has no effect? When `wlan_settings.band_enabled` is 0, `master_socket` was never set, and now you always set it.

tomereli (changes requested):
Can you explain a bit more in the commit message?

tomereli (commented):
The first commit message is very informative, but it looks more like a brainstorming between you and yourself than a commit message. I would rephrase it to be something like:

> The ucc listener in the agent should ideally reside in the backhaul manager, since there should be only one. Unfortunately, doing so causes inconsistency on send_cmdu function which on the controller uses a socket, while in the backhaul manager is using the bus.
The son slave which runs the ucc listener is defined in the agent configuration file for the sake of simplicity.

However, I have several issues with this approach:

1. Having 2 different ports for UCC listeners for controller and agent is confusing. Why should we run both? We are either doing controller certification OR agent certification, never in the same time, right? So I would expect a single UCC listener which supports all types of commands. In this case I would still prefer the listener to be an external application which has access to the bus and can send CMDUs to the agent and controller - this way we can have a single termination point for all UCC commands.

2. Regarding "First I thought to implement it in the backhaul manager, but it caused
inconsistency on send cmdu function which on the controller, uses
a socket, and on the backhaul manager uses the bus, so I have decided to
implement it on son_slave." - I think the backhaul manager is the only logical place to run this thread, if we decide to go with this approach. Also you said yourself that we need to remove the mrouter (https://github.com/prplfoundation/prplMesh/issues/435#issuecomment-542319551) so why not do that and align `send_cmdu`?

tomereli (commented):
The second commit is impossible to review, it would have been better to do it gradually so we can see which parts are copy paste from the wfa_ca / ucc listener. Is there any way you can make it easier to review? Perhaps add comments where the new implementation was modified?

tomereli (changes requested):
Fourth commit - I assume that a followup commit fixes test_flows.sh? Note that it has to be part of this PR, otherwise it won't pass CI..

tomereli (changes requested):
8th commit - I think that from this point this deserves a separate PR, after fixing the test_flows.sh.

tomereli (commented):
"Modify backhaul manager constructor" commit - Based on the commit message, this should be a fixup commit to the one that added `agent_ucc_listener`

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: The opposite, if the band enable is 0 the master socket was set immediately.
Basically, the cases in which the socket is not set are on all the cases we do slave_reset(), which we unset it there anyway. So if we didn't go to slave reset, we need to set it.

morantr:
1)
> The first commit message is very informative, but it looks more like a brainstorming between you and yourself than a commit message. I would rephrase it to be something like:
> 
> > The ucc listener in the agent should ideally reside in the backhaul manager, since there should be only one. Unfortunately, doing so causes inconsistency on send_cmdu function which on the controller uses a socket, while in the backhaul manager is using the bus.
> > The son slave which runs the ucc listener is defined in the agent configuration file for the sake of simplicity.

Ok, acceptable, I will change it when I'll rebase.

2)
> However, I have several issues with this approach:
> 
> 1. Having 2 different ports for UCC listeners for controller and agent is confusing. Why should we run both? We are either doing controller certification OR agent certification, never in the same time, right? So I would expect a single UCC listener which supports all types of commands. In this case I would still prefer the listener to be an external application which has access to the bus and can send CMDUs to the agent and controller - this way we can have a single termination point for all UCC commands.

You can use whatever port you wish, it is configurable on the controller/agent config file.
You can even shut the ucc_listener by setting the port number to zero.
But you can't use them both on the same port, because it is impossible to bind a specific port twice.
Anyway, If I remember correctly, the ucc uses different ports for Agent and Controller. Also, even if we would have use external ucc_listener we still would have to use 2 instances which listen to 2 different port because it is impossible to know which command is destined for for each entity.

3) 
> 2. Regarding "First I thought to implement it in the backhaul manager, but it caused
>    inconsistency on send cmdu function which on the controller, uses
>    a socket, and on the backhaul manager uses the bus, so I have decided to
>    implement it on son_slave." - I think the backhaul manager is the only logical place to run this thread, if we decide to go with this approach. Also you said yourself that we need to remove the mrouter ([#435 (comment)](https://github.com/prplfoundation/prplMesh/issues/435#issuecomment-542319551)) so why not do that and align `send_cmdu`?

You are right, but for that to happen, first need to do some refactoring - deleting mrouter, and turn the master thread to bus socket thread.

4)
> The second commit is impossible to review, it would have been better to do it gradually so we can see which parts are copy paste from the wfa_ca / ucc listener. Is there any way you can make it easier to review? Perhaps add comments where the new implementation was modified?

Basically I didn't change much. Most of the code is copy-past. The only changes I did is to remove all bml related code from ucc_listener parts , and on wfa_ca part, create use the newly defined virtual functions on the command handler instead of just use them directly on the file. 

5)
> Fourth commit - I assume that a followup commit fixes test_flows.sh? Note that it has to be part of this PR, otherwise it won't pass CI..

I know. I am waiting for @rmelotte to finish the python script which will simulate the UCC, and then I'll push a fixup for that commit that aligns test_flows.sh with the new script.

6)
> 8th commit - I think that from this point this deserves a separate PR, after fixing the test_flows.sh.

These commit stand on its own and can be part of this PR since they are related. Besides, I already did it here, and I don't think it will give us any benefit to move it now.

7)
> "Modify backhaul manager constructor" commit - Based on the commit message, this should be a fixup commit to the one that added `agent_ucc_listener`

Why? How is it related to it? This commit is something I "fixed" which is not related to the PR (except the reason I did it).

rmelotte:
@morantr I didn't want to add commits on your branch directly in case you need to rewrite history, so I pushed to a separate branch instead: `dev/test_flows_capi_commands`.
At the moment, only one test is failing: test higher layer data payload.
It fails with:
```
status,INVALID,errorCode,param name 'tlv_value' has value with invalid format
```
I will look into this tomorrow, but as you can see there seem to be an issue with one of the validations of the `tlv_value` field.

rmelotte:
I added [a commit](https://github.com/prplfoundation/prplMesh/commit/a29d3effefa5e189e88a21c0413622a12db6ab55) on `dev/test_flows_capi_commands` to fix the test that was failing. 
The issue was in test_flows.sh, not the validation.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -0,0 +1,1027 @@
...
+ */
+
+/*
+ * In order to to send WFA-CA command, send message with netcat a specific on port which is set on
+ * beerocks_agent.conf or beerocks_controller.conf under the field "ucc_listener_port".
+ * Example:
+ *     echo -n <command> | nc <bridge_ip_addr> <port>
+ */
+
+#include "../include/beerocks/bcl/beerocks_ucc_listener.h"
```
arnout, line 16: AFAICS the include directory is passed with a `-I` option, so `#include <beerocks/bcl/beerocks_ucc_listener.h>` should work.

(nitpick)

tomereli (commented):
I really like commit 162f9e2 and especially the elaborate commit message. Kudos.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/**
+ * @brief Clears configuration on Controller database.
+ * 
+ * @return None.
+ */
+void agent_ucc_listener::clear_configuration()
+{
+    // TODO
```
tomereli: Why leave it as TODO? And if so, please refer to an open task.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -0,0 +1,133 @@
...
+ * @brief Return socket to Agent with bridge 'dest_alid` MAC address.
+ * 
+ * @param[in] dest_alid Agent bridge MAC address.
+ * @return Socket* Socket to the Agent.
+ */
+Socket *agent_ucc_listener::get_dev_send_1905_destination_socket(const std::string &dest_alid)
+{
+    // On the agent side, the dest_alid is not really needed since the destination socket will
+    // always be the controller socket.
+    return *m_controller_sd;
```
tomereli, line 76: Can't `m_controller_sd` be nullptr? this may result with segfault, won't it?
Also, didn't you mean to do `return m_controller_sd` as you return a `Socket *`?

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+ * @brief Get preprepared buffer with CMDU in it.
+ * 
+ * @return std::shared_ptr<uint8_t> Buffer pointer.
+ */
+std::shared_ptr<uint8_t> agent_ucc_listener::get_buffer_filled_with_cmdu()
+{
+    // Currently, no such buffer on agent side.
+    return std::shared_ptr<uint8_t>();
```
tomereli: so why not return `nullptr`?

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -0,0 +1,133 @@
...
+/**
+ * @brief Send CMDU to destined Agent.
+ * 
+ * @param[in] sd Agent socket
+ * @param[in] cmdu_tx CMDU object
+ * @return true if successful, false if not.
+ */
+bool agent_ucc_listener::send_cmdu_to_destination(Socket *sd, ieee1905_1::CmduMessageTx &cmdu_tx)
+{
+    if (*m_controller_sd == nullptr) {
```
tomereli, line 99: What type is `m_controller_sd`? is it `shared_ptr`? why do you need to dereference it?

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -0,0 +1,133 @@
...
+        err_string = "parameter 'bss_info' is not relevant to the agent";
+        return false;
+    }
+
+    if (params.find("backhaul") == params.end()) {
+        err_string = "parameter 'backhaul' is missing";
+        return false;
+    }
+
+    // TODO
```
tomereli: Please refer to an open task in the TODO

agent/src/beerocks/slave/agent_ucc_listener.h
```diff
@@ -0,0 +1,49 @@
...
+    std::shared_ptr<uint8_t> get_buffer_filled_with_cmdu() override;
+    bool send_cmdu_to_destination(Socket *sd, ieee1905_1::CmduMessageTx &cmdu_tx) override;
+    bool handle_dev_set_config(std::unordered_map<std::string, std::string> &params,
+                               std::string &err_string) override;
+
+    const std::string &m_vendor;
+    const std::string &m_model;
+    const std::string &m_bridge_iface;
+    std::string m_bridge_mac;
+    Socket **const m_controller_sd;
```
tomereli, line 42: Oh, now I see it.... Why is it `**`?? Why not a simple pointer or shared_ptr?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -119,6 +119,17 @@ bool slave_thread::init()
         return false;
     }
 
+    if (config.ucc_listener_slave_hostap_iface == config.hostap_iface &&
```
tomereli, line 122: Please also add a flag which disables UCC completely in both agent and controller since it has effect on the code.

agent/src/beerocks/slave/agent_ucc_listener.h
morantr: It is pointer to pointer since the son_slave is manging the controller socket pointer and it can modify it. That way it is promised that if the son_slave will change it, the ucc_listener would "know" about it.

agent/src/beerocks/slave/agent_ucc_listener.cpp
morantr: see https://github.com/prplfoundation/prplMesh/pull/452#discussion_r342985412

agent/src/beerocks/slave/agent_ucc_listener.cpp
morantr: It is the same thing.

agent/src/beerocks/slave/agent_ucc_listener.cpp
morantr: see https://github.com/prplfoundation/prplMesh/pull/452#discussion_r342985412

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: There is already. It is the port number. If the port is zero, it means disable. (look one line down).

agent/src/beerocks/slave/agent_ucc_listener.cpp
morantr: done

agent/src/beerocks/slave/agent_ucc_listener.cpp
morantr: done

agent/src/beerocks/slave/agent_ucc_listener.cpp
tomereli: `nullptr` would be more appropriate IMHO. This just raises questions for anyone who will bump into this code. Also, should add a TODO to the comment with reference to an open task if this is bound to be changed.

agent/src/beerocks/slave/agent_ucc_listener.h
tomereli: This should be a shared_ptr then, and the son_slave should notify the listener on changes to the socket instead of relying on the pointer to be changed "under your feet".

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: So this should be mentioned in a comment in the configuration file if it isn't already.

tomereli (changes requested):
Mostly nitpicks, apart from the socket handling by the UCC listener. It should be notified by the slave on any changes to the socket. After this we can merge it.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: done

agent/src/beerocks/slave/agent_ucc_listener.cpp
morantr: By definition shared_ptr<something>() is pointing to nullptr. 😒 
Whatever I'll change it.
There is no need to add a TODO, since we might not need certification buffer.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -83,7 +84,7 @@ Socket *agent_ucc_listener::get_dev_send_1905_destination_socket(const std::stri
 std::shared_ptr<uint8_t> agent_ucc_listener::get_buffer_filled_with_cmdu()
 {
     // Currently, no such buffer on agent side.
-    return std::shared_ptr<uint8_t>();
+    return std::shared_ptr<uint8_t>(nullptr);
```
tomereli, line 87: I meant just `return nullptr` .... anyway (nitpick)

tomereli (approved):
I'm approving already so not to block this. But if you can fix the remaining comments it would be great.

agent/src/beerocks/slave/agent_ucc_listener.h
morantr: I don't agree with you. IMO this is the most clean/transparent way to do it. The design of the ucc_listener is to have a shared memory with another thread. On the master thread is easy since it has the controller database which holds the relevant sockets is on another file. Here it is more problematic since the socket is on the son_slave and I can't create a circular #include.
As well this is not the purpose of shared_ptr.
shared_ptr purourse is ti hold and not free a memory block as long as it is used. This is not the case here. Even if I turned the pointer to shared_ptr I would still have to pass it as a pointer or reference to shared_ptr to the ucc listener.
As far as I can see these are the alternative solutions:
1. Instead of passing the pointer to the controller socket, passing a callback function (lambda) which will give the current controller socket. The only drawback in this solution is that every time the ucc listener will need the controller socket, it will have to call to the callback function.
2. Create a public function on the ucc_listener that updates the controller socket.
This is a worse idea since every place the son_slave will want to change the controller socket, it will have to call this function. And of course, if a developer will want to change the controller socket in the future, it will have to remember to call this function. This can easily can be forgotten and can lead eventually to bugs.

tomereli: > I don't agree with you. IMO this is the most clean/transparent way to do it. The design of the ucc_listener is to have a shared memory with another thread. On the master thread is easy since it has the controller database which holds the relevant sockets is on another file. Here it is more problematic since the socket is on the son_slave and I can't create a circular #include.

Can't you just use `extern`? no need for circular `#include`.

morantr: I'll tell you what, When I will create the unified agent process, I'll create for it a database class like the controller has. That way the agent class will be much cleaner.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -29,9 +30,9 @@ namespace backhaul_manager {
 class main_thread : public btl::transport_socket_thread {
 
 public:
-    main_thread(std::string temp_path_, std::set<std::string> slave_ap_ifaces_,
-                std::set<std::string> slave_sta_ifaces_, int stop_on_failure_attempts,
-                std::string const_bh_slave_);
+    main_thread(const config_file::sConfigSlave &config,
+                const std::set<std::string> &slave_ap_ifaces_,
+                const std::set<std::string> &slave_sta_ifaces_);
```
arnout, line 35: Since you're storing references to the arguments in the main_thread object, I think there should be a (doxygen) comment that clarifies that the lifetime of those arguments must surpass that of the main_thread object itself.

arnout (approved):
This is an excellent pull request, with nice commit messages and of course good content as well. I really like how the UCC listener has been implemented.

I've got two nitpick comments so it can be merged as is.

I'm not actually so sure about the "controller: Change database cetification buffer return value type" commit. It doesn't explain why we need to do that. But then, I don't really understand when smart pointers should be passed by reference and when by value.

PR #454, opened 2019-10-29
Feature/test for optimal path task

### LINKED ISSUE #443: [TASK] Test for optimal path task
## Description

In the demo at prpl Summit, we made sure that the controller's optimal path task is working properly. To make sure this keeps on working, we should have a test of it. The test should run in Docker, similar to test_flows.sh, but in a separate script.

## Exit criteria

A test script exists that runs a gateway and two repeaters and tests the following features:
- Optimal path task is started when a client connects.
- Check that 11k request is done and that its arguments are as expected.
- Test gives 11k response that shows best link quality for current BSS.
- Check that no steering is done.
- Check that after a while optimal path task runs again *TBD what should trigger this?*
- Give 11k response that shows best link quality in a different BSS.
- Check that steering is done.
- Connect a second station, check that optimal path task runs.
- Give error response to 11k request.
- Check that error response is handled correctly (no crashes, falls back to RSSI).

YuriyMasechko:
New test for optimal path task was added.
Test has three scenarios:
- test gives 11k response that shows best link quality for current BSS, no steering done
- test gives 11k response that shows best link quality in a different BSS, steer client
- error response to 11k request, optimal path falls back to RSSI measurement

Additionally dhcp events were added as association task
requires sta to be connected with IP. New data path to change sta stats was added as event.

fixes #443

### DIFF
--- a/common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
@@ -120,9 +120,14 @@ static void map_event_obj_parser(std::string event_str, parsed_obj_map_t &map_ob
     std::stringstream ss(event_str.c_str() + idx_start);
     std::string str_storage;
     bool opcode = true;
+    bool type   = true;
     while (std::getline(ss, str_storage, ' ')) {
-        if (opcode) {
-            // assume that the first param is event name
+        if (type) {
+            // assume that the first param is type - DATA or EVENT
+            map_obj[DUMMY_EVENT_KEYLESS_PARAM_TYPE] = str_storage;
+            type                                    = false;
+        } else if (opcode) {
+            // assume that the second param is data or event name
             map_obj[DUMMY_EVENT_KEYLESS_PARAM_OPCODE] = str_storage;
             opcode                                    = false;
         } else if (beerocks::net::network_utils::is_valid_mac(str_storage)) {
@@ -343,9 +348,22 @@ bool base_wlan_hal_dummy::process_ext_events()
     //base_wlan_hal_dummy::parsed_obj_debug(event_obj);
 
     // Process the event
-    if (!process_dummy_event(event_obj)) {
-        LOG(ERROR) << "Failed processing DUMMY event: "
-                   << event_obj[DUMMY_EVENT_KEYLESS_PARAM_OPCODE];
+    if (event_obj[DUMMY_EVENT_KEYLESS_PARAM_TYPE] == "EVENT") {
+        if (!process_dummy_event(event_obj)) {
+            LOG(ERROR) << "Failed processing DUMMY event: "
+                       << event_obj[DUMMY_EVENT_KEYLESS_PARAM_OPCODE];
+            return false;
+        }
+    }
+    // Process data
+    else if (event_obj[DUMMY_EVENT_KEYLESS_PARAM_TYPE] == "DATA") {
+        if (!process_dummy_data(event_obj)) {
+            LOG(ERROR) << "Failed processing DUMMY data: "
+                       << event_obj[DUMMY_EVENT_KEYLESS_PARAM_OPCODE];
+            return false;
+        }
+    } else {
+        LOG(ERROR) << "Unsupported type " << event_obj[DUMMY_EVENT_KEYLESS_PARAM_TYPE];
         return false;
     }
 
--- a/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
@@ -40,6 +40,14 @@ namespace dummy {
 /////////////////////////// Local Module Functions ///////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
+static mon_wlan_hal::Event dummy_to_bwl_event(const std::string &opcode)
+{
+    if (opcode == "RRM-BEACON-REP-RECEIVED")
+        return mon_wlan_hal::Event::RRM_Beacon_Response;
+
+    return mon_wlan_hal::Event::Invalid;
+}
+
 //////////////////////////////////////////////////////////////////////////////
 /////////////////////////////// Implementation ///////////////////////////////
 //////////////////////////////////////////////////////////////////////////////
@@ -142,6 +150,8 @@ bool mon_wlan_hal_dummy::process_dummy_data(parsed_obj_map_t &parsed_obj) { retu
 
 bool mon_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
 {
+    char *tmp_str;
+    int64_t tmp_int;
     // Filter out empty events
     std::string opcode;
     if (!(parsed_obj.find(DUMMY_EVENT_KEYLESS_PARAM_OPCODE) != parsed_obj.end() &&
@@ -149,6 +159,95 @@ bool mon_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
         return true;
     }
     LOG(TRACE) << __func__ << " - opcode: |" << opcode << "|";
+
+    auto event = dummy_to_bwl_event(opcode);
+
+    switch (event) {
+    case Event::RRM_Beacon_Response: {
+        // Allocate response object
+        auto resp_buff = ALLOC_SMART_BUFFER(sizeof(SBeaconResponse11k));
+        auto resp      = reinterpret_cast<SBeaconResponse11k *>(resp_buff.get());
+
+        if (!resp) {
+            LOG(FATAL) << "Memory allocation failed!";
+            return false;
+        }
+
+        // Initialize the message
+        memset(resp_buff.get(), 0, sizeof(SBeaconResponse11k));
+
+        // STA MAC
+        if (!dummy_obj_read_str(DUMMY_EVENT_KEYLESS_PARAM_MAC, parsed_obj, &tmp_str)) {
+            LOG(ERROR) << "Failed reading mac parameter!";
+            return false;
+        }
+
+        beerocks::net::network_utils::mac_from_string(resp->sta_mac.oct, tmp_str);
+
+        // Channel
+        if (!dummy_obj_read_int("channel", parsed_obj, tmp_int)) {
+            LOG(ERROR) << "Failed reading channel parameter!";
+            return false;
+        }
+        resp->channel = tmp_int;
+
+        // Dialog Token
+        if (!dummy_obj_read_int("dialog_token", parsed_obj, tmp_int)) {
+            LOG(ERROR) << "Failed reading dialog_token parameter!";
+            return false;
+        }
+        resp->dialog_token = tmp_int;
+
+        // Measurement Response Mode
+        if (!dummy_obj_read_int("measurement_rep_mode", parsed_obj, tmp_int)) {
+            LOG(ERROR) << "Failed reading measurement_rep_mode parameter!";
+            return false;
+        }
+        resp->rep_mode = tmp_int;
+
+        // Operating Class
+        if (!dummy_obj_read_int("op_class", parsed_obj, tmp_int)) {
+            LOG(ERROR) << "Failed reading op_class parameter!";
+            return false;
+        }
+        resp->op_class = tmp_int;
+
+        // Measurement Duration
+        if (!dummy_obj_read_int("duration", parsed_obj, tmp_int)) {
+            LOG(ERROR) << "Failed reading duration parameter!";
+            return false;
+        }
+        resp->duration = tmp_int;
+
+        // RCPI
+        if (!dummy_obj_read_int("rcpi", parsed_obj, tmp_int)) {
+            LOG(ERROR) << "Failed reading rcpi parameter!";
+            return false;
+        }
+        resp->rcpi = tmp_int;
+
+        // RSNI
+        if (!dummy_obj_read_int("rsni", parsed_obj, tmp_int)) {
+            LOG(ERROR) << "Failed reading rsni parameter!";
+            return false;
+        }
+        resp->rsni = tmp_int;
+
+        // BSSID
+        if (!dummy_obj_read_str("bssid", parsed_obj, &tmp_str)) {
+            LOG(ERROR) << "Failed reading mac parameter!";
+            return false;
+        }
+        beerocks::net::network_utils::mac_from_string(resp->bssid.oct, tmp_str);
+
+        // Add the message to the queue
+        event_queue_push(event, resp_buff);
+        break;
+    }
+    default:
+        LOG(WARNING) << "Unhandled event received: " << opcode;
+        break;
+    }
     return true;
 }
 
--- a/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
@@ -52,6 +52,8 @@ static mon_wlan_hal_dummy::Data dummy_to_bwl_data(const std::string &opcode)
 {
     if (opcode == "STA-UPDATE-STATS")
         return mon_wlan_hal_dummy::Data::STA_Update_Stats;
+    else if (opcode == "RRM-BEACON-REP-RECEIVED")
+        return mon_wlan_hal_dummy::Data::RRM_Update_Beacon_Measurements;
 
     return mon_wlan_hal_dummy::Data::Invalid;
 }
@@ -91,7 +93,7 @@ bool mon_wlan_hal_dummy::update_stations_stats(const std::string vap_iface_name,
         return false;
     }
 
-    dummy_sta_stats = dummy_sta->second;
+    dummy_sta_stats = dummy_sta->second.sta_stats;
 
     sta_stats.rx_rssi_watt             = dummy_sta_stats.rx_rssi_watt;
     sta_stats.rx_rssi_watt_samples_cnt = dummy_sta_stats.rx_rssi_watt_samples_cnt;
@@ -119,6 +121,20 @@ bool mon_wlan_hal_dummy::sta_beacon_11k_request(const SBeaconRequest11k &req, in
     LOG(DEBUG) << "Beacon 11k request to sta " << sta_mac << " on bssid " << bssid << " channel "
                << std::to_string(req.channel);
 
+    auto dummy_sta = m_dummy_stas_map.find(sta_mac);
+    if (dummy_sta != m_dummy_stas_map.end()) {
+        auto dummy_bssid_event = dummy_sta->second.beacon_measurment_events.find(bssid);
+        if (dummy_bssid_event == dummy_sta->second.beacon_measurment_events.end()) {
+            LOG(WARNING) << "No RRM beacon response for bssid " << bssid;
+        } else {
+            //TODO: remove sleep. Sleep fixes missing requests sent by controller but not
+            //received by agent
+            sleep(1);
+            process_dummy_event(dummy_bssid_event->second);
+        }
+    } else
+        LOG(WARNING) << "No RRM beacon response for sta " << sta_mac;
+
     return true;
 }
 
@@ -138,6 +154,7 @@ bool mon_wlan_hal_dummy::process_dummy_data(parsed_obj_map_t &parsed_obj)
 {
     char *tmp_str;
     char *sta_mac;
+    char *bssid;
     int64_t tmp_int;
     // Filter out empty data
     std::string opcode;
@@ -200,7 +217,21 @@ bool mon_wlan_hal_dummy::process_dummy_data(parsed_obj_map_t &parsed_obj)
         }
         dummy_sta_stats.rx_phy_rate_100kb = (tmp_int / 100);
 
-        m_dummy_stas_map[sta_mac] = dummy_sta_stats;
+        m_dummy_stas_map[sta_mac].sta_stats = dummy_sta_stats;
+        break;
+    }
+    case Data::RRM_Update_Beacon_Measurements: {
+        if (!dummy_obj_read_str(DUMMY_EVENT_KEYLESS_PARAM_MAC, parsed_obj, &sta_mac)) {
+            LOG(ERROR) << "Failed reading mac parameter!";
+            return false;
+        }
+
+        if (!dummy_obj_read_str("bssid", parsed_obj, &bssid)) {
+            LOG(ERROR) << "Failed reading bssid!";
+            return false;
+        }
+
+        m_dummy_stas_map[sta_mac].beacon_measurment_events[bssid] = parsed_obj;
         break;
     }
     default:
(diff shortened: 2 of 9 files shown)


framework/platform/bpl/linux/bpl_dhcp_mon.c
```diff
@@ -8,8 +8,74 @@
...
+    tmpTok = strtok(line, ",");
+    while (tmpTok != NULL && numTok < 4) {
+        arrTok[numTok++] = tmpTok;
+        tmpTok           = strtok(NULL, ",");
+    }
+
+    // Execute the callback
+    if (s_pCallback)
+        s_pCallback((arrTok[0]) ? arrTok[0] : "", (arrTok[1]) ? arrTok[1] : "",
+                    (arrTok[2]) ? arrTok[2] : "", (arrTok[3]) ? arrTok[3] : "");
```
morantr: I think it is a good idea to add a description of the tokens:
<op: {"add","del", "old"}> <client_mac> <ip> <hostname>

Also, create some data validation. But IMO it should be on platform_manager_thread.cpp:send_dhcp_notification().
since it is more convenient to validate there the mac, the IP and the op string.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -120,9 +120,14 @@ static void map_event_obj_parser(std::string event_str, parsed_obj_map_t &map_ob
...
+    bool type   = true;
     while (std::getline(ss, str_storage, ' ')) {
-        if (opcode) {
-            // assume that the first param is event name
+        if (type) {
+            // assume that the first param is type - DATA or EVENT
+            map_obj[DUMMY_EVENT_KEYLESS_PARAM_TYPE] = str_storage;
+            type                                    = false;
+        } else if (opcode) {
+            // assume that the second param is name
```
morantr: please change the comment -> "....is data or event name"

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -75,7 +83,13 @@ bool mon_wlan_hal_dummy::update_vap_stats(const std::string vap_iface_name, SVap
 bool mon_wlan_hal_dummy::update_stations_stats(const std::string vap_iface_name,
```
morantr, line 85: If you can, please change the string argument to be const reference.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.h
```diff
@@ -54,6 +62,7 @@ class mon_wlan_hal_dummy : public base_wlan_hal_dummy, public mon_wlan_hal {
     // Private data-members:
 private:
     std::shared_ptr<char> m_temp_dummy_value;
+    SStaStats m_dummy_sta_stats;
```
morantr: This implementation supports only one station.
I think it will be better if you store it on unrodered_map which it's key is "sta_mac".
Optional: create a handler for a client disconnect event to remove the client element from the map.

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+    check_error=0
+
+    # No steer scenario
+    dbg "Connect dummy STA to wlan0"
+    check docker exec -it repeater1 sh -c \
+        'echo "EVENT AP-STA-CONNECTED 11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan0/EVENT'
+
+    dbg "DHCP event"
+    check docker exec -it gateway sh -c \
+        'echo "0,11:22:33:44:55:66,192.168.1.10,test" > /tmp/$USER/beerocks/dhcp/EVENT'
```
morantr: What is the zero at the string beginning? shouldn't it be one of {"add","del","old"} ?
Please look for main_thread::send_dhcp_notification on platform manager.

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+
+    dbg "Confirming 11k request is done"
+    check docker exec -it repeater1 sh -c \
+        "grep -i -q 'Beacon 11k request 11:22:33:44:55:66 aa:bb:cc:00:00:10 1' /tmp/$USER/beerocks/logs/beerocks_monitor_wlan0.log"
+
+    dbg "RPM Beacon Response"
```
morantr: I think you meant "RRM"? If not, what is "RPM" then?

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+    check docker exec -it repeater1 sh -c \
+        'echo "EVENT RRM-BEACON-REP-RECEIVED 11:22:33:44:55:66 channel=1 dialog_token=0 measurement_rep_mode=0 op_class=0 duration=50 rcpi=-40 rsni=40 bssid=aa:bb:cc:00:00:10" > /tmp/$USER/beerocks/wlan0/EVENT'
+
+    dbg "RPM Update Stats"
```
morantr: If indeed "RRM" is what you meant, so it is defintly not related to regular measurement, so please remove it from the comment.

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+    check docker exec -it repeater1 sh -c \
+        'echo "DATA STA-UPDATE-STATS rssi=-38,-39,-40,-41 snr=38,39,40,41 uplink=1000 downlink=800" > /tmp/$USER/beerocks/wlan0/EVENT'
+    sleep 15
+
+    dbg "Confirming 11k request is done"
+    check docker exec -it repeater1 sh -c \
+        "grep -i -q 'Beacon 11k request 11:22:33:44:55:66 aa:bb:cc:11:00:20 149' /tmp/$USER/beerocks/logs/beerocks_monitor_wlan0.log"
+
+    dbg "RPM Beacon Response"
```
morantr: RRM?

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -80,6 +80,49 @@ bool mon_wlan_hal_dummy::sta_channel_load_11k_request(const SStaChannelLoadReque
...
+                      std::to_string(req.duration) + " " +  // Duration
+                      measurement_mode;                     // Measurement Mode
+
+    if (req.use_optional_ssid) {
+        std::string req_ssid = '"' + std::string((char *)req.ssid) + '"';
+
+        cmd += " ssid=" + req_ssid;
+    }
+
+    LOG(DEBUG) << "Beacon 11k request " << cmd;
```
morantr: What is the purpose of the request struct processing? I mean, we don't send it to hostapd so it makes it makes the hostapd string construction to redundant.
The print only is enough: "Beacon measurement request to sta <sta_mac> on bssid <bssid_mac>" for the test_flow.sh.

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+        "grep -i -q 'Beacon 11k request 11:22:33:44:55:66 aa:bb:cc:00:00:10 1' /tmp/$USER/beerocks/logs/beerocks_monitor_wlan0.log"
+
+    dbg "RPM Beacon Response"
+    check docker exec -it repeater1 sh -c \
+        'echo "EVENT RRM-BEACON-REP-RECEIVED 11:22:33:44:55:66 channel=1 dialog_token=0 measurement_rep_mode=0 op_class=0 duration=50 rcpi=-40 rsni=40 bssid=aa:bb:cc:00:00:10" > /tmp/$USER/beerocks/wlan0/EVENT'
+
+    dbg "RPM Update Stats"
+    check docker exec -it repeater1 sh -c \
+        'echo "DATA STA-UPDATE-STATS rssi=-38,-39,-40,-41 snr=38,39,40,41 uplink=1000 downlink=800" > /tmp/$USER/beerocks/wlan0/EVENT'
+    sleep 15
```
morantr: Why sleeping 15 seconds? Isn't 1-2 seconds is enough?

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+    # No steer scenario
+    dbg "Connect dummy STA to wlan0"
+    check docker exec -it repeater1 sh -c \
+        'echo "EVENT AP-STA-CONNECTED 11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan0/EVENT'
+
+    dbg "DHCP event"
+    check docker exec -it gateway sh -c \
+        'echo "0,11:22:33:44:55:66,192.168.1.10,test" > /tmp/$USER/beerocks/dhcp/EVENT'
+
+    dbg "Confirming 11k request is done"
```
morantr: To make it more clear, I think that it will help to mention as part of which procedure/task this is done.
For example, this request is done by the client association task (right?).
And of course, apply it on the others as well.
To be honest, I find it very difficult to review it, since I see in this file so many events of requests, responses, etc..
that I'm not able to understand what exactly is going on the controller. I would appreciate comments that explain what is happening as getting forward with this test.

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+        'echo "DATA STA-UPDATE-STATS rssi=-38,-39,-40,-41 snr=38,39,40,41 uplink=1000 downlink=800" > /tmp/$USER/beerocks/wlan0/EVENT'
+    sleep 15
+
+    dbg "Confirming 11k request is done"
+    check docker exec -it repeater1 sh -c \
+        "grep -i -q 'Beacon 11k request 11:22:33:44:55:66 aa:bb:cc:11:00:20 149' /tmp/$USER/beerocks/logs/beerocks_monitor_wlan0.log"
+
+    dbg "RPM Beacon Response"
+    check docker exec -it repeater1 sh -c \
+        'echo "EVENT RRM-BEACON-REP-RECEIVED 11:22:33:44:55:66 channel=149 dialog_token=0 measurement_rep_mode=0 op_class=0 duration=50 rcpi=-80 rsni=10 bssid=aa:bb:cc:11:00:20" > /tmp/$USER/beerocks/wlan0/EVENT'
```
morantr: I don't understand how this script is synced with the controller. I mean, you can't know when exactly the request is sent, so how can't you know when to send an event on time so you will not reach timeout.
I think that the right thing to do is to store on the beacon measurement responses on the dummy for each bssid by sending it as DATA, and when receiving a measurement request, pull the data and simulate an internal "RRM-BEACON-REP-RECEIVED" event.
You can use the unordered map which I suggested on other comment to hold the sta measurement, to hold the beacon measurement response as well. 
That way you can pre-prepare the responses and make sure it will be sent on time.

```cpp
struct sDummySta {
    SStaStats sta_stats;
    std::unordered_map<std::string, std::string> beacon_measurment_events; // key=bssid, value=event string
};
std::unordered_map<std::string, sDummySta> dummy_sta; // key=sta_mac
```

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
         
 }
 
+test_optimal_path_dummy() {
```
morantr, line 278: hmm, does this test should be here? I thought we test here only certification flows.
@arnout

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+        'echo "EVENT RRM-BEACON-REP-RECEIVED 11:22:33:44:55:66 channel=1 dialog_token=0 measurement_rep_mode=0 op_class=0 duration=50 rcpi=-50 rsni=30 bssid=00:11:22:33:00:10" > /tmp/$USER/beerocks/wlan0/EVENT'
+    sleep 1
+
+    dbg "Confirming 11k request is done"
+    check docker exec -it repeater1 sh -c \
+        "grep -i -q 'Beacon 11k request 11:22:33:44:55:66 00:11:22:33:00:20 149' /tmp/$USER/beerocks/logs/beerocks_monitor_wlan0.log"
+
+    dbg "RPM Beacon Response"
+    check docker exec -it repeater1 sh -c \
+        'echo "EVENT RRM-BEACON-REP-RECEIVED 11:22:33:44:55:66 channel=149 dialog_token=0 measurement_rep_mode=0 op_class=0 duration=50 rcpi=-80 rsni=10 bssid=00:11:22:33:00:20" > /tmp/$USER/beerocks/wlan0/EVENT'
```
morantr: You already requested measurement one time for each bssid (I assume it is part of optimal path task), why did you 2 more measurement on wlan0 and 1 more measurement on wlan2?
Same thing on the second optimal task phase.

tools/docker/tests/test_flows.sh
```diff
@@ -207,6 +208,220 @@ test_client_association_dummy(){
...
+
+    #TODO// check for "disconnected after successful steering, proceeding to unblock" message 
+
+    dbg "Connect dummy STA to wlan2"
+    check docker exec -it repeater1 sh -c \
+        'echo "EVENT AP-STA-CONNECTED 11:22:33:44:55:66" > /tmp/$USER/beerocks/wlan2/EVENT'
+
+    dbg "Confirm steering success by client connected"
+    check docker exec -it gateway sh -c \
+        'grep -i -q "steering successful for sta 11:22:33:44:55:66" /tmp/$USER/beerocks/logs/beerocks_controller.log'
```
morantr: If I'm not mistaken, these tests are already done by other test flows, so isn't it redundant to do it again?
I think that it is enough to check that optimal_path has decided to steer the client, without checking the whole steering flow.
Another option is to use the functions that already test some flows (like steering) and combine them in the current test (as a function call I mean).

morantr (changes requested):
Besides my comments, please update the dummy bwl readme.md file.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
tomereli: Fully agree with Moran on this one.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.h
```diff
@@ -20,6 +20,14 @@ namespace dummy {
  */
 class mon_wlan_hal_dummy : public base_wlan_hal_dummy, public mon_wlan_hal {
 
+    // Public definitions
+public:
+    enum class Data {
+        Invalid = 0,
+
```
tomereli: redundant line breaks

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.h
tomereli: Good idea, but if it is too much work than I suggest to open a task for it. We won't need dummy implementation to test more than one dummy station for now.

framework/platform/bpl/linux/bpl_dhcp_mon.c
YuriyMasechko: Comment was changed.
Open separate task to add data validation to manager?

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
YuriyMasechko: Done.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
YuriyMasechko: It will require to change all bwl implementations, maybe separate task?

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.h
YuriyMasechko: Extended to map.

tools/docker/tests/test_flows.sh
YuriyMasechko: Will revert this test and commit new one. To big diff to review.

tools/docker/tests/test_flows.sh
YuriyMasechko: Will revert this test and commit new one. To big diff to review.

tools/docker/tests/test_flows.sh
YuriyMasechko: New DATA event was added to store beacon measurement.

tools/docker/tests/test_flows.sh
YuriyMasechko: @arnout should it go to separate file?

tools/docker/tests/test_flows.sh
YuriyMasechko: Because there are 2 IREs.

tools/docker/tests/test_flows.sh
YuriyMasechko: Will revert this test and commit new one. To big diff to review.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.h
YuriyMasechko: Done.

tools/docker/tests/test_flows.sh
YuriyMasechko: Will revert this test and commit new one. To big diff to review

tools/docker/tests/test_flows.sh
YuriyMasechko: Will revert this test and commit new one. To big diff to review

tools/docker/tests/test_flows.sh
YuriyMasechko: Will revert this test and commit new one. To big diff to review

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
YuriyMasechko: Changed to print only mac, bssid and channel.

tools/docker/tests/test_flows.sh
YuriyMasechko: Will revert this test and commit new one. To big diff to review

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
morantr: So leave it for now...

tools/docker/tests/test_flows.sh
morantr: @tomereli can you please reply

morantr (approved):

tools/docker/tests/test_flows.sh
arnout: It should be a separate file (and the client steering test as well). However, that can be done in a follow-up PR. Putting it in a separate file will require a bit of refactoring, e.g. moving some functions to functions.sh, and maybe not doing the startup from this script but instead do it in a separate step. And it requires an update of travis.yml.

So yes, it should be a separate file, but that should be done in a new PR. @YuriyMasechko if you have time you could work on that.

arnout (approved):

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -52,6 +52,8 @@ static mon_wlan_hal_dummy::Data dummy_to_bwl_data(const std::string &opcode)
 {
     if (opcode == "STA-UPDATE-STATS")
         return mon_wlan_hal_dummy::Data::STA_Update_Stats;
+    else if (opcode == "RRM-BEACON-REP-RECEIVED")
+        return mon_wlan_hal_dummy::Data::RRM_Update_Beacon_Measurements;
```
arnout, line 56: We end up using only this DATA. So I think it's not actually needed to make this split into DATA/EVENT. Instead, we can put everything in process_dummy_event().

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -80,6 +114,27 @@ bool mon_wlan_hal_dummy::sta_channel_load_11k_request(const SStaChannelLoadReque
...
+    auto dummy_sta = m_dummy_stas_map.find(sta_mac);
+    if (dummy_sta != m_dummy_stas_map.end()) {
+        auto dummy_bssid_event = dummy_sta->second.beacon_measurment_events.find(bssid);
+        if (dummy_bssid_event == dummy_sta->second.beacon_measurment_events.end()) {
+            LOG(WARNING) << "No RRM beacon response for bssid " << bssid;
+        } else {
+            //TODO: remove sleep. Sleep fixes missing requests sent by controller but not
+            //received by agent
+            sleep(1);
+            process_dummy_event(dummy_bssid_event->second);
```
arnout, line 133: This is the only place where the RRM event is used. So instead, that code can be moved here directly.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -95,15 +150,198 @@ bool mon_wlan_hal_dummy::sta_link_measurements_11k_request(const std::string &st
...
     }
     LOG(TRACE) << __func__ << " - opcode: |" << opcode << "|";
+
+    auto event = dummy_to_bwl_event(opcode);
+
+    switch (event) {
+    case Event::RRM_Beacon_Response: {
+        // Allocate response object
+        auto resp_buff = ALLOC_SMART_BUFFER(sizeof(SBeaconResponse11k));
+        auto resp      = reinterpret_cast<SBeaconResponse11k *>(resp_buff.get());
```
arnout, line 263: So this code can move directly into sta_beacon_11k_request().

arnout (changes requested):
I rebased on master again. I also removed the commit adding the dummy dhcp, since it's no longer needed (requirement for IP address has been removed).

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
YuriyMasechko: Data path is necessary for 2 reasons:
1. to prevent sleeping and un-synced operation on the test_flows.sh.
2. It is a preperation for the future, for commands that needs to return data immediatly that is not base on events, and we want the ability to change it.

Of course we could handle everything in process_event(), but it will be not such clear.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
YuriyMasechko: See comment above.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
YuriyMasechko: See comment above.

arnout (approved):
Since my remaining comment was overridden, I'll approve and merge after CI finishes.

PR #458, opened 2019-10-30
tlvf: CmduMessageRx: automatic tlv parse support 

tomereli:
CmduMessageRx provides a `parse` API which as for now parses only the
CMDU header. Users are then expected to call `addClass` for each TLV manually.
This commit adds support for parse_tlvs option to the class's parse API,
disabled by default for backward compatability:
When parse_tlvs is true, the `parse` API goes through all the TLVs in
the CMDU and calls addClass for each found TLV.
If Unknown TLVs are encountered, addClass<unknownTlv> is called.
Parsing ends successfully only once tlvEndOfMessage is found and added.

This option is intended to be used only for standard CMDUs, such as the
ones defined in the MultiAP and 1905.1a specifications, since it is
using the TLV type for the casting.

Note that for WSC TLV, we support only M1 and M2 which are required by
the MultiAP specification. The two WSC types are differenciated via the
MSG_TYPE attribute. In the future, this will be also handled
automatically by the parser by adding each attribute manually without
using the tlvWscMx.yaml files which are wrong since they assume that the
attributes have to be the same order they are defined in the yaml.

Also add a test in tlvf_test which verifies the new functionality works. No effect on other components.

### DIFF
--- a/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
@@ -125,6 +125,7 @@ public:
     uint16_t getNextTlvLength() const;
     uint8_t *getNextTlvData() const;
     void swap();
+    bool swap_needed() {return m_swap; }
     bool is_finalized() const { return m_finalized; };
     bool is_swapped() const { return m_swapped; };
     eMessageType getMessageType();
--- a/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
@@ -55,11 +55,64 @@ public:
 
     static uint16_t getCmduHeaderLength() { return kCmduHeaderLength; }
 
+    /**
+     * @brief Get the Class object at index idx in the all classes array
+     * 
+     * @param idx index in the all classes array
+     * @return std::shared_ptr<BaseClass> to the class object at index idx, nullptr if not found
+     */
     std::shared_ptr<BaseClass> getClass(size_t idx) const;
 
+    /**
+     * @brief Get the (first) Class object
+     * 
+     * @tparam T class template
+     * @return std::shared_ptr<T> to the first object found of type T, nullptr if not found
+     */
+    template <class T> std::shared_ptr<T> getClass() const
+    {
+        for (size_t idx = 0; idx < getClassCount(); idx++) {
+            if (auto c = std::dynamic_pointer_cast<T>(getClass(idx)))
+                return c;
+        }
+        return nullptr;
+    }
+
+    /**
+     * @brief Get a class object of type T in index `idx` in the logical array containing
+     *        all classes of type T.
+     * 
+     * @tparam T class template
+     * @param idx index in the class T array
+     * @return std::shared_ptr<T> to the T class at index `idx` in the class T array
+     */
     template <class T> std::shared_ptr<T> getClass(size_t idx) const
     {
-        return std::dynamic_pointer_cast<T>(getClass(idx));
+        size_t idx_ = 0;
+        for (size_t i = 0; i < getClassCount(); i++) {
+            if (auto c = std::dynamic_pointer_cast<T>(getClass(i))) {
+                if (idx_++ == idx)
+                    return c;
+            }
+        }
+        return nullptr;
+    }
+
+    /**
+     * @brief Get the number of classes of type T
+     * 
+     * @tparam T class template
+     * @return size_t number of classes of type T
+     */
+    template <class T> size_t getClassCount() const
+    {
+        size_t count = 0;
+        for (size_t i = 0; i < getClassCount(); i++) {
+            if (auto c = std::dynamic_pointer_cast<T>(getClass(i))) {
+                count++;
+            }
+        }
+        return count;
     }
 
     size_t getClassCount() const;
--- a/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
@@ -123,6 +123,7 @@ public:
     bool getNextTlvType(eTlvType &tlvType) const;
     int getNextTlvType() const;
     uint16_t getNextTlvLength() const;
+    uint8_t *getNextTlvData() const;
     void swap();
     bool is_finalized() const { return m_finalized; };
     bool is_swapped() const { return m_swapped; };
--- a/framework/tlvf/AutoGenerated/include/tlvf/CmduMessageRx.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/CmduMessageRx.h
@@ -12,12 +12,15 @@
 #ifndef _CmduMessageRX_H_
 #define _CmduMessageRX_H_
 
-#include "ieee_1905_1/cCmduHeader.h"
-#include "ieee_1905_1/tlvVendorSpecific.h"
+#include <tlvf/ieee_1905_1/cCmduHeader.h>
+#include <tlvf/CmduMessage.h>
 
-#include "CmduMessage.h"
+#include <list>
+#include <memory>
 
 namespace ieee1905_1 {
+
+class CmduParser;
 
 class CmduMessageRx : public CmduMessage {
 
@@ -25,10 +28,11 @@ public:
     CmduMessageRx();
     CmduMessageRx(CmduMessageRx &original);
     ~CmduMessageRx();
-
-public:
-    std::shared_ptr<cCmduHeader> parse(uint8_t *buff, size_t buff_len, bool swap_needed = true);
+    bool parse(uint8_t *buff, size_t buff_len, bool swap_needed = true, bool parse_tlvs = false);
     CmduMessageRx &operator=(const CmduMessageRx &) = delete;
+
+private:
+    std::list<std::shared_ptr<CmduParser>> parsers_;
 };
 
 }; // close namespace: ieee1905_1
--- a/framework/tlvf/AutoGenerated/include/tlvf/CmduParser.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/CmduParser.h
@@ -1,0 +1,34 @@
+///////////////////////////////////////
+// AUTO GENERATED FILE - DO NOT EDIT //
+///////////////////////////////////////
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _CmduParser_H_
+#define _CmduParser_H_
+
+namespace ieee1905_1 {
+
+class CmduMessageRx;
+
+class CmduParser {
+
+public:
+    CmduParser() = delete;
+    CmduParser(CmduMessageRx &cmdu) : cmdu_(cmdu) {}
+    ~CmduParser() {};
+    virtual bool parse() = 0;
+    CmduParser &operator=(const CmduParser &) = delete;
+
+protected:
+    CmduMessageRx &cmdu_;
+};
+
+}; // close namespace: ieee1905_1
+
+#endif //_CmduParser_H_
--- a/framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
+++ b/framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
@@ -10,10 +10,15 @@
  */
 
 #include <tlvf/CmduMessageRx.h>
+#include <tlvf/CmduParser.h>
+#include <tlvf/CmduTlvParser.h>
 
 using namespace ieee1905_1;
 
-CmduMessageRx::CmduMessageRx() : CmduMessage() {}
+CmduMessageRx::CmduMessageRx() : CmduMessage()
+{
+    parsers_.emplace_back(std::make_shared<CmduTlvParser>(*this)); // default parser
+}
 
 CmduMessageRx::CmduMessageRx(CmduMessageRx &original) : CmduMessage()
 {
@@ -34,12 +39,7 @@ CmduMessageRx::~CmduMessageRx()
     }
 }
 
-/*
- * TODO
- * change all pointer arguments to const type where applicable
- */
-
-std::shared_ptr<cCmduHeader> CmduMessageRx::parse(uint8_t *buff, size_t buff_len, bool swap_needed)
+bool CmduMessageRx::parse(uint8_t *buff, size_t buff_len, bool swap_needed, bool parse_tlvs)
 {
     reset();
     m_parse       = true;
@@ -49,7 +49,17 @@ std::shared_ptr<cCmduHeader> CmduMessageRx::parse(uint8_t *buff, size_t buff_len
     m_cmdu_header = std::make_shared<cCmduHeader>(buff, buff_len, true, false);
     if (!m_cmdu_header || m_cmdu_header->isInitialized() == false) {
         m_cmdu_header = nullptr;
+        return false;
+    }
+    
+    if (!parse_tlvs)
+        return true;
+
+    for (auto parser : parsers_) {
+        if (parser->parse())
+            return true;
     }
 
-    return m_cmdu_header;
+    
+    return false;
 }
--- a/framework/tlvf/AutoGenerated/src/CmduTlvParser.cpp
+++ b/framework/tlvf/AutoGenerated/src/CmduTlvParser.cpp
@@ -1,0 +1,219 @@
+///////////////////////////////////////
+// AUTO GENERATED FILE - DO NOT EDIT //
+///////////////////////////////////////
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include <tlvf/CmduMessageRx.h>
+#include <tlvf/CmduTlvParser.h>
+#include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
+#include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
+#include <tlvf/ieee_1905_1/tlvMacAddress.h>
+#include <tlvf/ieee_1905_1/tlvDeviceInformation.h>
+#include <tlvf/ieee_1905_1/tlvDeviceBridgingCapability.h>
+#include <tlvf/ieee_1905_1/tlvNon1905neighborDeviceList.h>
+#include <tlvf/ieee_1905_1/tlv1905NeighborDevice.h>
+#include <tlvf/ieee_1905_1/tlvLinkMetricQuery.h>
+#include <tlvf/ieee_1905_1/tlvTransmitterLinkMetric.h>
+#include <tlvf/ieee_1905_1/tlvReceiverLinkMetric.h>
+#include <tlvf/ieee_1905_1/tlvVendorSpecific.h>
+#include <tlvf/ieee_1905_1/tlvLinkMetricResultCode.h>
+#include <tlvf/ieee_1905_1/tlvSearchedRole.h>
+#include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
+#include <tlvf/ieee_1905_1/tlvSupportedRole.h>
+#include <tlvf/ieee_1905_1/tlvSupportedFreqBand.h>
+#include <tlvf/ieee_1905_1/tlvPushButtonEventNotification.h>
+#include <tlvf/ieee_1905_1/tlvPushButtonJoinNotification.h>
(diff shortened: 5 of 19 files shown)


framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -31,12 +71,145 @@ CmduMessageRx::~CmduMessageRx()
...
+    if (*wsc->data() == WSC::WSC_MSG_TYPE_M2)
+        return addClass<tlvWscM2>();
+    else
+        return addClass<tlvWscM1>();
+
+    return nullptr;
+}
 
-std::shared_ptr<cCmduHeader> CmduMessageRx::parse(uint8_t *buff, size_t buff_len, bool swap_needed)
+std::shared_ptr<BaseClass> CmduMessageRx::parseNextTlv()
```
LiorAmram: Did you manually write this method.........?

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -46,7 +219,17 @@ std::shared_ptr<cCmduHeader> CmduMessageRx::parse(uint8_t *buff, size_t buff_len
     m_cmdu_header = std::make_shared<cCmduHeader>(buff, buff_len, true, false);
     if (!m_cmdu_header || m_cmdu_header->isInitialized() == false) {
         m_cmdu_header = nullptr;
+        return false;
+    }
+    
+    if (!parse_tlvs)
+        return true;
+
+    while (auto tlv = parseNextTlv()) {
```
LiorAmram: What does the standard say about additional TLVs after TLV EOM?
I think that if there are additional TLVs the frame is invalid, so it should be checked

framework/tlvf/src/src/CmduMessageRx.cpp
tomereli: No, @Galvien updated tlvf in a different commit which generates the switch statement. I think however that since it is not bound to be changed too often, that we can have it here statically instead of generating it each time.

framework/tlvf/src/src/CmduMessageRx.cpp
tomereli: The standard doesn't say anything about that, but having extraneous data after EOM TLV is fine and actually needed on some occasions, such as small packets (padding). We also saw that almost all golden units have extraneous data after EOM TLVs in some CMDUs, so we certainly shouldn't ignore such CMDUs.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
```diff
@@ -62,6 +62,15 @@ class CmduMessage {
...
     }
 
+    template <class T> std::shared_ptr<T> getClass() const
+    {
+        for (size_t idx = 0; idx < getClassCount(); idx++) {
+            if (auto c = getClass<T>(idx))
+                return c;
+        }
+        return nullptr;
+    }
```
morantr: Am I missing something here? I don't understand the purpose of this loop.
The routine is looping through all the classes, from class idx=0 to the last class which was added, but return always on the first iteration (idx=0). So isn't it like always `return getClass<T>(0);`.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessageRx.h
```diff
@@ -25,10 +23,12 @@ class CmduMessageRx : public CmduMessage {
...
     ~CmduMessageRx();
-
-public:
-    std::shared_ptr<cCmduHeader> parse(uint8_t *buff, size_t buff_len, bool swap_needed = true);
+    bool parse(uint8_t *buff, size_t buff_len, bool swap_needed = true, bool parse_tlvs = false);
     CmduMessageRx &operator=(const CmduMessageRx &) = delete;
+
+private:
+    std::shared_ptr<BaseClass> parseNextTlv();
+    std::shared_ptr<BaseClass> parseWscTlv();
```
morantr: I wonder if it correct to refer to a specific TLV in that class, and not give it a more general name like "parseAttributesTlv".

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
     }
 }
 
-/*
- * TODO
- * change all pointer arguments to const type where applicable
- */
+#pragma pack(push, 1)
+struct Tlv {
+    protected:
```
morantr: I wonder why did you choose protected and not private? Do you have plans to inherit it?
Same things on the "WSC".

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
...
+        uint8_t tlvType;
+        uint16_t tlvLength;
+
+    public:
+        uint8_t type() { return tlvType; }
+        Tlv *next() { return (Tlv *)((uint8_t *)this + size()); }
+        uint8_t *data() { return (uint8_t *)this + sizeof(Tlv); }
+        size_t size() { return sizeof(Tlv) + ntohs(tlvLength); }
```
morantr: You used here C-style casting. Since you can't use function notation casting on pointers, then the correct way to do it on CPP is with reinterpret_cast.
Same on WSC class, and parseWscTlv().

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
...
+    };
+#pragma pack(pop)
+
+// packed WSC Attribute struct to help in the parsing of the packet
+#pragma pack(push, 1)
+struct Wsc {
```
morantr: A more correct name would be "Attribute" IMO.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
...
+    protected:
+        uint16_t wsc_type;
+        uint16_t length;
+
+    public:
+        uint16_t type() { return ntohs(wsc_type); }
+        Wsc *next() { return (Wsc *)((uint8_t *)this + size()); }
+        uint8_t *data() { return (uint8_t *)this + sizeof(Wsc); }
+        size_t size() { return sizeof(Wsc) + ntohs(length); }
```
morantr: Isn't `ntohs()` is required only if swap is needed?
Same thing on Tlv struct.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
...
+std::shared_ptr<BaseClass> CmduMessageRx::parseNextTlv()
+{
+    switch(getNextTlvType())
+    {
+        case (0):{
+        return addClass<tlvEndOfMessage>();
+        }
+        case (1):{
+        return addClass<tlvAlMacAddressType>();
+        }
```
morantr: 1. The clang-format.sh script would force zero indentation between the `switch` and `case`, and put an indentation between the `case` and the case body.
2. The brackets wrapping the tlv type values are redundant and can be removed.
3. the curly brackets can be removed as well because scoping is irrelevant here. It will squeeze the code length so I think it is a good idea.
4. I think it will be better to use the enum of the tlv type instead of an integer, even if it will be needed to create several switch cases for every TlvType enum (eTlvTypeMap/eTlvType). It will be more readable, and the reader will not need to guess of integer '#' is actually represent a specific tlv or not.
If not, at least represent it in hex notation is it will be easier to find.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
...
+        case (14):{
+        return addClass<tlvAutoconfigFreqBand>();
+        }
+        case (15):{
+        return addClass<tlvSupportedRole>();
+        }
+        case (16):{
+        return addClass<tlvSupportedFreqBand>();
+        }
+        #if 1
```
morantr: why `#if 1` ?

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
...
+        case (161):{
+        return addClass<wfa_map::tlvApCapability>();
+        }
+        default:{
+        return addClass<tlvUnknown>();
+        }
+    }
+}
+
+bool CmduMessageRx::parse(uint8_t *buff, size_t buff_len, bool swap_needed, bool parse_tlvs)
```
morantr, line 42: If you return a bool, you need to check if some places on the code still refer to the previous cmdu_header return value.
For example, on beerocks_message.cpp: print_cmdu_type();

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -49,7 +250,17 @@ std::shared_ptr<cCmduHeader> CmduMessageRx::parse(uint8_t *buff, size_t buff_len
     m_cmdu_header = std::make_shared<cCmduHeader>(buff, buff_len, true, false);
     if (!m_cmdu_header || m_cmdu_header->isInitialized() == false) {
         m_cmdu_header = nullptr;
+        return false;
+    }
+    
+    if (!parse_tlvs)
+        return true;
+
+    while (auto tlv = parseNextTlv()) {
```
morantr: If there is an error and tlvEndofMessage is not exist, we could get stuck here on an infinite loop. I think you should add stop condition with a maximum number of iteration we would accept, and return false if we reach it.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -49,7 +250,17 @@ std::shared_ptr<cCmduHeader> CmduMessageRx::parse(uint8_t *buff, size_t buff_len
...
+        if (std::dynamic_pointer_cast<tlvEndOfMessage>(tlv)) {
```
morantr: I wonder if the dynamic pointer cast could return success if the TLV is not tlvEndOfMessage. I'm not sure.
I would explicitly compare the tlv type to the tlvEndOfMessage type.

morantr (changes requested):

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: You are missing something probably, I'll try to explain:
`getClass<T>(idx)` is actually `std::dynamic_pointer_cast<T>(getClass(idx))`.
So, this basically goes through each class in the vector and tries to dynamically cast it to the requested type.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessageRx.h
tomereli: should be a nitpick - but note that this is initial implementation, which will be soon changed:
We will have a parent `CmduParser` class from which the following will derive:
- `TlvParser` which parses 1905 and mAP TLVs
- `WscAttrParser` which parses WSC attributes for WSC TLVs
- `IntelVsParser` which parses Intel Vendor specific messages

So name changes here are irrelevant, since again - they are bound to be changed soon.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: Actually I took this from the transport code base, you are right - these can be private, or even public - we can define them as simple structs with functions. I'll change that in a fixup commit.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: Accepted. Will change to `reinterpret_cast`.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: right - will fix. nice catch.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: How about WscAttribute?

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: This was autogenerated and in the future will be autogenerated during build, so @Galvien please take this into account and fix your script accordingly - see #467

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: leftovers, will remove it.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: Since it passed compilation I didn't bother (and since in most cases it is just used to check if the function call was successful - `if (!parse(...))`) - but i'll double check that I didn't miss other places.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: Not needed, we will fail on parseNextTlv() which will reach the end of the buffer. No infinite loop here, error code will be handled internally.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: This is exactly what is done in `parseNextTlv()`. It will fail if it is not exactly this class, so no need to do any explicit comparisons.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
morantr: Then why the `return? it will return in the first iteration, no?
Still, I don't see what benefit it gives to cast it.
Also, I didn't see any use if it.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
morantr: The problem with WSC, it that it is specific to WSC. But theoretically, each tlv could have attributes like WSC. So I think a general name like "attribute" is more fit.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: > It will return in the first iteration, no?

No - `if (auto c = getClass<T>(idx))` will be `true` only if `getClass<T>(idx)` returned none NULL - which means the class at index `idx` is of type `T`.

> Also, I didn't see any use of it.

This is because this PR prepares all that's needed for automatic parsing but doesn't change any logic outside of TLVF to do the automatic parsing. It is used in `tlvf_test`:

```
CmduMessageRx received_message;
received_message.parse(recv_buffer, sizeof(recv_buffer), true, true);
auto tlv1_ = received_message.getClass<tlvNon1905neighborDeviceList>();
  if (!tlv1_)
    errors++;
```

As you can see - this function allows the user to simply call getClass<TYPE> and get nullptr if this TLV was not part of the CMDU, otherwise it returns the class it self already casted.

**Bottom line - this is one of the most important changes in this PR.**

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: I've decided to go with `wscAttribute`. I don't like to depend on theory, better stick to what we know and have.

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
tomereli: I looked and its fine, cmdu_header is only used for checking if the function call was successful.

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
morantr: Thanks for the clarification! I can now understand that the loop will continue to iterate until the class in index `idx` will match to the function template T.
It wasn't so clear at the first look, so thanks.
Ok, but what you do if you have several TLVs from the same type?
For example Channel Selection Response message may contain One or **more** Channel Selection Response TLVs.
So I think you should always iterate from beginning to the end, and when you got a match, push it to std::list<std::shared_ptr<T>> which you will return as the function return value. If the list will be empty, then it means that the TLV was not found.

framework/tlvf/src/src/CmduMessageRx.cpp
LiorAmram: I think that's a bad practice to insert auto generated code into non-auto generated files.
A better approach would be exporting this logic to an external class (which would be autogenerated), and have this class as a member of CmduMessageRx. That way, "parseNextTlv" would only call this class' method.

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -7,6 +7,45 @@
  */
 
 #include <tlvf/CmduMessageRx.h>
+#include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
```
LiorAmram: Same here, all of these includes don't belong here.
Same external class solution I suggested would solve this as well.

LiorAmram (changes requested):

framework/tlvf/AutoGenerated/include/tlvf/CmduParser.h
```diff
@@ -0,0 +1,34 @@
...
+#define _CmduParser_H_
+
+namespace ieee1905_1 {
+
+class CmduMessageRx;
+
+class CmduParser {
+
+public:
+    CmduParser() = delete;
```
morantr, line 22: AFAIK There is no need to explicitly delete the empty constructor because another constructor is defined and it will be the default constructor. So even if you delete this line, this should get you a compilation error:
```CmduParser parser;```

File: framework/tlvf/AutoGenerated/src/CmduTlvParser.cpp [hunk 16]
```diff
@@ -0,0 +1,219 @@
...
+class wscAttr {
+        uint16_t type_;
+        uint16_t length_;
+    public:
+        uint16_t type(bool swap) { return swap ? ntohs(type_) : type_; }
+        wscAttr *next(bool swap)
+        {
+            return reinterpret_cast<wscAttr*>(reinterpret_cast<uint8_t*>(this) + size(swap));
+        }
+        uint8_t *data() { return reinterpret_cast<uint8_t*>(this) + sizeof(wscAttr); }
```
morantr, line 71: Although I don't agree with it, you defined that the `*` sign shall not be attached to the type.
So it should be `<uint8_t *> ` (with space between)
If the clang-format script would have run on this file, it would have to change it. Just saying... 
This comment apply to several places on this PR.

framework/tlvf/src/src/CmduMessageRx.cpp
tomereli: Fixed in a big fixup splitting to new classes and files. We planned to do that anyway but since you asked so nicely.... :)

framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
tomereli: Most of the time you will just look for one class, so always returning a list is an overkill.
Moreover - The user of this API knows in advance if he has more than one element, in this case we will provide an additional API to return the class of type T at index idx.
We currently have such an API but it just tries to cast the class at idx `idx` out of *all classes* into type `T`. I will change it so that it will iterate through all classes, and only those which are of type `T` will be counted for - so it will behave as if there is an array which contains all the classes of type `T` and return the one at index `idx`. See last fixup for this.

framework/tlvf/src/src/CmduMessageRx.cpp
tomereli: Done

morantr (changes requested):

framework/tlvf/AutoGenerated/include/tlvf/CmduParser.h
tomereli: There is no need - in this you are right, but the compilation error, in this case, is more readable and therefore I prefer to keep it. 
It is considered good coding practice to explicitly delete default constructor if it is not used.

framework/tlvf/AutoGenerated/src/CmduTlvParser.cpp
tomereli: The whole of TLVF is not bound to clang format, so it serves little value to do that now.

morantr (approved):

LiorAmram (approved):

Issue #457, opened 2019-10-30
[TASK] Extend BWL to get supported and disabled channels


CoralMalachi:
## Description
Currently, the BWL does not report about the following :

- Statically inoperable channels.

- Supported channels.

The task is to extend the BWL to get the info about operating classes and statically inoperable channels and to set the acs_report properly (in ap_wlan_hal::read_acs_report).

The following iw tool command can be used as a reference to how it can be done :

- iw list (which output the following info ...)
![image](https://user-images.githubusercontent.com/38204874/67870967-51c33a00-fb38-11e9-8d2a-6da5ed7bac75.png)

Also, we can use the iw implementation for this method:
![image](https://user-images.githubusercontent.com/38204874/67874487-a74e1580-fb3d-11e9-9fa7-0b80b5ab5870.png)

Please look here https://godoc.org/github.com/remyoudompheng/go-netlink/nl80211.

- To get the channels, we need to use FREQUENCY_ATTR_FREQ and the "ieee80211_frequency_to_channel" method which is defined in util.c file  (https://github.com/Distrotech/iw/blob/distrotech-iw/util.c)

- To check if a current channel is disabled use FREQUENCY_ATTR_DISABLED.

- To get a radar detection channel use FREQUENCY_ATTR_RADAR.

- We can use the "print_phy_handler" method as a reference, which is defined in info.c file ----> https://github.com/Distrotech/iw/blob/distrotech-iw/info.c

tomereli:
![](https://github.trello.services/images/mini-trello-icon.png) [BWL nl80211 implementation](https://trello.com/c/xocDS5qw/23-bwl-nl80211-implementation)

PR #459, opened 2019-10-30
Feature/AP Radio Basic Capabilities and Channel Preference Report

CoralMalachi:
When the controller sends a Channel Preference Query, the agent should send
a Channel Preference Report that corresponds to the actual behavior of the AP.

Currently, the agent fills the Channel Preference Report with a dummy
operating class and dummies channels. Therefore, Change it so the
Channel Preference Report contains actual operating classes. 

Please note that this PR is depended on #457.
The part of setting the inoperable channels in Radio Basic Capabilities TLV will be done after #457 is implemented.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4230,6 +4230,20 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
 }
 
 /**
+ * @param[in] None
+ * @return Map of Operating classes and matching supported channels
+ */
+std::map<uint8_t, std::set<uint8_t>> slave_thread::get_supported_channels_map()
+{
+    auto channels_list = hostap_params.supported_channels;
+    std::set<uint8_t> channels_set;
+    for (size_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+        channels_set.insert(channels_list[i].channel);
+    }
+    return wireless_utils::channel_set_to_operating_classes(channels_set);
+}
+
+/**
  * @brief Diffie-Hellman public key exchange keys calculation
  *        class member params authkey and keywrapauth are computed
  *        on success.
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -5179,24 +5179,31 @@ bool slave_thread::add_radio_basic_capabilities()
     //TODO get maximum supported VAPs from DWPAL
     radio_basic_caps->maximum_number_of_bsss_supported() = 2;
 
-    auto operationClassesInfo = radio_basic_caps->create_operating_classes_info_list();
-    if (!operationClassesInfo) {
-        LOG(ERROR) << "Failed creating operating classes info list";
-        return false;
-    }
-
     // TODO: Currently sending dummy values, just one operating class based on the band. Need
     // to read them from DWPAL and use the correct WiFi Parameters based on the regulatory
     // domain
-    uint8_t operating_class                            = hostap_params.iface_is_5ghz ? 110 : 81;
-    operationClassesInfo->operating_class()            = operating_class;
-    operationClassesInfo->maximum_transmit_power_dbm() = 0;
 
-    if (!radio_basic_caps->add_operating_classes_info_list(operationClassesInfo)) {
-        LOG(ERROR) << "add_operating_classes_info_list failed";
-        return false;
+    // Get supported channels and operating classes .
+    auto m = get_supported_channels_map();
+
+    for (auto it = m.begin(); it != m.end(); ++it) {
+        // Create operating class object
+        auto operationClassesInfo = radio_basic_caps->create_operating_classes_info_list();
+        if (!operationClassesInfo) {
+            LOG(ERROR) << "Failed creating operating classes info list";
+            return false;
+        }
+        operationClassesInfo->operating_class() = it->first;
+        //TODO: Get real Maximum transmit
+        operationClassesInfo->maximum_transmit_power_dbm() = 0;
+        //TODO: Set number of statically Non-operable channels in the operating class and their numbers
+
+        // Push operating class object to the list of operating class objects
+        if (!radio_basic_caps->add_operating_classes_info_list(operationClassesInfo)) {
+            LOG(ERROR) << "add_operating_classes_info_list failed";
+            return false;
+        }
     }
-
     return true;
 }
 
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4944,12 +4944,8 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
 
     channel_preference_tlv->radio_uid() = network_utils::mac_from_string(config.radio_identifier);
 
-    // Create operating class object
-    auto op_class_channels = channel_preference_tlv->create_operating_classes_list();
-    if (!op_class_channels) {
-        LOG(ERROR) << "create_operating_classes_list() has failed!";
-        return false;
-    }
+    // Get supported channels and operating classes .
+    auto m = get_supported_channels_map();
 
     // TODO: check that the data is parsed properly after fixing the following bug:
     // Since sFlags is defined after dynamic list cPreferenceOperatingClasses it cause data override
@@ -4957,33 +4953,42 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
     // See: https://github.com/prplfoundation/prplMesh/issues/8
 
     // Fill operating class object
-    op_class_channels->operating_class() = 80; // random operating class for test purpose
+    for (auto it = m.begin(); it != m.end(); ++it) {
+        // Create operating class object
+        auto op_class_channels = channel_preference_tlv->create_operating_classes_list();
 
-    // Fill up channels list in operating class object, with test values
-    for (uint8_t ch = 36; ch < 50; ch += 2) {
-        // allocate 1 channel
-        if (!op_class_channels->alloc_channel_list()) {
-            LOG(ERROR) << "alloc_channel_list() has failed!";
+        if (!op_class_channels) {
+            LOG(ERROR) << "create_operating_classes_list() has failed!";
             return false;
         }
-        auto channel_idx = op_class_channels->channel_list_length();
-        auto channel     = op_class_channels->channel_list(channel_idx - 1);
-        if (!channel) {
-            LOG(ERROR) << "getting channel entry has failed!";
+
+        op_class_channels->operating_class() = it->first;
+        // Fill up channels list in operating class object
+        for (auto &ch : it->second) {
+            // allocate 1 channel
+            if (!op_class_channels->alloc_channel_list()) {
+                LOG(ERROR) << "alloc_channel_list() has failed!";
+                return false;
+            }
+            auto channel_idx = op_class_channels->channel_list_length();
+            auto channel     = op_class_channels->channel_list(channel_idx - 1);
+            if (!channel) {
+                LOG(ERROR) << "getting channel entry has failed!";
+                return false;
+            }
+            *channel = ch;
+        }
+
+        //TODO: get real preference score, and set Non-operable channel score to zero.
+        op_class_channels->flags().preference = 15; // channels preference
+        op_class_channels->flags().reason_code =
+            wfa_map::cPreferenceOperatingClasses::eReasonCode::UNSPECIFIED;
+
+        // Push operating class object to the list of operating class objects
+        if (!channel_preference_tlv->add_operating_classes_list(op_class_channels)) {
+            LOG(ERROR) << "add_operating_classes_list() has failed!";
             return false;
         }
-        *channel = ch;
-    }
-
-    // Update channel list flags
-    op_class_channels->flags().preference = 15; // channels preference
-    op_class_channels->flags().reason_code =
-        wfa_map::cPreferenceOperatingClasses::eReasonCode::UNSPECIFIED;
-
-    // Push operating class object to the list of operating class objects
-    if (!channel_preference_tlv->add_operating_classes_list(op_class_channels)) {
-        LOG(ERROR) << "add_operating_classes_list() has failed!";
-        return false;
     }
 
     LOG(DEBUG) << "sending channel preference report for ruid=" << config.radio_identifier;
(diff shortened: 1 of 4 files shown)


common/beerocks/bcl/include/beerocks/bcl/son/son_wireless_utils.h
```diff
@@ -248,6 +249,8 @@ class wireless_utils {
         {1300, 189}, {1350, 157}, {1560, 196}, {1620, 202}, {1755, 208}, {1800, 161}, {2160, 213},
         {2340, 241}, {2430, 220}, {2633, 169}, {2700, 232}, {2925, 189}, {3240, 237}, {3510, 251},
         {3600, 246}, {3900, 328}, {4680, 251}, {5265, 253}, {5850, 261}, {7020, 266}, {7800, 266}};
+
+    static const std::map<uint8_t, std::set<uint8_t>> operating_class_to_channels_map;
```
morantr, line 255: What is the purpose of declaring it here, and define it on the cpp file?
static members should be `constexpr` and not `const` and should have a definition on the cpp file.
Personally, since it is a private static member, I would move it outside of the class definition to the cpp file as static const, and use it without `wirless_utils::` on the class functions.

Whatever you decide to do, please align the rest of the class static members with it.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -464,45 +501,8 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
  */
 std::set<uint8_t> wireless_utils::operating_class_to_channel_set(uint8_t operating_class)
```
morantr, line 528: You could add to `operating_class_to_channel_set` operatinng class `0` with an empty set of channels that will be used in cases of errors. That way you could return `const std::set<uint8_t>&` and prevent the creation of a copy.
In case of error, you will return the `std::set` on operation class `0` of course (which is empty).

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -493,6 +493,32 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+                } else {
+                    std::set<uint8_t> s;
+                    s.insert(ch);
+                    map_result.insert({op_class, s});
+                }
+            }
+        }
+    }
+    return map_result;
+}
```
morantr, line 520: Several things about this function:
1. For std::set functions argument, const reference should be used.
2.  I don't understand why the return value is not just a list of operating classes? Inserting the channels from the input to the output is redundant. As well, inserting each time one element to std::set is not a good idea for performance. 
3. The lookup isn't accurate. A list of channels cannot imply operating class alone. It needs to be coupled with bandwidth. You need to find a way to add properties to the channel list and add there the bandwidth (and you might as well add band_5g, dfs) and lookup using the properties.
I suggest replacing the channel set to std::list, and to create a union in 1 byte size which will hold the preference and be pushed to the operating class channel list at the end/beginning:
```cpp
union sOpClassPreference { 
    struct {
        uint8_t reserved  : 4;
        uint8_t band_5g   : 1;
        uint8_t bandwidth : 2;
        uint8_t dfs       : 1;
    } bits;        
    uint8_t byte;
}; 
```

4. There is no need to loop over the channels argument, you can iterate on the `operating_class_to_channels_map` immediately and lookup if all the channels with a specific bandwidth comply with the operating class.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4229,6 +4229,20 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
                                   backhaul_params.bridge_mac);
 }
 
+/**
+ * @param[in] None
```
morantr, line 4233: Missing `@breif`. Also, I don't think that writing `@param[in] None` is needed when function doesn't receive an argument.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4229,6 +4229,20 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+ * @return Map of Operating classes and matching supported channels
+ */
+std::map<uint8_t, std::set<uint8_t>> slave_thread::get_supported_channels_map()
+{
+    auto channels_list = hostap_params.supported_channels;
+    std::set<uint8_t> channels_set;
+    for (size_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+        channels_set.insert(channels_list[i].channel);
+    }
+    return wireless_utils::channel_set_to_operating_classes(channels_set);
```
morantr, line 4243: I don't see how why this function is needed.
First, as I have mentioned in another comment, inserting elements to `std::set` is not a good idea from a performance perspective, but it is ok if it happens once. This is not the case here, because you will call it on every time the agent will receive a channel preference query.
I recommend holding a container on the class .h file and update it when the ap_manager joins.
Second, since this function is called only from one place, it can be removed. To shorten things up you could use the constructor of the container to immediately construct the list.
```cpp
  int myints[]= {10,20,30,40,50};
  std::set<int> second (myints,myints+5);        // range
```
This is also supported by `std::list`.
Of course, it will cause to insert zero to the container since `channels_list` is an array with fixed size so it will be a great idea to insert the actual length to the cmdu.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4944,46 +4944,51 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
     channel_preference_tlv->radio_uid() = network_utils::mac_from_string(config.radio_identifier);
 
-    // Create operating class object
-    auto op_class_channels = channel_preference_tlv->create_operating_classes_list();
-    if (!op_class_channels) {
-        LOG(ERROR) << "create_operating_classes_list() has failed!";
-        return false;
-    }
+    // Get supported channels and operating classes .
+    auto m = get_supported_channels_map();
```
morantr, line 4948: Please try to give it a more meaningful name, and define it right before you actually use it.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4944,46 +4944,51 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
 
     // TODO: check that the data is parsed properly after fixing the following bug:
     // Since sFlags is defined after dynamic list cPreferenceOperatingClasses it cause data override
     // on the first channel on the list and sFlags itself.
     // See: https://github.com/prplfoundation/prplMesh/issues/8
 
     // Fill operating class object
-    op_class_channels->operating_class() = 80; // random operating class for test purpose
+    for (auto it = m.begin(); it != m.end(); ++it) {
```
morantr, line 4956: Always prefer ranged for-loops over iterator-style for-loops if possible:
```cpp
for (const auto &element : elements) {
    // code
}
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4944,46 +4944,51 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
         }
-        *channel = ch;
-    }
 
-    // Update channel list flags
-    op_class_channels->flags().preference = 15; // channels preference
-    op_class_channels->flags().reason_code =
-        wfa_map::cPreferenceOperatingClasses::eReasonCode::UNSPECIFIED;
+        //TODO: get real preference score, and set Non-operable channel score to zero.
+        op_class_channels->flags().preference = 15; // channels preference
```
morantr, line 4983: The problem here is that you assume that the radio will always support all the channels on a specific operating class.
This is not will always be true.
For example, lets take the operating class 125 which contain channels {149, 153,157, 161,165, 169} with BW=20 MHz.
The radio can report it supports these channels, except for the last channel - 169.
In that case you need to add operating class 2 times:
1) With channel list {149, 153,157, 161,165} and preference "15".
2) With channel list {169} with preference "0" (non-operable).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -5179,24 +5179,31 @@ bool slave_thread::add_radio_basic_capabilities()
...
-    operationClassesInfo->maximum_transmit_power_dbm() = 0;
 
-    if (!radio_basic_caps->add_operating_classes_info_list(operationClassesInfo)) {
-        LOG(ERROR) << "add_operating_classes_info_list failed";
-        return false;
-    }
+    // Get supported channels and operating classes .
+    auto m = get_supported_channels_map();
+
+    for (auto it = m.begin(); it != m.end(); ++it) {
```
morantr, line 5189: Always prefer ranged for-loops.

morantr (changes requested):

common/beerocks/bcl/include/beerocks/bcl/son/son_wireless_utils.h
CoralMalachi: I'll change it to constexpr 👍

common/beerocks/bcl/include/beerocks/bcl/son/son_wireless_utils.h
```diff
@@ -111,6 +111,8 @@ class wireless_utils {
                                                                beerocks::eWiFiBandwidth bw,
                                                                uint16_t vht_center_frequency);
     static std::set<uint8_t> operating_class_to_channel_set(uint8_t operating_class);
+    static std::map<uint8_t, std::set<uint8_t>>
+    channel_set_to_operating_classes(std::set<uint8_t> channels);
```
arnout, line 115: An operating class is a bit more than just a set of channels: it also encodes bandwidth, max power, ....

Therefore, I think there should be a struct or class definition for operating class. For now, it can have the channel set as its only member, and we can add the other information later.

Note that we'll probably also need several kinds of operating class definitions. We have the static definition that can be used in the `operating_class_to_channels_map` but then there is also a radio-specific value that may include statically inoperable channels, the hostap-specific value that includes channel preferences, and a dynamic value that includes current DFS information.

I'm not 100% sure what would be the best way to go. I see two large possibilities:

- We could move the struct WiFiChannel that is currently defined in bwl to bcl and use it everywhere - possibly restructuring it a bit so it contains only static information (I think noise, bss_overlap and radar_affected are dynamic). The operclass would then just be a set of WiFiChannels, and the operclass's bandwidth and power would be implied by it. Since the WiFiChannel contains all information, there's a 1-to-1 mapping of set-of-WiFiChannels to set-of-operclasses-with-excluded-channels, so we could use WiFiChannels everywhere internally and only use operclasses for the EasyMesh messages.
- The alternative (and the direction we were going in, but I'm putting that back into question now) is to have a WiFiChannel that is just a channel number + operating class (it could be useful to have fields for the channel's band and DFS, but that's just caching). Internally we'd still use WiFiChannels everywhere, and the list of operclasses is simply constructed by taking the list of WiFiChannels, summarizing their operclasses, and adding non-operable markers for any channel of that operclass that is not used.

I still believe the second way is the best way to go, but I just want to make the options explicit here.

In either case, it's clear that we need to work with WiFiChannel objects, *not* with uint8_t channel numbers. As noted already by @morantr, the concept of a set of channels is useless, it needs to be combined with bandwidth and power. If we use the first approach (bandwidth and power are encoded in the WiFiChannel object directly), then we'd have a conversion function from channel number + bandwidth + power to WiFiChannel, and a conversion function from set-of-WiFiChannel to set-of-operclass-with-excluded-channels. If we use the second approach, we'd also have a conversion function channel number + bandwidth + power to WiFiChannel, but it would return a set of WiFiChannel instead of a single one. Which is probably better anyway: when you support 80MHz bandwidth, you also support the 40MHz bandwidth option, etc. There is no need for a WiFiChannels -> operating classes function in that case because it's dead simple, it can be done directly from the place where the EasyMesh CMDU is constructed.

For sure, the bwl interface should use WiFiChannel objects and not channels. I also don't think it should use operating classes.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Please note though that he standard specified the following:

> Channels for which the Multi-AP Agent’s radio is statically (permanently) unable to operate are reported in the AP
> Radio Basic Capabilities TLV. A Channel Preference Report message does not include information on operating classes
> that are not supported by the corresponding radio of the Multi-AP Agent per the AP Radio Basic Capabilities TLV.

That second sentence makes it ambiguous, but I read it as: statically inoperable channels should not be reported in the channel preference TLV.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4944,46 +4944,51 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
+            *channel = ch;
         }
-        *channel = ch;
-    }
 
-    // Update channel list flags
-    op_class_channels->flags().preference = 15; // channels preference
-    op_class_channels->flags().reason_code =
-        wfa_map::cPreferenceOperatingClasses::eReasonCode::UNSPECIFIED;
+        //TODO: get real preference score, and set Non-operable channel score to zero.
```
arnout, line 4982: Note that the specification says that channels which are left out of the TLV are assumed to have preference level 15. So until we have real preference values, the only thing we really need to do here is to set dynamic non-operable channels to 0.

I'm not sure yet how we will deal with channel preferences other than 0 or 15: should they come from prplMesh itself or from hostapd? You obviously need some input from hostapd to get a channel availability report, but it kind of makes sense to do the channel selection and therefore also preference assignment in prplMesh itself.

For now though, I'd just use 0 for radar detection. But we need a wifi expert to give some input about this :-)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4944,46 +4944,51 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
 
-    // Push operating class object to the list of operating class objects
-    if (!channel_preference_tlv->add_operating_classes_list(op_class_channels)) {
-        LOG(ERROR) << "add_operating_classes_list() has failed!";
-        return false;
+        // Push operating class object to the list of operating class objects
+        if (!channel_preference_tlv->add_operating_classes_list(op_class_channels)) {
+            LOG(ERROR) << "add_operating_classes_list() has failed!";
+            return false;
+        }
```
arnout, line 4991: Since it is also possible to send an unsolicited channel report, I think the construction of the TLV (or maybe even the entire CMDU) should be moved into a separate function.

arnout (changes requested):
There's a bug in the first commit message: it says "prevent reuse", but of course we want to encourage reuse... So it should either be "To be able to reuse code", or "To prevent duplication of code".

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Maybe it's better to have list of operating classes with list of non-operable channels inside it. Radio Basic Caps should have list of non-operable channel in the operating class.

arnout:
Superseded by #507

PR #464, opened 2019-11-03
Dev/tlv parser hierarchy

tomereli:


framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -59,6 +59,15 @@ class CmduMessage {
         return std::dynamic_pointer_cast<T>(getClass(idx));
     }
 
+    template <class T> std::shared_ptr<T> getClass() const
+    {
+        for (size_t idx = 0; idx < getClassCount(); idx++) {
+            if (auto c = getClass<T>(idx))
```
arnout, line 65: I think it makes very little sense to go through `getClass<T>` which itself goes through the BaseClass version of the function. IMO it is a lot easier to understand what's happening here if you expand it:

```
    for (auto it& : m_class_vector) {
        if (auto c = std::dynamic_pointer_cast<T>(it))
```

Note however that this change has already been committed to master in 9c7b8fb2a68 as part of #458, without my approval :-P

framework/tlvf/AutoGenerated/src/CmduMessageRx.cpp
```diff
@@ -34,12 +73,174 @@ CmduMessageRx::~CmduMessageRx()
...
+struct Tlv {
+    protected:
+        uint8_t tlvType;
+        uint16_t tlvLength;
+
+    public:
+        uint8_t type() { return tlvType; }
+        Tlv *next() { return (Tlv *)((uint8_t *)this + size()); }
+        uint8_t *data() { return (uint8_t *)this + sizeof(Tlv); }
+        size_t size() { return sizeof(Tlv) + ntohs(tlvLength); }
```
arnout: Due to the swapping that is done, the ntohs here is dangerous...

framework/tlvf/src/include/tlvf/CmduParser.h
```diff
@@ -0,0 +1,35 @@
...
+#include <memory>
+#include <BaseClass.h>
+
+namespace ieee1905_1 {
+    class CmduParser
+    {
+    private:
+    public:
+CmduParser();
+    virtual std::shared_ptr<BaseClass> Parse(CmduMessageRx &cmdu_rx)=0;
```
arnout, line 20: I don't think we have any function anywhere that is using upper camel case. In other words, it should be called parse in lowercase.

framework/tlvf/src/include/tlvf/CmduParser.h
```diff
@@ -0,0 +1,35 @@
...
+ * See LICENSE file for more details.
+ */
+#ifndef _CmduParser_H_
+#define _CmduParser_H_
+
+#include <memory>
+#include <BaseClass.h>
+
+namespace ieee1905_1 {
+    class CmduParser
```
arnout, line 15: Formatting is all wrong, please use clang-format on it (clang-format.sh skips it so it has to be done manually).

framework/tlvf/src/include/tlvf/CmduParser.h
```diff
@@ -0,0 +1,35 @@
...
+    {
+    private:
+    public:
+CmduParser();
+    virtual std::shared_ptr<BaseClass> Parse(CmduMessageRx &cmdu_rx)=0;
+        // CmduParser(/* args */);
```
arnout, line 21: Please don't include any commented out code.

framework/tlvf/src/include/tlvf/CmduParser.h
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+#ifndef _CmduParser_H_
```
arnout, line 8: I'm not sure if CmduParser is the right name. It is not just for CMDUs, it's for any type,length,value-structured sequence.

So I'd rather call it TLVListParser, and also create a TLVList class (from which CmduMessage derives) to collect all the functions that manipulate lists of TLVs.

framework/tlvf/src/include/tlvf/TlvMapParser.h
```diff
@@ -0,0 +1,25 @@
...
+namespace ieee1905_1 {
+
+class TlvMapParser : public CmduParser {
+public:
+    TlvMapParser();
+    std::shared_ptr<BaseClass> Parse(CmduMessageRx &cmdu_rx);
+
+};
+};
+#endif //_TlvMapParser_H_
```
arnout, line 25: Missing newline at EOF

framework/tlvf/src/include/tlvf/CmduParser.h
```diff
@@ -0,0 +1,35 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
```
arnout, line 3: 2019 only.

framework/tlvf/src/include/tlvf/TlvMapParser.cpp
```diff
@@ -0,0 +1,171 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
```
arnout, line 3: 2019 only.

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -227,8 +227,12 @@ bool CmduMessageRx::parse(uint8_t *buff, size_t buff_len, bool swap_needed, bool
...
     if (!parse_tlvs)
         return true;
-
-    while (auto tlv = parseNextTlv()) {
+    //need to beter classify message types in order to fit them into dictionary values,
+    //currently -1 is the default TlvMapParser.
+    //for example 4 is for vendor specific messages.
+    int msgtype = -1;
+    CmduParser* parser = (parsers[msgtype]).get();
+    while (std::shared_ptr<BaseClass> tlv = parser->Parse(*this)){
```
arnout, line 235: I'm probably missing something major here, but why do we need a dynamic list of parsers?

The only tricky one is the M1/M2 distinction. But that's actually easy: we can add a single tlvWSC that only has fields type, length and value. Then it's up to the caller to cast it to M1 or M2 as appropriate.

PR #465, opened 2019-11-04
controller: DEV_SET_CONFIG: remove old configs for AL-MAC 

### LINKED ISSUE #195: [TASK] Pass flow 5.4.2 test in testbed
## Description

Perform all steps necessary to pass the test associated with flow 5.4.2

## Exit criteria

### LINKED ISSUE #199: 5.4.3 AP configuration BSS Tear Down
## Description

The prplMesh controller needs to pass the test case as defined in 5.4.3

## Exit criteria

The flow is completed once the controller passes this test in the scenario and testbed described in the upstream definition.

- [ ] #197 
- [ ] #198

CoralMalachi:
currently, test 5.4.2 keeps on failing with the following message :

image

It means that the SSID was not changed as it was supposed to.
The solution is to remove all old configs for the AL-MAC address when we
get a DEV_SET_CONFIG command.

After this change, it was verified that the AP configuration renew
succeeded and the SSID was changed as required.
Fixes #195 

In addition, this PR is also fixed #199, by adding a validation check for the 
bss_info_conf content.

### DIFF
--- a/controller/src/beerocks/master/controller_ucc_listener.cpp
+++ b/controller/src/beerocks/master/controller_ucc_listener.cpp
@@ -8,6 +8,7 @@
 
 #include "controller_ucc_listener.h"
 #include "son_actions.h"
+#include <unordered_set>
 
 using namespace beerocks;
 
@@ -80,6 +81,7 @@ bool controller_ucc_listener::send_cmdu_to_destination(Socket *sd,
 bool controller_ucc_listener::handle_dev_set_config(
     std::unordered_map<std::string, std::string> &params, std::string &err_string)
 {
+    std::unordered_set<sMacAddr> al_mac_cleared_conf;
     if (params.find("backhaul") != params.end()) {
         err_string = "parameter 'backhaul' is not relevant to the controller";
         return false;
@@ -103,8 +105,12 @@ bool controller_ucc_listener::handle_dev_set_config(
             return false;
         }
 
-        m_database.add_bss_info_configuration(net::network_utils::mac_from_string(al_mac),
-                                              bss_info_conf);
+        auto mac = net::network_utils::mac_from_string(al_mac);
+        if (al_mac_cleared_conf.find(mac) == al_mac_cleared_conf.end()) {
+            m_database.clear_bss_info_configuration(mac);
+            al_mac_cleared_conf.insert(mac);
+        }
+        m_database.add_bss_info_configuration(mac, bss_info_conf);
     }
     return true;
 }
--- a/controller/src/beerocks/master/controller_ucc_listener.cpp
+++ b/controller/src/beerocks/master/controller_ucc_listener.cpp
@@ -110,6 +110,11 @@ bool controller_ucc_listener::handle_dev_set_config(
             m_database.clear_bss_info_configuration(mac);
             al_mac_cleared_conf.insert(mac);
         }
+
+        //If SSID is empty, tear down - do not add bss_info_conf to database.
+        if (bss_info_conf.ssid.empty()) {
+            continue;
+        }
         m_database.add_bss_info_configuration(mac, bss_info_conf);
     }
     return true;
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -3190,6 +3190,8 @@ std::list<db::bss_info_conf_t> &db::get_bss_info_configuration(const sMacAddr &a
 
 void db::clear_bss_info_configuration() { bss_infos.clear(); }
 
+void db::clear_bss_info_configuration(const sMacAddr &al_mac) { bss_infos[al_mac].clear(); }
+
 //
 // PRIVATE FUNCTIONS
 //   must be used from a thread safe context
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -638,7 +638,7 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1> 
         config_data.multiap_attr().subelement_value = bss_info_conf->bss_type;
 
         LOG(DEBUG) << "WSC config_data:" << std::hex << std::endl
-                   << "     ssid: " << config_data.ssid() << std::endl
+                   << "     ssid: " << config_data.ssid_str() << std::endl
                    << "     authentication_type: "
                    << int(config_data.authentication_type_attr().data) << std::endl
                    << "     encryption_type: " << int(config_data.encryption_type_attr().data)
--- a/tools/docker/tests/test_flows.sh
+++ b/tools/docker/tests/test_flows.sh
@@ -82,8 +82,32 @@ test_ap_config_renew() {
 }
 
 test_ap_config_bss_tear_down() {
-    err "ap_config_bss_tear_down not implemented yet."
-    return 0
+    status "test ap config bss tear down"
+    # Regression test: MAC address should be case insensitive
+    MAC_AGENT1=$(echo $mac_agent1 | tr a-z A-Z)
+    # Configure the controller and send renew
+    send_CAPI_command gateway "DEV_RESET_DEFAULT" $redirect
+    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$mac_agent1 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1" $redirect
+
+    gw_mac_without_colons="$(printf $mac_gateway | tr -d :)"
+    send_CAPI_command gateway "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x000A,tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,0x${gw_mac_without_colons},tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{0x00},tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{0x00}}" $redirect
+
+    sleep 3
+    check docker exec -it repeater1 sh -c \
+        'grep -i -q "ssid: Multi-AP-24G-1, .* fronthaul" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+    check docker exec -it repeater1 sh -c \
+        'grep -i -q "ssid: .* teardown" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+
+    # SSIDs have been removed for the CTT Agent1's front radio
+    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$MAC_AGENT1 8x" $redirect
+    # Send renew message
+    send_CAPI_command gateway "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x000A,tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,0x${gw_mac_without_colons},tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{0x00},tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{0x00}}" $redirect
+
+    sleep 3
+    check docker exec -it repeater1 sh -c \
+        'grep -i -q "ssid: .* teardown" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+    
+    return $check_error
 }
 
 test_channel_selection() {
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -529,6 +529,7 @@ public:
     void add_bss_info_configuration(const sMacAddr &al_mac, const bss_info_conf_t &bss_info);
     std::list<db::bss_info_conf_t> &get_bss_info_configuration(const sMacAddr &al_mac);
     void clear_bss_info_configuration();
+    void clear_bss_info_configuration(const sMacAddr &al_mac);
 
     //
     // tasks


tomereli (approved):

arnout:
I haven't reviewed the details yet, but one overall comment: instead of prepending to the list, we should remove all old configs for the AL-MAC address when we get a DEV_SET_CONFIG command.

tomereli (changes requested):
please update PR headline to reflect the new changes we do here, fix issues with CI and test it at least once in the certification testbed before we actually merge it.

CoralMalachi:
Please do not merge this PR.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -3190,6 +3190,8 @@ std::list<db::bss_info_conf_t> &db::get_bss_info_configuration(const sMacAddr &a
 
 void db::clear_bss_info_configuration() { bss_infos.clear(); }
 
+void db::clear_bss_info_configuration(const sMacAddr &al_mac) { bss_infos[al_mac].clear(); }
```
morantr, line 3193: It will be more suitable to edit the exist clear function and add it to it an optional value of MacAddr.
Of course that it will cause the function to be with more that one line so I will be needed to move the implementation to the CPP file.

controller/src/beerocks/master/wfa_ca.cpp
```diff
@@ -708,7 +710,11 @@ void wfa_ca::handle_wfa_ca_message(
                 reply(sd, cmdu_tx, eWfaCaStatus::INVALID, err_string);
                 return;
             }
-
+            //remove all old configs for the AL-MAC address when we get a DEV_SET_CONFIG command
+            if (!old_configs_removed) {
+                database.clear_bss_info_configuration(net::network_utils::mac_from_string(al_mac));
+                old_configs_removed = true;
+            }
```
morantr: What is the purpose of the flag `old_configs_removed` and the `if` statement? 
`old_configs_removed` is always set to `false` a couple of lines before, so the code inside the if statement will always happen, so it makes the `if` statement redundant. Also in the `if` scope you change the value old_configs_removed  to `true` but it is not used afterward so it turns out that this flag is redundant too.

morantr (changes requested):
I don't understand one thing. Before the UCC set config it should remove the config first with DEV_RESET_DEFAULT, so how come this bug happens?

CoralMalachi:
> I don't understand one thing. Before the UCC set config it should remove the config first with DEV_RESET_DEFAULT, so how come this bug happens?

The UCC does it only at the beginning of the test before the first DEV_SET_CONFIG.
In 5.4.2 test the UCC sends DEV_SET_CONFIG twice, and before the second time there is no DEV_RESET_DEFAULT, so the old config is still there...

controller/src/beerocks/master/wfa_ca.cpp
CoralMalachi: I did it since I want to remove the old configuration and then set the new one.
In each DEV_SET_CONFIG there are two SSID one for each VAP,
If I call the clear method in each iteration, at the second iteration I'll remove the first SSID I just added to the database, so the boolean variable is used to clear the old configuration the right way.
If you have a better way to do this I'll try it .

morantr: Oh sorry, you are right. I didn't notice that the `old_configs_removed` is defined outside the loop.
But still, this solution has a hole in it.
Since you used a flag, which you set it one time, it will create a problem if there are more than one AL_MACs. So, in that case, it will work for the first AL_MAC, but for the second, third and go on, it will not delete the configuration.
I suggest creating unordered_set of al_mac which we have cleared configuration on:

```cpp
std::unordered_set<std::string> al_mac_cleared_conf;
// code 
// for loop
    // code
    if (al_mac_cleared_conf.find(al_mac) == al_mac_cleared_conf.end()) {
          database.clear_bss_info_configuration(net::network_utils::mac_from_string(al_mac));
          al_mac_cleared_conf.insert(al_mac);
    }
```

tomereli (approved):
Fixes #195 - that's not entirely true. Say that it fixes some of the issues which are causing #195 to fail.
Other than that - looks fine.

controller/src/beerocks/master/db/db.cpp
arnout: > It will be more suitable to edit the exist clear function and add it to it an optional value of MacAddr.

I disagree, I think it makes more sense that it's an overload. An optional argument has two disadvantages:

- It needs to be a pointer (otherwise it's not optional) which makes the code slightly more difficult to read.
- The implementation will simply have a condition on the argument being nullptr or not; two different function signature are much easier to read.

> Of course that it will cause the function to be with more that one line so I will be needed to move the implementation to the CPP file.

It already is in the cpp file :-)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -621,7 +621,7 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
     // Create ConfigData
     uint8_t buf[1024];
     WSC::cConfigData config_data(buf, sizeof(buf), false, true);
-    if (bss_info_conf) {
+    if (bss_info_conf && !bss_info_conf->ssid.empty() && !bss_info_conf->network_key.empty()) {
```
arnout: I don't like this. I think this empty SSID is an artifact of the CAPI and not generic enough to warrant this special treatment here. So I think the empty SSID should be handled in handle_dev_set_config() instead.

Also, an empty network key definitely doesn't mean that it should be cleared. If it's an open network, there will be no network key.

arnout (changes requested):
Note that this needs to be redone entirely because the code has moved. So please rebase on master before resubmitting for review.

I think the commit messages could be much improved by saying what commands UCC sends and what result those commands are expected to have. The details of how this is found in the certification logs are not so important. So, for example, for the first fix, it could be:

```
UCC sends the following commands:

dev_set_config,program,map,bss_info1,be:66:8b:4c:f7:0e 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,bss_info2,be:66:8b:4c:f7:0e 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0
...
dev_set_config,program,map,bss_info1,be:66:8b:4c:f7:0e 8x Multi-AP-24G-1-diff 0x0020 0x0008 maprocks1 0 1,bss_info2,be:66:8b:4c:f7:0e 8x Multi-AP-24G-2-diff 0x0020 0x0008 maprocks2 1 0

The intention is that the second `dev_set_config` replaces 
the first one entirely, i.e. all bss_info for the specified AL-MAC are removed.
However, currently we append the new bss_info to the existing ones.
Therefore, clear the existing bss_info when dev_set_config is done.
```

Similar for the second fix.

Also I wonder if it wouldn't be possible to simply clear the entire bss_info for all AL-MAC when the dev_set_config is given. But it's still not clear to me what the intention is for UCC. I think the only case where dev_set_config is given multiple times is in 5.4.2 and 5.4.3, where the intention really is to clear it.

controller/src/beerocks/master/db/db.cpp
morantr: > It already is in the cpp file :-)

Yeah I confused 😏 

> >It will be more suitable to edit the exist clear function and add it to it an optional value of MacAddr.
> 
> I disagree, I think it makes more sense that it's an overload. An optional argument has two disadvantages:
> 
> It needs to be a pointer (otherwise it's not optional) which makes the code slightly more difficult to read.
> The implementation will simply have a condition on the argument being nullptr or not; two different function signature are much easier to read.

It does not have to be a pointer, it can still be a reference. I tested the following code and it worked:

```cpp
#include <iostream>
#include <string>

std::string ret_str(const std::string &str = "hello world") 
{
  return str;
}

int main() {
  std::cout << ret_str() << std::endl;
}
```
`output: hello world`

arnout: > > > It will be more suitable to edit the exist clear function and add it to it an optional value of MacAddr.
> > I disagree, I think it makes more sense that it's an overload. An optional argument has two disadvantages:
> > It needs to be a pointer (otherwise it's not optional) which makes the code slightly more difficult to read.
> It does not have to be a pointer, it can still be a reference. I tested the following code and it worked:
> ```c++
> std::string ret_str(const std::string &str = "hello world") 
> ```

Yes, but what would you give as default?

```
    void clear_bss_info_configuration(const sMacAddr &al_mac = ????);
```

You could use `ZERO_MAC`, indeed, then the issue with the pointer is gone. But still you would need a condition in the implementation, and you'd need it as well in the documentation (which is now missing anyway). If you have it as separate function signatures, the condition is implied.

Clearly this is a nitpicking issue, but it would be good to decide if we in general prefer optional arguments to be declared with default parameter values and a condition in the implementation, or as two separate overloaded functions (where one potentially calls the other).

controller/src/beerocks/master/controller_ucc_listener.cpp
```diff
@@ -80,6 +81,7 @@ bool controller_ucc_listener::send_cmdu_to_destination(Socket *sd,
 bool controller_ucc_listener::handle_dev_set_config(
     std::unordered_map<std::string, std::string> &params, std::string &err_string)
 {
+    std::unordered_set<std::string> al_mac_cleared_conf;
```
arnout: I would have used sMacAddr instead of std::string, but that's probably nitpicking.

Hang on, it's not nitpicking at all: you need the first commit because you're using a string here instead of sMacAddr! This shows that there is a good reason to keep both the upper and lower case in the test.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -630,15 +630,15 @@ bool master_thread::autoconfig_wsc_add_m2(std::shared_ptr<ieee1905_1::tlvWscM1>
...
+    if (bss_info_conf) {
         config_data.set_ssid(bss_info_conf->ssid);
         config_data.authentication_type_attr().data = bss_info_conf->authentication_type;
         config_data.encryption_type_attr().data     = bss_info_conf->encryption_type;
         config_data.set_network_key(bss_info_conf->network_key);
         config_data.multiap_attr().subelement_value = bss_info_conf->bss_type;
 
         LOG(DEBUG) << "WSC config_data:" << std::hex << std::endl
-                   << "     ssid: " << config_data.ssid() << std::endl
+                   << "     ssid: " << config_data.ssid_str() << std::endl
```
arnout, line 641: I would have done this change in a separate commit, but again that's nitpicking.

tools/docker/tests/test_flows.sh
```diff
@@ -59,7 +59,7 @@ test_initial_ap_config() {
     MAC_AGENT1=$(echo $mac_agent1 | tr a-z A-Z)
     # Configure the controller and send renew
     send_CAPI_command gateway "DEV_RESET_DEFAULT" $redirect
-    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$MAC_AGENT1 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,bss_info2,$mac_agent1 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0" $redirect
+    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$MAC_AGENT1 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,bss_info2,$MAC_AGENT1 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0" $redirect
```
arnout: NACK this. The whole point of this MAC_AGENT1 is to verify that it is really case insensitive, so we want *both* upper and lower case version in the same configuration.

tools/docker/tests/test_flows.sh
```diff
@@ -82,8 +82,34 @@ test_ap_config_renew() {
 }
 
 test_ap_config_bss_tear_down() {
-    err "ap_config_bss_tear_down not implemented yet."
-    return 0
+    status "test ap config bss tear down"
+    # Regression test: MAC address should be case insensitive
```
arnout, line 86: To make it a real regression test, I think it's better to add it with one case and to remove it with a different case. So the simplest would be to only use the uppercase version in the second dev_set_config.

tools/docker/tests/test_flows.sh
```diff
@@ -82,8 +82,34 @@ test_ap_config_renew() {
...
+    MAC_AGENT1=$(echo $mac_agent1 | tr a-z A-Z)
+    # Configure the controller and send renew
+    send_CAPI_command gateway "DEV_RESET_DEFAULT" $redirect
+    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$MAC_AGENT1 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,bss_info2,$MAC_AGENT1 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0" $redirect
```
arnout:
```suggestion
    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$mac_agent1 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,bss_info2,$mac_agent1 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0" $redirect
```

And maybe a single BSS is sufficient too, so remove bss_info2.

tools/docker/tests/test_flows.sh
```diff
@@ -82,8 +82,34 @@ test_ap_config_renew() {
...
+        'grep -i -q "ssid: Multi-AP-24G-1, .* fronthaul" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+    check docker exec -it repeater1 sh -c \
+        'grep -i -q "ssid: Multi-AP-24G-2, .* backhaul" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+    check docker exec -it repeater1 sh -c \
+        'grep -i -q "ssid: .* teardown" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+
+    # SSIDs have been removed for the CTT Agent1's front radio
+    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$MAC_AGENT1 8x" $redirect
+    # Send renew message
+    send_CAPI_command gateway "DEV_SEND_1905,DestALid,$MAC_AGENT1,MessageTypeValue,0x000A,tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,0x${gw_mac_without_colons},tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{0x00},tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{0x00}}" $redirect
```
arnout:
```suggestion
    send_CAPI_command gateway "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x000A,tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,0x${gw_mac_without_colons},tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{0x00},tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{0x00}}" $redirect
```

arnout (changes requested):

arnout (approved):
It's an excellent idea to add a test for the teardown flow!

Even though I have change requests that are essential, I'm going to approve now already because the fixes are trivial to implement, and because I won't be back until Tuesday to approve an updated version. @morantr can do the final review.

morantr (approved):

Issue #469, opened 2019-11-04
[TASK] Add possibility for building statically


arnout:
## Description

prplMesh now consists of a many relatively small shared libraries. These have a significant memory footprint. We would like to have the possibility to build static libraries instead, which may reduce the size (especially if it is combined with link-time optimisation).

Instead of forcing shared libraries, allow to build static libraries with the standard CMake option `BUILD_SHARED_LIBS`.

bwl and bpl are currently even dlopen'ed, but there's no need for that. The libraries are not replaced dynamically, they are fixed for a specific platform.

## Exit criteria

- It is possible to build with `-DBUILD_SHARED_LIBS=OFF`
- Analyze executable size difference
- Identify other routes to decrease size

vitalybu:
Memory or disk footprint?
Maybe I'm missing something, but shared libraries usually reduce memory footprint, as they are loaded only once into the memory.
If the BCL for example will link statically to all the executables, they'll take considerably more space both on disk and in-memory.

arnout:
![](https://github.trello.services/images/mini-trello-icon.png) [Support static linking](https://trello.com/c/gjsv6UWn/35-support-static-linking)

arnout:
An attempt was made in #560 but doesn't completely work. For the time being, we're putting this on the back burner.

ghost:
This is needed for Econet

ghost:
@arnout Can this issue be closed? Or is further work needed?

juanschroeder:
What I did for Econet (econet/econet_static) was to do a full static build, including libc and dependencies. I added a new CMake option BUILD_FULL_STATIC.
Is that what was needed here?

arnout:
I think the BUILD_FULL_STATIC is indeed what what we need to aim for. Only I'm not so sure if we really should include that workaround for the buggy Ubuntu toolchain...

arnout:
Closing - we're currently building with CMAKE_SHARED_LIBS=OFF in CI. Full static building is more of an integration thing - we'll take bugfixes for it if someone encounters a problem, but we don't need a separate task for it.

PR #478, opened 2019-11-08
Top-level CMake

### LINKED ISSUE #468: [TASK] Merge CMakeLists into a single project
## Description

Currently, each of the components (framework, common, agent, controller) has a separate "top-level" CMakeLists.txt. That makes it unnecessarily difficult to build the whole - in practice, it has to be done with maptools.py. But using such a script bypasses the standard cmake integration in OpenWrt and in RDK-B entirely. In addition, the separate top-level CMakeLists.txt make it more difficult to set or change global options.

Therefore, there should be a single top-level CMakeLists.txt that builds all the components.

## Exit criteria

The full prplMesh stack can be built with a single top-level call to `cmake`.

## Detailed description

- A previous attempt is still available in the feature/cmake-top-level but that went too far.
- BML (and its dependencies) should still be exported; the rest doesn't have to be. See #1.

arnout:
For proper integration with build systems (OpenWrt and RDK-B) it is much
easier if we can build all of prplMesh in a single go with cmake,
without the need for maptools.py to build all components.

This PR moves the framework, common, controller and agent components to the top level.

Fixes #468.


tomereli (approved):
So far, so good. But this means that we can't build each component by its own anymore, doesn't it?

morantr:
I think we should wait before merge it until we will merge all the changes from the turris omnia.

arnout:
> I think we should wait before merge it until we will merge all the changes from the turris omnia.

OK for me - it indeed creates conflicts that have to be resolved. Then I can include agent and controller into the top-level build in the same PR as well.

And anyway, CI is currently failing on this so it can't be merged as is.

CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+cmake_minimum_required(VERSION 3.0)
+project (prplmesh C CXX)
+
+if(NOT DEFINED TARGET_PLATFORM)
+set(TARGET_PLATFORM "linux")
+endif()
+
+option (BUILD_AGENT "Build EasyMesh agent" ON)
+option (BUILD_CONTROLLER "Build EasyMesh agent" ON)
```
vitalybu: Incorrect option description. Should be "Build EasyMesh controller".

CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+add_subdirectory(framework)
+add_subdirectory(common)
+if (BUILD_AGENT)
+add_subdirectory(agent)
+endif()
+if (BUILD_CONTROLLER)
+add_subdirectory(controller)
+endif()
```
vitalybu: Missing new line at the end of the file

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -121,9 +121,9 @@ bool slave_thread::init()
 
     if (config.ucc_listener_slave_hostap_iface == config.hostap_iface &&
         config.ucc_listener_port != 0) {
-        m_agent_ucc_listener =
-            std::make_unique<agent_ucc_listener>(config.ucc_listener_port, config.vendor,
-                                                 config.model, config.bridge_iface, &master_socket);
+        m_agent_ucc_listener = std::make_unique<agent_ucc_listener>(
+            config.ucc_listener_port, config.vendor, config.model, config.bridge_iface,
+            (Socket **)&master_socket);
```
vitalybu, line 126: A better solution is to change the type of the controller socket to `SocketClient` in the `agent_ucc_listener` class.
Already did it in the UGW Support branch, so can be fixed later.

common/beerocks/btl/CMakeLists.txt
```diff
@@ -27,17 +23,19 @@ set(MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
...
 # Include paths
-target_include_directories(${PROJECT_NAME} 
+target_include_directories(${PROJECT_NAME}
+    PUBLIC
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>
     PRIVATE 
         ${MODULE_PATH} 
         ${MODULE_PATH}/../../../external
-        $<TARGET_PROPERTY:beerocks::bcl,INTERFACE_INCLUDE_DIRECTORIES>
+        $<TARGET_PROPERTY:bcl,INTERFACE_INCLUDE_DIRECTORIES>
```
vitalybu: Why is this required anyway?
It makes sense for BWL and BPL as they are loaded dynamically, but for the BCL that it linked during build time, `target_link_libraries` should be enough, no?
If not, perhaps we should update the target to correctly set the INTERFACE_INCLUDE_DIRECTORIES.

common/beerocks/bwl/brcm/ap_wlan_hal_brcm.h
```diff
@@ -9,8 +9,8 @@
 #ifndef _BWL_AP_WLAN_HAL_BRCM_H_
 #define _BWL_AP_WLAN_HAL_BRCM_H_
 
-#include "../common/ap_wlan_hal.h"
 #include "base_wlan_hal_brcm.h"
+#include <ap_wlan_hal.h>
```
vitalybu: I think that relative paths for files in the same module is better practice.
Can prevent issues when files from different modules have the same name.

This comment is relevant for all BPL/BWL files.

controller/src/beerocks/master/CMakeLists.txt
```diff
@@ -24,19 +20,12 @@ endif()
...
 set_target_properties(${PROJECT_NAME} PROPERTIES LINK_FLAGS "-Wl,-z,defs")
 
-target_link_libraries(${PROJECT_NAME} beerocks::bcl beerocks::btl tlvf elpp beerocks::btlvf)
+target_link_libraries(${PROJECT_NAME} bcl btl tlvf elpp btlvf)
 target_include_directories(${PROJECT_NAME} PRIVATE
     ${MODULE_PATH}/../bml
-    ${MODULE_PATH}/../../../external
-    $<TARGET_PROPERTY:beerocks::bcl,INTERFACE_INCLUDE_DIRECTORIES>
-    $<TARGET_PROPERTY:beerocks::btlvf,INTERFACE_INCLUDE_DIRECTORIES>
+    # $<TARGET_PROPERTY:beerocks::bcl,INTERFACE_INCLUDE_DIRECTORIES>
```
vitalybu: I guess it's not necessary after all...
Should be removed from all other CMake files as well.

vitalybu (changes requested):

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: ok, so once you merge the UGW Support branch you can override it.

common/beerocks/btl/CMakeLists.txt
tomereli: You're right - it's not needed, and neither does the external (which no longer exists nor the MODULE_PATH

common/beerocks/bwl/brcm/ap_wlan_hal_brcm.h
tomereli: The idea is to be able to easily move the BWL/BPL to the framework so we don't maintain 2 different common libraries. I also agree with you that using only the file name is problematic - so how about a compromise, adding a `bwl` directory prefix? for example, `#include <bwl/base_wlan_hal.h>`

tomereli: If you agree with this approach, I will add this change in a separate commit right after this one.

vitalybu: You can keep the relative paths and still move the modules between the projects.
Relative paths are used only for including files from the same logical group of files (e.g. inside the BWL module)

tomereli: OK, so I'll remove this commit. @arnout if you still think this is required, we change it later.

tomereli: On second thought we decided to keep Arnout's change but to use relative paths to `<bwl/...>` `<bpl/,,,>` `<btl/,,` etc. So I will change this commit by Arnout accordingly.

arnout (commented):
I don't agree with making btlvf a static lib.

1. The point of #469 is *not* to make everything static. Rather, the point is to make it *possible* to build static.
2. Linking a static library into a dynamic library only works if the static library is built with `-fPIC`. We happen to do that, but it's quite hidden.
3. This is really bad for code size: the btlvf (which is very large) will be linked into both libbml and the controller executable as weak symbols. LTO is not going to save is there.
4. It's not really needed. Even if btlvf is built as a shared library, that doesn't mean it has to be exported. Since bml is linked dynamically with btlvf, the dynamic linker will resolve it. So it is only needed if bml is built as a static library - but then this doesn't solve it, because the bml static library will *not* include btlvf so we still need to link with btlvf explicitly.
5. btlvf is no different from bcl, elpp, tvlf, so if we do this, the same should be done for those libraries.

So, I removed the last two commits and pushed again. That's what I will merge.

I pushed the last two commits as branch feature/clean-up-bml-export.

I'll make a new branch that does further cmake cleanups which will include most of that, except for the shared->static change.

PR #481, opened 2019-11-11
tlvf: add class_cast for tlv casting support

tomereli:
In order to support automatic parsing for unknown TLVs - for example, VS TLVs which are part of a normal TLV, we need a castTlv API. 
This is because the tlvParser is not aware of beerocks messages for example, for which automatic parsing support can be added but is overkill (they all contain a single VS TLV with the beerocks action header).

Consider the WSC simple Autoconfiguration CMDU for example. It contains standard TLVs which includes the M1 and radio basic capabilities TLVs (in the agent -> controller flow), but also the VS TLV in the Intel agent case which adds extra configuration parameters which are needed only by the Intel controller functionality.

The automatic parser will add the VS TLV as an Unkown TLV, which is fine for most, but not for Intel - since we need the VS TLV itself.

For this reason, this PR adds support for TLV casting, a static function in every TLV class which accepts a base class pointer as input, and returns the casted TLV to its own type.
It does so by simply taking the start buffer pointer from the source TLV, its full length, and calls its own constructor.

### DIFF
--- a/framework/tlvf/src/include/tlvf/BaseClass.h
+++ b/framework/tlvf/src/include/tlvf/BaseClass.h
@@ -13,6 +13,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <string>
+#include <memory>
 
 class BaseClass {
 protected:
@@ -29,6 +30,14 @@ public:
     size_t getLen();
     bool isInitialized();
     virtual void class_swap() = 0;
+    template <class T> std::shared_ptr<T> class_cast()
+    {
+        if (!m_parse__)
+            return nullptr;
+        if (m_swap__)
+            class_swap();
+        return std::make_shared<T>(m_buff__, m_buff_len__, true, m_swap__);
+    }
 
 protected:
     uint8_t *m_buff__;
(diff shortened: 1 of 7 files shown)


morantr (approved):

framework/tlvf/src/include/tlvf/BaseClass.h
```diff
@@ -29,6 +30,14 @@ class BaseClass {
     size_t getLen();
     bool isInitialized();
     virtual void class_swap() = 0;
+    template <class T> std::shared_ptr<T> class_cast()
```
LiorAmram: better have this addition to the signature:

```suggestion
template <class T, class = typename std::enable_if<std::is_base_of<BaseClass, T>::value>::type> std::shared_ptr<T> class_cast()
```

That way this cast would only be available for derived classes.
Check my syntax as it might be a bit off.

LiorAmram (changes requested):

LiorAmram (approved):

tomereli:
@arnout do you want to review this before I merge it?

arnout (changes requested):
As discussed orally with @tomereli, I don't think this PR is going in the right direction.

The `class_cast` function is not something I think we will use, and we shouldn't add dead code.

The first two commits do have value though, since they test a case which was not tested now. So I'd make a new PR with just those two.

tomereli:
superseded by #489. Keeping the branch for future reference if we'll need `class_cast`.

PR #484, opened 2019-11-12
encryption: Fix padding by using openssl internal padding

### LINKED ISSUE #192: [TASK] Pass flow 5.4.1 test in testbed
## Description

Perform all steps necessary to pass the test associated with flow 5.4.1

## Exit criteria

### LINKED ISSUE #195: [TASK] Pass flow 5.4.2 test in testbed

tomereli:
Previously, the aes_encrypt() required the plaintext to be aligned to
block size (16 bytes due to aes128).
Due to this, the config data was zero padded before encryption, which
was wrong since zero padding can only be used if the 0 can't be part of
the plaintext.

The correct padding is described in the WSC specifiction [5]:
"ENCkeywrapkey(...) This notation indicates symmetric encryption of the
values in parentheses using the key keywrapkey. The encryption algorithm
is AES-CBC per FIPS 197, *with PKCS#5 v2.0 padding*".
This padding scheme simply defines each padded character to be the value
of the number of padded characters.

This was the root cause for various failures in autoconfiguration flow
with multiple agents.
Since AES-CBC is a block cipher, it encrypts each block seperately.
So, if the input plaintext padding is wrong, the output ciphertext will
only be wrong in the last block.
Last block checking is optional in openssl (enabled by default), but is
*disabled* in BBF reference implementation, causing most agents to
*accept* invalid encrypted settings and fail spradically since the last
block is all messed up.
Only Qualcomm agent apparently let openssl (or using some internal
impelemntation) perform correctness checks for the last blcoks, which is
why prplMesh never could pass autoconfiguration with Qualcomm agent.

This commit solves all these issues, thus allowing all tests involving
Qualcomm as agent1 or agent2 to pass.
In addition, it solves issues observed with other agents when changing
default SSID which resulted with different padding which caused sporadic
failures in Broadcomm, Marvell, and Mediatek agents.

It does so by letting openssl take care of internal padding, but still
verifies there is enough room in the given ciphertext output buffer for
the necessary padding.

Fixes #192
Fixes #195

References
[1] https://www.di-mgt.com.au/cryptopad.html
[2] https://crypto.stackexchange.com/questions/48628/why-is-padding-used-in-cbc-mode
[3] https://asecuritysite.com/encryption/padding
[4] http://aes.online-domain-tools.com/
[5] WiFi Simple Configuration Technical Specification v2.0.6


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4333,14 +4333,20 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
 {
     auto encrypted_settings = m2->encrypted_settings();
     uint8_t *iv             = reinterpret_cast<uint8_t *>(encrypted_settings->iv());
-    auto buf                = reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings());
+    auto ciphertext         = reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings());
+    int clen                = encrypted_settings->encrypted_settings_length();
```
arnout: I prefer to use `size_t` for sizes. (nitpick)

I'd like somewhat more descriptive naming, even if it's longer, e.g. `cipherlen` or `ciphertext_length`. The latter is probably a bit too much... Similar for dlen of course. (nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4333,14 +4333,20 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
...
+    // leave room for up to 16 bytes internal padding length - see aes_decrypt()
+    int dlen = clen + 16;
+    uint8_t decrypted[dlen];
 
     LOG(DEBUG) << "M2 Parse: aes decrypt";
-    mapf::encryption::aes_decrypt(keywrapkey, iv, buf,
-                                  encrypted_settings->encrypted_settings_length());
+    if (!mapf::encryption::aes_decrypt(keywrapkey, iv, ciphertext, clen, decrypted, dlen)) {
+        LOG(ERROR) << "aes decrypt";
```
arnout: Error message could be improved :-)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,57 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
-    // Calculate length of data to encrypt
-    // (= plaintext length + 64 bits HMAC aligned to 16 bytes boundary)
-    // The Key Wrap Authenticator is 96 bits long
-    size_t len = (config_data.getLen() + sizeof(WSC::sWscAttrKeyWrapAuthenticator) + 15) & ~0xFU;
-
-    auto encrypted_settings = m2->create_encrypted_settings();
-    if (!encrypted_settings)
+    // Step 1 - prepare the plaintext: [config_data | keywrapauth]:
+    // We use the config_data buffer as the plaintext buffer for encryption.
+    // The config_data buffer has room for 8 bytes, but check it anyway
```
arnout: KeyWrapAuth is 96 bits == 12 bytes, so where does this 8 bytes come from?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,57 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
     uint8_t keywrapkey[16])
 {
-    // Calculate length of data to encrypt
-    // (= plaintext length + 64 bits HMAC aligned to 16 bytes boundary)
-    // The Key Wrap Authenticator is 96 bits long
-    size_t len = (config_data.getLen() + sizeof(WSC::sWscAttrKeyWrapAuthenticator) + 15) & ~0xFU;
-
-    auto encrypted_settings = m2->create_encrypted_settings();
-    if (!encrypted_settings)
+    // Step 1 - prepare the plaintext: [config_data | keywrapauth]:
```
arnout, line 446: Good one, a comment like this was missing!

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,57 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
         return false;
-
-    if (!mapf::encryption::aes_encrypt(keywrapkey, iv, buf, len)) {
-        LOG(DEBUG) << "aes encrypt";
+    LOG(DEBUG) << "iv: " << std::endl
+               << utils::dump_buffer(iv, 16) << std::endl
+               << "key: " << std::endl
+               << utils::dump_buffer(keywrapkey, 16) << std::endl
+               << "plaintext: " << std::endl
+               << utils::dump_buffer(ciphertext, plen);
```
arnout: Dumping key and ciphertext is very verbose and rarely useful, so I'd skip them.

framework/common/encryption.cpp
```diff
@@ -223,61 +223,112 @@ bool hmac::digest(uint8_t *digest)
...
+bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int plen,
+                 uint8_t *ciphertext, int &clen)
 {
     EVP_CIPHER_CTX *ctx;
+    int len;
 
-    int clen, len;
-    uint8_t buf[128];
-
+    /* Create and initialise the context */
```
arnout: This comment is pretty useless, "new EVP cipher context" is kind of the same, no?

framework/common/encryption.cpp
```diff
@@ -223,61 +223,112 @@ bool hmac::digest(uint8_t *digest)
...
+        MAPF_ERR("Failed to create context");
         return false;
     }
+
+    /*
+     * Initialise the encryption operation.
+     * IMPORTANT - key and IV should be
+     * The same as block size, which 16 bytes since we are using 
+     * 128 bit AES (i.e. a 128 bit key). The IV size for *most* modes is
+     * the same as the block size.
```
arnout: This kind of comment would be a lot more relevant in the doxygen in the header file, since it's about the interface, not the implementation.

framework/common/encryption.cpp
```diff
@@ -223,61 +223,112 @@ bool hmac::digest(uint8_t *digest)
...
+    /*
+     * Provide the message to be encrypted, and obtain the encrypted output.
+     * EVP_EncryptUpdate() encrypts inl bytes from the buffer in and writes
+     * the encrypted version to out. The amount of data written depends on
+     * the block alignment of the encrypted data: as a result the amount of
+     * data written may be anything from zero bytes to
+     * (inl + cipher_block_size - 1) so out should contain sufficient room.
+     * In this case - clen has to be >= plaintext_len + 15)
+     */
+    if ((clen < plen + 15) || (EVP_EncryptUpdate(ctx, ciphertext, &len, plaintext, plen) != 1)) {
```
arnout: The `clen < plen + 15` check is a check of the input parameters, so it doesn't make sense IMO to do it in the same condition as the EVP_EncryptUpdate call.

Also, the check is not semantically correct: there should just be sufficient space to do padding, not always 15 additional bytes. But I'm OK with using this approximation, just note it in the comment.

framework/common/encryption.cpp
```diff
@@ -223,61 +223,112 @@ bool hmac::digest(uint8_t *digest)
...
-    if (EVP_EncryptUpdate(ctx, data, &clen, data, data_len) != 1 || clen != (int)data_len) {
+    /*
+     * Provide the message to be encrypted, and obtain the encrypted output.
+     * EVP_EncryptUpdate() encrypts inl bytes from the buffer in and writes
+     * the encrypted version to out. The amount of data written depends on
+     * the block alignment of the encrypted data: as a result the amount of
+     * data written may be anything from zero bytes to
+     * (inl + cipher_block_size - 1) so out should contain sufficient room.
+     * In this case - clen has to be >= plaintext_len + 15)
+     */
```
arnout, line 265: This *is* a relevant comment here. :-)

framework/common/encryption.cpp
```diff
@@ -223,61 +223,112 @@ bool hmac::digest(uint8_t *digest)
...
         return false;
     }
-    EVP_CIPHER_CTX_set_padding(ctx, 0);
 
-    plen = data_len;
-    if (EVP_DecryptUpdate(ctx, data, &plen, data, data_len) != 1 || plen != (int)data_len) {
+    /*
+     * Provide the message to be decrypted, and obtain the decrypted output.
+     * The decrypted data buffer out passed to EVP_DecryptUpdate() should have
+     * sufficient room for (inl + cipher_block_size) bytes.
```
arnout: This makes no sense. The decrypted buffer will be *smaller* than the original, because padding is added.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4333,14 +4333,20 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
 {
     auto encrypted_settings = m2->encrypted_settings();
     uint8_t *iv             = reinterpret_cast<uint8_t *>(encrypted_settings->iv());
-    auto buf                = reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings());
+    auto ciphertext         = reinterpret_cast<uint8_t *>(encrypted_settings->encrypted_settings());
+    int clen                = encrypted_settings->encrypted_settings_length();
+    // leave room for up to 16 bytes internal padding length - see aes_decrypt()
+    int dlen = clen + 16;
```
arnout: That's not true. The decrypted length will always be equal to the encrypted length. The padding is added when you encrypt, not when you decrypt. The decrypted plaintext will include the padding bytes. AFAIK there is no "standard" in AES to detect where the actual plaintext ends and where the padding begins - this is something that has to be encoded in other metadata. In the case of WSC, it's encoded in the attribute length. Anything that comes after the keywrap attribute will be ignored.

framework/common/include/mapf/common/encryption.h
```diff
@@ -112,28 +112,43 @@ bool create_iv(uint8_t *iv, unsigned iv_length);
...
- * @param[in] key 32 byte KeyWrapKey calculated according to WSC v2.0.6 specification
- * @param[in] iv random 128bit input vector
- * @param[in/out] plaintext bytestream, aligned to 16 bytes boundary
- * @param[in] data_len plaintext buffer length
+ * @brief Perform AES128 CBC (cipher block chaining) encryption on given plaintext
+ *        and write the output encrypted data into the ciphertext buffer.
+ *        plaintext and ciphertext can be the same buffer (in place encryption).
+ *        Padding is handled internally, so the encrypted buffer can be up to 15
+ *        bytes larger than the plaintext buffer, and therefore this function
+ *        will fail if the given clen is not big enough (>= plen + 15).
```
arnout: This is not exactly brief :-) The @brief should be a single sentence, the rest should be in a separate paragraph.

framework/common/include/mapf/common/encryption.h
```diff
@@ -112,28 +112,43 @@ bool create_iv(uint8_t *iv, unsigned iv_length);
...
+ * 
+ * @param[in] key 16 bytes KeyWrapKey calculated according to WSC v2.0.6 specification
+ * @param[in] iv 16 bytes random initialization vector 
+ * @param[in] plaintext bytestream (data to encrypt)
+ * @param[in] plen plaintext buffer length
+ * @param[in/out] ciphertext output buffer of encrypted data (can be the same as plaintext buffer)
```
arnout: This is just an output parameter.

framework/common/test/encryption_test.cpp
```diff
@@ -25,6 +25,17 @@ static bool check(int &errors, bool check, std::string message)
     return check;
 }
 
+std::string dump_buffer(uint8_t *buffer, size_t len)
```
arnout: Could we promote beerocks_utils::dump_buffer to somewhere in framework, so it can be used here?

framework/common/test/encryption_test.cpp
```diff
@@ -61,27 +72,34 @@ int main()
           "keywrapkeys should be equal");
 
     {
-        uint8_t plaintext[1024];
+        uint8_t plaintext[50];
```
arnout: It would be good to re-run this test for various plaintext sizes, so we actually test different amounts of padding. Ideally also include plaintext size < 16 bytes.

arnout (commented):
I don't think there is really any reason to change the decrypt side. That said, it's a good idea to make separate input and output parameters to make it similar to the encrypt side, but when calling it we can still use the same buffer.

Rest of the comments are mostly nitpick.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: The library takes `int` and not `size_t` so isn't it better to do the casting here instead of inside the `aes_decrypt` function?

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Changed to `aes decypt failure`. A further reason is printed inside `aes_decrypt`.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: From https://linux.die.net/man/3/evp_decryptupdate:
> EVP_DecryptInit_ex(), EVP_DecryptUpdate() and EVP_DecryptFinal_ex() are the corresponding decryption operations. EVP_DecryptFinal() will return an error code if padding is enabled and the final block is not correctly formatted. The parameters and restrictions are identical to the encryption operations except that if padding is enabled the decrypted data buffer out passed to EVP_DecryptUpdate() should have sufficient room for (inl + cipher_block_size) bytes unless the cipher block size is 1 in which case inl bytes is sufficient.

According to this, we can either enable padding checks to be done by the library and be on the safe side, like this commit is doing (and therefore the change to the decrypt function). The downside for letting the library do these checks for us is that we need to provide an output buffer which is one block bigger then the input ciphertext, as done here.

However, we can be less strict and just decrypt the data without this check - if we do we can keep the original implementation as it was but just change the prototype to separate the input and output arguments as you suggested in another comment. I'm fine either way - what do you prefer?

controller/src/beerocks/master/son_master_thread.cpp
tomereli: You are right of course - the key itself is 8 bytes, 2 bytes for the type and 2 for the length, just like any other WSC attribute, will be fixed.

framework/common/encryption.cpp
tomereli: I disagree - IMHO it is more appropriate here since it is our take on the `EVP_EncryptUpdate` function which *will* fail if we don't give it enough buffer space.

tomereli: Fixed.

framework/common/encryption.cpp
tomereli: Again, see https://linux.die.net/man/3/evp_decryptupdate
Having the library perform correctness checks for the last buffer requires the output buffer to have an extra block. Probably since it is doing the check in place..

framework/common/include/mapf/common/encryption.h
tomereli: No, it is in/out since it also states the length of the output buffer for the final block check (should be >16 bytes from plen)

framework/common/test/encryption_test.cpp
tomereli: I forgot to remove this, we don't need to dump the buffer in the test. In any case we can - perhaps @CoralMalachi can do that?

framework/common/test/encryption_test.cpp
tomereli: Great idea, will do that.

tomereli:
> I don't think there is really any reason to change the decrypt side. That said, it's a good idea to make separate input and output parameters to make it similar to the encrypt side, but when calling it we can still use the same buffer.
> 
> Rest of the comments are mostly nitpick.

See comments inside the review about the decrypt side. I pushed a fixup to most your comments so you have a chance to review those before I split this into commits.

framework/common/test/encryption_test.cpp
CoralMalachi: Yes. @tomereli

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Coming to think about it again - I think that *this* is the reason Qualcomm always failed to decrypt while the rest of the agents "succeeded" - I believe that Qualcomm uses this capability in the openssl to verify the correctness of the last block.

So - based on that - I think we should be strict when decrypting instead of just ignoring potential wrong padding. It will make it much easier to identify faulty 3rd party agents / controllers.

arnout: Hm, so apparently I was wrong... I'm pretty sure that specifically for AES, the decrypted size is always equal to the encrypted size, but it's indeed better to follow the EVP_DecryptUpdate constraints.

framework/common/include/mapf/common/encryption.h
arnout: If it is in/out, it means that the ciphertext argument would have to be initialized with something before passing it into the function. Is that really the case? The documentation of the underlying `EVP_Encrypt*` functions also only talks of writing to `out`, not reading from it.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: > The library takes `int` and not `size_t` so isn't it better to do the casting here instead of inside the `aes_decrypt` function?

Argh, what a stupid library openssl is... Yes, in that case it has to be int, otherwise passing it as reference becomes problematic. I should have seen that.

framework/common/include/mapf/common/encryption.h
tomereli: You are right, I mistook it for the plen. Will change it to [out].

framework/common/include/mapf/common/encryption.h
```diff
@@ -156,6 +157,8 @@ bool wps_calculate_keys(const diffie_hellman &dh, const uint8_t *remote_pubkey,
                         unsigned remote_pubkey_length, const uint8_t *m1_nonce, const uint8_t *mac,
                         const uint8_t *m2_nonce, uint8_t *authkey, uint8_t *keywrapkey);
 
+std::string dump_buffer(uint8_t *buffer, size_t len);
```
arnout: This change is not needed, nor is the header change, since it's in utils.h.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,55 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
-    uint8_t *iv = reinterpret_cast<uint8_t *>(m2->encrypted_settings()->iv());
-
-    if (!mapf::encryption::create_iv(iv, WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH))
+    // Step 2 - AES encryption using temporary buffer. This is needed since we only
+    // know the encrypted length after encryption.
+    // Calculate initialization vector (IV), and encrypt the plaintext using aes128 cbc.
+    // leave room for up to 15 bytes internal padding length - see aes_encrypt()
+    int cipherlen = plaintextlen + 15;
+    uint8_t ciphertext[cipherlen];
+    uint8_t iv[WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH];
```
arnout: There's no need for an additional buffer for iv. We can create encrypted_settings here and use its iv directly. It's only alloc_encrypted_settings that has to be delayed until after encryption is done.

framework/common/encryption.cpp
```diff
@@ -223,33 +223,57 @@ bool hmac::digest(uint8_t *digest)
     return HMAC_Final(m_ctx, digest, &digest_length);
 }
 
-bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *data, uint32_t data_len)
+bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int plen,
+                 uint8_t *ciphertext, int &clen)
```
arnout, line 227: I don't like inout length parameters like this very much - I'd prefer passing `size_t clen` as input argument and returning `size_t encrypted_len`. But that's probably more of a personal opinion.

framework/common/encryption.cpp
```diff
@@ -223,33 +223,57 @@ bool hmac::digest(uint8_t *digest)
...
     EVP_CIPHER_CTX *ctx;
+    int len;
 
-    int clen, len;
-    uint8_t buf[128];
+    /* Verify that the ciphertext buffer has enough storage room
+     * for block size alignment padding which will be added
+     * during encryption.
+     */
+    int padlen = plen - ((plen + 15) & 0xFU);
```
arnout: This is actually the part of plen that is full blocks, not the padded length. The padded length is one block more than this. So I think `padlen` is not a good name.

Avoid using `&` to do modulo. It just makes things harder to read, and GCC is perfectly capable of converting `((plen + 15) % 16)` to an AND operation (if that is more efficient, which is not necessarily the case).

framework/common/encryption.cpp
```diff
@@ -223,33 +223,57 @@ bool hmac::digest(uint8_t *digest)
...
+    if (clen < padlen) {
```
arnout: I think this should be `clen < padlen + 16` because pkcs#5 padding always adds at least one padding byte, so if it's a whole number of blocks, one more block will be added.

This is also what is written in the OpenSSL man pages: for EVP_EncryptUpdate, the number of bytes written is up to inl + cipher_block_size - 1, and EVP_EncryptFinal_ex writes one additional block.

framework/common/include/mapf/common/encryption.h
```diff
@@ -113,16 +113,29 @@ bool create_iv(uint8_t *iv, unsigned iv_length);
...
- * @param[in] key 32 byte KeyWrapKey calculated according to WSC v2.0.6 specification
- * @param[in] iv random 128bit input vector
- * @param[in/out] plaintext bytestream, aligned to 16 bytes boundary
- * @param[in] data_len plaintext buffer length
+ * @brief AES128-CBC Encryption
+ * 
+ * Perform AES128 CBC (cipher block chaining) encryption on given plaintext
+ * and write the output encrypted data into the ciphertext buffer.
+ * plaintext and ciphertext can be the same buffer (in place encryption).
+ * Padding is handled internally, so the encrypted buffer can be up to 15
```
arnout: ... so it's up to 16 bytes larger.

framework/common/include/mapf/common/encryption.h
```diff
@@ -113,16 +113,29 @@ bool create_iv(uint8_t *iv, unsigned iv_length);
...
+ * bytes larger than the plaintext buffer, and therefore this function
+ * will fail if the given clen is not big enough (>= plen + 15).
+ * Key and IV should be the same as block size, which 16 bytes
+ * since we are using 128 bit AES (i.e. a 128 bit key).
+ * The IV size for *most* modes is the same as the block size.
+ * 
+ * @param[in] key 16 bytes KeyWrapKey calculated according to WSC v2.0.6 specification
```
arnout: This function is not necessarily specific to WSC. It can be used for any AES128-CBC with PKCS padding. So I wouldn't specify this KeyWrapKey here.

framework/common/include/mapf/common/encryption.h
arnout: It's still in/out...

framework/common/include/mapf/common/encryption.h
```diff
@@ -113,16 +113,29 @@ bool create_iv(uint8_t *iv, unsigned iv_length);
...
+ * @param[in] iv 16 bytes random initialization vector 
+ * @param[in] plaintext bytestream (data to encrypt)
+ * @param[in] plen plaintext buffer length
+ * @param[in/out] ciphertext output buffer of encrypted data (can be the same as plaintext buffer)
+ * @param[in/out] clen in - length of ciphertext buffer (>= plen + 15) out - length of output ciphertext
```
arnout: It's in,out instead of in/out. Cfr. http://doxygen.nl/manual/commands.html#cmdparam

framework/common/test/encryption_test.cpp
```diff
@@ -61,21 +61,26 @@ int main()
...
-        uint8_t data[len]; // last 64 bytes are the KWA
-        std::fill(data, data + sizeof(plaintext), 1);
+        // calculate length of data to encrypt = [plaintext | kwa]
+        // last 8 bytes are the KWA
+        int plaintextlen = sizeof(plaintext) + 8;
+        // Use the same buffer as the ciphertext output buffer
+        // for that, leave enough room for padding (16 bytes aes 128 block size)
+        int padlen    = plaintextlen - ((plaintextlen + 15) & 0xFU);
+        int cipherlen = plaintextlen + padlen;
+        uint8_t data[cipherlen];
```
arnout: We don't really need to test the in-place operation of aes_encrypt() since it's not used. So I'd use `plaintext` directly and calculate `plaintextlen = sizeof(plaintext) - 8` to leave room for the kwa.

framework/common/test/encryption_test.cpp
```diff
@@ -61,21 +61,26 @@ int main()
...
-        size_t len = (sizeof(plaintext) + 8 + 16) & ~0xFU;
-        uint8_t data[len]; // last 64 bytes are the KWA
-        std::fill(data, data + sizeof(plaintext), 1);
+        // calculate length of data to encrypt = [plaintext | kwa]
+        // last 8 bytes are the KWA
+        int plaintextlen = sizeof(plaintext) + 8;
+        // Use the same buffer as the ciphertext output buffer
+        // for that, leave enough room for padding (16 bytes aes 128 block size)
+        int padlen    = plaintextlen - ((plaintextlen + 15) & 0xFU);
+        int cipherlen = plaintextlen + padlen;
```
arnout: You calculated padlen as plaintextlen rounded down to a whole number of blocks. So now you get cipherlen = 2*plaintextlen rounded down. That's definitely big enough, but probably not what you want...

Since padlen is not used anywhere else, a much simpler calculation is to just round up: `int cipherlen = ((plaintextlen + 16) / 16) * 16;`

I think it would actually be better to make the buffers large enough and leave plenty of unused space at the end, e.g. both plaintext and data sized at 1000 bytes, but use no more than 800 bytes.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -221,10 +221,12 @@ bool add_encrypted_settings(std::shared_ptr<tlvWscM2> m2, uint8_t *keywrapkey, u
                << std::endl;
     config_data.class_swap();
 
-    size_t len             = (config_data.getLen() + 16) & ~0xFU;
-    uint8_t encrypted[len] = {0};
+    int datalen                  = config_data.getLen();
+    int padlen                   = datalen - ((datalen + 15) & 0xFU);
+    int cipherlen                = datalen + padlen;
```
arnout: Again, cipherlen is a bit less than two times datalen when calculated like this.

framework/common/include/mapf/common/encryption.h
```diff
@@ -138,16 +137,29 @@ bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int
...
- * @param[in] data_len cyphertext buffer length
+ * @brief AES128-CBC Decryption
+ * 
+ * Perform AES128 CBC (cipher block chaining) decryption on given ciphertext
+ * and write the output decrypted data into the plaintext buffer.
+ * plaintext and ciphertext can be the same buffer (in place decryption).
+ * Padding is handled internally, so the decrypted buffer can be up to 16
+ * bytes larger than the plaintext buffer, and therefore this function
+ * will fail if the given plen is not big enough (>= plen + 16).
+ * Key and IV should be the same as block size, which 16 bytes
```
arnout: which *is* 16 bytes

framework/common/include/mapf/common/encryption.h
```diff
@@ -138,16 +137,29 @@ bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int
...
+ * since we are using 128 bit AES (i.e. a 128 bit key).
+ * The IV size for *most* modes is the same as the block size.
+ * 
+ * @param[in] key 16 byte KeyWrapKey calculated according to WSC v2.0.6 specification
+ * @param[in] iv 16 bytes random initialization vector 
+ * @param[in] ciphertext bytestream (data to deccrypt)
```
arnout: decrypt

framework/common/include/mapf/common/encryption.h
```diff
@@ -138,16 +137,29 @@ bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int
...
+ * @param[in] clen length of ciphertext buffer
+ * @param[in/out] plaintext output buffer of encrypted data (can be the same as ciphertext buffer)
```
arnout: out only

framework/common/include/mapf/common/encryption.h
```diff
@@ -138,16 +137,29 @@ bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int
...
+ * @param[in/out] plen in - length of plaintext buffer (>= clen + 16) out - length of output plaintext
```
arnout: in,out

framework/common/test/encryption_test.cpp
```diff
@@ -80,8 +83,11 @@ int main()
         check(errors,
               mapf::encryption::aes_encrypt(keywrapkey1, iv, data, plaintextlen, data, cipherlen),
               "AES encrypt");
```
arnout, line 90: I'd add a check here that plaintextlen < cipherlen <= plaintextlen + 16

framework/common/test/encryption_test.cpp
```diff
@@ -60,22 +61,28 @@ int main()
     check(errors, std::equal(keywrapkey1, keywrapkey1 + sizeof(keywrapkey1), keywrapkey2),
           "keywrapkeys should be equal");
 
-    {
-        uint8_t plaintext[50];
+    std::list<int> sizes = {24 /* no padding */, 32, 50, 80, 96, 128};
```
arnout: Makes more sense to use a vector or an array for lists that don't change (nitpick).

PKCS#5 *always* adds padding, so instead of "no padding" it should be "multiple of block size". Probably better though to put that in a comment above it, like: "Sizes 24 and 96 yield a multiple of block size after adding kwa. The other sizes are arbitrary samples." But actually, I think it would be better to use sizes that include the kwa because that's easier to understand. This requires bigger changes in the code below though.

Also, I think it's important to have sizes (including kwa) 15, 16, 17 and 815, 816, 817. The other sizes are not so relevant I think. In general, when there is some natural boundary, you want the smallest possible value that is exactly at that boundary and one just above and below, and then you want a pretty large value that is at the boundary and one just above and below. The only other potentially relevant ones are 8, 24, 808 because they're at exactly half the boundary. For the current implementation they're not relevant, but they might become relevant after some code change so it's a good one to catch regressions.

framework/common/test/encryption_test.cpp
```diff
@@ -60,22 +61,28 @@ int main()
...
+        LOG(INFO) << "Test encryption with plaintext size " << size;
+        uint8_t plaintext[size];
         std::fill(plaintext, plaintext + sizeof(plaintext), 1);
         // calculate length of data to encrypt = [plaintext | kwa]
         // last 8 bytes are the KWA
         int plaintextlen = sizeof(plaintext) + 8;
         // Use the same buffer as the ciphertext output buffer
         // for that, leave enough room for padding (16 bytes aes 128 block size)
-        int padlen    = plaintextlen - ((plaintextlen + 15) & 0xFU);
+        int padlen    = ((plaintextlen + 15) & ~0xFU) - plaintextlen;
```
arnout: Now it's correct!

framework/common/test/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ if(BUILD_TESTS AND NOT MSGLIB STREQUAL "None")
 		install(TARGETS ${target} DESTINATION bin/tests/common)
 		add_test(NAME ${target} COMMAND $<TARGET_FILE:${target}>)
 	endforeach(test ${tests})
-else()
+elseif(BUILD_TESTS)
```
arnout: Nitpick: it would be better to have an outer condition BUILD_TESTS and then an inner condition on MSGLIB, which allows to refactor a bit because the code below will always be there so outside of the MSGLIB condition.

arnout (changes requested):
Since there were some mistakes in size calculations, I want to re-check before merging so changes requested.

The first commit message has a spelling mistake: Ofcourese -> Of course

The second commit message is not entirely accurate: although [1] says that zero-padding is only possible for ASCII data, the real limitation is that zero-padding is only possible if there is another way of determining the plaintext length - which is in fact the case of WSC. The real problem is that the WSC specification requires PKCS#5 padding. Also, there are a couple of spelling errors in the text. So I'd rewrite it as follows:

```
encryption: Use openssl internal padding

Previously, aes_encrypt() required the plaintext to be aligned to block
size (16 bytes due to aes128), and zero padding was added before
encryption. However, the WSC specification [5] describes the required
padding: "ENCkeywrapkey(...) This notation indicates symmetric 
encryption of the values in parentheses using the key keywrapkey. The
encryption algorithm is AES-CBC per FIPS 197, *with PKCS#5 v2.0 
padding*". This padding scheme simply defines each padded character to 
be the value of the number of padded characters.

With a wrong padding scheme, the length of the decrypted message can't
be derived correctly. However, it is not really needed to know that
length, because the content is a sequence of attributes which have
their length encoded. The BBF implementation of 1905.1 therefore
disabled the padding check in openssl. Since most agent impelemntations
are derived from the BBF implementation, most agents work with 
incorrect padding.

Only the Qualcomm agent apparently lets openssl (or using some internal
implementation) perform correctness checks for the last block, which is
why prplMesh never could pass autoconfiguration with Qualcomm agent.

This commit enables PKCS#5 padding in encryption. It does so by letting
openssl take care of internal padding, but still verifies there is
enough room in the given ciphertext output buffer for the necessary
padding. This allows all tests involving Qualcomm as agent1 or agent2 
to pass. In addition, it solves issues observed with other agents when
changing default SSID which resulted with different padding which
caused sporadic failures in Broadcomm, Marvell, and Mediatek agents.

Fixes #192
Fixes #195

References
[1] https://www.di-mgt.com.au/cryptopad.html
[2] https://crypto.stackexchange.com/questions/48628/why-is-padding-used-in-cbc-mode
[3] https://asecuritysite.com/encryption/padding
[4] http://aes.online-domain-tools.com/
[5] WiFi Simple Configuration Technical Specification v2.0.6
```

framework/common/include/mapf/common/utils.h
```diff
@@ -0,0 +1,35 @@
...
+ * @brief dump buffer in hex mode
+ * 
+ * Debugging utility to get a string representation of a given
+ * buffer in hexadecimal, split to bytes.
+ * 
+ * @param buffer input buffer
+ * @param len input buffer length
+ * @return std::string string containing the hexadecimal buffer contets
+ */
+std::string dump_buffer(uint8_t *buffer, size_t len);
```
morantr, line 31: We already have hex_dump() on beerocks_utils, why did you add it?

framework/common/encryption.cpp
tomereli: We already discussed this - openssl takes `int`. And regarding the inout parameters - I normally would agree with you on that, but in this case I think it makes the code easier to understand since this is the way the internal openssl APIs work, with in-out clen.

framework/common/encryption.cpp
tomereli: Changed to 
```
int padlen = plen % 16;
if (clen < plen + padlen) {
    LOG(ERROR) << "Insufficient room for padding in ciphertext buffer";
    return false;
}
```

framework/common/encryption.cpp
tomereli: Changed to 
```
int padlen = plen % 16;
if (clen < plen + (padlen ? padlen : 16)) {
    LOG(ERROR) << "Insufficient room for padding in ciphertext buffer";
    return false;
}
```
Which matches PKCS#5 - if the plaintext length is a multiple of block size, the buffer for the encrypted output should be `plen + 16`, otherwise it is `plen + paddlen`.

framework/common/encryption.cpp
tomereli: OK - see next comment with the fix.

framework/common/include/mapf/common/encryption.h
tomereli: Fixed finally.

framework/common/test/encryption_test.cpp
tomereli: I'll just use the maximum padlen which can be 16, regardless of the plaintext length.

framework/tlvf/test/tlvf_test.cpp
tomereli: This was a bug, fixed.

framework/common/test/encryption_test.cpp
tomereli: Changed, however, this should have been a (nitpick) IMO (caused me a lot of merge conflicts :-( )

framework/common/test/encryption_test.cpp
tomereli: But nevertheless no need to do this padding calculation here, it is changed to be simple +16 for max padding size :)

framework/common/test/CMakeLists.txt
tomereli: It's ugly either way, but OK I guess, I'll refactor it :)

framework/common/include/mapf/common/utils.h
tomereli: We've been over this discussion already, we need both.

framework/common/test/encryption_test.cpp
tomereli: @CoralMalachi already done, don't worry about it :)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,54 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
-    size_t len = (config_data.getLen() + sizeof(WSC::sWscAttrKeyWrapAuthenticator) + 15) & ~0xFU;
+    // Step 1 - prepare the plaintext: [config_data | keywrapauth]:
+    // We use the config_data buffer as the plaintext buffer for encryption.
+    // The config_data buffer has room for 12 bytes for the keywrapauth (2 bytes
+    // attribute type, 2 bytes attribute length, 8 bytes data), but check it anyway
+    // to be on the safe side. Then, we add keywrapauth at its end.
+    uint8_t *plaintext = config_data.getStartBuffPtr();
+    int plaintextlen   = config_data.getLen() + sizeof(WSC::sWscAttrKeyWrapAuthenticator);
+    if (config_data.getBuffRemainingBytes() < sizeof(WSC::sWscAttrKeyWrapAuthenticator)) {
+        LOG(ERROR) << "No room for 8 bytes keywrap auth in config_data";
```
arnout: 12 bytes. But it's not really needed to specify here.

(nitpick)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,54 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
         return false;
+    }
+    LOG(DEBUG) << "ciphertext: " << std::endl << utils::dump_buffer(ciphertext, cipherlen);
 
-    std::copy_n(reinterpret_cast<uint8_t *>(&keywrapauth), sizeof(keywrapauth),
-                &buf[config_data.getLen()]);
-    uint8_t *iv = reinterpret_cast<uint8_t *>(m2->encrypted_settings()->iv());
-
-    if (!mapf::encryption::create_iv(iv, WSC::WSC_ENCRYPTED_SETTINGS_IV_LENGTH))
+    // Step 3 - Allocate encrypted settings with with the size of the output ciphertext,
```
arnout: with with -> with (nitpick)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,54 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
+    // and copy the ciphertext to it.
+    if (!encrypted_settings->alloc_encrypted_settings(cipherlen))
         return false;
-
-    if (!mapf::encryption::aes_encrypt(keywrapkey, iv, buf, len)) {
-        LOG(DEBUG) << "aes encrypt";
+    if (!m2->add_encrypted_settings(encrypted_settings))
         return false;
-    }
+    std::copy_n(ciphertext, cipherlen, encrypted_settings->encrypted_settings());
```
arnout: It doesn't matter much, but it feels more natural to me to do the add_encrypted_settings after the copy. I see the add as "closing" the struct. (nitpick)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -443,40 +443,54 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(
...
+    uint8_t *plaintext = config_data.getStartBuffPtr();
+    int plaintextlen   = config_data.getLen() + sizeof(WSC::sWscAttrKeyWrapAuthenticator);
+    if (config_data.getBuffRemainingBytes() < sizeof(WSC::sWscAttrKeyWrapAuthenticator)) {
+        LOG(ERROR) << "No room for 8 bytes keywrap auth in config_data";
+        return false;
+    }
+    WSC::sWscAttrKeyWrapAuthenticator *keywrapauth =
+        reinterpret_cast<WSC::sWscAttrKeyWrapAuthenticator *>(&plaintext[config_data.getLen()]);
+    keywrapauth->struct_init();
+    keywrapauth->struct_swap();
```
arnout: It doesn't matter much, but it feels more natural to me to do the swap after writing kwa. kwa itself isn't swapped so it makes not difference, but it's more logical to do swapping at the end. (nitpick)

framework/common/encryption.cpp
```diff
@@ -223,33 +223,61 @@ bool hmac::digest(uint8_t *digest)
...
+                 uint8_t *ciphertext, int &clen)
 {
     EVP_CIPHER_CTX *ctx;
+    int len;
 
-    int clen, len;
-    uint8_t buf[128];
+    /* Verify that the ciphertext buffer has enough storage room
+     * for block size alignment padding which will be added
+     * during encryption, which is up to plen + cipher_block_size -1.
```
arnout: It's actually up to plen + cipher_block_size - 1 for the Update, and another cipher_block_size for the final one. The calculation below is correct though. (nitpick)

framework/common/encryption.cpp
```diff
@@ -223,33 +223,61 @@ bool hmac::digest(uint8_t *digest)
...
+     *
```
arnout: Spurious empty line (nitpick).

framework/common/encryption.cpp
```diff
@@ -223,33 +223,61 @@ bool hmac::digest(uint8_t *digest)
...
+     * the block alignment of the encrypted data: as a result the amount of
+     * data written may be anything from zero bytes to
+     * (inl + cipher_block_size - 1) so out should contain sufficient room.
+     */
+    if (EVP_EncryptUpdate(ctx, ciphertext, &len, plaintext, plen) != 1) {
+        MAPF_ERR("EVP_EncryptUpdate Failed");
+        EVP_CIPHER_CTX_free(ctx);
         return false;
     }
+    clen = len;
```
arnout: This is redundant: `outl` is a pure output parameter of `EVP_EncryptFinal_ex`, so clen will just be overwritten by it. If it were an in,out parameter, then we should do `clen -= len` here to pass in the remaining length of the ciphertext buffer. (nitpick)

framework/common/include/mapf/common/encryption.h
```diff
@@ -112,24 +112,49 @@ bool create_iv(uint8_t *iv, unsigned iv_length);
...
+ * Key and IV should be the same as block size, which 16 bytes
+ * since we are using 128 bit AES (i.e. a 128 bit key).
+ * The IV size for *most* modes is the same as the block size.
+ * 
+ * @param[in] key 16 byte encryption key
+ * @param[in] iv 16 bytes random initialization vector 
+ * @param[in] ciphertext bytestream (data to deccrypt)
+ * @param[in] clen length of ciphertext buffer
+ * @param[out] plaintext output buffer of encrypted data (can be the same as ciphertext buffer)
+ * @param[in,out] plen in - length of plaintext buffer (>= clen + 16) out - length of output plaintext
```
arnout, line 157: These documentation changes should have been in the commit that also changes the implementation. (nitpick)

framework/tlvf/test/tlvf_test.cpp
arnout: It's not fixed... cipherlen = datalen + padlen = datalen + datalen - ((datalen + 15) % 16) = datalen + (datalen rounded down to block size)

The easiest would be `cipherlen = datalen + 16` and remove padlen. We don't really care about the exact size here - for that, we would have to use the same calculation as in the implementation, and then what's the point of using it.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4372,7 +4378,7 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
     // substructure of encrypted_settings, it is easier to define it separately and just append to
     // config_data.
     WSC::sWscAttrKeyWrapAuthenticator *keywrapauth =
-        reinterpret_cast<WSC::sWscAttrKeyWrapAuthenticator *>(&buf[len]);
+        reinterpret_cast<WSC::sWscAttrKeyWrapAuthenticator *>(&decrypted[len]);
```
arnout, line 4385: Before doing this, we should add a check that `len + sizeof(WSC::sWscAttrKeyWrapAuthenticator) <= datalen`. This is a security issue, an attacker could construct an M2 that overflows the buffer.

cConfigData makes these checks internally while parsing, but we bypass the TLVF parsing here so we have to add the check manually.

framework/common/encryption.cpp
```diff
@@ -281,33 +281,59 @@ bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int
...
     }
-    EVP_CIPHER_CTX_set_padding(ctx, 0);
 
-    plen = data_len;
-    if (EVP_DecryptUpdate(ctx, data, &plen, data, data_len) != 1 || plen != (int)data_len) {
+    /*
+     * Provide the message to be decrypted, and obtain the decrypted output.
+     * The decrypted data buffer out passed to EVP_DecryptUpdate() should have
+     * sufficient room for (inl + cipher_block_size) bytes.
+     * In this case - plen has to be >= plaintext_len + 16)
```
arnout: This comment was already made above, and the check is also no longer here, so it's useless at this point.

The first sentence is also not saying much, the function name EVP_Decrypt is saying the same.

(nitpick)

framework/common/encryption.cpp
```diff
@@ -281,33 +281,59 @@ bool aes_encrypt(const uint8_t *key, const uint8_t *iv, uint8_t *plaintext, int
...
+     */
+    if (EVP_DecryptUpdate(ctx, plaintext, &len, ciphertext, clen) != 1) {
+        MAPF_ERR("EVP_DecryptUpdate Failed");
+        EVP_CIPHER_CTX_free(ctx);
         return false;
     }
+    plen = len;
```
arnout, line 316: Note that unlike the encrypt function, this statement is *not* useless, because below we pass len as the argument instead of plen.

framework/common/test/encryption_test.cpp
```diff
@@ -60,8 +61,15 @@ int main()
     check(errors, std::equal(keywrapkey1, keywrapkey1 + sizeof(keywrapkey1), keywrapkey2),
           "keywrapkeys should be equal");
 
-    {
-        uint8_t plaintext[50];
+    /* size 16 and 816 yeild a multiple of block size.
```
arnout: yield (nitpick)

framework/common/test/CMakeLists.txt
```diff
@@ -1,16 +1,18 @@
...
+		file(GLOB tests *_test.cpp)
+		foreach(test ${tests})
+			get_filename_component(target ${test} NAME_WE)
+			add_executable(${target} ${test})
+			target_link_libraries(${target} common dummy_messages mapf::elpp)
+			install(TARGETS ${target} DESTINATION bin/tests/common)
+			add_test(NAME ${target} COMMAND $<TARGET_FILE:${target}>)
+		endforeach(test ${tests})
+	endif()
+endif()
```
arnout: Missing newline at end of file. (nitpick)

arnout (approved):
Bunch of nitpicks so approving. There's one potential security issue (though minor) but that can be fixed later.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: wow nice catch.

PR #486, opened 2019-11-14
tlvf: add static type assertion for addclass

Galvien:
Similarly to the `cast_class()<T>` function of `BaseClass` as modified in #481,
add a change to `addClass<T>()` to assert the types of objects the template
function is called on are derived from `BaseClass`.
Currently the *compile script* fails when `addClass()` is called on a unsupported objects that don't implement _certain methods_ being called.
The more proper way to implement a mechanism for type matching is the following change, which checks for them at compile time, and on any compilation environment.
The line
`template <class T, class = typename std::enable_if<std::is_base_of<BaseClass, T>::value>::type>` 
(courtesy of @LiorAmram) prevents the compilation immediately in case the template class doesn't inherit from `BaseClass`.

### DIFF
--- a/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/CmduMessage.h
@@ -27,8 +27,9 @@ public:
 
 public:
     std::shared_ptr<cCmduHeader> getCmduHeader() const;
-
-    template <class T> std::shared_ptr<T> addClass()
+    
+    template <class T, class = typename std::enable_if<std::is_base_of<BaseClass, T>::value>::type>
+    std::shared_ptr<T> addClass()
     {
         std::shared_ptr<T> ptr;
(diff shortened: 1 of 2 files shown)


LiorAmram (approved):

tomereli:
@LiorAmram what's the point? `addClass<int>()` will also fail in compilation since it is immediately calling `std::make_shared<int>(m_cmdu_header, m_parse, m_swap)` - `int` doesn't have such a constructor hence it will fail. I think this commit is just bragging off some cpp skills :)

LiorAmram:
@tomereli Well first I think it's good practice to make sure our templates force using the right classes.
Second, you took a simplified example. Any class with a constructor that accepts `(uint8_t*, bool, bool)` (or even `(void*, bool, bool)` will pass compilation with this template.

Either way, I can't see any harm it could cause, and it helps us stay on the safe side.

tomereli (approved):

PR #307, opened 2019-09-19
Turris Omnia Support

vitalybu:
WIP - DO NOT MERGE!

PR #488, opened 2019-11-18
Feature/agent add turris omnia rdkb support

morantr:
Port Agent changes from the Turris Omnia support branch (#307).


agent/CMakeLists.txt
```diff
@@ -7,37 +7,40 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 message(STATUS "Beerocks Build Type: ${CMAKE_BUILD_TYPE}")
 
+# Default values
+    set(BEEROCKS_WLAN1_IFACE "wlan0")
```
tomereli: Wrong indentation. Also - @vitalybu , @arnout - isn't this better if we save the interface names in the platform DB instead of here? Then we can completely remove Turris Omnia with generic OpenWRT.

agent/CMakeLists.txt
```diff
@@ -7,37 +7,40 @@ if(NOT CMAKE_BUILD_TYPE)
...
     add_definitions(-DBEEROCKS_LINUX)
     set(BEEROCKS_PLATFORM "linux")
     set(BEEROCKS_TMP_PATH "/tmp/$ENV{USER}/beerocks")
-    set(BEEROCKS_NOTIFY 0)
-    set(BEEROCKS_BRIDGE_IFACE "br-lan")
     set(BEEROCKS_BH_WIRE_IFACE "sim-eth0")
-    set(BEEROCKS_REPEATER_MODE 0)
-    set(BEEROCKS_AUTOMATIC_CREDENTIALS_UNIFY 0)
-    set(BEEROCKS_STOP_ON_FAILURE_ATTEMPTS 1)
+elseif (TARGET_PLATFORM STREQUAL "turris-omnia")
```
tomereli, line 39: Again - we shouldn't have turris-omnia target, just openwrt target.

agent/CMakeLists.txt
```diff
@@ -108,6 +111,12 @@ add_definitions(-DBEEROCKS_REVISION="${${PROJECT}_REVISION}")
 if(TARGET_PLATFORM STREQUAL "linux")
 add_definitions(-DBEEROCKS_BIN_PATH="${CMAKE_INSTALL_PREFIX}/bin/")
 add_definitions(-DBEEROCKS_CONF_PATH="${CMAKE_INSTALL_PREFIX}/config/")
+elseif(TARGET_PLATFORM STREQUAL "turris-omnia")
```
tomereli, line 114: I don't like this. We should use the same paths on all platforms. no more `/opt/beerocks` please.

common/beerocks/tlvf/CMakeLists.txt
```diff
@@ -3,7 +3,13 @@ cmake_minimum_required (VERSION 2.8.12)
 set(PROJECT_NAME btlvf)
 project(${PROJECT_NAME})
 
-find_package(PythonInterp REQUIRED)
+# TODO: FIXME: For some reason CMake detects only the Python from the cross-toolchain
```
tomereli: I suspect that this happened in the demo preparations. Now is the time for fixing this - don't leave it as TODO. Same goes for all other CMAKE hacks (but I'll comment on those separately).

tomereli: Also - What is the problem with the python from the cross-toolchain? Why can't we just use it?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -616,13 +616,15 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
             FSM_MOVE_STATE(MASTER_DISCOVERY);
         } else { // link establish
 
+#ifdef BEEROCKS_LINUX
```
tomereli: This commit doesn't explain *why* this is done and also why is it done only for Linux.
The commit message should be something like "agent: backhaul: skip removing STA interfaces from the bridge" with a description which explains why.

In addition - the *only* mode is passive mode so we should *never* do any bridge operations. This commit should remove this logic completely.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -1518,7 +1518,7 @@ bool main_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
             LOG(DEBUG) << "slave is backhaul manager, updating";
             m_pBackhaulManagerSlave = sd;
 
-#if !defined(BEEROCKS_LINUX) && !defined(BEEROCKS_TURRIS_OMNIA)
+#ifndef BEEROCKS_TURRIS_OMNIA
```
tomereli: Whole commit is wrong. Doesn't say why it disables arp monitor on Turris Omnia, and also it Re-enables arp monitor for Linux.
What was wrong with how it used to be? Looks like this commit should be completely removed.

tomereli (changes requested):

agent/CMakeLists.txt
morantr: I'll fix the indentation, but to your question about the interfaces, the answer is no.
The interfaces names are set in the configuration files and are needed before we have the ability to get them from bpl.
For example, the son_slave_main needs the interfaces from the configuration to decide which radio son_slave to bring up. I guess we can find a way to remove completely the configuration file and move everything to the database, but we need to make DR for it.

tomereli: ok.

agent/CMakeLists.txt
vitalybu: I'll change it as part of the UGW porting PR

agent/CMakeLists.txt
vitalybu: I'll change it as part of the UGW porting PR

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
vitalybu: Turris Omnia currently uses the same dummy BPL and the Linux platform, so the ARP monitor should be disabled for both. Later on we should create a Linux BPL with ARP monitor support and use it for all Linux based platforms.

common/beerocks/tlvf/CMakeLists.txt
vitalybu: The problem is with the pyYAML library dependency. Not all cross-toolchains contain this package.
As for this specific hack, it may be an issue with the older Turris tree (3.x), perhaps it doesn't exist in the latest one.
I suggest opening an issue and handling it separably from this PR.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Add a fix up that remove it

common/beerocks/tlvf/CMakeLists.txt
tomereli: Can't we quickly check if this is needed in the new Turris tree? I think @arnout or @rmelotte can help with a quick check.
Of the answer is that it is needed for now, then @morantr please open an issue on that and add the issue link in the TODO comment.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
tomereli: But now arp monitor will be enabled for Linux, is this what you meant to happen?

Does the arp monitor use the BPL? Is this why it has to be disabled when dummy BPL is used?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -581,16 +581,6 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
             FSM_MOVE_STATE(MASTER_DISCOVERY);
         } else { // link establish
 
-#ifdef BEEROCKS_LINUX
```
tomereli: This is fine but don't forget to fix the commit message once you rebase

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
vitalybu: No, enabling the ARP monitor for Linux is a bug.
It should be disabled for both Linux and Turris Omnia platforms.

tomereli: Still not fixed.

tomereli (approved):
This is a critical path change so I'm already approving - but please fix the bug you added that enables arp monitor for linux, and address the remaining comments. After that, you can merge IMO.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
morantr: removing the commit

common/beerocks/tlvf/CMakeLists.txt
arnout: I've reverted this bit on the dev/turris-omnia-support branch and it built fine.

Anyway it doesn't make any sense that this would be needed in common but not in framework.

Therefore, please don't include this change.

agent/src/beerocks/slave/CMakeLists.txt
```diff
@@ -34,6 +35,11 @@ add_executable(${PROJECT_NAME} ${beerocks_agent_sources})
 set_target_properties(${PROJECT_NAME} PROPERTIES LINK_FLAGS "-Wl,-z,defs")
 target_link_libraries(${PROJECT_NAME} rt dl beerocks::bcl beerocks::btl tlvf elpp beerocks::btlvf ${LIBS})
 
+# For some reason, this isn't found automatically from btl...
```
arnout, line 38: I'll try to fix this once top-level CMake is merged.

arnout (changes requested):
Reviewed only the first 3 commits for now. I'm already submitting this review so the find_package(PythonInterp) change can be undone. Feel free to rebase and force-push after fixing.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -589,18 +589,19 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
 
         LOG(TRACE) << "backhaul manager state=ENABLED";
 
-        // Connect to the platform manager
+        // Connect/Reconnect to the platform manager
         if (!m_scPlatform) {
-            m_scPlatform.reset();
-        }
-
         m_scPlatform = std::make_shared<SocketClient>(
```
arnout: Indentation is wrong here after commit "agent: Add to backhaul manager pointer validation and prints" (but fixed later). (nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -589,18 +589,19 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
...
             SocketClient(beerocks_temp_path + std::string(BEEROCKS_PLAT_MGR_UDS)));
         std::string err = m_scPlatform->getError();
         if (!err.empty()) {
             m_scPlatform.reset();
             LOG(ERROR) << "Failed connecting to Platform Manager: " << err;
         }
+        } else {
+            LOG(DEBUG) << "Using existing platform_manager_socket=0x"
+                       << intptr_t(m_scPlatform.get());
```
arnout, line 577: AFAIU this will print decimal, not hex (intptr_t is an integer type, not a pointer type, so default representation is decimal). (nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -690,18 +690,15 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
...
         bool sent_for_2g = false;
         bool sent_for_5g = false;
-        /*
-             * TODO
-             * this is a workaround, need to find a better way to know each slave's band
-             */
+
         for (auto soc : slaves_sockets) {
-            if (soc->hostap_iface == "wlan0") {
+            if (soc->freq_type == eFreqType::FREQ_24G) {
```
arnout, line 661: This code is still incorrect for a 3-radio system (2.4G, 5GL, 5GH). But that can be fixed later.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1010,6 +1023,14 @@ bool main_thread::backhaul_fsm_wireless(bool &skip_select)
             }
         }
 
+        if (connected) {
+            FSM_MOVE_STATE(MASTER_DISCOVERY);
+            state_attempts = 0; // for next state
+        } else {
+            FSM_MOVE_STATE(INITIATE_SCAN);
```
arnout, line 994: I actually don't think prplMesh should manage the wireless backhaul link at all (where "manage" means: configure it; it should still monitor the backhaul link, and trigger roaming if the controller requests it). The platform should already have brought the backhaul link up before prplMesh starts (though of course it's possible that the link is not up yet, therefore monitoring is needed). And wpa_supplicant is perfectly capable of selecting a BSS all by itself.

The only reasons for prplMesh to be involved is for channel selection, and for roaming between bands which would require a switch of radio. For channel selection, I think it's up to the controller to make such a decision - it can take into account the channel the backhaul station will force the radio to after steering to a different BSS.

Roaming between radios is more tricky, there is unfortunately nothing in EasyMesh that stops a controller from doing that. It's a tricky thing because you generally don't want to bring down the old STA before the new one is up and running. So I wouldn't try to deal with that for now. And anyway, I think it should be handled by having the backhaul STA config set up for both radios, but leaving one of them disabled.

Note that in the latter context, the "backhaul preferred radio band feature" *is* relevant and should be kept. So +1 for that change.

Anyway, all of this is not for this PR, just an observation.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1140,63 +1140,57 @@ bool main_thread::backhaul_fsm_wireless(bool &skip_select)
         if (roam_flag) {
             selected_bssid         = roam_selected_bssid;
             selected_bssid_channel = roam_selected_bssid_channel;
-        }
-
-        if (wifi_reconnect_flag) {
             if (!active_hal->roam(selected_bssid, selected_bssid_channel)) {
```
arnout, line 1107: Before the "Modify backhaul manager WIRELESS_WAIT_FOR_RECONNECT state", no roaming was actually done when roam_flag was set (or rather, it was only done after a disconnect event). Now, roaming is done immediately. I assume that that is intentional - but it's not so clear from the commit message. Could you double check that this roam call should be done when roam_flag is set, and should not be removed instead?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1535,6 +1539,15 @@ bool main_thread::handle_slave_backhaul_message(std::shared_ptr<SSlaveSockets> s
...
                    << " on channel=" << int(roam_selected_bssid_channel);
+
+        /*
+        * NOTE: Why moving to RESTART state? It causing BACKHAUL DISCONNECTED to the son slaves,
+        * which in the past (when we had TCP socket to the controller instead of the bus)
+        * disconnected the socket, but now when we have bus it is not happening anyway. 
+        * Another thing it is causing is to re-scan the for networks. Is it really necessary?
+        * Can we just move immediately to WIRELESS_ASSOCIATE_4ADDR, and then when successfully
+        * connected, notify the agent about the ne bssid?
+        */
```
arnout, line 1469: Great that you added this explanation!

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1517,7 +1518,7 @@ bool main_thread::handle_slave_backhaul_message(std::shared_ptr<SSlaveSockets> s
                     m_sConfig.wire_iface_type = (beerocks::eIfaceType)request->wire_iface_type();
                     m_sConfig.wired_backhaul  = request->wired_backhaul();
                     LOG(DEBUG) << "All slaves ready, proceeding" << std::endl
-                               << "SSID: " << m_sConfig.ssid << ", Pass: " << m_sConfig.pass
+                               << "SSID: " << m_sConfig.ssid << ", Pass: ****"
```
arnout, line 1439: I'm not sure why to use *** here. The credentials will be readable in the hostapd config file anyway.

If you're removing the password, just remove it.

(nitpick).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1922,9 +1951,11 @@ bool main_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event_pt
...
             roam_flag      = false;
             state_attempts = 0;
-            if (FSM_IS_IN_STATE(WIRELESS_ASSOCIATE_4ADDR_WAIT)) {
-                FSM_MOVE_STATE(WIRELESS_BRIDGE_DHCP);
-            }
+
+            // Send slaves to enable the AP's
+            send_slaves_enable();
+
+            FSM_MOVE_STATE(MASTER_DISCOVERY);
```
arnout, line 1871: This change is done as part of "Add AP enable flow" but I don't see how it's related. I think it should be part of "Remove DHCP flows".

BTW, why is the condition on WIRELESS_ASSOICATE_4ADDR_WAIT removed? I think this condition is there to handle the cases where the state has changed in the loop above. If you now unconditionally change the state, those state changes above are useless. Oh, I see this is fixed in the commit "agent: Fix bug which caused to move to the wrong backhaul state" - so perhaps that should be squashed into this one (or rather, in "Remove DHCP flows" which should contain this change as well).

(nitpick)

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -441,6 +441,70 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+            response->success() = false;
+            message_com::send_cmdu(slave_socket, cmdu_tx);
+            break;
+        }
+
+        // Enable the radio interface to apply the new configuration.
+        if (!ap_wlan_hal->enable()) {
+            LOG(ERROR) << "Failed enable";
+            response->success() = false;
+        }
```
arnout, line 501: As far as I can see, the only thing that we do here is to set the channel, and to enable the AP if it wasn't enabled yet. I think the disable, if it is needed, should be handled by bwl itself instead of by ap_manager, because a good implementation should be able to change the channel without disabling the AP (and it should also be able to tell clients to roam to the new channel).

And of course, I think all of this shouldn't be needed because we shouldn't manage the backhaul from prplMesh, but that's a different issue.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1500,6 +1500,7 @@ bool main_thread::handle_slave_backhaul_message(std::shared_ptr<SSlaveSockets> s
                     m_sConfig.ssid.assign(request->ssid(message::WIFI_SSID_MAX_LENGTH));
                     m_sConfig.pass.assign(request->pass(message::WIFI_PASS_MAX_LENGTH));
                     m_sConfig.security_type = static_cast<bwl::WiFiSec>(request->security_type());
+                    m_sConfig.mem_only_psk  = request->mem_only_psk();
```
arnout, line 1421: This mem_only_psk change belongs to a different commit (nitpick).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1833,11 +1833,9 @@ bool main_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event_pt
                             !son::wireless_utils::is_low_subband(radio_info.channel)) {
                             LOG(DEBUG) << "iface " << iface
                                        << " is connected on low 5G band with filter, aborting";
-                            if (FSM_IS_IN_STATE(WIRELESS_ASSOCIATE_4ADDR_WAIT)) {
                                 FSM_MOVE_STATE(WIRELESS_CONFIG_4ADDR_MODE);
```
arnout: Indentation is wrong in commit "agent: Fix bug which caused to move to the wrong backhaul state", fixed in a later commit. Same a few lines below. (nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3249,23 +3249,25 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
...
             LOG(ERROR) << "backhaul_manager_socket: " << err;
+                backhaul_manager_stop();
             platform_notify_error(BPL_ERR_SLAVE_CONNECTING_TO_BACKHAUL_MANAGER,
                                   "iface=" + config.backhaul_wireless_iface);
             stop_on_failure_attempts--;
             slave_reset();
+                break;
         } else {
             add_socket(backhaul_manager_socket);
+            }
```
arnout: Indentation is wrong in "agent: Backhaul manager socket reuse if exist" commit, fixed later. (nitpick)

arnout (approved):
Lots of comments but mostly nitpicks or bigger issues.

I want the find_package(PythonInterp) change to be removed before merging, but otherwise good to go. The nitpicks are mostly about wrong commit splitting so not so important.

We should have a discussion about how to handle wireless backhaul. I made some notes in my review comments here, but it warrants a bigger discussion.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: No, it should be done as part of this commit, because we must go to discovery when enabling the APs.
The WIRELESS_ASSOICATE_4ADDR_WAIT condition was removed because we already check it before (see line 1905).

PR #489, opened 2019-11-19
tlvf: make tlvVarlist a valid TLV

tomereli:
tlvVarList cannot be used in automatic parsing since it is not a valid
TLV - it includes a 2 bytes type instead of 1 byte, and the type value
is 1 which collides with other real TLV - TLV_AL_MAC_ADDRESS_TYPE.

This commit changes the type to be uint8_t and the value to be 255,
which does not and will not collide with any other real TLV, so it can
be used in automatic parsing.

We will use this later on to test various automatic parse
functionalities.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -146,7 +146,7 @@
             "request": "launch",
             "program": "${workspaceFolder}/../build/install/bin/tests/tlvf_test",
             "args": [],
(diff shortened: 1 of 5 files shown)


tomereli:
@arnout , @rmelotte can we assign someone to debug these CI failures?

arnout (approved):

PR #491, opened 2019-11-20
Fix build failure with GCC 4.9

kcego:
Namespace and template specialization problem with GCC 4.9

### DIFF
--- a/common/beerocks/bcl/include/beerocks/bcl/network/net_struct.h
+++ b/common/beerocks/bcl/include/beerocks/bcl/network/net_struct.h
@@ -22,7 +22,8 @@ inline bool operator==(sMacAddr const &lhs, sMacAddr const &rhs)
 
 inline bool operator!=(sMacAddr const &lhs, sMacAddr const &rhs) { return !(rhs == lhs); }
 
-template <> struct std::hash<sMacAddr> {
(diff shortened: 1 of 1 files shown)


common/beerocks/bcl/include/beerocks/bcl/network/net_struct.h
```diff
@@ -22,17 +22,19 @@ inline bool operator==(sMacAddr const &lhs, sMacAddr const &rhs)
...
         uint64_t value_to_hash = 0;
         for (size_t byte = 0; byte < sizeof(m.oct); byte++) {
             value_to_hash <<= 8;
             value_to_hash += m.oct[byte];
         }
-        return std::hash<std::uint64_t>()(value_to_hash);
+        return hash<std::uint64_t>()(value_to_hash);
     }
 };
+}
```
arnout: There should be a comment here that marks the end of the namespace. `clang-format.sh` adds it.

I'll rebase on master, add that comment, and push it.

arnout (approved):
Small issue which makes CI fail (see https://travis-ci.com/prplfoundation/prplMesh/builds/137366312) which I'll fix. Otherwise, LGTM.

Thank you for this patch!

PR #492, opened 2019-11-20
tlvf: add support for optional MultiAP-Ext in M1

### LINKED ISSUE #213: [TASK] Pass flow 5.8.1 test in testbed
## Description

Perform all steps necessary to pass the test associated with flow 5.8.1

## Exit criteria

### LINKED ISSUE #490: [BUG] tlvf: Multi-AP-Extension in M1 is not supported
Some devices, such as the Mediatek agent of the testbed, add the Multi-AP-Extension attribute to the vendor extension.
The current parsing of the M1 expects a vendor extension containing only the version2 attribute. Because of this, after parsing the Mediatek M1, the first byte of the version2 attribute (which is 0x00) is used as the next TLV type, and the AP radio basic capabilities aren't parsed at all.

To solve this issue, we need to allow an optional Multi-AP Extension as part of the vendor extension.

Unfortunately as described in #471, optional attributes are not currently supported (unless using  unknown length lists).

This issue is preventing the 5.7.1 and 5.8.1 tests from passing, since they require the Mediatek agent.

tomereli:
This is a first attempt at adding support for an optional
MultiAP-Extension attribute in the vendor extension of M1/M2.

First, sWscAttrVersion2 is renamed to sWscVExtAttrVersion2. It will
not be used as part of the M1 anymore, but it may still be used
somewhere else, like in the M2.

Second, a new sWscAttrVersion2 attribute is added. It contains only
the Version2 subitems, and not the whole vendor extension.

A new class is also added: cWscVendorExt. It contains both the vendor
extension fields, and unknown length list vs_data, which may contain
the version2 and/or MultiAP-Extension items.

son_slave_thread.cpp also needs to be modified. Now that the vendor
extension uses an unknown length list, it no longer contains the
default value for the Version2 attribute, so they have to be added
manually.

Fixes #490
Fixes #213

### DIFF
--- a/framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
+++ b/framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
@@ -265,6 +265,18 @@ sWscAttrVersion2:
     _type: uint8_t
     _value: WSC_VERSION2
 
+sWscIeVersion2:
+  _type: struct
+  subelement_id:
+    _type: uint8_t
+    _value: 0x0
+  subelement_length:
+    _type: uint8_t
+    _value: 0x1
+  subelement_value:
+    _type: uint8_t
+    _value: WSC_VERSION2
+
 sWscAttrVendorExtMultiAp:
   _type: struct
   attribute_type:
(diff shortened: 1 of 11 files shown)


framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -265,6 +265,18 @@ sWscAttrVersion2:
     _type: uint8_t
     _value: WSC_VERSION2
 
+sWscAttrVersion2:
```
arnout: It's not an attribute, it's a vendor extension IE. So `sWscIeVersion2` would be better. (nitpick)

With that, it is also no longer needed to change the name of sWscAttrVersion2.

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -395,3 +407,25 @@ cWscAttrEncryptedSettings:
   encrypted_settings:
     _type: char
     _length: []
+
+cWscVendorExt:
```
arnout: Since the WFA ID is hardcoded here, it would be better to name it `cWscVendorExtWfa`.

arnout (approved):
Nitpick comment, and second commit needs an actual commit message, otherwise LGTM.

PR #495, opened 2019-11-22
A series of small improvements to the cmake infrastructure

arnout:
Nothing essential, just cleanup.


arnout:
I will push more commits with more improvements, but it can already be merged as is as well.

As soon as someone approves it, I'll stop pushing additional commits and instead create a new PR for more improvements.

arnout:
Darn, apparently this approach doesn't work with older CMake :-( Back to the drawing board!

PR #496, opened 2019-11-22
Remove maptools reference from README

arnout:
While checking that it really works when building with normal cmake, I noticed that some libraries are installed in the wrong place and are not found by the
dynamic linker, so fix that as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ To build prplMesh, you need (on Ubuntu) the following packages:
 ```bash
 sudo apt install curl gcc cmake binutils git autoconf autogen libtool pkg-config \
      libreadline-dev libncurses-dev libssl-dev libjson-c-dev libnl-genl-3-dev libzmq3-dev \
-     python python-yaml python-paramiko repo bridge-utils clang-format
+     python python-yaml python-paramiko repo bridge-utils clang-format ninja
 ```
 
 If you haven't done so already, set up your git configuration:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -43,30 +43,32 @@ repo forall -p -c 'git checkout $REPO_RREV'
...
 
-Each component can be built with CMAKE, or use the [tools/maptools.py](tools/README.md) build command.
+Use standard CMake to build prplMesh, with a configure-build-install cycle.
 
-To build prplMesh in debug mode (for being able to debug with gdb), with all features and tests, run
+To build prplMesh natively in debug mode (for being able to debug with gdb), with all features and tests, and installed in a local directory, run
 
 ```bash
-./prplMesh/tools/maptools.py build map -f MSGLIB=zmq BUILD_TESTS=ON CMAKE_BUILD_TYPE=Debug
+cmake -DBUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=$(dirname $PWD)/install .
```
tomereli: It is recommended to build CMAKE out of tree, and it is better to align to the way we built so far with `maptools`, so I suggest to change this to 

```
cmake -DBUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=./install -H. -B./build -G Ninja
```
We can add a comment that `make` can be used instead of `Ninja`.

README.md
```diff
@@ -43,30 +43,32 @@ repo forall -p -c 'git checkout $REPO_RREV'
...
+make
```
tomereli: `cd build && ninja install` or `cd build && make -j`

tomereli (approved):
Mainly nitpicks, so approving.

README.md
arnout: For consistency with scripts and other README files referring to the build directory, I've used `../build` and `../build/install`. Though actually, I think the only places where it is hardcoded is in maptools and in the scripts that call docker. But still, it's good to keep the same location.

PR #498, opened 2019-11-24
Bugfix/Ap Metrics TLV with single unknown length list

### LINKED ISSUE #473: [TASK] revert w/a for TLV_AP_METRIC optional parameters TLV_AP_METRIC
## Description
Need to define TLV_AP_METRIC parameters as optional after tlvf will support optional parameters (#471)

## Exit criteria
All relevant 4 parameters dynamically defined in  TLV_AP_METRIC.yaml and pass compilation.
Wifi certification  test 5.7.1 is passing in test bed.

## Detailed description
As part of controller certification flows TLV_AP_METRIC needed to be saved to db.
TLV_AP_METRIC has to be defined with the following 4 optional parameters.
 estimated_service_info_field_ac_be:
 estimated_service_info_field_ac_bk:
 estimated_service_info_field_ac_vo:
 estimated_service_info_field_ac_vi:

Since Tlvf doesn’t support optional parameters at that time  a temp solution was applied to pass certification.
W/A: define max size instead of dynamic.
Todo:
	1. Need to revert w/a and   define parameters as  dynamic.
	2. Need to make appropriate changes so during  addClass<wfa_map::tlvApMetric>
    these optional parameters will be added correctly according to  estimated_service_parameters status.

For further info See tlvApMetric definetion in Multi-AP_Specification_v1.0 spec

CoralMalachi:
According to specification (17.2.2 AP Metrics TLV format), Ap Metrics TLV
may contain the Estimated Service Parameters Information field for AC=BK,
AC=VO, AC=VI (always contains a field for AC=BE).

Since there is no guarantee that all fields will always be sent by an AP,
it's wrong to initialize the Ap Metric TLV to contain all of them from
the beginning.

The TLVF cannot accommodate more than one unknown length list OR an
unknown length list which is not the last member in the TLV.
Therefore, replace all the Estimated Service Parameters Information
fields with a single unknown length list.

Fixes #473
Verified on the testbed.

### DIFF
--- a/controller/src/beerocks/master/db/node.cpp
+++ b/controller/src/beerocks/master/db/node.cpp
@@ -350,28 +350,19 @@ bool node::ap_metrics_data::add_ap_metric_data(std::shared_ptr<wfa_map::tlvApMet
     channel_utilization                 = ApMetricData->channel_utilization();
     number_of_stas_currently_associated = ApMetricData->number_of_stas_currently_associated();
 
-    if (ApMetricData->estimated_service_parameters().include_ac_be) {
-        std::copy(ApMetricData->estimated_service_info_field_ac_be(),
-                  ApMetricData->estimated_service_info_field_ac_be() + 3,
-                  estimated_service_info_field_ac_be);
-    }
-
+    //copy all fields to database vector
+    std::copy(ApMetricData->estimated_service_info_field(),
+              ApMetricData->estimated_service_info_field() +
+                  ApMetricData->estimated_service_info_field_length(),
+              std::back_inserter(estimated_service_info_fields));
     if (ApMetricData->estimated_service_parameters().include_ac_bk) {
-        std::copy(ApMetricData->estimated_service_info_field_ac_bk(),
-                  ApMetricData->estimated_service_info_field_ac_bk() + 3,
-                  estimated_service_info_field_ac_bk);
+        include_ac_bk = true;
     }
-
     if (ApMetricData->estimated_service_parameters().include_ac_vo) {
-        std::copy(ApMetricData->estimated_service_info_field_ac_vo(),
-                  ApMetricData->estimated_service_info_field_ac_vo() + 3,
-                  estimated_service_info_field_ac_vo);
+        include_ac_vo = true;
     }
-
     if (ApMetricData->estimated_service_parameters().include_ac_vi) {
-        std::copy(ApMetricData->estimated_service_info_field_ac_vi(),
-                  ApMetricData->estimated_service_info_field_ac_vi() + 3,
-                  estimated_service_info_field_ac_vi);
+        include_ac_vi = true;
     }
     return true;
 }
--- a/controller/src/beerocks/master/db/node.h
+++ b/controller/src/beerocks/master/db/node.h
@@ -212,10 +212,10 @@ public:
         sMacAddr bssid;
         uint8_t channel_utilization;
         uint16_t number_of_stas_currently_associated;
-        uint8_t estimated_service_info_field_ac_be[3];
-        uint8_t estimated_service_info_field_ac_bk[3];
-        uint8_t estimated_service_info_field_ac_vo[3];
-        uint8_t estimated_service_info_field_ac_vi[3];
+        std::vector<uint8_t> estimated_service_info_fields;
+        bool include_ac_vo = false;
+        bool include_ac_bk = false;
+        bool include_ac_vi = false;
 
         bool add_ap_metric_data(std::shared_ptr<wfa_map::tlvApMetric> ApMetricData);
     };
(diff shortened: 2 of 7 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1408,15 +1408,24 @@ bool master_thread::construct_combined_infra_metric()
...
-                metric_data_per_agent.estimated_service_info_field_ac_vo[i];
-            *ap_metrics_tlv->estimated_service_info_field_ac_vi(i) =
-                metric_data_per_agent.estimated_service_info_field_ac_vi[i];
+        auto len = metric_data_per_agent.estimated_service_info_fields.size();
+        if (!ap_metrics_tlv->alloc_estimated_service_info_field(len)) {
+            LOG(ERROR) << "alloc_estimated_service_info_field() has failed!";
+            return false;
+        }
+        size_t info_idx = 0;
+        for (auto &info : metric_data_per_agent.estimated_service_info_fields) {
```
arnout: Can't use use std::copy here?

```
std::copy(metric_data_per_agent.estimated_service_info_fields.begin(),
                metric_data_per_agent.estimated_service_info_fields.end(),
                 ap_metrics_tlv->estimated_service_info_field());
```

controller/src/beerocks/master/db/node.cpp
```diff
@@ -350,28 +350,19 @@ bool node::ap_metrics_data::add_ap_metric_data(std::shared_ptr<wfa_map::tlvApMet
...
     number_of_stas_currently_associated = ApMetricData->number_of_stas_currently_associated();
 
-    if (ApMetricData->estimated_service_parameters().include_ac_be) {
-        std::copy(ApMetricData->estimated_service_info_field_ac_be(),
-                  ApMetricData->estimated_service_info_field_ac_be() + 3,
-                  estimated_service_info_field_ac_be);
+    auto len = ApMetricData->estimated_service_info_field_length();
+    //copy all fields to database vector
+    for (unsigned int i = 0; i < len; ++i) {
+        estimated_service_info_fields.push_back(*ApMetricData->estimated_service_info_field(i));
```
arnout: The normal way to do this would be

```
std::copy(ApMetricData->estimated_service_info_field(),
                ApMetricData->estimated_service_info_field() + ApMetricData->estimated_service_info_field_length(),
                std::back_inserter(estimated_service_info_fields));
```

controller/src/beerocks/master/db/node.h
```diff
@@ -212,10 +212,10 @@ class node {
         sMacAddr bssid;
         uint8_t channel_utilization;
         uint16_t number_of_stas_currently_associated;
-        uint8_t estimated_service_info_field_ac_be[3];
-        uint8_t estimated_service_info_field_ac_bk[3];
-        uint8_t estimated_service_info_field_ac_vo[3];
-        uint8_t estimated_service_info_field_ac_vi[3];
+        std::vector<uint8_t> estimated_service_info_fields;
```
arnout, line 215: I'm not entirely happy with this. If we were to actually use this information for any purpose, it would make more sense to have something like this:

```
struct estimated_service_info {
    bool is_valid;
    uint8_t data_format;
    uint8_t ba_window_size;
    ...
};
estimated_service_info estimated_service_info_be;
estimated_service_info estimated_service_info_vo;
estimated_service_info estimated_service_info_bk;
estimated_service_info estimated_service_info_vi;
```

However, that is a *lot* more work for no real benefit (at this time), so OK.

That said, I think the original was better: 4 explicit lists. The booleans can be replaced with a comparison with `{0, 0, 0}`. To make that comparison easy, we can use `std::array<uint8_t, 3>` instead of a C array. But this option is also a nitpick so only change it if you really think it's better.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1202,21 +1203,23 @@ print_ap_metric_map(std::unordered_map<sMacAddr, son::node::ap_metrics_data> &ap
...
+                   << int(pair_agent.second.estimated_service_info_fields[0]) << std::hex
+                   << int(pair_agent.second.estimated_service_info_fields[1]) << std::hex
+                   << int(pair_agent.second.estimated_service_info_fields[2]) << std::endl;
+        if (len > 3) {
+            int lists_count = (len - 3) / 3;
+            int i           = 3;
+            for (int k = 0; k < lists_count; k++) {
+
+                LOG(DEBUG) << "  estimated_service_info_field_ac_** = 0x" << std::hex
+                           << int(pair_agent.second.estimated_service_info_fields[i]) << std::hex
```
arnout: A single `std::hex` is enough, I think.

arnout (approved):
Nitpicks as usual.

The first commit actually breaks things, and only the third commit fixes it. So I think the first three should be squashed together. But it's not that important.

tomereli (approved):

PR #501, opened 2019-11-24
Feature/rax40 support

tomereli:


agent/CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ if (TARGET_PLATFORM STREQUAL "openwrt")
     elseif (TARGET_PLATFORM_TYPE STREQUAL "turris-omnia")
         set(BEEROCKS_PLATFORM "turris-omnia")
         set(BEEROCKS_WLAN2_IFACE "wlan1")
+    elseif (TARGET_PLATFORM_TYPE STREQUAL "nighthawk")
+        set(BEEROCKS_PLATFORM "nighthawk")
```
arnout, line 33: This is not going in the right direction... IMO, we shouldn't even have a TARGET_PLATFORM_TYPE. Instead, we should have a variant for BPL and BWL (and possibly BML as well in the future).

What we can do, is to define platform types that give defaults to BPL and BWL and some other things like the interface names in the config file - but more as a helper, not really as something to use in "production". But such platform types should just be rdkb, openwrt, ugw and linux IMO. *Not* nighthawk and turris-omnia.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -1518,25 +1518,25 @@ bool main_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
...
-#if !defined(BEEROCKS_LINUX) && !defined(BEEROCKS_TURRIS_OMNIA) && !defined(BEEROCKS_NIGHTHAWK)
-            // Start ARP monitor
-            if (enable_arp_monitor) {
-                if (!init_arp_monitor()) {
-                    LOG(ERROR) << "can't start ARP monitor";
-                    return false;
+            if (bpl_cfg_monitoring_supported()) {
+                // Start ARP monitor
+                if (enable_arp_monitor) {
+                    if (!init_arp_monitor()) {
```
arnout, line 1524: It would make more sense to me if this were implemented by just having an implementation of `init_arp_monitor` that returns true without doing anything. However, since init_arp_monitor is implemented in platform_manager instead of in bpl, that's not possible, so OK (for now).

arnout (changes requested):
First commit I don't think is going in the right direction, but the second commit is good. It can be pushed directly to master (in a new PR, obviously).

agent/CMakeLists.txt
vitalybu: Or just move all platform specific variables into the configuration files (`beerocks_agent.conf`/`beerocks_controller.conf`).
We can even parse/load the configuration file from the `prplmesh_utils.sh` script and use any platform-specific values (interfaces names etc.)

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
vitalybu: It's also possible to extend the return value of the `bpl_arp_mon_start()` function (called internally from `init_arp_monitor()`)
For example, 0 on success, -1 on error and -2 if not supported/implemented.

arnout: Yes, I like that!

tomereli: you 2 are making me work harder :)

OK - we'll go with this approach, but know that it will require some compromises - I will update bpl_err.h to include -1 and -2 for BPL_ERR_UNKNOWN (-1) and BPL_ERR_OPERATION_NOT_SUPPORTED (-2). I will not however replace all -1 to BPL_ERR_UNKNOWN in all of the BPL :)

tomereli: See #508

tomereli:
Superceeded by #508

PR #507, opened 2019-11-25
Feature/radio basic capabilities channel preferences report

### LINKED ISSUE #363: [TASK] AP Radio Basic Capabilities and Channel Preference Report for nl80211
## Description

When the controller does a Channel Preference Query, the agent should send a Channel Preference Report that corresponds to the actual behaviour of the AP. Similarly, the AP Radio Basic Capabilities TLV should contain the real operating classes and channels. Since these two go hand in hand, they're included in a single issue.

From Multi-AP specification:

> 
> 
> The Channel Preference Report message shall
> contain information regarding all the channels that a given radio of the Multi-AP Agent transmitting the report is
> temporarily unable to operate in, or prefers not to operate in, for each supported operating class as specified by the Multi-
> AP Agent (per Operating Classes field in the AP Radio Basic Capabilities TLV in section 17.2.7). Additionally, the Channel
> Preference Report message shall contain information regarding the channel separation (if any) that a given radio of the
> Multi-AP Agent requires between each of its own potential operating channels and the operating channel of another radio
> of the Multi-AP Agent, where the two radios may have one transmitting and the other receiving or transmitting
> simultaneously.
> Note: Channels for which the Multi-AP Agent’s radio is statically (permanently) unable to operate are reported in the AP
> Radio Basic Capabilities TLV. A Channel Preference Report message does not include information on operating classes
> that are not supported by the corresponding radio of the Multi-AP Agent per the AP Radio Basic Capabilities TLV.

Most likely, the statically inoperable channels list will always be empty, and the channel preference will be limited to DFS info, but this should be checked.

## Exit criteria

- Radio Basic Capabilities TLV contains actual operating classes and inoperable channels on OpenWrt, based on netlink information.
- Channel Preference Report TLV contains actual operating classes and channel preferences from netlink.
- This is validated manually on Turris Omnia.

YuriyMasechko:
When the controller does a Channel Preference Query, the agent should send a Channel Preference Report that corresponds to the actual behaviour of the AP. Similarly, the AP Radio Basic Capabilities TLV should contain the real operating classes and channels.

- Only radar affected channels are added to non-operable list for Channel Preference.

- Netlink code is based on iw utility. Not sure that part with channel bandwidth is correct.

Fixes #363


common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -16,6 +16,40 @@
...
+    beerocks::eWiFiBandwidth band;
+};
+
+//Based on hostapd global_op_class struct, file ieee802_11_common.c
+static const std::map<uint8_t, sOperatingClass> operating_classes_list = {
+    {81, {{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13}, beerocks::BANDWIDTH_20}},
+    {82, {{14}, beerocks::BANDWIDTH_20}},
+    {83, {{1, 2, 3, 4, 5, 6, 7, 8, 9}, beerocks::BANDWIDTH_40}},
+    {115, {{36, 40, 44, 48}, beerocks::BANDWIDTH_20}},
+    {116, {{36, 44}, beerocks::BANDWIDTH_40}},
```
morantr: Please wrap make this not be formatted by the clang and align the columns:

```cpp
// clang-format off
static const std::map<uint8_t, sOperatingClass> operating_classes_list = {
//  {OP Class   {Channels List,                                 Bandwidth             }}  
    {81,        {{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13},   beerocks::BANDWIDTH_20}},
    {82,        {{14},                                          beerocks::BANDWIDTH_20}},
    {83,        {{1, 2, 3, 4, 5, 6, 7, 8, 9},                   beerocks::BANDWIDTH_40}},
    {115,       {{36, 40, 44, 48},                              beerocks::BANDWIDTH_20}},
    {116,       {{36, 44},                                      beerocks::BANDWIDTH_40}},

// clang-format on
```

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -462,49 +496,13 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
-        {180, {1, 2, 3, 4, 5, 6}}};
-
-    auto it = operating_class_to_channel_set.find(operating_class);
-    if (it == operating_class_to_channel_set.end()) {
+
+    auto it = operating_classes_list.find(operating_class);
+    if (it == operating_classes_list.end()) {
         LOG(ERROR) << "reserved operating class " << int(operating_class);
-        return std::set<uint8_t>();
+        return operating_classes_list.at(0).channels;
```
morantr: The place `0` is not defined on the map so it will lead to segfault.
You should create local `static const std::set<uint8_t> dummy` on the function and return it instead.

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
```diff
@@ -93,6 +93,30 @@ typedef struct sRadioCapabilities {
...
+    sPreference(){};
+    sPreference(uint8_t oper_class_, int8_t preference_, uint8_t reason_)
+    {
+        oper_class = oper_class_;
+        preference = preference_;
+        reason     = reason_;
+    }
+    bool operator<(const sPreference &rhs) const
+    {
+        return oper_class < rhs.oper_class || preference < rhs.preference || reason < rhs.reason;
```
morantr: I think that the `<` operator should compare the preference only.

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
```diff
@@ -93,6 +93,30 @@ typedef struct sRadioCapabilities {
...
+    int8_t preference  = 0;
+    uint8_t reason     = 0;
+    sPreference(){};
+    sPreference(uint8_t oper_class_, int8_t preference_, uint8_t reason_)
+    {
+        oper_class = oper_class_;
+        preference = preference_;
+        reason     = reason_;
+    }
+    bool operator<(const sPreference &rhs) const
```
morantr: If we turn this struct to be auto-generated in the future, the operator function will be gone (There is no way to define an operator on yaml file). So this operator will have to move somewhere else (son_wireless_utils).

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
```diff
@@ -93,6 +93,30 @@ typedef struct sRadioCapabilities {
     }
 } __attribute__((packed)) sRadioCapabilities;
 
+typedef struct sPreference {
+    uint8_t oper_class = 0;
+    int8_t preference  = 0;
```
morantr: Why is it `int8_t` and not `uint8_t`? Can the preference be negative?

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -490,6 +492,89 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+void wireless_utils::get_channel_preferences(
+    beerocks::message::sWifiChannel supported_channels[],
+    std::map<beerocks::message::sPreference, std::list<beerocks::message::sWifiChannel>>
+        &preferences)
+{
+    std::list<uint8_t> oper_classes;
+    wireless_utils::get_supported_operating_classes(supported_channels, oper_classes);
+
+    for (auto oper_class : oper_classes) {
+        std::set<uint8_t> oper_class_channels = operating_classes_list.at(oper_class).channels;
```
morantr: Please make it reference so you will not make a copy of the lists of channel.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -490,6 +492,89 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+        }
+        if (!radar_affected_channels.empty()) {
+            preferences.insert(std::make_pair<beerocks::message::sPreference,
+                                              std::list<beerocks::message::sWifiChannel>>(
+                beerocks::message::sPreference(
+                    {oper_class, 0,
+                     wfa_map::cPreferenceOperatingClasses::eReasonCode::
+                         OPERATION_DISALLOWED_DUE_TO_RADAR_DETECTION_ON_A_DFS_CHANNEL}),
+                std::list<beerocks::message::sWifiChannel>(radar_affected_channels)));
+        }
```
morantr: This code inserts one element to the map container. It has too many lines of code and with the clang-format, it is difficult to read.
To make it look better you can insert element with `[]`operator instead of with`make_pair<>()`:
```cpp
preferences[beerocks::message::sPreference({oper_class, 0, 
     wfa_map::cPreferenceOperatingClasses::eReasonCode::
     OPERATION_DISALLOWED_DUE_TO_RADAR_DETECTION_ON_A_DFS_CHANNEL})] =
                std::list<beerocks::message::sWifiChannel>(radar_affected_channels);
```
This is also ugly, so I'll prefer to build `sPreference` separately, and push it. Yes it will make an unnecessary copy but it will be more readable so I prefer it:

```cpp
auto pref = beerocks::message::sPreference({oper_class, 0, 
     wfa_map::cPreferenceOperatingClasses::eReasonCode::
     OPERATION_DISALLOWED_DUE_TO_RADAR_DETECTION_ON_A_DFS_CHANNEL});

preferences[pref] = std::list<beerocks::message::sWifiChannel>(radar_affected_channels);
```

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -490,6 +492,89 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
     return channels;
 }
 
+void wireless_utils::get_channel_preferences(
+    beerocks::message::sWifiChannel supported_channels[],
+    std::map<beerocks::message::sPreference, std::list<beerocks::message::sWifiChannel>>
```
morantr: Why does `sPreference` is defined as `map` and not `unordered_map`? Does the order of the element matter?

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -490,6 +492,89 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+uint8_t wireless_utils::get_operating_class_max_tx_power(
+    beerocks::message::sWifiChannel supported_channels[], uint8_t operating_class)
+{
+    uint8_t max_tx_power = 0;
+    for (auto op_class_channel : operating_classes_list.at(operating_class).channels) {
+        for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+            if (op_class_channel == supported_channels[i].channel &&
+                max_tx_power < supported_channels[i].tx_pow) {
+                max_tx_power = supported_channels[i].tx_pow;
+            }
```
morantr:
```cpp
if (op_class_channel == supported_channels[i].channel) {
    max_tx_power = std::max(max_tx_power, supported_channels[i].tx_pow);
}
```

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -490,6 +492,89 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+            for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+                if (channel == supported_channels[i].channel &&
+                    oper_class.second.band == supported_channels[i].channel_bandwidth) {
+                    operating_classes.push_back(oper_class.first);
+                    found = 1;
+                    break;
+                }
+            }
+            if (found)
+                break;
```
morantr:
```cpp
auto supported_channels_end = supported_channels + beerocks::message::SUPPORTED_CHANNELS_LENGTH;
auto it = std::find(supported_channels, supported_channels_end, op_class_channel);
found = it != supported_channels && oper_class.second.band == it->channel_bandwidth;
if (found) {
    operating_classes.push_back(oper_class.first);
    break;
}
```

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -490,6 +492,89 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+    for (auto op_class_channel : operating_classes_list.at(operating_class).channels) {
+        uint8_t found = 0;
+        for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+            if (op_class_channel == supported_channels[i].channel) {
+                found = 1;
+                break;
+            }
+        }
+        if (!found)
+            non_oper_channels.push_back(op_class_channel);
```
morantr:
```cpp
for (auto op_class_channel : operating_classes_list.at(operating_class).channels) {
    auto supported_channels_end = supported_channels + beerocks::message::SUPPORTED_CHANNELS_LENGTH;
    found = std::find(supported_channels, supported_channels_end, op_class_channel) != supported_channels_end;
    if (!found) {
         non_oper_channels.push_back(op_class_channel);
         break;
    }
}
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -265,8 +265,10 @@ eActionOp_APMANAGER:
 
   ACTION_APMANAGER_SUPPORTED_CHANNELS_REQUEST: 72
   ACTION_APMANAGER_SUPPORTED_CHANNELS_RESPONSE: 73
+  ACTION_APMANAGER_READ_ACS_REPORT_REQUEST: 74
+  ACTION_APMANAGER_READ_ACS_REPORT_RESPONSE: 75
```
morantr: I don't understand why did you add it and create additional states for it on the son_slave, when the ap_manager sends it now on APMANAGER_JOINED_NOTIFICATION?

morantr (changes requested):

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: at(0) will return this - {0, {{}, beerocks::BANDWIDTH_MAX}}};

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
YuriyMasechko: APMANAGER_JOINED_NOTIFICATION will send it only on connect, but ACS_REPORT is used for channel preference query. Still not sure that added logic is correct. Should I use acs report for channel preference or read supported channels?
@tomereli @arnout could you please take a look at this part of code? I added logic to read acs report when getting channel preference. For netlink acs report will just call read_supported_channels, but for dwpal it calls different cmd.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
YuriyMasechko: Will it work for different oper classes with same preference?
It is used for std::map<beerocks::message::sPreference,std::list<beerocks::message::sWifiChannel>>.
So when inserting different oper classes with same preference, how will it compare them?
Even same preferences for same oper class could have different reasons.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: It is used to build preference report, I think it's better to have oper classes ordered.

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
morantr: I guess you right if you plan to insert it into `std::map`.
The question is why did you choose `std::map` and not `std::unordered_map` in the first place?
Is it important that the key will be ordered it a particular way when iterating the container?
If not, use `std::unordered_map` then, and then you could get rid of this operator.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
morantr: Sorry, I didn't notice this element.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -85,6 +92,15 @@ static ap_wlan_hal::Event nl80211_to_bwl_event(const std::string &opcode)
     return ap_wlan_hal::Event::Invalid;
 }
 
+static uint32_t ieee80211_frequency_to_channel(uint32_t freq)
```
morantr: you have this function on beerocks_utils.cpp

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -417,10 +433,132 @@ bool ap_wlan_hal_nl80211::read_acs_report()
...
 bool ap_wlan_hal_nl80211::read_supported_channels()
 {
-    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED!";
-    return true;
+    LOG(TRACE) << __func__ << " for interface: " << get_radio_info().iface_name;
+
+    // Clear the supported channels vector
+    for (auto &chan : m_radio_info.supported_channels) {
+        memset(&chan, 0, sizeof(chan));
+    }
```
morantr: This is not the cpp way to do it :)
Since supported_channels is a vector, it has a built-in `clear` method.
Please do this instead:
`m_radio_info.supported_channels.clear()`

morantr (dismissed):

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
YuriyMasechko: If it's just a case of operator, then unordered_map will require additional hash operator.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
YuriyMasechko: Done.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
YuriyMasechko: Done.

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
morantr: ok

common/beerocks/bcl/source/son/son_wireless_utils.cpp
morantr: ok

common/beerocks/bcl/include/bcl/beerocks_message_structs.h
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
morantr: It is not aligned nicely. Check that the clang does not mess it up

YuriyMasechko: Corrected.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
arnout: I don't like this dummy element in the table very much. As proven by @morantr's mistake, it's not very obvious. Also, this dummy element is not used anywhere else. So I would follow @morantr's proposal and add a static dummy variable here.

common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
```diff
@@ -135,6 +145,20 @@ class wireless_utils {
...
+    get_operating_class_max_tx_power(beerocks::message::sWifiChannel supported_channels[],
+                                     uint8_t operating_class);
+    static void
+    get_operating_class_non_oper_channels(beerocks::message::sWifiChannel supported_channels[],
+                                          uint8_t operating_class,
+                                          std::list<uint8_t> &non_oper_channels);
+    static void get_channel_preferences(
+        beerocks::message::sWifiChannel supported_channels[],
+        std::map<beerocks::message::sPreference, std::list<beerocks::message::sWifiChannel>>
+            &preferences);
```
arnout: It's not very useful to pass the return value as an out argument here (for all three get functions above). That would only be useful if we're appending to the lists. So instead, return the results. Note that thanks to the move constructor and move-assignment, this doesn't hurt performance (much).

Also, these four functions should have a doxygen comment.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -557,6 +559,82 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+                    radar_affected_channels.push_back(supported_channels[i]);
+            }
+        }
+        if (!radar_affected_channels.empty()) {
+            auto pref = beerocks::message::sPreference(
+                {oper_class, 0,
+                 wfa_map::cPreferenceOperatingClasses::eReasonCode::
+                     OPERATION_DISALLOWED_DUE_TO_RADAR_DETECTION_ON_A_DFS_CHANNEL});
+
+            preferences[pref] = std::list<beerocks::message::sWifiChannel>(radar_affected_channels);
```
arnout: I don't think the explicit cast is needed here.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -557,6 +559,82 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+{
+    //TODO handle regulatory domain operating classes
+    auto supported_channels_end = supported_channels + beerocks::message::SUPPORTED_CHANNELS_LENGTH;
+
+    for (auto oper_class : operating_classes_list) {
+        for (auto op_class_channel : oper_class.second.channels) {
+            auto it = std::find(supported_channels, supported_channels_end, op_class_channel);
+            auto found =
+                it != supported_channels_end && oper_class.second.band == it->channel_bandwidth;
+            if (found) {
```
arnout: No need for an extra variable. It is more readable if the condition is directly in the `if`.

But actually, I think it would be useful to swap the loops and factor this into an extra function:

```
static bool has_operating_class_channel(const sOperatingClass& oper_class, const 
    beerocks::message::sWifiChannel& channel)
{
    if (oper_class.band != channel.channel_bandwidth)
        return false;
    auto it = oper_class.channels.find(channel.channel);
    return it != oper_class.channels.end();
}
```

and change this function into:

```
for (auto oper_class : operating_classes_list) {
    for (auto channel : supported_channels) {
        if (has_operating_class_channel(oper_class.second, channel)) {
            operating_classes.push_back(oper_class.first);
            break;
        }
    }
}
```

(yes, I know, you can't use this kind of for loop because it's a C-style array, but I was lazy)

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -557,6 +559,82 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+
+void wireless_utils::get_supported_operating_classes(
+    beerocks::message::sWifiChannel supported_channels[], std::list<uint8_t> &operating_classes)
+{
+    //TODO handle regulatory domain operating classes
+    auto supported_channels_end = supported_channels + beerocks::message::SUPPORTED_CHANNELS_LENGTH;
+
+    for (auto oper_class : operating_classes_list) {
+        for (auto op_class_channel : oper_class.second.channels) {
+            auto it = std::find(supported_channels, supported_channels_end, op_class_channel);
```
arnout: op_class_channel is a uint8_t and supported_channels is an array of sWifiChannel... You had to add this comparison between sWifiChannel and uint8_t just to support this find operation. I think that's a bit annoying. Therefore, I think it's better to do it differently as described below.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -557,6 +559,82 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
     return channels;
 }
 
+void wireless_utils::get_channel_preferences(
+    beerocks::message::sWifiChannel supported_channels[],
+    std::map<beerocks::message::sPreference, std::list<beerocks::message::sWifiChannel>>
+        &preferences)
+{
+    std::list<uint8_t> oper_classes;
+    wireless_utils::get_supported_operating_classes(supported_channels, oper_classes);
```
arnout: There's no need to do get_supported_operating_classes first, it is implied by the matching in the loop below. The loop below can directly iterate over all operating classes.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -557,6 +559,82 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
...
+    for (auto oper_class : oper_classes) {
+        const std::set<uint8_t> &oper_class_channels =
+            operating_classes_list.at(oper_class).channels;
+        std::list<beerocks::message::sWifiChannel> radar_affected_channels;
+        for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+            if (oper_class_channels.find(supported_channels[i].channel) !=
+                oper_class_channels.end()) {
+                if (supported_channels[i].radar_affected)
+                    radar_affected_channels.push_back(supported_channels[i]);
+            }
```
arnout: I think the bandwidth must be matched as well here, no?

So then, the helper function that I describe below can be used instead:

```
if (has_operating_class_channel(oper_class_channels, supported_channels[i])) {
```

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -557,6 +559,82 @@ std::vector<uint8_t> wireless_utils::calc_5g_20MHz_subband_channels(
     return channels;
 }
 
+void wireless_utils::get_channel_preferences(
+    beerocks::message::sWifiChannel supported_channels[],
+    std::map<beerocks::message::sPreference, std::list<beerocks::message::sWifiChannel>>
+        &preferences)
+{
+    std::list<uint8_t> oper_classes;
```
arnout: For all these things, I actually think a std::vector makes more sense than a std::list. In theory std::list is more efficient for appending, but that's only due to resize operations. The cost, however, is 2 pointers overhead per element. If we presize the vector to e.g. 8 elements (which makes it about as large as a single-entry std::list), it will almost never be resized.

Anyway, that's just an optimisation so not very important.

common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
```diff
@@ -69,6 +69,11 @@ inline bool operator<(const sPreference &rhs, const sPreference &lhs)
     return rhs.oper_class < lhs.oper_class || rhs.preference < lhs.preference ||
            rhs.reason < lhs.reason;
 }
+
+inline bool operator==(const sWifiChannel &wifi_channel, const uint8_t channel)
+{
+    return wifi_channel.channel == channel;
+}
```
arnout: As noted somewhere else, this is only needed for the std::find calls, and I think we can avoid them.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -201,3 +201,12 @@ cACTION_APMANAGER_WIFI_CREDENTIALS_UPDATE_REQUEST:
...
   _type: class
 
+cACTION_APMANAGER_SUPPORTED_CHANNELS_REQUEST:
+  _type: class
+
+cACTION_APMANAGER_SUPPORTED_CHANNELS_RESPONSE:
+  _type: class
+  supported_channels_list:
+    _type: beerocks::message::sWifiChannel
+    _length: [ "beerocks::message::SUPPORTED_CHANNELS_LENGTH" ]
```
arnout: I don't like very much that this is a fixed-size list, but it's already done like that in other places so OK.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3110,6 +3132,30 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             break;
         }
```
arnout: This check should not be part of STATE_GET_SUPPORTED_CHANNELS. It is there in STATE_JOIN_MASTER because we do send_cmdu_to_controller there. In this state, we don't do anything like that, so it's not needed.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2201,6 +2201,28 @@ bool slave_thread::handle_cmdu_ap_manager_message(
...
+        auto response =
+            cmdu_rx.addClass<beerocks_message::cACTION_APMANAGER_SUPPORTED_CHANNELS_RESPONSE>();
+        if (!response) {
+            LOG(ERROR) << "addClass cACTION_APMANAGER_SUPPORTED_CHANNELS_RESPONSE failed";
+            return false;
+        }
+
+        auto tuple_supported_channels = response->supported_channels_list(0);
+        std::copy_n(&std::get<1>(tuple_supported_channels), message::SUPPORTED_CHANNELS_LENGTH,
+                    hostap_params.supported_channels);
```
arnout: I think we could avoid the additional state by including the supported channels list in ACTION_APMANAGER_JOINED_NOTIFICATION. I think that for now, we can just do the read_supported_channels in the else branch of the acs_enabled condition. In either case, the supported_channels_list will be available.

And actually, the acs_enabled condition should (eventually) be removed entirely and it should be replaced by the preference query that you added.

arnout (approved):
Although I have a lot of comments again, some of them rather big ones, they are all nitpicky.

The only comment that may be important is that the get_channel_preferences, get_operating_class_max_tx_power and get_operating_class_non_oper_channels functions should not just match the channel but also the bandwidth. But I'm not 100% certain that I'm right about that.

So, please decide for each comment if you will take it into account or not. If not, just mark it as resolved. When ready (possibly without any change at all), tell me on slack and I'll merge.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: It was taken from moran's comment to 459:
https://github.com/prplfoundation/prplMesh/pull/459#discussion_r341846613

arnout: I think @morantr could have lived with a static dummy variable as well. The main point was to have something static so you can return a const ref instead of a copy of the set.

Anyway, (nitpick)

YuriyMasechko: Done.

common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
YuriyMasechko: Done.

agent/src/beerocks/slave/son_slave_thread.cpp

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Removed and added read_supported_channels() in else.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
YuriyMasechko: Done.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
YuriyMasechko: State was removed.

arnout (approved):

PR #508, opened 2019-11-26
platform_manager: remove hardcoded platform checks for arp and dhcp monitors

tomereli:
Remove hard coded #ifdefs which doesn't start arp and dhcp monitors when
the BPL backend is Linux, where arp and dhcp monitors are not
implemented.
Instead, update the init_arp_monitor and init_dhcp_monitor functions to
return true if the BPL backend does not support monitoring.

While at it, also make dummy BWL available for non Linux platforms. This is useful in cases where initial bringup on a different OS (UGW, OpenWRT, etc.) should be done without WLAN bring up, and integration of the BWL can come later, like what will be done for the Nighthawk (RAX40) platform.

### DIFF
--- a/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
+++ b/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
@@ -1518,7 +1518,6 @@ bool main_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
             LOG(DEBUG) << "slave is backhaul manager, updating";
             m_pBackhaulManagerSlave = sd;
 
-#if !defined(BEEROCKS_LINUX) && !defined(BEEROCKS_TURRIS_OMNIA)
             // Start ARP monitor
             if (enable_arp_monitor) {
                 if (!init_arp_monitor()) {
@@ -1534,9 +1533,6 @@ bool main_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
                     return false;
                 }
             }
-#else
-            LOG(INFO) << "*** ARP & DHCP Monitors are disabled ***";
-#endif
         }
     } break;
 
@@ -2126,13 +2122,18 @@ bool main_thread::init_arp_monitor()
             return false;
         }
 
-        m_uiArpMonIP   = info.ipa.s_addr;
-        m_uiArpMonMask = info.nmask.s_addr;
-
-        if (bpl_arp_mon_start(&m_ctxArpMon, config.bridge_iface.c_str()) != 0) {
+        int ret = bpl_arp_mon_start(&m_ctxArpMon, config.bridge_iface.c_str());
+        if (ret < 0) {
+            if (ret == -BPL_ERR_OPERATION_NOT_SUPPORTED) {
+                LOG(INFO) << "Skip starting ARP monitor (not supported)";
+                return (true);
+            }
             LOG(ERROR) << "Failed starting ARP monitor!";
             return (false);
         }
+
+        m_uiArpMonIP   = info.ipa.s_addr;
+        m_uiArpMonMask = info.nmask.s_addr;
 
         // Create wrapper socket classes
         if (!m_pArpRawSocket) {
@@ -2198,11 +2199,15 @@ bool main_thread::init_dhcp_monitor()
 
     if (!m_pDHCPMonSocket) {
 
-        int dhcp_mon_fd;
-        if ((dhcp_mon_fd = bpl_dhcp_mon_start(
-                 [](const char *op, const char *mac, const char *ip, const char *hostname) {
-                     dhcp_monitor_cb_wrapper(op, mac, ip, hostname);
-                 })) < 0) {
+        int dhcp_mon_fd = bpl_dhcp_mon_start(
+            [](const char *op, const char *mac, const char *ip, const char *hostname) {
+                dhcp_monitor_cb_wrapper(op, mac, ip, hostname);
+            });
+        if (dhcp_mon_fd < 0) {
+            if (dhcp_mon_fd == -BPL_ERR_OPERATION_NOT_SUPPORTED) {
+                LOG(INFO) << "Skip starting DHCP monitor (not supported)";
+                return (true);
+            }
             LOG(ERROR) << "Failed starting DHCP monitor: " << dhcp_mon_fd;
             return (false);
         }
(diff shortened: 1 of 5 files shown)


framework/platform/bpl/include/bpl/bpl_err.h
```diff
@@ -27,69 +27,71 @@ extern "C" {
...
+    ERROR_CODE(BPL_ERR_MONITOR_REPORT_PROCESS_FAIL)                                                \
+    ERROR_CODE(BPL_ERR_CONFIG_PLATFORM_REPORTED_INVALID_CONFIGURATION)                             \
+    ERROR_CODE(BPL_ERR_CONFIG_BACKHAUL_WIRED_INTERFACE_IS_UNSUPPORTED)                             \
+    ERROR_CODE(BPL_ERR_CONFIG_BACKHAUL_WIRELESS_INTERFACE_IS_UNSUPPORTED)                          \
+    ERROR_CODE(BPL_ERR_CONFIG_NO_VALID_BACKHAUL_INTERFACE)                                         \
+    ERROR_CODE(BPL_ERR_CONFIG_BRIDGE_INTERFACE_IS_EMPTY)                                           \
+    ERROR_CODE(BPL_ERR_WIFI_CONFIGURATION_CHANGE_TIMEOUT)                                          \
+    ERROR_CODE(BPL_ERR_WATCHDOG_PROCESS_STUCK)                                                     \
+    ERROR_CODE(BPL_ERR_WATCHDOG_PROCESS_ZOMBIE)                                                    \
+    ERROR_CODE(BPL_ERR_LAST)
```
morantr, line 96: As part of my changes on remove non-passive mode code, I deleted a lot of unused BPL error since it is not being used anymore, and aligned it as you did. Please add the new error codes at the end of the list so it would be easy to rebase.

framework/platform/bpl/linux/bpl_arp.cpp
```diff
@@ -17,7 +17,7 @@
 
 using namespace beerocks::bpl;
 
-int bpl_arp_mon_start(BPL_ARP_MON_CTX *ctx, const char *iface) { return 0; }
+int bpl_arp_mon_start(BPL_ARP_MON_CTX *ctx, const char *iface) { return -2; }
```
arnout, line 20: Why don't we use `-BPL_ERR_OPERATION_NOT_SUPPORTED` here? (Not actually a change for this PR because we use int instead of enum everywhere, so fixing that would be a separate PR.)

framework/platform/bpl/include/bpl/bpl_err.h
```diff
@@ -27,69 +27,71 @@ extern "C" {
...
-            BPL_ERR_CONFIG_PLATFORM_REPORTED_INVALID_CONFIGURATION) /* 57 */                       \
-        ERROR_CODE(BPL_ERR_CONFIG_BACKHAUL_WIRED_INTERFACE_IS_UNSUPPORTED) /* 58 */ ERROR_CODE(    \
-            BPL_ERR_CONFIG_BACKHAUL_WIRELESS_INTERFACE_IS_UNSUPPORTED) /* 59 */                    \
-        ERROR_CODE(BPL_ERR_CONFIG_NO_VALID_BACKHAUL_INTERFACE) /* 60 */ ERROR_CODE(                \
-            BPL_ERR_CONFIG_BRIDGE_INTERFACE_IS_EMPTY) /* 61 */                                     \
-        ERROR_CODE(BPL_ERR_WIFI_CONFIGURATION_CHANGE_TIMEOUT) /* 62 */ ERROR_CODE(                 \
-            BPL_ERR_WATCHDOG_PROCESS_STUCK) /* 63 */ ERROR_CODE(BPL_ERR_WATCHDOG_PROCESS_ZOMBIE)   \
-                                                                                                   \
-        /* 74 */ ERROR_CODE(BPL_ERR_LAST)
+    ERROR_CODE(BPL_ERR_NONE)                                                                       \
```
arnout, line 30: It took me a bit of figuring out what was going on here... Basically, the comment was removed, and this allowed clang-format to properly format it. An alternative would be to put the number as a second argument to the macro:

```
#define GENERATE_ERROR_ENUM(ENUM, VALUE) ENUM = VALUE,
#define GENERATE_ERROR_STRING(STRING, VALUE) #STRING "=" #VALUE,

#define FOREACH_ERROR_CODE(ERROR_CODE) \
    ERROR_CODE(BPL_ERR_NONE, 0) \
...
```

Anyway, this thing about the comment that is removed should have been part of the commit message.

(nitpick)

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -2129,7 +2125,12 @@ bool main_thread::init_arp_monitor()
         m_uiArpMonIP   = info.ipa.s_addr;
         m_uiArpMonMask = info.nmask.s_addr;
 
-        if (bpl_arp_mon_start(&m_ctxArpMon, config.bridge_iface.c_str()) != 0) {
+        int ret = bpl_arp_mon_start(&m_ctxArpMon, config.bridge_iface.c_str());
+        if (ret < 0) {
+            if (ret == -BPL_ERR_OPERATION_NOT_SUPPORTED) {
+                LOG(INFO) << "Skip starting ARP monitor (not supported)";
+                return (true);
```
arnout, line 2129: I would add an explicit comment here to explain that it is not needed to initialize m_pArpRawSocket and m_pArpMonSocket and m_tpArpEntriesCleanup because they won't be used.

Also, maybe it would be better to not initialize m_uiArpMonIP and m_uiArpMonMask either in this case?

(nitpick)

framework/platform/bpl/include/bpl/bpl_err.h
tomereli: The new error codes values have to be 1 and 2 so they can't be added to the end of the list. Rebase shouldn't be too hard even as it is now.

arnout (approved):
Just nitpicks

framework/platform/bpl/linux/bpl_arp.cpp
tomereli: We can do this in a followup PR indeed.

framework/platform/bpl/include/bpl/bpl_err.h
tomereli: Updated commit message accordingly. Didn't think it is worth the extra effort to manually add the enum value for each error code line, if we really need it we can do it in the future.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
tomereli: I'm not sure I understand why this comment is needed. I do however agree about moving the m_uiArpMonIP and m_uiArpMonMask initialization - and did that.

PR #511, opened 2019-11-26
Cleanups / preparation for UGW

arnout:
This PR contains the commits from #497 that were already reviewed or just had nitpicks (which are fixed in this PR).


cmake/Findsafec.cmake
```diff
@@ -0,0 +1,31 @@
...
+    SAFEC_LIBRARY
+    SAFEC_INCLUDE_DIRS
+)
+
+if (safec_FOUND)
+    add_library(safec UNKNOWN IMPORTED)
+
+    # Includes
+    set_target_properties(safec PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${SAFEC_INCLUDE_DIRS}/libsafec"
```
tomereli, line 22: As you mentioned in the other review - this should be `"${SAFEC_INCLUDE_DIRS}"`

tomereli (approved):
One comment but since I can't test it I'm approving, we can fix it later.

PR #512, opened 2019-11-26
Bugfix/ TLVF: empty class member initialization

### LINKED ISSUE #480: [BUG] TLVF: empty class member initialization
**Describe the bug** -
The following tlv class is used by tlvf_test to verify various TLVF functionalities:
```yaml
 _type: class
  _is_tlv_class : True
  # TODO tlvf currently requires type to be first field
  type:
    _type: uint8_t
    _value_const: 99
  # TODO tlvf currently requires length to be second field
  length: uint16_t
  var0: uint16_t

  simple_list_length: uint8_t
  simple_list:
    _type: uint16_t
    _length: [ simple_list_length ]
  
  test_string_length: uint8_t
  test_string:
    _type: char
    _length: [ test_string_length ]
    _length_max: 8

  complex_list_length: uint8_t
  complex_list:
    _type: cInner
    _length: [ complex_list_length ]

  var1: cInner
  var2: uint32_t
  unknown_length_list:
    _type: cInner
    _length: []

cInner:
  _type: class
  _is_tlv_class: true
  type:
    _type: uint16_t
    _value_const: 1
  # TODO tlvf currently requires length to be second field
  length: uint16_t
  list_length: uint8_t
  list:
    _type: uint8_t
    _length: [ list_length ]
  var1: uint32_t
  unknown_length_list_inner:
    _type: char
    _length: []

```

Since `var1` is a class member, it might contain variable-length / unknown-length lists. For this reason, its `init()` function doesn't create `var1` when it is first created in a `CmduMessageTx` (`parse == false`).
However, when `tlvVarlist` is created in a `CmduMessageRx` (`parse == true`), `var1` is created automatically by the `init()` function since its length is already known and cannot change.

This leads to the following confusing code when the user just wants to send the default values - `var1` still has to be added manually:

```
auto tlv4 = msg.addClass<tlvTestVarList>();
tlv4->add_var1(tlv4->create_var1());
msg.finalize(true);
```

This is not very intuitive and leads to cases where the user simply does not add `var1` which causes various and very hard to find errors.

A better outcome should be that `finalize()` will fail if such variables are present in one of the TLVs and were not added.

## Steps to reproduce

1. Remove the initialization of `var1` in tlvf_test.cpp: test_parser() function
2. Build
3. Run tlvf_test

## Expected behavior

Finalize will fail with a clear error message if any TLV inside the CMDU is not fully initialized.

CoralMalachi:
When a TLV contains a class member (lets call it var1), since it might
contain a variable-length/unknown-length lists, its its init() function
doesn't create var1 when it is first created in a CmduMessageTx (m_parse == false).

However, when the same TLV is created in a CmduMessageRx (m_parse == true),
var1 is created automatically by the init() function since its length is
already known.

Due to this, when creating a CmduMessageTx containing a class member, the
caller needs to create-add it even if it is not used.

In the next commits, Finalize will fail with a clear
error message if any TLV inside the CMDU is not fully initialized.

In order to illustrate the situation of an error case, remove the
manually adding of var1, and change the finalize() to fail if not each of the members of the type class in TLV is fully initialized

Fixes #480

### DIFF
--- a/framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/tlvWscM1.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/tlvWscM1.h
@@ -88,6 +88,7 @@ class tlvWscM1 : public BaseClass
         WSC::sWscAttrDevicePasswordID& device_password_id_attr();
         WSC::sWscAttrConfigurationError& configuration_error_attr();
         WSC::sWscAttrOsVersion& os_version_attr();
+        bool isPostInitSucceeded();
         std::shared_ptr<WSC::cWscVendorExtWfa> create_vendor_ext();
         bool add_vendor_ext(std::shared_ptr<WSC::cWscVendorExtWfa> ptr);
         std::shared_ptr<WSC::cWscVendorExtWfa> vendor_ext() { return m_vendor_ext_ptr; }
@@ -138,6 +139,7 @@ class tlvWscM1 : public BaseClass
         WSC::sWscAttrOsVersion* m_os_version_attr = nullptr;
         WSC::cWscVendorExtWfa *m_vendor_ext = nullptr;
         std::shared_ptr<WSC::cWscVendorExtWfa> m_vendor_ext_ptr = nullptr;
+        bool m_vendor_ext_init = false;
         bool m_lock_allocation__ = false;
 };
 
--- a/framework/tlvf/AutoGenerated/src/BaseClass.cpp
+++ b/framework/tlvf/AutoGenerated/src/BaseClass.cpp
@@ -13,7 +13,7 @@
 
 BaseClass::BaseClass(uint8_t *buff, const size_t buff_len, const bool parse, const bool swap_needed)
     : m_buff__(buff), m_buff_ptr__(buff), m_buff_len__(buff_len), m_parse__(parse),
-      m_swap__(swap_needed), m_init_succeeded(false)
+      m_swap__(swap_needed)
 {
 }
 BaseClass::~BaseClass() {}
--- a/framework/tlvf/AutoGenerated/src/CmduMessageTx.cpp
+++ b/framework/tlvf/AutoGenerated/src/CmduMessageTx.cpp
@@ -88,6 +88,10 @@ bool CmduMessageTx::finalize(bool swap_needed)
 
     // Update vendor specific length which is seperated to 3 classes
     for (auto class_it = m_class_vector.begin(); class_it != m_class_vector.end(); class_it++) {
+        if(!(*class_it)->isPostInitSucceeded()){
+            TLVF_LOG(ERROR) << "TLV post init failed";
+            return false;
+        }
         // Type is the first byte in the tlv buffer
         uint8_t tlv_type = *(*class_it)->getStartBuffPtr();
         if (tlv_type == uint8_t(eTlvType::TLV_VENDOR_SPECIFIC)) {
--- a/framework/tlvf/AutoGenerated/src/tlvf/ieee_1905_1/tlvWscM1.cpp
+++ b/framework/tlvf/AutoGenerated/src/tlvf/ieee_1905_1/tlvWscM1.cpp
@@ -465,6 +465,14 @@ WSC::sWscAttrOsVersion& tlvWscM1::os_version_attr() {
     return (WSC::sWscAttrOsVersion&)(*m_os_version_attr);
 }
 
+bool tlvWscM1::isPostInitSucceeded() {
+    if (!m_vendor_ext_init) {
+        TLVF_LOG(ERROR) << "vendor_ext is not initialized";
+        return false;
+    }
+    return true; 
+}
+
 std::shared_ptr<WSC::cWscVendorExtWfa> tlvWscM1::create_vendor_ext() {
     if (m_lock_order_counter__ > 5) {
         TLVF_LOG(ERROR) << "Out of order allocation for variable length list vendor_ext, abort!";
@@ -504,6 +512,7 @@ bool tlvWscM1::add_vendor_ext(std::shared_ptr<WSC::cWscVendorExtWfa> ptr) {
         TLVF_LOG(ERROR) << "Not enough available space on buffer";
         return false;
     }
+    m_vendor_ext_init = true;
     size_t len = ptr->getLen();
     m_vendor_ext_ptr = ptr;
     if (!buffPtrIncrementSafe(len)) { return false; }
--- a/framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
+++ b/framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
@@ -225,6 +225,14 @@ bool tlvTestVarList::add_complex_list(std::shared_ptr<cInner> ptr) {
     return true;
 }
 
+bool tlvTestVarList::isPostInitSucceeded() {
+    if (!m_var1_init) {
+        TLVF_LOG(ERROR) << "var1 is not initialized";
+        return false;
+    }
+    return true; 
+}
+
 std::shared_ptr<cInner> tlvTestVarList::create_var1() {
     if (m_lock_order_counter__ > 3) {
         TLVF_LOG(ERROR) << "Out of order allocation for variable length list var1, abort!";
@@ -266,6 +274,7 @@ bool tlvTestVarList::add_var1(std::shared_ptr<cInner> ptr) {
         TLVF_LOG(ERROR) << "Not enough available space on buffer";
         return false;
     }
+    m_var1_init = true;
     size_t len = ptr->getLen();
     m_var2 = (uint32_t *)((uint8_t *)(m_var2) + len - ptr->get_initial_size());
     m_unknown_length_list = (cInner *)((uint8_t *)(m_unknown_length_list) + len - ptr->get_initial_size());
--- a/framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
+++ b/framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
@@ -73,6 +73,7 @@ bool tlvTestVarList::alloc_simple_list(size_t count) {
     m_complex_list_length = (uint8_t *)((uint8_t *)(m_complex_list_length) + len);
     m_complex_list = (cInner *)((uint8_t *)(m_complex_list) + len);
     m_var1 = (cInner *)((uint8_t *)(m_var1) + len);
+    m_var3 = (cInner *)((uint8_t *)(m_var3) + len);
     m_var2 = (uint32_t *)((uint8_t *)(m_var2) + len);
     m_unknown_length_list = (cInner *)((uint8_t *)(m_unknown_length_list) + len);
     m_simple_list_idx__ += count;
@@ -142,6 +143,7 @@ bool tlvTestVarList::alloc_test_string(size_t count) {
     m_complex_list_length = (uint8_t *)((uint8_t *)(m_complex_list_length) + len);
     m_complex_list = (cInner *)((uint8_t *)(m_complex_list) + len);
     m_var1 = (cInner *)((uint8_t *)(m_var1) + len);
+    m_var3 = (cInner *)((uint8_t *)(m_var3) + len);
     m_var2 = (uint32_t *)((uint8_t *)(m_var2) + len);
     m_unknown_length_list = (cInner *)((uint8_t *)(m_unknown_length_list) + len);
     m_test_string_idx__ += count;
@@ -186,6 +188,7 @@ std::shared_ptr<cInner> tlvTestVarList::create_complex_list() {
         std::copy_n(src, move_length, dst);
     }
     m_var1 = (cInner *)((uint8_t *)(m_var1) + len);
+    m_var3 = (cInner *)((uint8_t *)(m_var3) + len);
     m_var2 = (uint32_t *)((uint8_t *)(m_var2) + len);
     m_unknown_length_list = (cInner *)((uint8_t *)(m_unknown_length_list) + len);
     return std::make_shared<cInner>(src, getBuffRemainingBytes(src), m_parse__, m_swap__);
@@ -216,6 +219,7 @@ bool tlvTestVarList::add_complex_list(std::shared_ptr<cInner> ptr) {
     if (!m_parse__) { (*m_complex_list_length)++; }
     size_t len = ptr->getLen();
     m_var1 = (cInner *)((uint8_t *)(m_var1) + len - ptr->get_initial_size());
+    m_var3 = (cInner *)((uint8_t *)(m_var3) + len - ptr->get_initial_size());
     m_var2 = (uint32_t *)((uint8_t *)(m_var2) + len - ptr->get_initial_size());
     m_unknown_length_list = (cInner *)((uint8_t *)(m_unknown_length_list) + len - ptr->get_initial_size());
     m_complex_list_vector.push_back(ptr);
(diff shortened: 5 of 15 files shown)


framework/tlvf/AutoGenerated/src/BaseClass.cpp
```diff
@@ -13,7 +13,7 @@
 
 BaseClass::BaseClass(uint8_t *buff, const size_t buff_len, const bool parse, const bool swap_needed)
     : m_buff__(buff), m_buff_ptr__(buff), m_buff_len__(buff_len), m_parse__(parse),
-      m_swap__(swap_needed), m_init_succeeded(false)
+      m_swap__(swap_needed), m_init_succeeded(false) , m_post_init_succeeded(true)
```
morantr: There is no point in initializing it here to a default value. It could be done in the member declaration. Same for the `m_init_succeeded`.
(nitpick)

framework/tlvf/src/include/tlvf/BaseClass.h
```diff
@@ -28,6 +28,7 @@ class BaseClass {
 
     size_t getLen();
     bool isInitialized();
+    virtual bool isPostInitSucceeded() { return m_post_init_succeeded; };
```
morantr: If you want to override it, then why did you do an implementation here? 
I think that the right thing is to define it as pure virtual with no implementation.

framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
```diff
@@ -225,6 +225,13 @@ bool tlvTestVarList::add_complex_list(std::shared_ptr<cInner> ptr) {
     return true;
 }
 
+bool tlvTestVarList::isPostInitSucceeded() {
+    if( !is_var1_initialized() ){
```
morantr: Please remove the spaces between the `if` brackets to the internal condition, and add space between the `if` to the open bracket.

framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
```diff
@@ -225,6 +225,13 @@ bool tlvTestVarList::add_complex_list(std::shared_ptr<cInner> ptr) {
...
+        m_post_init_succeeded=false;
```
morantr: Please add space before and after the `=` sign.

framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
```diff
@@ -266,6 +273,7 @@ bool tlvTestVarList::add_var1(std::shared_ptr<cInner> ptr) {
         TLVF_LOG(ERROR) << "Not enough available space on buffer";
         return false;
     }
+    m_var1_init = true;
```
morantr, line 286: I gotta say I don't fully understand what is the difference between setting this boolean to true, to return true in the function.
If the function is failing, then this member will also be set to false.

morantr (changes requested):

framework/tlvf/src/include/tlvf/BaseClass.h
CoralMalachi: Since I want to override this method only in cases that the TLV contains a member of type class.

tomereli (commented):
I would change the start of the first commit message (and PR description):

```
When a TLV contains a class member (lets call it var1), since it might contain a
variable-length/unknown-length lists, its its init() function
doesn't create var1 when it is first created in a CmduMessageTx (m_parse == false).

However, when the same TLV is created in a CmduMessageRx (m_parse == true), var1 is created
automatically by the init() function since its length is already known.

Due to this, when creating a CmduMessageTx containing a class member, the caller needs to
create-add it even if it is not used.
```

The rest of the

tomereli (commented):
- commit e8afd796d57506b85e4ae10e147fd33deea7a2db: `isPostInitSucceede` should be `isPostInitSucceeded`

- each of the members of the type class in TLV **are** fully initialized.

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/tlvWscM1.h
```diff
@@ -88,6 +88,8 @@ class tlvWscM1 : public BaseClass
         WSC::sWscAttrDevicePasswordID& device_password_id_attr();
         WSC::sWscAttrConfigurationError& configuration_error_attr();
         WSC::sWscAttrOsVersion& os_version_attr();
+        bool is_vendor_ext_initialized() { return m_vendor_ext_init;}
```
tomereli: No need for a function, you can access these member variables from `isPostInitSucceeded()` in the derived class.

framework/tlvf/AutoGenerated/src/tlvf/ieee_1905_1/tlvWscM1.cpp
```diff
@@ -465,6 +465,13 @@ WSC::sWscAttrOsVersion& tlvWscM1::os_version_attr() {
     return (WSC::sWscAttrOsVersion&)(*m_os_version_attr);
 }
 
+bool tlvWscM1::isPostInitSucceeded() {
+    if( !is_vendor_ext_initialized() ){
```
tomereli: Just check `m_vendor_ext_init` instead of using a function here.

framework/tlvf/AutoGenerated/src/tlvf/ieee_1905_1/tlvWscM1.cpp
```diff
@@ -465,6 +465,13 @@ WSC::sWscAttrOsVersion& tlvWscM1::os_version_attr() {
...
+        m_post_init_succeeded=false;
```
tomereli: add a space before and after the `=`

framework/tlvf/AutoGenerated/src/tlvf/ieee_1905_1/tlvWscM1.cpp
tomereli: Remove the space before the `!`. Remove the space before the `)`. Add a space before the `{`
So:
```
if (!is_vendor_ext_initialized()) {
```

(obviously should be done in tlvf.py)

framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
tomereli: When you return true, the caller knows that add succeeded. But if the caller never called the add function, tlvf needs to be aware of that so in Finalize we check the value of this variable.

framework/tlvf/tlvf.py
```diff
@@ -209,6 +209,7 @@ def __init__(self, fname, name, dict):
         self.list_index = 0
         self.lock_allocation_member_added = False
         self.lock_order_member_added = False
+        self.marker_post_init_method_added = False
```
tomereli, line 212: Can we call it `post_init_method_added` instead and set it to True after we encounter the first class member and really add the marker and the function?

framework/tlvf/tlvf.py
```diff
@@ -336,7 +337,7 @@ def __init__(self, src_path, yaml_path, out_path, conf_path, print_dependencies,
         self.CODE_CLASS_PUBLIC_FUNC_INSERT          = "//~class_public_func_insert"
         self.CODE_CLASS_PRIVATE_FUNC_INSERT         = "//~class_private_func_insert"
         self.CODE_CLASS_INIT_FUNC_INSERT            = "//~class_init_func_insert"
-        self.CODE_CLASS_INIT_FUNC_SWAP_INSERT            = "//~class_init_func_swap_insert"
+        self.CODE_CLASS_INIT_FUNC_SWAP_INSERT       = "//~class_init_func_swap_insert"
```
tomereli, line 340: Should be on a separate commit (nitpick)

framework/tlvf/tlvf.py
```diff
@@ -346,6 +347,7 @@ def __init__(self, src_path, yaml_path, out_path, conf_path, print_dependencies,
         self.CODE_STRUCT_REVERSED_BITFIELD_INSERT   = "//~struct_reversed_bitfield_insert"
         self.CODE_STRUCT_SWAP_FUNC_INSERT           = "//~struct_swap_func_insert"
         self.CODE_STRUCT_INIT_FUNC_INSERT           = "//~struct_init_func_insert"
+        self.CODE_CLASS_IS_INIT_FUNC                = "//~class_is_init_func"
```
tomereli: Please move this to after line 342 so it will be with its family :)

framework/tlvf/tlvf.py
```diff
@@ -644,6 +646,41 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
             if TypeInfo(param_type).type == TypeInfo.CLASS:                
                 lines_h.append("%s *m_%s = nullptr;" % (param_type, param_name))
                 lines_h.append("std::shared_ptr<%s> m_%s_ptr = nullptr;" % (param_type, param_name))
+                lines_h.append("bool m_%s_init = false;" % (param_name))
+                
+                #Add a get method in the H file
```
tomereli: Should go into a function with documentation explaining all we are doing here and why. Its about time that tlvf gets some good comments.

framework/tlvf/tlvf.py
```diff
@@ -644,6 +646,41 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
             if TypeInfo(param_type).type == TypeInfo.CLASS:                
                 lines_h.append("%s *m_%s = nullptr;" % (param_type, param_name))
                 lines_h.append("std::shared_ptr<%s> m_%s_ptr = nullptr;" % (param_type, param_name))
+                lines_h.append("bool m_%s_init = false;" % (param_name))
```
tomereli: This should go to the same function below

framework/tlvf/tlvf.py
```diff
@@ -1057,6 +1094,8 @@ def addClassVarLenMethods(self, obj_meta, param_type, param_name, param_meta, pa
                 lines_cpp.append( "%sm_%s_idx__++;" % (self.getIndentation(1), param_name) )
                 if is_var_len:
                     lines_cpp.append( "%sif (!m_parse__) { (*m_%s)++; }" % (self.getIndentation(1), param_length) )
+            else: 
+                lines_cpp.append("%sm_%s_init = true;" %(self.getIndentation(1),param_meta.name))
```
tomereli, line 1121: please add a comment here which explains this - refer to the function which adds the member that you will add later.

framework/tlvf/tlvf.py
```diff
@@ -1167,7 +1206,6 @@ def addAllocationMarkersAlloc(self, obj_meta, param_meta, param_length, memmove)
             marker = "%s_%s_%s_%s" %(self.CODE_CLASS_ALLOC_INSERT, obj_meta.name, "alloc", param_meta.name)
             obj_meta.alloc_list.append((marker, "alloc"))
             lines_cpp.append(marker)
-
```
tomereli: why remove the line break? same goes to the below.

framework/tlvf/AutoGenerated/src/CmduMessageTx.cpp
```diff
@@ -88,6 +88,10 @@ bool CmduMessageTx::finalize(bool swap_needed)
 
     // Update vendor specific length which is seperated to 3 classes
     for (auto class_it = m_class_vector.begin(); class_it != m_class_vector.end(); class_it++) {
+        if(!(*class_it)->isPostInitSucceeded()){
```
tomereli, line 91: Is there any way we can add a print here saying which class postInit failed? Otherwise the error print will not say much.
Perhaps @arnout has an idea..

(nitpick)

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -303,7 +303,10 @@ int test_parser()
     // tlv4->add_var1(tlv4->create_var1());
 
     LOG(DEBUG) << "Finalize";
-    msg.finalize(true);
+    if (!msg.finalize(true)) {
```
tomereli, line 323: Just to make sure - do we have this in the controller and agent?

tomereli (changes requested):

framework/tlvf/tlvf.py
```diff
@@ -583,6 +583,62 @@ def addStructParam(self, obj_meta, param_name, param_type, param_type_info, para
         else:
             self.insertLineH(obj_meta.name, self.CODE_STRUCT_INSERT, line)
 
+    ##########################################################################################
```
tomereli, line 586: BIG LIKE!

framework/tlvf/tlvf.py
```diff
@@ -583,6 +583,62 @@ def addStructParam(self, obj_meta, param_name, param_type, param_type_info, para
...
+        #if this is the first member of type class override the method 
+        if not obj_meta.marker_post_init_method_added:
+            #Add decleration in header file
+            line_h = "bool isPostInitSucceeded();" 
+            self.insertLineH(obj_meta.name, self.CODE_CLASS_PUBLIC_FUNC_INSERT, line_h)
+                    
+            #Override isPostInitSucceeded method to verify the class is completly initialized
+            lines_cpp.append("bool %s::isPostInitSucceeded() {" % (obj_meta.name))
+            #if this member is not added by the user return false
+            lines_cpp.append("%sif(!m_%s_init) {" % (self.getIndentation(1) ,param_name))
```
tomereli: we don't run clang-format here but we can at least try to align to our coding rules, so add a space after the `if`:
```
lines_cpp.append("%sif (!m_%s_init) {" % (self.getIndentation(1) ,param_name))
```

Same comment for other places you introduced an `if` without a space.

framework/tlvf/tlvf.py
```diff
@@ -583,6 +583,62 @@ def addStructParam(self, obj_meta, param_name, param_type, param_type_info, para
...
+            lines_cpp.append("%sm_post_init_succeeded = false;" %(self.getIndentation(2)))
+            lines_cpp.append("%sTLVF_LOG(ERROR) << \"%s is not initialized\";" %(self.getIndentation(2), param_name))
+            lines_cpp.append("%s}" % (self.getIndentation(1)))
+
+            #in case there is more than one member of type class, add a marker 
+            #before the return statement
+            marker = "%s_%s" %(self.CODE_CLASS_IS_INIT_FUNC, obj_meta.name)
+            lines_cpp.append(marker)
+
```
tomereli: why all these line breaks? (nitpick)

framework/tlvf/tlvf.py
```diff
@@ -1095,6 +1118,8 @@ def addClassVarLenMethods(self, obj_meta, param_type, param_name, param_meta, pa
                 if is_var_len:
                     lines_cpp.append( "%sif (!m_parse__) { (*m_%s)++; }" % (self.getIndentation(1), param_length) )
             else: 
+                # When the user calls the add_member method, the value of the bool variable is set to TRUE.
+                # Fore more information, please read overrideIsPostInitSucceeded method documentation
```
tomereli: `Fore` --> `For`

framework/tlvf/AutoGenerated/src/CmduMessageTx.cpp
```diff
@@ -88,6 +88,10 @@ bool CmduMessageTx::finalize(bool swap_needed)
...
+            TLVF_LOG(ERROR) << "A TLV with a class member is not fully initialized";
```
tomereli: Log should be more generic, saying "TLV post init failed".

tomereli (approved):
Mainly nitpicks and small typos changes, so please fix before merging, but otherwise LGTM. Nice

morantr (approved):

PR #513, opened 2019-11-26
Feature/gitlab ci

### LINKED ISSUE #504: [TASK] Create a Gitlab mirror with CI
## Description

Some features, such as uploading artifacts, are needed to run the certification tests as part of the CI, and are not available on TravisCI. 
Therefore, we would like to setup a Gitlab mirror and setup our own Gitlab runners.

- [x] Create a mirror of this repository on the prpl.dev Gitlab instance 
- [x] Port the travis CI configuration to Gitlab CI
- [x] Setup and assign Gitlab runners
- [x] Add a first CI configuration which reproduces the Travis CI

rmelotte:
Closes #504.
We would like to switch from Travis to Gitlab CI because it's a lot more flexible.

This PR adds a first .gitlab-ci.yml file, which more or less reproduces the Travis configuration. 
Once merged, it will then be synced to the existing [mirror](https://git.prpl.dev/prplmesh/prplMesh).
The content of this PR has already been tested on Gitlab CI, the results are [here](https://git.prpl.dev/prplmesh/prplMesh/pipelines/61).


.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    - git update-index -q --refresh
+    - git diff-index --exit-code HEAD
+  tags:
+    - shell
+
+run-tests:
+  stage: test
+  script:
+    - cd tools/docker
+    - ./run.sh -f -d --entrypoint "$PWD/../../../build/install/bin/tests/tlvf_test"
```
tomereli, line 29: What about encryption test?

arnout: It is currently not in .travis.yml. I think it's a good idea for the first version of .gitlab-ci.yml to be as close as possible to .travis.yml.

arnout:
> I think thus will mean that running test_gateway_repeater.sh will now kill the container after it is complete? If not, will we still be able to run docker exec - it gateway bash and debug as we normally do?

`--interactive --tty` is completely unrelated to killing the container.

`--interactive` will make sure that stdin is connected. So it is completely useless on any run or exec command that doesn't actually use stdin.

`--tty` will allocate a pty. In practice, this means that isatty() will return true, which will make some programs behave differently. For example, there won't be any colours in the output of `make`.

That the containers keep on running is proven by the fact that Travis CI succeeded on this PR: test_flows.sh requires the gateway, repeater1 and repeater2 containers to be running...

.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
+before_script:
+  - docker info
```
arnout: Nice for your debugging, but is it useful in production?

.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+  # check the availability of some commands:
+  - command >/dev/null -v python3 jq
+
+build-in-docker:
+  stage: build
+  script:
+    - cd tools/docker
+    # Try to pull from dockerhub, build the container images if that fails
+    - ./image-pull.sh || ./image-build.sh
```
arnout, line 13: For building, it should be possible to use the prplmesh-builder:ubuntu-18.04 container directly. Of course we can't use the shell runner then, so remove the shell tag.

Note that this is something we should probably do in a follow-up PR - for this PR, it's good to stay as close as possible to .travis.yml.

D'oh, I realize now that the builder container has maptools as the entrypoint, so that won't work... We'd first have to remove the entrypoint from that container so we have full flexibility for running stuff in it.

.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    # Run clang-format and check there are no modified files by checking git status
+    - ./clang-format.sh
```
arnout, line 15: We can use the top-level clang-format.sh script itself if we run it in the appropriate docker container.

It would also be nice to put this in a separate job so it can be done in parallel (of course the git diff-index has to be included in the same job).

Again, both of these comments are for follow-up PRs.

.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    - git diff-index --exit-code HEAD
+  tags:
+    - shell
+
+run-tests:
+  stage: test
+  script:
+    - cd tools/docker
+    - ./run.sh -f -d --entrypoint "$PWD/../../../build/install/bin/tests/tlvf_test"
+    - ./tests/test_flows.sh -v
```
arnout, line 30: This will leave the test containers running directly in the machine which has the runner (i.e., mind2). Not nice... Please add a after_script to kill the containers. This **is** something that must be fixed for this PR.

Also, since this is in a shell runner, the containers started by different instances all have the same name "gateway", "repeater1", "repeater2". Therefore, if two jobs are executing concurrently, they'll conflict with each other. We need to find a way to resolve that - but again, can be done later.

.gitlab-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+    - ./image-pull.sh || ./image-build.sh
+    # Run clang-format and check there are no modified files by checking git status
+    - ./clang-format.sh
+    - git diff-index --exit-code HEAD
+    - ./build.sh map -f CMAKE_BUILD_TYPE=Debug BUILD_TESTS=ON
+    # Check that the AutoGenerated files are correct by checking git status
+    - git update-index -q --refresh
+    - git diff-index --exit-code HEAD
+  tags:
+    - shell
```
arnout, line 22: Please use the "dockerbuild" tag. If we later switch to one of the other approaches, the tag will stay the same, it will just be a differently-configured runner.

arnout (approved):
Lots of comments but mostly for follow-up PRs. Just two things that need to be fixed: "dockerbuild" tag (for both) and an after_script to kill the containers.

.gitlab-ci.yml
rmelotte: It's useful to have when a test fails, but as in our case we will always have access to the runners it's probably not needed. I'll remove it.

rmelotte:
I made the required changes, and opened an issue (#515) for the comments that won't be solved as part of this PR.
The new Gitlab pipeline is [here](https://git.prpl.dev/prplmesh/prplMesh/pipelines/63).

rmelotte:
I forgot to remove the "docker info" part, which is now done. 
Pipeline: https://git.prpl.dev/prplmesh/prplMesh/pipelines/64

PR #516, opened 2019-11-27
build: cmake: remove -DDEBUG

tomereli:
Remove -DDEBUG from common and framework CMakeLists.txt since it
is not used, and since it clashes with other DEBUG macros defined
in the UCI BPL when compiling for RDKB or UGW.


arnout (approved):
I had the same commit with a slightly different commit message as the first on #514

PR #519, opened 2019-11-28
bwl: more TARGET_PLATFORM cmake cleanups

tomereli:
TARGET_PLATFORM can no longer be "ugw", for that we have TARGET_PLATFORM_TYPE
(and this is also temporary), so remove these sections from the CMakeLists.txt.
While at it, since we now have find module for dwpal, no need to use the include_directories
for rdkb.

### DIFF
--- a/common/beerocks/bwl/CMakeLists.txt
+++ b/common/beerocks/bwl/CMakeLists.txt
@@ -16,32 +16,19 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
 set(BWL_TYPE "DUMMY")
 add_definitions(-DBEEROCKS_TMP_PATH="${TMP_PATH}")
 
-# UGW
-if (TARGET_PLATFORM STREQUAL "ugw")
-
-    set(BWL_TYPE "DWPAL")
-    # Extra libraries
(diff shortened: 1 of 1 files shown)


tomereli:
Adam Dov must review this PR prior to approving it since it might have an effect on the standalone repeater.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -16,32 +16,19 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
...
 
-# Turris Omnia
 elseif(TARGET_PLATFORM STREQUAL "openwrt")
 
-    set(BWL_TYPE "NL80211")
+    if (TARGET_PLATFORM_TYPE STREQUAL "ugw")
+        set(BWL_TYPE "DWPAL")
+    else()
+        set(BWL_TYPE "NL80211")
+    endif()
```
arnout, line 31: What I would really prefer is that BWL_TYPE is set explicitly by the build system, so either by the openwrt Makefile (if building as part of the firmware) or by the openwrt.cmake file (if building out-of-tree).

But that's for a different PR.

arnout (approved):

PR #521, opened 2019-11-29
cmake: allow any hostapd variant

rmelotte:
When building with OpenWRT, it's possible to use different variant of
hostapd: hostapd, hostapd-full, hostapd-mini, and so on.

This commit makes it possible to use those different variants, and not
only hostapd-full.


tomereli (approved):

arnout (approved):
Just to be clear: to actually work, prplMesh requires the hostapd-full variant. So in the openwrt Makefile for prplmesh we'll have to make sure that hostapd-full is selected.

However, it is perfectly possible to build prplmesh outside of openwrt, using one of the smaller variants (one of them is (almost) always built so it will be available), and install it later by copying the files over. This PR handles that particular use case.

PR #522, opened 2019-11-29
Hotfix/fix nl80211 compilation

rmelotte:
Since commit dfe30fae1e0de29c0acb0f42af3c8ffbbf867ef9 (common: bwl:
move reject status handle to bwl), the reject_sta argument of the
sta_deny function was removed from the base ap_wlan_hal class.

Because it was not removed in the nl80211 implementation, the
compilation currently fails with:
error: 'virtual bool bwl::nl80211::ap_wlan_hal_nl80211::sta_deny(const
     string&, int)' marked 'override', but does not override virtual
     bool sta_deny(const std::string &mac, int reject_sta) override;

Since the current nl80211 implementation of sta_deny makes no use of
reject_sta, remove it.

### DIFF
--- a/common/beerocks/bwl/CMakeLists.txt
+++ b/common/beerocks/bwl/CMakeLists.txt
@@ -44,7 +44,7 @@ elseif(TARGET_PLATFORM STREQUAL "openwrt")
     set(BWL_TYPE "NL80211")
 
     # hostapd directory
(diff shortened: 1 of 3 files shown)


arnout (approved):

arnout:
Excellent commit message BTW.

PR #529, opened 2019-12-02
agent: set BEEROCKS_PLATFORM for the turris-omnia

rmelotte:
BEEROCKS_PLATFORM is not set for the turris-omnia, so
"agent/config/beerocks_agent.conf.in" will end up with an empty
platform.

Set it to turris-omnia for now, as it's what's being done for other
platforms.

This is just a quick fix to get it working on the Turris Omnia.
Since we want to avoid TARGET_PLATFORM_TYPE, shouldn't the platform just be set to "openwrt" for all the targets using OpenWRT?


tomereli:
> BEEROCKS_PLATFORM is not set for the turris-omnia, so
> "agent/config/beerocks_agent.conf.in" will end up with an empty
> platform.
> 
> Set it to turris-omnia for now, as it's what's being done for other
> platforms.
> 
> This is just a quick fix to get it working on the Turris Omnia.
> Since we want to avoid TARGET_PLATFORM_TYPE, shouldn't the platform just be set to "openwrt" for all the targets using OpenWRT?

I think that you're right - `TARGET_PLATFORM` default should be "openwrt" in this case, and `BEEROCKS_PLATFORM` should always use `TARGET_PLATFORM`. So why a workaround and not just fix it?

tomereli (approved):
Approved, but I would rather see this commit fix the issue and not add a workaround. I don't see why we can't set `BEEROCKS_PLATFORM=@TARGET_PLATFORM`.

rmelotte:
@tomereli I didn't know whether the other platforms (rdkb for instance) make any use of the "platform" field in the agent configuration file. Is it only used for logging/display purposes or do some features depend on it?
In the former case we could just drop BEEROCKS_PLATFORM and use `platform=@TARGET_PLATFORM@` directly (then `platform` will either be "rdkb", "linux" or "openwrt").
Does that sound good to you?

arnout:
> I didn't know whether the other platforms (rdkb for instance) make any use of the "platform" field in the agent configuration file. Is it only used for logging/display purposes or do some features depend on it?

In fact, the platform is only used for debug info. And in a complicated way, because it is read from the config file as a string, then converted to an enum in the agent, then sent to the controller as a uint8_t, then converted back to a different enum, then converted to a string in bml.

So my vote is to eliminate all this craziness. It should in the end be replaced with a manufacturer etc. strings from M1.

That said, eliminating it is a bit of work, so for now we can live with this workaround I suppose.

tomereli:
> > I didn't know whether the other platforms (rdkb for instance) make any use of the "platform" field in the agent configuration file. Is it only used for logging/display purposes or do some features depend on it?
> 
> In fact, the platform is only used for debug info. And in a complicated way, because it is read from the config file as a string, then converted to an enum in the agent, then sent to the controller as a uint8_t, then converted back to a different enum, then converted to a string in bml.
> 
> So my vote is to eliminate all this craziness. It should in the end be replaced with a manufacturer etc. strings from M1.
> 
> That said, eliminating it is a bit of work, so for now we can live with this workaround I suppose.

I agree with @arnout. Please open a task for it and feel free to merge this for now as a workaround. - unless you'd like to do this as part of this PR, in which case I vote for it :)

PR #531, opened 2019-12-02
Bugfix/ucc agent listener

### LINKED ISSUE #336: [TASK] Support wfa_ca in agent
## Description

## Exit criteria

   - `DEV_RESET_DEFAULT`
   - `DEV_SET_CONFIG,backhaul`

## Detailed description

tomereli:
Any test case triggered by the UCC application starts with `dev_reset_defaults` which "tells" the agent that it should go back to its default state, meaning unconfigured and *not* onboarded.
Next, onboarding is triggered as part of the `dev_set_config` command.

This PR completes the agent wfa_ca listener (agent ucc listener) by adding agent implementation to dev_reset_default and dev_set_config.

The complete prplMesh agent flow starts with the UCC application triggering
dev_reset_default, which results in a full slave reset triggered by the backhaul manager as a response to the ucc agent's request. 
The ucc agent now waits in `BACKHAUL_SEND_ENABLE` state until the `dev_set_config` command from the UCC application is received, thus blocking the agent from starting onboarding + autoconfig flows. 

Next, the UCC sends `dev_set_config,name,IntelCTTDUT,program,map,backhaul,eth` command (can be either wired or wireless), for which the ucc agent listener replies with `RUNNING`, and sets the onboarding and onboarding_in_progress flags to `true`, to signal the son_agent that it should continue now with onboarding.

The son_slave now unblocks and proceeds with the state machine, which results with BACKHAUL_ENABLE, autoconfig search, autoconfig M1/M2 exchange by *both* radio agents, and finally the slave gets to operational state.

In this state, since the onboarding_inprogress flag is set, the son_slave ends the flow by replying COMPLETE to the UCC and resetting the flags in the agent's UCC listener.

Fixes #336 


common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -731,17 +731,7 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
 
         // TODO: Find out what to do with value of param "name".
 
-        // send back first reply
-        if (!reply_ucc(eWfaCaStatus::RUNNING)) {
-            LOG(ERROR) << "failed to send reply";
-            break;
-        }
-
```
arnout, line 739: This change is semantically wrong. The CAPI implementation should do the following steps:

- validate command and arguments;
- send status RUNNING or INVALID (in the latter case, terminate);
- do whatever the command is supposed to do;
- send status COMPLETE (with response arguments) or ERROR (with error code).

So, the RUNNING status should stay here, only the COMPLETE status should move to the reset implementation.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -876,18 +876,7 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
         if (!handle_dev_set_config(params, err_string)) {
             LOG(ERROR) << err_string;
             reply_ucc(eWfaCaStatus::INVALID, err_string);
-            break;
-        }
-
-        // Send back first reply
-        if (!reply_ucc(eWfaCaStatus::RUNNING)) {
-            LOG(ERROR) << "failed to send reply";
-            break;
```
arnout, line 895: If the RUNNING reply moves inside handle_dev_set_config(), then it makes sense to move the INVALID reply there as well. Therefore, I think handle_dev_set_config() should return void instead, and you simply end this function with

```
handle_dev_set_config(params);
break;
```

arnout (commented):
I've only made a cursory review of the implementation because I think there are architectural problems with this approach.

First of all, I want to revisit @morantr's decision to put the UCC listener on the slave. The reasoning behind it was the following:
> First I thought to implement it in the backhaul manager, but it caused
> inconsistency on send cmdu function which on the controller, uses
> a socket, and on the backhaul manager uses the bus, so I have decided to
> implement it on son_slave.

However, I think it's actually the controller that is doing things "the wrong way": it should use the bus instead of sockets. What happens now is that mrouter creates a new socket for each agent that talks to the controller, and passes messages from the bus through such a socket and vice versa. That's just crazy... In fact, the two calls to send_cmdu_to_destination currently require an additional call to get_dev_send_1905_destination_socket() to get the Socket from the AL-MAC. Instead, we should just use the AL-MAC directly as destination in send_cmdu_to_bus(). Then, we can move the UCC listener to the backhaul manager. For this PR, the impact of that is not too big yet (it just avoids a few of the additional state bits); but for future agent UCC operations, they *will* need to address a specific slave, so it has to go through the backhaul manager thread in the end. Better do the listen there immediately.

Second, I think we should use the state machine as much as possible instead of adding additional flags. In this case, I think the reset command should move the backhaul manager state to RESTART and let it wait in state WAIT_ENABLE until the set_config command is received. I think the set_config command can simply set pending_enable to true, but maybe the state machine needs to be updated instead.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
morantr: No, you should revert it all. It all should be in the base class. If you want to wait that the configuration will finish or something, then block the thread of the agent_ucc_listener.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4257,6 +4263,23 @@ bool slave_thread::monitor_heartbeat_check()
...
+void slave_thread::handle_ucc_listener_reset_defaults()
+{
+    if (m_agent_ucc_listener && m_agent_ucc_listener->dev_reset_default_needed()) {
+        LOG(INFO) << "UCC listener received dev_reset_default, trigger slave_reset !!";
+        m_agent_ucc_listener->set_dev_reset_default_inprogress();
+        // Send BACKHHAUL_RESET so full reset is done
+        LOG(TRACE) << "send ACTION_BACKHAUL_RESET";
+        auto request =
+            message_com::create_vs_message<beerocks_message::cACTION_BACKHAUL_RESET>(cmdu_tx);
+        if (request == nullptr) {
```
morantr, line 4312: if (!request)

agent/src/beerocks/slave/agent_ucc_listener.h
```diff
@@ -26,6 +26,27 @@ class agent_ucc_listener : public beerocks_ucc_listener {
     void lock() override { mutex.lock(); }
     void unlock() override { mutex.unlock(); }
 
+    bool dev_reset_default_needed() { return m_dev_reset_default; }
+    bool dev_reset_default_inprogress() { return m_dev_reset_default_inprogress; }
+    void set_dev_reset_default_inprogress()
+    {
+        m_dev_reset_default            = false;
+        m_dev_reset_default_inprogress = true;
```
morantr, line 35: Why did you concatenate the word "in" and "progress" everywhere? It is two separate words.
Should be `in_progress` instead.
(nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3554,14 +3556,22 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
...
+            m_agent_ucc_listener->set_dev_reset_default_complete();
+            // Now wait for onboading command from UCC, so don't send backhaul manager enable yet
+        } else if (m_agent_ucc_listener && !m_agent_ucc_listener->dev_set_config_onboarding()) {
+            slave_state = STATE_BACKHAUL_ENABLE;
         } else {
-            // backhaul manager will request for backhaul iface and tx enable after receiving ACTION_BACKHAUL_ENABLE,
-            // when wireless connection is required
-            LOG(TRACE) << "goto STATE_SEND_BACKHAUL_MANAGER_ENABLE";
-            slave_state = STATE_SEND_BACKHAUL_MANAGER_ENABLE;
+            if (error) {
```
morantr, line 3595: If you move to a state put `break` immidiatly after. This will promise that no one would be able to change the state after that. Also, you could get rid of the `else` statement.

morantr (changes requested):

morantr:
I don't like what you did moving the sending of the COMPLETE and RUNNING from the base class to the derived class.
You should have blocked the agent_ucc_listener thread until the operation is complete by polling the status until you reach timeout (which is defined in the specification).
On each polling iteration, you need to use `unlock()` to unlock the mutex and use linux `yield()` to make the thread sleep.

tomereli:
> I've only made a cursory review of the implementation because I think there are architectural problems with this approach.
> 
> First of all, I want to revisit @morantr's decision to put the UCC listener on the slave. The reasoning behind it was the following:
> 
> > First I thought to implement it in the backhaul manager, but it caused
> > inconsistency on send cmdu function which on the controller, uses
> > a socket, and on the backhaul manager uses the bus, so I have decided to
> > implement it on son_slave.
> 
> However, I think it's actually the controller that is doing things "the wrong way": it should use the bus instead of sockets. What happens now is that mrouter creates a new socket for each agent that talks to the controller, and passes messages from the bus through such a socket and vice versa. That's just crazy... In fact, the two calls to send_cmdu_to_destination currently require an additional call to get_dev_send_1905_destination_socket() to get the Socket from the AL-MAC. Instead, we should just use the AL-MAC directly as destination in send_cmdu_to_bus(). Then, we can move the UCC listener to the backhaul manager. For this PR, the impact of that is not too big yet (it just avoids a few of the additional state bits); but for future agent UCC operations, they _will_ need to address a specific slave, so it has to go through the backhaul manager thread in the end. Better do the listen there immediately.
> 
> Second, I think we should use the state machine as much as possible instead of adding additional flags. In this case, I think the reset command should move the backhaul manager state to RESTART and let it wait in state WAIT_ENABLE until the set_config command is received. I think the set_config command can simply set pending_enable to true, but maybe the state machine needs to be updated instead.

As discussed - I agree that the path in which this PR is going is the "dark path", and we should get rid of the `mrouter` and move the ucc agent listener to the backhaul manager.
However, in the meanwhile, in order to have something working that can be used with the certification testbed to actually progress with tests, I will go further down this "dark path" only to the point we have it working (missing dev_get_param for the agent to get `macaddr`). Then we can parallelize tests between developers and do this right.

The right way would be to have the agent listener as part of the backhaul. We would still need to cache the VAP list in the backhaul manager in certification mode for the `dev_get_param` implementation.

More design thoughts:
- I think that the base class should implement only common commands such as `dev_send_1905`.
- For any other command, the implementation should be in the derived class, so `reply_ucc` has to be protected and not private
- `parse_params` should also be protected and not private for the same reason
- This PR does not take care at all to timeouts. This should be handled
- I agree about the part of utilizing the state machine and not adding conditional variables.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
tomereli: OK, agree.

agent/src/beerocks/slave/agent_ucc_listener.h
tomereli: too long. I think `inprogress` is just fine, but in any case I'll think of how to get rid of these completely.

tomereli:
> I don't like what you did moving the sending of the COMPLETE and RUNNING from the base class to the derived class.
> You should have blocked the agent_ucc_listener thread until the operation is complete by polling the status until you reach timeout (which is defined in the specification).
> On each polling iteration, you need to use `unlock()` to unlock the mutex and use linux `yield()` to make the thread sleep.

Sounds reasonable enough, You're right, I did it quickly to get it working. This is the "dark path" so we don't care about edge cases here :-D

morantr:
> > I don't like what you did moving the sending of the COMPLETE and RUNNING from the base class to the derived class.
> > You should have blocked the agent_ucc_listener thread until the operation is complete by polling the status until you reach timeout (which is defined in the specification).
> > On each polling iteration, you need to use `unlock()` to unlock the mutex and use linux `yield()` to make the thread sleep.
> 
> Sounds reasonable enough, You're right, I did it quickly to get it working. This is the "dark path" so we don't care about edge cases here :-D

But why not doing it the right way? It is not like you need to change the whole infrastructure to do it (like getting rid of mrouter)...

tomereli:
[Agent UCC Listener refactor](https://trello.com/c/ir89AEav/4-agent-ucc-listener-refactor)

tomereli:
[Pass test 4.2.1 POC](https://trello.com/c/XfJsCVb7/4-pass-test-421-poc)

tomereli:
[Pass test 4.2.1 POC](https://trello.com/c/XfJsCVb7/4-pass-test-421-poc)

tomereli:
[mrouter refactoring](https://trello.com/c/6CLMwoNb/6-mrouter-refactoring)

tomereli:
[agent ucc listener refactor](https://trello.com/c/paMpixDg/7-agent-ucc-listener-refactor)

tomereli:
[add vap list update for backhaul](https://trello.com/c/E424Xwp2/8-add-vap-list-update-for-backhaul)

tomereli:
DO NOT MERGE - this should have been a draft PR. We will keep it until the UCC issues this PR addresses are merged in mainline.

tomereli:
Superceeded by #636

Issue #346, opened 2019-10-09
[TASK] AP configuration on Turris Omnia / RAX40 OpenWrt


arnout:
## Description

The agent parses the WSC M2 message, but it only logs the configuration. The configuration should also be applied to the system. bpl functions exist for this, but they probably need some refactoring.

## Exit criteria

- Agent WSC M2 handling calls bwl functions to add APs and set credentials.
- Dummy and OpenWrt implementations of these exist.
- `test_flows.sh` is updated to check for bpl call instead of arrival of M2.
- Manually validated on Turris Omnia / RAX40

ghost:
@YuriyMasechko could you please evaluate what else needs to be done for this task?

YuriyMasechko:
There is only dummy implementation of bwl.
So first of all need to implement bwl and then update test_flows.sh.

tomereli:
Already done

Issue #343, opened 2019-10-09
[TASK] Test flow for 4.2.1 Multi-AP Ethernet Onboarding and Initialization


arnout:
## Description

Implement a test flow script that can be executed in docker to validate 4.2.1 Multi-AP Ethernet Onboarding and Initialization.

Example PASS logs with prplMesh attached - [4.2.1_PASS.zip](https://github.com/prplfoundation/prplMesh/files/4281497/4.2.1_PASS.zip)

arnout:
@tomereli wrote in https://github.com/prplfoundation/prplMesh/pull/533#discussion_r354843238

> Perhaps we can store the credentials in some class member, and use it in refresh_radio_info which is sent to the controller. This way even in Dummy mode in Linux containers we will see the credentials in bml_conn_map.
> Based on this, we can add/update the corresponding test flow to verify that the correct VAP credentials were applied.

tomereli:
[4.2.1_PASS.zip](https://github.com/prplfoundation/prplMesh/files/4281497/4.2.1_PASS.zip)

tomereli:
This is no longer needed since 4.2.1 test is covered by channel selection and autoconfig tests in test_flows.sh and test_flows.py, hence closing.

PR #533, opened 2019-12-03
Feature/ap configuration from m2

YuriyMasechko:
Agent part of #346 .

Bwl was added as a stub. Need to implement bwls separately and modify test_flows.sh accordingly.


common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -184,6 +184,23 @@ cACTION_APMANAGER_STEERING_EVENT_AUTH_FAIL_NOTIFICATION:
...
+  pass:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_PASS_MAX_LENGTH" ]
+  auth_type:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH" ]
+  enc_type:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH" ]
+  is_backhaul: uint8_t
```
arnout: A BSS has 3 possible modes: fronthaul, backhaul, or backhaul+fronthaul combined. So this should be either:
- a bitmap with fronthaul and backhaul bits;
- an enum with 3 values;
- two boolean fields.

For the first two options, the name is_backhaul is not appropriate. In hostapd, it's an enum with 4 values (because there's also the "not EasyMesh" case), and it's called "multi_ap".

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -184,6 +184,23 @@ cACTION_APMANAGER_STEERING_EVENT_AUTH_FAIL_NOTIFICATION:
...
+    _length: [ "beerocks::message::WIFI_SSID_MAX_LENGTH" ]
+  pass:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_PASS_MAX_LENGTH" ]
+  auth_type:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH" ]
+  enc_type:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH" ]
```
arnout: These two should be enums instead of freeform strings.

I would even use `WSC::eWscAuth` directly - @tomereli @morantr  what do you think?

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -184,6 +184,23 @@ cACTION_APMANAGER_STEERING_EVENT_AUTH_FAIL_NOTIFICATION:
...
+  is_backhaul: uint8_t
+  teardown: uint8_t
```
arnout: Teardown should be a separate action. If we receive a teardown, that should be the only M2 we receive; anything else is an error.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -184,6 +184,23 @@ cACTION_APMANAGER_STEERING_EVENT_AUTH_FAIL_NOTIFICATION:
   _type: class
   params: sSteeringEvAuthFail
 
+cACTION_APMANAGER_WIFI_CREDENTIALS_SET_REQUEST:
+  _type: class
+  ssid:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_SSID_MAX_LENGTH" ]
```
arnout: Maybe it was implemented like that before, but this is wrong: the length should be dynamic. So we should have a separate ssid_length field:
```
  ssid_length:
    _type: uint8_t
  ssid:
    _type: char
    _length: [ ssid_length ]
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -184,6 +184,23 @@ cACTION_APMANAGER_STEERING_EVENT_AUTH_FAIL_NOTIFICATION:
...
+  pass:
+    _type: char    
+    _length: [ "beerocks::message::WIFI_PASS_MAX_LENGTH" ]
```
arnout: Same here, length should be dynamic.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4622,6 +4662,32 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
                    << ", authentication_type: " << std::hex << int(authtype)
                    << ", encryption_type: " << int(enctype) << (fronthaul ? " fronthaul" : "")
                    << (backhaul ? " backhaul" : "") << (teardown ? " teardown" : "");
+
+        if (!fronthaul && !backhaul && !teardown)
+            continue;
```
arnout: This should be logged as an error. But the `continue` part is good.

In addition, I would also add a check that there's either a single teardown, or one or more front/backhaul M2s. So it should not be allowed that there is no M2 at all, or that a teardown is combined with anything else. Of course, the question is what to do in that situation: tear down? partially configure? I would say to do a teardown. So the check is simple: if teardown is true, then m2_list should have size 1.

arnout (changes requested):
It would be useful if there was also an action "reset_wifi_credentials" before the loop over the M2s, and "apply_wifi_credentials" afterwards, because we want to apply all credentials in one shot. Actually, it would be even better if all credentials were passed in a single action, and the bwl call has a vector of credentials structs.

Actually, we already have bss_info_conf_t in db.h, so maybe we could move that (except for the operating_class list) to some common header and use it in bwl as well.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
morantr: But add property `_length_var: True` to the `ssid length`:
```
  ssid_length:
    _type: uint8_t
    _length_var: True
  ssid:
    _type: char
    _length: [ ssid_length ]
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
morantr: Agree.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
morantr: Why not using `eWscVendorExtsubelementBssType`? This will also include the teardown.

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
```diff
@@ -157,6 +157,23 @@ class ap_wlan_hal : public virtual base_wlan_hal {
...
+     * @param [in] enc Encryption type.
+     * @param [in] is_backhaul Type of VAP - backhaul STA or fronthaul BSS.
+     * @param [in] teardown Remove VAP.
+     *
+     * @return true on success or false on error.
+     */
+
+    virtual bool set_wifi_credentials(const std::string &ssid, const std::string &pass,
+                                      const std::string &auth, const std::string &enc,
+                                      uint8_t is_backhaul, uint8_t teardown) = 0;
```
morantr: As @arnout has said, `auth` and `enc` should be enums, and `is_backhauk` and `teardown` should be bitfield or enum (cpp enum is valid if its value moves between min to max value).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -69,6 +69,42 @@ static bwl::WiFiSec platform_to_bwl_security(const std::string &sec)
...
+    case WSC::eWscEncr::WSC_ENCR_WEP:
+        return "WEP";
+    case WSC::eWscEncr::WSC_ENCR_TKIP:
+        return "TKIP";
+    case WSC::eWscEncr::WSC_ENCR_AES:
+        return "AES";
+    default:
+        return "INVALID";
+    }
+}
```
morantr: I think it will better be on son_wireless_utils.

morantr (changes requested):

tomereli (changes requested):
Comments inline

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
tomereli: I don't think this is required. Check out the var length lists in M2.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
tomereli: Agree

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
tomereli: I agree but the its another set of messages:/

tomereli:
[autoconfig M2 support in agent](https://trello.com/c/hPNsRYNF/15-autoconfig-m2-support-in-agent)

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
YuriyMasechko: Changed to list of configdata.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
YuriyMasechko: Changed to list of configdata.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
YuriyMasechko: Changed to list of configdata.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: According new design error cases are ignored.

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
YuriyMasechko: Changed to bssinfo list.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Moved.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -172,6 +172,22 @@ bool ap_wlan_hal_dummy::sta_bss_steer(const std::string &mac, const std::string
...
+    std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list)
+{
+    for (auto bss_info_conf : bss_info_conf_list) {
+        LOG(DEBUG) << "Received credentials for ssid: " << bss_info_conf.ssid << " auth_type: "
+                   << son::wireless_utils::wsc_to_bwl_authentication(
+                          bss_info_conf.authentication_type)
+                   << " encr_type: "
+                   << son::wireless_utils::wsc_to_bwl_encryption(bss_info_conf.encryption_type)
+                   << " network_key: " << bss_info_conf.network_key << " bss_type: "
+                   << son::wireless_utils::wsc_to_bwl_bss_type(bss_info_conf.bss_type);
```
tomereli, line 186: Perhaps we can store the credentials in some class member, and use it in `refresh_radio_info` which is sent to the controller. This way even in Dummy mode in Linux containers we will see the credentials in `bml_conn_map`.
Based on this, we can add/update the corresponding test flow to verify that the correct VAP credentials were applied.
This can be done in a followup PR, but I think it *should* be done.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -635,6 +635,13 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
     return true;
 }
 
+bool ap_wlan_hal_dwpal::update_vap_credentials(
+    std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list)
+{
+    //TODO Implement #346
```
tomereli, line 652: Better to have a full link to the issue, in the future we might move to gitlab so perhaps the issues numbers won't be the same (nitpick)

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -811,6 +812,41 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+
+            bss_info_conf.ssid                = config_data.ssid_str();
+            bss_info_conf.authentication_type = config_data.authentication_type_attr().data;
+            bss_info_conf.encryption_type     = config_data.encryption_type_attr().data;
+            bss_info_conf.network_key         = config_data.network_key_str();
+            bss_info_conf.bss_type            = static_cast<WSC::eWscVendorExtSubelementBssType>(
+                config_data.multiap_attr().subelement_value);
+
+            bss_info_conf_list.push_back(bss_info_conf);
+        }
```
tomereli, line 848: Just wanted to say that this looks much better now, good work with the yaml :)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3577,7 +3577,7 @@ bool slave_thread::autoconfig_wsc_authenticate(std::shared_ptr<ieee1905_1::tlvWs
 bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
     std::shared_ptr<ieee1905_1::tlvWscM2> m2, uint8_t authkey[32], uint8_t keywrapkey[16],
     std::string &ssid, sMacAddr &bssid, WSC::eWscAuth &auth_type, WSC::eWscEncr &encr_type,
-    bool &backhaul, bool &fronthaul, bool &teardown)
+    bool &backhaul, bool &fronthaul, bool &teardown, std::shared_ptr<WSC::cConfigData> &credentials)
```
tomereli, line 3605: Don't we have the backhaul, fronthaul and teardown bits as part of `cConfigData`? If so, now that you pass the whole `configData` I don't see a reason to pass another 3 booleans.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3823,8 +3838,21 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
                    << (backhaul ? " backhaul" : "") << (teardown ? " teardown" : "");
+
+        if (!request->add_wifi_credentials(credentials)) {
+            LOG(ERROR) << "add_wifi_credentials() has failed!";
+            return false;
+        }
+
+        if (teardown && m2_list.size() > 1) {
+            // Teardown is not compatible with any other bit in the M2.
+            LOG(DEBUG) << "Teardwon combined with other M2, ignoring all";
```
tomereli: should be `LOG(ERROR)`

tomereli (commented):
Looks good in general, although I have some more comments I think could be useful as part of this PR.
Especially regarding the saving of the credentials in dummy mode and using them in `refresh_radio_info` so they will show in the `bml_conn_map` in the controller. This however can be done later.

Not approving yet so Arnout has a chance to review. Please fix the remaining issues and I think we should merge this today or on Sunday since this is a blocker for the autoconfig task.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
arnout: I added this comment to #343.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: > Don't we have the backhaul, fronthaul and teardown bits as part of `cConfigData`?

In fact, *all* the credentials are in cConfigData; we don't need ssid, bssid, auth_type, encr_type either. However, the backhaul, fronthaul and teardown bits are a bit difficult to retrieve since they're part of the vendor extension attribute.

I originally proposed to use cConfigData, but I'm thinking that maybe we should define a new, cleaner structure instead...

Anyway, I'm OK with using cConfigData for now.

arnout: Just to clarify: I agree with @tomereli that it doesn't make sense to return anything other than `credentials`.

Another thing, however: instead of a shared_ptr, it should be a reference. A reference makes it clear that the function will not keep the object alive after it returns. Cfr. [CppCoreGuildeline R.30](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rr-smartptrparam).

arnout (approved):
Excellent splitting into commits! Every commit is easy to review on its own, and the tests succeed after every commit.

I agree with @tomereli's comments and I added one small additional comment. So I'm approving already.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: Removed ssid, bssid, auth_type, encr_type. Backhaul, fronthaul and teardown left as is because they should be parsed as bit map.
It should be shared_ptr because inside tlvf it is shared ptr. Could change yaml to have it as reference.

agent/src/beerocks/slave/son_slave_thread.cpp

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -241,10 +241,8 @@ class slave_thread : public beerocks::socket_thread {
                                        uint8_t authkey[32], uint8_t keywrapkey[16]);
     bool autoconfig_wsc_parse_m2_encrypted_settings(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
                                                     uint8_t authkey[32], uint8_t keywrapkey[16],
-                                                    std::string &ssid, sMacAddr &bssid,
-                                                    WSC::eWscAuth &auth_type,
-                                                    WSC::eWscEncr &encr_type, bool &backhaul,
-                                                    bool &fronthaul, bool &teardown);
+                                                    bool &backhaul, bool &fronthaul, bool &teardown,
```
tomereli: I'm confused, didn't we talk about removing the backhaul fronthaul and teardown booleans since they are part of the configData?

tomereli (approved):
LGTM

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -173,6 +173,22 @@ bool ap_wlan_hal_dummy::sta_bss_steer(const std::string &mac, const std::string
     return true;
 }
 
+bool ap_wlan_hal_dummy::update_vap_credentials(
+    std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list)
+{
+    for (auto bss_info_conf : bss_info_conf_list) {
```
morantr, line 179: please use const reference
`const auto &bss_info_conf`

I guess you can use const on the function signature ass well.

morantr (approved):
Small comment, nothing important.

tomereli:
Please rebase and merge

Issue #471, opened 2019-11-04
[TASK] TLVF add WSC attributes support


tomereli:
## Description

Currently, TLVF supports only basic types of TLVs, where TLV data can be either a simple type (uint8_t, uint16_t, etc,) or a list (variable known/unknown length).

The WSC specification defines attributes that can be added to TLVs where each attribute includes a TLV like structure (Type, Length, Value) - the difference being the Type which is 16 bits instead of 8 bits as in regular 1905/mAP TLVs.

Due to the insufficiant support for WSC attributes inside TLVs, the whole M1 and M2 implementation is wrong - it defines two different yaml files with the same TLV type, and all attributes defined in the order they are defined in the WSC specification.
This is wrong since the attributes can come in any order, and parsing of such TLVs should take this into account.

### Suggested solution
In order to support attributes in any order, the M2 and M1 yamls (tlvWscM1.yaml and tlvWscM2.yaml) should be removed, and instead each attribute should be defined in a separate yaml file.

**sub tasks**
- [x] Automatic length calculation for adding attributes - when `addClass` is called for a WSC Attribute, the TLV length should be updated along with the Attribute length
- [x] "Optional" support - WSC Attributes should be optional, without the hack of using unknown length lists
- [x] Automatic parser for WSC attributes - see #318

tomereli:
![](https://github.trello.services/images/mini-trello-icon.png) [tlvList refactoring](https://trello.com/c/JHrwp6IN/19-tlvlist-refactoring)

Issue #300, opened 2019-09-16
Topology discovery full support


arnout:
## Description

For certification, it is sufficient to only support part of topology discovery (cfr. #120, #186). For a fully operational controller, however, it is needed to fully support topology discovery at least in the controller. It is also required in the agent if IEEE 1905.1 compliance is desired.

- [x] #120 Topology discovery -  Controller certification
- [x] #186 5.3.1 Supported discovery
- [x] #525 Construct topology response message
- [x] #830 Make a design for complete topology discovery support
- [x] #1160 per-interface Topology Discovery message
- [x] #1425 Extend topology database to store neighbor's interface and MAC address 
- [x] #1428 Use neighbor's interface info in Topology Response 
- [ ] #701 Add Device Bridging Capability TLV to Topology Response message 
- [ ] #925 Detect topology changes on the network 

## Exit criteria

- Topology agent (framework/discovery_agent/) is built
- Topology agent is started by prplmesh_utils.sh
- Topology agent sends a topology notification every 60 seconds
- Topology agent listens for topology query and constructs topology response (#525)

### Problem statement

Both the agent and the controller, when operating independently, must perform the topology discovery functions. Therefore, it is not simply possible to implement this functionality in one of them.

Both the agent and the controller need to know some of the topology information. Mainly the controller needs it to discover stations connecting and disconnecting.

### Possible solutions

1. The functionality is provided by a library in common or framework and linked with both controller and agent. If a device has both agent and controller, the agent signals its presence to the controller and the controller disables its topology functionality.
2. An additional component takes care of the topology discovery. This component is always running (much like the ieee1905_transport component), in agent, controller, and agent+controller mode.

The second option is conceptually cleaner so that one is chosen.

### Components affected

A topology discovery agent already exists in framework/discovery_agent, but it is currently not built and probably broken. So that has to be revived.

The discovery is responsible for sending out notifications regularly. It will also be responsible for most of the rest of the IEEE1905.1 functionality, though that doesn't need to be implemented for EasyMesh certification.

Topology information is needed by agent and controller as well. The topology discovery agent will manage the protocol messages (e.g. sending out regular topology discovery messages and responding with topology query). The other components will keep a local topology database and listen to the relevant CMDUs (e.g. responses to the topology queries sent out by the topology discovery agent) to update that database.

Agent and controller may also emit topology related messages autonomously. For example, the agent emits a topology notification message when a station connects or disconnects. That functionality will stay in the agent.

[...]

PR #538, opened 2019-12-05
framework: tag possible issues found while reviewing source code

mariomaz:
I have recently joined the prplMesh project as a contributor. While
reviewing source code to get familiar with it, I have found what could
possibly be some bugs.

I have created a personal development branch and added a comment close
to each such possible issue to describe it. This way it is easy to read
through all of them and decide, case by case, if they are really issues
or just rookie missunderstandings.


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -66,6 +66,10 @@ const char *main_thread::s_arrStates[] = {FOREACH_STATE(GENERATE_STRING)};
 /////////////////////////////// Implementation ///////////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
+/**
+ * MMZ: Eclipse complains about uninitialized member variables.
+ * Should use CppCheck or some other static analysis tool to check code.
+ */
```
arnout, line 72: #539

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -585,6 +589,14 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
...
 
             // If a wired (WAN) interface was provided, try it first, check if the interface is UP
+            /**
+             * MMZ
+             * When the wired connection uses G.hn, link quality has to be
+             * considered to decide if such wired connection must be used or not
+             * instead of the wireless one.
+             * Moreover, such link quality may vary over time so it should be
+             * regularly checked to select the best backhaul link available.
+             */
```
arnout, line 599: As briefly discussed in the arch call, backhaul link management is out of scope for prplMesh at the moment. It is true that it's a feature that is really needed, but we'll first concentrate on getting certification and a minimal usable feature set for a product. That said, we (i.e., you :-) could/should probably already create a wiki page to track our understanding of this problem.

The current backhaul manager FSM needs serious review, because it probably still contains a lot of leftovers from "active mode" (where prplMesh was responsible for bringing up the backhaul link). We decided to remove that active mode, and instead require that the backhaul link(s) are configured up front by the platform. prplMesh responsibility limits itself to steering (possibly to another radio) and enabling/disabling a link (which would be the way for doing wired/wireless link management).

Note that EasyMesh doesn't have anything to support backhaul link selection by the controller. There isn't even a way for the controller to get beacon RSSI values from a backhaul STA, except through 11k and Beacon Metrics Query. However, it may be possible to use 1905.1a messages for backhaul link selection by the controller.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -679,6 +691,9 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
         }
         break;
     }
+    /**
+     * MMZ this is a transient state and thus could be removed
+     */
```
arnout, line 696: :+1: 

#540

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -696,6 +711,10 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
         LOG(DEBUG) << "Sent notifications, goto OPERATIONAL state";
 
         eth_link_poll_timer = std::chrono::steady_clock::now();
+        /**
+         * MMZ  m_sConfig.wire_iface could be empty or m_sConfig.wired_backhaul could
+         * be false. iface status should not be checked in either case.
+         */
```
arnout, line 717: Well, it's not really harmful either, the function will just return `false`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -707,11 +726,27 @@ bool main_thread::backhaul_fsm_main(bool &skip_select)
...
+            /**
+             * MMZ  m_sConfig.wire_iface could be empty or m_sConfig.wired_backhaul could
+             * be false. iface status should not be checked in either case.
+             */
             bool eth_link_up = network_utils::linux_iface_is_up_and_running(m_sConfig.wire_iface);
             if (eth_link_up != m_eth_link_up) {
+              /**
+               * MMZ  there is no need to compute m_sConfig.wire_iface status back again
+               * as the result is already in eth_link_up variable
+               */
```
arnout, line 749: For such small improvements, it's better that you immediately create a commit/PR that fixes it instead of going through this process.

FYI, the commit message would be:

```
agent: backhaul_manager: remove redundant iface status check

There is no need to compute m_sConfig.wire_iface status back again
as the result is already in eth_link_up variable.

Signed-off-by: ...
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1686,6 +1721,10 @@ bool main_thread::handle_1905_autoconfiguration_response(ieee1905_1::CmduMessage
         return true;
     }
 
+    /**
+     * MMZ maybe supported service TLV should be checked prior to assigning
+     * controller_bridge_mac variable
+     */
```
arnout, line 1727: It's a bit more complicated than that. If there are two controllers/registrars replying to the autoconfig search, we have a problem anyway, even if one of them is Multi-AP and the other one is not.

I created #541 instead.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1703,6 +1742,9 @@ bool main_thread::handle_1905_autoconfiguration_response(ieee1905_1::CmduMessage
                                    << soc->hostap_iface;
                         LOG(DEBUG) << FSM_CURR_STATE_STR;
                         soc->controller_discovered = true;
+                        /**
+                         * MMZ there is no need to assign variable again
+                         */
```
arnout, line 1747: Please create a hotfix PR for this.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1714,6 +1756,9 @@ bool main_thread::handle_1905_autoconfiguration_response(ieee1905_1::CmduMessage
                 }
             }
         } else if (tlvSupportedFreqBand->value() == ieee1905_1::tlvSupportedFreqBand::BAND_5G) {
+          /**
+           * MMZ this code is duplicated except for the frequency value
+           */
```
arnout, line 1761: Please create a hotfix PR for this.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1723,6 +1768,9 @@ bool main_thread::handle_1905_autoconfiguration_response(ieee1905_1::CmduMessage
                                    << soc->hostap_iface;
                         LOG(DEBUG) << FSM_CURR_STATE_STR;
                         soc->controller_discovered = true;
+                        /**
+                         * MMZ there is no need to assign variable again
+                         */
```
arnout, line 1773: Please create a hotfix PR for this.

agent/src/beerocks/slave/platform_manager/bpl_loader.c
```diff
@@ -19,6 +19,16 @@
...
 //////////////////////////////////////////////////////////////////////////////
 
+/**
+ * MMZ:
+ * It is not necessary to dynamically load this library during execution since
+ * it is available during compile/link phase.
+ *
+ * Dynamic load is suitable/required for plug-in libraries but I think it adds an
+ * unnecessary overhead in this case. Moreover, code readability and
+ * maintainability are more difficult.
```
arnout, line 29: I fully agree. I thought that I had mentioned this in #469 but apparently not. Added there now.

common/beerocks/bcl/include/bcl/beerocks_config_file.h
```diff
@@ -29,6 +29,16 @@ class config_file {
...
 
+    /**
+     * MMZ: Master and slave configuration structures as well as methods to
+     * read configuration file and fill in such structures should not be defined
+     * in this class but in their respective applications.
+     * If we had to modify master or slave configuration or create a brand new
+     * application, then we would have to modify this class and recompile all
+     * the applications that use it.
+     * For the sake of cleanliness and reusability, this module should not know
+     * anything about the modules that depend on it.
```
arnout, line 40: Indeed. I've created #542 which covers this and a whole lot more.

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -36,6 +36,13 @@ namespace beerocks {
 
 namespace message {
 
+  /**
+   * MMZ: It is not clear from the name of the consts if there is room for the
+   * null termination char or not. The suffix SIZE should be used to mean that
+   * it is a buffer size, including termination char. The suffix LENGTH or
+   * MAX_LENGTH should be used for values that can be compared to the strlen()
+   * return value, that is, not including a termination char.
```
arnout, line 44: Good point. Feel free to create a PR that improves this.

Ideally, however, this really should be a maximum length, and the actual length should be encoded in the messages.

Even more ideal would be a string type in tlvf for a dynamic-length string with a 1-byte length, and that can be read and written as std::string.

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -49,6 +49,20 @@ template <class T> class promise {
...
+         * specified a nanosecond value less than zero or greater than or equal
+         * to 1000 million.
+         * The clock_gettime system call has higher precision than its successor
+         * the gettimeofday(). Decide on whether to use it with CLOCK_REALTIME
+         * or CLOCK_MONOTONIC. If CLOCK_MONOTONIC is used, condition variable
+         * attributes must then be initialized accordingly.
+         *
+         * Alternatively and maybe better off, use std::mutex and
+         * std::condition_variable, where wait_for() method accepts a duration
+         * parameter.
```
arnout, line 64: Or even better, throw this thing away and replace it with std::promise and std::future. @vitalybu do you know why this wasn't done?

common/beerocks/bcl/include/bcl/beerocks_ucc_listener.h
```diff
@@ -116,6 +116,10 @@ class tlvPrefilledData : public BaseClass {
     {
         m_init_succeeded = true;
     };
+    /**
+     * MMZ: Destructor should be virtual because this class has virtual method(s)
+     * (detected by static source code analyzer in Eclipse)
```
arnout, line 121: Please create a hotfix PR for this.

common/beerocks/bcl/include/bcl/network/net_struct.h
```diff
@@ -15,6 +15,11 @@
 
 #include <tlvf/common/sMacAddr.h>
 
+/**
+ * MMZ for coherence with other structs in this same file, operators == and !=
+ * should be defined inside sMacAddr struct (which is autogenerated code).
+ * Alternatively, move other structs so they are also autogenerated as sMacAddr
```
arnout, line 21: This is historicaly accident, because:

- we needed this operator for sMacAddr;
- we were too lazy to update tlvf to generate operator== for all structs (it's not so simple, and it is not needed for most of them);
- we needed a place for this operator, and we didn't want to create a new header file just for that.

common/beerocks/bcl/source/beerocks_os_utils.cpp
```diff
@@ -97,6 +97,14 @@ void os_utils::kill_pid(std::string path, std::string file_name)
 
 bool os_utils::is_pid_running(std::string path, std::string file_name, int *pid_out)
 {
+  /**
+   * MMZ: a function called get_pid_file_name() should be created and used all
+   * over the project instead of using "pid/" so if path to file ever changes
+   * then nothing will break
```
arnout, line 103: IMO this whole process management thing should be revisited and either rely on the system's process manager (procd, systemd) or be implemented as a single process manager to manages all of the subprocesses. That process manager could do things the right way and store PIDs in memory and use wait/waitpid() instead of relying on race-ridden PID files.

Actually, I'd prefer if prplMesh were a single multicall binary and it would plain fork() internally, instead of relying on fork-exec or the horrible system().

common/beerocks/bcl/source/beerocks_os_utils.cpp
```diff
@@ -188,6 +199,13 @@ int os_utils::redirect_console_std(std::string log_file_name)
 
 void os_utils::close_file(int fd)
 {
+  /**
+   * MMZ: All non-zero integers cast to true so if descriptor were -1, then it
+   * would also try to be closed.
+   * Since we are not interested in return code from close() we could call it
+   * directly, without previously checking file descriptor's value (probably it
+   * is already being being checked inside the close() function).
```
arnout, line 207: I agree that close_file is useless. Please create a hotfix PR for this.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -53,6 +53,11 @@ bool socket_thread::init()
     if (unix_socket_path.empty()) {
         int port = server_port();
         if (port > 0) {
+          /**
+           * MMZ: This server will listen for connection requests in all
+           * available interfaces ¿was this done on purpose or shall we bind
+           * server socket to a particular interface?
```
arnout, line 59: This is only used in the CLI and the certification agent, both of which are meant for external access, so yes this is intentional.

(Never mind the use in mrouter; @morantr will remove that soon.)

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -153,6 +158,9 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     size_t message_size             = uds_header->length + sizeof(message::sUdsHeader);
 
     // Try to read all message
+    // MMZ: Maybe not all bytes in message are ready to be read at this moment
+    // and that's not necessarily an error as those pending bytes would be
+    // eventually received if iterating.
```
arnout, line 163: To be honest, there's a lot more wrong with this code... For example, if the length is larger than the buffer, no recv() is done, so the bytes aren't abandoned, so you've essentially DoS'ed the socket...

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -312,6 +320,11 @@ bool socket_thread::work()
                     break;
                 }
 
+                /**
+                 * MMZ: contents of rx_buffer are undetermined when calling
+                 * custom_message_handler() and not used here after the function
+                 * returns. So rx_buffer should not be an argument of custom_message_handler
```
arnout, line 326: The rx buffer is just used as buffer space in custom_message_handler, to avoid the need for allocating a separate buffer there. The contents are immediately overwritten by a call to readBytes.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -512,6 +512,9 @@ bool beerocks_ucc_listener::socket_disconnected(Socket *sd)
     LOG(DEBUG) << "Client has disconnected";
     if (sd != m_ucc_sd) {
         LOG(DEBUG) << "Unfamiliar socket has disconnected";
+        /**
+         * MMZ redundant code
```
arnout, line 516: Please create a hotfix PR for this.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -534,7 +537,23 @@ bool beerocks_ucc_listener::custom_message_handler(Socket *sd, uint8_t *rx_buffe
...
+     * remaining bytes, then those bytes should be discarded as an incomplete command
+     */
+    /**
+     * MMZ readBytes could return two or more commands in a row and they would
+     * be treated as if just one command had been received, others being ignored
+     */
     auto available_bytes       = sd->readBytes(rx_buffer, rx_buffer_size, true);
+    /**
+     * MMZ if available_bytes == rx_buffer_size, then end char is written past
+     * the end of buffer
```
arnout, line 555: Please create a hotfix PR for this.

common/beerocks/bcl/source/network/socket.cpp
```diff
@@ -31,8 +31,15 @@ typedef int socklen_t;
 
 int Socket::m_ref = 0;
 
+/**
+ * MMZ: The readTimeout type should be unsigned to disallow negative values with
+ * the method signature itself
```
arnout, line 36: Please create a hotfix PR for this.

common/beerocks/bcl/source/network/socket.cpp
```diff
@@ -31,8 +31,15 @@ typedef int socklen_t;
...
+ */
 Socket::Socket(const std::string &uds_path, long readTimeout)
 {
+  /**
+   * MMZ: Windows code should be removed for better readability
```
arnout, line 41: #544

common/beerocks/bcl/source/network/socket.cpp
```diff
@@ -134,6 +141,9 @@ bool Socket::setReadTimeout(long msec)
 #else
     long sec  = 0;
     long usec = 0;
+    /**
+     * MMZ Should be != instead of ==
```
arnout, line 145: Please create a hotfix PR for this. Actually the condition is redundant.

I wonder how we didn't notice before...

common/beerocks/bcl/source/network/socket.cpp
```diff
@@ -182,6 +192,8 @@ ssize_t Socket::readBytes(uint8_t *buf, size_t buf_size, bool blocking, size_t b
             buf_len = buf_size;
     }
 
+    // MMZ: Comparison might be true if there were two or more messages ready to
+    // be read, one after the other. This case should not be considered an error
```
arnout, line 196: No, if buf_len is set, then it is the size of one message, so it should never be bigger than the buffer size. If buf_len is not set, it will be set equal to buf_size above so the condition will not be true.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -534,7 +537,23 @@ bool beerocks_ucc_listener::custom_message_handler(Socket *sd, uint8_t *rx_buffe
...
+     * MMZ readBytes might return an incomplete command without that being an
+     * error. Remaining bytes could successfully be read in next call to readBytes.
+     * New line character should be used to know if received string contains one
+     * full command or, on the contrary, some bytes are yet to be read.
+     * If bytes were read without a new line ending char and a timeout expires before receiving
+     * remaining bytes, then those bytes should be discarded as an incomplete command
+     */
+    /**
+     * MMZ readBytes could return two or more commands in a row and they would
+     * be treated as if just one command had been received, others being ignored
```
arnout, line 550: In practice, no, because UCC doesn't send another command until a reply is received for the previous one.

common/beerocks/btl/btl_local_bus.cpp
```diff
@@ -80,6 +80,16 @@ bool transport_socket_thread::bus_subscribe(const std::vector<ieee1905_1::eMessa
...
     for_each(msg_types.begin(), msg_types.end(), [&](const ieee1905_1::eMessageType msg_type) {
+        /**
+         * MMZ
+         * What is the actual intention here? Subscribe to as many topics as
+         * possible or stop subscribing on first error?
+         * Bitwise AND never does short-circuit so subscribe_topic_to_bus is
+         * invoked on each iteration, no matter the result.
+         * A logical AND would do short-circuit and thus it would be more
+         * efficient if the intention is to stop on error:
+         * ret = ret && ...
```
arnout, line 91: Please create a hotfix PR for this. Indeed, if one of them fails we're dead anyway so it doesn't make sense to continue.

Actually, there's no real reason to use `for_each` here, a ranged for loop with a `return false` on failure  would be simpler.

common/beerocks/btl/btl_uds.cpp
```diff
@@ -20,13 +20,16 @@ bool transport_socket_thread::configure_ieee1905_transport_interfaces(
     return true;
 }
 
+// MMZ No need to reimplement this method because behavior is the same
```
arnout, line 23: Please create a hotfix PR for this.

common/beerocks/btl/btl_uds.cpp
```diff
@@ -20,13 +20,16 @@ bool transport_socket_thread::configure_ieee1905_transport_interfaces(
...
 void transport_socket_thread::add_socket(Socket *s, bool add_to_vector)
 {
     socket_thread::add_socket(s, add_to_vector);
 }
 
+// MMZ No need to reimplement this method because behavior is the same
```
arnout, line 29: Please create a hotfix PR for this.

common/beerocks/btl/btl_uds.cpp
```diff
@@ -20,13 +20,16 @@ bool transport_socket_thread::configure_ieee1905_transport_interfaces(
...
 void transport_socket_thread::remove_socket(Socket *s) { socket_thread::remove_socket(s); }
 
+// MMZ No need to reimplement this method because behavior is the same
```
arnout, line 32: Please create a hotfix PR for this.

common/beerocks/btl/btl_uds.cpp
```diff
@@ -100,6 +103,9 @@ bool transport_socket_thread::from_bus(Socket *sd) { return sd == bus; }
 
 bool transport_socket_thread::work()
 {
+    /**
+     * MMZ: This method is copy&paste of socket_thread::work() (the one at base class)
```
arnout, line 107: Please create a hotfix PR for this.

common/beerocks/btl/include/btl/btl.h
```diff
@@ -54,6 +54,7 @@ class transport_socket_thread : public socket_thread {
 
 private:
     void bus_init();
+    // MMZ: cmdu parameter should be const&
```
arnout, line 57: Please create a hotfix PR for this. It will require making come methods of CmduMessage const as well though.

common/beerocks/tlvf/src/src/beerocks_message.cpp
```diff
@@ -137,6 +137,13 @@ bool message_com::send_cmdu(Socket *sd, ieee1905_1::CmduMessageTx &cmdu_tx,
         return false;
     }
 
+    /**
+     * MMZ
+     * The cost of byte reversal on modern processors is normally negligible.
+     * IMHO, Network Byte Order should always be used for the sake of having a
+     * simpler code (swap_needed could be removed as an option and finalize()
+     * and parse() methods would always swap bytes)
```
arnout, line 145: :+1: 

Actually, I think the swapping should even be done immediately while writing the values. But that's a bit a bigger change... #547

framework/common/include/mapf/common/socket.h
```diff
@@ -74,24 +74,50 @@ class SubSocket : public Socket {
         return Subscribe(topic);
     }
     int Subscribe(const std::initializer_list<std::string> &topics);
+    /**
+     * MMZ
+     * Parameters that are not modified inside the method should be passed as
+     * const reference
```
arnout, line 80: Please create a hotfix PR for this.

framework/common/include/mapf/common/socket.h
```diff
@@ -74,24 +74,50 @@ class SubSocket : public Socket {
...
         for (const auto &topic : topics)
+          /**
+           * MMZ
+           * Method return value is ignored
+           */
             Subscribe<T>(topic);
+        /**
+         * MMZ
+         * Method should return a value
+         */
```
arnout, line 94: This function is actually unused. I'm surprised though that gcc doesn't warn about it (and -Werror is turned on...)

Please create a hotfix PR for this.

framework/common/include/mapf/common/socket.h
```diff
@@ -74,24 +74,50 @@ class SubSocket : public Socket {
...
+        /**
+         * MMZ
+         * Method returns on first iteration
+         */
         for (const auto &topic : topics)
             return Unsubscribe<T>(topic);
+        /**
+         * MMZ
+         * Method should return a value
+         */
```
arnout, line 120: Also unused.

framework/discovery_agent/include/mapf/topology_discovery_agent/topology_discovery_agent.h
```diff
@@ -159,6 +159,9 @@ class DiscAgent : public TopologyDiscoveryAgentLocalDB {
         } bss_list[MAX_BSS];
     };
 
+    /**
+     * MMZ This structure is defined several times in the project
```
arnout, line 163: topology_discovery is not built at the moment and hasn't been kept up to date.

These comments are definitely very helpful for reviving topology_discovery:
#300

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -52,6 +52,10 @@ bool CmduMessageRx::parse(uint8_t *buff, size_t buff_len, bool swap_needed, bool
     if (!parse_tlvs)
         return true;
 
+    /**
+     * MMZ
+     * Why using a list of parsers? It seems like just one parser should suffice
```
arnout, line 57: I had the same feedback but the PR got merged regardless :-)

Should be fixed soon by @CoralMalachi and @Galvien as part of #318 #471 #160

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvDeviceInformation.yaml
```diff
@@ -21,5 +21,8 @@ sInfo:
   _type: struct
   mac: sMacAddr
   media_type: eMediaType
+# MMZ Media-specific information
+# 1 octet: Number of octets in ensuing field
+# n octets: Media-specific information of the local interface (as defined in the "Media-specific information" column of Table 6-12.
```
arnout, line 26: Many of the IEEE1905.1 structures are currently unused. Some of them will be revived and corrected as part of #300.

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport_messages.h
```diff
@@ -151,6 +151,10 @@ class CmduRxMessage : public CmduXxMessage {
         return build_topic(kTopicPrefix, ETH_P_1905_1, msg_type);
     }
 
+    /**
+     * MMZ: LLDP packets are not IEEE 1905.1 CMDUs (maybe base class should be
+     * renamed to avoid confusion)
```
arnout, line 156: I think @CoralMalachi will push a PR soon that fixes this.

arnout (changes requested):
I have tried to respond to all of the comments. I've created a lot of new issues to take them into account.

A lot of the comments are small fixes. For such things, just create a hotfix PR that fixes them.

Thank you for this thorough review.

I'm marking the PR as request changes since it's obviously not meant for merging.

common/beerocks/bcl/include/bcl/beerocks_promise.h
vitalybu: It was created to workaround a bug in some older GCC compiler (probably 4.7-4.8).
The standard future/promise implementations caused a deadlock in some cases.

common/beerocks/bcl/source/beerocks_os_utils.cpp
vitalybu: 👍

common/beerocks/bcl/source/beerocks_socket_thread.cpp
tomereli: True - we should check if the message_size is smaller than the size of the buffer. But other than that (edge case) - its a blocking read, we know from the UDS header *exactly* how many bytes are there to read, so we try to read all of them.
I don't see an issue here.. @vitalybu , your thoughts?

vitalybu: The socket is not read only when the number of available bytes is less than the size of the UDS header.
`recv()` is always called inside `readBytes()` (data is truncated to the buffer size).
Not sure I understand the issue here...

common/beerocks/bcl/include/bcl/beerocks_promise.h
arnout: Could we require GCC 4.9+ and change it to std::future/promise then?

tomereli: I think we can, UGW8.x and RDKB toolchains both >4.9 so I don't see an issue with that.

vitalybu: Yup

common/beerocks/bcl/source/beerocks_socket_thread.cpp
arnout: > its a blocking read, we know from the UDS header _exactly_ how many bytes are there to read, so we try to read all of them.

No we don't. There's only a single recv(). If the full packet has not been received yet, recv() will return with a length that is smaller than the full packet. So the recv() should be continued in a loop until the full packet is received.

However, in practice, this cannot happen:

- UDS on Linux will transport everything passed to send() atomically (at least up to 4K, but I think even it is bigger), and we always send the full CMDU with one send();
- IEEE1905 and LLDP frames from the network fit in a single Ethernet packet, and that is passed up to recv() in a single chunk.

PR #543, opened 2019-12-05
Feature/implement agent topology response

Galvien:
## Overview

The last step of test 4.2.1 in the MAUT EasyMesh certification test plan is the Topology Response sent by the MAUT to the controller, this change implements the flow, i.e. sending a topology response message to the controller upon receiving a topology query from it.
Currently the message's contents are mainly filled with dummy data, to be updated later by real platform data.

## Changes

- Update TLVf to generate __AP Operational BSS__ and __Associated Clients__ TLVs
- Add topology _query_ __RX__ handling
- Add topology _reponse_ __TX__ creation
- Send the topology response to bus

## Required future changes and related changes
- Adding real platform data to the response message
- Adding agent test flows, as the current flows are controller-based.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1770,7 +1770,12 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
         return false;
     }
 
-    if (!tlvSupportedService->alloc_supported_service_list()) {
+    size_t number_of_supported_services = 1;
+    if (local_master) {
+        number_of_supported_services++;
+    }
+
+    if (!tlvSupportedService->alloc_supported_service_list(number_of_supported_services)) {
         LOG(ERROR) << "alloc_supported_service_list failed";
         return false;
     }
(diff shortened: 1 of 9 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1782,7 +1787,18 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+
+    if (local_master) {
+        auto supportedServiceTuple = tlvSupportedService->supported_service_list(1);
+        if (!std::get<0>(supportedServiceTuple)) {
+            LOG(ERROR) << "Failed accessing supported_service_list";
+            return false;
+        }
+
+        std::get<1>(supportedServiceTuple) =
+            wfa_map::tlvSupportedService::eSupportedService::MULTI_AP_CONTROLLER;
```
tomereli, line 1800: Should be `wfa_map::tlvSupportedService::eSupportedService::MULTI_AP_CONTROLLER | wfa_map::tlvSupportedService::eSupportedService::MULTI_AP_AGENT` since in local GW we are also running local agent.

tomereli (changes requested):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: I think there must be one item in the list for each supported role, it is not a bitmask.
The case that is not supported is that in which the only supported role is 'controller'

tomereli: Of course, you're right. So in the case we are local_gw we add another supported role which is controller?

tomereli (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: I don't know how to handle the case where the only supported role is controller. If I've understood it correctly, backhaul_manager_thread is created by the agent process so agent role is always in the list and, if local_master is true, then controller role is also added. Controller code should, if not running together with an agent, also respond to the topology request adding the controller role to the list of the supported roles.

mariomaz:
[BUG: Topology response missing Agent role](https://trello.com/c/DKfp0gu1/65-bug-topology-response-missing-agent-role)

arnout (approved):
There was an #if 0 added. That is never a good idea, and especially in this case because the code inside of it was there already outside the #if 0 as well. So I removed that and force-pushed.

I think we could have just added empty TLVs instead of filling them with dummy data, but this way it's more interesting to see if the certified controllers can parse it, so OK.

PR #545, opened 2019-12-05
Feature/docker pre built openwrt

rmelotte:
This PR adds a Dockerfile and scripts to cross-compile prplMesh for OpenWrt target.

Although it should be possible to use it for the rax40 platform, it's currently not supported due to wrong dependencies in the package Makefiles.
We will have to fix the existing prplWRT feeds first.


rmelotte:
DO NOT MERGE
This needs to be updated to use another prplmesh feed.

arnout:
[Build for target as part of CI](https://trello.com/c/AoN6HvdK/30-build-for-target-as-part-of-ci)

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+####
+FROM openwrt-prerequisites as openwrt-prebuilt
+
+ARG HOME
+COPY --chown=openwrt:openwrt configs/sdk_config $HOME/sdk_config
+USER openwrt
+
+ARG OPENWRT_REPOSITORY=https://git.prpl.dev/prplwrt/prplwrt.git
+ARG OPENWRT_VERSION=9d2efd
+ARG TARGET=mvebu
```
arnout: I'd like some documentation comments above the ARGs.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+WORKDIR $HOME
+
+RUN git clone "$OPENWRT_REPOSITORY" openwrt \
+    && cd openwrt \
+    && git checkout "$OPENWRT_VERSION"
+
+WORKDIR $HOME/openwrt
+
+RUN cp "${HOME}/sdk_config" .config \
```
arnout: Since sdk_config is just a oneliner, and we definitely always want that one line in the config, I'd do it with echo here instead of requiring a separate file. If it becomes more than that one line, we can still make it a file.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+    && echo "CONFIG_TARGET_${TARGET}=y" >> .config \
+    && cat .config \
```
arnout: I guess this is for debugging. However, I don't think it's that useful, it really can't be anything else than those two lines, right? Anyway, if it is kept, there should also be an echo before to tell us what we're seeing.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+    && make defconfig \
+    && make -j$(nproc) 2>&1 | tee build.log
```
arnout: Not sure if tee'ing to build.log makes a whole lot of sense, but OK.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+RUN TAR="$(find bin/targets -name 'openwrt-sdk*' -exec echo {} \;)" \
```
arnout: Why `-exec echo {} \;` instead of the normal `-print`?

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+    && mkdir -p "${HOME}/openwrt_sdk" \
+    && mv -v "$TAR" "${HOME}/openwrt_sdk" \
+    && TAR="$(basename $TAR)" \
```
arnout: Please use a different variable name, reusing the same one is just confusing.

However, all this cd and rm stuff is redundant, you can just do

```
    && mkdir -p "${HOME}/openwrt_sdk" \
    && tar -C "${HOME}/openwrt_sdk" -xf "${TAR}" --strip-components=1
```

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
+ARG HOME=/home/openwrt
```
arnout: Is it useful to make this an argument?

If yes, please document it (including how it could be used).

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+FROM openwrt-sdk as prplmesh-builder
+
+ARG HOME
+COPY --chown=openwrt:openwrt configs/prplmesh_config $HOME/prplmesh_config
+ARG PRPLMESH_VERSION=82f1e83
+ARG TARGET_PLATFORM_TYPE=turris-omnia
+ARG PRPL_FEED=https://git.prpl.dev/prplmesh/feed-prpl.git
+# Depending on the target, you may want to use dwpal instead of nl80211:
+ARG USE_DWPAL=false
```
arnout: That's what I'm talking about: document the ARGs!

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+RUN apt-get update \
+    && apt-get -y install --no-install-recommends \
+    python3 python-yaml rsync less vim libgetopt-complete-perl \
+    && rm -rf /var/lib/apt/lists/*
+    # Building tlvfs currently uses the docker
+    # container's python (not the one from OpenWrt)
```
arnout: I would put all this in the base image already. Although I wonder why less, vim and libgetopt-complete are useful..

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+USER openwrt
+
+RUN scripts/feeds update -a \
+    && scripts/feeds install -a \
+    && echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
```
arnout: If you do this at the beginning, you don't need the additional update below.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+    && scripts/feeds update prpl \
+    && sed -ri "s/-DTARGET_PLATFORM_TYPE=[^ ]*/-DTARGET_PLATFORM_TYPE=${TARGET_PLATFORM_TYPE}/" \
+       feeds/prpl/prplmesh/Makefile \
+    && sed -ri "s/PKG_REV:=.*$/PKG_REV:=${PRPLMESH_VERSION}/" feeds/prpl/prplmesh/Makefile \
+    && scripts/feeds install -p prpl prplmesh\
```
arnout: Since this is a hopefully temporary hack, I'd do it in a separate RUN. And add a comment to explain this hack.

TBH I don't understand why the feeds install has to be done after the patching. AFAIU it should be possible to locally modify a Makefile without re-running feeds.

tools/docker/builder/openwrt/build.sh
```diff
@@ -0,0 +1,78 @@
...
+       .
+
+# Next we'll build the specified prplMesh version and copy out the ipk
+docker run -i \
+       --name "$container_name" \
+       -e TARGET \
+       -e TARGET_PLATFORM_TYPE \
+       -e OPENWRT_VERSION \
+       -e USE_DWPAL \
+       -e PRPLMESH_VERSION \
```
arnout, line 69: I don't like this, but it's OK for the time being.

What we really want is that the current source gets built inside the container, not some random commit that you must have pushed already and that you have to provide explicitly on the command line.

A first step to achieve that is pretty simple: mount the .git directory inside the container as /home/openwrt/.../prplmesh/git-src. Then we just require the current source to be checked in as HEAD.

arnout (changes requested):
Lots of small things.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Done.

tools/docker/builder/openwrt/Dockerfile
rmelotte: It's not useful anymore indeed. I removed it.

tools/docker/builder/openwrt/Dockerfile
rmelotte: No particular reason, I switched to `-print`.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Indeed, that's so much simpler. I also skipped the "rm" part entirely, since there is no real reason to delete it (from the `openwrt` dir; it won't be in `openwrt_sdk`).

tools/docker/builder/openwrt/Dockerfile
rmelotte: If we ever need to add a new build-time dependency (one that could reasonably taken from the host) or add other debugging tools, we could do so without recompiling the base image (the `openwrt-prebuilt` one takes significantly more time to build than the other ones).

`less`, `vim` et al. have been added to be able to debug failing prplMesh builds (when running the container of course, not when buiding) interactively. 
I don't remember why `libgetopt-complete` though. I'll have to check if it's still used and either remove it or add a comment.

On the other hand it's not like we can keep the same base image forever, we will have to update it from time to time anyway.

What do you think? Is it worth keeping these packages in a separate step?

arnout: > If we ever need to add a new build-time dependency (one that could reasonably taken from the host) or add other debugging tools, we could do so without recompiling the base image (the `openwrt-prebuilt` one takes significantly more time to build than the other ones).

We can still use this approach when we need that. But I think even then we'll at some point consolidate it into the base image.

> `less`, `vim` et al. have been added to be able to debug failing prplMesh builds (when running the container of course, not when buiding) interactively.

OK, good reason to keep them. Add a comment about it though.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Done.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Done.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Actually this is not needed anymore, I removed it.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Indeed it was most useful when debugging the first (successful) builds, removing.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Alright, I moved them to the base image.
NB: `libgetopt-complete-perl` was not needed anymore.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -67,30 +68,13 @@ COPY --chown=openwrt:openwrt configs/prplmesh_config $HOME/prplmesh_config
 ARG PRPLMESH_VERSION=82f1e83
 ARG TARGET_PLATFORM_TYPE=turris-omnia
 ARG PRPL_FEED=https://git.prpl.dev/prplmesh/feed-prpl.git
```
arnout: Could you add a documentation comment for these three as well?

tools/docker/builder/openwrt/configs/sdk_config
```diff
@@ -0,0 +1 @@
+CONFIG_SDK=y
```
arnout: This file should be removed now.

tools/docker/builder/openwrt/scripts/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+sed -ri "s/-DTARGET_PLATFORM_TYPE=[^ ]*/-DTARGET_PLATFORM_TYPE=${TARGET_PLATFORM_TYPE}/" feeds/prpl/prplmesh/Makefile
+sed -ri "s/PKG_REV:=.*$/PKG_REV:=${PRPLMESH_VERSION}/" feeds/prpl/prplmesh/Makefile
+make V=sc
+find bin -name 'prplmesh*.ipk' -exec cp {} "prplmesh-${TARGET}-${TARGET_PLATFORM_TYPE}-${OPENWRT_VERSION}-${PRPLMESH_VERSION}.ipk" \;
```
arnout: Is there a chance that the find will match several ipk files?

arnout (changes requested):

tools/docker/builder/openwrt/configs/sdk_config
rmelotte: Done.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Not really useful no, it has been removed.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Done.

tools/docker/builder/openwrt/scripts/build.sh
rmelotte: In theory yes, if another package builds an ipk matching "prplmesh*".
I don't see any easy way to get an exact match at this point though. 
We could check that the version number is following the package name maybe.
For reference, here is the full package name for the rax40:  `prplmesh_1.0-1_mips_24kc_nomips16.ipk`.

tools/docker/builder/openwrt/build.sh
rmelotte: I went with [USE_SOURCE_DIR](https://openwrt.org/docs/guide-developer/packages#use_source_dir) since it doesn't require to commit changes in order to build. As it requires a separate "prepare" step, we will have to document how to rebuild manually  somewhere.

rmelotte:
Rebased on master.
Also made the git describe output shorter, so that the "prplmesh version" part of the resulting .ipk is only a hash (and potentially "-dirty").

tools/docker/builder/openwrt/build.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/sh -e
+
+scriptdir="$(cd "${0%/*}"; pwd)"
+rootdir="${scriptdir%/*/*/*/*}"
+
+if [ ! -f "$1" ] ; then
+    echo "Please specify a file containing the build args as the
+    first argument"
+    exit 1
+fi
```
arnout, line 10: AFAICS there are now defaults for everything, so it is not actually needed to pass a config file.

However, I think it's nicer to use command line arguments instead of a config file, using getopt, like we do in most other scripts.

This is something we can easily update later so (nitpick).

arnout (approved):
One nitpick. Please rebase on master so we can merge.

tools/docker/builder/openwrt/build.sh
rmelotte: Indeed there are now defaults for everything. Now that you mention it, I should remove the default for TARGET and TARGET_PLATFORM_TYPE, as this build script will be used to build for other targets as well.
The idea of the config file was to make it easy for anyone to build for a specific target, without the need to know what TARGET and TARGET_PLATFORM_TYPE should be.

What I could do instead is to add a mandatory argument for the target and list the targets in the same script (currently only turris-omnia).
I'll rebase on master and keep it like that for now, please ping me if you want that done before merging.

PR #549, opened 2019-12-07
tlvf: remove kMinimumLength from BaseClass

tomereli:
kMinimumLength is 3, and is used to verify that each TLV has at least 3
bytes - 1 byte type, 2 bytes length.
TLVF generates a check for this in the init() function of each class,
even if it is not a TLV class, which verifies that there is "enough"
room in the buffer.
This check is wrong for non TLV classes which may not contain even a
single parameter (many of the VS classes for example), and is wrong also
for TLV classes since it only verifies there is enough room for the TLV
header.
It is even more silly to use it since we have a better API to use -
get_initial_size(), which takes to account not only the TLV header size
but also each class members.

Therefore, delete kMinimumLength and use get_initial_size() instead for
all init() functions.

Since this change is big, a followup commit will produce all the
autogenerated files.


CoralMalachi (approved):

LiorAmram (approved):

arnout (approved):
Nice catch!

PR #550, opened 2019-12-08
tlvf: remove copied source and include files

tomereli:
Copied files are confusing - they appear twice in the editor, and in
order to modify tlvf file we need to edit the src folder file and clean
make which takes time or do the changes to the Autogenerated files but
risk override by accidental clean make.

This commit first removes the copy command from tlvf.py to copy the source
files to the Autogenerated folder, and removes all the files themselves
from the autogenerated folder.

Then, we update tlvf and btlvf CMakeLists.txt to include the source
files in the build and install targets.


framework/tlvf/tlvf.py
```diff
@@ -386,7 +386,7 @@ def __init__(self, src_path, yaml_path, out_path, conf_path, print_dependencies,
             sys.exit(0)
 
         self.loadAllYamlFilesToDB()
-        self.copyIncludeSource()
+        #self.copyIncludeSource()
```
arnout: Don't comment it out, just remove it. And remove the definition of copyIncludeSource as well, and recursively anything else that is no longer needed.

arnout (approved):
Excellent idea! I've wanted to do this myself for some time now...

Nitpick: this should have been at least two, maybe even three or even four commits:

- Change tlvf CMakeLists.txt
- Change btlvf CMakeLists.txt (could be merged with the previous one)
- Change tlvf.py
- Remove Autogenerated (could be merged with the previous one)

PR #551, opened 2019-12-08
Dev/tlv list example 2

tomereli:


framework/tlvf/src/include/tlvf/TlvList.h
```diff
@@ -0,0 +1,136 @@
...
+        } else {
+            return std::shared_ptr<T>();
+        }
+        return ptr;
+    }
+
+    template <class T> std::shared_ptr<T> dynamicCast(std::shared_ptr<BaseClass> ptr) const
+    {
+        return std::dynamic_pointer_cast<T>(ptr);
+    }
```
arnout: When this PR is more or less complete, could you check if this method is really used? I just copied it from CmduMessage but I suspect some of these methods aren't used, or maybe only once.

 This one in particular looks pretty useless, it just abbreviates `std::dynamic_pointer_cast` to `dynamicCast`. And it should be static.

framework/tlvf/src/include/tlvf/TlvList.h
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * @brief Get the Class object at index idx in the all classes array
+     *
+     * @param idx index in the all classes array
+     * @return std::shared_ptr<BaseClass> to the class object at index idx, nullptr if not found
+     */
+    std::shared_ptr<BaseClass> getClass(size_t idx) const;
```
arnout: Same here: is it really used? It is used by the two functions below, but those really should be using `for (auto it : m_class_vector)` instead.

framework/tlvf/src/include/tlvf/TlvList.h
```diff
@@ -0,0 +1,136 @@
...
+     * @tparam T class template
+     * @param idx index in the class T array
+     * @return std::shared_ptr<T> to the T class at index `idx` in the class T array
+     */
+    template <class T> std::shared_ptr<T> getClass(size_t idx) const
+    {
+        size_t idx_ = 0;
+        for (size_t i = 0; i < getClassCount(); i++) {
+            if (auto c = std::dynamic_pointer_cast<T>(getClass(i))) {
+                if (idx_++ == idx)
```
arnout: Please do the increment as a separate statement.

A simpler way could be:

```
for (auto it : m_class_vector) {
    if (auto c = std::dynamic_pointer_cast<T>(it)) {
        if (idx == 0)
            return c;
        idx--;
    }
}
```

However I'm not particularly fond of modifying arguments or decrementing an iterator myself, so probably not such a good idea :-)

framework/tlvf/src/include/tlvf/TlvList.h
```diff
@@ -0,0 +1,136 @@
...
+        for (size_t i = 0; i < getClassCount(); i++) {
+            if (auto c = std::dynamic_pointer_cast<T>(getClass(i))) {
+                count++;
+            }
+        }
+        return count;
+    }
+
+    size_t getClassCount() const;
+    const std::vector<std::shared_ptr<BaseClass>> &getClassVector() const;
```
arnout: Again, please check if they are really used, and if yes, if they can't be replaced by iterators and ranged loops.

framework/tlvf/src/src/TlvList.cpp
```diff
@@ -0,0 +1,73 @@
...
+        return nullptr;
+    }
+}
+
+size_t TlvList::getClassCount() const { return m_class_vector.size(); }
+
+const std::vector<std::shared_ptr<BaseClass>> &TlvList::getClassVector() const
+{
+    return (const std::vector<std::shared_ptr<BaseClass>> &)m_class_vector;
+}
```
arnout, line 36: It would be better to make these inline. Also, the cast isn't needed.

framework/tlvf/src/include/tlvf/TlvList.h
```diff
@@ -0,0 +1,136 @@
...
+    template <class T> size_t getClassCount() const
+    {
+        size_t count = 0;
+        for (size_t i = 0; i < getClassCount(); i++) {
+            if (auto c = std::dynamic_pointer_cast<T>(getClass(i))) {
+                count++;
+            }
+        }
+        return count;
+    }
```
arnout: I think this is only used for a pattern like this:
```
for (int idx = 0; idx < t.getClassCount<Foo>(); idx++) {
   auto foo = t.getClass<Foo>(idx);
   ...
}
```

It would be a lot better to convert this to an iterator pattern, but I guess that's a lot more complicated.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -27,6 +30,35 @@
...
+        uint8_t action_op() { return actionhdr()->action_op(); };
+        uint16_t id() { return actionhdr()->id(); }
+        template <class T> std::shared_ptr<T> addClass()
+        {
+            return m_contents->addClass<T>();
+        }
+        template <class T> std::shared_ptr<T> getClass()
+        {
+            return m_contents->getClass<T>();
+        }
```
arnout: Note that the idea was that these would be removed again at the end of the PR, because the users instead use a wrapper, e.g. `addAction()` and `getAction()` that take into account the action header. Or that m_contents is public so callers can use it directly.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -52,41 +84,47 @@ class message_com {
...
+        std::unique_ptr<ieee1905_1::TlvList> actions =
+        std::make_unique<ieee1905_1::TlvList>(vs_tlv->payload(), payload_length, false, false);
+        if (!actions)
+            return nullptr;
+        std::shared_ptr<beerocks_header> hdr = std::make_shared<beerocks_header>(std::move(actions));
+        if (!hdr)
+            return nullptr;
+        if (!hdr->addClass<beerocks_message::cACTION_HEADER>())
+            return nullptr;
+        if (!hdr->addClass<T>())
```
arnout: So here we could use `hdr->m_contents` directly.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -139,7 +175,15 @@ class message_com {
             return nullptr;
         }
 
-        return add_intel_vs_data<T>(cmdu_tx, tlvhdr, id);
+        auto beerocks_header = add_intel_vs_data<T>(tlvhdr, id);
+        cmdu_tx.header = beerocks_header;
```
arnout: Why isn't this done in add_intel_vs_data?

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -139,7 +175,15 @@ class message_com {
...
+
+        // According to C++'03 Standard 14.2/4:
+        // When the name of a member template specialization appears after . or -> in a postfix-expression,
+        // or after nested-name-specifier in a qualified-id, and the postfix-expression or qualified-id
+        // explicitly depends on a template-parameter (14.6.2), the member template name must be prefixed by
+        // the keyword template. Otherwise the name is assumed to name a non-template.
+        return beerocks_header->template getClass<T>();
```
arnout, line 211: So if we remove the non-templated getClass, this wouldn't be needed anymore, right?

If we don't remove it, we could still rename it.

controller/src/beerocks/master/tasks/association_handling_task.h
```diff
@@ -23,9 +23,9 @@ class association_handling_task : public task {
 
 protected:
     virtual void work() override;
-    virtual void handle_response(std::string radio_mac,
-                                 beerocks_message::eActionOp_CONTROL action_op,
-                                 ieee1905_1::CmduMessageRx &cmdu_rx) override;
+    virtual void handle_response(
+        std::string radio_mac, beerocks_message::eActionOp_CONTROL action_op,
+        std::shared_ptr<beerocks::beerocks_header> beerocks_header) override;
```
arnout, line 28: [CppCoreGuideline R.30](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#r30-take-smart-pointers-as-parameters-only-to-explicitly-express-lifetime-semantics) Use reference not smartptr as function argument.

I think I need to make a CODING_STYLE entry for this...

framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -19,62 +20,36 @@ namespace ieee1905_1 {
...
-            }
-        } else {
-            return std::shared_ptr<T>();
-        }
-        return ptr;
-    }
 
+    ieee1905_1::TlvList tlvs;
+
+    template <class T> std::shared_ptr<T> addClass() { return tlvs.addClass<T>(); }
```
arnout, line 31: Same here: please remove the "forwarding" wrappers and use tlvs directly, or make a more relevant wrapper if appropriate.

framework/tlvf/src/src/BaseClass.cpp
```diff
@@ -35,7 +35,7 @@ bool BaseClass::isInitialized() { return m_init_succeeded; }
  */
 bool BaseClass::buffPtrIncrementSafe(size_t length)
 {
-    if (getBuffRemainingBytes() <= length) {
+    if (getBuffRemainingBytes() < length) {
```
arnout, line 38: Please add a test to tlvf_tests that proves this is needed.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -77,42 +56,5 @@ std::shared_ptr<tlvVendorSpecific> CmduMessageTx::add_vs_tlv(tlvVendorSpecific::
 
 bool CmduMessageTx::finalize(bool swap_needed)
 {
-    if (m_finalized)
-        return true;
-
-    if (!m_cmdu_header)
-        return false; 
-
-    // Update vendor specific length which is seperated to 3 classes
```
arnout, line 86: As I mentioned in the daily: if you don't find a better solution, you can just keep this hack for the time being. Allocate the vendorspecific TLV with a minimum size (which is how it was done before) and overwrite the length field like below.

arnout (commented):
Please make one or more PRs for the preparatory commits already, so the big change becomes easier to review.

framework/tlvf/src/src/BaseClass.cpp
tomereli: I updated the commit message, no need for a test it is definitely a bug - if you have n bytes left and you want to increment by n, then it would fail since n <= n is true.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: You were right, removed it.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: Done

framework/tlvf/src/include/tlvf/CmduMessage.h
tomereli: Thinking about this again I don't get what you mean here. `addClass` is widely used, changing everywhere to `cmdu_rx.tlvs.addClass` is a huge change, and although it can be done I see no reason not to encapsulate that with this forwarding wrapper - can you explain whats wrong with it?

framework/tlvf/src/include/tlvf/TlvList.h
tomereli: Agree, removing.

framework/tlvf/src/include/tlvf/TlvList.h
tomereli: I'm not sure I agree that `idx-- == idx` is not readable enough. People should know the order of operations which happen here, if they don't then its a red sign.

framework/tlvf/src/include/tlvf/TlvList.h
tomereli: It would, and it is, so I'll leave it as is for now :)

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: I kept the wrapper functions for backward compatability. Plus I think having some forward wrappers such as this is OK - it also makes lines shorter.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: Obsolete and now it is.

controller/src/beerocks/master/tasks/association_handling_task.h
tomereli: You are right basically but for the time being, I prefer to keep it as a shared_ptr since changing it to reference will cause a massive change to all the callers I would prefer to avoid at this stage.

framework/tlvf/src/src/CmduMessageTx.cpp
tomereli: Did it a bit differently, but it works.

tomereli:
Replaced by #572

framework/tlvf/src/include/tlvf/CmduMessage.h
arnout: I think I put this comment on the wrong line. Or maybe I was just wrong :-). Anyway in the final review it will turn up again if it really was relevant.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
arnout: There is no backward compatibility for these, because they didn't exist before this PR.

Avoiding the forwarding wrappers is because it makes class instantiation more difficult. But the main point actually is that instead of a simple wrapper, you would typically want forwarding to be done with more intelligence. That's why I was talking about addAction() and getAction().

controller/src/beerocks/master/tasks/association_handling_task.h
arnout: You're *introducing* the shared_ptr here, so I don't see what you mean with the massive change to all callers. I mean, yes it's a massive change to all callers, but you're doing that change anyway.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: But now that `beerocks_header` extends `TlvList`, it gets `addClass` and `getClass` by default. We can add wrappers now `addAction` and `getAction` but we won't be able to "hide" `addClass` and `getClass`.

PR #552, opened 2019-12-08
Refactor/remove mrouter

morantr:
Remove Controllers Mrouter from the code.


controller/src/beerocks/master/beerocks_master_main.cpp
```diff
@@ -304,7 +304,14 @@ int main(int argc, char *argv[])
     std::string master_uds  = beerocks_master_conf.temp_path + std::string(BEEROCKS_MASTER_UDS);
     std::string mrouter_uds = beerocks_master_conf.temp_path + std::string(BEEROCKS_MROUTER_UDS);
 
-    son::db master_db(master_conf, logger);
```
tomereli, line 307: Why not just use `beerocks::net::network_utils::get_iface_info(bridge_info, bridge_iface)` when we need it?

morantr:
For some reason the test_flows.sh fails on some steering tests. I tried to run it on local GW and it worked for me. So I couldn't find the reason why it is not working.

controller/src/beerocks/master/db/db.h
```diff
@@ -748,6 +758,8 @@ class db {
     // certification
     std::shared_ptr<uint8_t> certification_tx_buffer;
     std::unordered_map<sMacAddr, std::list<bss_info_conf_t>> bss_infos; // key=al_mac
+
+    master_thread *m_master_thread_ctx;
```
arnout, line 759: Instead of making this a pointer, I'd make it a reference and pass it in the constructor (obviously) instead of set_master_thread_ctx(). This has the following advantages:

- Can remove the set function;
- It's guaranteed not to change over the lifetime;
- It indicates that the lifetime of db is tied to the lifetime of masther_thread;
- get_master_thread_ctx() returns a reference, so no need to check nullptr.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -112,7 +138,7 @@ bool master_thread::init()
         }
     }
 
-    return socket_thread::init();
+    return true;
```
arnout, line 157: This change should have been part of "controller: Make son_master_thread to be derived from btl" commit.

controller/src/beerocks/master/db/db.h
```diff
@@ -136,7 +136,8 @@ class db {
         bool client_optimal_path_roaming_prefer_signal_strength = false;
     } sDbMasterSettings;
 
-    db(sDbMasterConfig &config_, beerocks::logging &logger_) : config(config_), logger(logger_)
+    db(sDbMasterConfig &config_, beerocks::logging &logger_, const std::string &local_bridge_mac)
+        : config(config_), logger(logger_), m_local_bridge_mac(local_bridge_mac)
```
arnout, line 140: Although this address is indeed the local bridge address at the moment, we will have to change this at some point (because it's wrong: it should be a locally administered address according to 1905.1, while the bridge address is usually a OUI address). Therefore, I wouldn't create a new variable that has "bridge" in the name. Instead, use `al_mac`, `m_al_mac` etc.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -187,6 +187,42 @@ bool master_thread::socket_disconnected(Socket *sd)
 
 bool master_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
+    // Note: sd can be equal to nullptr it the master_thread is derived from btl_local_bus and the
```
arnout: The part `it the master_thread is derived from btl_local_bus` is not clear and IMO redundant.  Actually, the entire comment is redundant - we're only using `sd` if `from_bus` returns false, so all is well...

controller/src/beerocks/master/son_actions.cpp
```diff
@@ -294,23 +296,18 @@ void son_actions::handle_dead_node(std::string mac, std::string hostap_mac, db &
      */
     if ((mac_type == beerocks::TYPE_IRE_BACKHAUL || mac_type == beerocks::TYPE_CLIENT) &&
         database.is_node_wireless(mac)) {
-        Socket *hostap_mac_sd = database.get_node_socket(hostap_mac);
-        if (hostap_mac_sd) {
```
arnout, line 298: We still need this condition, but then on the MAC address:

```
if (!agent_mac.empty()) {
```

Or actually, looking at the rest, we don't do this anywhere else, so probably not needed. In that case, it should be mentoined in the commit message. Which is anyway is not entirely clear, better would be:

```
 controller: send_cmdu_to_agent: use bus instead of mrouter socket

Change "send_cmdu_to_agent" on son_actions to send the message to the
Bus instead of UDS from mrouter.

To replace the Socket, we need the master_thread as an argument to be able
to call the send_cmdu_to_bus function, and the MAC address of the agent.
Convert all callers to get the MAC address of the agent from the database 
with get_node_parent_ire instead of getting the Socket with get_node_socket.

There were four cases where the return value of get_node_socket() was
checked. Remove that check now, because we don't do it anywhere else: a
hostap node is guaranteed to have a ire parent, and if the MAC address is
empty, it will fail in the sending stage.
```

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -87,7 +87,6 @@ void client_steering_task::work()
 
 void client_steering_task::steer_sta()
 {
-    Socket *sta_socket = database.get_node_socket(sta_mac);
```
arnout, line 90: This one is different from the rest of the changes, so it should be mentioned explicitly in the commit message. Actually, I even think it could be a separate commit.

arnout (changes requested):
Reviewed up to "controller: Change "send_cmdu_to_agent" on son_actions "

Nothing major so far, it's definitely going in the right direction.

Are there any cleanups that can still be done but haven't been taken in this PR? If yes, it would be nice to mention this as a comment in the PR itself.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -64,16 +64,16 @@ void agent_ucc_listener::clear_configuration()
...
- * @return Socket* Socket to the Agent.
+ * @return true if successful, false if not.
  */
-Socket *agent_ucc_listener::get_dev_send_1905_destination_socket(const std::string &dest_alid)
+bool agent_ucc_listener::validate_destination_alid(const std::string &dest_alid)
 {
     // On the agent side, the dest_alid is not really needed since the destination socket will
     // always be the controller socket.
-    return *m_controller_sd;
+    return (*m_controller_sd)->getPeerMac() == dest_alid;
```
arnout, line 76: Nice! I hadn't thought of the possibility to validate the destination.

arnout (approved):
Reviewed the rest of the PR.

Since my comments in the first review were all of the nitpicky type, I'm approving already.

arnout:
One more thing: `transport_socket_thread::init()` must be done already in commit `controller: Make son_master_thread to be derived from btl` otherwise the UCC class will trigger an error because it tries to add a socket when the poller isn't initialized yet.

tomereli (approved):

controller/src/beerocks/master/db/db.h
tomereli: +1 for `m_al_mac`

controller/src/beerocks/master/db/db.h
morantr: It can't be done since the database is being created before the son_master is created and being passed as a reference to it.
I guess there is a solution for that, like not creating the database outside to the son_master thread and initialize it in the son_master constructor.
Do you want me to do it? It will need an extra commit on top.
@arnout @tomereli

controller/src/beerocks/master/son_master_thread.cpp
morantr: what? the returning of `true`? it is not critical...

controller/src/beerocks/master/beerocks_master_main.cpp
morantr: Because we only need to read it once. There is no point in reading it all over again every time.

controller/src/beerocks/master/db/db.h
morantr: I don't understand, in the future `al_mac` and the bridge mac won't be equal?

controller/src/beerocks/master/son_master_thread.cpp
morantr: Removed the comment.

controller/src/beerocks/master/son_actions.cpp
morantr: Will change the commit message

controller/src/beerocks/master/tasks/client_steering_task.cpp
morantr: It really not. This is used for backhaul roaming. Here we get the backhaul socket which is basically the same mac address of the parent bridge mac which we later get.
On line 144 I replace the validation of this socket by comparing the node type to `beerocks::TYPE_IRE_BACKHAUL`.
Since I remove all `get_node_socket` calls in this commit, this needs to be part of this commit/

controller/src/beerocks/master/son_master_thread.cpp
morantr: Not critical.

controller/src/beerocks/master/db/db.h
tomereli: Sounds legit - the son master is the "owner" of the DB so it makes sense that it will create it in its constructor, so I would say go for it.

controller/src/beerocks/master/db/db.h
tomereli: The name should be al_mac. Each device can define its own al mac, currently it is the bridge mac but it can change, so the answer is "maybe".

controller/src/beerocks/master/db/db.h
arnout: I'm OK either way. It was anyway just a nitpick comment.

common/beerocks/btl/btl_local_bus.cpp
```diff
@@ -82,16 +71,26 @@ bool transport_socket_thread::bus_init()
 
 bool transport_socket_thread::bus_subscribe(const std::vector<ieee1905_1::eMessageType> &msg_types)
 {
-    bool ret = true;
-
     LOG_IF(!bus, FATAL) << "Bus is not allocated!";
     LOG_IF(!poller, FATAL) << "Poller is not allocaed!";
     bus->Init();
```
tomereli, line 76: I think we should have a `bus_Init()` which will call `bus->init()` so in the future if we call `bus_subscribe` more than once it will not cause an issue.

tomereli (approved):

common/beerocks/btl/btl_local_bus.cpp
morantr: we have it already.

PR #553, opened 2019-12-09
doc: fix README.md wrong cmake command

tomereli:
Remove extra whitespace after -H and -B CMAKE options which cause the
command to fail.


arnout (approved):
This change triggered me to check since when the -H and -B options exist, and it's actually complicated.

-B exists since pre-3.0 but it was undocumented and supposed to be internal.
-H also exists since pre-3.0.
Both of them required the path to be followed immediately, without space.

All this changed in 3.13: -H was deprecated in favour of -S, -B was documented, and both of them gained the possibility to have a space between the option and the path.

Since 3.13 is pretty recent, we want to support older versions, so indeed we should use the old -H option and without spaces.

Issue #405, opened 2019-10-10
4.8.1,2,3,5 Client Steering


arnout:
The prplMesh agent on Turris Omnia with OpenWrt needs to pass the certification test case 4.8.1, 4.8.2, 4.8.3 and 4.8.5 as defined by the Wi-Fi CERTIFIED EasyMesh (tm) Test Plan.

- [x] #429 Correct information in BTMSteeringReport message
- [x] #400 Pass flow 4.8.1 test in testbed
- [x] #401 Pass flow 4.8.2 test in testbed
- [x] #986 Handle Client Steering Opportunity
- [x] #402 Pass flow 4.8.3 test in testbed
- [x] #404 Pass flow 4.8.5 test in testbed

rmelotte:
Here is the current status:
- MAP-4.8.2_ETH_FH24G
Fails [with the Qualcomm controller](https://gitlab.com/prpl-foundation/prplMesh/-/jobs/454093091), but passes [with the Marvell controller](https://gitlab.com/prpl-foundation/prplMesh/-/jobs/454165724).

There is an issue to track the failures with some of the certified controllers here: https://github.com/prplfoundation/prplMesh/issues/921

- MAP-4.8.3_ETH_FH5GL
Job here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/454253682
It currently fails with:
```
2020-02-28 19:05:49.099 - INFO - MSG:  No 1905 Ack message found 
```

- 4.8.4 and 4.8.5 fail because they use the Mediatek and Qualcomm controllers respectively, with which prplMesh currently fails on the onboarding

PR #555, opened 2019-12-09
Bugfix - Client Steering EasyMesh Messages

vitalybu:
This PR fixes the client steering logic after switching to EasyMesh messages.
Current implementation sends incorrect values and missing local agent BSSID filtering.
This caused the target agent to add the steering station into it's DENY list which completely broke the mechanism.

### DIFF
--- a/common/beerocks/bwl/include/bwl/ap_wlan_hal.h
+++ b/common/beerocks/bwl/include/bwl/ap_wlan_hal.h
@@ -104,19 +104,21 @@ public:
      * Allow the station with the given MAC address to connect.
      *
      * @param [in] mac The MAC address of the station.
+     * @param [in] bssid The BSSID to which the operation is applicable.
      *
      * @return true on success or false on error.
      */
-    virtual bool sta_allow(const std::string &mac) = 0;
+    virtual bool sta_allow(const std::string &mac, const std::string &bssid) = 0;
 
     /*!
      * Deny the station with the given MAC address from connecting to the AP.
      *
      * @param [in] mac The MAC address of the station.        
+     * @param [in] bssid The BSSID to which the operation is applicable.
      * 
      * @return true on success or false on error.
      */
-    virtual bool sta_deny(const std::string &mac) = 0;
+    virtual bool sta_deny(const std::string &mac, const std::string &bssid) = 0;
 
     /*!
      * Disassociate the station with the given MAC address.
--- a/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_1905_vs.yaml
+++ b/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_1905_vs.yaml
@@ -35,8 +35,3 @@ tlvVsClientAssociationEvent:
   disconnect_type: 
     _type: uint8_t
     _comment: relevant only on disconnect event
-
-
-tlvVsClientAssociationControlRequest:
-  _type: class
-  ipv4: beerocks::net::sIpv4Addr 
--- a/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
+++ b/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
@@ -133,11 +133,12 @@ cACTION_APMANAGER_CLIENT_DISCONNECT_RESPONSE:
 cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST:
   _type: class
   mac: sMacAddr
+  bssid: sMacAddr
 
 cACTION_APMANAGER_CLIENT_ALLOW_REQUEST:
   _type: class
   mac: sMacAddr
-  ipv4: beerocks::net::sIpv4Addr
+  bssid: sMacAddr
 
 cACTION_APMANAGER_CLIENT_RX_RSSI_MEASUREMENT_REQUEST:
   _type: class
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -682,9 +682,12 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             LOG(ERROR) << "addClass cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST failed";
             return false;
         }
+
         std::string sta_mac = network_utils::mac_to_string(request->mac());
-        LOG(DEBUG) << "CLIENT_DISALLOW, mac = " << sta_mac;
-        ap_wlan_hal->sta_deny(sta_mac);
+        std::string bssid   = network_utils::mac_to_string(request->bssid());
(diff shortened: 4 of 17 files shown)


tomereli (approved):

tomereli (changes requested):
please explain in the PR description though how this was tested and on which target.
Also, it looks like the implantation for nl80211 BWL is not taking the bssid into account. Perhaps we can filter the bssid in the ap manager instead of the BWL itself? As a matter of fact, even the controller has the vap list so you can save the messages to the other agents.

tomereli (commented):
Not approving yet so Arnout has also a chance to review.

vitalybu:
> please explain in the PR description though how this was tested and on which target.
> Also, it looks like the implantation for nl80211 BWL is not taking the bssid into account. Perhaps we can filter the bssid in the ap manager instead of the BWL itself? As a matter of fact, even the controller has the vap list so you can save the messages to the other agents.

This fix was tested on CGR running RDKB.
I thought about the right place to implement the filtering and decided to go with BWL for the following reasons:
- We can't assume anything about 3rd party controllers, so the filtering must be done in the agent.
- Currently the Agent and the AP Manager are somewhat agnostic to the underlying networking structure. They don't even store a list of VAPs. We can change this of course, and perhaps as more and more radio/vap specific messages will be implemented it'll make more sense, but at this point I came to the decision that BWL filtering is good enough.
- BWL is platform specific, and some platform may need to handle this differently.

As for the NL80211 BWL implementation, the BSSID is indeed ignored there, but it doesn't break the implementation. The fix was made in the controller, so any NL80211 platform will keep Allowing/Disallowing stations for the entire radio (we don't currently support any NL80211 platforms with VAPs anyway).

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -133,11 +133,12 @@ cACTION_APMANAGER_CLIENT_DISCONNECT_RESPONSE:
 cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST:
   _type: class
   mac: sMacAddr
+  bssid: sMacAddr
 
 cACTION_APMANAGER_CLIENT_ALLOW_REQUEST:
   _type: class
   mac: sMacAddr
-  ipv4: beerocks::net::sIpv4Addr
```
arnout, line 140: If you remove ipv4, then the same commit should also remove its uses from son_slave_thread and ap_manager_thread, otherwise the build breaks.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_1905_vs.yaml
```diff
@@ -35,8 +35,3 @@ tlvVsClientAssociationEvent:
   disconnect_type: 
     _type: uint8_t
     _comment: relevant only on disconnect event
-
-
-tlvVsClientAssociationControlRequest:
```
arnout, line 40: At this point, it *is* actually still used in slave_thread::handle_client_association_request, so that has to be removed first.

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
```diff
@@ -104,19 +104,21 @@ class ap_wlan_hal : public virtual base_wlan_hal {
      * Allow the station with the given MAC address to connect.
      *
      * @param [in] mac The MAC address of the station.
+     * @param [in] bssid The BSSID to which the operation is applicable.
      *
      * @return true on success or false on error.
      */
-    virtual bool sta_allow(const std::string &mac) = 0;
+    virtual bool sta_allow(const std::string &mac, const std::string &bssid) = 0;
```
arnout, line 111: I don't want any new code that uses std::string instead of sMacAddr. Ideally there would be a separate commit that changes the mac to sMacAddr as well, but for the time being I'm OK with just adding `const sMacAddr &bssid`.

The changes to the caller have to be done in the same commit. To make the transition smoother, you could give the bssid a default value of ZERO_MAC here and remove that default later.

tomereli:
> > please explain in the PR description though how this was tested and on which target.
> > Also, it looks like the implantation for nl80211 BWL is not taking the bssid into account. Perhaps we can filter the bssid in the ap manager instead of the BWL itself? As a matter of fact, even the controller has the vap list so you can save the messages to the other agents.
> 
> This fix was tested on CGR running RDKB.
> I thought about the right place to implement the filtering and decided to go with BWL for the following reasons:
> 
> * We can't assume anything about 3rd party controllers, so the filtering must be done in the agent.
> * Currently the Agent and the AP Manager are somewhat agnostic to the underlying networking structure. They don't even store a list of VAPs. We can change this of course, and perhaps as more and more radio/vap specific messages will be implemented it'll make more sense, but at this point I came to the decision that BWL filtering is good enough.
> * BWL is platform specific, and some platform may need to handle this differently.
> 
> As for the NL80211 BWL implementation, the BSSID is indeed ignored there, but it doesn't break the implementation. The fix was made in the controller, so any NL80211 platform will keep Allowing/Disallowing stations for the entire radio (we don't currently support any NL80211 platforms with VAPs anyway).

Good points, thank you for the explanation. I'll approve, but let's give @arnout also a chance to review before merging?

arnout (approved):
The commit splitting/ordering is not OK, but that's nitpicking.

I really don't want new uses of string for MAC address. However, since the functions in bwl are already using strings, it's also something that can be fixed later.

I also think the slave should have a copy of the VAP list. But even if it does, the BSSID has to be passed to the AP manager so this PR still applies.

Note that we will also need to handle the validity period of the blocking request at some point. However, this is not part of the test plan, so not included in #405 

Another thing that I'm wondering about is: wouldn't it be better if we sent the wfa_map tlvClientAssociationControlRequest as-is to the AP Manager? Why do we need to do this conversion in the slave?

tomereli (approved):
Arnout raised valid comments, so consider fixing those before merging.

vitalybu:
> The commit splitting/ordering is not OK, but that's nitpicking.
> 
> I really don't want new uses of string for MAC address. However, since the functions in bwl are already using strings, it's also something that can be fixed later.
> 
> I also think the slave should have a copy of the VAP list. But even if it does, the BSSID has to be passed to the AP manager so this PR still applies.
> 
> Note that we will also need to handle the validity period of the blocking request at some point. However, this is not part of the test plan, so not included in #405
> 
> Another thing that I'm wondering about is: wouldn't it be better if we sent the wfa_map tlvClientAssociationControlRequest as-is to the AP Manager? Why do we need to do this conversion in the slave?

I agree that the commit splitting and ordering is not optimal, but in this case I preferred to split in this way for easier review (assuming that the stand-alone commits won't be cherry-picked anywhere).

Thought about passing the `tlvClientAssociationControlRequest` message as-is to the AP Manager, but currently the AP Manager handles vendor-specific messages [only](https://github.com/prplfoundation/prplMesh/blob/bugfix/client-steering-easymesh-messages/agent/src/beerocks/slave/ap_manager_thread.cpp#L425). Should probably be changed later.

I'll merge this PR now to unblock work on UGW8.x and RDKB, but let's discuss the Agent<->AP Manager messaging mechanism in the next architectural meeting.

PR #559, opened 2019-12-11
Feature/ci docker in docker

rmelotte:
Shell executors have a number of limitations:
- they may keep build results from one run to the other.
- they may not be able to remove previous built results made by
  root (a docker container for instance).
- when they use artifacts from a docker executor, they may have
  problems with file ownership and issue warnings like the following:
' WARNING: build/: lchown build/: operation not permitted (suppressing repeats) '
- the runners using shell executors are harder to reproduce on other
  hosts.

For those reasons, avoid using shell executors at all, and use docker
executors only.

One of the big required changes is that the build directory (when using the docker builder) is now inside the repository itself (under "build").

Also, since running the tests requires to use docker _inside_ docker, the docker-dind image and privileged runners are used.


.gitlab-ci.yml
```diff
@@ -14,14 +14,14 @@ build-in-docker:
...
-      - ../build
+      - build
 run-tests:
   stage: test
   script:
     - cd tools/docker
     # Try to pull from dockerhub, build the container images if that fails
     - ./image-pull.sh || ./image-build.sh
-    - ./run.sh -f -d --entrypoint "$PWD/../../../build/install/bin/tests/tlvf_test"
+    - ./run.sh -f -d --entrypoint "$PWD/../../build/install/bin/tests/tlvf_test"
```
arnout: The same change needs to be done in `.travis-ci.yml`.

tools/docker/builder/Dockerfile
```diff
@@ -34,4 +34,4 @@ RUN apt-get update && apt-get install -y \
 	vim \
     && rm -rf /var/lib/apt/lists/*
 
-ENTRYPOINT ["./prplMesh/tools/maptools.py", "build"]
+ENTRYPOINT ["./tools/maptools.py", "build"]
```
arnout, line 37: This is really annoying - it requires a rebuild and push of the docker containers or things will break.

I think this can be solved by adding a symlink `prplMesh -> .` to the rootdir. So please do that as well in the same commit.

.gitlab-ci.yml
```diff
@@ -15,6 +15,10 @@ build-in-docker:
   artifacts:
     paths:
       - build
+  tags:
+    - docker
+    - specific
```
arnout: NACK this. We can just configure the project on gitlab.com to not use the shared runners. But I don't think we need to - I believe the "private" runners are used in preference to shared runners, so the shared runners would only be used when the private one is already overloaded.

.gitlab-ci.yml
```diff
@@ -21,7 +21,7 @@ run-tests:
     - cd tools/docker
     # Try to pull from dockerhub, build the container images if that fails
     - ./image-pull.sh || ./image-build.sh
-    - ./run.sh -f -d --entrypoint "$PWD/../../build/install/bin/tests/tlvf_test"
+    - ./run.sh -f --entrypoint "$PWD/../../build/install/bin/tests/tlvf_test"
```
arnout, line 31: A bit useless to do this since in the end we want to move this to a separate stage anyway, but OK.

tools/docker/tests/test_flows.sh
```diff
@@ -35,7 +35,7 @@ send_bml_command() {
 
 send_CAPI_command() {
     # send a CAPI command to a container.
-    ${rootdir}/tools/docker/tests/send_CAPI_command.py $(container_ip "$1") $(container_CAPI_port "$1") "$2"
+    ${rootdir}/tools/docker/tests/send_CAPI_command.py $(container_ip "$1") $(container_CAPI_port "$1") "$2" "$verbose_option"
```
arnout: I think it's more appropriate to just put `dbg` in front instead.

tools/docker/run.sh
```diff
@@ -105,6 +105,7 @@ main() {
                 --expose ${PORT}
                 -v ${installdir}:${installdir}
                 -v ${rootdir}:${rootdir}
+		-v ${rootdir}/logs/${NAME}:/tmp/${SUDO_USER:-${USER}}/beerocks/logs
```
arnout: Indentation is wrong here.

.gitlab-ci.yml
```diff
@@ -33,6 +33,11 @@ run-tests:
     # start the docker daemon and give it some time to start:
     - dockerd-entrypoint.sh > /dev/null 2>&1 & sleep 2
 
+  artifacts:
+    paths:
+      - logs
+    when: always
```
arnout, line 42: Maybe `on_failure` would be more appropriate?

.gitlab-ci.yml
```diff
@@ -1,31 +1,49 @@
...
-    - ./build.sh map -f CMAKE_BUILD_TYPE=Debug BUILD_TESTS=ON
+    - cmake -DBUILD_TESTS=ON -DCMAKE_BUILD_TYPE=$CMAKE_BUILD_TYPE -DCMAKE_INSTALL_PREFIX=build/install -H. -Bbuild -G Ninja
+    - ninja -C build install
     # Check that the AutoGenerated files are correct by checking git status
     - git update-index -q --refresh
     - git diff-index --exit-code HEAD
+
+  artifacts:
+    paths:
+      - build
```
arnout: Do we really need to save the entire build directory? Isn't build/install sufficient?

.gitlab-ci.yml
```diff
@@ -1,31 +1,49 @@
...
     - cd tools/docker
-    - ./run.sh -f -d --entrypoint "$PWD/../../../build/install/bin/tests/tlvf_test"
+    # Try to pull from dockerhub, build the container images if that fails
+    - ./image-pull.sh || ./image-build.sh
+    - ./run.sh -f --entrypoint "$PWD/../../build/install/bin/tests/tlvf_test"
     - ./tests/test_flows.sh -v
+
+  before_script:
+    # start the docker daemon and give it some time to start:
+    - dockerd-entrypoint.sh > /dev/null 2>&1 & sleep 2
```
arnout: Maybe better do an appropriate wait loop, e.g. `for i in $(seq 1 10); do docker info -f '.' && break; sleep .5; false; done`. Note that the `false` makes sure we error out if docker is not alive.

However, why can't we just set this as the ENTRYPOINT in our tests-runner image? Or specify it explicitly with

```
image:
  name: $CI_REGISTRY_IMAGE/tests-runner
  entrypoint: dockerd-entrypoint.sh
```

arnout (changes requested):
I'd like to try to pass both gitlab-ci and Travis before merging.

.gitlab-ci.yml

tools/docker/run.sh
rmelotte: That's only Github playing tricks on us. When you open the file, the indentation looks correct.

tools/docker/tests/test_flows.sh
rmelotte: Done, I dropped the entire commit and used dbg instead.

tools/docker/builder/Dockerfile
rmelotte: Done.

.gitlab-ci.yml
rmelotte: It doesn't hurt to keep them, they could be used as a reference for someone wanting to check what the logs look like in a normal flow. They are relatively small and will be deleted after some time anyway.
What do you think?

.gitlab-ci.yml

.gitlab-ci.yml
rmelotte: We cannot use the entrypoint as-is because gitlab-ci needs the entrypoint to either be empty, or to be a shell. For reference, you can see _part_ of what gitlab-ci is doing with the entrypoint here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/378489864
In a few words: it first exec the entrypoint, then gives it a script that, among other things, tries to exec another shell. After that (not visible in this log), the content of the "script" part of `.gitlab-ci.yml` is passed through stdin (along with a bunch of added commands and variables; `set -eo pipefail` is one of them).

I created a new entrypoint that uses the loop you describe, then runs the gitlab-ci script at the end (or fail if dockerd could not be started).

.gitlab-ci.yml
rmelotte: Removed the "specific" tag.

arnout (approved):
There are still a few references to ../build left:

- .vscode/launch.json
- README.md

But that can be fixed later.

There was really no need to do the last commit as a separate commit, it could be done immediately like that in 22540d55. But that's a nitpick.

tools/docker/tests/test_flows.sh
```diff
@@ -35,7 +35,7 @@ send_bml_command() {
 
 send_CAPI_command() {
     # send a CAPI command to a container.
-    ${topdir}/prplMesh/tools/docker/tests/send_CAPI_command.py $(container_ip "$1") $(container_CAPI_port "$1") "$2"
+    ${rootdir}/prplMesh/tools/docker/tests/send_CAPI_command.py $(container_ip "$1") $(container_CAPI_port "$1") "$2"
```
tomereli: How can this work? shouldn't this be `${rootdir}/tools/docker/tests/send_CAPI_command.py $(container_ip "$1") $(container_CAPI_port "$1") "$2"` ?

tools/docker/functions.sh
```diff
@@ -5,8 +5,8 @@
 # See LICENSE file for more details.
 ###############################################################
 
-installdir="${topdir}/build/install"
-sourcesdir="${topdir}/prplMesh"
+installdir="${rootdir}/build/install"
+rootdir="${rootdir}/prplMesh"
```
tomereli: Should be `rootdir="${scriptdir%/*/*}"` like in other places, shouldn't it?

tools/docker/tests/test_flows.sh
rmelotte: This commit was generated using the command in the commit message. The next commit fixes this manually.
It would have been possible to include this particular change in the first (generated) commit, but other manual fixes are needed, so the 2 commits _have_ to be applied together anyway.

tomereli (commented):
Commit 2c435b4af4a12dcc88c23a5a33a16fbc93958455 mentions that some manual fixes are needed after the mechanical replace of `topdir` to `rootdir` but I didn't see these fixed in followup commits, so I think these are real bugs - can you please double check?

tomereli:
I now see that e74d848c7e71dc3bba8275f5a3db2c1436885cc3 but IMHO it should be squashed to the commit before that, otherwise this commit "breaks" `build.sh`, doesn't it?

tools/docker/tests/test_flows.sh
tomereli: I now see that e74d848c7e71dc3bba8275f5a3db2c1436885cc3 fixes that but IMHO it should be squashed to the commit before that, otherwise this commit "breaks" `build.sh`, doesn't it? In that case I think it would be better to squash them so build will not break.

rmelotte: 2c435b4a indeed does break things, it has to be applied with the next one (e74d848c).
I remember @arnout mentioned he prefers to have "generated" commits separate from their manual fixes, but I don't know if it applies in this case.

rmelotte: I asked @arnout, the reason to have them split at first is to make it easier to review.
Since you both already did, I'll squash them now.

PR #564, opened 2019-12-11
RDKB build support

vitalybu:
Add support for out-of-tree RDKB build

### DIFF
--- a/cmake/multiap-helpers.cmake
+++ b/cmake/multiap-helpers.cmake
@@ -18,6 +18,17 @@ if(NOT WIN32)
   set(BoldWhite   "${Esc}[1;37m")
 endif()
 
+# Force the compiler to produce ANSI-colored output when using ninja build
+# https://github.com/ninja-build/ninja/wiki/FAQ
+option (FORCE_COLORED_OUTPUT "Always produce ANSI-colored output (GNU/Clang only)." ON)
+if (${FORCE_COLORED_OUTPUT})
+    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+       add_compile_options (-fdiagnostics-color=always)
+    elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+       add_compile_options (-fcolor-diagnostics)
+    endif ()
+endif ()
+
 if (NOT ${PROJECT}_REVISION)
     execute_process(
         COMMAND "git" "rev-parse" "HEAD"
--- a/tools/cmake/toolchain/rdkb.cmake
+++ b/tools/cmake/toolchain/rdkb.cmake
@@ -1,0 +1,73 @@
+## General configuration
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_EXPORT_NO_PACKAGE_REGISTRY ON)
+
+# Workaround for https://www.cmake.org/Bug/view.php?id=14075
+set(CMAKE_CROSS_COMPILING ON)
+
+# Platform specific
+set(TARGET_PLATFORM "rdkb" CACHE STRING "Target Platform")
+
+# Base RDKB build dir (e.g. /path/to/atom_rdkbos)
+if (NOT DEFINED ENV{PRPLMESH_PLATFORM_BASE_DIR})
+    message(FATAL_ERROR "PLATFORM_BASE_DIR environment variable not defined!")
+endif()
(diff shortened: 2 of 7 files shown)


tools/cmake/toolchain/rdkb.cmake
```diff
@@ -0,0 +1,80 @@
+## General configuration
+set(CMAKE_SYSTEM_NAME Linux)
+set(CMAKE_SYSTEM_PROCESSOR ARMV7 CACHE STRING "System Processor")
```
tomereli: Do we need it to be ARMV7?

tools/cmake/toolchain/rdkb.cmake
```diff
@@ -0,0 +1,80 @@
...
+set(PLATFORM_INCLUDE_DIR        ${PLATFORM_STAGING_DIR}/usr/include CACHE STRING "Platform Include Directory")
+set(ENV{PKG_CONFIG_PATH}        "${PLATFORM_STAGING_DIR}/usr/lib/pkgconfig")
+
+# Compiler Definitions
+add_definitions(
+    -DBEEROCKS_RDKB
+)
+
+## Library config
+set(BUILD_SHARED_LIBS ON CACHE BOOL "BUILD_SHARED_LIBS" FORCE)
```
tomereli: I recall that @arnout said these are not needed.

cmake/multiap-helpers.cmake
```diff
@@ -18,6 +18,17 @@ if(NOT WIN32)
   set(BoldWhite   "${Esc}[1;37m")
 endif()
 
+# Force the compiler to produce ANSI-colored output when using ninja build
```
tomereli, line 21: Nice!

tomereli (approved):

morantr (approved):

tools/cmake/toolchain/rdkb.cmake
vitalybu: Nope, I'll remove it

tools/cmake/toolchain/rdkb.cmake
vitalybu: Will remove that as well

tools/commands/build.py
```diff
@@ -196,9 +196,12 @@ def __init__(self, args):
         modules_dir = os.path.realpath(args.map_path)
 
         map_cmake_flags = args.cmake_flags
-        if not args.native:
-            cmake_toolchain_file = os.environ.get('CMAKE_TOOLCHAIN_FILE', 'external_toolchain.cmake')
-            map_cmake_flags = ["CMAKE_TOOLCHAIN_FILE=" + cmake_toolchain_file] + map_cmake_flags
+        if not args.native: 
+            if "CMAKE_TOOLCHAIN_FILE" in os.environ:
+                map_cmake_flags += ["CMAKE_TOOLCHAIN_FILE=" + os.path.abspath(os.environ.get('CMAKE_TOOLCHAIN_FILE'))]
```
arnout, line 201: Weird, I was convinced I had merged this exact change some time ago...

arnout (changes requested):
The Dockerfile is not the approach we want to take, because it still requires you to have the whole RDKB build on the host, so there is not much of an advantage of building it in Docker. Worse, if your host is not Ubuntu 18.04, it may fail in weird ways. See #545 for the proper approach. Note that #545 is not finished yet. So I'm OK with merging this RDKB Dockerfile as is for the moment; just be warned that I'm planning to kill it :-)

Do not make merge commits; rebase on master instead of merging it.

Otherwise LGTM. Still changes requested because of the merge commit.

vitalybu:
> The Dockerfile is not the approach we want to take, because it still requires you to have the whole RDKB build on the host, so there is not much of an advantage of building it in Docker. Worse, if your host is not Ubuntu 18.04, it may fail in weird ways. See #545 for the proper approach. Note that #545 is not finished yet. So I'm OK with merging this RDKB Dockerfile as is for the moment; just be warned that I'm planning to kill it :-)
> 
> Do not make merge commits; rebase on master instead of merging it.
> 
> Otherwise LGTM. Still changes requested because of the merge commit.

I agree that having the entire toolchain and staging directory inside a docker is a much better approach.
The purpose of this dockerfile is to remove the dependency on the build host (for example, one can run an Ubuntu 18.04 docker on Ubuntu 14.04 without issues).

As for the merge commit, it was automatically applied by GitHub once it detected that the branch is out of date (with a press of a button), but I'll remove it

tools/commands/build.py
vitalybu: ¯\\_(ツ)_/¯

arnout (approved):

PR #566, opened 2019-12-11
Feature/tlvf print error on init failure

tomereli:
Till now, whenever buffPtrIncrementSafe() failed, it returned "silently"
without logging the failure so it was rather difficult to find out where
the error is.

Since this is autogenerated code, it is easy to add explicit error log
to all these cases - so add them in tlvf.py.

### DIFF
--- a/common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_1905_vs.cpp
+++ b/common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_1905_vs.cpp
@@ -80,22 +80,43 @@ bool tlvVsClientAssociationEvent::init()
         return false;
     }
     m_mac = (sMacAddr*)m_buff_ptr__;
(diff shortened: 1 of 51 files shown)


common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_1905_vs.cpp
```diff
@@ -80,22 +80,43 @@ bool tlvVsClientAssociationEvent::init()
         return false;
     }
     m_mac = (sMacAddr*)m_buff_ptr__;
-    if (!buffPtrIncrementSafe(sizeof(sMacAddr))) { return false; }
+    if (!buffPtrIncrementSafe(sizeof(sMacAddr))) {
+        LOG(ERROR) << "buffPtrIncrementSafe(" << int(sizeof(sMacAddr)) << ") Failed!";
```
arnout: Why the cast to int? It shouldn't be needed, sizeof() is a size_t type.

arnout (approved):

morantr (approved):

PR #567, opened 2019-12-11
Bugfix/fix verify cmdu

tomereli:
verify_cmdu is using the tlvVendorSpecific constructor directly instead
of using addClass, so it doesn't check for isInitialized() like done in
addClass method of CmduMessageTx.

The length used for the tlvVendorSpecific constructor is wrong - it is
the length of the TLV without the TLV header length, and therefore can
fail. It never failed till now becuase the length of the TLV was set to
much more than the size of the TLV defined in the yaml file - TLV vendor
specific TLVs simply has extranous data after the OUI which is the
Action header and the actual message class.

When moving to TLVlist, we change the yaml to include an unknown length
list which will be used as an internal TLVlist, which will cause this to
fail.


arnout (approved):
Excellent commit message, thanks!

PR #569, opened 2019-12-12
Load bwl without dlopen

arnout:
Loading bwl with dlopen serves no purpose at all.
Link it as a normal library instead.


arnout:
Required for #469.

arnout:
I only did it for monitor now (hence the draft-ness of this PR). If it looks OK, I'll do it for the other two bwl and the bpl as well.

tomereli (approved):

tomereli:
LGTM

vitalybu (commented):
Why do we need the create functions at all?
Can instantiate the classes directly.

vitalybu (commented):
Looks fine to me. What about the Agent (AP Manager)?

arnout:
[Support static linking](https://trello.com/c/gjsv6UWn/35-support-static-linking)

arnout:
> Why do we need the create functions at all?
> Can instantiate the classes directly.

The functions are there to abstract the subclasses, otherwise the caller would have to know which bwl type to instantiate.

The subclasses are not strictly needed; the different implementations could instead all instantiate the same class. There are a few advantages of putting the implementations in unique classes:

- it makes it easy to put the common functionality in the base class;
- it makes it somewhat easier for an IDE to include the different implementations in a single project.

None of these advantages are very strong though so I would definitely be OK with changing this and renaming mon_wlan_hal_dwpal to mon_wlan_hal. But that is a *much* bigger change, and I think the benefit is not big enough.

arnout:
[Support static linking](https://trello.com/c/gjsv6UWn/35-support-static-linking)

tomereli (approved):

PR #571, opened 2019-12-15
cmake: building for generic openwrt targets

dob71:
- Removed remaining refences to turris-omnia
- Added ability to customize interface names and BWL_TYPE for generic
  openwrt targets using the following environment variables:
    PRPLMESH_WLAN1_IFACE (if not set default is wlan0)
    PRPLMESH_WLAN2_IFACE (if not set default is wlan1)
    PRPLMESH_BH_WIRE_IFACE (if not set default is eth1)
    PRPLMESH_BWL_TYPE (if not set default is NL80211)

Note: it's no longer needed to set PRPLMESH_PLATFORM_TYPE for turris-omnia
      or any other openwrt target except ugw, but the changes are backward
      compatible and the build should still work if the above environment
      variable is set.


tomereli (changes requested):
Why is it preferred to have these as environment variables instead of simple cmake variables? I'm asking since in the OpenWrt Makefile we will now need to define environment variables instead of simple cmake variables which I'd prefer to avoid.
This should be the usage in the OpenWrt feed:

```
CMAKE_OPTIONS += \
-DBEEROCKS_WLAN1_IFACE=wlan0 \
-DBEEROCKS_WLAN2_IFACE=wlan2
-DBWL_TYPE=NL89211

```

So I don't see why we need all these environment variables. 

These are used for out of tree build with tool chain file, which uses env vars and sets cmake vars.

rmelotte (changes requested):
I agree with what @tomereli said. It should be possible to use cmake variables directly?
Also (I might be wrong) I don't think new variables names are required, you could reuse the existing ones.

We need a way to change them using the "-D" option though.
You could either:
- use set(...CACHE) to cache them (making it possible to change them using "-D").
- check if they are already defined:
```
if (NOT DEFINED BEEROCKS_WLAN1_IFACE)
  set(BEEROCKS_WLAN1_IFACE "wlan2")
endif()
```
I don't know if there is any specific reason why they were not using CACHE already, since it's already the case for the target platform and so on. Could anyone else comment on that?

Also since this PR is about making it possible to change the interface names, I suggest you also grep for the remaining places where they are hard-coded (grep for `wlan0` for example: it's still present in `prplmesh_utils.sh.in`) and change them to use the cmake variables (only for the places which are already generated by cmake, so only the `.in` files).

arnout:
> I don't know if there is any specific reason why they were not using CACHE already, since it's already the case for the target platform and so on. Could anyone else comment on that?

Because the original philosophy was that these variables were "internal" variables that are set based on PLATFORM. We are now changing our mind and saying that the platform's build system should specify those details and there should not be a PLATFORM at all.

Note that specifically for the interface names, we don't need to bother too much with the details since they're going to be removed entirely by #574.

> Also since this PR is about making it possible to change the interface names, I suggest you also grep for the remaining places where they are hard-coded (grep for `wlan0` for example: it's still present in `prplmesh_utils.sh.in`) and change them to use the cmake variables (only for the places which are already generated by cmake, so only the `.in` files).

:+1:

CMakeLists.txt
```diff
@@ -69,8 +69,23 @@ if (TARGET_PLATFORM STREQUAL "openwrt")
...
+        if (NOT DEFINED ENV{PRPLMESH_WLAN2_IFACE})
+            set(BEEROCKS_WLAN2_IFACE "wlan1")
+        else()
+            set(BEEROCKS_WLAN2_IFACE "$ENV{PRPLMESH_WLAN2_IFACE}")
+        endif()
+        if (NOT DEFINED ENV{PRPLMESH_BH_WIRE_IFACE})
+            set(BEEROCKS_BH_WIRE_IFACE "eth1")
+        else()
+            set(BEEROCKS_BH_WIRE_IFACE "$ENV{PRPLMESH_BH_WIRE_IFACE}")
+        endif()
```
arnout: Since all of this will anyway be deprecated by #574 I think we should go for the simple solution:

```
... ugw stuff ...
else()
    set(BEEROCKS_WLAN2_IFACE "wlan1")
endif()
```

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -27,7 +27,12 @@ elseif(TARGET_PLATFORM STREQUAL "openwrt")
     if (TARGET_PLATFORM_TYPE STREQUAL "ugw")
         set(BWL_TYPE "DWPAL")
     else()
-        set(BWL_TYPE "NL80211")
+        # For gneric openwrt target use NL80211 interface by default
+        if (NOT DEFINED ENV{PRPLMESH_BWL_TYPE})
+            set(BWL_TYPE "NL80211")
+        else()
+            set(BWL_TYPE "$ENV{PRPLMESH_BWL_TYPE}")
+        endif()
```
arnout: Here, I think the existing code already works, so this change is not needed.

Long term, we want BWL_TYPE to be a cache variable and only set it based on PLATFORM if it is not defined already. Longer term, we'll want to remove PLATFORM entirely.

framework/platform/bpl/CMakeLists.txt
```diff
@@ -47,11 +47,9 @@ if (TARGET_PLATFORM STREQUAL "openwrt")
         add_definitions(-DHAVE_C99)
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c99")
 
-    elseif (TARGET_PLATFORM_TYPE STREQUAL "turris-omnia")
-
-        # Platform definitions
-        add_definitions(-DBEEROCKS_TURRIS_OMNIA)
+    else()
 
+        # Generic openwrt platform definitions
```
arnout: The comment "Platform defintiions" was about the add_definitions statement, so this comment is not correct. Just remove it.

tools/cmake/toolchain/openwrt.cmake
```diff
@@ -7,7 +7,11 @@ set(CMAKE_CROSS_COMPILING ON)
 
 # Platform specific
 set(TARGET_PLATFORM             "openwrt" CACHE STRING "Target Platform")
+if (DEFINED ENV{PRPLMESH_PLATFORM_TYPE})
 set(TARGET_PLATFORM_TYPE        "$ENV{PRPLMESH_PLATFORM_TYPE}" CACHE STRING "Target Platform Type")
+else()
+set(TARGET_PLATFORM_TYPE        "generic" CACHE STRING "Target Platform Type")
+endif()
```
arnout: As you mention in the commit message, after this change only "ugw" is used in TARGET_PLATFORM_TYPE. So there's no need to define it as generic, you can just as well leave it empty. So I don't think this change is needed.

arnout (changes requested):

CMakeLists.txt
dob71: Do you mean revert it to what it was, just set BEEROCKS_WLAN2_IFACE to "wlan1"?
At minimum I need to be able to set wired backhaul to eth0 fro Gl.Inet and  keep it eth1 for Turris. I don't mind removing all the wlan# interface stuff if it will no longer be needed soon.

common/beerocks/bwl/CMakeLists.txt
dob71: Ok, will remove

framework/platform/bpl/CMakeLists.txt
dob71: Ok

tools/cmake/toolchain/openwrt.cmake
dob71: I was thinking that it might become useful for printing a message, but sure can remove.

dob71 (commented):
@tomereli The changes were made to allow compiling prplMesh separately outside of the openwrt tree per the instructions on the WiKi pages (following the same pattern that was used for Turris): 
https://github.com/prplfoundation/prplMesh/wiki/prplMesh-on-Gl.Inet-B1300#cross-compile-prplmesh-for-glinet-b1300
The environment vars were just easier to use in that case.

tools/cmake/toolchain/openwrt.cmake
```diff
@@ -7,7 +7,9 @@ set(CMAKE_CROSS_COMPILING ON)
 
 # Platform specific
 set(TARGET_PLATFORM             "openwrt" CACHE STRING "Target Platform")
+if (DEFINED ENV{PRPLMESH_PLATFORM_TYPE})
 set(TARGET_PLATFORM_TYPE        "$ENV{PRPLMESH_PLATFORM_TYPE}" CACHE STRING "Target Platform Type")
+endif()
```
arnout: Why is this condition needed? If TARGET_PLATFORM_TYPE is not defined in the environment, it will just be defined to empty string and that's OK, no?

dob71: I don't know cmake, so trying to stay safe, I'll try what happens if it's left undefined and if it works  remove the condition.

arnout (approved):
I squashed everything into a single commit and updated the commit message a bit.

tomereli (approved):

rmelotte (changes requested):
Typo in the commit message: `PRPLMESH_PLATFORM_TYPE` should be `TARGET_PLATFORM_TYPE`.
Otherwise, LGTM.

arnout:
> `PRPLMESH_PLATFORM_TYPE`

Fixed.

If you only have such a minor comment, please approve the PR already so it can be merged without your re-review. You can use the new "don't merge" tag to indicate that it still needs some fixup before merge (anybody can remove that tag, while admin privilege is required to override a reviewer's "changes requested").

rmelotte:
> If you only have such a minor comment, please approve the PR already so it can be merged without your re-review. You can use the new "don't merge" tag to indicate that it still needs some fixup before merge (anybody can remove that tag, while admin privilege is required to override a reviewer's "changes requested").

Alright, will do. I was thinking maybe anyone could bypass/reset my "change request" because I'm not in the code owners.

rmelotte (approved):

PR #572, opened 2019-12-15
tlvf: CmduMessageTx ->TlvList refactor

tomereli:
Currently, CmduMessage (and its derived classes CmduMessageTx & CmduMessageRx) include the TLV class vector and the buffer.

Some TLVs, such as vendor-specific and WSC, may contain arbitrary data that can be parsed/created with the same mechanics that we use in the CmduMessage and are required by multiple features (#471, #160, #318).

It would be easier to implement support for such TLVs using a TLV list class, which will include all the buffer handling that is currently part of the CmduMessage classes.

This PR refactors TLVF base files and adds a new class, `TlvList`, which can be used to create and parse TLVs with arbitrary data represented with an internal TLV list.

It also adds support for the first outer TLV, the vendor-specific TLV, by adding an unknown length buffer to the end of its YAML file, which is used as the buffer for the inner TLV list.
The inner TLV list is the `beerocks_header` which derives from `TlvList` and adds some backward compatible APIs so the change will not be too big to swallow in a single PR (Nevertheless, this kind of change is pretty big since in order for not breaking the build in each commit).

Now, each vendor-specific CMDU is built using a vendor-specific TLV with a payload buffer (`uint8_t`).
The buffer size is preallocated in`add_intel_vs_data()` to the maximum allowed TLV size - which is `MTU - CMDU header length - EOM TLV length - TS TLV init length` for standard CMDUs since the ieee1905 transport performs fragmentation on TLV boundary and the maximum buffer size for CMDUs containing oversized TLVs (referred to as "jumbo" CMDUs) which are not meant to be sent externally.
Having a preallocated big enough buffer allows the internal TLV list to accommodate variable-length data in the internal classes, which is mandatory since by the time we are allocating the buffer we do not know if the caller will need more space.

This brings the need to resize the inner TLVs and update the outer TLVs length on `finalize()`, which does so by going through the CMDU's inner tlv map, resizing each inner TLVlist to its final message length and moving the rest of the CMDU buffer back to the new end of the inner TLV list.

### DIFF
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -209,8 +209,7 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
             if (tlv_vendor_specific.vendor_oui() ==
                 ieee1905_1::tlvVendorSpecific::eVendorOUI::OUI_INTEL) {
                 // assuming that the magic is the first data on the beerocks header
-                auto beerocks_magic = *(
-                    uint32_t *)((uint8_t *)tlv + ieee1905_1::tlvVendorSpecific::get_initial_size());
+                auto beerocks_magic = *(uint32_t *)(tlv_vendor_specific.payload());
                 if (uds_header->swap_needed) {
                     swap_32(beerocks_magic);
                 }
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -23,8 +23,9 @@ using namespace beerocks;
 
 socket_thread::socket_thread(const std::string &unix_socket_path_)
     : thread_base(), cmdu_tx(TX_BUFFER_UDS, TX_BUFFER_UDS_SIZE),
-      unix_socket_path(unix_socket_path_), server_socket(nullptr),
-      server_max_connections(DEFAULT_MAX_SOCKET_CONNECTIONS)
+      unix_socket_path(unix_socket_path_), cmdu_rx(rx_buffer + sizeof(message::sUdsHeader),
+                                                   sizeof(rx_buffer) - sizeof(message::sUdsHeader)),
+      server_socket(nullptr), server_max_connections(DEFAULT_MAX_SOCKET_CONNECTIONS)
 {
     memset(TX_BUFFER_UDS, 0, TX_BUFFER_UDS_SIZE);
     set_select_timeout(500);
@@ -162,8 +163,7 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
         return false;
     }
 
-    if (!cmdu_rx.parse(rx_buffer + sizeof(message::sUdsHeader), uds_header->length,
-                       uds_header->swap_needed)) {
+    if (!cmdu_rx.parse(uds_header->swap_needed)) {
         THREAD_LOG(ERROR) << "parsing cmdu failure, rx_buffer" << std::hex << rx_buffer << std::dec
                           << ", uds_header->length=" << int(uds_header->length)
                           << ", uds_header->swap_needed=" << int(uds_header->swap_needed);
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -194,7 +194,7 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
         swap_16(length);
         if (static_cast<ieee1905_1::eTlvType>(type) == ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC) {
             auto tlv_vendor_specific = ieee1905_1::tlvVendorSpecific(
-                (uint8_t *)tlv, length + sizeof(ieee1905_1::TlvHeader), true, true);
+                (uint8_t *)tlv, length + sizeof(ieee1905_1::TlvHeader), true);
             if (!tlv_vendor_specific.isInitialized()) {
                 LOG(ERROR) << "tlvVendorSpecific init() failure";
                 ret = false;
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -17,11 +17,6 @@
 
 using namespace beerocks;
 
-typedef struct sTlvHeader {
-    uint8_t type;
-    uint16_t length;
-} __attribute__((packed)) sTlvHeader;
-
 #define DEFAULT_MAX_SOCKET_CONNECTIONS 10
 #define TX_BUFFER_UDS (tx_buffer + sizeof(beerocks::message::sUdsHeader))
 #define TX_BUFFER_UDS_SIZE (sizeof(tx_buffer) - sizeof(beerocks::message::sUdsHeader))
@@ -185,10 +180,10 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
 // FIXME - WLANRTSYS-6360 - should be moved to transport
 bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
 {
-    sTlvHeader *tlv =
-        reinterpret_cast<sTlvHeader *>((uint8_t *)uds_header + sizeof(message::sUdsHeader) +
-                                       ieee1905_1::cCmduHeader::get_initial_size());
-    sTlvHeader *first_tlv = tlv;
+    ieee1905_1::TlvHeader *tlv = reinterpret_cast<ieee1905_1::TlvHeader *>(
+        (uint8_t *)uds_header + sizeof(message::sUdsHeader) +
+        ieee1905_1::cCmduHeader::get_initial_size());
+    ieee1905_1::TlvHeader *first_tlv = tlv;
 
     bool ret = true;
 
@@ -203,7 +198,8 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
 
         if (static_cast<ieee1905_1::eTlvType>(type) == ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC) {
             auto tlv_vendor_specific = ieee1905_1::tlvVendorSpecific(
-                (uint8_t *)tlv, length + sizeof(sTlvHeader), true, uds_header->swap_needed);
+                (uint8_t *)tlv, length + sizeof(ieee1905_1::TlvHeader), true,
+                uds_header->swap_needed);
             if (!tlv_vendor_specific.isInitialized()) {
                 LOG(ERROR) << "tlvVendorSpecific init() failure";
                 ret = false;
@@ -241,7 +237,7 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
         }
 
         // set the next tlv
-        tlv    = (sTlvHeader *)((uint8_t *)tlv + sizeof(sTlvHeader) + length);
+        tlv    = (ieee1905_1::TlvHeader *)((uint8_t *)tlv + sizeof(ieee1905_1::TlvHeader) + length);
         type   = tlv->type;
         length = tlv->length;
 
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -163,10 +163,9 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
         return false;
     }
 
-    if (!cmdu_rx.parse(uds_header->swap_needed)) {
+    if (!cmdu_rx.parse()) {
         THREAD_LOG(ERROR) << "parsing cmdu failure, rx_buffer" << std::hex << rx_buffer << std::dec
-                          << ", uds_header->length=" << int(uds_header->length)
-                          << ", uds_header->swap_needed=" << int(uds_header->swap_needed);
+                          << ", uds_header->length=" << int(uds_header->length);
         return false;
     }
 
@@ -192,14 +191,10 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
 
     do {
 
-        if (uds_header->swap_needed) {
-            swap_16(length);
-        }
-
+        swap_16(length);
         if (static_cast<ieee1905_1::eTlvType>(type) == ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC) {
             auto tlv_vendor_specific = ieee1905_1::tlvVendorSpecific(
-                (uint8_t *)tlv, length + sizeof(ieee1905_1::TlvHeader), true,
-                uds_header->swap_needed);
+                (uint8_t *)tlv, length + sizeof(ieee1905_1::TlvHeader), true, true);
             if (!tlv_vendor_specific.isInitialized()) {
                 LOG(ERROR) << "tlvVendorSpecific init() failure";
                 ret = false;
@@ -210,9 +205,7 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
                 ieee1905_1::tlvVendorSpecific::eVendorOUI::OUI_INTEL) {
                 // assuming that the magic is the first data on the beerocks header
                 auto beerocks_magic = *(uint32_t *)(tlv_vendor_specific.payload());
-                if (uds_header->swap_needed) {
-                    swap_32(beerocks_magic);
-                }
+                swap_32(beerocks_magic);
 
                 if (beerocks_magic != message::MESSAGE_MAGIC) {
                     THREAD_LOG(WARNING) << "mismatch magic " << std::hex << int(beerocks_magic)
@@ -225,10 +218,8 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
                 THREAD_LOG(INFO) << "Not an Intel vendor specific message!";
             }
 
-            if (uds_header->swap_needed) {
-                // cancel the swap we did
-                tlv_vendor_specific.class_swap();
-            }
+            // cancel the swap we did
+            tlv_vendor_specific.class_swap();
         } else if (static_cast<ieee1905_1::eTlvType>(type) ==
                        ieee1905_1::eTlvType::TLV_END_OF_MESSAGE &&
                    length == 0) {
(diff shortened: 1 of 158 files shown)


tomereli:
[tlvList refactoring](https://trello.com/c/JHrwp6IN/19-tlvlist-refactoring)

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
```diff
@@ -0,0 +1,66 @@
...
+        if (m_finalized)
+            return true;
+
+        if (swap_needed)
+            swap();
+
+        m_finalized = true;
+        return true;
+    }
+};
```
CoralMalachi: I did not notice this before, but since this method overrides the original finalize method, shouldn't we add here the post init check also?

tomereli: You are correct. Please add a fix up commit to fix this.

framework/tlvf/src/include/tlvf/BaseClass.h
```diff
@@ -38,6 +37,19 @@ class BaseClass {
...
+     * @brief trim buffer by tailroom size
+     *
+     * This is intended to be used __only__ by friend class TlvList on finalize
+     * When resizing buffers.
+     *
+     * @param tailroom size to trim
+     * @return true on success
+     * @return false on failure
+     */
+    friend void trim(BaseClass &obj, size_t tailroom) { obj.m_buff_ptr__ -= tailroom; };
```
CoralMalachi, line 50: A friend function of a class is defined outside of a class scope, so I think we should use the friend keyword in the TlvList class, and declare the trim method as a friend method of TlvList class, so it has the right to access all private and protected members of the class and use this method.

CoralMalachi: Nevermind, that is not going to work since `trim` is a private class method, but still making `trim` a friend method is wrong since it can be used by other classes, for example, CmduMessageTx, and that was not Tomer's intention.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -17,19 +17,15 @@
...
-
 #define DEFAULT_MAX_SOCKET_CONNECTIONS 10
 #define TX_BUFFER_UDS (tx_buffer + sizeof(beerocks::message::sUdsHeader))
 #define TX_BUFFER_UDS_SIZE (sizeof(tx_buffer) - sizeof(beerocks::message::sUdsHeader))
 
 socket_thread::socket_thread(const std::string &unix_socket_path_)
     : thread_base(), cmdu_tx(TX_BUFFER_UDS, TX_BUFFER_UDS_SIZE),
-      unix_socket_path(unix_socket_path_), server_socket(nullptr),
-      server_max_connections(DEFAULT_MAX_SOCKET_CONNECTIONS)
+      unix_socket_path(unix_socket_path_), cmdu_rx(rx_buffer + sizeof(message::sUdsHeader),
```
arnout, line 26: This change should be explained in the commit message (why it is needed, why it worked without initialisation before). Possibly in a separate commit.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
```diff
@@ -0,0 +1,70 @@
...
+    std::shared_ptr<beerocks_message::cACTION_HEADER> actionhdr()
+    {
+        return getClass<beerocks_message::cACTION_HEADER>();
+    }
+    beerocks_message::eAction action()
+    {
+        return actionhdr() ? actionhdr()->action() : beerocks_message::ACTION_NONE;
+    }
+    uint8_t action_op() { return actionhdr() ? actionhdr()->action_op() : 0; };
+    uint16_t id() { return actionhdr() ? actionhdr()->id() : 0; }
```
arnout, line 44: All these methods should be const.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
```diff
@@ -0,0 +1,70 @@
...
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _TLVF_BEEROCKS_HEADER_H_
+#define _TLVF_BEEROCKS_HEADER_H_
+
+#include <beerocks/tlvf/beerocks_message_header.h>
+#include <tlvf/TlvList.h>
+#include <tlvf/tlvflogging.h>
```
arnout, line 14: Is this header used?

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -47,46 +47,49 @@ class message_com {
...
     parse_intel_vs_message(ieee1905_1::CmduMessageRx &cmdu_rx);
 
-    template <class T> static std::shared_ptr<T> get_vs_class(ieee1905_1::CmduMessage &cmdu)
-    {
-        return std::dynamic_pointer_cast<T>(cmdu.getClass(2));
-    }
-
-    static std::shared_ptr<beerocks_message::cACTION_HEADER>
-    get_vs_class_header(ieee1905_1::CmduMessage &cmdu)
+    static std::shared_ptr<beerocks_header> get_beerocks_header(ieee1905_1::CmduMessage &cmdu)
```
arnout, line 53: This get_beerocks_header() function smells funny to me...

The pattern seems to be that you do an `add_vs_tlv` which creates both the beerocks_header and the specific action object, but only returns the specific action object. Maybe it would be easier to directly return a `std::pair<std::shared_ptr<beerocks_header>, std::shared_ptr<T>>`?

It probably doesn't cover all the cases, but I feel it's a more sensible approach.

Another possibility is to go for a completely different approach. A beerocks TLV is always a VS TLV with a beerocks_action and a T. So we could have a beerocks_vs_tlv template class that has these two as members which can be accessed directly. And on the receive side we could still use beerocks_header as it is now (for receive it works fine).

(nitpick probably, or something that can be refactored later)

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -47,46 +47,49 @@ class message_com {
...
     {
-        return std::dynamic_pointer_cast<beerocks_message::cACTION_HEADER>(cmdu.getClass(1));
+        return cmdu.tlvs.getInnerTlvList<beerocks_header>(
+            ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC);
     }
 
     template <class T>
-    static std::shared_ptr<T> add_intel_vs_data(ieee1905_1::CmduMessageTx &cmdu_tx,
-                std::shared_ptr<ieee1905_1::tlvVendorSpecific> vs_tlv, uint16_t id = 0)
+    static bool add_intel_vs_data(ieee1905_1::CmduMessageTx &cmdu_tx, uint16_t id = 0)
```
arnout, line 60: I think it would be better to inline this directly in add_vs_tlv, so you can use the return values of the addClass calls instead of doing a getClass (and checking return value *again*) afterwards.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
```diff
@@ -0,0 +1,70 @@
...
+public:
+    beerocks_header() = delete;
+    beerocks_header(uint8_t *buff, size_t buff_len, bool parse, bool swap)
+        : TlvList(buff, buff_len, parse, swap)
+    {
+    }
+    std::shared_ptr<beerocks_message::cACTION_HEADER> actionhdr()
+    {
+        return getClass<beerocks_message::cACTION_HEADER>();
+    }
```
arnout, line 38: I don't know if it's worth it, but we could instead define

```
beerocks_message::cACTION_HEADER *operator->()
{
    auto header = getClass<beerocks_message::cACTION_HEADER();
    if (header)
        return header;
    else
        return nullptr;
}
```

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -103,27 +106,58 @@ class message_com {
...
+     * be only sent internally over UDS by setting the maximum TLV size
+     * of the CMDU to be bigger than the MTU, based on the actual buffer size.
+     *
+     * @tparam T beerocks operation template type
+     * @param cmdu_tx Transmit CMDU descriptor
+     * @param id message id
+     * @return std::shared_ptr<T> to the created operation class
+     */
+    template <class T>
+    static std::shared_ptr<T> create_jumbo_vs_message(ieee1905_1::CmduMessageTx &cmdu_tx,
```
arnout, line 140: I don't like that we need to add all this special-case code (this function, max_tlv_length() and all related stuff) for something that should be rather exceptional (an internal VS message that is larger than 1450-ish bytes). But OK, I guess there's a really good reason to need this (I first thought that it was the conn_map, but that one already splits over several VS messages so it has to be something else).

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -103,27 +106,58 @@ class message_com {
...
+            return false;
         }
 
-        actionhdr->action()    = (beerocks_message::eAction)action;
-        actionhdr->action_op() = action_op;
-        actionhdr->id()        = id;
+        hdr->actionhdr()->action()    = (beerocks_message::eAction)action;
+        hdr->actionhdr()->action_op() = action_op;
+        hdr->actionhdr()->id()        = id;
+        cmdu_tx.tlvs.addInnerTlvList(ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC, hdr);
```
arnout, line 119: (I briefly discussed this orally with @tomereli)

What this call essentially does is to register a finalization callback - the inner tlv list is only really used in finalize (it is also used by get_beerocks_header but that's just a shorthand for passing through `getClass<ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC>`).

So maybe we should make this more explicit and instead call it setFinalizeCallback with a callable as argument. The finalize callback can be called in addClass before actually adding the class, so the resizing can be done then. Unfortunately, a showstopper for this is the swap_needed argument, which we don't know yet in addClass...

framework/tlvf/src/include/tlvf/TlvList.h
```diff
@@ -0,0 +1,182 @@
...
+#include <tlvf/ieee_1905_1/eTlvType.h>
+#include <unordered_map>
+#include <vector>
+
+namespace ieee1905_1 {
+
+typedef struct TlvHeader {
+    uint8_t type;
+    uint16_t length;
+} __attribute__((packed)) TlvHeader;
```
arnout, line 24: Not good. TlvList should be a generic class that can be used for a list of ieee1905_1 TLVs but also for a list of WSC attributes. You can't make assumptions on the layout of the header. Unless you make it a template class which takes the header as template argument.

framework/tlvf/src/include/tlvf/TlvList.h
```diff
@@ -0,0 +1,182 @@
...
+    size_t getMessageLength() const;
+    size_t getMessageBuffLength() const;
+    uint8_t *getMessageBuff() const;
+    void swap();
+    bool swap_needed() { return m_swap; }
+    bool is_finalized() const { return m_finalized; };
+    bool is_swapped() const { return m_swapped; };
+    virtual bool finalize(bool swap_needed);
+    void reset(bool parse, bool swap);
+    void addInnerTlvList(eTlvType type, std::shared_ptr<TlvList> list)
```
arnout, line 109: Not good, we can't rely on eTlvType here because it's specific for ieee1905_1.

framework/tlvf/src/src/CmduMessage.cpp
```diff
@@ -10,149 +10,76 @@
...
 
-std::shared_ptr<BaseClass> CmduMessage::getClass(size_t idx) const
+std::shared_ptr<cCmduHeader> CmduMessage::getCmduHeader() const
 {
-    if (m_class_vector.size() > idx) {
-        return std::shared_ptr<BaseClass>(m_class_vector.at(idx));
-    } else {
-        return nullptr;
-    }
+    return tlvs.getClass<cCmduHeader>();
```
arnout, line 25: No this can't be right. The CMDU header can't be part of the list of TLVs, it's not a TLV. It doesn't start with a type and length.

framework/tlvf/src/include/tlvf/BaseClass.h
arnout, line 50: I think the trim is not appropriate here. I think it would make more sense to make a special case in tlvf for when an unknown-length list is at the end of the class (or maybe even make a special _payload type for this). In that specific case, we could add two functions: a prepare that reserves the rest of the buffer and returns the remaining length, and a commit with the final length as argument that updates the end pointer and the length field. Both functions are pretty much the same as allocate (the trim is like allocate with a negative value) in that they update both length and end pointer.

arnout (changes requested):
I did a "quick" review to give feedback if I thought this was going in the right direction. Unfortunately, the answer is "not entirely". I'm not convinced of the solution for updating the length of inner tlv lists, it feels too complicated and too error-prone to me. That said, we *may* decide to merge it like this anyway and refactor later.

To make some progress, I would like a separate PR with the following commits:

- 41d99bcc tlvf: Fix bug in BuffPtrIncrementSafe()
- 4531edde remove tlvf & btlvf src from .clang-ignore
- e80a2ce1 tlvf_conf.yaml: Remove include_source_path (note that the commit sha referenced in the commit message is not correct)
- cda27dbe common: beerocks_message: remove unclear TODO
- b7792fa6 tlvf: Delete dynamicCast API from TlvList


And maybe a second PR with:

- f5bacbbf framework: tlvf: add TlvList class
- 3a1c3f47 tlvf: add payload to vendor specific messages

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: Perhaps we can just set swap_needed=True always, no matter for which message, then here we'll assume that swap_needed is true.

framework/tlvf/src/include/tlvf/TlvList.h
tomereli: I think we needn't be too generic - outer TLVs, which are the only TLV types represented by a TlvList are 1905_1 TLVs - I don't see any future use for any other type. Inner TLVs are different, for those we don't need and don't use the TLV Header.

framework/tlvf/src/include/tlvf/TlvList.h
tomereli: As I commented before, IMO outer TLVs are 1905 TLVs so we can rely on eTlvType.

framework/tlvf/src/src/CmduMessage.cpp
tomereli: IMHO A TLV list is just a wrapper for TLVF classes which do not necessarily need to be TLVs. With this assumption, and the fact that an outer TLV list is only used for 1905 TLVs we can decide that the first element is always the CMDU header. I think it makes the code simpler.

tomereli: beerocks_header derives from TlvList and its classes also don't have type and length. Maybe a better name would be TlvfClassList since it's a container for TLVF classes more than it is a TLV list. Then if we really want to, we can always extend it and create a strict TLV only list.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
tomereli: Before, m_buff and m_buff_length were always set in create() and parse(). This is just silly, it makes much more sense to save the buffer and length in the constructor. I will update the relevant commit message.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
tomereli: Perhaps not, will delete it if not.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: The approach I took here was to change as little as possible the current code, I didn't refractor the VS APIs although there surely are much better ways to do what they do. I think we can consider refactoring these as well, but in a separate PR.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: I don't like it either but it's the most generic method I could think of to allow that. I reviewed the network map and it is a much bigger change to refractor that into sending smaller chunks.
If you think of a nicer way I'd love to hear it.

framework/tlvf/src/include/tlvf/BaseClass.h
tomereli: Sound OK, but the commit API doesn't need an argument, it knows the buffer length and we know the current buffer pointer so we can calculate that ourselves

framework/tlvf/src/src/CmduMessage.cpp
arnout: > beerocks_header derives from TlvList and its classes also don't have type and length. 

Good point, I didn't realize that. I was too fixated on WSC :-)

> Maybe a better name would be TlvfClassList since it's a container for TLVF classes more than it is a TLV list.

That's a different idea than what I originally had, but makes a lot more sense. Indeed, if you look at it as a sequence of tlvf-generated structures, it makes a lot more sense.

So yes, a rename would be a good idea. ClassList is maybe more appropriate, like we have BaseClass.

tomereli:
> 
> To make some progress, I would like a separate PR with the following commits:
> 
> * [41d99bc](https://github.com/prplfoundation/prplMesh/commit/41d99bcc3d6f26e47392b6a6f12dd437602bc3b3) tlvf: Fix bug in BuffPtrIncrementSafe()
> * [4531edd](https://github.com/prplfoundation/prplMesh/commit/4531edde8dd65be85c8fa3fe82d27c779e783490) remove tlvf & btlvf src from .clang-ignore
> * [e80a2ce](https://github.com/prplfoundation/prplMesh/commit/e80a2ce1a48e67ed8003d4e3d458707f1fb5fe7f) tlvf_conf.yaml: Remove include_source_path (note that the commit sha referenced in the commit message is not correct)
> * [cda27db](https://github.com/prplfoundation/prplMesh/commit/cda27dbe263978a7599a038ad703edcfd47040a6) common: beerocks_message: remove unclear TODO

This

> * [b7792fa](https://github.com/prplfoundation/prplMesh/commit/b7792fa6dc2277202406debd6f0486bb99a83934) tlvf: Delete dynamicCast API from TlvList

can't be part of this PR since TlvList is not yet added.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: Or just return `std::shared_ptr<beerocks_header>`

PR #576, opened 2019-12-17
agent: slave: stop loading BPL dynamically with dlopen

mariomaz:
Loading with dlopen only makes sense in a few cases:
- for plugins that you don't always want to load;
- to be able to handle different library versions with different
interfaces;
- to be able to specify the file to load from the program (e.g. based on
a command-line argument) instead of letting the dynamic loader find it.

We are in none of these situations. The only thing that is desired is
that it is possible to replace bpl with a different implementation (but
with the same ABI) without rebuilding the rest of prplMesh. But that is
also possible with a normal dynamic library.

Therefore, remove the dlopen() and just rely on the dynamic library
loader to find bpl.

Since bpl is now a link_library dependency, the include path no longer
needs to be specified explicitly either.


mariomaz:
Required for #469.

framework/platform/bpl/include/bpl/bpl.h
```diff
@@ -9,9 +9,7 @@
 #ifndef _BPL_H_
 #define _BPL_H_
 
-#ifdef __cplusplus
-extern "C" {
-#endif
+namespace bpl {
```
arnout, line 12: Since the namespace is such a big change, I would have done the removal of the `extern "C"` in a separate commit. But that's just a nitpick.

More importantly, though: I would take this occasion to also remove the `bpl_` prefix from all names. The namespace replaces it.

Also, did you make this change automatically (e.g. with a sed script)? If yes, please mention the command you used in the commit message, and explicitly mention which additional manual changes you made.

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -28,6 +28,8 @@ using namespace beerocks::bpl;
 
 const char *s_error_strings[] = {FOREACH_ERROR_CODE(GENERATE_ERROR_STRING)};
 
+namespace bpl {
```
arnout, line 29: Shouldn't the s_error_strings above also be in the bpl namespace? Hm, actually, it doesn't even seem to be used...

arnout (commented):
Looks good so far.

Is there anything else that needs to be done? To me it looks like it's ready for merge.

framework/platform/bpl/include/bpl/bpl.h
mariomaz: I was going to make this very same proposal 
> More importantly, though: I would take this occasion to also remove the bpl_ prefix from all names. The namespace replaces it.

I didn't use any script to make the changes, all changes where made manually.

framework/platform/bpl/linux/bpl_cfg.cpp
mariomaz: I'll remove this variable.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -239,7 +239,7 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
             m_sConfig.slave_iface_socket.erase(iface);
 
             LOG(INFO) << "sending platform_notify: slave socket disconnected " << iface;
-            platform_notify_error(BPL_ERR_BH_SLAVE_SOCKET_DISCONNECTED,
+            platform_notify_error(bpl::BPL_ERR_BH_SLAVE_SOCKET_DISCONNECTED,
```
tomereli: So the error should become `enum class`
```
platform_notify_error(bpl_error::BH_SLAVE_SOCKET_DISCONNECTED;
```
(nitpick)

tomereli (approved):
LGTM, please rebase and merge

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done.
It's not a nitpicking comment actually: using enum classes over enum types has some advantages that allow for better quality code.

tomereli (approved):

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -26,8 +26,6 @@ using namespace beerocks::bpl;
 #define RETURN_OK 0
 #define RETURN_ERR -1
 
-const char *s_error_strings[] = {FOREACH_ERROR_CODE(GENERATE_ERROR_STRING)};
```
arnout, line 29: With this, the GENERATE_ERROR_STRING macro is also no longer used.

arnout (approved):
Nice work.

PR #577, opened 2019-12-17
common: ap_wlan_hal_create() - fix typo

adam1985d:
the ap_wlan_hal_dwpal build fails (UGW) due to typo:
'adwpal' instead of 'dwpal'.

after the typo fix build succeeds.


arnout (approved):
You had unrelated changes in this commit, which actually broke clang-format. So I fixed that up and force-pushed.

I also added the full component information to the commit message: "common: bwl: dwpal:" instead of just "common:". Nitpick but I was anyway fixing up the commit so I took it along.

PR #579, opened 2019-12-17
Feature/agent ucc listener onboarding

### LINKED ISSUE #336: [TASK] Support wfa_ca in agent

morantr:
Any test case triggered by the UCC application starts with dev_reset_defaults which "tells" the agent that it should go back to its default state, meaning unconfigured and not onboarded.
Next, onboarding is triggered as part of the dev_set_config command.

This PR completes the agent wfa_ca listener (agent ucc listener) by adding agent implementation to dev_reset_default and dev_set_config.

The complete prplMesh agent flow starts with the UCC application triggering
dev_reset_default, which results in a full slave reset triggered by the backhaul manager as a response to the ucc agent's request.
The ucc agent now waits in BACKHAUL_SEND_ENABLE state until the dev_set_config command from the UCC application is received, thus blocking the agent from starting onboarding + autoconfig flows.

Next, the UCC sends "dev_set_config,name,IntelCTTDUT,program,map,backhaul,eth" command (can be either wired or wireless), for which the ucc agent listener replies with RUNNING, and sets the onboarding and onboarding to "in progress" state, continue with the onboarding with the requested radio/eth interface.

The son_slave now unblocks and proceeds with the state machine, which results in receiving M2 which unblocks the ucc listener the return "COMPLETE" status to ucc.

Fixes #336


arnout:
[Fix dev_reset_default for agent](https://trello.com/c/Vcn26ZWp/12-fix-devresetdefault-for-agent)

tomereli (approved):

Issue #556, opened 2019-12-10
[TASK] Initial BU Intel IP repeater - port existing Repeater functionality to prplMesh


adam1985d:
As part of porting RDKB code to prplMesh the Standalone Repeater functionality was broken due to:
1.  send_slaves_enable() implementation was changed to send bandwidth and center_channel, which are read from "radio_info" struct after a "refresh_radio_info()" was performed - but the refresh_radio_info() doesn't yet support STA radio info.
- [x] revert implementation change which was done as part of port of RDKB code to prplMesh
- [x] create tasks: 1) add STA support in refresh_radio_info(). 2) Extend send_slaves_enable() to provide bandwidth and center_frequency
2. Waver in RDKB could not have been ported to prplMesh.
In RDKB MESH is not yet supported and the check "reject join if GW haven't joined yet" done as part of slave_join in controller (and relevant for MESH only) was commented-out. The commenting-out was not ported but no other solution was provided. Need to encapsulate this check with not-standalone-repeater condition.
- [x] add not-standalone-repeater condition to "reject join if GW haven't joined yet" done as part of slave_join in controller

#581

tomereli:
@adam1985d please also create a card in our Trello board for this - https://trello.com/b/LYqZPvqT/prplmesh-agile-board

adam1985d:
> @adam1985d please also create a card in our Trello board for this - https://trello.com/b/LYqZPvqT/prplmesh-agile-board

done - please review it in Trello board

adam1985d:
mistakenly closed

PR #581, opened 2019-12-18
Feature/port repeater from rdkb to prpl with ugw

adam1985d:
This PR is for porting exiting RDKB repeater functionality to the PRPL
 #556

Testing of the changes in this PR was done over Linux as well as UGW.

As part of RDKB implementation (not supporting MESH) a waiver was introduced:
- check if Agent tries to join before GW joined was commented out
- the above-mentioned condition is required for MESH and waiver could not be ported as-is
- the waiver was removed in the slave-join handling and a fix to the flow was introduced

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -998,14 +998,15 @@ bool backhaul_manager::backhaul_fsm_wireless(bool &skip_select)
             } else {
                 UTILS_SLEEP_MSEC(1000);
             }
+            break;
         }
+
+        state_attempts = 0; // for next state
 
         if (connected) {
             FSM_MOVE_STATE(MASTER_DISCOVERY);
-            state_attempts = 0; // for next state
         } else {
             FSM_MOVE_STATE(INITIATE_SCAN);
-            state_attempts = 0; // for next state
         }
 
         break;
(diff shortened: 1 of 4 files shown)


morantr (approved):

adam1985d:
[Initial BU Intel IP repeater to prplMesh](https://trello.com/c/eYDjEaBZ/32-initial-bu-intel-ip-repeater-to-prplmesh)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1805,24 +1805,17 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
 
     for (auto soc : slaves_sockets) {
         auto notification =
             message_com::create_vs_message<beerocks_message::cACTION_BACKHAUL_ENABLE_APS_REQUEST>(
                 cmdu_tx);
 
         if (soc->sta_iface == m_sConfig.wireless_iface) {
-            notification->channel()   = iface_hal->get_channel();
-            notification->bandwidth() = radio_info.bandwidth;
-            notification->center_channel() =
```
arnout, line 1814: I think I've asked this question before, but I don't remember the answer:

This will leave the bandwidth and center_channel uninitialized, so they will use whatever was in the buffer before. Why is that acceptable? (Note that we already are in that situation if son->sta_iface != wireless_iface...)

arnout (approved):
Truly *excellent* commit messages. Thank you for that!

I have one remark but it's more a question.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
adam1985d: You are right, you did ask.

Currently, the refresh_radio_info doesn't support STAs, and thus the BW and center_channel for STAs.
I can set it to zeros (although not a valid value) until we add this functionality to the refresh_radio_info

PR #583, opened 2019-12-18
Feature/ci turris omnia target

rmelotte:
We would like to build and test prplmesh on all the targets we currently support.
This first PR does that for the Turris Omnia only.

- [x] Build for the turris-omnia using the pre-built OpenWrt.
- [x] Deploy to the target
- [x] Run the status script to check that prplmesh is operational

Pipeline result: https://gitlab.com/prpl-foundation/prplMesh/pipelines/103987368

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,14 +36,45 @@ run-tests:
       - logs
     when: always
 
-  artifacts:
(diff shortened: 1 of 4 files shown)


rmelotte:
This PR has been plit between #586 #599, so closing.

PR #584, opened 2019-12-18
tlvf: Add ClassList implementation

### LINKED ISSUE #597: TLVF - ClassList design (Refactoring)
### Overview

Currently, CmduMessage (and its derived classes CmduMessageTx & CmduMessageRx) include the TLV class vector and the buffer.

Some TLVs, such as vendor-specific and WSC, may contain arbitrary data that can be parsed/created with the same mechanics that we use in the CmduMessage and are required by multiple features.

Currently, the way we handle the VS TLV, for example, is to preallocate its buffer to the maximum allowed, which leads to extraneous data at the end.

### Motivation
The motivation for this design is to have an infrastructure for creating and parsing all types of TLV, including VS TLV and WSC.

To allow support for such TLVs, we can create a new class called **ClassList** which will include all the buffer handling that is currently part of the CmduMessage classes. 
In simple words, ClassList is a list of any TLVF generated classes, which also include a list of TLVs.
Including an inner Class List can be done by adding an unknown length buffer to the end of the YAML file (WSC / VS), which will be used as the buffer for the inner ClassList. 

That way, we can easily support Vendor-Specific TLV and WSC.

[ For example, in order to support VS TLV we will add an unknown length buffer at the end of this YAML file, which will be used as the buffer for the inner Class list. 
This inner Class list, which derives from ClassList contain the beerocks
operation header and classes. ]

Since the inner ClassList may contain variable-length data, we pre-allocate the payload to the maximum allowed size.

As a result, if this inner ClassList is not using all the buffer, there is a need to trim this tailroom, since 
some TLVs such as the WSC TLV does not allow to have **extraneous data** at the end.

### Design Details : 

To add support for this ClassList design, the following idea is suggested :

Each time we call the addClass method, the addClass method itself calls BaseClass::Finalize() for the previous class in the list.
The method first validates the class is fully initialized  (it means that if this class contains a member of a type class, it is initialized) and do **swap** always. 

In addition, we check if there is an inner ClassList in this class, and if it does, call ClassList::Finalize(), which returns the size of the unused buffer of this classList (the tailroom).
Now, when we know the size of the tailroom we know the correct length of this TLV, so update the length and the m_buff_ptr to point to the correct address, by subtracting the tailroom value.

How can we know if a TLV contains an inner ClassList?
the BaseClass should have a shared pointer of type ClassList which initially initialized to null.
when we create a new inner ClassList in a TLV, we set the shared pointer to point on this ClassList.

Finally, when CmduMessageTx::Finalize() is called, EOM TLV is added and we are done (there is no need to do finalize to the last class of the list since when we add EOM class we already do it by calling the addClass method)

It is important to note that ClassList::Finalize() method should do finalize to the last class of the list (since each time we call addClass we do finalize to the previous class, we never do to the last one)

This way we can support a case of TLV which contains an inner ClassList, which is very useful for Vendor-Specific TLV and for WSC message, and maybe for more needs in the future.

This can be split into the following subtasks :
- [x] remove the `swap_needed` completely, and always call swap() on finalizing and parsing
- [x] add ClassList class which include all the buffer handling
- [x] add `ClassList ::Finalize` method which always returns the tailroom size  and finalize the last class of the list
- [x] add Finalize, pure virtual method, to BaseClass which validates that the class is fully initialized, do swap, and finalize an inner class list if exists and update m_buff_ptr accordingly. if the class has a length member, update the length also.
- [x] add a shared pointer of type BaseClass which initially initialized to null.
- [x] add a new method called `addClassList` which set the shared pointer to point the new ClassList
- [x] addClass method should finalize the previous class added

image

tomereli:
Currently, the CmduMessage class conflates the CMDU-specific things with
the general concept of a list of TLVs.

Create a new class ClassList that generalizes the handling of lists of
any TLVF generated classes (this includes lists of TLVs).

In a later commit/PR, this will be used in CmduMessage to represent the
list of TLVF classes that compose a CMDU - the CmduHeader and the list
of TLVs themselves.

In addition, this new container class will be used to represent inner
TLVs data such as vendor specific TLVs which contain a beerocks
operation header and class, and WSC TLVs which contains a list of
attributes.

All those will extend the ClassList to provide the parsing and creating
capabilities for the inner data.

Fixes #597

### DIFF
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -209,9 +209,7 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
             if (tlv_vendor_specific.vendor_oui() ==
                 ieee1905_1::tlvVendorSpecific::eVendorOUI::OUI_INTEL) {
                 // assuming that the magic is the first data on the beerocks header
-                auto beerocks_magic = *(
-                    uint32_t *)((uint8_t *)tlv + ieee1905_1::tlvVendorSpecific::get_initial_size());
-
+                auto beerocks_magic = *(uint32_t *)(tlv_vendor_specific.payload());
                 swap_32(beerocks_magic);
                 if (beerocks_magic != message::MESSAGE_MAGIC) {
                     THREAD_LOG(WARNING) << "mismatch magic " << std::hex << int(beerocks_magic)
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -198,8 +198,8 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
 
         swap_16(length);
         if (static_cast<ieee1905_1::eTlvType>(type) == ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC) {
-            auto tlv_vendor_specific = ieee1905_1::tlvVendorSpecific(
-                (uint8_t *)tlv, length + sizeof(sTlvHeader), true);
+            auto tlv_vendor_specific =
+                ieee1905_1::tlvVendorSpecific((uint8_t *)tlv, length + sizeof(sTlvHeader), true);
             if (!tlv_vendor_specific.isInitialized()) {
                 LOG(ERROR) << "tlvVendorSpecific init() failure";
                 ret = false;
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -17,11 +17,6 @@
 
 using namespace beerocks;
 
-typedef struct sTlvHeader {
-    uint8_t type;
-    uint16_t length;
-} __attribute__((packed)) sTlvHeader;
-
 #define DEFAULT_MAX_SOCKET_CONNECTIONS 10
 #define TX_BUFFER_UDS (tx_buffer + sizeof(beerocks::message::sUdsHeader))
 #define TX_BUFFER_UDS_SIZE (sizeof(tx_buffer) - sizeof(beerocks::message::sUdsHeader))
@@ -184,10 +179,10 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
 // FIXME - WLANRTSYS-6360 - should be moved to transport
 bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
 {
-    sTlvHeader *tlv =
-        reinterpret_cast<sTlvHeader *>((uint8_t *)uds_header + sizeof(message::sUdsHeader) +
-                                       ieee1905_1::cCmduHeader::get_initial_size());
-    sTlvHeader *first_tlv = tlv;
+    ieee1905_1::sTlvHeader *tlv = reinterpret_cast<ieee1905_1::sTlvHeader *>(
+        (uint8_t *)uds_header + sizeof(message::sUdsHeader) +
+        ieee1905_1::cCmduHeader::get_initial_size());
+    ieee1905_1::sTlvHeader *first_tlv = tlv;
 
     bool ret = true;
 
@@ -198,8 +193,8 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
 
         swap_16(length);
         if (static_cast<ieee1905_1::eTlvType>(type) == ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC) {
-            auto tlv_vendor_specific =
-                ieee1905_1::tlvVendorSpecific((uint8_t *)tlv, length + sizeof(sTlvHeader), true);
+            auto tlv_vendor_specific = ieee1905_1::tlvVendorSpecific(
+                (uint8_t *)tlv, length + sizeof(ieee1905_1::sTlvHeader), true);
             if (!tlv_vendor_specific.isInitialized()) {
                 LOG(ERROR) << "tlvVendorSpecific init() failure";
                 ret = false;
@@ -231,8 +226,8 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
         }
 
         // set the next tlv
-        tlv    = (sTlvHeader *)((uint8_t *)tlv + sizeof(sTlvHeader) + length);
-        type   = tlv->type;
+        tlv  = (ieee1905_1::sTlvHeader *)((uint8_t *)tlv + sizeof(ieee1905_1::sTlvHeader) + length);
+        type = tlv->type;
         length = tlv->length;
 
     } while ((uint8_t *)tlv <= (uint8_t *)first_tlv + uds_header->length);
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -168,10 +168,9 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
         return false;
     }
 
-    if (!cmdu_rx.parse(uds_header->length, uds_header->swap_needed)) {
+    if (!cmdu_rx.parse()) {
         THREAD_LOG(ERROR) << "parsing cmdu failure, rx_buffer" << std::hex << rx_buffer << std::dec
-                          << ", uds_header->length=" << int(uds_header->length)
-                          << ", uds_header->swap_needed=" << int(uds_header->swap_needed);
+                          << ", uds_header->length=" << int(uds_header->length);
         return false;
     }
 
@@ -197,13 +196,10 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
 
     do {
 
-        if (uds_header->swap_needed) {
-            swap_16(length);
-        }
-
+        swap_16(length);
         if (static_cast<ieee1905_1::eTlvType>(type) == ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC) {
             auto tlv_vendor_specific = ieee1905_1::tlvVendorSpecific(
-                (uint8_t *)tlv, length + sizeof(sTlvHeader), true, uds_header->swap_needed);
+                (uint8_t *)tlv, length + sizeof(sTlvHeader), true, true);
             if (!tlv_vendor_specific.isInitialized()) {
                 LOG(ERROR) << "tlvVendorSpecific init() failure";
                 ret = false;
@@ -215,10 +211,8 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
                 // assuming that the magic is the first data on the beerocks header
                 auto beerocks_magic = *(
                     uint32_t *)((uint8_t *)tlv + ieee1905_1::tlvVendorSpecific::get_initial_size());
-                if (uds_header->swap_needed) {
-                    swap_32(beerocks_magic);
-                }
 
+                swap_32(beerocks_magic);
                 if (beerocks_magic != message::MESSAGE_MAGIC) {
                     THREAD_LOG(WARNING) << "mismatch magic " << std::hex << int(beerocks_magic)
                                         << " != " << int(message::MESSAGE_MAGIC) << std::dec;
@@ -230,10 +224,8 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
                 THREAD_LOG(INFO) << "Not an Intel vendor specific message!";
             }
 
-            if (uds_header->swap_needed) {
-                // cancel the swap we did
-                tlv_vendor_specific.class_swap();
-            }
+            // cancel the swap we did
+            tlv_vendor_specific.class_swap();
         } else if (static_cast<ieee1905_1::eTlvType>(type) ==
                        ieee1905_1::eTlvType::TLV_END_OF_MESSAGE &&
                    length == 0) {
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -28,8 +28,9 @@ typedef struct sTlvHeader {
 
 socket_thread::socket_thread(const std::string &unix_socket_path_)
     : thread_base(), cmdu_tx(TX_BUFFER_UDS, TX_BUFFER_UDS_SIZE),
-      unix_socket_path(unix_socket_path_), server_socket(nullptr),
-      server_max_connections(DEFAULT_MAX_SOCKET_CONNECTIONS)
+      unix_socket_path(unix_socket_path_), cmdu_rx(rx_buffer + sizeof(message::sUdsHeader),
+                                                   sizeof(rx_buffer) - sizeof(message::sUdsHeader)),
+      server_socket(nullptr), server_max_connections(DEFAULT_MAX_SOCKET_CONNECTIONS)
 {
     memset(TX_BUFFER_UDS, 0, TX_BUFFER_UDS_SIZE);
     set_select_timeout(500);
@@ -167,8 +168,7 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
         return false;
     }
 
-    if (!cmdu_rx.parse(rx_buffer + sizeof(message::sUdsHeader), uds_header->length,
-                       uds_header->swap_needed)) {
+    if (!cmdu_rx.parse(uds_header->length, uds_header->swap_needed)) {
         THREAD_LOG(ERROR) << "parsing cmdu failure, rx_buffer" << std::hex << rx_buffer << std::dec
                           << ", uds_header->length=" << int(uds_header->length)
                           << ", uds_header->swap_needed=" << int(uds_header->swap_needed);
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -199,7 +199,7 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
         swap_16(length);
         if (static_cast<ieee1905_1::eTlvType>(type) == ieee1905_1::eTlvType::TLV_VENDOR_SPECIFIC) {
             auto tlv_vendor_specific = ieee1905_1::tlvVendorSpecific(
-                (uint8_t *)tlv, length + sizeof(sTlvHeader), true, true);
+                (uint8_t *)tlv, length + sizeof(sTlvHeader), true);
             if (!tlv_vendor_specific.isInitialized()) {
                 LOG(ERROR) << "tlvVendorSpecific init() failure";
                 ret = false;
--- a/common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
+++ b/common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
@@ -1,0 +1,50 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2019 Tomer Eliyahu (Intel Corporation)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _TLVF_BEEROCKS_HEADER_H_
+#define _TLVF_BEEROCKS_HEADER_H_
+
+#include <beerocks/tlvf/beerocks_message_header.h>
+#include <tlvf/ClassList.h>
+
+namespace beerocks {
+
+/**
+ * @brief beerocks_header
+ *
+ * Each vendor specific message used for internal and external communication
+ * between the different components in prplMesh, via UDS or the BTL, includes
+ * an action header and the actual message data.
+ * These stored in the vendor specific TLV payload as a separate TLVlist.
+ *
+ * Note that this is not really a "header", a more appropriate name would be
+ * beerocks_actions - but since the name beerocks_header is widely used, we keep
+ * it for backward comparability. Same goes to all the getter functions.
+ *
+ */
+class beerocks_header : public ClassList {
+public:
+    beerocks_header() = delete;
+    beerocks_header(uint8_t *buff, size_t buff_len, bool parse) : ClassList(buff, buff_len, parse)
+    {
+    }
+    std::shared_ptr<beerocks_message::cACTION_HEADER> actionhdr()
+    {
+        return getClass<beerocks_message::cACTION_HEADER>();
+    }
+    beerocks_message::eAction action()
+    {
+        return actionhdr() ? actionhdr()->action() : beerocks_message::ACTION_NONE;
+    }
+    uint8_t action_op() { return actionhdr() ? actionhdr()->action_op() : 0; };
+    uint16_t id() { return actionhdr() ? actionhdr()->id() : 0; }
+};
+
+} // namespace beerocks
+
+#endif // _TLVF_BEEROCKS_HEADER_H_
--- a/common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
+++ b/common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
@@ -9,7 +9,7 @@
 #ifndef _TLVF_BEEROCKS_MESSAGE_H_
 #define _TLVF_BEEROCKS_MESSAGE_H_
 
-#include <beerocks/tlvf/beerocks_message_header.h>
+#include <beerocks/tlvf/beerocks_header.h>
 #include <tlvf/CmduMessageRx.h>
 #include <tlvf/CmduMessageTx.h>
 #include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
@@ -61,30 +61,42 @@ public:
         return std::dynamic_pointer_cast<beerocks_message::cACTION_HEADER>(cmdu.getClass(1));
     }
 
-    template <class T>
-    static std::shared_ptr<T>
-    add_intel_vs_data(ieee1905_1::CmduMessageTx &cmdu_tx,
-                      std::shared_ptr<ieee1905_1::tlvVendorSpecific> vs_tlv, uint16_t id = 0)
+    static std::shared_ptr<beerocks_header> get_beerocks_header(ieee1905_1::CmduMessage &cmdu)
     {
-        beerocks_message::eAction action;
-        std::shared_ptr<T> p_class = nullptr;
+        auto tlv = cmdu.msg.getClass<ieee1905_1::tlvVendorSpecific>();
+        if (!tlv)
+            return nullptr;
+        if (!intel_oui(tlv))
+            return nullptr;
+        return std::dynamic_pointer_cast<beerocks_header>(tlv->getInnerClassList());
+    }
 
-        auto actionhdr = cmdu_tx.addClass<beerocks_message::cACTION_HEADER>();
-        if (!actionhdr) {
+    template <class T>
+    static std::shared_ptr<beerocks_header>
+    add_intel_vs_data(std::shared_ptr<ieee1905_1::tlvVendorSpecific> tlv,
+                                  uint16_t id = 0)
+    {
+        // Allocate maximum allowed length for the payload, so it can accommodate variable length
+        // data inside the internal TLV list.
+        // On finalize(), the buffer is shrunk back to its real size.
+        size_t payload_length =
+            tlv->getBuffRemainingBytes() - ieee1905_1::tlvEndOfMessage::get_initial_size();
+        if (!tlv->alloc_payload(payload_length)) {
             std::cout << "beerocks_message.h[ " << __LINE__ << "]: " << __FUNCTION__ << " failed!"
                       << std::endl;
             return nullptr;
         }
-        vs_tlv->length() += beerocks_message::cACTION_HEADER::get_initial_size();
-
-        p_class = cmdu_tx.addClass<T>();
-        if (!p_class) {
-            std::cout << "beerocks_message.h[ " << __LINE__ << "]: " << __FUNCTION__ << " failed!"
-                      << std::endl;
+        std::shared_ptr<beerocks_header> hdr =
+            std::make_shared<beerocks_header>(tlv->payload(), tlv->payload_length(), false);
+        if (!hdr)
             return nullptr;
-        }
+        if (!hdr->addClass<beerocks_message::cACTION_HEADER>())
+            return nullptr;
+        if (!hdr->addClass<T>())
+            return nullptr;
 
-        auto action_op = p_class->get_action_op();
+        beerocks_message::eAction action = beerocks_message::eAction::ACTION_NONE;
+        auto action_op                   = hdr->getClass<T>()->get_action_op();
         if (typeid(action_op) == typeid(beerocks_message::eActionOp_1905_VS))
             action = beerocks_message::eAction::ACTION_1905_VS;
         else if (typeid(action_op) == typeid(beerocks_message::eActionOp_CONTROL))
@@ -101,8 +113,6 @@ public:
             action = beerocks_message::eAction::ACTION_CLI;
         else if (typeid(action_op) == typeid(beerocks_message::eActionOp_BML))
             action = beerocks_message::eAction::ACTION_BML;
-        else
-            action = beerocks_message::eAction::ACTION_NONE;
 
         if (action == beerocks_message::eAction::ACTION_NONE) {
             std::cout << "beerocks_message.h[ " << __LINE__ << "]: " << __FUNCTION__ << " failed!"
@@ -110,11 +120,12 @@ public:
             return nullptr;
         }
 
-        actionhdr->action()    = (beerocks_message::eAction)action;
-        actionhdr->action_op() = action_op;
-        actionhdr->id()        = id;
+        hdr->actionhdr()->action()    = (beerocks_message::eAction)action;
+        hdr->actionhdr()->action_op() = action_op;
+        hdr->actionhdr()->id()        = id;
+        tlv->addInnerClassList(hdr);
 
-        return p_class;
+        return hdr;
     }
 
     template <class T>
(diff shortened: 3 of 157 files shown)


common/beerocks/tlvf/src/src/beerocks_message.cpp
```diff
@@ -164,7 +162,7 @@ bool message_com::send_cmdu(Socket *sd, ieee1905_1::CmduMessageTx &cmdu_tx,
     }
 
     uds_header->length      = cmdu_tx.getMessageLength();
-    uds_header->swap_needed = swap;
+    uds_header->swap_needed = true;
```
morantr: If it is always swapped then it seems to me that you can also remove the `swap_needed' from `uds_header`.

morantr (changes requested):

common/beerocks/tlvf/src/src/beerocks_message.cpp
```diff
@@ -137,9 +137,7 @@ bool message_com::send_cmdu(Socket *sd, ieee1905_1::CmduMessageTx &cmdu_tx,
         return false;
     }
 
-    bool swap = !dst_mac.empty();
-
-    if (!cmdu_tx.finalize(swap)) {
+    if (!cmdu_tx.finalize(true)) {
```
morantr: If the swap is always true, then maybe it is a good idea to remove the swap argument from the function.

morantr (changes requested):

tomereli:
@morantr please remove you changes requested, this is not ready yet.

common/beerocks/tlvf/src/src/beerocks_message.cpp
tomereli: it happens on the followup commit

common/beerocks/tlvf/src/src/beerocks_message.cpp
tomereli: happens in the followup commit

framework/tlvf/src/include/tlvf/ClassList.h
```diff
@@ -0,0 +1,139 @@
...
+    ClassList(uint8_t *buff, size_t buff_len, bool parse = false, bool swap = false);
+    ~ClassList();
+
+public:
+    const std::vector<std::shared_ptr<BaseClass>> &getClassVector() const
+    {
+        return (const std::vector<std::shared_ptr<BaseClass>> &)m_class_vector;
+    }
+
+    template <class T> std::shared_ptr<T> addClass()
```
arnout, line 41: Since this one is here to stay, it might be good to add a doxygen comment now.

framework/tlvf/src/include/tlvf/ClassList.h
```diff
@@ -0,0 +1,139 @@
...
+    {
+        std::shared_ptr<T> ptr;
+        if (m_buff) {
+            if (m_class_vector.size() == 0) {
+                ptr = std::make_shared<T>(m_buff, m_buff_len, m_parse, m_swap);
+            } else {
+                // do not allow to use addClass method if the previous tlv is not fully initialized
+                if (!m_class_vector.back()->isPostInitSucceeded()) {
```
arnout: Since we have this anyway, might be nice to change the code into:

```
auto prev = m_class_vector.back();
if (prev) {
   if (!prev->isPostInitSucceeded()) {
      ...
   }
   ptr = std::make_shared<T>(prev, m_parse, m_swap);
} else {
   ptr = std::make_shared<T>(m_buff, m_buff_len, m_parse, m_swap);
}
```

framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -20,133 +21,41 @@ namespace ieee1905_1 {
...
-    bool swap_needed() { return m_swap; }
-    bool is_finalized() const { return m_finalized; };
-    bool is_swapped() const { return m_swapped; };
-    eMessageType getMessageType();
-    uint16_t getMessageId();
 
-protected:
-    void reset();
+    // Forward wrapper functions
+    // TODO check which of them can be removed
```
arnout, line 45: Good!

framework/tlvf/src/src/CmduMessage.cpp
```diff
@@ -10,149 +10,46 @@
...
-    }
-
-    if (m_class_vector.size() == 0) {
-        tlvValue = *m_cmdu_header->getBuffPtr();
-    } else {
-        tlvValue = *m_class_vector.back()->getBuffPtr();
-    }
-
-    return int(tlvValue);
+    uint8_t *tlv = msg.getClassVector().back()->getBuffPtr();
```
arnout: I would like an explanatory comment here, saying that the first byte of the next TLV is the type.

More importantly, though: if we've consumed the entire buffer (so in practice, after the EOM TLV), getBuffPtr() may return a pointer beyond the buffer. So I think we should still check getBuffRemainingBytes explicitly. I would actually keep the original condition:

```
if (prev->getBuffRemainingBytes() < kTlvHeaderLength) {
```

Also it would be nicer to (in a preparatory commit) add a private header function that returns a sTlvHeader and that can be used by all of the below functions.

framework/tlvf/src/src/CmduMessage.cpp
```diff
@@ -10,149 +10,46 @@
...
-    } else {
-        return nullptr;
-    }
-}
-
 int CmduMessage::getNextTlvType() const
 {
-    uint8_t tlvValue = 0;
-    if (!m_cmdu_header && (m_cmdu_header->getBuffRemainingBytes() < kTlvHeaderLength)) {
+    if (!getCmduHeader())
```
arnout, line 15: I think it will be a lot easer to understand what's going on if we make use of the ClassList:

```
auto prev = msg.getClassVector().back();
if (!prev)
    return -1;
```

(prev in a variable because I'll use it again below.)

Also, I think the msg.getClassVector().back() is going to be used so often that it pays of to add a method for it in ClassList.

framework/tlvf/src/src/CmduMessage.cpp
```diff
@@ -10,149 +10,46 @@
...
-    } else {
-        tlvValue = *m_class_vector.back()->getBuffPtr();
-    }
-
-    return int(tlvValue);
+    uint8_t *tlv = msg.getClassVector().back()->getBuffPtr();
+    return *tlv;
 }
 
 bool CmduMessage::getNextTlvType(eTlvType &tlvType) const
```
arnout, line 21: AFAICS this function only gets used in tlvf_test, so maybe we should just remove it in a preparatory commit...

framework/tlvf/src/include/tlvf/CmduMessageRx.h
```diff
@@ -22,14 +22,14 @@ class CmduParser;
 class CmduMessageRx : public CmduMessage {
 
 public:
-    CmduMessageRx();
-    CmduMessageRx(CmduMessageRx &original);
+    CmduMessageRx() = delete;
+    CmduMessageRx(uint8_t *buff, size_t buff_len);
     ~CmduMessageRx();
-    bool parse(uint8_t *buff, size_t buff_len, bool swap_needed = true, bool parse_tlvs = false);
+    bool parse(bool swap_needed = true, bool parse_tlvs = false);
```
arnout: This change (moving the buffer from parse to constructor) could be a separate, preparatory commit. It is relatively easy to make that change and change the calls to parse() and the constructors in the same commit.

framework/tlvf/src/include/tlvf/CmduMessageRx.h
```diff
@@ -22,14 +22,14 @@ class CmduParser;
...
     CmduMessageRx &operator=(const CmduMessageRx &) = delete;
 
 private:
-    std::list<std::shared_ptr<CmduParser>> parsers_;
+    std::shared_ptr<CmduParser> parser;
```
arnout, line 32: This change is definitely easy to do in a separate, preparatory commit.

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -12,47 +12,26 @@
...
-    m_cmdu_header = std::make_shared<cCmduHeader>(buff, buff_len, true, false);
-    if (!m_cmdu_header || m_cmdu_header->isInitialized() == false) {
-        m_cmdu_header = nullptr;
+    msg.reset(true, swap_needed);
+    auto cmduhdr = msg.addClass<cCmduHeader>();
+    if (!cmduhdr)
         return false;
-    }
+    if (swap_needed)
+        cmduhdr->class_swap(); //swap back cmduheader
```
arnout: I think this is the only change that is directly related to the ClassList, so it can easily be done together with the previous commit.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -13,62 +13,39 @@
...
+CmduMessageTx::CmduMessageTx(uint8_t *buff, size_t buff_len) : CmduMessage(buff, buff_len) {}
 
 CmduMessageTx::~CmduMessageTx() {}
 
 std::shared_ptr<cCmduHeader> CmduMessageTx::create(uint16_t id, eMessageType message_type)
 {
-    if (!m_buff || !m_buff_len)
+    // parse - false since this is for TX
+    // swap - false since on TX we do swap in Finalize(swap_needed)
+    msg.reset(false, false);
```
arnout: This could also be done in a preparatory commit: add a protected reset function to CmduMessage which has the parse and swap arguments, and use that instead of setting parse and swap explicitly. And since we always do an addClass<cCmduHeader> afterwards (or a std::make_shared in the original code), we can move that into the reset function as well. And the memset should be moved into the function as well, of course.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -13,62 +13,39 @@
...
-        return nullptr;
-
-    if (!m_cmdu_header || m_cmdu_header->isInitialized() == false) {
+    if (!getCmduHeader())
         return nullptr;
-    }
 
     auto tlv = addClass<tlvVendorSpecific>();
-    if (!tlv) {
-        m_cmdu_header.reset();
```
arnout, line 68: This reset is removed now. Probably it shouldn't be done (we don't automatically reset in other cases), but better in a separate, explicit commit.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -77,47 +54,7 @@ std::shared_ptr<tlvVendorSpecific> CmduMessageTx::add_vs_tlv(tlvVendorSpecific::
...
-            // Classes order: +0 = tlvVendorSpecific, +1 = beerocks_header, +2 = beerocks_message
-            if (std::next(class_it, 1) == m_class_vector.end() ||
-                std::next(class_it, 2) == m_class_vector.end()) {
-                TLVF_LOG(ERROR) << "vendor specific TLV has not been added correctly";
-                return false;
-            }
-            *tlv_length += (std::next(class_it, 2)->get())->getLen();
-            class_it += 2;
-        }
-    }
```
arnout, line 113: I'm not convinced that it's a good idea to remove this. This hack should still just work with the ClassList.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -77,47 +54,7 @@ std::shared_ptr<tlvVendorSpecific> CmduMessageTx::add_vs_tlv(tlvVendorSpecific::
...
         return false;
-
-    if (swap_needed)
-        swap();
-
-    m_finalized = true;
-    return true;
-}
+    return msg.finalize(swap_needed);
+};
```
arnout: Missing newline.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
```diff
@@ -0,0 +1,50 @@
...
+ * These stored in the vendor specific TLV payload as a separate TLVlist.
+ *
+ * Note that this is not really a "header", a more appropriate name would be
+ * beerocks_actions - but since the name beerocks_header is widely used, we keep
+ * it for backward comparability. Same goes to all the getter functions.
+ *
+ */
+class beerocks_header : public ClassList {
+public:
+    beerocks_header() = delete;
```
arnout, line 32: Since the base class ClassList already deleted it, this is implied (nitpick).

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
```diff
@@ -0,0 +1,50 @@
...
+    beerocks_header(uint8_t *buff, size_t buff_len, bool parse) : ClassList(buff, buff_len, parse)
+    {
+    }
+    std::shared_ptr<beerocks_message::cACTION_HEADER> actionhdr()
+    {
+        return getClass<beerocks_message::cACTION_HEADER>();
```
arnout, line 38: Would it make sense to instead do a dynamic_cast on the front() of the class vector?

framework/tlvf/tlvf.py
```diff
@@ -1556,16 +1555,16 @@ def addClassSizeMethod(self, insert_name, insert_marker, name):
     def closeObject(self, obj_meta):
         if obj_meta.type == MetaData.TYPE_CLASS: # add class constractor
             # constractor 1
-            self.insertLineH(obj_meta.name, self.CODE_CLASS_CONSTRACTOR, "%s(uint8_t* buff, size_t buff_len, bool parse = false, bool swap_needed = false);" % (obj_meta.name))
-            self.insertLineCpp(obj_meta.name, self.CODE_CLASS_CONSTRACTOR, ("%s::%s(uint8_t* buff, size_t buff_len, bool parse, bool swap_needed) :" % (obj_meta.name, obj_meta.name)))
-            self.insertLineCpp(obj_meta.name, self.CODE_CLASS_CONSTRACTOR, ("%sBaseClass(buff, buff_len, parse, swap_needed) {" % self.getIndentation(1) ))
+            self.insertLineH(obj_meta.name, self.CODE_CLASS_CONSTRACTOR, "%s(uint8_t* buff, size_t buff_len, bool parse = false);" % (obj_meta.name))
```
arnout, line 1593: framework/tlvf/README.md still refers to the swap_needed argument.

There's also one in son_slave_thread.cpp comment and in beerocks_message.cpp.

framework/tlvf/src/src/BaseClass.cpp
```diff
@@ -41,3 +43,22 @@ bool BaseClass::buffPtrIncrementSafe(size_t length)
...
+    if (m_finalized) {
+        TLVF_LOG(DEBUG) << "finalize() called for already finalized class";
+        return true;
+    }
+    if (!isPostInitSucceeded()) {
+        TLVF_LOG(ERROR) << "post init check failed";
+        return false;
+    }
+    class_swap();
+    m_finalized = true;
```
arnout: Hm, this is risky...

Before, the m_swapped field of ClassList was checked in finalize and it was not done again if it was already swapped. Now, only the finalized flag is checked. Thus, the following sequence would have worked previously but no longer works: `cmdu.swap(); cmdu.finalize();` And there are probably more corner cases.

I think it would be good to do a review of all explicit cmdu.swap and foo.class_swap calls and explain why they're OK (for example, in a reply to this comment). And of course testing.

framework/tlvf/src/src/BaseClass.cpp
```diff
@@ -41,3 +43,22 @@ bool BaseClass::buffPtrIncrementSafe(size_t length)
...
+    return true;
+}
```
arnout: Missing newline.

arnout (commented):
Not finished yet but already posting my comments.

Mostly it's about better ways to split and order the commits or nitpick comments. The only important ones so far are:

- check that getNextTlvType and friends can't read beyond the buffer;
- check the explicit swap calls.

framework/tlvf/src/include/tlvf/ClassList.h
```diff
@@ -11,10 +11,10 @@
 #define _ClassList_H_
 
 #include <memory>
-#include <tlvf/BaseClass.h>
```
arnout: There's no need to remove this include.

framework/tlvf/src/include/tlvf/ClassList.h
```diff
@@ -11,10 +11,10 @@
...
-#include <tlvf/ieee_1905_1/eTlvType.h>
 #include <vector>
 
+class BaseClass;
```
arnout: There's no need for a forward declaration here.

framework/tlvf/src/include/tlvf/ClassList.h
```diff
@@ -36,7 +36,7 @@ class ClassList {
                 ptr = std::make_shared<T>(m_buff, m_buff_len, m_parse);
             } else {
                 // before adding a new class, finalize the previous one
-                if (!m_class_vector.back()->finalize())
+                if (!finalize(m_class_vector.back()))
```
arnout: There's no need for this weird finalize function.

framework/tlvf/src/include/tlvf/BaseClass.h
```diff
@@ -42,7 +42,7 @@ class BaseClass {
     uint8_t *m_buff_ptr__;
     const size_t m_buff_len__;
     const bool m_parse__;
-    bool m_finalized;
+    bool m_finalized__;
```
arnout, line 45: Why is this rename needed?

framework/tlvf/src/src/BaseClass.cpp
```diff
@@ -44,24 +44,3 @@ bool BaseClass::buffPtrIncrementSafe(size_t length)
...
-    if (!isPostInitSucceeded()) {
-        TLVF_LOG(ERROR) << "post init check failed";
-        return false;
-    }
-    if (m_inner)
-        m_inner->finalize();
-    class_swap();
-    m_finalized = true;
-    return true;
-}
```
arnout: I would prefer to keep this function in BaseClass, and add a private virtual finalize_impl function or something like that that is optionally generated by tlvf and that defaults to nothing. That keeps most of the code outside of tlvf which is nicer.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
```diff
@@ -61,30 +61,48 @@ class message_com {
...
+                                  uint16_t id = 0)
+    {
+        // Allocate maximum allowed length for the payload, so it can accommodate variable length
+        // data inside the internal TLV list.
+        // On finalize(), the buffer is shrunk back to its real size.
+        size_t payload_length;
+        const size_t max_vs_tlv_length = ieee1905_1::CmduMessage::kMaxCmduLength -
+                                         ieee1905_1::CmduMessage::kCmduHeaderLength -
+                                         ieee1905_1::tlvVendorSpecific::get_initial_size() -
+                                         ieee1905_1::tlvEndOfMessage::get_initial_size();
```
arnout: I still don't understand why we don't just allocate the rest of the buffer here. That's basically what we've done up to now and it seems to work...

framework/tlvf/src/src/BaseClass.cpp
tomereli: Well, it passed all tests so I guess it's fine. Nevertheless - to be reviewed again as part of #601, so keeping open.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_message.h
tomereli: Didn't cherry pick the commit yet, we will

framework/tlvf/src/include/tlvf/BaseClass.h
tomereli: To align with other variable names which are explicitly used in tlvf.py so we use the __ suffix to mark that.

framework/tlvf/src/include/tlvf/ClassList.h
```diff
@@ -0,0 +1,139 @@
...
+    const std::vector<std::shared_ptr<BaseClass>> &getClassVector() const
+    {
+        return (const std::vector<std::shared_ptr<BaseClass>> &)m_class_vector;
+    }
+
+    template <class T> std::shared_ptr<T> addClass()
+    {
+        std::shared_ptr<T> ptr;
+        if (m_buff) {
+            if (m_class_vector.size() == 0) {
```
morantr: m_class_vector.empty() (nitpick)

morantr: Please remove nesting as most as possible by handling an error right away and avoid `else`, it makes the code much more readable and clearer:
```cpp
template <class T> std::shared_ptr<T> addClass()
{
    std::shared_ptr<T> ptr;
    if (m_buff == nullptr) {
        return ptr;
    }
        
    if (m_class_vector.size() == 0) {
        ptr = std::make_shared<T>(m_buff, m_buff_len, m_parse, m_swap);
    } else {
        // do not allow to use addClass method if the previous tlv is not fully initialized
        if (!m_class_vector.back()->isPostInitSucceeded()) {
            TLVF_LOG(ERROR) << "TLV post init failed";
            return ptr;
        }
        ptr = std::make_shared<T>(m_class_vector.back(), m_parse, m_swap);
    }

    if (!ptr || ptr->isInitialized() == false) {
        return std::shared_ptr<T>();
    }
    
    m_class_vector.push_back(std::shared_ptr<BaseClass>(ptr));

    return std::shared_ptr<T>();
}
```

framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -20,133 +21,40 @@ namespace ieee1905_1 {
...
-                return std::shared_ptr<T>();
-            } else {
-                m_class_vector.push_back(std::shared_ptr<BaseClass>(ptr));
-            }
-        } else {
-            return std::shared_ptr<T>();
-        }
-        return ptr;
-    }
+    ClassList msg;
```
morantr, line 33: I don't understand why did you make `ClassList` be a member of `CmduMessage` instead of just be it.
I mean you took out all kinds of functions out from `CmduMessage` to `ClassList`, why couldn't you just change `CmduMessage`?

Also, why did you choose to use a member over heritage? It would have save you all the wrapper functions.

framework/tlvf/src/src/CmduMessage.cpp
```diff
@@ -10,149 +10,46 @@
...
-            swap_16((uint16_t &)tlvLength);
-        }
-    } else {
-        tlvLength = 0;
-    }
-    return tlvLength;
+    if (!getCmduHeader())
+        return -1;
+    uint8_t *tlv = msg.getClassVector().back()->getBuffPtr();
+    return *(uint16_t *)(tlv + sizeof(uint8_t));
```
morantr: For pointer casting, please you `reinterpret_cast`

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -13,62 +13,39 @@
...
-    m_parse = true;
-    reset();
-    m_cmdu_header = std::make_shared<cCmduHeader>(m_buff, m_buff_len, m_parse);
-    if (!m_cmdu_header || m_cmdu_header->isInitialized() == false) {
-        return nullptr;
-    }
-
-    return m_cmdu_header;
+    msg.reset(true, false);
+    return addClass<cCmduHeader>();
```
morantr, line 36: We talked about it. This function implementation is not correct. Add a comment so you could resolve it when you will finish with cmdu copy constructor.

framework/tlvf/src/include/tlvf/CmduMessageRx.h
morantr: Why did you remove the `uint8_t *buff, size_t buff_len` from the parse function and move it to the constructor?
What does is the benefit of that? I don't like it since it would force calling the constructor in class initialize list if cmdu_rx is a member of a class.

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/tlvVendorSpecific.h
```diff
@@ -39,8 +40,11 @@ class tlvVendorSpecific : public BaseClass
         };
         
         const eTlvType& type();
-        uint16_t& length();
+        const uint16_t& length();
```
morantr, line 43: What is the point of passing the length as const reference? We pass const reference only if the size is greater than 4 bytes. which is not the case here. (nitpick)

https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md#functions

morantr (changes requested):

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -209,9 +209,7 @@ bool socket_thread::verify_cmdu(message::sUdsHeader *uds_header)
             if (tlv_vendor_specific.vendor_oui() ==
                 ieee1905_1::tlvVendorSpecific::eVendorOUI::OUI_INTEL) {
                 // assuming that the magic is the first data on the beerocks header
-                auto beerocks_magic = *(
-                    uint32_t *)((uint8_t *)tlv + ieee1905_1::tlvVendorSpecific::get_initial_size());
-
+                auto beerocks_magic = *(uint32_t *)(tlv_vendor_specific.payload());
```
morantr, line 207: We should get rid of the magic in the beerocks action header. It is not relevant anymore, and it is basically a leftover of the area before we used cmdu.

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/tlvVendorSpecific.h
tomereli: This is autogenerated code, this means that the length field should never change. We only pass references in tlvf, so const is needed here.

framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -52,10 +53,21 @@ class CmduMessage {
...
     static const uint16_t kTlvHeaderLength  = 3;
+
+private:
+    // By default, we assume that the CMDU will be sent over the ieee1905_1 transport.
+    // Since the ieee1905_1 transport only supports CMDU fragmantation on TLVs boundaries,
+    // a TLV max size can't be larger than the MTU - CMDU header length - EOM TLV length.
+    // However, when the CMDU is sent internally via UDS, this restriction does not apply.
+    // So we allo changing it via set_max_tlv_length() API.
+    size_t m_max_tlv_length =
+        kMaxCmduLength - kCmduHeaderLength - ieee1905_1::tlvEndOfMessage::get_initial_size();
```
morantr: I'm sorry but this whole commit is featuring a bad solution IMO. Limiting CMDU size to MTU and creating special cases for "jumbo" tlvs is not an elegant solution.
There should not be any size limit expect the defined by the given buffer size, which is:
`buff_len - cmdu_header_len - tlv_end_of_message_len'

Limiting the TLV size should not be handled in the tlvf, it should be handled in the transport layer.
If we have exceeded MTU size, we should get that failure on the BTL or the transport when trying to send the CMDU.

framework/tlvf/src/include/tlvf/ClassList.h
tomereli: Done

framework/tlvf/src/src/CmduMessage.cpp
tomereli: Done

framework/tlvf/src/src/CmduMessageTx.cpp
tomereli: No need for the copy constructor. The latest works - you are looking at a broken commit.
Since this change is so big, I had to include some commits which break things so it would be reviewable. I agreed with Arnout to keep the broken commits so in the future it would be easy to review this change for someone who is not familiar with TLVF.

framework/tlvf/src/include/tlvf/CmduMessageRx.h
tomereli: It makes perfect sense to move the stuff which should never change to the constructor, such as the buffer and buffer length. The benefit is that this is the only way I could easilly integrate ClassList to CmduMessage - otherwise I would have had to do some pretty ugly adaptations whenever create / parse is called.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
tomereli: It still works and it still validates the TLV, so I don't think removing it now is beneficial.
Removing this will be done as part of moving verify_cmdu to the transport.

framework/tlvf/src/include/tlvf/CmduMessage.h
tomereli: The ieee1905 defines fragmantation to be based on TLV boundary, so each TLV must never exceed the MTU - CMDU header length.
Adding the check here is also necessary since we don't know if this CMDU will be transmitted over the transport.

> Limiting the TLV size should not be handled in the tlvf, it should be handled in the transport layer.

Correct - and it is handled there, which is how this was reproduced in the first place - there is an error in the transport when trying to send oversized TLVs - see `ieee1905_transport_packet_processing.cpp::fragment_and_send_packet_to_network_interface()` function.

We can't know in advance if the TLV should be sent over to the network, in which case it will fail if it has an oversized TLV. The best solution is never to send a TLV which is larger then `MTU - CMDU header length - TLV end of message length` - but this is problematic since `network_map` minimum node size is already bigger than 2KB.

The solution of using the jumbo TLVs is the most elegant I could think of, I'm sorry you don't like it, but I think that for now we should take it as is since this PR is already huge and took too long to be merged, we can refactor that in the future if we decide to do so.

framework/tlvf/src/include/tlvf/CmduMessage.h
tomereli: >I don't understand why did you make ClassList be a member of CmduMessage instead of just be it.

Don't understand what you mean by "just be it." - the __whole__ point in this change is to decouple the TLVs from CmduMessage class, which is why we crearted `ClassList` and put it as a member of `CmduMessage`.
It also makes perfect sense - a Cmdu Message is a descriptor of TLVF generated classes, so it includes a member ClassList which holds all the TLVs, where the first class is the CmduHeader and the rest are the TLVs themselves.

>Also, why did you choose to use a member over heritage? It would have save you all the wrapper functions.

The wrapper functions are only for backward compatibility. We chose member over inheritance cause it makes more sense and we don't want CmduMessage to access all the the private and protected members of ClassList.

framework/tlvf/src/include/tlvf/CmduMessage.h
morantr: > We can't know in advance if the TLV should be sent over to the network, in which case it will fail if it has an oversized TLV. The best solution is never to send a TLV which is larger then MTU - CMDU header length - TLV end of message length - but this is problematic since network_map minimum node size is already bigger than 2KB.

You don't need to know it in advance. You should get an error when trying to send such TLV to the bus. So if such an error exists on the transport, then the only change you can do in this PR is to set the tlv limitation to `buf_len - CMDU header length - TLV end of message length` and remove the mumbo jumbo commits.

tomereli: Moran - Well done, I just misunderstood you but now with the resize solution in place there is no longer any need to limit the TLV size, so I removed the relevant commits. Thanks for finding that!

https://media.giphy.com/media/vCKC987OpQAco/giphy.gif

morantr (approved):

framework/tlvf/src/src/CmduMessage.cpp
tomereli: Will be addressed in #601, hence leaving open.

framework/tlvf/src/include/tlvf/CmduMessageRx.h
tomereli: I tried doing it in a separate commit but it became a spaghetti of changes so decided not to do that eventually.

framework/tlvf/src/include/tlvf/CmduMessageRx.h
tomereli: Yes, but at this point what is there to gain from it? I prefer not to block merge for this PR for small things like this one since every new change that enters prplMesh may break this one (we'll need to find all those places we need to replace `cmdu_tx.addClass()` with `beerocks_header->addClass()`)

framework/tlvf/src/src/CmduMessageRx.cpp
tomereli: https://github.com/prplfoundation/prplMesh/pull/584#discussion_r360714833

framework/tlvf/src/src/CmduMessageTx.cpp
tomereli: https://github.com/prplfoundation/prplMesh/pull/584#discussion_r360714833

framework/tlvf/src/src/CmduMessageTx.cpp
tomereli: https://github.com/prplfoundation/prplMesh/pull/584#discussion_r360714833

framework/tlvf/src/src/CmduMessageTx.cpp
tomereli: I don't understand why we should keep this, but keeping open so we can review it again in #601

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
tomereli: I think it is still good practice to delete this explicitly.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
tomereli: I think not, better to keep it generic by using `getClass` to make sure we cover all corner cases.

framework/tlvf/tlvf.py
tomereli: Will be fixed as part of #601

framework/tlvf/src/src/BaseClass.cpp
tomereli: Fixed

framework/tlvf/src/include/tlvf/ClassList.h
tomereli: But is there any need to keep it if it's not used? I think not.

framework/tlvf/src/include/tlvf/ClassList.h
tomereli: How not? we are declaring multiple functions which return shared_ptr to BaseClass, this is needed.

framework/tlvf/src/src/BaseClass.cpp
tomereli: Will be reviewed as part of #601

PR #585, opened 2019-12-18
Feature/move post init check to addClass

CoralMalachi:
Preparative PR.

Move the post init check to addClass method to prevent the user from adding a
new TLV before the last inserted TLV is not fully initialized.

In the future, this check also validate that last TLV does not contain a
tailroom the was not trimmed yet.

Also, a test is added to demonstrate a situation of trying to use the addClass
method before fully initialize the last TLV

### DIFF
--- a/framework/tlvf/src/include/tlvf/CmduMessage.h
+++ b/framework/tlvf/src/include/tlvf/CmduMessage.h
@@ -12,6 +12,7 @@
 #include <memory>
 #include <tlvf/ieee_1905_1/cCmduHeader.h>
 #include <tlvf/ieee_1905_1/eTlvType.h>
+#include <tlvf/tlvflogging.h>
 #include <vector>
 
 namespace ieee1905_1 {
@@ -28,6 +29,16 @@ public:
     template <class T, class = typename std::enable_if<std::is_base_of<BaseClass, T>::value>::type>
     std::shared_ptr<T> addClass()
     {
+
+        auto last_tlv = m_class_vector.empty() ? nullptr : m_class_vector.back();
+        // do not allow to use addClass method if the last tlv is not fully initialized
+        if (last_tlv) {
+            if (!last_tlv->isPostInitSucceeded()) {
+                TLVF_LOG(ERROR) << "TLV post init failed";
+                return nullptr;
+            }
+        }
+
         std::shared_ptr<T> ptr;
         if (m_cmdu_header) {
             if (m_class_vector.size() == 0) {
--- a/framework/tlvf/src/src/CmduMessageTx.cpp
+++ b/framework/tlvf/src/src/CmduMessageTx.cpp
@@ -83,12 +83,17 @@ bool CmduMessageTx::finalize(bool swap_needed)
     if (!m_cmdu_header)
         return false;
 
-    // Update vendor specific length which is seperated to 3 classes
-    for (auto class_it = m_class_vector.begin(); class_it != m_class_vector.end(); class_it++) {
-        if (!(*class_it)->isPostInitSucceeded()) {
+    // validate that last tlv of the CMDU is fully initialized
+    auto last_tlv = m_class_vector.empty() ? nullptr : m_class_vector.back();
+    if (last_tlv) {
+        if (!last_tlv->isPostInitSucceeded()) {
             TLVF_LOG(ERROR) << "TLV post init failed";
             return false;
         }
+    }
+
+    // Update vendor specific length which is seperated to 3 classes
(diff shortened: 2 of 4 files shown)


framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -28,6 +29,16 @@ class CmduMessage {
...
     {
+        
+        // do not allow to use addClass method if the last tlv is not fully initialized        
+        if(m_class_vector.size() != 0){
+            auto last_tlv = m_class_vector.back();
+            if(!(*last_tlv).isPostInitSucceeded()){
+            TLVF_LOG(ERROR) << "TLV post init failed";
+            return nullptr;
+            }
+        }
```
arnout: This is full of formatting errors. Did you run clang-format?

Instead of checking m_class_vector.size(), it would be more natural to do:

```
auto last_tlv = m_class_vector.back();
if (last_tlv) {
    TLVF_LOG(ERROR) << "TLV post init failed";
    return nullptr;
}
```

framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -28,6 +29,16 @@ class CmduMessage {
     template <class T, class = typename std::enable_if<std::is_base_of<BaseClass, T>::value>::type>
     std::shared_ptr<T> addClass()
     {
+        
+        // do not allow to use addClass method if the last tlv is not fully initialized        
+        if(m_class_vector.size() != 0){
+            auto last_tlv = m_class_vector.back();
+            if(!(*last_tlv).isPostInitSucceeded()){
+            TLVF_LOG(ERROR) << "TLV post init failed";
```
tomereli: Indentation error

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -81,14 +81,19 @@ bool CmduMessageTx::finalize(bool swap_needed)
...
 
     if (!m_cmdu_header)
-        return false; 
+        return false;
 
-    // Update vendor specific length which is seperated to 3 classes
-    for (auto class_it = m_class_vector.begin(); class_it != m_class_vector.end(); class_it++) {
-        if(!(*class_it)->isPostInitSucceeded()){
+    // validate that last tlv of the CMDU is fully initialized
+    if(m_class_vector.size() != 0) {
```
arnout: Same remarks:

- no need to check size, just check if last_tlv is valid;
- formatting is not OK.

framework/tlvf/src/include/tlvf/CmduMessage.h
tomereli: Much shorter option:
```
if (!m_class_vector.empty() &&
    !m_class_vector.back()->isPostInitSucceeded()) {
    TLVF_LOG(ERROR) << "TLV post init failed";
    return nullptr;
}
```

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -307,14 +307,22 @@ int test_parser()
     auto tlv2 = msg.addClass<tlvLinkMetricQuery>();
     auto tlv3 = msg.addClass<tlvWscM1>();
     // TODO https://github.com/prplfoundation/prplMesh/issues/480
```
arnout: This todo is about the `tlv3->add_vendor_ext(tlv3->create_vendor_ext())` part, so it should stay close to it. Or rather, it should move to just in front of `tlv4->add_var1(tlv4->create_var1())` which has a copy of the same comment because it also applies to it.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -307,14 +307,22 @@ int test_parser()
...
-    tlv3->add_vendor_ext(tlv3->create_vendor_ext());
+
+    // Trying to add a new class before fully initialize last class should lead to error
     auto tlv4 = msg.addClass<tlvTestVarList>();
+    if (tlv4) {
+        LOG(ERROR) << "addClass should fail since the the last tlv is not fully initialized";
```
arnout: previous tlv is more appropriate here than last tlv (nitpick).

arnout (approved):
A PR that includes a test and a documentation update! I love it!

A few comments that are either nitpicky or can be applied while rebasing, so approving already.

tomereli (approved):
small nitpick comments

framework/tlvf/src/include/tlvf/CmduMessage.h
tomereli: Actually I like @arnout 's style better so use his :)

PR #586, opened 2019-12-18
.gitlab-ci.yml: build for the Turris Omnia

rmelotte:
Add a new hidden job: build-for-openwrt. It is defined to make it easy
to build for different targets. It has to be a hidden job because it's
not supposed to be run as-is.

Also add a new job extending .build-for-openwrt, to build for the
Turris Omnia.

Add a "needs" for build-in-docker to the run-tests job, so that it
doesn't wait for the OpenWrt build nor use its artifacts.

While we're at it, remove the duplicate "artifacts" for the
run_tests job.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,14 +36,30 @@ run-tests:
       - logs
     when: always
 
-  artifacts:
-    paths:
-      - logs
-    when: always
-
   tags:
     - docker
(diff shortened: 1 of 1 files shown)


tomereli (approved):

arnout (approved):
I've added to the commit message something about the reuse of docker layers.

PR #596, opened 2019-12-19
Feature/ci unique openwrt builder names

rmelotte:
- Make it possible to choose the image tag. While doing that, switch to using command line options instead of files when calling the build script.
- Use unique names for the openwrt builder, so that we can safely run multiple jobs at the same time. 

This should make the CI pipelines to be picked up faster, since it will no longer be limited to 1 job at a time (once we make the related change to the runner).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -49,7 +49,7 @@ run-tests:
   stage: build
   script:
     - cd tools/docker/builder/openwrt
-    - ./build.sh targets/$TARGET.args 2>&1 | tee build.log | grep -E '^make\[[12]\]|^Step' --color=never --line-buffered
+    - ./build.sh -v -d "$TARGET" -t "prplmesh-builder-$TARGET:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee build.log | grep '^make\[[12]\]\|^Step\|^ --->' --color=never --line-buffered
   artifacts:
     paths:
       - tools/docker/builder/openwrt/*.ipk
@@ -70,7 +70,7 @@ run-tests:
 build-for-turris-omnia:
   extends: .build-for-openwrt
   variables:
-    TARGET: turris-omnia
+    TARGET: "turris-omnia"
 
 test-on-turris-omnia:
   extends: .test-on-target
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -49,7 +49,7 @@ run-tests:
   stage: build
   script:
     - cd tools/docker/builder/openwrt
-    - ./build.sh -v -d "$TARGET" -t "prplmesh-builder-$TARGET:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee build.log | grep '^make\[[12]\]\|^Step\|^ --->' --color=never --line-buffered
+    - ./build.sh -v -d "$TARGET_DEVICE" -t "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee build.log | grep '^make\[[12]\]\|^Step\|^ --->' --color=never --line-buffered
   artifacts:
     paths:
       - tools/docker/builder/openwrt/*.ipk
@@ -62,18 +62,18 @@ run-tests:
 .test-on-target:
   stage: test
   script:
-    - tools/deploy_ipk.sh $TARGET_DEVICE tools/docker/builder/openwrt/*.ipk
-    - tools/docker/tests/openwrt/test_status.sh $TARGET_DEVICE
+    - tools/deploy_ipk.sh $TARGET_DEVICE_NAME tools/docker/builder/openwrt/*.ipk
+    - tools/docker/tests/openwrt/test_status.sh $TARGET_DEVICE_NAME
   tags:
     - targets
 
 build-for-turris-omnia:
   extends: .build-for-openwrt
   variables:
-    TARGET: "turris-omnia"
+    TARGET_DEVICE: "turris-omnia"
 
 test-on-turris-omnia:
   extends: .test-on-target
   variables:
-    TARGET_DEVICE: turris-omnia-1
+    TARGET_DEVICE_NAME: turris-omnia-1
   needs: ["build-for-turris-omnia"]
--- a/tools/docker/builder/openwrt/build.sh
+++ b/tools/docker/builder/openwrt/build.sh
@@ -3,75 +3,120 @@
 scriptdir="$(cd "${0%/*}"; pwd)"
 rootdir="${scriptdir%/*/*/*/*}"
 
-if [ ! -f "$1" ] ; then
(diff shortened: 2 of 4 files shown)


tools/docker/builder/openwrt/build.sh
```diff
@@ -3,75 +3,123 @@
...
-       "$image_tag" \
-       ./build_scripts/build.sh
-
-docker cp "${container_name}:/home/openwrt/openwrt_sdk/prplmesh-${TARGET}-${TARGET_PLATFORM_TYPE}-${OPENWRT_VERSION}-${PRPLMESH_VERSION}.ipk" .
-
-docker rm "${container_name}"
+. "${rootdir}/tools/docker/functions.sh"
+
+usage() {
+    echo "usage: $(basename $0) [-ipt]"
```
tomereli: short options are missing some of the long options

tools/docker/builder/openwrt/build.sh
```diff
@@ -3,75 +3,123 @@
...
+    fi
+
+    if [ -n "$TAG" ] ; then
+        image_tag=$TAG
+    else
+        image_tag="prplmesh-builder-${TARGET}:${OPENWRT_VERSION}"
+        dbg "image tag not set, using default value $image_tag"
+    fi
+
+    export OPENWRT_REPOSITORY
```
tomereli, line 97: Why do we need empty exports?

tools/docker/builder/openwrt/build.sh
```diff
@@ -3,75 +3,123 @@
...
+            TARGET_PLATFORM_TYPE=turris-omnia
+            ;;
+        *)
+            err "Unknown target device: $TARGET_DEVICE"
+            info "Currently supported targets are: turris-omnia"
+            ;;
+    esac
+
+    if [ -z "$OPENWRT_REPOSITORY" ] ; then
+        OPENWRT_REPOSITORY=https://git.prpl.dev/prplmesh/prplwrt.git
```
tomereli: I prefer to declare all defaults outside of `main()` where we put `VERBOSE`, then simply update them directly in the `getopt` loop if selected.

tools/docker/builder/openwrt/build.sh
```diff
@@ -3,75 +3,123 @@
...
+            -o | --openwrt-version)    OPENWRT_VERSION="$2"; shift; shift ;;
+            -r | --openwrt-repository) OPENWRT_REPOSITORY="$2"; shift; shift ;;
+            -t | --tag)                TAG="$2"; shift ; shift ;;
+            -v | --verbose)            VERBOSE=true; shift ;;
+            -- ) shift; break ;;
+            * ) err "unsupported argument $1"; usage; exit 1 ;;
+        esac
+    done
+
+    case "$TARGET_DEVICE" in
```
tomereli: I prefer to add this `case` to the `while` loop above and treat it like any other argument - by overriding defaults which are declared outside of `main`.

tools/docker/builder/openwrt/build.sh
```diff
@@ -3,75 +3,123 @@
...
+    # We first need to build the corresponding images
+    docker build --tag "$image_tag" \
+           --build-arg OPENWRT_REPOSITORY \
+           --build-arg OPENWRT_VERSION \
+           --build-arg TARGET \
+           --build-arg PRPL_FEED \
+           "$scriptdir/"
+}
+
+build_prplmesh() {
```
tomereli, line 32: I think it is better to explicitly pass parameters to this function and use $1 $2 etc. into local parameters (nitpick)

.gitlab-ci.yml
```diff
@@ -49,7 +49,7 @@ run-tests:
   stage: build
   script:
     - cd tools/docker/builder/openwrt
-    - ./build.sh targets/$TARGET.args 2>&1 | tee build.log | grep -E '^make\[[12]\]|^Step' --color=never --line-buffered
+    - ./build.sh -v -d "$TARGET" -t "prplmesh-builder-$TARGET:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee build.log | grep '^make\[[12]\]\|^Step\|^ --->' --color=never --line-buffered
```
tomereli: I think its better to change `$TARGET` to `$TARGET_DEVICE` since it will be more readable this way IMO.

tomereli (approved):
Mainly nitpicks so already approving, please fix the comments or reply before you merge.

tools/docker/builder/openwrt/build.sh
rmelotte: They need to be exported for `docker build` to be able to use them. The other option is to use `docker build --build-arg OPTION=$OPTION` everywhere (i.e. specifying the arg value explicitly), but I didn't want to type the args twice everywhere.
If there is a good reason to do it I can still change it.

tools/docker/builder/openwrt/build.sh
rmelotte: Done except for "image_tag", which depend on other options.

tools/docker/builder/openwrt/build.sh
rmelotte: Done.

tools/docker/builder/openwrt/build.sh
rmelotte: Done.

.gitlab-ci.yml
rmelotte: Done in a separate commit, since the variables are already in master, and the changes in this PR are not related.

tools/docker/builder/openwrt/build.sh
```diff
@@ -56,7 +57,19 @@ main() {
...
+		    turris-omnia)
+			TARGET=mvebu
+			;;
+		    *)
+			err "Unknown target device: $TARGET_DEVICE"
+			info "Currently supported targets are: turris-omnia"
+			;;
+		esac
+		;;
+	    -f | --prpl-feed)          PRPL_FEED="$2"; shift; shift ;;
```
tomereli: missing a space..

tools/docker/builder/openwrt/build.sh
```diff
@@ -118,5 +113,8 @@ main() {
 
 IMAGE_ONLY=false
 VERBOSE=false
+OPENWRT_REPOSITORY='https://git.prpl.dev/prplmesh/prplwrt.git'
+OPENWRT_VERSION='9d2efd'
+PRPL_FEED='https://git.prpl.dev/prplmesh/iwlwav.git^06a0126d5fb53b1d65bad90757a5f9f5f77419ca'
```
tomereli, line 118: Great!

tomereli (approved):
LGTM

tools/docker/builder/openwrt/build.sh
rmelotte: Crap, tabs vs spaces again. Let me untabify it.

rmelotte:
Don't merge yet, the last commit is missing something (the fixup also needs to be squashed ofc).

rmelotte:
Rebased on master, squashed the fixups.
There are 2 open review comments, other than I think it can be merged.

PR #602, opened 2019-12-23
Feature/wsc attrlist

### LINKED ISSUE #160: More flexible to parse TLVs in autoconfiguration WSC message.

### LINKED ISSUE #471: [TASK] TLVF add WSC attributes support

tomereli:
The design done as part of #597 and implemented in #584 provided a generic infrastructure for creating and parsing lists of classes. The reason for this refactor was to provide the infrastructure for supporting a list of wsc attributes in TLVF since currently, prplMesh has YAML files for WSC M1 and WSC M2 TLVs with fixed attributes, which means that if a WSC message received from a *different* controller or agent, with slightly different attributes order and/or less/more attributes, we will fail in parsing.

This PR adds `class AttrList` class which derives from `ClassList` and is used to create and parse WSC attributes. The `AttrList` is a pure virtual class that provides `create()` and `parse()` factory methods which are used to create or parse a WSC attributes list using derived classes - m1 and m2.
The design can be extended to support any WSC message, but for EasyMesh we only need m1 and m2.

The implementation of m1 and m2 is simple and aims to provide full backward compatability to old prplMesh versions since it follows the same order of attributes defined in the tlvWscM2.yaml and tlvWscM1.yaml files.

With this, prplMesh *should* be able to parse any WSC message received from any controller and agent.

Fixes #471 
Fixes #160


tomereli:
[TLVF WSC TLV parser](https://trello.com/c/Hhp90qPO/21-tlvf-wsc-tlv-parser)

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -680,3 +680,107 @@ cWscAttrOsVersion:
...
+
+cWscAttrMac:
+  _type: class
+  _is_tlv_class: True
+  type: 
+    _type: eWscAttributes
+    _value: ATTR_MAC_ADDR
+  length:
+    _type: uint16_t
+    _value_const: WSC_MAC_LENGTH
```
vitalybu, line 440: Why do we need const values inside the message?
It's part of the EasyMesh specification?

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+    uint16_t type;
+    uint16_t length;
+    uint8_t *data() { return ((uint8_t *)this + sizeof(*this)); }
+} __attribute__((packed)) sWscAttrHeader;
+
+/**
+ * @brief wsc configuration
+ * 
+ */
+struct config {
```
vitalybu, line 30: `config` is a very generic name... I would consider changing this to a more meaningful name (especially since it's in a header file)

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+#include <tlvf/ieee_1905_1/tlvWsc.h>
+#include <tlvf/tlvflogging.h>
+
+namespace WSC {
+
+typedef struct sWscAttrHeader {
+    uint16_t type;
+    uint16_t length;
+    uint8_t *data() { return ((uint8_t *)this + sizeof(*this)); }
+} __attribute__((packed)) sWscAttrHeader;
```
vitalybu, line 24: I'm not sure what's the behavior of all compilers when packing a struct that also contains methods...

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+    uint8_t iv[WSC_ENCRYPTED_SETTINGS_IV_LENGTH];
+};
+
+class AttrList : public ClassList {
+protected:
+    AttrList(uint8_t *buff, size_t buff_len, bool parse) : ClassList(buff, buff_len, parse) {}
+
+public:
+    virtual ~AttrList()                  = default;
+    virtual eWscMessageType type()       = 0;
```
vitalybu: Why is this a pure-virtual method and not say a simple getter method?
Won't it be easier to pass the Type in the constructor and have the base class simply return it?
Anyhow, getter method should probably be defined as const (e.g. `virtual eWscMessageType type() const = 0;`)

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+    uint16_t getNextAttrType() { return ntohs(getNextAttrHdr()->type); };
+    size_t getRemainingBytes()
+    {
+        return m_class_vector.empty() ? getMessageBuffLength()
+                                      : m_class_vector.back()->getBuffRemainingBytes();
+    };
+    virtual bool valid_custom() = 0;
+
+private:
+    static eWscMessageType wsc_msg_type_from_buffer(uint8_t *buff, size_t buff_len);
```
vitalybu: Any reason to not convert this method to static function in the cpp file?
I prefer using stand-alone static functions as it cleans up the header file (which is the actual interface to the class).

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+};
+
+class m1 : public AttrList {
```
vitalybu: Why did you decide to keep it in the same header file?

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+    virtual eWscMessageType type() override { return eWscMessageType::WSC_MSG_TYPE_M2; };
+    bool init(const config &cfg) override;
+    bool valid_custom() override;
+
+    // getters
+    uint8_t *authenticator() { return getAttr<cWscAttrAuthenticator>()->data(); };
+    uint8_t *registrar_nonce() { return getAttr<cWscAttrRegistrarNonce>()->nonce(); };
+    cWscAttrEncryptedSettings &encrypted_settings()
+    {
+        return *getAttr<cWscAttrEncryptedSettings>();
```
vitalybu: What if this attribute doesn't exists? You'll dereference a nullptr?
You should either make sure it exists or return the `shared_ptr`

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+        if (!attributes->init(cfg))
+            return nullptr;
+        attributes->finalize();
+        tlv.addInnerClassList(attributes);
+    } else if (cfg.msg_type == eWscMessageType::WSC_MSG_TYPE_M2) {
+        attributes = std::make_shared<m2>(tlv.payload(), tlv.payload_length(), false);
+        if (!attributes->init(cfg))
+            return nullptr;
+        attributes->finalize();
+        tlv.addInnerClassList(attributes);
```
vitalybu: This part is duplicated between the M1 and M2. You can move it outside the if statement.

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+}
+
+eWscMessageType AttrList::wsc_msg_type_from_buffer(uint8_t *buff, size_t buff_len)
+{
+    auto attr                = reinterpret_cast<sWscAttrHeader *>(buff);
+    size_t attributes_length = 0;
+    while (ntohs(attr->type) != ATTR_MSG_TYPE) {
+        attributes_length += ntohs(attr->length) + sizeof(sWscAttrHeader);
+        if (attributes_length > buff_len)
+            return eWscMessageType::WSC_MSG_TYPE_INVALID;
```
vitalybu, line 49: Is this the correct error code for this?

framework/tlvf/src/include/tlvf/WSC/AttrList.h
vitalybu: Also, the input pointer type should probably be `void*`. Can prevent casting issues.

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+        attr = reinterpret_cast<sWscAttrHeader *>(buff + attributes_length);
+    }
+    return static_cast<eWscMessageType>(*attr->data());
+}
+
+std::shared_ptr<AttrList> AttrList::parse(ieee1905_1::tlvWsc &tlv)
```
vitalybu, line 55: `tlv` isn't modified in this method. Can be changed to `const`.

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+    //Check if its M1 or M2
+    auto msg_type = wsc_msg_type_from_buffer(tlv.payload(), tlv.payload_length());
+    if (msg_type == eWscMessageType::WSC_MSG_TYPE_M1) {
+        attributes = std::make_shared<m1>(tlv.payload(), tlv.payload_length(), true);
+        if (!attributes->init())
+            return nullptr;
+    } else if (msg_type == eWscMessageType::WSC_MSG_TYPE_M2) {
+        attributes = std::make_shared<m2>(tlv.payload(), tlv.payload_length(), true);
+        if (!attributes->init())
+            return nullptr;
```
vitalybu: Move outside the if statement

vitalybu: Also, you should make sure that the allocation was successful:
`if (!attributes || !attributes->init())`

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+    }
+    return attributes;
+}
+
+bool AttrList::valid()
+{
+    if (!getAttr<cWscAttrVersion>())
+        return false;
```
vitalybu, line 80: Why there's no logging for errors?

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+        case ATTR_AUTHENTICATOR:
+            if (!addAttr<cWscAttrAuthenticator>())
+                return false;
+            break;
+        // TODO these are not expected attributes (part of config data)
+        case ATTR_SSID:
+        case ATTR_NETWORK_KEY:
+        case ATTR_KEY_WRAP_AUTH:
+        case ATTR_ENCR_TYPE:
+        case ATTR_AUTH_TYPE:
```
vitalybu: Should add our `FALLTHROUGH` macro. Otherwise it can generate build errors on some compilers.

vitalybu:
Reviewed up to the "agent: create M1 using WSC::m1" commit

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
tomereli: A WSC attribute has a length and data, as a TLV does.
Specifying the length as `_value_const` is a tlvf feature which basically sets the `length()` getter to return a const so it can't be changed accidentally.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: It is part of the WSC namespace - so the usage is `WSC::config`, don't you think its enough? it is precisely that - a WSC configuration for creating WSC messages. Do you have a better name to suggest?

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: It is not compiler related AFAIK - the only difference between a struct and class in C++ is the default accessibility of member variables and methods. In a struct they are public; in a class they are private, so there shouldn't be any dependency on one compiler or the other.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: I guess it could be a simple getter, just like the other common attributes getters defined in the AttrList class.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
vitalybu: Depends on the amount of classes in the WSC namespace. If you believe there won't be any other "configuration" structure and/or related data-types it's probably fine.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: No specific reason, just for convenience. I'll split to separate headers and files now.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
vitalybu: Right, but you don't use classes the same way you use structs.
For example, you wouldn't `memcpy` into a class or "pack" its' members...
What I meant is that I don't know how the compiler will "pack" the members, since it also contains a method.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: The design of these classes is that the `AttrList` is an abstract class with factory methods to create and parse M1 and M2 messages. The `AttrList::create()` uses `wsc_msg_type_from_buffer()` to get the message type and then creates a `shared_ptr` of the appropriate derived class - M1 or M2.
Then it calls one of the 2 versions of the `init()` virtual function which calls `addAttr<T>` for all attributes which belong in this message.
Finally, the `valid()` function validates that the message includes all the mandatory attributes, for those we have the getter functions.
Bottom line - M2 and M1 should __only__ be created using one of the factory methods - 
 `AttrList::create()` or `AttrList::parse()`, which ensures all the attributes are there and can be dereferenced. This makes the rest of the code much simpler, especially for the callers which remain almost identical to how they were before this change.

framework/tlvf/src/src/WSC/AttrList.cpp
tomereli: Yes, why would you think otherwise?

framework/tlvf/src/src/WSC/AttrList.cpp
tomereli: Good point, I'll add logging in the valid() methods.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
vitalybu: ACK

framework/tlvf/src/src/WSC/AttrList.cpp
tomereli: It is modified - we are adding the WSC attributes in the TLV payload, can't be const.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: I see, but I prefer not to change this based on a hunch. We can count on the CI to catch it if there is an issue with it.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: I think its fine, if there will be an issue we can always change it in the future.

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -558,7 +558,7 @@ cWscAttrModelName:
     _length_var: True
   model:
     _type: char
-    _length: [ length ]
+    _length: []
```
morantr, line 309: This change is fishy, I don't understand how `length` will be updated when allocating the string.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+    template <class T> std::shared_ptr<T> addAttr() { return this->addClass<T>(); };
+    bool finalize() { return ClassList::finalize(); };
+    size_t len() { return this->getMessageLength(); };
+    uint8_t *buffer() { return this->getMessageBuff(); };
+
+    // getter functions for common attributes for m1 and m2
+    std::string manufacturer() { return getAttr<cWscAttrManufacturer>()->manufacturer_str(); };
+    std::string model_name() { return getAttr<cWscAttrModelName>()->model_str(); };
+    std::string device_name() { return getAttr<cWscAttrDeviceName>()->device_name_str(); };
+    std::string serial_number() { return getAttr<cWscAttrSerialNumber>()->serial_number_str(); };
```
morantr: Can you change the return value to `const std::string &`?

framework/tlvf/src/include/tlvf/WSC/AttrList.h
```diff
@@ -0,0 +1,139 @@
...
+    uint16_t rf_bands() { return getAttr<cWscAttrRfBands>()->bands(); };
+    bool valid();
+
+protected:
+    sWscAttrHeader *getNextAttrHdr()
+    {
+        return reinterpret_cast<sWscAttrHeader *>(
+            m_class_vector.empty() ? getMessageBuff() : m_class_vector.back()->getBuffPtr());
+    }
+    uint16_t getNextAttrType() { return ntohs(getNextAttrHdr()->type); };
```
morantr, line 100: I think it would be nicer if we use `swap_16()` from "swap.h" instead of `ntohs`.
Same on function `AttrList::wsc_msg_type_from_buffer(uint8_t *buff, size_t buff_len)`

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+
+std::shared_ptr<AttrList> AttrList::create(ieee1905_1::tlvWsc &tlv, const config &cfg)
+{
+    std::shared_ptr<AttrList> attributes = nullptr;
+    if (cfg.msg_type == eWscMessageType::WSC_MSG_TYPE_M1) {
+        attributes = std::make_shared<m1>(tlv.payload(), tlv.payload_length(), false);
+        if (!attributes->init(cfg))
+            return nullptr;
+        attributes->finalize();
+        tlv.addInnerClassList(attributes);
```
morantr: Need to add a check that the memory has been successfully allocated:
`if (!attributes || !attributes->init(cfg)) { return nullptr; }`
This should be done in several places on this file. Do it in all the places you use `make_shared`.

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+
+    auto mac_attr = addAttr<WSC::cWscAttrMac>();
+    if (!mac_attr)
+        return false;
+    mac_attr->data() = cfg.mac;
+
+    auto enrolee_nonce_attr = addAttr<WSC::cWscAttrEnrolleeNonce>();
+    if (!enrolee_nonce_attr)
+        return false;
+    std::copy(cfg.enrollee_nonce, cfg.enrollee_nonce + WSC_NONCE_LENGTH,
```
morantr: If you will use `std::copy_n(src, len, dest)` it would look nices since you will not have to do `cfg.enrollee_nonce + WSC_NONCE_LENGTH`, instead replace it with just `WSC_NONCE_LENGTH`.
(nitpick)

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,504 @@
...
+    if (!getAttr<cWscAttrRegistrarNonce>())
+        return false;
+    if (!getAttr<cWscAttrVersion2>())
+        return false;
+    if (!getAttr<cWscAttrEncryptedSettings>())
+        return false;
+    if (!getAttr<cWscAttrAuthenticator>())
+        return false;
+    return false;
+}
```
morantr: missing last line break

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4569,65 +4571,45 @@ bool slave_thread::add_radio_basic_capabilities()
...
+    size_t payload_length =
+        tlv->getBuffRemainingBytes() - ieee1905_1::tlvEndOfMessage::get_initial_size();
+    tlv->alloc_payload(payload_length);
+
+    WSC::config cfg;
+    cfg.msg_type = WSC::eWscMessageType::WSC_MSG_TYPE_M1;
+    cfg.mac      = network_utils::mac_from_string(backhaul_params.bridge_mac);
+    dh           = std::make_unique<mapf::encryption::diffie_hellman>();
+    std::copy(dh->nonce(), dh->nonce() + dh->nonce_length(), cfg.enrollee_nonce);
+    std::copy(dh->pubkey(), dh->pubkey() + dh->pubkey_length(), cfg.pub_key);
```
morantr, line 4592: std::copy_n (nitpick)

morantr (approved):

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
tomereli: As explained oraly, since WSC is also a TLV, it will be updated automatically as its a feature of unknown length lists in TLVF TLV classes.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: No:
```
./framework/tlvf/src/include/tlvf/WSC/AttrList.h: In member function 'const string& WSC::AttrList::serial_number() const':                                                                                 
../framework/tlvf/src/include/tlvf/WSC/AttrList.h:74:106: error: returning reference to temporary [-Werror=return-local-addr]                                                                               
     const std::string& serial_number() const { return getAttr<cWscAttrSerialNumber>()->serial_number_str(); };  
```

This is because the `manufacturer_str()` returns a temporary std::string() object:

```
std::string cWscAttrManufacturer::manufacturer_str() {
    char *manufacturer_ = manufacturer();
    if (!manufacturer_) { return std::string(); }
    return std::string(manufacturer_, m_manufacturer_idx__);
}
```

tomereli: I did however go over the code and added const to functions which can be const

framework/tlvf/src/src/WSC/AttrList.cpp
tomereli: Done

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Was already like that so I prefer not to change it as part of this PR.

framework/tlvf/src/include/tlvf/WSC/AttrList.h
tomereli: Actually I really hate that we use swap_* instead of normal linux API ntohs and friends. Using the `ntohX` version makes it perfectly clear that we are changing endianess to/from network/host byte order and not just swapping bytes without a reason. This is why these APIs exist..

tomereli:
> Reviewed up to the "agent: create M1 using WSC::m1" commit

I rebased and squashed with fixups to all your and Moran's comments, leaving open so you can finish reviewing tomorrow.

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -0,0 +1,344 @@
...
+                return false;
+            }
+            break;
+        case ATTR_AUTHENTICATOR:
+            if (!addAttr<cWscAttrAuthenticator>()) {
+                TLVF_LOG(ERROR) << "Failed to add cWscAttrAuthenticator";
+                return false;
+            }
+            break;
+        // TODO these are not expected attributes (part of config data)
```
vitalybu: Although I commented on the missing FALLTHROUGH attribute, maybe it's better to simply remove unexpected cases.
All unexpected attributes will be caught by the "default" case anyway.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -843,23 +840,21 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(const std::string &sr
...
-                      << tlvwscM1->encryption_type_flags_attr().data << " need 0x"
+                      << m1->encr_type_flags() << " need 0x"
                       << uint16_t(bss_info_conf.encryption_type);
         }
         if (num_bsss >= radio_basic_caps->maximum_number_of_bsss_supported()) {
             LOG(INFO) << "Configured #BSS exceeds maximum for " << al_mac << " radio " << ruid;
             break;
         }
-        if (!autoconfig_wsc_add_m2(tlvwscM1, &bss_info_conf)) {
+        if (!autoconfig_wsc_add_m2(*m1, &bss_info_conf)) {
```
vitalybu, line 853: Why not change the argument of this method to reference as well?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3622,23 +3622,23 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
  * @param[out] keywrapkey 16 bytes calculated key wrap key
  * @return true on success
  * @return false on failure
  */
-bool slave_thread::autoconfig_wsc_calculate_keys(std::shared_ptr<ieee1905_1::tlvWscM2> m2,
-                                                 uint8_t authkey[32], uint8_t keywrapkey[16])
+bool slave_thread::autoconfig_wsc_calculate_keys(WSC::m2 &m2, uint8_t authkey[32],
+                                                 uint8_t keywrapkey[16])
 {
     auto mac = network_utils::mac_from_string(backhaul_params.bridge_mac);
```
vitalybu: No need to "waste" time on converting the MAC address if the DH member is not initialized

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3854,7 +3853,7 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
     LOG(DEBUG) << "Received AP_AUTOCONFIGURATION_WSC_MESSAGE";
 
     std::shared_ptr<wfa_map::tlvApRadioIdentifier> ruid = nullptr;
-    std::vector<std::shared_ptr<ieee1905_1::tlvWscM2>> m2_list;
+    std::vector<std::shared_ptr<WSC::m2>> m2_list;
```
vitalybu: As a general rule of thumb, unless there's a need for index based access to elements, it's better to use `std::list` as it's more predictable in runtime complexity and memory allocations.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -235,46 +236,58 @@ bool add_encrypted_settings(std::shared_ptr<tlvWscM2> m2, uint8_t *keywrapkey, u
     int cipherlen                = datalen + 16;
     uint8_t encrypted[cipherlen] = {0};
     std::copy_n(config_data.getStartBuffPtr(), config_data.getLen(), encrypted);
-    if (!mapf::encryption::aes_encrypt(keywrapkey, iv, encrypted, datalen, encrypted, cipherlen)) {
+    if (!mapf::encryption::aes_encrypt(keywrapkey, cfg.iv, encrypted, datalen, encrypted,
+                                       cipherlen)) {
         LOG(DEBUG) << "aes encrypt";
```
vitalybu: That's an interesting ERROR message :)

controller/src/beerocks/master/son_master_thread.cpp
tomereli: m1 is created in this method, it is not passed as an argument.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Will change it in a separate commit.

PR #603, opened 2019-12-23
framework: bpl: allow any target folder when looking for libbridge

rmelotte:
When compiling for the rax40 using OpenWrt, libbridge is not found
because it doesn't match the GLOB that is being used.

When compiling with OpenWrt's SDK for example, libbridge can be in:
target-mips_24kc_musl/linux-ath79_generic/bridge-utils-1.6/libbridge/libbridge.h

Change the GLOB so that libbridge can be found no matter what the
target folder is.


tomereli (approved):

rmelotte:
The target being different was a bug, this is not needed anymore.

PR #608, opened 2019-12-26
Feature/vap list for backhaul

YuriyMasechko:
Each son slave sends a VAP list update to the controller once the VAPs are read. Send this VAP list update also to the backhaul, with the RUID. This will be used to store the VAPs configuration in the UCC listener so it can use it to answer `dev_get_parameter` from the UCC.


agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -64,6 +64,25 @@ void agent_ucc_listener::clear_configuration()
...
+ * @brief Update list of VAPs for ruid.
+ * 
+ * @return None.
+ */
+
+typedef struct {
+    std::string mac;
+    std::string ssid;
+    bool backhaul_vap;
+} sVapElement;
```
tomereli, line 77: Missing line break

tomereli (changes requested):
https://github.com/prplfoundation/prplMesh/pull/608/commits/9548ade76570ae6f59310f60c1cde18ee51e2ba0 should be split to 2 - one commit which adds the message passing from son slave to the backhaul and just prints that the VAP configuration update received in the backhaul manager, and one which adds the actual map and updates it. But I guess it's a nitpick.

tomereli (approved):

agent/src/beerocks/slave/agent_ucc_listener.cpp
morantr, line 77: Why did you define `sVapElement`  and not using it?

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -64,6 +64,25 @@ void agent_ucc_listener::clear_configuration()
...
+void agent_ucc_listener::update_vaps_list(std::string ruid, beerocks_message::sVapsList &vaps)
+{
+    LOG(INFO) << "Update VAP map for ruid " << ruid << " bssid "
+              << network_utils::mac_to_string(vaps.vaps->mac) << " ssid "
+              << std::string(vaps.vaps->ssid, 36);
+    vaps_map[ruid] = vaps;
```
morantr, line 83: Besides to insert data into the container, there is no use for it.

morantr:
This PR is partial, I don't think it should have been merged.

PR #610, opened 2019-12-26
tools: docker: openwrt: increase the status test timeout

rmelotte:
When running the status test, we only wait for a specified time for
prplMesh to become operational. This time was set to 10 seconds, which
can be a little bit too small. As a consequence, the test would fail
from time to time, while re-running it would make it succeed.

Set the timeout to 30 seconds, to give prplMesh more time to restart.


rmelotte:
The timeout was not the main issue, this change will be part of another PR.

PR #611, opened 2019-12-26
Klocwork - solving the nullptr issue

CoralMalachi:
Resolving a klocwork issue which leads to many errors.

Since a null pointer may cause undefined behavior when the program attempts to
access memory through it, each time we tried to access to an object that
may be null without validates it is not, the klocwork created an issue
for it with the following message:

`Pointer *** returned from call to function *** may be NULL and will be dereferenced at ...`

To solve all of those issues, a validation check is added each time we try
to call a member function on an object that can be an invalid object (a
null pointer doesn't point to a valid object).

Signed-off-by: CoralMalachi <coral.malachi@intel.com>

### DIFF
--- a/agent/src/beerocks/monitor/monitor_stats.cpp
+++ b/agent/src/beerocks/monitor/monitor_stats.cpp
@@ -109,7 +109,7 @@ void monitor_stats::process()
     if (send_activity_mode_notification) {
         auto notification = message_com::create_vs_message<
             beerocks_message::cACTION_MONITOR_HOSTAP_ACTIVITY_NOTIFICATION>(cmdu_tx);
-        if (notification == nullptr) {
+        if (!notification) {
             LOG(ERROR) << "Failed building cACTION_MONITOR_HOSTAP_ACTIVITY_NOTIFICATION message!";
             return;
         }
@@ -122,7 +122,7 @@ void monitor_stats::process()
     if (!requests_list.empty()) {
         auto response = message_com::create_vs_message<
             beerocks_message::cACTION_MONITOR_HOSTAP_STATS_MEASUREMENT_RESPONSE>(cmdu_tx);
-        if (response == nullptr) {
+        if (!response) {
             LOG(ERROR)
                 << "Failed building cACTION_MONITOR_HOSTAP_STATS_MEASUREMENT_RESPONSE message!";
             return;
@@ -153,7 +153,7 @@ void monitor_stats::process()
         for (auto it = mon_db->sta_begin(); it != mon_db->sta_end(); ++it) {
             auto sta_mac  = it->first;
             auto sta_node = it->second;
-            if (sta_node == nullptr) {
+            if (!sta_node) {
                 continue;
             }
 
@@ -181,7 +181,12 @@ void monitor_stats::process()
             sta_stats_msg.rx_rssi           = sta_stats.rx_rssi_curr;
         }
 
-        beerocks_header->actionhdr()->id() = requests_list.front();
+        auto actionhdr = beerocks_header->actionhdr();
+        if (!actionhdr) {
+            return;
+        }
+
+        actionhdr->id() = requests_list.front();
         requests_list.pop_front();
         message_com::send_cmdu(slave_socket, cmdu_tx);
     }
@@ -209,8 +214,8 @@ void monitor_stats::process()
     for (auto it = mon_db->sta_begin(); it != mon_db->sta_end(); ++it) {
         auto sta_mac  = it->first;
         auto sta_node = it->second;
-        if (sta_node == nullptr) {
-            LOG(ERROR) << "sta_node == nullptr !";
+        if (!sta_node) {
+            LOG(ERROR) << "!sta_node !";
             return;
         }
 
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -292,7 +292,7 @@ void backhaul_manager::platform_notify_error(bpl::eErrorCode code, const std::st
         message_com::create_vs_message<beerocks_message::cACTION_PLATFORM_ERROR_NOTIFICATION>(
             cmdu_tx);
 
-    if (error == nullptr) {
+    if (!error) {
         LOG(ERROR) << "Failed building message!";
         return;
     }
@@ -363,7 +363,7 @@ bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
         auto notification = message_com::create_vs_message<
             beerocks_message::cACTION_BACKHAUL_CONNECTED_NOTIFICATION>(cmdu_tx);
 
-        if (notification == nullptr) {
+        if (!notification) {
             LOG(ERROR) << "Failed building message!";
             return false;
         }
@@ -508,7 +508,7 @@ bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
                            << " skipping " << sc->hostap_iface;
                 continue;
             }
-            if (sc->slave == nullptr) {
+            if (!sc->slave) {
                 LOG(ERROR) << "slave " << sc->hostap_iface << " socket is nullptr!";
                 continue;
             }
@@ -539,7 +539,7 @@ bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
                 continue;
             }
 
-            if (sc->slave == nullptr) {
+            if (!sc->slave) {
                 continue;
             }
 
@@ -547,7 +547,7 @@ bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
 
             auto notification = message_com::create_vs_message<
                 beerocks_message::cACTION_BACKHAUL_DISCONNECTED_NOTIFICATION>(cmdu_tx);
-            if (notification == nullptr) {
+            if (!notification) {
                 LOG(ERROR) << "Failed building message!";
                 return false;
             }
@@ -1487,7 +1487,7 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
         } else if (pending_enable) {
             auto notification = message_com::create_vs_message<
                 beerocks_message::cACTION_BACKHAUL_BUSY_NOTIFICATION>(cmdu_tx);
-            if (notification == nullptr) {
+            if (!notification) {
                 LOG(ERROR) << "Failed building cACTION_BACKHAUL_BUSY_NOTIFICATION message!";
                 break;
             }
@@ -1602,7 +1602,7 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
             auto response = message_com::create_vs_message<
                 beerocks_message::cACTION_BACKHAUL_CLIENT_RX_RSSI_MEASUREMENT_CMD_RESPONSE>(
                 cmdu_tx, beerocks_header->id());
-            if (response == nullptr) {
+            if (!response) {
                 LOG(ERROR) << "Failed building "
                               "ACTION_BACKHAUL_CLIENT_RX_RSSI_MEASUREMENT_CMD_RESPONSE message!";
                 break;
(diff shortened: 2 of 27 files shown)


CoralMalachi:
Please do not merge yet.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -181,7 +181,12 @@ void monitor_stats::process()
             sta_stats_msg.rx_rssi           = sta_stats.rx_rssi_curr;
         }
 
-        beerocks_header->actionhdr()->id() = requests_list.front();
+        auto actionhdr = beerocks_header->actionhdr();
+        if (!actionhdr) {
+            return;
+        }
+
+        actionhdr->id() = requests_list.front();
```
arnout, line 189: The rest of the commit is an almost mechanical change `foo == nullptr` -> `!foo`. Therefore it would be better to do this bit in a separate commit.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1874,6 +1874,11 @@ bool backhaul_manager::send_slaves_enable()
             message_com::create_vs_message<beerocks_message::cACTION_BACKHAUL_ENABLE_APS_REQUEST>(
                 cmdu_tx);
 
+        if (!notification) {
+            LOG(ERROR) << "create_vs_message method failed";
+            return true;
+        }
+
```
arnout, line 1881: Same here, better as a separate commit.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -1312,7 +1312,7 @@ bool main_thread::handle_arp_monitor()
     sd = get_slave_socket_from_hostap_iface_name(strIfaceName);
 
     // Use the Backhaul Manager Slave as the default destination
-    if ((sd == nullptr) && ((sd = get_backhaul_socket()) == nullptr)) {
+    if ((!sd) && ((sd = get_backhaul_socket()))) {
```
arnout, line 1315: You changed the semantics here:

```suggestion
    if ((!sd) && (!(sd = get_backhaul_socket()))) {
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1562,15 +1562,20 @@ bool slave_thread::handle_cmdu_platform_manager_message(
             std::string client_mac = network_utils::mac_to_string(notification->mac());
             std::string client_ip  = network_utils::ipv4_to_string(notification->ipv4());
 
+            auto hostname = notification->hostname(message::NODE_NAME_LENGTH);
+            if (!hostname) {
+                LOG(ERROR) << "hostname() failed";
+                return false;
+            }
```
arnout, line 1569: Again, better in a separate commit.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
```diff
@@ -39,10 +39,31 @@ class beerocks_header : public ClassList {
...
     beerocks_message::eAction action()
     {
-        return actionhdr() ? actionhdr()->action() : beerocks_message::ACTION_NONE;
+        auto action_hdr = actionhdr();
+        if (!action_hdr) {
+            return beerocks_message::ACTION_NONE;
+        } else {
+            return action_hdr->action();
+        }
+    }
```
arnout, line 48: This is exactly the same as the original... Does klocwork really barf on `?:` expressions? Or isn't it be able to recognise that `actionhdr()` always returns the same value? Hm, probably the latter... Would it help if we would add `__attribute__((pure))` to the actionhdr declaration?

Oh, and we probably need to make all those functions const as well. But that's unrelated.

controller/src/beerocks/master/db/network_map.cpp
```diff
@@ -109,20 +115,41 @@ void network_map::send_bml_network_map_message(db &database, Socket *sd,
...
+
+                auto actionhdr = beerocks_header->actionhdr();
+                if (!actionhdr) {
+                    LOG(ERROR) << "actionhdr() failed ";
+                    return;
+                }
+
+                actionhdr->last() = 0;
+                num_of_nodes      = response->node_num(); // prepare for next message
+                num_of_nodes      = 0;
```
arnout, line 137: I realize that you didn't change this code, but something is fishy here - the first assignment is redundant...

controller/src/beerocks/master/db/network_map.cpp
```diff
@@ -109,20 +115,41 @@ void network_map::send_bml_network_map_message(db &database, Socket *sd,
...
                 }
 
-                if (data_start == nullptr) {
-                    data_start = (uint8_t *)response->buffer(0);
+                if (!data_start) {
+                    auto buf = response->buffer(0);
+                    if (!buf) {
+                        LOG(ERROR) << "buffer is nullptr";
+                        return;
+                    }
```
arnout, line 151: Does klocwork really require this? To me, it would be more natural to do:

```
    data_start = (uint8_t *)response->buffer(0);
    if (!data_start) {
```

controller/src/beerocks/master/db/network_map.cpp
```diff
@@ -311,17 +343,24 @@ void network_map::send_bml_nodes_statistics_message_to_listeners(
...
-    if (response == nullptr) {
+    if (!response) {
         LOG(ERROR) << "Failed building ACTION_BML_STATS_UPDATE message!";
         return;
     }
 
-    auto beerocks_header = message_com::get_beerocks_header(cmdu_tx);
-    if (!beerocks_header) {
+    auto beerocks_hdr = message_com::get_beerocks_header(cmdu_tx);
+    if (!beerocks_hdr) {
```
arnout, line 352: Why?

controller/src/beerocks/master/db/network_map.cpp
```diff
@@ -373,12 +417,26 @@ void network_map::send_bml_nodes_statistics_message_to_listeners(
...
+            auto beerocks_hdr = message_com::get_beerocks_header(cmdu_tx);
+            if (!beerocks_hdr) {
+                LOG(ERROR) << "get_beerocks_header() has failed";
+                return -1;
+            }
+            auto actionhdr = beerocks_hdr->actionhdr();
+            if (!actionhdr) {
+                LOG(ERROR) << "actionhdr() has failed";
+                return -1;
+            }
```
arnout, line 434: Hm, this pattern is repeating a couple of times. And actually it is a bit silly: if create_vs_message doesn't return nullptr, then we're guaranteed to have a beerocks_hrd and an actionhdr...

Maybe we should just extend create_vs_message with an optional std::shared_ptr<beerocks_message::cACTION_HEADER> argument that gets updated with the action header for the cases that need it?

controller/src/beerocks/master/db/network_map.cpp
```diff
@@ -476,15 +539,21 @@ void network_map::send_bml_bss_tm_req_message_to_listeners(db &database,
...
+    auto event = reinterpret_cast<BML_EVENT *>(buf);
+    if (!event) {
         LOG(ERROR) << "event is nullptr";
         return;
     }
 
     event->type = BML_EVENT_TYPE_BSS_TM_REQ;
     auto size   = sizeof(BML_EVENT);
-    event->data = GET_MESSAGE_POINTER(BML_EVENT_BSS_TM_REQ, response->buffer(0), size);
+    event->data = GET_MESSAGE_POINTER(BML_EVENT_BSS_TM_REQ, buf, size);
```
arnout, line 556: Meh. This stuff was ugly already, and it doesn't get better after the change... Why aren't bml events generated like the rest of the code? @morantr @vitalybu 

At least, if it is not generated, it would be better to add a helper template function that does all of the parsing and checking. And get rid of the GET_MESSAGE_POINTER macro.

arnout (changes requested):
Other than the bug in `main_thread::handle_arp_monitor` it looks good to me. However, because there's so much squashed into a single commit, I can't be sure that I didn't miss anything. Please split the commits.

controller/src/beerocks/master/db/network_map.cpp
CoralMalachi: Yes, I saw that too, @morantr can I remove line 136?

tomereli: For something like this you don't need approval from Moran.. the first assignment is simply redundant so just remove it

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
CoralMalachi: Yes, klocwork complainants whenever we try to access the return value of any method that returns a shared pointer unless we validate that it is not null

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
CoralMalachi: nice catch

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
tomereli: But ` = actionhdr() ? ACTION_NONE : actionhd->action();` is checking exactly the same thing. Weird.

arnout:
One more thing: could you add a commit that updates CODINGSTYLE and explains not to use `== nullptr`?

CoralMalachi:
> One more thing: could you add a commit that updates CODINGSTYLE and explains not to use `== nullptr`?

Sure.

common/beerocks/tlvf/src/include/beerocks/tlvf/beerocks_header.h
arnout: No, because there is no guarantee that actionhdr() will return the same value in both cases. That's why I suggested to declare it as pure - if klocwork knows the pure attribute, it can use it to determine that in fact the same value will be returned in both cases.

framework/tlvf/src/src/CmduMessage.cpp
```diff
@@ -14,7 +14,12 @@ int CmduMessage::getNextTlvType() const
 {
     if (!getCmduHeader())
         return -1;
-    sTlvHeader *tlv = reinterpret_cast<sTlvHeader *>(msg.prevClass()->getBuffPtr());
+
+    auto prev_class = msg.prevClass();
+    if (!prev_class) {
+        return -1;
+    }
```
arnout, line 21: This is a bit annoying. If getCmduHeader() doesn't return nullptr, then prevClass() is guaranteed to return non-null because the cmdu header is the first class in the list. So this additional check is entirely redundant, and it's just a matter of insufficient intelligence in the tool (i.e., a false positive).

In this case, we don't actually make use of the getCmduHeader result, so we can get away with just removing the first check.

controller/src/beerocks/master/db/network_map.cpp
CoralMalachi: I'll do it in next commit, great idea

controller/src/beerocks/master/db/network_map.cpp
morantr: Actually it is a serious bug now that I look at it. `num_of_nodes` is defined as reference of `response->node_num()` at the function beginning. On that block, we create a new message but we still use the reference to the previous message. It is accidentally working since the `num_of_nodes` location on the buffer is not changed when creating a new message.

I think that regardless of deleting the redundant line, we need also to use `response->node_num()` directly instead of the reference, and delete the reference declaration at the function beginning.

controller/src/beerocks/master/db/network_map.cpp
```diff
@@ -144,8 +176,8 @@ std::ptrdiff_t network_map::fill_bml_node_data(db &database, std::string node_ma
                                                bool force_client_disconnect)
 {
     auto n = database.get_node(node_mac);
-    if (n == nullptr) {
-        LOG(ERROR) << "get_node(), node_mac=" << node_mac << " , n == nullptr !!!";
+    if (!n) {
+        LOG(ERROR) << "get_node(), node_mac=" << node_mac << " , !n !!!";
```
morantr, line 180: I don't think it was right to replace it also on debug prints.

morantr (changes requested):

PR #613, opened 2019-12-26
Enhance logging configuration

vitalybu:
This PR adds several improvements to the existing logging configuration:
- Allow controlling files/stdout/syslog output from the configuration file
- Allow controlling the internal log files rolling mechanism
- Enable easylogging++ syslog support (ELPP_SYSLOG)

NOTE: Some commits in this PR are not independent, but were separated to ease the review process.

### DIFF
--- a/cmake/Findelpp.cmake
+++ b/cmake/Findelpp.cmake
@@ -2,10 +2,11 @@ set(ELPP_LIB_NAME "mapf::elpp")
 set(ELPP_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/framework/external/easylogging)
 set(ELPP_LIBRARY ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libelpp.so.1.4.0)
 
-add_definitions(-DELPP_THREAD_SAFE)
-add_definitions(-DELPP_FORCE_USE_STD_THREAD)
-add_definitions(-DELPP_NO_DEFAULT_LOG_FILE)
 add_library(${ELPP_LIB_NAME} UNKNOWN IMPORTED)
+
+set_target_properties(${ELPP_LIB_NAME} PROPERTIES
+    INTERFACE_COMPILE_DEFINITIONS "ELPP_SYSLOG;ELPP_THREAD_SAFE;ELPP_NO_DEFAULT_LOG_FILE;ELPP_FORCE_USE_STD_THREAD"
+)
 
 # Include directory
 set_target_properties(${ELPP_LIB_NAME} PROPERTIES
--- a/common/beerocks/bcl/include/bcl/beerocks_config_file.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_config_file.h
@@ -20,13 +20,14 @@ public:
 
     struct SConfigLog {
         //[log]
-        std::string path;
         std::string global_levels;
         std::string syslog_levels;
         std::string global_size;
+        std::string files_enabled;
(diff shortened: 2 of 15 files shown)


cmake/Findelpp.cmake
```diff
@@ -2,11 +2,12 @@ set(ELPP_LIB_NAME "mapf::elpp")
 set(ELPP_INCLUDE_DIRS ${CMAKE_SOURCE_DIR}/framework/external/easylogging)
 set(ELPP_LIBRARY ${CMAKE_LIBRARY_OUTPUT_DIRECTORY}/libelpp.so.1.4.0)
 
-add_definitions(-DELPP_THREAD_SAFE)
-add_definitions(-DELPP_FORCE_USE_STD_THREAD)
-add_definitions(-DELPP_NO_DEFAULT_LOG_FILE)
 add_library(${ELPP_LIB_NAME} UNKNOWN IMPORTED)
 
+set_target_properties(${ELPP_LIB_NAME} PROPERTIES
```
tomereli, line 7: This should be 2 separate commits - one which unifies the `add_definitions` into a single `set_target_properties`, and one commit which adds the `ELPP_SYSLOG`

common/beerocks/bcl/source/beerocks_logging.cpp
```diff
@@ -440,10 +448,11 @@ void logging::handle_logging_rollover(const char *log_name, std::size_t)
     remove(rollover_name.c_str());
     rename(log_name, rollover_name.c_str());
 }
+
 void logging::apply_settings()
 {
     // Disable The instance of RollMonitor to start fresh
-    {
+    if (m_log_files_auto_roll) {
```
tomereli, line 455: Should probably be a separate commit which adds the auto-roll option

common/beerocks/bcl/include/bcl/beerocks_config_file.h
```diff
@@ -20,13 +20,14 @@ class config_file {
...
-        std::string path;
         std::string global_levels;
         std::string syslog_levels;
         std::string global_size;
+        std::string files_enabled;
+        std::string files_path;
+        std::string files_auto_roll;
+        std::string stdout_enabled;
         std::string syslog_enabled;
-        std::string netlog_host;
```
tomereli, line 28: Removing the netlog logging should probably have been done in a separate commit.

tomereli (approved):
commit https://github.com/prplfoundation/prplMesh/pull/613/commits/c7f170a664bbe7d162ba60f3362953c9c8ecb479 should have been split to 3 commits:
1. cmake: add log files autoroll option
2. cmake: add stdout logging option
3. cmake: add syslog logging option

Then the following commit should be split also to 3 which adds the actual support for each feature. This will make this PR much easier to review and also easier to bisect issues in the future if such arise.

Other than that, LGTM, so already approving.

cmake/Findelpp.cmake
vitalybu: You're probably right, but since it doesn't change the previous behavior, I'll keep it as it

common/beerocks/bcl/source/beerocks_logging.cpp
vitalybu: Could've split the changes into more atomic commits, but since most of the changes are concentrated in just a couple of files (mostly `beerocks_logging.cpp`), I thought it'll be easier to review by seeing the bigger picture.
Will take that into account for the next PRs.

PR #614, opened 2019-12-26
Feature/automatic parsing

### LINKED ISSUE #318: [POC] Automatically parsing CMDUs into a collection of TLVs (blindly parsing attributes)

### LINKED ISSUE #609: [TASK] TLVF: system wide use of CmduTlvParser
## Description

Now that both vendor-specific and WSC messages provide automatic parsers as introduced in #597 and #471, we can switch to using the `CmduTlvParser` when calling `cmdu_rx.parse()`.
This means that *all* TLVs will be parsed completely by the TLVF (soon to be) autogenerated `CmduTlvParser`.

Vendor specific TLV parsing will be partial - we will parse the outer VS TLV but the parsing of the beerocks header will be done by the caller since TLVF is not aware of each vendor implementation details.

WSC TLV parsing will be complete and will call `AttrList::parse(*tlv);` which will parse all the internal attributes and store the m1 or m2 in the TLV inner class list pointer.

This will remove the need for using while loops before calling `cmdu_rx.addClass` in the callers and replace that with a simple `cmdu_rx.getClass`.

tomereli:
This PR completes the automatic TLV parsing feature by enabling `cmdu_rx.parse()` to *always* parse the entire CMDU. As a result, the automatic parser `CmduTlvParser` is blindely parsing the entire CMDU, including WSC internal Attribute Lists for M1 and M2 messages.
Vendor specific inner classes are *not* automatically parsed since they may differ between different vendors.
 
Fixes #609 
Fixes #318


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -311,116 +311,74 @@ bool master_thread::handle_cmdu_1905_1_message(const std::string &src_mac,
...
-
-    if (tlvAlMacAddressType) {
-        al_mac =
-            network_utils::mac_to_string((const unsigned char *)tlvAlMacAddressType->mac().oct);
-        LOG(DEBUG) << "mac=" << al_mac;
-    } else {
-        LOG(ERROR) << "tlvAlMacAddressType missing - ignoring autconfig search message";
+    auto tlvAlMacAddressType = cmdu_rx.getClass<ieee1905_1::tlvAlMacAddressType>();
+    if (!tlvAlMacAddressType) {
+        LOG(ERROR) << "getClass<tlvAlMacAddressType> failed";
```
vitalybu, line 318: Really nitpicking here, but one can argue that this is not really an error.
I believe that errors should be reported when something went so wrong that regular operation can't be continued.
In this case, it's probably a malformed CMDU or a missing TLV, which doesn't really impact the Controller's operation, so a WARNING would be more appropriate here.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -311,116 +311,74 @@ bool master_thread::handle_cmdu_1905_1_message(const std::string &src_mac,
...
-        LOG(DEBUG) << "band=" << int(auto_config_freq_band);
-    } else {
-        LOG(ERROR) << "tlvAutoconfigFreqBand missing - ignoring autconfig search message";
+    auto al_mac =
+        network_utils::mac_to_string((const unsigned char *)tlvAlMacAddressType->mac().oct);
+    LOG(DEBUG) << "mac=" << al_mac;
+
+    LOG(DEBUG) << "searched_role=" << int(tlvSearchedRole->value());
+    if (tlvSearchedRole->value() != ieee1905_1::tlvSearchedRole::REGISTRAR) {
+        LOG(ERROR) << "invalid tlvSearchedRole value";
```
vitalybu, line 348: Continuing my previous comment, this is probably a warning as well.
Also, a good idea would be to print the received value... Will help debugging.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3848,77 +3848,47 @@ bool slave_thread::handle_autoconfiguration_renew(Socket *sd, ieee1905_1::CmduMe
...
+    // if an AP-Autoconfiguration WSC message is sent by the Multi-AP Agent,
+    // it includes one AP Radio basic capabilities TLV and one WSC TLV (containing M1).
+    // If the the message is sent by the Multi-AP controller, it includes one
+    // AP radio Identifier TLV and one or more WSC TLVs (containing M2).
+    // Check if this is the case and this is an M1 message we sent to the controller
+    // which was just looped back (when we send the M1 to the local controller,
+    // it will be published back on the local bus because the destination is our
+    // AL-MAC, and the agent does listen to this CMDU message type).
+    // If this is a looped back M1 CMDU, we can treat is as handled successfully.
+    if (cmdu_rx.getClass<wfa_map::tlvApRadioBasicCapabilities>())
```
vitalybu: Shouldn't we check that it's indeed our looped request?
What about non-local agents (repeaters) that send an autoconfig message to our controller?
The local agent won't receive the messages?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3862,77 +3862,47 @@ bool slave_thread::handle_autoconfiguration_renew(Socket *sd, ieee1905_1::CmduMe
...
+        if (tlvWsc) {
+            auto m2 = std::dynamic_pointer_cast<WSC::m2>(tlvWsc->getInnerClassList());
+            if (m2)
+                m2_list.push_back(*m2);
+        }
+    }
     if (m2_list.empty()) {
         LOG(ERROR) << "No M2s present";
         return false;
     }
```
Galvien: In my opinion, instead of iterating over the cmdu_rx with `getClassCount<T>()`  we should implement a function named something like `getAllClasses<T>()` (`getClassList<T>()` would be too confusing) that returns a vector of all tlvs of type T in the cmdu.
- Firstly, it would be much more befitting of the OOP oriented direction of this pull request and it's predecessors.
- Secondly, once implemented, we could iterate over the vector with a foreach loop, making boilerplates such as this one shorter and more readable.

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -19,18 +19,13 @@ CmduMessageRx::CmduMessageRx(uint8_t *buff, size_t buff_len) : CmduMessage(buff,
...
 
-bool CmduMessageRx::parse(bool parse_tlvs)
+bool CmduMessageRx::parse()
 {
     msg.reset(true);
     auto cmduhdr = msg.addClass<cCmduHeader>();
     if (!cmduhdr)
         return false;
 
-    cmduhdr->class_swap(); //swap back cmduheader
```
vitalybu, line 29: Was is swapped twice? How did it work before?

Galvien (commented):
1. This PR is a continuation of a few previous and more importantly - _recently merged_ PRs, I think it would be nice if it was mentioned in the PR comment, and even have references to relevant commits (especially cool if it's a github comment, as github provides the direct link to the commit page) when relevant.
2. When performing a tree-wide change, such as in febf29c, please provide references to the changes worth taking a look at. The changes replacing all of the `addClass<T>()` calls with `getClass<T>()` are pretty ignorable, but there are some more important, logic changes that should be noted.
Some suggested methods of highlighting such cases:
    - Split the commit into a 'boilerplate changing' commit and a 'meaty' commit.
    - In the commit message, add a suffix along the lines of "note line #1234"
    - On github, add comments on code parts worth checking out, and mention it in the commit message.

framework/tlvf/src/src/CmduTlvParser.cpp
```diff
@@ -108,7 +108,7 @@ std::shared_ptr<BaseClass> CmduTlvParser::parseNextTlv()
     case (17): {
         auto tlv = cmdu_.addClass<ieee1905_1::tlvWsc>();
         if (tlv)
-            WSC::AttrList::parse(*tlv);
+            tlv->addInnerClassList(WSC::AttrList::parse(*tlv));
         return tlv;
     }
```
Galvien: CmduTlvParser shouldn't handle special cases such as this one, the attribute list parsing should be done as part of `tlvWsc::addClass()`, this requires a TLVF change, as tlvWsc is an autogenerated class, and I will create an issue for that.
But still worth mentioning.

Galvien: CmduTlvParser shouldn't handle special cases such as this one, the attribute list parsing should be done as part of `tlvWsc::addClass()`, this requires a TLVF change, as tlvWsc is an autogenerated class.
It's still worth mentioning in my opinion, and I will create an issue for that.

tomereli:
[TLVF Automatic parsing full support](https://trello.com/c/y3wISsqB/22-tlvf-automatic-parsing-full-support)

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: It doesn't matter if its our looped request or an M1 CMDU from another agent, in both cases we are dropping the message.

framework/tlvf/src/src/CmduMessageRx.cpp
tomereli: Updated commit message with full explanation.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: I disagree - I think that errors are anything not expected which changes the way the system behaves. In this case, an agent tried to onboard our controller and we simply didn't answer it due to some missing TLV - we should easilly find this in the logs while looking for either error or warning, so changing to warning is OK, but not as part of this PR since it was ERROR before this change.

IMHO FATAL is more appropriate than error to describe an unrecoverable error, not ERROR.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Same anwer - perhaps you're right, but I didn't change the error logging as part of this PR and I don't want to do that either.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: We do get the other agents WSC messages and the handling is and should be the same as looped back M1 CMDUs - we should ignore them. Checking that this is indeed our looped request is a much harder thing to do with ZMQ which is our current MSGLIB backend. Perhaps it will become easier once we switch to a different implementation, but for now, this is the only way we can do that.

framework/tlvf/src/src/CmduTlvParser.cpp
tomereli: I generally agree, TLVF should add meta option to describe which inner class list should be part of each TLV and then we can autogenerate the parser to call the `addInnerClassList` but Its an overkill IMO at this stage.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Good idea - done.

framework/tlvf/src/src/CmduMessageRx.cpp
arnout: The improved commit message is excellent!

framework/tlvf/src/include/tlvf/ClassList.h
```diff
@@ -105,6 +106,23 @@ class ClassList {
         return count;
     }
 
+    /**
+     * @brief Get the list of classes of type T
+     *
+     * @tparam T class template
+     * @return std::list<std::shared_ptr<T>> list of class T pointers
+     */
+    template <class T> std::list<std::shared_ptr<T>> getClassList() const
```
arnout, line 93: I know that @vitalybu suggested to use a std::list instead of a std::vector because it's more efficient, but he's wrong according to [CppCoreGuildelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#slcon2-prefer-using-stl-vector-by-default-unless-you-have-a-reason-to-use-a-different-container). std::vector has

- the fastest general-purpose access (random access, including being vectorization-friendly);
- the fastest default access pattern (begin-to-end or end-to-begin is prefetcher-friendly);
- the lowest space overhead (contiguous layout has zero per-element overhead, which is cache-friendly).

Specifically in this case, where we expect the result set to contain 0, 1 or 2 elements and very rarely more than that, a vector is more efficient on all counts.

Obviously, though, since the result is usually small, the impact of this choice is not significant. So nitpick.

common/beerocks/tlvf/src/src/beerocks_message.cpp
```diff
@@ -61,7 +61,7 @@ message_com::parse_intel_vs_message(ieee1905_1::CmduMessageRx &cmdu_rx)
     if (!cmdu_rx.getMessageLength()) {
         LOG(ERROR) << "cmdu is not initialized!";
     }
-    auto tlv = cmdu_rx.addClass<ieee1905_1::tlvVendorSpecific>();
+    auto tlv = cmdu_rx.getClass<ieee1905_1::tlvVendorSpecific>();
```
arnout, line 64: It is weird that the commit message says that addClass will be converted to getClass in later commits, but then this one change is done in the same commit.

Also, why is the parse call on line 87 not converted in the same commit?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -311,116 +311,74 @@ bool master_thread::handle_cmdu_1905_1_message(const std::string &src_mac,
...
-    }
-
-    if (tlvAlMacAddressType) {
-        al_mac =
-            network_utils::mac_to_string((const unsigned char *)tlvAlMacAddressType->mac().oct);
-        LOG(DEBUG) << "mac=" << al_mac;
-    } else {
-        LOG(ERROR) << "tlvAlMacAddressType missing - ignoring autconfig search message";
+    auto tlvAlMacAddressType = cmdu_rx.getClass<ieee1905_1::tlvAlMacAddressType>();
+    if (!tlvAlMacAddressType) {
```
arnout, line 317: The removal of the loop and moving the checks immediately after the declaration is a bit more than what is strictly needed for "automatic parsing - fix autoconfig search".

In fact, I would have split the commits completely differently: first commit that just changes the default value of the parse_tlvs parameter from false to true, and in the same commit replace all cmdu_rx.addClass with cmdu_rx.getClass. That's an almost purely mechanical change which probably doesn't break anything. Then subsequent commits can de improvements and cleanups, like this one.

Anyway, very much nitpick.

**EDIT**

D'oh, I was stupid. Just replacing addClass with getClass *doesn't* fix things, because the while loop no longer works correctly (after parse, getNextTlvType will always return EOM). So never mind.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3862,77 +3862,44 @@ bool slave_thread::handle_autoconfiguration_renew(Socket *sd, ieee1905_1::CmduMe
...
                    << " != " << network_utils::mac_to_string(ruid->radio_uid());
         return true;
     }
 
+    std::list<WSC::m2> m2_list;
+    for (auto tlv : cmdu_rx.msg.getClassList<ieee1905_1::tlvWsc>()) {
+        auto m2 = std::dynamic_pointer_cast<WSC::m2>(tlv->getInnerClassList());
+        if (m2)
+            m2_list.push_back(*m2);
+    }
```
arnout: I don't like this very much. I would much prefer if we would do the recursive parsing directly here instead of in `CmduTlvParser::parse`. So:

```
std::vector<WSC::m2> m2_list;
for (auto tlv : cmd_rx.msg.getClassList<ieee1905_1::tlvWsc>()) {
    auto m2 = WSC::m2::parse(tlv);
    if (!m2) {
        LOG(INFO) << "Not a valid M2 - ignoring WSC CMDU";
        return false;
    }
    m2_list.push_back(m2);
}
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3862,77 +3862,44 @@ bool slave_thread::handle_autoconfiguration_renew(Socket *sd, ieee1905_1::CmduMe
...
         LOG(DEBUG) << "not to me - ruid " << config.radio_identifier
                    << " != " << network_utils::mac_to_string(ruid->radio_uid());
         return true;
     }
 
+    std::list<WSC::m2> m2_list;
+    for (auto tlv : cmdu_rx.msg.getClassList<ieee1905_1::tlvWsc>()) {
+        auto m2 = std::dynamic_pointer_cast<WSC::m2>(tlv->getInnerClassList());
+        if (m2)
+            m2_list.push_back(*m2);
```
arnout: I think we need to return false if m2 is not valid, like we did before.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3988,9 +3955,10 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
         return false;
     }
 
-    if (intel_agent) {
+    auto beerocks_header = message_com::parse_intel_vs_message(cmdu_rx);
+    if (beerocks_header) {
         LOG(INFO) << "Intel controller join response";
-        if (!parse_intel_join_response(sd, cmdu_rx)) {
+        if (!parse_intel_join_response(sd, *beerocks_header)) {
```
arnout, line 3946: It would have been nice to have done this change in a separate, preparative commit (removing intel_agent and relying on `if (beerocks_header)` is something that can be done independent of the change to `parse(true)`).

nitpick, of course.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3862,77 +3862,44 @@ bool slave_thread::handle_autoconfiguration_renew(Socket *sd, ieee1905_1::CmduMe
...
+    // According to section 17.1.3 in the Multi-AP Specification version 1.0,
+    // if an AP-Autoconfiguration WSC message is sent by the Multi-AP Agent,
+    // it includes one AP Radio basic capabilities TLV and one WSC TLV (containing M1).
+    // If the the message is sent by the Multi-AP controller, it includes one
+    // AP radio Identifier TLV and one or more WSC TLVs (containing M2).
+    // Check if this is the case and this is an M1 message we sent to the controller
+    // which was just looped back (when we send the M1 to the local controller,
+    // it will be published back on the local bus because the destination is our
+    // AL-MAC, and the agent does listen to this CMDU message type).
+    // If this is a looped back M1 CMDU, we can treat is as handled successfully.
```
arnout: Actually, we could just drop this special handling entirely now. The return value is anyway ignored (it gets passed up to socket_thread::work without further comment, and there it essentially gets ignored). We really should be checking for M1/M2, which we do implicitly by checking m2_list.empty().

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -703,60 +703,28 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(const std::string &sr
...
+    // it includes one AP Radio basic capabilities TLV and one WSC TLV (containing M1).
+    // If the the message is sent by the Multi-AP controller, it includes one
+    // AP radio Identifier TLV and one or more WSC TLVs (containing M2).
+    // Check if this is the case and this is an M2 message we sent to the agent
+    // which was just looped back (when we send the M2 to the local agent,
+    // it will be published back on the local bus because the destination is our
+    // AL-MAC, and the controller does listen to this CMDU message type).
+    // If this is a looped back M2 CMDU, we can treat is as handled successfully.
+    if (cmdu_rx.getClass<wfa_map::tlvApRadioIdentifier>())
+        return true;
```
arnout: Same here, I think we don't need this special handling and we can just rely on the m1 check.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -703,60 +703,28 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(const std::string &sr
...
-
+    auto radio_basic_caps = cmdu_rx.getClass<wfa_map::tlvApRadioBasicCapabilities>();
+    auto tlvWsc           = cmdu_rx.getClass<ieee1905_1::tlvWsc>();
     if (!tlvWsc) {
-        LOG(ERROR) << "Failed to get TLV_WSC TLV";
+        LOG(ERROR) << "getClass<ieee1905_1::tlvWsc> failed";
         return false;
     }
-    auto m1 = std::dynamic_pointer_cast<WSC::m1>(WSC::AttrList::parse(*tlvWsc));
+    auto m1 = std::dynamic_pointer_cast<WSC::m1>(tlvWsc->getInnerClassList());
```
arnout: Again, it would make more sense to me do do the inner parsing here instead of doing a down-cast.

arnout (commented):
Reviewed up to "autoparse: fix WSC autoconfig". Submitting the review already so you can start processing the feedback. Up to now, I would approve it (all changes are either nitpicks or possibly to be fixed later).

framework/tlvf/src/src/ClassList.cpp
```diff
@@ -34,6 +34,8 @@ void ClassList::swap()
     // call all tlv finalize functions
     for (auto const &c : m_class_vector) {
         c->class_swap();
+        if (c->getInnerClassList())
+            c->getInnerClassList()->swap();
```
arnout, line 38: There is also an explicit swap at the end of `message_com::print_cmdu_types`. I'm not even sure if it is appropriate there, but it should probably be mentioned in the commit message.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1749,7 +1749,7 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
     const auto mid = cmdu_rx.getMessageId();
     LOG(DEBUG) << "Received HIGHER_LAYER_DATA_MESSAGE , mid=" << std::hex << int(mid);
 
-    auto tlvHigherLayerData = cmdu_rx.addClass<wfa_map::tlvHigherLayerData>();
+    auto tlvHigherLayerData = cmdu_rx.getClass<wfa_map::tlvHigherLayerData>();
     if (!tlvHigherLayerData) {
         LOG(ERROR) << "addClass wfa_map::tlvHigherLayerData failed";
```
arnout, line 1754: There are still quite a lot of log messages that mention addClass instead of getClass. nitpick.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1553,51 +1474,22 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
-            auto beerocks_header = message_com::parse_intel_vs_message(cmdu_rx);
-            if (!beerocks_header) {
-                LOG(DEBUG) << "Failed to parse intel vs message (not Intel?)";
-                continue;
-            }
+    auto client_association_event_tlv = cmdu_rx.getClass<wfa_map::tlvClientAssociationEvent>();
+    if (!client_association_event_tlv) {
+        LOG(ERROR) << "addClass wfa_map::tlvClientAssociationEvent failed";
+        return false;
+    }
```
arnout, line 1474: This is not correct. The original code had the following, which can just stay:

```
   if (!client_association_event_tlv) {
        // The standard allows having zero TLV_CLIENT_ASSOCIATION_EVENT, in that case, there is no
        // point to continue the handling of this message.
        LOG(INFO) << "wfa_map::tlvClientAssociationEvent not found";
        return true;
    }
```

framework/tlvf/src/include/tlvf/BaseClass.h
```diff
@@ -36,6 +36,14 @@ class BaseClass {
     bool is_finalized() { return m_finalized__; };
     void addInnerClassList(std::shared_ptr<ClassList> list) { m_inner__ = list; };
     std::shared_ptr<ClassList> getInnerClassList() { return m_inner__; };
+    template <class T> std::shared_ptr<T> class_cast()
```
arnout, line 49: Please add documentation for new functions. In this case, it should be in framework/tlvf/README.md because that's where the rest of the documentation is (though personally I'd prefer to have everything as doxygen).

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -7,28 +7,176 @@
...
+#include <tlvf/wfa_map/tlvSteeringRequest.h>
+#include <tlvf/wfa_map/tlvSupportedService.h>
+#include <tlvf/wfa_map/tlvTransmitPowerLimit.h>
+
+#include <iostream>
 
 using namespace ieee1905_1;
 
-CmduMessageRx::CmduMessageRx(uint8_t *buff, size_t buff_len) : CmduMessage(buff, buff_len)
+std::shared_ptr<BaseClass> CmduMessageRx::parseNextTlv()
```
arnout, line 52: This change goes a bit against the automatically generated parser that @Galvien still has lined up AFAIU. A step in the right direction would be to make this an external function that takes the ClassList as an argument and that open-codes getNextTlvType, because that's what the generated code would look like.

Nitpick.

arnout (approved):
One important change that really needs to be fixed before the merge: client_association_event_tlv

Also, the commit message of 'tlvf: test: fix test_all() ' has a bunch of spelling mistakes, it would be nice to correct those.

Since all changes can be handled without re-review, I'm already approving, but mark as donotmerge.

framework/tlvf/src/include/tlvf/ClassList.h
tomereli: I think that we can stick with class list since it is in fact a class list, and since the impact is not significant.

common/beerocks/tlvf/src/src/beerocks_message.cpp
tomereli: You're right, this is the first time we change addClass to getClass so that the initial message passing of vendor speficic CMDUs will work. I will add this in a separate commit which will come right after this one.
Regarding line 87 - will fix this also.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Splitting now is too much extra effort but I agree, will take it into account next time.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Correct, removed.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Done.

framework/tlvf/src/src/ClassList.cpp
tomereli: Actually - now that parse() does not automatically parse the inner class list, it isn't swapped, so this commit can be removed altogether - the backhaul manager parse OR the notorious print_cmdu_types method will simply not parse the inner list. I'm debating weather we should keep this here for the future in case we do the parsing then decide to swap back for forwarding. What do you think?

In any case I updated the commit message for now cause it was wrong.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: Arr yep.. We can fix that in a mechanical change I guess. Anyway, different PR.

framework/tlvf/src/include/tlvf/BaseClass.h
tomereli: So lets start with updating the header only, we can remove the documentation later on from the README.md since it is obsolete. @CoralMalachi can you do that as part of the update to the README.md file?

framework/tlvf/src/src/CmduMessageRx.cpp
tomereli: I don't think we need an automatically generated parser anymore. We can add the missing TLVs for R2 manually, it shouldn't take this long - how many are there?

arnout: Hehe. There are about 30 new TLVs in R2... I agree though that the effort of maintaining that manually is pretty small. I just thought, since @Galvien has the code anyway, we could easily merge it.

tomereli: The code needs to be reviewed again and added to TLVF so it will take Gal at least half a day I assume, and I rather he work on KW issues for now. We can use his branch to regenerate the R2 TLVs when the time comes, but if someone has the time and will to do that, I won't object to merging the automatic parser, its just its really nice to have and not a must.

Galvien: Right now the code serves more as a method of quickly generating parsing code later to be copied and integrated manually into the codebase rather than a mechanism for creating autogenerated parser files.
So the course of action regarding adding new automatic parsers would be running TLVF once with the required configuration for the autogenerator, then copying it's output manually, the whole process should take 5-10 minutes instead of the hours needed for integrating it fully (and the far more difficult aspect - properly) into TLVF, that's at least where it stands ATM...

PR #616, opened 2019-12-27
Build and test on the glinet-b1300

rmelotte:
- build for the glinet-b1300
- test on the glinet-b1300
- collect logs and data and save them as artifacts
- other improvements to the CI

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -77,8 +77,19 @@ build-for-turris-omnia:
   variables:
     TARGET_DEVICE: "turris-omnia"
 
+build-for-glinet-b1300:
+  extends: .build-for-openwrt
+  variables:
+    TARGET_DEVICE: "glinet-b1300"
+
 test-on-turris-omnia:
   extends: .test-on-target
   variables:
     TARGET_DEVICE_NAME: turris-omnia-1
   needs: ["build-for-turris-omnia"]
+
+test-on-glinet-b1300:
+  extends: .test-on-target
+  variables:
+    TARGET_DEVICE_NAME: glinet-b1300-1
+  needs: ["build-for-glinet-b1300"]
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -69,6 +69,10 @@ run-tests:
   script:
     - tools/deploy_ipk.sh $TARGET_DEVICE_NAME tools/docker/builder/openwrt/*.ipk
     - tools/docker/tests/openwrt/test_status.sh $TARGET_DEVICE_NAME
+  artifacts:
+    paths:
+      - ${TARGET_DEVICE_NAME}_logs.tar.gz
+    when: always
   tags:
     - targets
 
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -61,6 +61,11 @@ run-tests:
 
 .test-on-target:
   stage: test
+  before_script:
+    # we need a list of all the targets to be able to kill them
+    - ALL_TARGETS="turris-omnia-1 glinet-b1300-1 netgear-rax40-1"
+    # kill prplmesh on each target to make sure they don't interfere with the test
+    - for i in $ALL_TARGETS ; do ssh "$i" 'pgrep -f beerocks | xargs kill -9 2>/dev/null' || true ; done
   script:
     - tools/deploy_ipk.sh $TARGET_DEVICE_NAME tools/docker/builder/openwrt/*.ipk
     - tools/docker/tests/openwrt/test_status.sh $TARGET_DEVICE_NAME
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -72,6 +72,7 @@ run-tests:
   artifacts:
     paths:
       - ${TARGET_DEVICE_NAME}_logs.tar.gz
+      - ${TARGET_DEVICE_NAME}_diags.log
     when: always
   tags:
     - targets
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -81,9 +81,12 @@ RUN echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
     && scripts/feeds install -a \
     && cp "/home/openwrt/prplmesh_config" .config \
     && make defconfig \
-    && make -j$(nproc) V=sc \
-    && make -j$(nproc) \
+    && cat .config \
+    && make package/feeds/prpl/prplmesh/compile -j$(nproc) V=sc \
+    && make package/feeds/prpl/prplmesh/compile -j$(nproc) \
     && make package/feeds/prpl/prplmesh/clean
     # For some reason, make has to be run twice because
     # it still rebuilds some packages on the second run.
     # On the third run, it's able to compile prplmesh only.
+    # V=sc turns on verbosity for the first command to
+    # make debugging easier.
--- a/tools/docker/builder/openwrt/build.sh
+++ b/tools/docker/builder/openwrt/build.sh
@@ -54,6 +54,8 @@ main() {
 
     eval set -- "$OPTS"
 
+    SUPPORTED_TARGETS="turris-omnia glinet-b1300"
+
     while true; do
         case "$1" in
             -h | --help)               usage; exit 0; shift ;;
@@ -63,9 +65,16 @@ main() {
                     turris-omnia)
                         TARGET=mvebu
                         ;;
+                    glinet-b1300)
+                        TARGET=ipq40xx
+                        ;;
                     *)
                         err "Unknown target device: $TARGET_DEVICE"
-                        info "Currently supported targets are: turris-omnia"
+                        info "Currently supported targets are:"
+			for i in $SUPPORTED_TARGETS ; do
+			    info "$i"
+			done
+                        exit 1
                         ;;
                 esac
                 ;;
--- a/tools/docker/builder/openwrt/build.sh
+++ b/tools/docker/builder/openwrt/build.sh
@@ -9,6 +9,7 @@ usage() {
     echo "usage: $(basename $0) -d <target_device> [-hfiortv]"
     echo "  options:"
     echo "      -h|--help - show this help menu"
+    echo "      -b|--build-options - add docker build options"
     echo "      -d|--target-device the device to build for"
     echo "      -f|--prpl-feed the prpl feed to use"
     echo "      -i|--image - build the docker image only"
@@ -26,6 +27,7 @@ build_image() {
            --build-arg OPENWRT_VERSION \
            --build-arg TARGET \
            --build-arg PRPL_FEED \
+           $BUILD_OPTIONS \
            "$scriptdir/"
 }
 
@@ -48,7 +50,7 @@ build_prplmesh() {
 }
 
 main() {
-    OPTS=`getopt -o 'hd:f:io:r:t:v' --long help,device:,prpl-feed:,image,openwrt-version:,openwrt-repository:,tag:,verbose -n 'parse-options' -- "$@"`
+    OPTS=`getopt -o 'hb:d:f:io:r:t:v' --long help,build-options:,device:,prpl-feed:,image,openwrt-version:,openwrt-repository:,tag:,verbose -n 'parse-options' -- "$@"`
 
     if [ $? != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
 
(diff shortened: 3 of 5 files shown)


tools/docker/builder/openwrt/Dockerfile
```diff
@@ -81,8 +81,9 @@ RUN echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
     && scripts/feeds install -a \
     && cp "/home/openwrt/prplmesh_config" .config \
     && make defconfig \
-    && make -j$(nproc) V=sc \
-    && make -j$(nproc) \
+    && cat .config \
```
tomereli, line 84: Grammar mistake in commit message - 
>Also "cat" the configuration file before compiling, so that we now what the result of "make defconfig" was.

Should be 

> Also "cat" the configuration file before compiling, so that we see what the result of "make defconfig" was.

tomereli, line 84: I'm not sure that dumping the entire .config (is it the whole .config or just the diff?) is the correct thing to do here, we better save it as an artifact like done for logs (I guess we don't dump the whole logs when things go well, do we?)

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -81,8 +81,9 @@ RUN echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
...
+    && make package/feeds/prpl/prplmesh/compile -j$(nproc) V=sc \
```
tomereli, line 85: What does `make V=sc` do? I think it should be documented in a comment

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -81,8 +81,9 @@ RUN echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
...
+    && make package/feeds/prpl/prplmesh/compile -j$(nproc) \
     && make package/feeds/prpl/prplmesh/clean
     # For some reason, make has to be run twice because
```
tomereli, line 88: With this change, do we still need to run make twice? shouldn't we simply clean prplmesh then run make? why the need to compile it twice, then clean it, then run full make? I guess I'm missing something here.

.gitlab-ci.yml
```diff
@@ -72,8 +72,19 @@ build-for-turris-omnia:
   variables:
     TARGET_DEVICE: "turris-omnia"
 
+build-for-glinet-b1300:
```
tomereli, line 85: I can't put the comment in the correct line but I think the previous commit should have added the .config to the artifact paths in this file.

tools/docker/builder/openwrt/build.sh
```diff
@@ -63,9 +63,15 @@ main() {
                     turris-omnia)
                         TARGET=mvebu
                         ;;
+                    glinet-b1300)
+                        TARGET=ipq40xx
+                        ;;
                     *)
                         err "Unknown target device: $TARGET_DEVICE"
-                        info "Currently supported targets are: turris-omnia"
+                        info "Currently supported targets are:"
```
tomereli: It makes sense to put the supported targets in a string and iterate those here.

tools/docker/tests/openwrt/test_status.sh
```diff
@@ -24,3 +24,5 @@ done
 exit 1
 EOF
 
+echo "Stopping prplMesh"
```
tomereli, line 69: This is really not needed since you are doing `restart` in the test.
Also, please add the `-d` flag to the restart command - this will delete the logs from the previous run.

tools/docker/tests/openwrt/test_status.sh
```diff
@@ -14,7 +14,7 @@ TARGET="$1"
 # the first start currently never succeeds, we need to restart it first.
 ssh "$TARGET" <<"EOF"
 /opt/prplmesh/prplmesh_utils.sh restart
-TIMEOUT=10
+TIMEOUT=30
```
tomereli, line 38: It makes more sense to add the TIMEOUT as a script argument IMO.

tools/docker/builder/openwrt/build.sh
```diff
@@ -57,6 +59,7 @@ main() {
     while true; do
         case "$1" in
             -h | --help)               usage; exit 0; shift ;;
+	    -b | --build-options)      BUILD_OPTIONS="$2"; shift ;;
```
tomereli: Indentation looks wrong here

tools/docker/builder/openwrt/build.sh
```diff
@@ -103,7 +107,7 @@ main() {
     export OPENWRT_REPOSITORY
     export OPENWRT_VERSION
     export TARGET
-    PRPLMESH_VERSION="$(git describe --always --dirty --exclude '*')"
+    PRPLMESH_VERSION="$(git describe --always --dirty | sed -e 's/.*-g//')"
```
tomereli, line 116: I would add a comment here saying what this sed command is doing, and change the commit message to mention that instead of exclude which is not supported in some git versions, we are using sed.

tools/docker/tests/openwrt/test_status.sh
```diff
@@ -13,6 +13,7 @@ TARGET="$1"
 
 # the first start currently never succeeds, we need to restart it first.
 ssh "$TARGET" <<"EOF"
+rm -rf /tmp/beerocks/logs/*
```
tomereli: Instead of `rm -rf` please use the `-d` option in the restart command:
```/opt/prplmesh/prplmesh_utils.sh restart -d```

File: tools/docker/tests/openwrt/test_status.sh [hunk 11]
```diff
@@ -11,7 +11,25 @@ fi
...
+ssh "$TARGET" <<"EOF" > "${TARGET}_diags.log"
+date
+echo "Release:"
+cat /etc/*_release
+echo "Version:"
+cat /etc/*_version
+echo
+echo "UCI Configuration (network and wireless):"
+uci export network
+uci export wireless
```
tomereli, line 26: `uci export prplmesh` is missing

tools/docker/tests/openwrt/test_status.sh
```diff
@@ -11,7 +11,25 @@ fi
...
+echo
+echo "ifconfig output:"
+ifconfig
```
tomereli, line 29: I would also add wireless information with `iw`

tools/docker/tests/openwrt/test_status.sh
```diff
@@ -24,15 +42,29 @@ for _ in $(seq 1 "$TIMEOUT") ; do
...
 TEST_STATUS=$?
+
+# Collect post-run diagnostics
+echo "Collecting post-test-run diagnostic data..."
+ssh "$TARGET" <<"EOF" >> "${TARGET}_diags.log"
+date
+echo "Output of bml_conn_map:"
+/opt/prplmesh/bin/beerocks_cli -c bml_conn_map 2>&1
+echo "Output of bml_nw_map_query:"
+/opt/prplmesh/bin/beerocks_cli -c bml_nw_map_query 2>&1
```
tomereli, line 57: What is `bml_nw_map_query`?

tomereli (changes requested):

tools/docker/builder/openwrt/Dockerfile
rmelotte: Although it's being dumped when building the dockerfile, it won't be visible directly in the CI log, so I think it's still ok.
It will only be visible in the build.log file, which already contains a very verbose build output.

It's currently not possible to add the file as an artifact, because the CI does not run _inside_ the container (if we really want to have it as an artifact, we could still copy it out of the container).

tools/docker/builder/openwrt/Dockerfile
rmelotte: `V` controls the verbosity, `sc` is the most verbose. We use that to know what's happening when building. 
For reference, it's documented here: https://openwrt.org/docs/techref/buildroot#warnings_errors_and_tracing
I will add a comment at the end.

tools/docker/builder/openwrt/Dockerfile
rmelotte: We previously had to run it twice because the build system would always try to rebuild the dependencies on the second run (but not on the third). I will check if it's still the case with this change.

Running make then clean is just a way to have the dependencies (but *not* prplmesh itself) already compiled inside the image. I don't know any way to build only the dependencies of a package in OpenWrt.

.gitlab-ci.yml
rmelotte: Is there a specific reason to keep it? It's already in the build.log file (see previous comment).

tools/docker/builder/openwrt/build.sh
rmelotte: Done.

tools/docker/builder/openwrt/build.sh
rmelotte: Done. 

Also fixed: this commit missed passing BUILD_OPTIONS to build_image() (i.e. BUILD_OPTIONS was unused).

tools/docker/builder/openwrt/build.sh
rmelotte: Done.

tools/docker/tests/openwrt/test_status.sh
rmelotte: Done.

tools/docker/tests/openwrt/test_status.sh
rmelotte: Done.

tools/docker/tests/openwrt/test_status.sh
rmelotte: Done.

tools/docker/tests/openwrt/test_status.sh
rmelotte: It's at least being used by the analyzer, and from what I recall its output can be different than `bml_conn_map`.
I don't know to what extent it's relevant to have it in the archive though.

@dob71 do you have a use case where `bml_nw_map_query` is useful?

tools/docker/tests/openwrt/test_status.sh
rmelotte: We need to stop it at the end of the test to avoid the multiple target from interfering with one another.
Now this one is not strictly needed because we will kill all the instances before the tests starts, but it gives it a chance to stop more gracefully.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Done.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Yes, running make twice is still needed.
We can probably avoid it by using CONFIG_AUTOREBUILD, but that can probably be done in another PR, because it needs testing.

tools/docker/builder/openwrt/build.sh
```diff
@@ -80,10 +96,11 @@ main() {
         esac
     done
 
+    dbg "BUILD_OPTIONS=$BUILD_OPTIONS"
     dbg "OPENWRT_REPOSITORY=$OPENWRT_REPOSITORY"
     dbg "OPENWRT_VERSION=$OPENWRT_VERSION"
     dbg "PRPL_FEED=$PRPL_FEED"
-    dbg "IMAGE_ONlY=IMAGE_ONLY"
+    dbg "IMAGE_ONlY=$IMAGE_ONLY"
```
arnout: There was still a typo in the prompt here, so I fixed that as well.

tools/docker/builder/openwrt/build.sh
```diff
@@ -48,24 +54,34 @@ build_prplmesh() {
...
     if [ $? != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
 
     eval set -- "$OPTS"
 
+    SUPPORTED_TARGETS="turris-omnia glinet-b1300"
+
     while true; do
         case "$1" in
             -h | --help)               usage; exit 0; shift ;;
+            -b | --build-options)      BUILD_OPTIONS="$2"; shift ;;
```
arnout: There's a missing shift here, so the option will be interpreted as a script option as well...

tools/docker/builder/openwrt/build.sh
```diff
@@ -48,24 +54,34 @@ build_prplmesh() {
...
+                        TARGET=ipq40xx
+                        ;;
                     *)
                         err "Unknown target device: $TARGET_DEVICE"
-                        info "Currently supported targets are: turris-omnia"
+                        info "Currently supported targets are:"
+			for i in $SUPPORTED_TARGETS ; do
+			    info "$i"
+			done
+                        exit 1
```
arnout: Tabs instead of spaces...

tools/docker/builder/openwrt/build.sh
```diff
@@ -48,24 +54,34 @@ build_prplmesh() {
...
     while true; do
         case "$1" in
             -h | --help)               usage; exit 0; shift ;;
+            -b | --build-options)      BUILD_OPTIONS="$2"; shift ;;
             -d | --target-device)
                 TARGET_DEVICE="$2"; shift ; shift
                 case "$TARGET_DEVICE" in
                     turris-omnia)
                         TARGET=mvebu
                         ;;
```
arnout: This code wasn't executed at all if the -d option is not given, so there is no error in that case. To simplify the code, I just moved this entire case block after the loop. That also makes it more readable IMO.

tools/docker/builder/openwrt/build.sh
```diff
@@ -20,13 +21,18 @@ usage() {
...
+	ARGS="$* $scriptdir/"
+    fi
     # We first need to build the corresponding images
     docker build --tag "$image_tag" \
            --build-arg OPENWRT_REPOSITORY \
            --build-arg OPENWRT_VERSION \
            --build-arg TARGET \
            --build-arg PRPL_FEED \
-           "$scriptdir/"
+           "$ARGS"
```
arnout: This isn't going to work: the whole thing will be passed as a single argument with an embedded space and won't be recognized as an option.

I changed this to instead use the BUILD_OPTIONS directly, unquoted (so if multiple options are provided they are properly passed as individual options).

arnout (approved):
The new -b option didn't work, so I did a bunch of fixes.

There were also a bunch of tabs/spaces indentation errors. Perhaps we should add a [modeline](https://help.gnome.org/users/gedit/stable/gedit-plugins-modelines.html) to all files so editors do the right thing? @tomereli @rmelotte @dob71 @morantr does your editor support modeline? Normal editors (i.e. vim, emacs, kate, gedit) of course do support it, and I know that for Atom you need an [extension](https://atom.io/packages/vim-modeline), but I don't know about weird editors like vscode.

tomereli (approved):

PR #617, opened 2019-12-27
Feature/ci build netgear rax40

rmelotte:
Keeping it as a draft until we can confirm all tests from CI pass. 
Will also need to integrate changes from #616 .

### DIFF
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -73,6 +73,8 @@ COPY --chown=openwrt:openwrt configs/prplmesh_config /home/openwrt/prplmesh_conf
 ARG TARGET_PLATFORM_TYPE
 # The feed to use to build prplMesh.
 ARG PRPL_FEED
+# The variant to build (nl80211 or dwpal)
+ARG PRPLMESH_VARIANT
 
 USER openwrt
 
@@ -80,6 +82,7 @@ RUN echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
     && scripts/feeds update -a \
     && scripts/feeds install -a \
     && cp "/home/openwrt/prplmesh_config" .config \
+    && echo "CONFIG_PACKAGE_prplmesh${PRPLMESH_VARIANT}=y" >> .config \
     && make defconfig \
     && cat .config \
     && make package/feeds/prpl/prplmesh/compile -j$(nproc) V=sc \
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -23,6 +23,9 @@ FROM openwrt-prerequisites as openwrt-prebuilt
 
 USER openwrt
 
+# optional: intel feed to use.
+ARG INTEL_FEED
+
 # The following args are mandatory, do not expect the build to
 # succeed without specifying them on the command line.
 #
@@ -41,7 +44,12 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
 
 WORKDIR /home/openwrt/openwrt
 
-RUN echo "CONFIG_SDK=y" > .config \
+RUN if [ -n "$INTEL_FEED" ] ; then \
+	echo "src-git intel $INTEL_FEED" >> feeds.conf.default \
+	&& scripts/feeds update intel \
+	&& scripts/feeds install intel_mips ; \
+    fi ; \
+    echo "CONFIG_SDK=y" > .config \
     && echo "CONFIG_TARGET_${TARGET}=y" >> .config \
     && make defconfig \
     && make -j$(nproc)
@@ -78,8 +86,12 @@ ARG PRPLMESH_VARIANT
 
 USER openwrt
 
+# We first install from the intel feed preferably, because
+# it replaces some packages from the "packages" feed.
+# If no intel feed is present, nothing will be done so it's still ok.
 RUN echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
     && scripts/feeds update -a \
+    && scripts/feeds install -a -p intel \
     && scripts/feeds install -a \
     && cp "/home/openwrt/prplmesh_config" .config \
     && echo "CONFIG_PACKAGE_prplmesh${PRPLMESH_VARIANT}=y" >> .config \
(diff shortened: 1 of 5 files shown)


tools/docker/builder/openwrt/build.sh
```diff
@@ -26,6 +26,7 @@ build_image() {
            --build-arg OPENWRT_VERSION \
            --build-arg TARGET \
            --build-arg PRPL_FEED \
+	   --build-arg PRPLMESH_VARIANT \
```
arnout: Yet another tab problem...

tools/docker/builder/openwrt/build.sh
```diff
@@ -86,6 +87,8 @@ main() {
     dbg "IMAGE_ONlY=IMAGE_ONLY"
     dbg "TARGET_DEVICE=$TARGET_DEVICE"
     dbg "TAG=$TAG"
+    dbg "TARGET=$TARGET"
```
arnout: This is an unrelated change and I'm not sure if it's very relevant.

tools/docker/builder/openwrt/scripts/build.sh
```diff
@@ -7,5 +7,5 @@ cp -r /home/openwrt/prplMesh_source /home/openwrt/prplMesh
 rm -rf /home/openwrt/prplMesh/build
 
 make package/feeds/prpl/prplmesh/prepare USE_SOURCE_DIR="/home/openwrt/prplMesh" V=s
-make package/feeds/prpl/prplmesh/compile V=sc -j"$(nproc)"
-find bin -name 'prplmesh_*.ipk' -exec cp -v {} "prplmesh-${TARGET}-${OPENWRT_VERSION}-${PRPLMESH_VERSION}.ipk" \;
+make package/feeds/prpl/prplmesh/compile VARIANT="$PRPLMESH_VARIANT" V=sc -j"$(nproc)"
```
arnout: This shouldn't be needed since the variant is specified in the config, right? 

Also, since it's an ARG and not an ENV, is it even saved in the environment when running the container?

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -41,7 +44,12 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
 
 WORKDIR /home/openwrt/openwrt
 
-RUN echo "CONFIG_SDK=y" > .config \
+RUN if [ -n "$INTEL_FEED" ] ; then \
+	echo "src-git intel $INTEL_FEED" >> feeds.conf.default \
+	&& scripts/feeds update intel \
+	&& scripts/feeds install intel_mips ; \
```
arnout, line 50: Can you explain better in the commit message why it is necessary to explicitly install this one?

tools/docker/builder/openwrt/build.sh
```diff
@@ -10,7 +10,6 @@ usage() {
     echo "  options:"
     echo "      -h|--help - show this help menu"
     echo "      -d|--target-device the device to build for"
-    echo "      -f|--prpl-feed the prpl feed to use"
```
arnout, line 14: Can you extend the help text with the environment variables that affect the build? I.e., PRPL_FEED and INTEL_FEED. And include their defaults as well.

tools/docker/builder/openwrt/build.sh
```diff
@@ -64,6 +64,12 @@ main() {
                     turris-omnia)
                         TARGET=mvebu
                         ;;
+                    netgear-rax40)
+                        TARGET=intel_mips
+                        PRPLMESH_VARIANT="-dwpal"
+                        PRPL_FEED="https://git.prpl.dev/prplmesh/iwlwav.git^e9ee6b109764dbb91b6cb399ccc0f78e5b04e260"
```
arnout: Is this really still needed? Can't we still use this feed by specifying the variant?

I'm running a test build, if successful I'll push that with fixup commits.

arnout (approved):
Easily fixed comments only.

I did a rebase on master to be able to test some things, so I'll force-push that after running a test build.

arnout:
Build didn't complete yet but I had to go home, so let's wait for CI to report if it works or not...

tools/docker/builder/openwrt/scripts/build.sh
rmelotte: Last time I tried it didn't use the correct variant even when it was selected in the config file.

This was an attempt to force it to use the correct variant, but it was a bad idea: `VARIANT` will be the same for all the packages, which means that the hostapd variant for example, will be `-dwpal` or `-nl80211`...

Removing.

tools/docker/builder/openwrt/build.sh
rmelotte: When using the latest feed for the nl80211 targets, dwpal gets compiled although it should not (the intel feed is missing so it fails with safeclibs missing).
I suggest we first merge this PR using the 2 separate feed versions and open an issue about it.
That way we can finally build the dwpal variant in CI and at least spot potential compilation issues introduced by other changes early.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Done.

tools/docker/builder/openwrt/build.sh
rmelotte: Done.

rmelotte:
For reference, latest pipeline was here: https://gitlab.com/prpl-foundation/prplMesh/pipelines/106383852
It still fails because it compiles dwpal even for the nl80211 targets.
I'll force push again with the 2 different feeds version, and the changes that were requested. 

I kept a copy of the 2 additional fixup commits in case we still want them.

rmelotte:
I changed the value of `INTEL_FEED` to be an empty string (to make it obvious that it's being used in the script, like we do for the other variables). Because of this, the CI needs to rebuild the whole image.

I had to cancel the latest one to avoid compiling the same image twice at the same time (which is why github shows the pipeline as canceled).

Marking it as ready to review already, but we will need to re-run the pipeline once the current build is done.

arnout (approved):

rmelotte:
The build failed because of a container name clash, so I added an additional commit to use unique container names (image names were already unique, containers names were only timestamped).

I also noticed the containers weren't always removed, so I added another commit to make sure containers are removed even when the build fails.

Finally, I also fixed some debug messages which printed wrong values.

rmelotte:
Added quotes to the target in gitlab-ci.yml, since every other target are quoted.

rmelotte:
Rebased on master.

PR #618, opened 2019-12-29
Bugfix/autoconfig fix credentials set

tomereli:
Fix bug in autoconfig which got passed test_flows.sh - the ap manager thread on receipt of the config_data containing the credentials that needs to be updated, did an extra class_swap which caused some of the fields to be swapped.

So fixed that - but while we're at it, make sure this doesn't happen again by updating test_flows.sh and the dummy implementation to test the final credentials.

Also - since the initial_autoconfig included also the renew part, split it to a separate test.

### DIFF
--- a/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
@@ -177,13 +177,25 @@ bool ap_wlan_hal_dummy::update_vap_credentials(
     std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list)
 {
     for (auto bss_info_conf : bss_info_conf_list) {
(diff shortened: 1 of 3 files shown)


common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -177,13 +177,19 @@ bool ap_wlan_hal_dummy::update_vap_credentials(
     std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list)
 {
     for (auto bss_info_conf : bss_info_conf_list) {
-        LOG(DEBUG) << "Received credentials for ssid: " << bss_info_conf.ssid << " auth_type: "
```
vitalybu, line 180: Why did you move the DEBUG print after the validations?
If something goes wrong you won't have any prints at all... The function simply returns `false`

vitalybu (approved):

Issue #562, opened 2019-12-11
[TASK] port Dynamic channel selection (DCS) feature from RDKB


kantera800:
## Description

This task will be focused on porting Dynamic Channel Selection code from Intel repositories (RDKB) .

## Exit criteria

- run DCS feature on RDKB platform and pass attached unitest.

- pass all prplMesh CI checks.

## Detailed description

The DCS as a feature can provides information on the supported channels which can be utilized by the management system to provide better availability and user experience.

The feature is divided into two sub-features:
1. Channel Scan using NL & DWPAL
Uses the DWPAL to perform a netlink channel scan to find all neighboring APs
2. Dymanic Channel Selection & Analysis
Returns the Channel Scan results to provide the current status of the channels

The feature is presented in five segments each will have its own PR as follows:

- [x] Low Level Infastructure (DWPAL): NL support in BWL

- [feature/port-dynamic-channel-selection-from-rdkb-bwl](https://github.com/prplfoundation/prplMesh/tree/feature/port-dynamic-channel-selection-from-rdkb-bwl) > [PR #637](https://github.com/prplfoundation/prplMesh/pull/637) **Merged**
- bwl: cmake: add nl libs for dwpal
- bwl: Add NL support (with stubs)
- bwl: Add NL DWPAL base implementation
- bwl: Add NL DWPAL monitor implementation

- [x] High Level Infastructure (Controller & yaml): Support for Channel Scan & DCS

- [feature/port-dynamic-channel-selection-from-rdkb-infastructure](https://github.com/prplfoundation/prplMesh/tree/feature/port-dynamic-channel-selection-from-rdkb-infastructure) > [PR #643](https://github.com/prplfoundation/prplMesh/pull/643) **Merged**
- beerocks defines: enum definitions for Channel Scan & DCS
- yaml: beerocks_message_common: Add channel scan structures
- yaml: beerocks_message_control: Add channel scan & DCS CMDUs
- controller: node: Add channel scan structures
- controller: DB: Add channel scan DB methods

- [x] Dymanic Channel Selection Task

- [feature/port-dynamic-channel-selection-from-rdkb-task](https://github.com/prplfoundation/prplMesh/tree/feature/port-dynamic-channel-selection-from-rdkb-task) > [PR #644](https://github.com/prplfoundation/prplMesh/pull/644) **Merged**
- controller: task: add clear pending events
- controller: dymanic channel selection task intro
- controller: DCS task: FSM
- controller: DCS task: events
[...]

PR #619, opened 2019-12-30
Feature/port dynamic channel selection from rdkb

itayx:
This PR is POC DO NOT COMMENT, DO NOT MERGE 

[DCS Patch 1/5] port dynamic channel selection from rdkb bwl #637 
[DCS Patch 2/5] port dynamic channel selection from rdkb infrastructure #643 
[DCS Patch 3/5] port dynamic channel selection from rdkb tasks #644 

The DCS as a feature can provide information on the supported channels which can be utilized by the management system to provide better availability and user experience. #562


common/beerocks/bwl/CMakeLists.txt
```diff
@@ -59,7 +59,7 @@ if(BWL_TYPE STREQUAL "DWPAL")
     )
 
     find_package(dwpal REQUIRED)
-    list(APPEND BWL_LIBS dwpal)
+    list(APPEND BWL_LIBS dwpal nl-3 nl-genl-3)
```
tomereli, line 62: You can't just assume nl-3 and nl-genl-3 are found and are part of the default library and header search paths. for this we have in CMAKE the concept of "Find module files" - for example, see `cmake/Finddwpal.cmake` in the root of prplmesh. This commit need to be split to two - first commit which adds `cmake/Findnl.cmake` and second commit which adds it in this file by first using find_package then adding it as a target to the list:

```
find_package(nl-3)
find_package(nl-genl-3)
list(APPNED BWL_LIBS dwpal nl-3 nl-genl-3)
```

tomereli: Also I think that the nl dependency should be global and not just for DWPAL - @arnout what do you think? will we need NLs in general?

common/beerocks/bwl/common/base_wlan_hal.cpp
```diff
@@ -27,7 +27,7 @@ base_wlan_hal::base_wlan_hal(HALType type, std::string iface_name, IfaceType ifa
     // Initialize radio info structure
     m_radio_info.iface_name = iface_name;
     m_radio_info.iface_type = iface_type;
-
+
```
tomereli, line 30: You accidentally added line breaks here. Please run clang-format for every commit.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -19,6 +19,8 @@ extern "C" {
 #include <dwpal.h>
 }
 
+#include <slibc/string.h>
```
tomereli, line 22:
```
#ifdef USE_LIBSAFEC
#define restrict __restrict
#include <libsafec/safe_str_lib.h>
#elif USE_SLIBC
#include <slibc/string.h>
#else
#error "No safe C library defined, define either USE_LIBSAFEC or USE_SLIBC"
#endif
```

tomereli: Actually, since it is used now by both base_wlan_hal_dwpal.cpp and ap_wlan_hal_dwpal.cpp - I suggest to move it to `base_wlan_hal_dwpal.h`.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -136,6 +138,21 @@ bool base_wlan_hal_dwpal::fsm_setup()
                     attached = true;
                 }
 
+                //Attach NL interface
+                if (m_dwpal_nl_ctx == nullptr) {
```
tomereli, line 142: KW error - change to `if (!m_dwpal_nl_ctx)`

And please make sure to run KW on this change before submitting it, next time.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -288,10 +305,27 @@ bool base_wlan_hal_dwpal::fsm_setup()
                     LOG(ERROR) << "Can't get event fd since dwpal ctx is NULL!";
                     return (transition.change_destination(dwpal_fsm_state::Detach));
                 }
-
                 // Success
                 LOG(DEBUG)
                     << "Open and attach an event interface to wpa_supplicant/hostapd - SUCCESS!";
+
+                // Get the nl event interface file descriptor
+                if (m_dwpal_nl_ctx != nullptr) {
```
tomereli, line 313: `if (m_dwpal_nl_ctx) {`

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -331,6 +365,17 @@ bool base_wlan_hal_dwpal::fsm_setup()
 
             m_fd_ext_events = -1;
 
+            // detach nl socket from main vap
+            if (!m_dwpal_nl_ctx) {
```
tomereli, line 369: Looks like a bug - this means that the context is NULL, and you give a reference to it to the `dwpal_driver_nl_detach` function.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+        LOG(ERROR) << "Invalid WPA Control socket used for nl events (m_dwpal_ctx == nullptr)";
+        return false;
+    }
+
+    // Passing a lambda with capture is not supported for standard C function
+    // pointers. As a workaround, we create a static (but thread local) wrapper
+    // function that calls the capturing lambda function.
+    static __thread std::function<DWPAL_Ret(struct nl_msg * msg)> nl_handler_cb_wrapper;
+    nl_handler_cb_wrapper = [&](struct nl_msg *msg) -> DWPAL_Ret {
+        if (process_dwpal_nl_event(msg) != true) {
```
tomereli, line 568: ` if (!process_dwpal_nl_event(msg))`

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            LOG(ERROR) << "User's netlink handler function failed!";
+            return DWPAL_FAILURE;
+        }
+        return DWPAL_SUCCESS;
+    };
+    auto nl_handler_cb = [](struct nl_msg *msg) -> DWPAL_Ret { return nl_handler_cb_wrapper(msg); };
+
+    if (m_fd_nl_events > 0) {
```
tomereli, line 576: How can this happen if the context is valid? in any case, move this as another sanity check to the top of this function.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+        //parsing will be done in callback function
+        if (dwpal_driver_nl_msg_get(m_dwpal_nl_ctx, DWPAL_NL_UNSOLICITED_EVENT, NULL,
+                                    nl_handler_cb) == DWPAL_FAILURE) {
+            LOG(ERROR) << " dwpal_driver_nl_msg_get failed,"
+                       << " ctx=" << m_dwpal_nl_ctx;
+            detach();
```
tomereli, line 582: Will detach clear the context and the file descriptors?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool base_wlan_hal_dwpal::dwpal_nl_cmd_set(const std::string &ifname, unsigned int nl_cmd,
+                                           unsigned char *vendor_data, size_t vendor_data_size)
```
tomereli, line 591: Why not pass the vendor data as a vector?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+    if ((vendor_data_size > 0) && (vendor_data == NULL)) {
+        LOG(ERROR) << __FUNCTION__ << "ifname or vendor_data is NULL ==> Abort!\n";
+        return false;
+    }
+
+    if (dwpal_driver_nl_cmd_send(m_dwpal_nl_ctx, DWPAL_NL_UNSOLICITED_EVENT, (char *)ifname.c_str(),
+                                 NL80211_CMD_VENDOR, DWPAL_NETDEV_ID,
+                                 (enum ltq_nl80211_vendor_subcmds)nl_cmd, vendor_data,
+                                 vendor_data_size) != DWPAL_SUCCESS) {
+        err = false;
```
tomereli, line 603: What? There should be an error log here and `return false`

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+    }
+
+    return err;
```
tomereli, line 606: if we got till here - `return true` is OK. Then get rid of the `err` variable.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+    }
+
+    return true;
+}
+
+bool base_wlan_hal_dwpal::dwpal_nl_cmd_set(const std::string &ifname, unsigned int nl_cmd,
+                                           unsigned char *vendor_data, size_t vendor_data_size)
+{
+    bool err = true;
+    if ((vendor_data_size > 0) && (vendor_data == NULL)) {
```
tomereli, line 594: Why are you using `NULL` and not `nullptr`? This pattern goes on in all this file. In general why do you use C structures and not CPP structures from the STL library? Does it have to be C compliant for some reason?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+                                 (enum ltq_nl80211_vendor_subcmds)nl_cmd, vendor_data,
+                                 vendor_data_size) != DWPAL_SUCCESS) {
+        err = false;
+    }
+
+    return err;
+}
+
+bool base_wlan_hal_dwpal::dwpal_nl_cmd_get(const std::string &ifname, unsigned int nl_cmd,
+                                           unsigned char *out_buffer, size_t *out_buffer_size,
```
tomereli, line 610: Here it makes sense to have the `out_buffer` as `unsigned char *` since it is given to a C API, but I still think we should not explicitly compare it to `NULL` but rather use the `if (!var)` style. @arnout what do you think?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+{
+
+    if ((out_buffer_size == NULL) || (out_buffer == NULL)) {
+        LOG(ERROR) << __FUNCTION__ << "out_buffer or out_buffer_size incorrect ==> Abort!\n";
+        return false;
+    }
+
+    /* Handle a command which invokes an event with the output data */
+    if (dwpal_driver_nl_cmd_send(
+            m_dwpal_nl_ctx, DWPAL_NL_SOLICITED_EVENT, (char *)ifname.c_str(), NL80211_CMD_VENDOR,
```
tomereli, line 621: Why do you need the C style cast here? doesn't `c_str()` return `char *`? Even if not, you should use `static_cast<char*>` and not C style cast. (nitpick)

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            DWPAL_NETDEV_ID, (enum ltq_nl80211_vendor_subcmds)nl_cmd, NULL, 0) == DWPAL_FAILURE) {
```
tomereli, line 622: same here - don't use C style cast (nitpick)

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+        LOG(DEBUG) << __FUNCTION__ << "ERROR for cmd = " << nl_cmd;
+        return false;
+    }
+
+    // Passing a lambda with capture is not supported for standard C function
```
tomereli, line 627: Can you add a URL which proves this is not supported? for example, the top answer in https://stackoverflow.com/questions/28746744/passing-capturing-lambda-as-function-pointer explains this perfectly with references to the c++11 standard.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+    // pointers. As a workaround, we create a static (but thread local) wrapper
+    // function that calls the capturing lambda function.
+    static __thread std::function<DWPAL_Ret(char *ifname, int event, int subevent, size_t len,
+                                            unsigned char *data)>
+        nl_handler_cb_wrapper;
```
tomereli, line 632: This line should go away, it does nothing.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+    nl_handler_cb_wrapper = [&](char *ifname, int event, int subevent, size_t len,
+                                unsigned char *data) -> DWPAL_Ret {
+        LOG(DEBUG) << __func__ << " ifname = " << std::string(ifname) << " event = " << event
```
tomereli, line 635: passing `ifname` to `std::string` constructor is assuming that `ifname` is null terminated - are you sure this is the case? I would use the `std::string(char *, size_t)` constructor instead, and define a BWL_IFNAME_LEN as we do in BPL, and use it as the maximum size.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            LOG(DEBUG) << __func__ << "len=0 and/or data is NULL ==> Abort!\n";
+            return DWPAL_FAILURE;
+        }
+        if (event == NL80211_CMD_VENDOR) {
+            if (len >= max_buffer_size) {
+                LOG(DEBUG) << __func__ << "NL size exceeds out_buffer size ==> Abort!\n";
+                return DWPAL_FAILURE;
+            }
+            // copy get result from nl buffer (data[i])
+            for (int i = 0; i < (int)len; i++) {
```
tomereli, line 648: Wait what? Why not use `std::copy` or `std::copy_n` or safe_c memcpy_s if there is such API?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+                out_buffer[i] = data[i];
+            }
+            *out_buffer_size = len;
+
+        } else {
+            LOG(DEBUG) << __func__ << "non vendor event not handled ";
```
tomereli, line 654: If you return ERROR, it should be `LOG(ERROR)`

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            return DWPAL_FAILURE;
+        }
+        return DWPAL_SUCCESS;
+    };
+    auto nl_handler_cb = [](char *ifname, int event, int subevent, size_t len,
+                            unsigned char *data) -> DWPAL_Ret {
+        return nl_handler_cb_wrapper(ifname, event, subevent, len, data);
+    };
+
+    while (*out_buffer_size == 0) { /* wait for reply with valid data */
```
tomereli, line 664: I think it makes much more sense to add a promise or something like this instead of this (ugly) hack.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool base_wlan_hal_dwpal::dwpal_nl_cmd_scan_dump()
+{
+    bool err = true;
```
tomereli, line 681: here as well as all other places - `bool err` should be removed, it is simply not needed.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            LOG(ERROR) << "User's netlink handler function failed!";
+            return DWPAL_FAILURE;
+        }
+        return DWPAL_SUCCESS;
+    };
+    auto nl_handler_cb = [](struct nl_msg *msg) -> DWPAL_Ret { return nl_handler_cb_wrapper(msg); };
+
+    if (dwpal_driver_nl_scan_dump(m_dwpal_nl_ctx, (char *)m_radio_info.iface_name.c_str(),
+                                  nl_handler_cb) != DWPAL_SUCCESS) {
+        err = false;
```
tomereli, line 697: add `LOG(ERROR)` and return false

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -732,5 +925,11 @@ std::string base_wlan_hal_dwpal::get_radio_mac()
     return mac;
 }
 
+bool base_wlan_hal_dwpal::process_dwpal_nl_event(struct nl_msg *msg)
+{
+    LOG(DEBUG) << __func__ << "isn't implemented by derived and shouldn't be called";
```
tomereli, line 930: So why isn't it pure virtual? then it MUST be implemented by the derived class.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
```diff
@@ -53,6 +54,7 @@ class base_wlan_hal_dwpal : public virtual base_wlan_hal,
 
     // Process dwpal event
     virtual bool process_dwpal_event(char *buffer, int bufLen, const std::string &opcode) = 0;
+    virtual bool process_dwpal_nl_event(struct nl_msg *msg);
```
tomereli, line 57: Should be pure virtual.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -10,13 +10,18 @@
...
 
 #include <easylogging++.h>
+#include <net/if.h>
 
 #include <cmath>
+#include <functional>
 
 extern "C" {
 #include <dwpal.h>
+#include <slibc/stdio.h>
```
tomereli, line 23: This won't work with UGW - this is exatcly why we have the `HAVE_SAFEC` flags.
You need (here and in all other places you use safec headers) to use this format:
```
#ifdef USE_LIBSAFEC
#define restrict __restrict
#include <libsafec/safe_str_lib.h>
#elif USE_SLIBC
#include <slibc/string.h>
#else
#error "No safe C library defined, define either USE_LIBSAFEC or USE_SLIBC"
#endif
```

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -39,6 +44,38 @@ namespace dwpal {
             delete[] obj;                                                                          \
     })
 
+// Helper macro for dwpal_nl_cmd_get
+// interfaceName, out_buffer, out_buffer_size, max_buffer_size, paramsObj
+#define DWPAL_GET_SCAN_PARAMS_HELPER_FG(IFNAME, OUT_BUFF, OUT_BUFF_SZ, MAX_BUFF_SZ, PRMS)          \
```
tomereli, line 49: I really don't like these style of MACROs especially that we have c++11. Please create a static inline functions and use them instead.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -62,6 +99,26 @@ static mon_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
     return mon_wlan_hal::Event::Invalid;
 }
 
+static mon_wlan_hal::Event dwpal_nl_to_bwl_event(uint8_t cmd, bool waiting_for_results_ready)
+{
+    if (cmd == NL80211_CMD_TRIGGER_SCAN) {
```
tomereli, line 104: I think that `switch-case` is more appropriate here.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
 }
 
+bool mon_wlan_hal_dwpal::dcs_scan_trigger(int dwell_time_msec,
+                                          const std::vector<unsigned int> &channel_pool)
+{
+    LOG(DEBUG) << __func__ << " received on interface=" << m_radio_info.iface_name;
+
+    //build scan parameters
+    ScanParams dcs_scan_params;
+    memset((void *)&dcs_scan_params, 0, sizeof(dcs_scan_params)); //must to ensure freq are zero
```
tomereli, line 476: Don't use `memset`. I think it is banned c function. Use the appropriate safec function, or simply use `ScanParams dcs_scan_params = {};`

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+    size_t out_len                                = 0;
+    unsigned char out_data[NL_MAX_REPLY_BUFFSIZE] = {'\0'};
```
tomereli, line 478: Why is this needed? you provide the buffer length so why should you set null termination explicitly?

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+    sScanCfgParams fg, new_fg;   //foreground scan param
+    sScanCfgParamsBG bg, new_bg; //background scan param
+
+    auto ifname = m_radio_info.iface_name;
+
+    DWPAL_GET_SCAN_PARAMS_HELPER_FG(ifname, out_data, out_len, NL_MAX_REPLY_BUFFSIZE, fg);
```
tomereli, line 484: Should use static inline functions and check return value.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+                return false;
+            }
+        }
+
+        dcs_scan_params.freq[freq_index] = beerocks::utils::wifi_channel_to_freq(int(channel));
+        LOG(DEBUG) << " dcs scan pool add center frequency=" << dcs_scan_params.freq[freq_index];
+        freq_index++;
+    }
+
+    dcs_scan_params.ap_force = 1;
```
tomereli, line 545: Please add a comment explaining what is this setting is done here and why is it not the default.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+    LOG(DEBUG) << "Trigger scan success!"
+               << " scan params dwell_time_msec restored, passive = " << fg.passive
+               << " active = " << fg.active;
+
+    return true;
+}
+
+bool mon_wlan_hal_dwpal::dcs_scan_dump_results()
+{
+    if (dwpal_nl_cmd_scan_dump() != true) {
```
tomereli, line 586: `if (!func())`

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -582,6 +764,111 @@ bool mon_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std
     return true;
 }
 
+bool mon_wlan_hal_dwpal::process_dwpal_nl_event(struct nl_msg *msg)
+{
+    struct nlmsghdr *nlh     = nlmsg_hdr(msg);
+    struct genlmsghdr *gnlh  = (genlmsghdr *)nlmsg_data(nlh);
+    char ifname[IF_NAMESIZE] = "\0";
+    struct nlattr *tb[NL80211_ATTR_MAX + 1];
+    const std::string temp = "abc";
```
tomereli, line 773: unused variable `temp`

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal_types.h
```diff
@@ -0,0 +1,40 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
```
tomereli, line 3: Its a new file - `* Copyright (c) 2019 Intel Corporation`

tomereli (changes requested):
Reviewed up till https://github.com/prplfoundation/prplMesh/pull/619/commits/e5b56810350c22ff9dfd6e53c310b3fe173ac6c3

The commit split is not ideal, to say the least.
I reviewed the first 2 commits since I believe the second commit should be split to at least 2 commits.
The rest of the commits should be a separate PR, or more than that, depends on their content.
So, I suggest to fix my review comments in a new PR and rebase this PR on top of it. The new PR should include the following commits:

First commit split:
1. cmake: Add Findnl-3.cmake
2. cmake: Add Findnl-genl-3.cmake
3. bwl: cmake: add nl libs for dwpal

Second commit split:
1. bwl: Add NL support (with stubs)
2. bwl: Add NL DWPAL base implementation
3. bwl: Add NL DWPAL monitor implementation

Documenataion - there is none. You added a huge change with lots of new function, you need to document each new method in the header file or the cpp file (align to what is currently used in the areas you changed) with doxygen format.

Finally, I would like @arnout to also review the first commits so you can fix the changes once at least for the first PR as I suggested.

common/beerocks/bwl/CMakeLists.txt
arnout: Although I agree that it should be done like that (though since we always need genl I would just do `find_package(nl-genl REQUIRED)`), it's already done in the hacked way in bpl, and it's already hacked in a different way for nl80211 which hardcodes nl-tiny, so I wouldn't add that change as a requirement for this PR.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: I had a vague plan that this should be factored into a separate header file, perhaps generated from cmake or symlinked or something from cmake, and that cmake is responsible for discovering if libsafec or slibc is available instead of tying it to the platform. But again, that's a bigger change.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -136,6 +138,21 @@ bool base_wlan_hal_dwpal::fsm_setup()
...
+                    if (dwpal_driver_nl_attach(&m_dwpal_nl_ctx) == DWPAL_FAILURE) {
+                        LOG(ERROR) << "dwpal_driver_nl_attach returned ERROR for interface ="
+                                   << m_radio_info.iface_name;
+                        return (transition.change_destination(dwpal_fsm_state::Detach));
+                    }
+                    LOG(DEBUG) << "dwpal_driver_nl_attach() success for radio ="
+                               << m_radio_info.iface_name;
+                } else {
+                    //Already attached
+                    attached = true;
```
arnout, line 153: This is wrong: if attached was set to false before because the attach_ctrl_interface() call failed, we shouldn't override it here.

I think there simply shouldn't be an else branch.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -136,6 +138,21 @@ bool base_wlan_hal_dwpal::fsm_setup()
                     attached = true;
                 }
 
+                //Attach NL interface
+                if (m_dwpal_nl_ctx == nullptr) {
+                    //Attach nl interface
+                    if (dwpal_driver_nl_attach(&m_dwpal_nl_ctx) == DWPAL_FAILURE) {
+                        LOG(ERROR) << "dwpal_driver_nl_attach returned ERROR for interface ="
+                                   << m_radio_info.iface_name;
+                        return (transition.change_destination(dwpal_fsm_state::Detach));
```
arnout, line 147: It may be worth to add a comment that explains why in case of error from attach_ctrl_interface we stay in Attach state, but in case of error from dwpal_driver_nl_attach we go to Detach state immediately.

(I think the reasoning is that retrying an nl attach will never lead to success, while retrying ctrl attach may be successful after a while because hostapd has started up).

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout, line 313: This check is kind of pointless: we can only get into this state if the nl attach has succeeded.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -288,10 +305,27 @@ bool base_wlan_hal_dwpal::fsm_setup()
...
                 }
-
                 // Success
                 LOG(DEBUG)
                     << "Open and attach an event interface to wpa_supplicant/hostapd - SUCCESS!";
+
+                // Get the nl event interface file descriptor
+                if (m_dwpal_nl_ctx != nullptr) {
+                    LOG(DEBUG) << "new  nl interface";
+                    if (dwpal_driver_nl_fd_get(m_dwpal_nl_ctx, &m_fd_nl_events, &m_fd_nl_cmd_get)) {
```
arnout, line 315: There is no reason to do this here, it is much more appropriate to do it as part of attaching the nl interface in the Init -> Attach transition.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: It can't be pure virtual because it will only be implemented by the monitor class.

Which makes me wonder why this stuff is in the base class anyway, if it's only used for dwpal monitor.

common/beerocks/bwl/include/bwl/base_wlan_hal.h
```diff
@@ -92,6 +92,15 @@ class base_wlan_hal {
...
     virtual bool process_ext_events() = 0;
 
+    /*!
+     * Process incoming nl events from the underlying hardware/middleware.
+     * This method should be called if the file descriptor returned by
+     * get_nl_events_fd() generated an event.
+     *
+     * @return true on success or false on error.
+     */
+    virtual bool process_nl_events() = 0;
```
arnout, line 102: If you add such a pure virtual function, you need to also add an implementation to each subclass (nl80211 and dummy in addition to dwpal).

However, it doesn't seem to be called anywhere...

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            LOG(ERROR) << "User's netlink handler function failed!";
+            return DWPAL_FAILURE;
+        }
+        return DWPAL_SUCCESS;
+    };
+    auto nl_handler_cb = [](struct nl_msg *msg) -> DWPAL_Ret { return nl_handler_cb_wrapper(msg); };
+
+    if (m_fd_nl_events > 0) {
+        //parsing will be done in callback function
+        if (dwpal_driver_nl_msg_get(m_dwpal_nl_ctx, DWPAL_NL_UNSOLICITED_EVENT, NULL,
```
arnout, line 578: All the code above seems to be quite a lot of complexity for essentially wrapping this one function call. Can't we just override process_nl_events() in the leaf class directly?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout, line 582: I don't think it's appropriate to detach just because we failed to parse an nl message...

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+{
+    bool err = true;
+    if ((vendor_data_size > 0) && (vendor_data == NULL)) {
+        LOG(ERROR) << __FUNCTION__ << "ifname or vendor_data is NULL ==> Abort!\n";
+        return false;
+    }
+
+    if (dwpal_driver_nl_cmd_send(m_dwpal_nl_ctx, DWPAL_NL_UNSOLICITED_EVENT, (char *)ifname.c_str(),
+                                 NL80211_CMD_VENDOR, DWPAL_NETDEV_ID,
+                                 (enum ltq_nl80211_vendor_subcmds)nl_cmd, vendor_data,
```
arnout, line 601: Why doesn't the nl_cmd argument have type enum ltq_nl80211_vendor_subcmds?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: I even wonder if the check is correct. Why wouldn't there be a vendor command with no data?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: I agree.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: It should actually be const_cast. c_str returns const char*, but dwpal is missing const in many places.

Better yet would be to fix dwpal, but I know that that's impossible :-)

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -39,6 +44,38 @@ namespace dwpal {
...
 
+// Helper macro for dwpal_nl_cmd_get
+// interfaceName, out_buffer, out_buffer_size, max_buffer_size, paramsObj
+#define DWPAL_GET_SCAN_PARAMS_HELPER_FG(IFNAME, OUT_BUFF, OUT_BUFF_SZ, MAX_BUFF_SZ, PRMS)          \
+    if (dwpal_nl_cmd_get(IFNAME, LTQ_NL80211_VENDOR_SUBCMD_GET_SCAN_PARAMS, OUT_BUFF,              \
+                         &OUT_BUFF_SZ, MAX_BUFF_SZ) != true) {                                     \
+        LOG(ERROR) << __func__ << " LTQ_NL80211_VENDOR_SUBCMD_GET_SCAN_PARAMS failed!";            \
+        return false;                                                                              \
+    }                                                                                              \
+    memcpy_s(&PRMS, sizeof(PRMS), &OUT_BUFF[NL_ATTR_HDR], sizeof(PRMS));
```
arnout, line 55: This is crazy - dwpal_nl_cmd_get already does a copy, and then immediately afterwards we do another copy... Just pass the params structure directly as argument to dwpal_nl_cmd_get.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout, line 591: Use uint8_t as data type, not unsigned char.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+                       << " ctx=" << m_dwpal_nl_ctx;
+            detach();
+            return false;
+        }
+    }
+
+    return true;
+}
+
+bool base_wlan_hal_dwpal::dwpal_nl_cmd_set(const std::string &ifname, unsigned int nl_cmd,
```
arnout, line 590: The name of this function is confusing with the dwpal functions. Just call it nl_cmd_set. Same for get.

Yes, I know there's already dwpal_send_cmd. No need to repeat that mistake :-)

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: Doing a select()/poll() on the appropriate fd should work.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -507,6 +552,154 @@ bool base_wlan_hal_dwpal::attach_ctrl_interface(int vap_id)
...
+            m_dwpal_nl_ctx, DWPAL_NL_SOLICITED_EVENT, (char *)ifname.c_str(), NL80211_CMD_VENDOR,
+            DWPAL_NETDEV_ID, (enum ltq_nl80211_vendor_subcmds)nl_cmd, NULL, 0) == DWPAL_FAILURE) {
+        LOG(DEBUG) << __FUNCTION__ << "ERROR for cmd = " << nl_cmd;
+        return false;
+    }
+
+    // Passing a lambda with capture is not supported for standard C function
+    // pointers. As a workaround, we create a static (but thread local) wrapper
+    // function that calls the capturing lambda function.
+    static __thread std::function<DWPAL_Ret(char *ifname, int event, int subevent, size_t len,
```
arnout, line 630: Why not use `DWPAL_nlEventCallback` instead of expanding the function signature?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: This pattern with the static capturing lambda variable is repeated a couple of times. AFAIU it is only needed because dwpal doesn't have a reentrant API. So maybe it's better to create a function that does this wrapping and that takes a C++ (instead of C) function as argument.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
arnout: These macros just obfuscate the control flow. No static inline functions, just keep the control flow explicit in the calling function.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -39,6 +44,38 @@ namespace dwpal {
...
+        LOG(ERROR) << __func__ << " LTQ_NL80211_VENDOR_SUBCMD_GET_SCAN_PARAMS_BG failed!";         \
+        return false;                                                                              \
+    }                                                                                              \
+    memcpy_s(&PRMS, sizeof(PRMS), &OUT_BUFF[NL_ATTR_HDR], sizeof(PRMS));
+
+// Helper macro for dwpal_nl_cmd_set
+// interfaceName, paramsObj, new_paramsObj, max_buffer_size, paramsObj
+#define DWPAL_SET_SCAN_PARAMS_HELPER_FG(IFNAME, PRMS, NEW_PRMS)                                    \
+    if (dwpal_nl_cmd_set(IFNAME, LTQ_NL80211_VENDOR_SUBCMD_SET_SCAN_PARAMS,                        \
+                         (unsigned char *)&NEW_PRMS, sizeof(PRMS)) != true) {                      \
```
arnout, line 68: This is very fishy, sizeof something else then the buffer argument...

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+    new_bg.active  = dwell_time_msec;
+
+    DWPAL_SET_SCAN_PARAMS_HELPER_FG(ifname, fg, new_fg);
+    DWPAL_GET_SCAN_PARAMS_HELPER_FG(ifname, out_data, out_len, NL_MAX_REPLY_BUFFSIZE, new_fg);
+
+    DWPAL_SET_SCAN_PARAMS_HELPER_BG(ifname, bg, new_bg);
+    DWPAL_GET_SCAN_PARAMS_HELPER_BG(ifname, out_data, out_len, NL_MAX_REPLY_BUFFSIZE, new_bg);
+
+    if ((new_fg.active != dwell_time_msec) && (new_fg.passive != dwell_time_msec) &&
+        (new_bg.active != dwell_time_msec) && (new_bg.passive != dwell_time_msec)) {
```
arnout, line 502: These should be || instead of &&

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+
+    if ((new_fg.active != fg.active) && (new_fg.passive != fg.passive) &&
+        (new_bg.active != bg.active) && (new_bg.passive != bg.passive)) {
+        LOG(ERROR) << __func__ << " failed to restore dwell time" << std::endl
+                   << "new fg: " << new_fg.passive << " " << new_fg.active << " "
+                   << new_fg.num_probe_reqs << " " << new_fg.probe_reqs_interval << " "
+                   << new_fg.passive_scan_valid_time << " " << new_fg.active_scan_valid_time
+                   << std::endl
+                   << "new bg: " << new_bg.passive << " " << new_bg.active << " "
+                   << new_bg.num_probe_reqs << " " << new_bg.probe_reqs_interval << " "
```
arnout, line 571: This is pretty much the identical code of above, so better factor that into a separate function.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+    dcs_scan_params.ap_force = 1;
+
+    if (dwpal_driver_nl_scan_trigger(get_dwpal_nl_ctx(), (char *)m_radio_info.iface_name.c_str(),
+                                     &dcs_scan_params) != DWPAL_SUCCESS) {
+        LOG(ERROR) << " scan trigger failed! Abort scan ";
+        return false;
+    }
+
+    // timeout for nl confirmation on scan trigger
+    sleep(1);
```
arnout, line 554: Err, that needs more explanation... If we're waiting for confirmation, please use the approprate polling and get calls to actually wait for confirmation.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -409,6 +466,131 @@ bool mon_wlan_hal_dwpal::sta_link_measurements_11k_request(const std::string &st
...
+{
+    LOG(DEBUG) << __func__ << " received on interface=" << m_radio_info.iface_name;
+
+    //build scan parameters
+    ScanParams dcs_scan_params;
+    memset((void *)&dcs_scan_params, 0, sizeof(dcs_scan_params)); //must to ensure freq are zero
+    size_t out_len                                = 0;
+    unsigned char out_data[NL_MAX_REPLY_BUFFSIZE] = {'\0'};
+    sScanCfgParams fg, new_fg;   //foreground scan param
+    sScanCfgParamsBG bg, new_bg; //background scan param
```
arnout, line 480: Instead of two copies, we can just use the one copy and separately remember the two values we override (nitpick).

arnout (changes requested):
I have only reviewed the first two commits, and only partly.

For the second commit, @tomereli proposed to split it into three parts, but I think in addition there should be a split of the new bwl interfaces (new events, dcs_scan_trigger, dcs_scan_results) from the addition of the NL functionality. And of course there should be a (dummy) implmentation of the new function for all platforms.

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -443,6 +446,27 @@ enum eBssType {
...
+enum eDcsScanErrCode : uint8_t {
+    DCS_SCAN_NO_ERROR = 0,
+    DCS_SCAN_INTERNAL_FAILURE,
+    DCS_SCAN_CHANNEL_POOL_TOO_BIG,
+    DCS_SCAN_SCAN_TRIGGERED_EVENT_TIMEOUT,
+    DCS_SCAN_RESULTS_READY_EVENT_TIMEOUT,
+    DCS_SCAN_RESULTS_DUMP_EVENT_TIMEOUT,
+    DCS_SCAN_ABORTED_BY_DRIVER,
+    DCS_SCAN_INVALID_SCAN_PARAMS
+};
```
arnout, line 458: I don't like the terminology here very much. DCS is not a term that is immediately obvious. In addition, this is about scanning which is not necessarily directly related to DCS (though I can't think of any other use right away). In addition, Multi-AP R2 has "channel scan" support, and I think that that is what we're doing here.  So instead of DCS scan I would like to call it channel scan.

common/beerocks/bwl/CMakeLists.txt
kantera800: no change required for this commit then.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,397 @@
...
+using namespace beerocks;
+using namespace net;
+using namespace son;
+
+#define FSM_MOVE_STATE(new_state)                                                                  \
+    ({                                                                                             \
+        TASK_LOG(TRACE) << "FSM: " << s_ar_states[int(m_fsm_state)] << " --> "                     \
+                        << s_ar_states[int(eState::new_state)];                                    \
+        m_fsm_state = eState::new_state;                                                           \
+    })
```
arnout, line 26: This could be a private member function instead of a macro.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+#define FSM_IS_IN_STATE(state) (m_fsm_state == eState::state)
+#define FSM_CURR_STATE_STR s_ar_states[int(m_fsm_state)]
+
+#define EVENT_STR(event) s_ar_events[int(event)]
```
arnout, line 31: These three macros only obfuscate the code IMO. @tomereli please confirm.

If they are useful, they should be private member functions instead.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,397 @@
...
+        if (m_is_single_scan_pending) {
+            m_is_single_scan         = true;
+            m_is_single_scan_pending = false;
+            FSM_MOVE_STATE(TRIGGER_SCAN);
+        } else if (database.get_dcs_is_enabled(m_radio_mac)) {
+            auto now = std::chrono::steady_clock::now();
+
+            if (now > m_next_scan_timestamp_interval) {
+                m_is_single_scan = false;
+                FSM_MOVE_STATE(TRIGGER_SCAN);
```
arnout, line 66: I prefer the FSM_MOVE_STATE at the end.

kantera800:
> I have only reviewed the first two commits, and only partly.
> 
> For the second commit, @tomereli proposed to split it into three parts, but I think in addition there should be a split of the new bwl interfaces (new events, dcs_scan_trigger, dcs_scan_results) from the addition of the NL functionality. And of course there should be a (dummy) implmentation of the new function for all platforms.

Concluding  from @arnout @tomereli  comments the split should be as follows:

1st PR:
First commit split - no change 
Second commit split -
bwl: Add NL support (with stubs)
bwl: Add NL DWPAL base implementation
bwl: Add NL DWPAL monitor implementation
bwl: Add NL DWPAL dcs scan interface
2nd PR:
rest of commits - @arnout @tomereli please review these commits.

Also, I think its better to change this PR to POC and open new PRs

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: do you mean to run KW from rdkb on prpl (off tree) ?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2117,6 +2117,322 @@ bool db::get_hostap_is_acs_enabled(std::string mac)
...
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid hostap!";
+        return false;
+    }
+    LOG(DEBUG) << __FUNCTION__ << "," << (single_scan ? "single" : "continuous")
+               << ", dwell time msec = " << dwell_time_msec;
+
+    if (dwell_time_msec <= 0) {
+        LOG(ERROR) << __FUNCTION__ << "Only >0 dwell time is supported!";
+        return false;
+    }
```
arnout, line 2278: This is silly... Just make dwell_time_msec a uint32_t and be done with it.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2117,6 +2117,322 @@ bool db::get_hostap_is_acs_enabled(std::string mac)
...
+        if (supported_channels.find(channel) == supported_channels.end()) {
+            //even if one channel is not supported reject all pool
+            LOG(ERROR) << __FUNCTION__ << "channel =" << int(channel)
+                       << " is invalid, setting channel pool failed !";
+            return false;
+        }
+    }
+    auto scan_config =
+        (single_scan) ? n->hostap->single_scan_config : n->hostap->continuous_scan_config;
+    scan_config->channel_pool = std::set<uint8_t>(channel_pool); // creates a copy of the pool
```
arnout, line 2335: This explicit conversion shouldn't be needed.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
kantera800: I can add "must as single wifi won't allow scan on ap without this flag" hope its clear. 
also can be changed to bool , will be much clearer.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: you are right .
also,  this  corner case is a hard catch in validation...

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: The reason that i initially thought of was just not to crash in case platform dosnt support NL , but then I see now that its not possible . 
On the other hand i dont see a point running with ctrl interfaces but without NL , so i think we should set attach = false in case of nl attach failure as well .
@arnout what do you think?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: no entirely true,
unless we add attach = false in case of nl attach failure we will reach this code.

Also, its a good  protection in case we decide not to support NL in some dwpal platforms.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
kantera800: Ported from wav-api.
Nothing to poll on .
We can add comment & comment it out for now, leave the decision for final PR validation stage.

arnout (commented):
> > I have only reviewed the first two commits, and only partly.
> > For the second commit, @tomereli proposed to split it into three parts, but I think in addition there should be a split of the new bwl interfaces (new events, dcs_scan_trigger, dcs_scan_results) from the addition of the NL functionality. And of course there should be a (dummy) implmentation of the new function for all platforms.
> 
> Concluding from @arnout @tomereli comments the split should be as follows:
> 
> 1st PR:
> First commit split - no change
> Second commit split -
> bwl: Add NL support (with stubs)
> bwl: Add NL DWPAL base implementation
> bwl: Add NL DWPAL monitor implementation

:+1: 

Just to be clear, the "with stubs" part should include the stub implementation for all bwl types.

> bwl: Add NL DWPAL dcs scan interface

I'm not sure what @tomereli meant with this though...

> 2nd PR:
> rest of commits - @arnout @tomereli please review these commits.

I'll do an initial, light review.


> Also, I think its better to change this PR to POC and open new PRs

:+1:

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: That's exactly what I meant in https://github.com/prplfoundation/prplMesh/pull/619#discussion_r362031408

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
arnout: > no entirely true,
> unless we add attach = false in case of nl attach failure we will reach this code.

In the current code we go to Detach state on failure, so we can't get into this state.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
arnout: > Nothing to poll on .

There are two possibilities.

Either there is a netlink event that confirms the scan was triggered. Then you can do poll()/select() on the fd and get the event.

Or there is no such event but we still need to give the driver some time to register the changed scan parameters. (IOW, the driver is broken.) In that case we need the sleep and a comment that explains why (i.e., the first sentence of this paragraph but then formulated a bit better).

If a sleep is needed, please use C++ `std::this_thread::sleep_for`

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,397 @@
...
+            TASK_LOG(DEBUG) << "TRIGGER_SINGLE_SCAN received";
+            auto trigger_single_scan_event = reinterpret_cast<sTriggerSingleScan_event *>(obj);
+            handle_event                   = true;
+            TASK_LOG(DEBUG) << "TRIGGER_SINGLE_SCAN handled on:"
+                            << network_utils::mac_to_string(
+                                   trigger_single_scan_event->radio_mac.oct);
+            m_is_single_scan_pending = true;
+            break;
+        }
+        case eEvent::SCAN_TRIGGERED: {
```
arnout, line 211: Does it really make sense to have these SCAN_TRIGGERED, SCAN_RESULTS_READY and SCAN_FINISHED event with corresponding WAIT states? In the end, the only one that matters is the SCAN_RESULTS_DUMP event. We can just add a finished flag to it to indicate that it's finished, and an error flag to indicate that it has no dump results but the scan was stopped due to error.

common/beerocks/bwl/include/bwl/base_wlan_hal.h
kantera800: its an INFRA PR.
It will be called from monitor_thread.cpp in the next PR.

arnout (commented):
I have one additional remark. In R2, there is support for channel scan. So ideally, we should use those R2 messages already. However, we should probably merge the current implementation first because:

- the vendor-specific messages implementation exists already;
- it will still be quite some work to change it to standard R2 messages + vendor-specific extensions (which are needed e.g. to specify dwell time);
- this feature is blocking the "single code base" at Intel, and we really want that to be done as soon as possible.

arnout:
Still two more remarks... 

I would prefer not to add *any* feature if there is no test for it. #454 is a good example of how a feature like this could be implemented. Note that this requires a dummy bwl implementation that takes an event in addition to the dwpal implementation.

For the later commits, this time I think it makes sense to *squash* them.

2. CMDU & Structure support -> I think it's better to split this into (1) support definitions outside of tlvf (2) yaml modifications and (3) autogenerated.

3. Database + Node and 4. DCS task -> I think it makes more sense to do all of that together. For completely new stuff, like DCS task, splitting it into commits is not very useful. A sensible way of splitting into separate commits is going to be a lot of extra work, so I don't think it's worth the bother. It's only worth it if the feature hasn't been fully implemented yet, and you want to merge it partially already while still developing the final bits. For example, a sensible way of splitting could be:

- Add DCS task with only INIT state.
- Add BML to trigger DCS (includes additional stuff in database for scan parameters).
- Add test flow for DCS trigger (checks in controller log that DCS was triggered).
- Add scan trigger and scan result states to DCS, store results in database (includes the additional stuff in database for storing results).
- Add test flow for scan (checks in agent that trigger was received, gives scan results, checks with bml that results were stored in database).
- Add flow for dynamic channel selection.
- Add test for channel selection (checks that decision is as expected based on provided scan results).

Again: splitting like that after the fact is significant extra work, because you need to change the state machine in each step. So I don't think it's worth doing that now.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
tomereli: Yes

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: I believe that having two copies of sScanCfgParams & sScanCfgParamsBG, makes the code easier to read and understand, and besides that, if we want to change other parameters later on having the two copies would make it easier to do so

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: so, to conclude the changes we agree on:
1. remove the "attached = true;" from the else case.
2. add "attach = false" in case dwpal_driver_nl_attach() fails

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: can't use static, they use protected methods called "dwpal_nl_cmd_get" and "dwpal_nl_cmd_set"

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: Since there is no point to continue if NL is not supported  i suggest 
1. add attach = false in case of nl attach failure (see my [comment](https://github.com/prplfoundation/prplMesh/pull/619#discussion_r362424962))
2. leave this check as is .

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
kantera800: second option,  and i'll explain why .
In current implementation we are changing scan parameters before triggering the scan , so 
we have to restore them no matter what. 
There could be a lot of failure points until the next select that i the reason why "wait" was chosen.

arnout:
[Port dynamic channel selection](https://trello.com/c/65NWF2dD/42-port-dynamic-channel-selection)

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: changed the sleep to use the sleep_for format

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: separated into a static function

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: fixed

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: added the attach = false to the attach() fail scenario

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: removed

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal_types.h
itayx: changed

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
vitalybu: @arnout is right, but we don't actually check make sure that this variable is not null.
We only move to Detach if `dwpal_driver_nl_attach` returns an error...
Theoretically, since dwpal is an external library, the call may succeed by the variable will point to NULL.
We should simply move this check to the Init state.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
tomereli: You can define static inline private methods in the header.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
tomereli: Confirm what? Agree that private methods are probably nicer.

arnout: Confirm that the macros only obfuscate the code and that it's better to replace (taken from channel_selection_task):

```
            if (FSM_IS_IN_STATE(WAIT_FOR_ACS_RESPONSE)) {
```

with

```
            if (m_fsm_state == WAIT_FOR_ACS_RESPONSE) {
```

tomereli: In this case of course I confirm.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
kantera800: length is not provided, its an output parameter in order to know how much bytes reply_buffer contains.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: its ported, we can try to remove it at validation stage

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: dwpal callback  has to be in this format.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: currently it's only used in the dwpal monitor
but since this can be used by all libraries with dwpal we didn't want to restrict it to just be available in the monitor

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: There method also used in base_wlan_hal_nl80211.cpp .
If you want to refactor this code it should be done in different PR.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
kantera800: ported code from wav-api.
need more testing if you want to remove.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: It must have been a remainder from when the code was C and not C++
Changed it to = { 0 } since we still need to zero out all the values before usage

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: but since the cmd get & set are protected member methods they need an instance of the base_wlan_hal_dwpal object (in this case the local "this" is mon_wlan_hal_dwpal which inherits it) to use the two functions

so if we want to remove the macros we have to either:
1) make static functions that will receive an reference to base_wlan_hal_dwpal, and in the function call will add "this" as the handler
2) create local member methods that wrap the protected functions, they will use additional utility we gain from the macros

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: added it as a comment to explain why we set ap_force = 1

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: fixed

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: Added the "ifdef" to base_wlan_hal_dwpal and removed from ap_wlan_hal_dwpal

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: fixed

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: Fixed

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: you're right, changed to std::copy_n

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: removed, must have need left-overs from the porting process from C

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
itayx: Changed to a pure virtual

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: Added a switch-case

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: changed it and removed the err since it has no other use

common/beerocks/bwl/common/base_wlan_hal.cpp
itayx: Done

controller/src/beerocks/master/db/db.cpp
itayx: we want to force the user to set new values, therefore the default dwelltime is -1, an invalid value
We could change the set function signature to uint32_t but then the stored value, the get function and the set function wont have the same type

controller/src/beerocks/master/db/db.cpp
itayx: if you're talking about the _std::set<uint8_t>(channel_pool);_
It's no conversion, it invokes the set's copy constructor, and I know that std::set has a = operator function that copies the content, but I wanted to make sure it will invoke a copy function

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: We could, but I think that having the different events will help with the readability of the feature, since they are different event that have different meanings to them

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Moved

common/beerocks/bcl/include/bcl/beerocks_defines.h
itayx: Changed to channel_scan_... throughout the code

controller/src/beerocks/master/db/db.cpp
arnout: Ah now I see...

But that means that you always have to set the dwell time to some valid value first before you can do channel scan? That also seems weird... Why isn't there simply a sane default value for dwell time?

Anyway, nitpick.

controller/src/beerocks/master/db/db.cpp
arnout: Indeed, it's not conversion, it's copy constructor. But why do you want an explicit copy constructor? There is no reason for it, it's just overhead (which will normally be optimised away of course).

itayx: 100% overhead, I wanted to make sure that what I'm sending back is a copy as to not damage the original value. I can remove it if it's really unneeded

controller/src/beerocks/master/db/db.cpp
itayx: We plan in the next phase (after the feature is merged) to include a default configuration that will be persistent by using the beerocks_controller.conf file
currently as far as the user is concerned the continuous scan will require a once-per-load minimum configuration

PR #620, opened 2019-12-30
prplmesh_utils.sh:start_function - platform_init fix

adam1985d:
in the current implementation, the prplmesh_platform_db_init is called
even if the '$PLATFORM_INIT' flag is not set to 'true'.

after this change, the prplmesh_platform_db_init is performed only if
the '$PLATFORM_INIT' flag is set to 'true'.


tomereli (approved):

adam1985d:
it seems that docker tests don't have the support to run the script after these changes.

as these changes are not mandatory and this improvement can be introduced at a later time with docker tests changes

PR #623, opened 2019-12-31
Bugfix: replace addClass with getClass in UCC listener

### LINKED ISSUE #622: UCC listener - dev_send_1905 hangs
**Describe the bug** -
Hang when sending 1905 command by the UCC:

```
DEBUG 12:07:27:259 <139851804546816> beerocks_ucc_listener.cpp[605] --> Received WFA-CA, command="dev_send_1905,DestALid,46:8c:f8:4b:6f:89,MessageTypeValue,0x8006,tlv_type1,0x8B,tlv_length1,0x005F,tlv_value1,{0x00037f590641 0x14 {0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x52 {0x00 0x00}} {0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}} {0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x73 {0x00 0x00}} {0x74 {0x00 0x00}} {0x75 {0x00 0x00}} {0x76 {0x00 0x00}} {0x77 {0x00 0x00}} {0x78 {0x00 0x00}} {0x79 {0x00 0x00}} {0x7A {0x00 0x00}} {0x7B {0x00 0x00}} {0x7C {0x00 0x00}} {0x7D {0x00 0x00}} {0x7E {0x00 0x00}} {0x7F {0x00 0x00}} {0x80 {0x00 0x00}} {0x81 {0x00 0x00}} {0x82 {0x00 0x00}},tlv_type2,0x8D,tlv_length2,0x0007,tlv_value2,{0x00037f590641 0x14}"
DEBUG 12:07:27:260 <139851804546816> beerocks_ucc_listener.cpp[578] --> Replying message: 'status,RUNNING'
```
## Steps to reproduce

Run 5.4.1 test in the UCC or send the command manually with `nc`

CoralMalachi:
Currently, when trying to send a 1905 message of type
`CHANNEL_SELECTION_REQUEST_MESSAGE` or `COMBINED_INFRASTRUCTURE_METRICS_MESSAGE`,
the UCC listener replay status running and stack there.

The reason for this is the using of addClass in `beerocks_ucc_listener::create_cmdu`
was not replaced with the new `getClass` method as it should.

As a result, the UCC listener got stuck in an infinite loop, since the
parse of the TLVs already happened, which caused a timeout error.

In order to solve it, both while loops are removed and replaced with two
simple calls to `getClass` method.
Also, `tlv_type` is removed since it is not needed anymore.

Fixes #622 (checked on testbed)

### DIFF
--- a/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
+++ b/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
@@ -266,8 +266,6 @@ bool beerocks_ucc_listener::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx,
 {
     std::shared_ptr<ieee1905_1::cCmduHeader> cmdu_header;
 
-    int tlv_type;
-
     switch (message_type) {
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE: {
 
@@ -277,23 +275,18 @@ bool beerocks_ucc_listener::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx,
             return false;
         }
 
-        tlv_type = cmdu_tx.getNextTlvType();
-        while (tlv_type != int(ieee1905_1::eTlvType::TLV_END_OF_MESSAGE)) {
-            if (tlv_type == int(wfa_map::eTlvTypeMap::TLV_CHANNEL_PREFERENCE)) {
-                if (!cmdu_tx.addClass<wfa_map::tlvChannelPreference>()) {
-                    LOG(ERROR) << "addClass tlvChannelPreference has failed";
-                    err_string = err_internal;
-                    return false;
-                }
-            } else if (tlv_type == int(wfa_map::eTlvTypeMap::TLV_TRANSMIT_POWER_LIMIT)) {
-                if (!cmdu_tx.addClass<wfa_map::tlvTransmitPowerLimit>()) {
-                    LOG(ERROR) << "addClass tlvTransmitPowerLimit has failed";
-                    err_string = err_internal;
-                    return false;
-                }
-            }
-            tlv_type = cmdu_tx.getNextTlvType();
+        auto tlvChannelPreference = cmdu_tx.getClass<wfa_map::tlvChannelPreference>();
+        if (!tlvChannelPreference) {
+            LOG(ERROR) << "getClass<tlvChannelPreference> failed";
+            return false;
         }
+
+        auto tlvTransmitPowerLimit = cmdu_tx.getClass<wfa_map::tlvTransmitPowerLimit>();
+        if (!tlvTransmitPowerLimit) {
+            LOG(ERROR) << "getClass<tlvTransmitPowerLimit> failed";
+            return false;
+        }
+
         break;
     }
     case ieee1905_1::eMessageType::COMBINED_INFRASTRUCTURE_METRICS_MESSAGE: {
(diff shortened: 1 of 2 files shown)


tomereli (approved):
Let's also make addClass protected and add a wrapper for it only in CmduMessageTx. On RX it should never be used

CoralMalachi:
Please do not merge yet, I want to validate on the testbed first.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -277,23 +275,18 @@ bool beerocks_ucc_listener::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx,
...
+        auto tlvChannelPreference = cmdu_tx.getClass<wfa_map::tlvChannelPreference>();
+        if (!tlvChannelPreference) {
+            LOG(ERROR) << "getClass<tlvChannelPreference> failed";
+            return false;
         }
+
+        auto tlvTransmitPowerLimit = cmdu_tx.getClass<wfa_map::tlvTransmitPowerLimit>();
+        if (!tlvTransmitPowerLimit) {
+            LOG(ERROR) << "getClass<tlvTransmitPowerLimit> failed";
+            return false;
```
arnout, line 287: I don't think these checks are needed anymore. We only did it before to make sure the structures were parsed properly. Now the complete parsing is done as part of load(), so the addClass calls simply aren't needed. It's up to the functions that prepare the certification buffer to make sure that it contains the appropriate CMDUs.

So basically, I think that this entire function can be simplified to:

```
bool beerocks_ucc_listener::create_cmdu(ieee1905_1::CmduMessageTx &cmdu_tx,
                                        ieee1905_1::eMessageType message_type,
                                        std::string &err_string)
{
    switch (message_type) {
    case ieee1905_1::eMessageType::COMBINED_INFRASTRUCTURE_METRICS_MESSAGE: {

        auto cmdu_header = cmdu_tx.load();
        if (!cmdu_header) {
            LOG(ERROR) << "load cmdu has failed";
            err_string = err_internal;
            return false;
        }
        // Force mid
        cmdu_header->message_id() = g_mid;
        
        return true;
    }

    default:
        // If message type handler is not implemented, then return false with empty error, so the
        // caller would know that it was not handled.
        err_string = std::string();
        return false;
    }
}
```

arnout (changes requested):

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
CoralMalachi: Agree. I'll validate again the everything Ok on testbed before I'll merge it.

CoralMalachi: I think we should use `getClass` here.
The reason for this is that we call `create_cmdu` method at line 790 in `beerocks_ucc_listener.cpp` to check if the current  CMDU message has a preprepared buffer which should be loaded, otherwise, we need to create it manually and use TLV data from the command if exists.
if we remove those checks we'll never know in which situation are we.

arnout: I disagree. But since this whole certification buffer thing is a horrible hack anyway, let's not discuss it further.

arnout (approved):

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
morantr: @CoralMalachi, Arnout has suggested a correct implementation IMO. We used `AddClass` before, to add the classes into the CMDU vector so it will be swapped on `finalize()`. Now that the data on the buffer is always swapped there is no need to do it anymore, so getting the classes is useless now. The only thing that needs to be done is to modify the `MID` on the CMDU header.

If I'm not mistaken, the data on the CMDU header we get is already swapped, so I think that we need to override its value with swapped value:
 `cmdu_header->message_id() = swap_16(g_mid);`

morantr (changes requested):

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
tomereli: One important thing to note is that any `getClass` (except for cmdu header) will return `nullptr` after `cmdu_tx.load()` - so these checks are definitely not needed since they will always result with the same outcome.

CoralMalachi: As @tomereli said since the `getClass` always returns `nullptr` we pass the test by accident. 
The `CHANNEL_SELECTION_REQUEST_MESSAGE` CMDU should be loaded from the certification buffer, but instead, it is created manually and use TLVs from the command (which do not exist).

To reveal this kind of bug, when a CMDU is not created from the certification buffer when it should be, update the test_flows to validate it in the last commit.

PR #624, opened 2019-12-31
Hotfix/remove add class from cmdu message

tomereli:
Prevent cases where parse is called more than once by making parse a CmduMessageTx method instead of CmduMessage. Also, make ClassList msg a protected member.

### DIFF
--- a/common/beerocks/tlvf/src/src/beerocks_message.cpp
+++ b/common/beerocks/tlvf/src/src/beerocks_message.cpp
@@ -107,7 +107,7 @@ std::string message_com::print_cmdu_types(ieee1905_1::CmduMessageRx &cmdu_rx, sC
         cmdu_info->cmdu_type = message_type;
     info = "cmdu_type=" + std::to_string(int(message_type));
     if (message_type == ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE) {
-        auto tlv_header = cmdu_rx.addClass<ieee1905_1::tlvVendorSpecific>();
+        auto tlv_header = cmdu_rx.getClass<ieee1905_1::tlvVendorSpecific>();
         if (!intel_oui(tlv_header))
             return std::string();
         if (!tlv_header) {
@@ -120,7 +120,7 @@ std::string message_com::print_cmdu_types(ieee1905_1::CmduMessageRx &cmdu_rx, sC
             LOG(ERROR) << "addClass<cACTION_HEADER> failed!";
             return info;
         }
-
+        tlv_header->addInnerClassList(hdr); // so it will also be swapped back
         if (cmdu_info) {
             cmdu_info->intel_vs_action    = hdr->action();
             cmdu_info->intel_vs_action_op = hdr->action_op();
--- a/framework/platform/bpl/linux/prplmesh_platform_db
+++ b/framework/platform/bpl/linux/prplmesh_platform_db
@@ -1,4 +1,4 @@
 management_mode=Multi-AP-Controller-and-Agent
 operating_mode=Gateway
-stop_on_failure_attempts=0
+stop_on_failure_attempts=1
 certification_mode=1
--- a/framework/tlvf/src/include/tlvf/CmduMessage.h
+++ b/framework/tlvf/src/include/tlvf/CmduMessage.h
@@ -43,7 +43,6 @@ public:
 
     // Forward wrapper functions
     // TODO check which of them can be removed
-    template <class T> std::shared_ptr<T> addClass() { return msg.addClass<T>(); }
     template <class T> std::shared_ptr<T> getClass() const { return msg.getClass<T>(); };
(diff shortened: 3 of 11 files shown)


CoralMalachi:
I don't sure I fully understand the commit message of the second commit.
Why did you say it will prevent "accidental usage of ClassList::getClass for parsed CMDUs"?
`getClass` method is not for parsed CMDU? did you mean to say for not parsed CMDUs?

CoralMalachi (commented):
I don't fully understand the commit message of the second commit.
Why did you say: "To prevent accidental usage of ClassList::getClass for parsed CMDUs", is not the `getClass` method is for parsed CMDUs ? or did you mean to say for not parsed CMDUs?

common/beerocks/tlvf/src/src/beerocks_message.cpp
```diff
@@ -120,7 +120,7 @@ std::string message_com::print_cmdu_types(ieee1905_1::CmduMessageRx &cmdu_rx, sC
             LOG(ERROR) << "addClass<cACTION_HEADER> failed!";
             return info;
         }
-
+        tlv_header->addInnerClassList(hdr); // so it will also be swapped back
```
arnout, line 123: This change isn't explained in the commit message.

framework/tlvf/src/include/tlvf/CmduMessage.h
```diff
@@ -44,16 +42,24 @@ class CmduMessage {
...
     // TODO check which of them can be removed
     template <class T> std::shared_ptr<T> getClass() const { return msg.getClass<T>(); };
+    template <class T> std::list<std::shared_ptr<T>> getClassList() const
+    {
+        return msg.getClassList<T>();
+    };
     size_t getMessageLength() const { return msg.getMessageLength(); };
     size_t getMessageBuffLength() const { return msg.getMessageBuffLength(); };
     uint8_t *getMessageBuff() const { return msg.getMessageBuff(); };
+    template <class T> std::shared_ptr<T> getClass() { return msg.getClass<T>(); }
```
arnout, line 52: There is already a const getClass version a bit higher up, why do we also need a non-const one?

arnout (approved):
As noted by @CoralMalachi, the commit message of the second commit is bogus. Suggested improvement:

```
To prevent accidental usage of ClassList::addClass for parsed CMDUs,
make the ClassList member protected instead of public.

Add a wrapper for getClassList<T> which was up to now accesses through
the ClassList member.

```

framework/tlvf/src/include/tlvf/CmduMessage.h
tomereli: we don't, nice catch.

tools/docker/tests/test_flows.sh
```diff
@@ -544,10 +544,13 @@ test_topology() {
     return $check_error
 }
 test_init() {
-    status "test initialization"    
-    eval ${scriptdir}/test_gw_repeater.sh -f -r "repeater1" -r "repeater2" -d 7 $redirect || {
-        err "start GW+Repeater failed, abort"
-        exit 1
+    status "test initialization"
+
+    [ "$1" = "false" ] && {
```
arnout: It would make more sense to use global SKIP_INIT here like we do for e.g. verbose. (nitpick)

framework/platform/bpl/linux/prplmesh_platform_db
```diff
@@ -1,4 +1,4 @@
 management_mode=Multi-AP-Controller-and-Agent
 operating_mode=Gateway
-stop_on_failure_attempts=0
+stop_on_failure_attempts=1
```
arnout, line 3: I'm not convinced... We also want a test_flows for agent certification, and that will have to do dev_reset_default...

arnout (approved):
I think "bpl: linux: set stop_on_failure=1 in dummy mode" should be dropped. For debugging, you can just set it to 0 in the install dir before starting test_flows.sh.

framework/platform/bpl/linux/prplmesh_platform_db
tomereli: we don't decrement stop_on_failure_attempts when backhaul disconnects so setting stop_on_failure=1 is fine and will not hurt dev_reset_default.

See line 1323 in son_slave_thread.cpp - we call slave_reset() without decrementing stop on failure count so it is not counted as failure.

tomereli:
> I think "bpl: linux: set stop_on_failure=1 in dummy mode" should be dropped. For debugging, you can just set it to 0 in the install dir before starting test_flows.sh.

See https://github.com/prplfoundation/prplMesh/pull/624#discussion_r362511205

tools/docker/tests/test_flows.sh
tomereli: Done.

PR #629, opened 2020-01-03
test_flows.sh: test_client_steering_policy: check for any ack with mid

rmelotte:
The test_client_steering_policy test first sends a steering policy to
the gateway, saves a mid, then later compare it to the mid of the
latest ACK_MESSAGE received by the controller.

However, it can happen that the ACK is present, but is not the latest
one. An example of such a situation is if the client gets unblocked
from the network, and son_master_thread receives an ACK_MESSAGE for
allow requests before test_flows.sh could check the ACK for the
steering.

Instead of comparing only the last mid, check if the controller
received any ACK_MESSAGE with the corresponding mid.


tomereli (approved):

rmelotte:
Rebased on master.

arnout (approved):

PR #630, opened 2020-01-03
.gitlab-ci.yml: split off in more separate jobs

arnout:
The main goal is to minimize the amount of work done in the run-tests job.


.gitlab-ci.yml
```diff
@@ -5,15 +5,21 @@
...
     - docker
 
+clang-format:
+  extends: .in-prplmesh-builder
+  stage: build
+  script:
+    # Run clang-format and check there are no modified files by checking git status
+    - ./clang-format.sh
+    - git update-index -q --refresh
+    - git diff-index -p --exit-code HEAD
```
tomereli, line 18: Does this create the patch? If so, what is the patch name? where will it be stored? I think it's worth adding it to the above comment.

framework/tlvf/test/tlvf_test.cpp
```diff
@@ -754,7 +754,13 @@ int test_all()
 
 int main(int argc, char *argv[])
 {
-    mapf::Logger::Instance().LoggerInit("tlvf_test");
```
tomereli, line 757: Is it possible to add the below to `LoggerInit()` as an optional argument? It can be useful for other unit tests so it will save us from duplicating this code.

tomereli (approved):
LGTM.

framework/tlvf/test/tlvf_test.cpp
arnout: I instead gave Logger a default config that logs to stdout, in a separate commit.

PR #631, opened 2020-01-05
Feature/remove network before scan

morantr:
Before BH-scan starts, the repeater should remove existing network configuration so STAs will not be busy in background scans.
Add removal of all existing networks before start scanning and thus make the station to be free from background scans.

### DIFF
--- a/common/beerocks/bwl/dummy/sta_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/sta_wlan_hal_dummy.cpp
@@ -40,7 +40,7 @@ bool sta_wlan_hal_dummy::connect(const std::string &ssid, const std::string &pas
     return true;
 }
 
-bool sta_wlan_hal_dummy::disconnect() { return true; }
+bool sta_wlan_hal_dummy::disconnect(bool remove_all_networks) { return true; }
 
 bool sta_wlan_hal_dummy::roam(const std::string &bssid, uint8_t channel) { return true; }
 
--- a/common/beerocks/bwl/dummy/sta_wlan_hal_dummy.h
+++ b/common/beerocks/bwl/dummy/sta_wlan_hal_dummy.h
@@ -41,7 +41,7 @@ public:
                          bool mem_only_psk, const std::string &bssid, uint8_t channel,
                          bool hidden_ssid) override;
 
-    virtual bool disconnect() override;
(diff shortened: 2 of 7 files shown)


arnout (commented):
Looks good to me at a low level. 

However, do we expect the same to be done for all bwl implementations?

If yes, then I think the disconnect(all) should be done in the backhaul manager itself, not in the dwpal bwl implementation.

If it is only for dwpal (which I think it is), then I think it's better to only implement this for dwpal. In other words, in sta_wlan_hal_dwpal.h:

```
bool disconnect() override { return disconnect(false); }
bool disconnect(bool all_networks);
```

morantr:
> Looks good to me at a low level.
> 
> However, do we expect the same to be done for all bwl implementations?
> 
> If yes, then I think the disconnect(all) should be done in the backhaul manager itself, not in the dwpal bwl implementation.
> 
> If it is only for dwpal (which I think it is), then I think it's better to only implement this for dwpal. In other words, in sta_wlan_hal_dwpal.h:
> 
> ```
> bool disconnect() override { return disconnect(false); }
> bool disconnect(bool all_networks);
> ```

It should be done all bwl implementations as long it uses the wpa_supplicant - which they are.
But I prefer it will be on the bwl since the backhaul manager shouldn't know about it. It's complicating it unnecessarily. We already do similar things with other bwl api, calling other bwl apis.

arnout (approved):
When it's ready for merge, please rebase on master right away so we can go through with the merge.

tomereli:
@morantr what's the status on this one?

morantr:
> @morantr what's the status on this one?

Still not tested.

ghost:
@morantr Is there any progress on this issue?

If not, could you explain what is blocking you and how we might be able to move forward?

morantr:
> @morantr Is there any progress on this issue?
> 
> If not, could you explain what is blocking you and how we might be able to move forward?

Currently, this task is on hold since I'm focusing my efforts on building rdkb with prplmesh.

tomereli:
@arnout Can we delete this PR and branch now?

arnout:
> @arnout Can we delete this PR and branch now?

Why are you asking me? This is @morantr's PR...

vitalybu:
@morantr, can we close this PR?

morantr:
> @morantr, can we close this PR?

Why are you asking me? This is @kantera800's decision...

PR #632, opened 2020-01-05
Bugfix/cmdu tx load fix

CoralMalachi:
The CmduMessageTx::load() API is used only for certification mode for
sending a prefilled CMDU when instructed to by the testbed UCC.

It got broken somewhere when we moved to ClassList and the current
implementation is now broken since load is only adding the first class
(CmduHeader) then on Finalize EOM TLV is added *in the middle of the
CMDU* overriding the actual buffer contents.

The solution is simple - since the prefilled buffer is ready to be sent
to the network and is only missing the EOM TLV, we change the load()
function to not only add the CMDU header (in parsing mode) but to also
add an Unknown TLV for the rest of the data.

This serves 2 purposes - the first is that the m_buff_ptr__ is updated
to point to the correct location in the buffer, which is the end of the
buffer. The second is that the last TLV is now an unknown length list
therefore is not swapped on finalize - which is what we want since all
the data is already swapped.

While we're at it - also refactor the certification buffer APIs in the master database to make it simpler to understand and use by adding only 2 APIs - set and get. Set allocates the buffer and presets it with a CmduMessageTx, and get returns the certification buffer shared_ptr.

This way, we can get rid of the if-else-if in the ucc listener and simple test for the certification buffer and send it as is, removing the need for create_cmdu API which is also removed.


controller/src/beerocks/master/db/db.h
```diff
@@ -510,18 +510,18 @@ class db {
...
-                                             std::default_delete<uint8_t[]>()));
-    };
-    void remove_certification_tx_buffer() { certification_tx_buffer.reset(); };
-    void fill_certification_tx_buffer(ieee1905_1::CmduMessageTx &cmdu_tx)
+    void clear_certification_tx_buffer() { certification_tx_buffer = nullptr; };
+    bool set_certification_tx_buffer(ieee1905_1::CmduMessageTx &cmdu_tx)
     {
+        certification_tx_buffer =
+            std::shared_ptr<uint8_t>(new uint8_t[beerocks::message::MESSAGE_BUFFER_LENGTH],
+                                     std::default_delete<uint8_t[]>());
```
morantr, line 519: Adding a check that checks if it is already allocated before allocating will be a good idea.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1329,13 +1316,7 @@ bool master_thread::construct_combined_infra_metric()
...
     }
 
-    auto certification_tx_buffer = database.get_certification_tx_buffer();
-    if (!certification_tx_buffer) {
-        LOG(ERROR) << "certification_tx_buffer is not allocated!";
-        return false;
-    }
-    database.fill_certification_tx_buffer(cmdu_tx);
-    return true;
+    return database.set_certification_tx_buffer(cmdu_tx);
```
morantr, line 1319: And if we are not in certification mode?

controller/src/beerocks/master/controller_ucc_listener.cpp
```diff
@@ -68,6 +68,7 @@ std::shared_ptr<uint8_t> controller_ucc_listener::get_buffer_filled_with_cmdu()
 bool controller_ucc_listener::send_cmdu_to_destination(ieee1905_1::CmduMessageTx &cmdu_tx,
                                                        const std::string &dest_mac)
 {
+    m_database.clear_certification_tx_buffer();
```
morantr: At first, I thought this call is putting zeros on the buffer, but now I see that the point of it is to free the buffer.
"free" is more correct word to describe what you are doing here.
But still, I don't understand the point of freeing the buffer after each use so you will need to allocate it again next time.
I would use it only when it is disabled by the bml or something.

controller/src/beerocks/master/db/db.h
```diff
@@ -510,18 +510,18 @@ class db {
...
+        if (!certification_tx_buffer)
+            return false;
         std::copy_n(cmdu_tx.getMessageBuff() - sizeof(beerocks::message::sUdsHeader),
                     cmdu_tx.getMessageBuffLength() + sizeof(beerocks::message::sUdsHeader),
                     certification_tx_buffer.get());
+        return true;
```
morantr, line 525: I would fill 4 more extra bytes with zeros just to make sure it has EOM.

controller/src/beerocks/master/db/db.h
```diff
@@ -510,18 +510,18 @@ class db {
...
-    std::shared_ptr<uint8_t> allocate_certification_tx_buffer()
-    {
-        return (certification_tx_buffer =
-                    std::shared_ptr<uint8_t>(new uint8_t[beerocks::message::MESSAGE_BUFFER_LENGTH],
-                                             std::default_delete<uint8_t[]>()));
-    };
-    void remove_certification_tx_buffer() { certification_tx_buffer.reset(); };
-    void fill_certification_tx_buffer(ieee1905_1::CmduMessageTx &cmdu_tx)
+    void clear_certification_tx_buffer() { certification_tx_buffer = nullptr; };
+    bool set_certification_tx_buffer(ieee1905_1::CmduMessageTx &cmdu_tx)
```
morantr, line 514: On the commit message, you mention that certification support only 2 APIs, but it is actually 3: clear, set and get.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -33,7 +34,9 @@ std::shared_ptr<cCmduHeader> CmduMessageTx::create(uint16_t id, eMessageType mes
 std::shared_ptr<cCmduHeader> CmduMessageTx::load()
 {
     msg.reset(true);
-    return addClass<cCmduHeader>();
+    auto cmdu_header = msg.addClass<cCmduHeader>();
+    msg.addClass<ieee1905_1::tlvUnknown>();
```
morantr: But what happens if it has more than 1 TLV? I don't think it will work.
I guess it will work for the specific CMDUs that we are using the certification buffer for, but maybe adding it in a loop until getting EOM might be a more correct solution. Of course, we will have to put 4 more zeros on the buffer, right after the CMDU we copied.

But generally, it is a GOOD solution and I like it. 😏

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -790,30 +691,27 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
...
-        if (!certification_tx_buffer) {
-            LOG(ERROR) << "certification_tx_buffer is not allocated";
-            reply_ucc(eWfaCaStatus::ERROR, err_internal);
-            break;
-        }
-        ieee1905_1::CmduMessageTx external_cmdu_tx(
-            certification_tx_buffer.get() + sizeof(beerocks::message::sUdsHeader),
-            message::MESSAGE_BUFFER_LENGTH - sizeof(beerocks::message::sUdsHeader));
-        if (create_cmdu(external_cmdu_tx, ieee1905_1::eMessageType(message_type), err_string)) {
+        if (certification_tx_buffer) {
```
morantr: Ok, now I understand why you have decided to remove the buffer after each use. Still, I think a better idea (instead of deallocating the buffer) will be to add a flag like "read_ready" that will be set to true after the buffer is filled, and that being cleared after the buffer is read by 
`get_buffer_filled_with_cmdu()`. If calling to `get_buffer_filled_with_cmdu()` when the "read_ready" is cleared, than return a nullptr.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1668,7 +1668,7 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
         LOG(DEBUG) << "ACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION received from iface "
                    << soc->hostap_iface;
         if (m_agent_ucc_listener) {
-            auto msg = cmdu_rx.getClass<
+            auto msg = beerocks_header->addClass<
```
morantr: This bug has slipped from my eyes, so I wonder if there is any way to fail while using the wrong API?

morantr (changes requested):

controller/src/beerocks/master/db/db.h
tomereli: If it is already allocated then we want to delete the old buffer and allocate a new one. So this does exactly that.

controller/src/beerocks/master/controller_ucc_listener.cpp
tomereli: This is on purpose, we want to free the previous certification buffer when we are done with it so we don't accidently think the following command should use it. Note that now we check if there is a preallocated buffer to send, and if we have one, we send it. So once we do, we must delete the old one.

controller/src/beerocks/master/db/db.h
tomereli: Will fix the commit message

controller/src/beerocks/master/db/db.h
tomereli: Not needed, finalize explicitely adds EOM TLV.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: This whole function is not called if we are not in certification mode.

framework/tlvf/src/src/CmduMessageTx.cpp
tomereli: That's the beauty of this solution - we don't care how many TLVs the CMDU contains.
Assume the buffer contents are a CMDU message with 3 TLVs: `[CMDU HEADER][TLV1][TLV2][TLV3]`
Then `load()` will treat `TLV1..TLV3` as one big list of unknown length without changing the buffer contents which is exactly what we want.
What it will do is to update `m_buff_ptr__` so on `finalize()` it will add the EOM TLV exactly at the end of the buffer, which is again, what we want.
And lastly, since it is an unknown length list of uint8_t, finalize will not swap the message contents which is again what we want since they are already swapped.
So bottom line, this covers all cases.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
tomereli: You are again paying more attention to performance than to code readability - in this case there is absolutly no reason to do that. The code is simpler to understand when certification buffer is allocated and used only for one message then freed. Performance is not an issue here, especially that this happens only in certification.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: We need to convert beerocks_header like I did the AttrList and make addClass protected so it can't be used from outside. This will uncover all bugs, but should be done in a separate PR.

controller/src/beerocks/master/controller_ucc_listener.cpp
morantr: change name to "free"

morantr: change name to "free"

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: After I looked on the init function of `tlvUnknown` it seems to me that it will only add one tlv.

tomereli: You are right, we need to add skipClass or something like that instead of unknown tlv.

tomereli: I removed this commit cause it was wrong - great catch, it worked by sheer luck!
I added 2 new commits which allow us to do this correctly and being able to verify the CMDU since getClass will now work.

arnout (commented):
IMO it would make sense to instead subclass CmduMessageTx to a message that contains the buffer (i.e. buffer is allocated by constructor), and use a shared_ptr to that everywhere. So the full message (both buffer and its wrapping class) is stored in the database, both are allocated and freed together. In the UCC handlers, you would not load() the cmdu_tx but just get it from the database.

This is a pretty big change though and I'm not sure if it's worth pursuing it.
to pass a shared_ptr to it here. In other words, where you set the certification buffer, you construct the full CmduMessageTx, but don't send it.

That way, there is no longer any need for load().

controller/src/beerocks/master/controller_ucc_listener.cpp
```diff
@@ -68,6 +68,7 @@ std::shared_ptr<uint8_t> controller_ucc_listener::get_buffer_filled_with_cmdu()
 bool controller_ucc_listener::send_cmdu_to_destination(ieee1905_1::CmduMessageTx &cmdu_tx,
                                                        const std::string &dest_mac)
 {
+    m_database.free_certification_tx_buffer();
```
arnout, line 88: I may be wrong, but: this should be the last shared_ptr reference to the certification buffer, so it will be freed. However, the cmdu_tx is still referring to it, just not as a shared_ptr. So cmdu_tx will be pointing to freed memory...

Anyway, I think it's the wrong thing to do. The certification buffer is filled when some specific CMDU arrives in the controller. In principle, this is more or less asynchronous to what the UCC is doing. So it's possible that the UCC first does some other dev_send_1905 before it does the dev_send_1905 which should use the certification buffer. So I think the free should be done only in the cases that we were using the certification buffer.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: Meh, if we ever do autoparsing of the Intel VS TLV, then we have to change this into getClass again...

Anyway, it clearly has to be on beerocks_header and not on cmdu_rx, so it's a step in the right direction (not to mention that it fixes the current issue :-)

arnout: It won't uncover anything... The issue that we had here is that we're calling cmdu_rx.getClass<...> with a class that is not a TLV.

If we want to detect this, we would need to check that the class is really a 1905 TLV, using that trick that @Galvien added to verify that a template instantiation is of the correct class. Small showstopper for that: there is currently no base class for 1905 TLV...

arnout (changes requested):
I don't really like this approach of parsing again the CmduMessageTx. I think it's hard to understand and likely to lead to mistakes.

But then, I don't like the certification buffer to begin with...

My preference would be to just remove the certification buffer entirely and replace it with a trigger that builds and sends the CMDU on demand. Since the ucc listener takes a lock on the db in after_select, there's really nothing stopping us from doing that.

A second option is to reimplement the certification buffer as a single shared_ptr that contains the fully constructed CmduMessageTx including the buffer and that is sent and freed as a whole. Then no load() is needed.

That said, I won't oppose merging this PR as it is either, since I think we shouldn't spend too much time on this certification stuff. However, the use-after-free must be solved.

morantr:
It came to my mind that the simplest thing we can do is to initially build the cmdu on a different cmdu_tx object (certification_cmdu_tx), So on the ucc_listener we just get it and send it as is.
With this, we could drop all the loading and everything...

morantr (approved):

tomereli:
> IMO it would make sense to instead subclass CmduMessageTx to a message that contains the buffer (i.e. buffer is allocated by constructor), and use a shared_ptr to that everywhere. 

This is not possbile AFAIK since we can't initialize the buffer before we pass it to the base class constructor:

```
C++11, 12.6.2
10 In a non-delegating constructor, initialization proceeds in the following order:

First, and only for the constructor of the most derived class (1.8), virtual base classes are initialized in the order they appear on a depth-ﬁrst left-to-right traversal of the directed acyclic graph of base classes, where "left-to-right" is the order of appearance of the base classes in the derived class base-speciﬁer-list.

Then, direct base classes are initialized in declaration order as they appear in the base-speciﬁer-list (regardless of the order of the mem-initializers).

Then, non-static data members are initialized in the order they were declared in the class deﬁnition (again regardless of the order of the mem-initializers).

Finally, the compound-statement of the constructor body is executed.
```
We can add another constructor to `CmduMessage` which will only get the buffer length as an argument and will create the buffer using `shared_ptr`, then delete this constructor in `CmduMessageRx` class since it should never be used for RX. What do you think?

tomereli:
> It came to my mind that the simplest thing we can do is to initially build the cmdu on a different cmdu_tx object (certification_cmdu_tx), So on the ucc_listener we just get it and send it as is.
> With this, we could drop all the loading and everything...

I like this and this is what @arnout is going to in his solution - just you suggest to duplicate cmdu_tx and tx_buffer to `certificaiton_cmdu_tx` and `certification_tx_buffer` which will not require any modifications to the base classes.

However I think a constructor for CmduMessageTx which accepts only a buffer size and internally allocates the buffer is a good idea anyway, so I'm still not sure which is the better approach.

arnout:
It's a bit of a hack, but instead of a shared_ptr, you could go for explicit free in the destructor, and access it through getMessageBuff(). So:

```
CmduMessageTxWithBuf(size_t buf_size) : CmduMessageTx(new uint8_t[buf_size], buf_size) {}
~CmduMessageTxWithBuf() { delete getMessageBuff(); }
```

Not great because the buffer gets deleted before the ClassList is completely destroyed, but OK-ish.

I think @morantr's proposal is to do the same for the certification buffer as for the other cmdu_tx instances: instantiate only once (both buffer and CmduMessageTx object) and reset it after sending. That actually sounds a lot more sensible than my proposal.

I also like the idea of an allocating constructor in CmduMessage itself. Note that you don't need to delete it from Rx, you only need to delete a copy constructor or default constructor. The allocating constructor can actually be used by all CmduMessageTx instances.

tomereli:
Superceeded by #651

PR #636, opened 2020-01-05
Feature/dev get param

### LINKED ISSUE #336: [TASK] Support wfa_ca in agent

tomereli:
For controller certification, dev_get_parameter is always ALid which is why it is hard-coded in prplMesh.
When the request is for the agent, it can either be ALid or macaddr, so this needs to be fixed.

When the parameter is macaddr, it includes the ruid and ssid, so the UCC agent listener should fetch the information from the VAPs configuration MAP - see attached card.

This PR adds generic dev_get_parameter support to both the controller and agent, with identical implementation for the case where `parameter=ALid`, and implements the case where `parameter=macaddr` in the agent ucc listener.

In addition, this PR fixes several minor bugs preventing the VAPs to be updated in the UCC agent, and finally adds a test in `test_flows.sh` which will make sure that `dev_get_parameter` will not be broken easily in the future by using and checking it in ap_config_renew test.

Fixes #336

### DIFF
--- a/agent/src/beerocks/slave/agent_ucc_listener.cpp
+++ b/agent/src/beerocks/slave/agent_ucc_listener.cpp
@@ -97,6 +97,22 @@ void agent_ucc_listener::update_vaps_list(std::string ruid, beerocks_message::sV
 }
 
 /**
+ * @brief get parameter command
+ *
+ * get agent parameter
+ *
+ * @param[in] params command parsed parameter map
+ * @param[out] value returned parameter or error on failure
+ * @return true on success
+ * @return false on failure
+ */
+bool agent_ucc_listener::handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
+                                              std::string &value)
+{
+    value = "parameter " + params["parameter"] + " not supported";
+    return false;
+}
+/**
  * @brief Validate if the 'dest_alid` MAC address matches the controllers MAC address.
  * 
  * @param[in] dest_alid Agent bridge MAC address.
--- a/agent/src/beerocks/slave/agent_ucc_listener.cpp
+++ b/agent/src/beerocks/slave/agent_ucc_listener.cpp
@@ -109,6 +109,32 @@ void agent_ucc_listener::update_vaps_list(std::string ruid, beerocks_message::sV
 bool agent_ucc_listener::handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
                                               std::string &value)
 {
+    if (params["parameter"] == "macaddr") {
+        if (params.find("ruid") == params.end()) {
+            value = "missing ruid";
+            return false;
+        }
+        if (params.find("ssid") == params.end()) {
+            value = "missing ssid";
+            return false;
+        }
+        auto ruid = network_utils::mac_to_string(std::stoull(params["ruid"], nullptr, 16));
+        auto ssid = params["ssid"];
+
+        auto it = vaps_map.find(ruid);
+        if (it == vaps_map.end()) {
+            value = "ruid " + ruid + " not found";
+            return false;
+        }
+        for (const auto &vap : it->second.vaps) {
+            if (std::string(vap.ssid) == ssid) {
+                value = network_utils::mac_to_string(vap.mac);
+                return true;
+            }
+        }
+        value = "macaddr not found for ruid " + ruid + " ssid " + ssid;
+        return false;
+    }
     value = "parameter " + params["parameter"] + " not supported";
     return false;
 }
--- a/common/beerocks/bcl/include/bcl/beerocks_ucc_listener.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_ucc_listener.h
@@ -46,6 +46,8 @@ protected:
                                           const std::string &dest_mac = std::string()) = 0;
     virtual bool handle_dev_set_config(std::unordered_map<std::string, std::string> &params,
                                        std::string &err_string)                        = 0;
+    virtual bool handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
+                                      std::string &value)                              = 0;
 
     enum class eUccListenerRunOn : uint8_t {
         CONTROLLER,
--- a/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
+++ b/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
@@ -651,21 +651,32 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
         break;
     }
     case eWfaCaCommand::DEV_GET_PARAMETER: {
-
-        // For controller certification, the following command is received to get the ALid:
-        // DUT (127.0.0.1:5000) ---> dev_get_parameter,program,map,parameter,ALid
-
+        std::unordered_map<std::string, std::string> params{{"parameter", ""}};
+        if (!parse_params(cmd_tokens_vec, params, err_string)) {
+            LOG(ERROR) << err_string;
+            reply_ucc(eWfaCaStatus::INVALID, err_string);
+            break;
+        }
         if (!reply_ucc(eWfaCaStatus::RUNNING)) {
             LOG(ERROR) << "failed to send reply";
             break;
         }
-
-        std::string alid;
-        if (!net::network_utils::linux_iface_get_mac("br-lan", alid)) {
-            reply_ucc(eWfaCaStatus::INVALID, "FAIL");
+        auto parameter = params["parameter"];
+        std::transform(parameter.begin(), parameter.end(), parameter.begin(), ::tolower);
+        std::string value;
+        if (parameter == "alid") {
+            if (!net::network_utils::linux_iface_get_mac("br-lan", value)) {
+                LOG(ERROR) << "failed to get br-lan mac address";
+                reply_ucc(eWfaCaStatus::ERROR, "failed to get br-lan mac address");
+                break;
+            }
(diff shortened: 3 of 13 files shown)


tomereli:
[Fix dev_get_parameter for agent](https://trello.com/c/VEs3QFn4/11-fix-devgetparameter-for-agent)

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -651,21 +651,18 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
...
-        if (!net::network_utils::linux_iface_get_mac("br-lan", alid)) {
-            reply_ucc(eWfaCaStatus::INVALID, "FAIL");
-        }
-
-        reply_ucc(eWfaCaStatus::COMPLETE, std::string("aLid,") + alid);
+        std::string value;
+        if (!handle_dev_get_param(params, value))
+            reply_ucc(eWfaCaStatus::ERROR, value);
+        else
+            reply_ucc(eWfaCaStatus::COMPLETE, value);
```
arnout: I think it makes more sense to keep the ALid handling in the superclass.

Also, I think it would be a good idea to make 'parameter' a mandatory parameter. So we end up with something like:

```
std::unordered_map<std::string, std::string> params {{"parameter"}};
if (!parse_params(cmd_tokens_vec, params, err_string)) {
    ...
}
 if (!reply_ucc(eWfaCaStatus::RUNNING)) {
       LOG(ERROR) << "failed to send reply";
        break;
}
auto parameter = params["parameter"];
if (parameter == "ALid") {
     std::string alid;
     ...
} else if (!handle_dev_get_param(parameter, params, value)
...
```

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -98,6 +98,19 @@ void agent_ucc_listener::update_vaps_list(std::string ruid, beerocks_message::sV
     vaps_map[ruid] = vaps;
 }
 
+bool agent_ucc_listener::handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
+                                              std::string &value)
+{
+    value = "FAIL";
+    if (params["parameter"] == "ALid") {
```
arnout: Shouldn't this be case-insensitive comparison, i.e. convert the parameter to lowercase first and then compare with "alid"?

common/beerocks/bcl/include/bcl/beerocks_ucc_listener.h
```diff
@@ -46,6 +46,8 @@ class beerocks_ucc_listener : public socket_thread {
                                           const std::string &dest_mac = std::string()) = 0;
     virtual bool handle_dev_set_config(std::unordered_map<std::string, std::string> &params,
                                        std::string &err_string)                        = 0;
+    virtual bool handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
```
arnout, line 49: Shouldn't we add a doxygen?

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -108,6 +108,27 @@ bool agent_ucc_listener::handle_dev_get_param(std::unordered_map<std::string, st
...
+        }
+        if (params.find("ssid") == params.end()) {
+            value += " missing ssid";
+            return false;
+        }
+        auto ruid = network_utils::mac_to_string(std::stoull(params["ruid"], nullptr, 16));
+        auto ssid = params["ssid"];
+
+        value += " macaddr not found for ruid " + ruid + " ssid " + ssid;
+        if (vaps_map.find(ruid) == vaps_map.end())
```
arnout:
```
auto vaps = vaps_map.find(ruid);
if (vaps = vaps_map.end())
    return false;
for (auto vap : *vaps)
    ...
```

(nitpick)

tools/docker/tests/test_flows.sh
```diff
@@ -26,7 +26,11 @@ container_ip() {
 
 container_CAPI_port() {
     # get the CAPI port based on the container name.
-    docker exec -t "$1" grep ucc_listener_port ${installdir}/config/beerocks_controller.conf  | cut -f2 -d= | cut -f1 -d' '
+    if [ "$1" = "gateway" ]; then
```
arnout, line 31: It is slightly weird that gateway == controller because the gateway also has agents. However, indeed, for test_flows, we always consider the gateway to be a controller-only device (i.e. the agent on the gateway is not addressed by CAPI).

So I'd add a comment here to explain why gateway == controller.

controller/src/beerocks/master/controller_ucc_listener.h
```diff
@@ -32,7 +32,8 @@ class controller_ucc_listener : public beerocks_ucc_listener {
                                   const std::string &dest_mac = std::string()) override;
     bool handle_dev_set_config(std::unordered_map<std::string, std::string> &params,
                                std::string &err_string) override;
-
+    bool handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
+                              std::string &value) override;
```
morantr, line 36: Using `std::string &value` instead of `std::string &err_string` for error is breaking the file convetion.
I get that you did it since it is not necessarily for error, and is using for the function positive output.
I would replace the return value to `std::string`, so in case it fails, it would return an empty string, and the err_string would be filled with the error description.
That way you won't have to define failure by the string "FAIL":

defintion:
```cpp
std::string handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
                                 std::string &err_string)
```
use:
```cpp
auto param_value = handle_dev_get_param(params, err_string)
if (handle_dev_get_param(params, value).empty()) {
    reply_ucc(eWfaCaStatus::ERROR, err_string);
    break;
}
    
reply_ucc(eWfaCaStatus::COMPLETE, param_value);
```

(nitpick)

tools/docker/tests/test_flows.sh
```diff
@@ -79,6 +79,11 @@ test_ap_config_renew() {
     # Wait a bit for the renew to complete
     sleep 3
 
+    TMP=$(send_CAPI_command repeater1 "dev_get_parameter,program,map,ruid,0x000000000000,ssid,Multi-AP-24G-1,parameter,macaddr")
+    echo $TMP
```
arnout: This should either be removed or `dbg`

tools/docker/tests/test_flows.sh
```diff
@@ -79,6 +79,11 @@ test_ap_config_renew() {
     # Wait a bit for the renew to complete
     sleep 3
 
+    TMP=$(send_CAPI_command repeater1 "dev_get_parameter,program,map,ruid,0x000000000000,ssid,Multi-AP-24G-1,parameter,macaddr")
```
arnout: (nitpick) it would make more sense to do this check after the checks of the log file.

tools/docker/tests/test_flows.sh
```diff
@@ -636,19 +641,19 @@ main() {
...
+        #         docker exec $dockerimage sh -c '
+        #             for logfile in controller agent agent_wlan0 agent_wlan2; do
+        #                 logpath=/tmp/$USER/beerocks/logs/beerocks_$logfile.log
+        #                 [ -e $logpath ] && {
+        #                     printf "\n\n==> %s <==\n" $logfile
+        #                     cat $logpath
+        #                 }
+        #             done'
+        #     done
+        # fi
```
arnout: Not related to this PR. But indeed a separate PR should be opened to remove this stuff, it was only relevant for Travis.

arnout (approved):
Nothing major so approving already.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -108,6 +108,27 @@ bool agent_ucc_listener::handle_dev_get_param(std::unordered_map<std::string, st
...
+            return false;
+        for (auto vap : vaps_map[ruid].vaps) {
+            if (std::string(vap.ssid) == ssid) {
+                value = "macaddr," + network_utils::mac_to_string(vap.mac);
+                return true;
+            }
+        }
```
morantr: I think it will be more readable and clear like that:
```cpp
// `value` is empty
auto ruid_it = vaps_map.find(ruid);
if (vaps_map.find(ruid) != vaps_map.end())
    for (const auto& vap : ruid_it->vaps) {
        if (std::string(vap.ssid) == ssid) {
            value = "macaddr," + network_utils::mac_to_string(vap.mac);
            return true;
        }
    }
}
value += " macaddr not found for ruid " + ruid + " ssid " + ssid;
return false;
```
(nitpick)

morantr (approved):

common/beerocks/bcl/include/bcl/beerocks_ucc_listener.h
tomereli: Done

controller/src/beerocks/master/controller_ucc_listener.h
tomereli: Not needed - in case of failure, value will be set to the error reason.

agent/src/beerocks/slave/agent_ucc_listener.cpp
tomereli: took a mix of yours and Moran's suggestions

PR #638, opened 2020-01-05
[KW]: Resolve uninitialized class member issue

Galvien:
After running the klocwork script on RDKB with prplMesh, the following error was received : 

" 'this->***' is not initialized in this constructor. "

This KW error indicates that there is a class member possibly unassigned.
This PR fixes all occurrences of this issue by updating each of those constructors to assign its uninitialized member a default value.

In addition, update .gitignore to ignore KW files irrelevant to the GitHub repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@ analyzer-venv
 /*.ipk
 /build/
 /logs/
+_GO_KW
+.kwlp
+.kwps
+kw_reports
--- a/common/beerocks/bcl/source/beerocks_logging.cpp
+++ b/common/beerocks/bcl/source/beerocks_logging.cpp
@@ -292,8 +292,7 @@ const std::string logging::syslogFormat("<%thread> %fbase[%line] --> %msg");
 
 logging::logging(const std::string config_path, std::string module_name)
     : m_module_name(module_name), m_logfile_size(LOGGING_DEFAULT_MAX_SIZE),
-      m_levels(LOG_LEVELS_GLOBAL_DEFAULT), m_syslog_levels(LOG_LEVELS_SYSLOG_DEFAULT),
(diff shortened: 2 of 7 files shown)


CoralMalachi:
The build failed, that's why it seemed it solved the issue.

.gitignore
```diff
@@ -2,3 +2,7 @@ analyzer-venv
 /*.ipk
 /build/
 /logs/
+_GO_KW
+.kwlp
+.kwps
+kw_reports
```
morantr, line 8: I think that we would like to see kw reports on the git, no?
@tomereli ?

common/beerocks/bcl/source/beerocks_logging.cpp
```diff
@@ -293,7 +293,7 @@ const std::string logging::syslogFormat("<%thread> %fbase[%line] --> %msg");
 logging::logging(const std::string config_path, std::string module_name)
     : m_module_name(module_name), m_logfile_size(LOGGING_DEFAULT_MAX_SIZE),
       m_levels(LOG_LEVELS_GLOBAL_DEFAULT), m_syslog_levels(LOG_LEVELS_SYSLOG_DEFAULT),
-      m_stdout_enabled("true"), m_syslog_enabled("false")
+        m_log_files_enabled(true), m_log_files_auto_roll(true), m_stdout_enabled(true), m_syslog_enabled(false)
```
morantr: Personally I am not a big fan of initializing class members in the initializer list, and do it only if I have to.
If it's possible, initialize the members on the class definition (.h file).

CoralMalachi: Agree.

morantr (approved):

PR #639, opened 2020-01-05
Feature/runtime raw socket disable

morantr:
If raw socket fails to be opened due to insufficient permissions continue without it.

### DIFF
--- a/agent/src/beerocks/monitor/monitor_rssi.cpp
+++ b/agent/src/beerocks/monitor/monitor_rssi.cpp
@@ -38,8 +38,6 @@ void monitor_rssi::stop()
         close(arp_socket);
         arp_socket = -1;
     }
-    mon_db       = nullptr;
-    slave_socket = nullptr;
 }
 
 bool monitor_rssi::start(monitor_db *mon_db_, Socket *slave_socket_)
@@ -61,6 +59,7 @@ bool monitor_rssi::start(monitor_db *mon_db_, Socket *slave_socket_)
     arp_socket_class = new Socket(arp_socket);
     std::string err  = arp_socket_class->getError();
     if (!err.empty()) {
+        arp_socket = -1;
         LOG(ERROR) << "Opening Socket err:" << err;
         return false;
     }
@@ -202,6 +201,13 @@ void monitor_rssi::process()
         auto arp_state  = sta_node->get_arp_state();
         auto &sta_stats = sta_node->get_stats();
 
+        // If the arp is disabled, there is no point being in any other state that is not IDLE,
+        // since all other states are arp related. Thus, override the arp state to IDLE when arp is
+        // disabled.
+        if (!arp_enabled()) {
+            sta_node->set_arp_state(monitor_sta_node::IDLE);
+        }
+
         if (arp_state == monitor_sta_node::IDLE) {
             if (!poll_last)
                 continue;
@@ -241,7 +247,7 @@ void monitor_rssi::process()
                                << " delta_val=" << int(delta_val);
                 }
             }
-            if (!conf_disable_initiative_arp) {
+            if (arp_enabled() && !conf_disable_initiative_arp) {
                 if (std::chrono::steady_clock::now() >=
                     (sta_node->get_last_change_time() +
                      std::chrono::milliseconds(mon_db->MONITOR_LAST_CHANGE_TIMEOUT_MSEC))) {
--- a/agent/src/beerocks/monitor/monitor_rssi.h
+++ b/agent/src/beerocks/monitor/monitor_rssi.h
@@ -43,6 +43,8 @@ private:
     void send_rssi_measurement_response(std::string &sta_mac, monitor_sta_node *sta_node);
     void monitor_idle_station(std::string &sta_mac, monitor_sta_node *sta_node);
 
+    bool arp_enabled() { return arp_socket > 0; }
+
     monitor_db *mon_db = nullptr;
     Socket *slave_socket;
     int arp_socket;
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -273,10 +273,10 @@ void monitor_thread::after_select(bool timeout)
 
             LOG(TRACE) << "mon_rssi.start()";
             if (!mon_rssi.start(&mon_db, slave_socket)) {
-                LOG(ERROR) << "mon_rssi.start() failed";
-                thread_last_error_code = MONITOR_THREAD_ERROR_ATTACH_FAIL;
-                stop_monitor_thread();
-                return;
+                // If monitor rssi failed to start, continue without it. It might failed due to
+                // insufficient permissions. Detailed error message is printed inside.
+                LOG(WARNING) << "mon_rssi.start() failed, ignore and continue without it";
+                mon_rssi.stop();
             }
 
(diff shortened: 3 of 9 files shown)


agent/src/beerocks/monitor/monitor_rssi.cpp
```diff
@@ -38,8 +38,6 @@ void monitor_rssi::stop()
         close(arp_socket);
         arp_socket = -1;
     }
-    mon_db       = nullptr;
```
tomereli, line 41: Why did you delete these lines? It is not clear from the commit message.

agent/src/beerocks/monitor/monitor_rssi.cpp
```diff
@@ -241,7 +247,7 @@ void monitor_rssi::process()
                                << " delta_val=" << int(delta_val);
                 }
             }
-            if (!conf_disable_initiative_arp) {
+            if (arp_enabled() || !conf_disable_initiative_arp) {
```
tomereli: Shouldn't this be `if (arp_enabled() && !conf_disable_initiative_arp)`?
And coming to think about it, shouldn't arp_enabled() be enough? i.e. move the conf_disable_initiative_arp check to inside the arp_enabled() API.

agent/src/beerocks/monitor/monitor_rssi.h
```diff
@@ -43,6 +43,8 @@ class monitor_rssi {
     void send_rssi_measurement_response(std::string &sta_mac, monitor_sta_node *sta_node);
     void monitor_idle_station(std::string &sta_mac, monitor_sta_node *sta_node);
 
+    bool arp_enabled() { return arp_socket > 0; }
```
tomereli, line 46: Here also add the initiative_arp config check.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -273,10 +273,10 @@ void monitor_thread::after_select(bool timeout)
 
             LOG(TRACE) << "mon_rssi.start()";
             if (!mon_rssi.start(&mon_db, slave_socket)) {
-                LOG(ERROR) << "mon_rssi.start() failed";
-                thread_last_error_code = MONITOR_THREAD_ERROR_ATTACH_FAIL;
-                stop_monitor_thread();
-                return;
+                // If monitor rssi failed to start, continue without it. It might failed due to
+                // insufficient permissions. Detailed error message is printed inside.
+                LOG(WARNING) << "mon_rssi.start() failed, ignore and continue without it";
```
tomereli, line 278: We should log the failure reason - i.e. if its because of missing permissions or some other failure. This has to be perfectly clear from this log.

agent/src/beerocks/slave/backhaul_manager/wan_monitor.cpp
```diff
@@ -33,7 +33,7 @@ static int netlink_open_socket()
     // Open a netlink socket
     int fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
     if (fd < 0) {
-        LOG(ERROR) << "Failed creating netlink socket";
+        LOG(ERROR) << "Failed creating netlink socket: " << strerror(errno);
```
tomereli, line 36: Good - will this print that there are no permissions to create the socket?

framework/platform/bpl/uci/arp/monitor/arp_monitor.cpp
```diff
@@ -217,21 +219,21 @@ bool arp_monitor::start(std::string strIface)
     // open netlink sock
     if ((m_fdMonSocket = socket(AF_NETLINK, SOCK_RAW | SOCK_CLOEXEC, NETLINK_ROUTE)) < 0) {
         LOG(ERROR) << "Failed opening socket: " << strerror(errno);
-        return false;
+        return -int(eErrorCode::OPERATION_NOT_SUPPORTED);
```
tomereli, line 222: Why is the cast needed here?

framework/platform/bpl/uci/arp/monitor/arp_monitor.cpp
```diff
@@ -217,21 +219,21 @@ bool arp_monitor::start(std::string strIface)
...
     }
 
     int sndbuf = SOCKET_SEND_BUF_SIZE;
     if (setsockopt(m_fdMonSocket, SOL_SOCKET, SO_SNDBUF, &sndbuf, sizeof(sndbuf)) < 0) {
         LOG(ERROR) << "Failed setting SO_SNDBUF option: " << strerror(errno);
         stop();
-        return false;
+        return -1;
```
tomereli, line 229: Why not return the enum as you do in other places?

tomereli (changes requested):
First commit message typo:
opening a raw a raw socket will fail. --> opening raw socket will fail.

agent/src/beerocks/monitor/monitor_rssi.cpp
morantr: nullptr assignment is not needed and not contributing anything. When the monitor calls to `mon_rssi.start()` and fails, it calls to `stop()` BUT it continue calling to `process()` on each work cycle, which uses these pointers.

agent/src/beerocks/monitor/monitor_rssi.cpp
morantr: Yes, you right it should be `&&`. But no, it can't be inside arp_enabled() since it is two different things.

agent/src/beerocks/monitor/monitor_rssi.h
morantr: No, the inactive arp is only for inactive arp, and not related to general arp.

agent/src/beerocks/slave/backhaul_manager/wan_monitor.cpp
morantr: yes

framework/platform/bpl/uci/arp/monitor/arp_monitor.cpp
morantr: because the function returns `int`. 😕

framework/platform/bpl/uci/arp/monitor/arp_monitor.cpp
morantr: `-1` represents general failure without any specific error code.  I used the enum to separate permissions error form other general errors. Basically bpl_err.h doesn't have a general error enum, and I wasn't sure if it is a good idea to add one. Anyway `1` is already taken for `UNKNOWN` on bpl_err.h, unless you would like to replace the `-1` with `-int(eErrorCode::UNKNOWN)`.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: We don't need it here. A detailed error will be printed inside (as I wrote on the comment here).

agent/src/beerocks/monitor/monitor_rssi.cpp
tomereli: Not sure I understand why it is two different things, but OK.

agent/src/beerocks/monitor/monitor_thread.cpp
tomereli: OK

tomereli (approved):
You answered all my comments so approving already.

PR #644, opened 2020-01-06
[DCS Patch 3/5] port dynamic channel selection from rdkb task

itayx:
> The Dynamic Channel Selection as a feature can provide information on the supported channels which can be utilized by the management system to provide better availability and user experience.
#562

The Dynamic Channel Selection feature is divided into several pull request, this is the third out of 5
start review from 676ac83
### Dynamic Channel Selection Task:
Changes in this PR include the following:
- controller: task: add clear pending events
- controller: dymanic channel selection task intro
- controller: DCS task: FSM
- controller: DCS task: events
- controller: DCS task: node & DB

### DIFF
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -32,14 +32,150 @@ void dynamic_channel_selection_task::work()
     TASK_LOG(DEBUG) << "FSM current state: " << s_ar_states[int(m_fsm_state)];
 
     switch (m_fsm_state) {
-    case eState::INIT:
-    case eState::IDLE:
-    case eState::TRIGGER_SCAN:
-    case eState::WAIT_FOR_SCAN_TRIGGERED:
-    case eState::WAIT_FOR_RESULTS_READY:
-    case eState::WAIT_FOR_RESULTS_DUMP:
-    case eState::SCAN_DONE:
-    case eState::ABORT_SCAN:
+    case eState::INIT: {
+        database.assign_dynamic_channel_selection_task_id(m_radio_mac, id);
+        // allow first scan
+        m_next_scan_timestamp_interval = std::chrono::steady_clock::now();
+        fsm_move_state(eState::IDLE);
+        break;
+    }
+    case eState::IDLE: {
+        if (m_is_single_scan_pending) {
+            m_is_single_scan         = true;
+            m_is_single_scan_pending = false;
+            fsm_move_state(eState::TRIGGER_SCAN);
+        } else if (database.get_channel_scan_is_enabled(m_radio_mac)) {
+            auto now = std::chrono::steady_clock::now();
+
+            if (now > m_next_scan_timestamp_interval) {
+                m_is_single_scan = false;
+                fsm_move_state(eState::TRIGGER_SCAN);
+                m_last_scan_try_timestamp = now;
+                LOG(DEBUG) << "interval condition is met, trigger scan";
+            }
+        }
+        break;
+    }
+    case eState::TRIGGER_SCAN: {
+        LOG(TRACE) << "TRIGGER_SCAN, mac=" << m_radio_mac << "scan_type is "
+                   << ((m_is_single_scan) ? "single-scan" : "continuous-scan");
+
+        // When a scan is requested send the scan parameters Channel pool & Dwell time
+        // Before sending the request set the scan_in_progress flag to true
+        // So another scan would not trigger on the same radio simultaneously
+
+        database.set_channel_scan_in_progress(m_radio_mac, true, m_is_single_scan);
+
+        auto request = beerocks::message_com::create_vs_message<
+            beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST>(cmdu_tx);
+        if (request == nullptr) {
+            LOG(ERROR)
+                << "Failed building message cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST !";
+            m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
+            fsm_move_state(eState::ABORT_SCAN);
+            break;
+        }
+
+        int32_t dwell_time_msec =
+            database.get_channel_scan_dwell_time_msec(m_radio_mac, m_is_single_scan);
+        if (dwell_time_msec < 0) {
+            LOG(ERROR) << "invalid dwell_time < 0";
+            m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
+            fsm_move_state(eState::ABORT_SCAN);
+            break;
+        }
+
+        //get current channel pool from DB
+        auto &curr_channel_pool = database.get_channel_scan_pool(m_radio_mac, m_is_single_scan);
+        if (curr_channel_pool.empty()) {
+            LOG(ERROR) << "empty channel pool is not supported. please set channel pool for mac="
+                       << m_radio_mac;
+            m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
+            fsm_move_state(eState::ABORT_SCAN);
+            break;
+        }
+
+        if (curr_channel_pool.size() > beerocks::message::SUPPORTED_CHANNELS_LENGTH) {
+            LOG(ERROR) << "channel_pool is too big";
+            m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_POOL_TOO_BIG;
+            fsm_move_state(eState::ABORT_SCAN);
+            break;
+        }
+
+        request->scan_params().radio_mac         = m_radio_mac;
+        request->scan_params().dwell_time_ms     = dwell_time_msec;
+        request->scan_params().channel_pool_size = curr_channel_pool.size();
+        std::copy(curr_channel_pool.begin(), curr_channel_pool.end(),
+                  request->scan_params().channel_pool);
+
+        // get the parent node to send the CMDU to the agent
+        auto radio_mac_str = beerocks::net::network_utils::mac_to_string(m_radio_mac);
+        auto ire           = database.get_node_parent_ire(radio_mac_str);
+        son_actions::send_cmdu_to_agent(ire, cmdu_tx, database, radio_mac_str);
+
+        set_events_timeout(SCAN_TRIGGERED_WAIT_TIME_MSEC);
+        dcs_wait_for_event(eEvent::SCAN_TRIGGERED);
+
+        fsm_move_state(eState::WAIT_FOR_SCAN_TRIGGERED);
+        break;
+    }
+    case eState::WAIT_FOR_SCAN_TRIGGERED: {
+        TASK_LOG(ERROR) << "This should not happen!";
+
+        m_last_scan_error_code =
+            beerocks::eChannelScanErrCode::CHANNEL_SCAN_TRIGGERED_EVENT_TIMEOUT;
+
+        fsm_move_state(eState::ABORT_SCAN);
+        break;
+    }
+    case eState::WAIT_FOR_RESULTS_READY: {
+        TASK_LOG(ERROR) << "This should not happen!";
+
+        m_last_scan_error_code =
+            beerocks::eChannelScanErrCode::CHANNEL_SCAN_RESULTS_READY_EVENT_TIMEOUT;
+
+        fsm_move_state(eState::ABORT_SCAN);
+        break;
+    }
+    case eState::WAIT_FOR_RESULTS_DUMP: {
+        TASK_LOG(ERROR) << "This should not happen!";
+
+        m_last_scan_error_code =
+            beerocks::eChannelScanErrCode::CHANNEL_SCAN_RESULTS_DUMP_EVENT_TIMEOUT;
+
+        fsm_move_state(eState::ABORT_SCAN);
+        break;
+    }
+    case eState::SCAN_DONE: {
+        LOG(TRACE) << "SCAN_DONE";
+
+        //update next continuous scan time
+        if (!m_is_single_scan) {
+            auto interval =
+                std::chrono::seconds(database.get_channel_scan_interval_sec(m_radio_mac));
+            m_next_scan_timestamp_interval = m_last_scan_try_timestamp + interval;
+        }
+
+        m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_SUCCESS;
+        database.set_channel_scan_results_status(m_radio_mac, m_last_scan_error_code,
+                                                 m_is_single_scan);
+
+        database.set_channel_scan_in_progress(m_radio_mac, false, m_is_single_scan);
+        m_is_single_scan = false;
+        fsm_move_state(eState::IDLE);
+        break;
+    }
+    case eState::ABORT_SCAN: {
+        LOG(ERROR) << "aborting scan for mac=" << m_radio_mac << ", last_scan_timestamp is not set";
+
+        database.set_channel_scan_results_status(m_radio_mac, m_last_scan_error_code,
+                                                 m_is_single_scan);
+        database.set_channel_scan_in_progress(m_radio_mac, false, m_is_single_scan);
+        m_is_single_scan = false;
+
+        fsm_move_state(eState::IDLE);
+        break;
+    }
     default: {
         break;
     }
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -1,0 +1,123 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2019-2020 Intel Corporation
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "dynamic_channel_selection_task.h"
+#include "../son_actions.h"
+#include <bcl/beerocks_defines.h>
+#include <bcl/network/network_utils.h>
+#include <easylogging++.h>
+
+#define SCAN_TRIGGERED_WAIT_TIME_MSEC 20000
+#define SCAN_RESULTS_DUMP_WAIT_TIME_MSEC 40000
+
+std::string s_ar_states[] = {FOREACH_DCS_STATE(GENERATE_STRING)};
+std::string s_ar_events[] = {FOREACH_DCS_EVENT(GENERATE_STRING)};
+
+dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
+                                                               ieee1905_1::CmduMessageTx &cmdu_tx_,
+                                                               task_pool &tasks_,
+                                                               sMacAddr radio_mac_)
+    : database(database_), cmdu_tx(cmdu_tx_), tasks(tasks_), m_radio_mac(radio_mac_)
+{
+    m_fsm_state = eState::INIT;
+}
+
+void dynamic_channel_selection_task::work()
+{
+    TASK_LOG(DEBUG) << "FSM current state: " << s_ar_states[int(m_fsm_state)];
+
+    switch (m_fsm_state) {
+    case eState::INIT:
+    case eState::IDLE:
+    case eState::TRIGGER_SCAN:
+    case eState::WAIT_FOR_SCAN_TRIGGERED:
+    case eState::WAIT_FOR_RESULTS_READY:
+    case eState::WAIT_FOR_RESULTS_DUMP:
+    case eState::SCAN_DONE:
+    case eState::ABORT_SCAN:
+    default: {
+        break;
+    }
+    }
+    TASK_LOG(DEBUG) << "FSM next state: " << s_ar_states[int(m_fsm_state)];
+}
+
+void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
+{
+    if (obj == nullptr) {
+        TASK_LOG(ERROR) << "obj == nullptr";
+        return;
+    }
+
+    auto event_radio_mac = *reinterpret_cast<sMacAddr *>(obj);
+    if (m_radio_mac != event_radio_mac) {
+        TASK_LOG(ERROR) << "event : " << s_ar_events[int(eEvent(event_type))]
+                        << " in state: " << s_ar_states[int(m_fsm_state)]
+                        << " for mac: " << event_radio_mac;
+        if (m_dcs_waiting_for_event != eEvent::INVALID_EVENT) {
+            TASK_LOG(DEBUG) << "calling wait_for_event for "
+                            << s_ar_events[int(m_dcs_waiting_for_event)];
+            wait_for_event(int(m_dcs_waiting_for_event));
+        }
+        return;
+    }
+
+    // This flag (event_handled) indicate if event has arived in the correct fsm state.
+    bool event_handled = false;
+    switch (eEvent(event_type)) {
+    case eEvent::TRIGGER_SINGLE_SCAN:
+    case eEvent::SCAN_TRIGGERED:
+    case eEvent::SCAN_RESULTS_READY:
+    case eEvent::SCAN_RESULTS_DUMP:
+    case eEvent::SCAN_FINISHED:
+    case eEvent::SCAN_ABORTED:
+    case eEvent::SCAN_ENABLE_CHANGE:
+    default: {
+        break;
+    }
+    }
+
+    if (!event_handled) {
+        TASK_LOG(ERROR) << "unexpected event: " << s_ar_events[int(eEvent(event_type))]
+                        << " in state: " << s_ar_states[int(m_fsm_state)];
+    } else {
+        m_dcs_waiting_for_event = eEvent::INVALID_EVENT;
+    }
+}
+
+void dynamic_channel_selection_task::handle_events_timeout(std::multiset<int> pending_events)
+{
+    // only one event is expected
+    for (std::multiset<int>::iterator it = pending_events.begin(); it != pending_events.end();
+         ++it) {
+        TASK_LOG(ERROR) << "event " << *it << " timed out on " << m_radio_mac
+                        << ", going to abort-scan state -> handle_nl_dead_node";
+
+        switch (m_fsm_state) {
+        case eState::WAIT_FOR_SCAN_TRIGGERED:
+        case eState::WAIT_FOR_RESULTS_READY:
+        case eState::WAIT_FOR_RESULTS_DUMP:
+        case eState::SCAN_DONE:
+        case eState::IDLE:
+        case eState::ABORT_SCAN: {
+            break;
+        }
+        default: {
+            TASK_LOG(ERROR) << "Unexpected event timeout!";
+            break;
+        }
+        }
+    }
+    m_dcs_waiting_for_event = eEvent::INVALID_EVENT;
+}
+
+void dynamic_channel_selection_task::dcs_wait_for_event(eEvent dcs_event)
+{
+    m_dcs_waiting_for_event = dcs_event;
+    wait_for_event((int)dcs_event);
+}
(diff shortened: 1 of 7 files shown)


controller/src/beerocks/master/tasks/task.cpp
```diff
@@ -89,6 +89,10 @@ void task::wait_for_task_end(int task_id, int ms)
     waiting                  = true;
     pending_task_id          = task_id;
 }
+void task::clear_pending_events()
```
tomereli: Please explain in the commit message why this is (or will be) needed so when we review this commit we understand the context. How come this was never needed so far?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
```
tomereli: New file - 2020

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "dynamic_channel_selection_task.h"
+
```
tomereli: Please remove empty lines between includes

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <bcl/beerocks_defines.h>
+#include <bcl/network/network_utils.h>
+#include <easylogging++.h>
+
+using namespace beerocks;
+using namespace net;
+using namespace son;
+
+#define FSM_MOVE_STATE(new_state)                                                                  \
```
tomereli: Please change all these to class private methods. We do not use MACROs which access class members.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+const char *dynamic_channel_selection_task::s_ar_events[] = {FOREACH_DCS_EVENT(GENERATE_STRING)};
+
+dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
+                                                               ieee1905_1::CmduMessageTx &cmdu_tx_,
+                                                               task_pool &tasks_,
+                                                               std::string radio_mac_)
+    : database(database_), cmdu_tx(cmdu_tx_), tasks(tasks_), m_radio_mac(radio_mac_)
+{
+    m_fsm_state = eState::INIT;
```
tomereli, line 27: Why not move the `m_fsm_state` to the initializer list like all other members?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+    m_dcs_waiting_for_event = eEvent::INVALID_EVENT;
+}
+
+void dynamic_channel_selection_task::dcs_wait_for_event(eEvent dcs_event)
+{
+    m_dcs_waiting_for_event = dcs_event;
+    wait_for_event((int)dcs_event);
+}
```
tomereli: Missing line break at end of file.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -0,0 +1,118 @@
...
+namespace son {
+
+class dynamic_channel_selection_task : public task, public std::enable_shared_from_this<task> {
+public:
+    dynamic_channel_selection_task(db &database, ieee1905_1::CmduMessageTx &cmdu_tx_,
+                                   task_pool &tasks_, std::string radio_mac_);
+    virtual ~dynamic_channel_selection_task() {}
+
+    typedef struct {
+        sMacAddr radio_mac;
```
tomereli: No need for `typedef` - you can declare the struct as is and access it without the `struct` keyword in c++

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -0,0 +1,118 @@
...
+    std::chrono::steady_clock::time_point m_last_scan_try_timestamp;
+
+    beerocks::eChannelScanErrCode m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_NO_ERROR;
+    eEvent m_dcs_waiting_for_event                   = eEvent::INVALID_EVENT;
+
+    bool m_is_single_scan_pending = false;
+    bool m_is_single_scan         = false;
+};
+} //namespace son
+#endif
```
tomereli: Missing line break at end of file.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -46,14 +46,142 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+        son_actions::send_cmdu_to_agent(ire, cmdu_tx, database, m_radio_mac);
+
+        set_events_timeout(SCAN_TRIGGERED_WAIT_TIME_MSEC);
+        dcs_wait_for_event(eEvent::SCAN_TRIGGERED);
+
+        FSM_MOVE_STATE(WAIT_FOR_SCAN_TRIGGERED);
+        break;
+    }
+    case eState::WAIT_FOR_SCAN_TRIGGERED: {
+        TASK_LOG(ERROR) << "This should not happen!";
```
tomereli, line 123: OK, why this shouldn't happen though? should be a comment explaining this. I really don't understand, you move the FSM to this state so the next work iteration you will get here..

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -46,14 +46,142 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+
+        m_last_scan_error_code = beerocks::CHANNEL_SCAN_TRIGGERED_EVENT_TIMEOUT;
+
+        FSM_MOVE_STATE(ABORT_SCAN);
+        break;
+    }
+    case eState::WAIT_FOR_RESULTS_READY: {
+        TASK_LOG(ERROR) << "This should not happen!";
```
tomereli, line 132: Same comment - needs explaining.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -46,14 +46,142 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+
+        m_last_scan_error_code = beerocks::CHANNEL_SCAN_RESULTS_READY_EVENT_TIMEOUT;
+
+        FSM_MOVE_STATE(ABORT_SCAN);
+        break;
+    }
+    case eState::WAIT_FOR_RESULTS_DUMP: {
+        TASK_LOG(ERROR) << "This should not happen!";
```
tomereli, line 141: same here.

tomereli (changes requested):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include "../son_actions.h"
+
+#include <bcl/beerocks_defines.h>
+#include <bcl/network/network_utils.h>
+#include <easylogging++.h>
+
+using namespace beerocks;
+using namespace net;
+using namespace son;
```
vitalybu: Adding `using namespace` globally is bad practice.
Namespaces were created for a reason, one of which is to encapsulate different types and functions.
Just use the namespaces explicitly where it's needed (e.g. `beerocks::...` or `beerocks::net::...`)

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+    switch (m_fsm_state) {
+    case eState::INIT:
+    case eState::IDLE:
+    case eState::TRIGGER_SCAN:
+    case eState::WAIT_FOR_SCAN_TRIGGERED:
+    case eState::WAIT_FOR_RESULTS_READY:
+    case eState::WAIT_FOR_RESULTS_DUMP:
+    case eState::SCAN_DONE:
+    case eState::ABORT_SCAN:
```
vitalybu: Why is this empty? no `break;` and no `FALLTHROUGH`?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+}
+
+void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
+{
+    if (obj == nullptr) {
+        TASK_LOG(ERROR) << "obj == nullptr";
+        return;
+    }
+    auto event_radio_mac = network_utils::mac_to_string(((sMacAddr *)obj)->oct);
+    if (m_radio_mac == event_radio_mac) {
```
vitalybu: Always check for errors or have the shorter code section first.
In this case, you can check if `m_radio_mac != event_radio_mac` first, do whatever you do, and then continue to the actual logic without the additional indentation:
```
if (m_radio_mac != event_radio_mac) {
    ... handle ...
    return
}

switch (eEvent(event_type)) {
    case ...
    case ...
}
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,131 @@
...
+        case eEvent::SCAN_RESULTS_DUMP:
+        case eEvent::SCAN_FINISHED:
+        case eEvent::SCAN_ABORTED:
+        case eEvent::SCAN_ENABLE_CHANGE:
+        default: {
+            break;
+        }
+        }
+
+        if (!handle_event) {
```
vitalybu: Won't this be handled automatically by the `default:` case?
Why is there a need for the `handle_event` variable?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -0,0 +1,118 @@
...
+    enum class eState { FOREACH_DCS_STATE(GENERATE_ENUM) };
+
+    // Strings Array
+    static const char *s_ar_states[];
+    static const char *s_ar_events[];
+
+    eState m_fsm_state;
+
+    const static int SCAN_TRIGGERED_WAIT_TIME_MSEC    = 20000;
+    const static int SCAN_RESULTS_DUMP_WAIT_TIME_MSEC = 40000;
```
vitalybu: Why `const static int` and not a define in the cpp file?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -0,0 +1,118 @@
...
+    STATE(WAIT_FOR_RESULTS_DUMP)                                                                   \
+    STATE(SCAN_DONE)                                                                               \
+    STATE(ABORT_SCAN)
+
+    // State ENUM
+    enum class eState { FOREACH_DCS_STATE(GENERATE_ENUM) };
+
+    // Strings Array
+    static const char *s_ar_states[];
+    static const char *s_ar_events[];
```
vitalybu: Why not simply static arrays in the cpp file?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -46,14 +46,142 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+    }
+    case eState::TRIGGER_SCAN: {
+        LOG(TRACE) << "TRIGGER_SCAN, mac=" << m_radio_mac << "scan_type is "
+                   << ((m_is_single_scan) ? "single-scan" : "continuous-scan");
+
+        // When a scan is requested send the scan parameters Channel pool & Dwell time
+        // Before sending the request set the scan_in_progress flag to true
+        // So another scan would not trigger on the same radio simultaneously
+
+        database.set_channel_scan_in_progress(m_radio_mac, true, m_is_single_scan);
```
vitalybu, line 67: Why do we store this in the database? Is it accessed from additional places?

vitalybu (changes requested):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
arnout: If the file was first created (and distributed to customers, even if in binary form) in 2016, then the starting copyright year really should be 2016.

controller/src/beerocks/master/tasks/task.cpp
kantera800: we clear events before task is restarted (ABORT or SCAN_FINISH) since if we dont handle_event() will not handle them and throw error in the log file.

maybe we should start using in other tasks also.

kantera800: will fix the commit comment after approval  in order not to break the order.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: it was created in in 2019 in intel repos so i changed to 2019-2020

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: not sure we need to start the task in a specific state.
I think we can remove it or just initialize m_fsm_state  in the header to INIT.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: Its a preparation for  next commit "controller: DCS task: FSM"

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: fixed

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: no.
"handle_event" flag is true  when a legal event received in the wrong fsm state.
we prefer to terminate the scan immediately and not let it reach timeout.
@itayx  do you think it could be removed?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
kantera800: then i would have to add reference of it to fsm_move_state() or copy the FSM log print to the cpp file.
much uglier option for my opinion.

maybe if i remove the FSM print from fsm_move_state() completely.

BTW same implementation in channel_selection_task.h

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: yes, we cannot allow user to trigger chan scan on the same radio in the middle of a already running scan

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: in ::work()
we TRIGGER_SCAN and got to WAIT_FOR_SCAN_TRIGGERED state and set the timout.
in  handle_event()
if  we dont recieve SCAN_TRIGGERED event from radio thats why we need the WAIT_FOR_SCAN_TRIGGERED  timout.

same for:
    case eState::WAIT_FOR_RESULTS_READY:
    case eState::WAIT_FOR_RESULTS_DUMP:

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: [see](https://github.com/prplfoundation/prplMesh/pull/644/commits/cd5b4d9a5530c2fb52d93c3f08cbef8c4d71e160#r367269605)

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: [see](https://github.com/prplfoundation/prplMesh/pull/644/commits/cd5b4d9a5530c2fb52d93c3f08cbef8c4d71e160#r367269605)

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: It's less a "This should not happen" and more on the lines of "If this happens it means something somewhere went wrong"

The WAIT_FOR_ states are for timeout handling

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: It's used to validate that if an event was not only received but also handled
the switch case by it self isn't enough to validate an event, since some events are only useful is certain states

So I don't think it can be removed, unless there is a better way to validate if an event was handled

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
tomereli: nipick anyway

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -194,159 +185,164 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
...
-                dcs_wait_for_event(eEvent::SCAN_RESULTS_DUMP);
-
-                FSM_MOVE_STATE(WAIT_FOR_RESULTS_READY);
-            } else {
-                TASK_LOG(DEBUG) << "ignoring SCAN_TRIGERRED event,"
-                                << " not in state WAIT_FOR_SCAN_TRIGGERED";
-            }
-            break;
+
+    auto event_radio_mac = beerocks::net::network_utils::mac_to_string(((sMacAddr *)obj)->oct);
```
tomereli: please changhe to `mac_to_string(obj)`

tomereli: actually no need to convert this to string at all, so please don't do this.
@arnout :)

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -194,159 +185,164 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
...
+    if (m_radio_mac != event_radio_mac) {
```
tomereli, line 194: m_radio_mac and event_radio_mac should be of type `sMacAddr` and the have the `==` operator.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -194,159 +185,164 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
...
+        TASK_LOG(ERROR) << "event : " << s_ar_events[int(eEvent(event_type))]
+                        << " in state: " << s_ar_states[int(m_fsm_state)]
+                        << " for mac: " << event_radio_mac;
```
tomereli, line 197: also here - `event_radio_mac` if it will be a simple `sMacAddr` - has the `<<` operator for `cout` so you can use it in prints.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -194,159 +185,164 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
...
+    }
+
+    bool handle_event = false;
+    switch (eEvent(event_type)) {
+    case eEvent::TRIGGER_SINGLE_SCAN: {
+        TASK_LOG(DEBUG) << "TRIGGER_SINGLE_SCAN received";
+        auto trigger_single_scan_event = reinterpret_cast<sTriggerSingleScan_event *>(obj);
+        handle_event                   = true;
+        TASK_LOG(DEBUG) << "TRIGGER_SINGLE_SCAN handled on:"
+                        << beerocks::net::network_utils::mac_to_string(
```
tomereli: No more `mac_to_string` calls, please

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -29,34 +29,28 @@ class dynamic_channel_selection_task : public task, public std::enable_shared_fr
...
     virtual ~dynamic_channel_selection_task() {}
 
-    typedef struct {
+    struct sTriggerSingleScan_event {
         sMacAddr radio_mac;
-    } sTriggerSingleScan_event;
-
-    typedef struct {
+    };
+    struct sScanTriggered_event {
```
tomereli: I think it would be much better to define a single event struct and an enum for the event type, instead of duplicating event structures like that.

```
enum class eDcsEventType { SCAN_TRIGGERED, SINGLE_SCAN, ...};

struct sDcsEvent {
    eDcsEventType type;
    sMacAddr radio_mac;
    union data {
        sChannelScanResultsElement scan_results;
        ...
    };
```

tomereli (changes requested):

kantera800:
[[DCS Patch 3/5]Port dynamic channel selection](https://trello.com/c/vLpLldqk/77-dcs-patch-3-5port-dynamic-channel-selection)

controller/src/beerocks/master/tasks/task.cpp
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done.
had to change all database.channel_scan methods (12 methods) .

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
kantera800: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: done

kantera800:
ready for re-review:

fixed all code review issues.
rebased to patch 2/5

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -443,6 +447,30 @@ enum eBssType {
     BSS_TYPE_INVALID
 };
 
+enum class eChannelScanErrCode : uint8_t {
```
tomereli: Since its an enum class, there is no need to have the `CHANNEL_SCAN_` prefix

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "dynamic_channel_selection_task.h"
+#include "../son_actions.h"
+#include <bcl/beerocks_defines.h>
+#include <bcl/network/network_utils.h>
+#include <easylogging++.h>
+
+#define SCAN_TRIGGERED_WAIT_TIME_MSEC 20000
+#define SCAN_RESULTS_DUMP_WAIT_TIME_MSEC 40000
+
+const char *dynamic_channel_selection_task::s_ar_states[] = {FOREACH_DCS_STATE(GENERATE_STRING)};
```
tomereli: Why not const std::string?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -0,0 +1,106 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
```
tomereli: 2020

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -31,14 +31,151 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+    case eState::ABORT_SCAN: {
+        LOG(ERROR) << "aborting scan for mac=" << m_radio_mac << ", last_scan_timestamp is not set";
+
+        database.set_channel_scan_results_status(m_radio_mac, m_last_scan_error_code,
+                                                 m_is_single_scan);
+        database.set_channel_scan_in_progress(m_radio_mac, false, m_is_single_scan);
+        m_is_single_scan = false;
+
+        fsm_move_state(eState::IDLE);
+        work(); // Skip pause, move to Idle immediately
```
tomereli: I don't understand this - we shouldn't call work() directly, should we? So why is it done here?

tomereli (approved):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: Will add comment that explains, and rename event_handled

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: check if can read task fsm state instead #665

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: proposed by @moran in order not to wait for next after_select to reach Idle.

common/beerocks/bcl/include/bcl/beerocks_defines.h
kantera800: #665

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
kantera800: removed dependence from fsm_move_state() .
moved to cpp.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
tomereli: But why? if it is for "performance" - please remove it. Readability is more important.

kantera800: yes performance.
removed

vitalybu (approved):

PR #645, opened 2020-01-06
[KW]: Resolve nullptr issue

CoralMalachi:
After running the klocwork script on RDKB with prplMesh, the following error was received :

" Pointer *** returned from call to function *** may be NULL and will be dereferenced at ... "

To solve this error, add a check to validate that the return value of `create_vs_message`
method is not nullptr.

Signed-off-by: CoralMalachi <coral.malachi@intel.com>

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1463,6 +1463,12 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
         auto register_response =
             message_com::create_vs_message<beerocks_message::cACTION_BACKHAUL_REGISTER_RESPONSE>(
                 cmdu_tx);
+
+        if (register_response == nullptr) {
+            LOG(ERROR) << "Failed building message!";
+            return false;
+        }
+
         message_com::send_cmdu(soc->slave, cmdu_tx);
         break;
     }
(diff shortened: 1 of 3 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1463,6 +1463,12 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
         auto register_response =
             message_com::create_vs_message<beerocks_message::cACTION_BACKHAUL_REGISTER_RESPONSE>(
                 cmdu_tx);
+
+        if (register_response == nullptr) {
```
tomereli, line 1467: `if (!register_response)`?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1895,6 +1901,11 @@ bool backhaul_manager::send_slaves_enable()
             message_com::create_vs_message<beerocks_message::cACTION_BACKHAUL_ENABLE_APS_REQUEST>(
                 cmdu_tx);
 
+        if (notification == nullptr) {
```
tomereli, line 1904: same here

tomereli (changes requested):
Didn't we say that KW complains when we compare to nullptr instead of `if (!ptr)`? even if in this case it doesn't, I still think we should align our code to one convention.

CoralMalachi:
> Didn't we say that KW complains when we compare to nullptr instead of `if (!ptr)`? even if in this case it doesn't, I still think we should align our code to one convention.

Yes, but we also decided to ignore this issue. if we change it here we need to change everywhere we use it. BTW, the in case of failure, `create_vs_message` returns a nullptr so comparing it to nullptr looks correct to me.

tomereli (approved):

arnout:
> > Didn't we say that KW complains when we compare to nullptr instead of `if (!ptr)`? even if in this case it doesn't, I still think we should align our code to one convention.
> 
> Yes, but we also decided to ignore this issue. if we change it here we need to change everywhere we use it.

The original idea of #611 was indeed to replace all `foo == nullptr` with `!foo`, and you had already done that work. So why did we change our mind?

PR #646, opened 2020-01-07
.gitlab-ci.yml: remove docker images after building

rmelotte:
Since we now use unique names for the docker images, we have to delete
the images we create or they will be kept around forever (they are not
and should not be overridden by the next builds).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -73,6 +73,8 @@ run-tests:
   script:
     - cd tools/docker/builder/openwrt
     - ./build.sh -v -d "$TARGET_DEVICE" -t "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee build.log | grep '^make\[[12]\]\|^Step\|^ --->' --color=never --line-buffered
+  after_script:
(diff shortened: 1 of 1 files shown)


tomereli (approved):

.gitlab-ci.yml
```diff
@@ -73,6 +73,8 @@ run-tests:
   script:
     - cd tools/docker/builder/openwrt
     - ./build.sh -v -d "$TARGET_DEVICE" -t "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee build.log | grep '^make\[[12]\]\|^Step\|^ --->' --color=never --line-buffered
+  after_script:
+    - docker rmi "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID"
```
arnout, line 77: Since it's used twice it would make sense to put the image name+tag in a variable. This would also allow it to be overridden by a manually triggered build.

Nitpick

arnout (approved):

.gitlab-ci.yml
rmelotte: I agree and I did that at some point, but gitlab-ci doesn't really support nested variables...
See [gitlab-ci's doc](https://docs.gitlab.com/ee/ci/variables/where_variables_can_be_used.html#gitlab-runner-internal-variable-expansion-mechanism): 

> the expansion is done only once, so nested variables may or may not work, depending on the ordering of variables definitions

PR #30, opened 2020-01-07
ieee1905_transport: use bind() instead of bpf filter

ghost:
The IEEE1905 transport opens a raw socket to capture 1905.1 and LLDP packets. It currently uses a BPF filter to do that. However, on some platforms, a BPF filter causes the acceleration fabric to stop working or to degrade.

A better solution is to use bind() to tie the raw socket to a specific ethernet protocol. This looks more or less like this:
```c
int openPacketSocket(int ifindex, uint16_t eth_type)
{
    int                 s;
    struct sockaddr_ll  socket_address;

    s = socket(AF_PACKET, SOCK_RAW, htons(eth_type));
    if (-1 == s)
    {
        return -1;
    }

    memset(&socket_address, 0, sizeof(socket_address));
    socket_address.sll_family   = AF_PACKET;
    socket_address.sll_ifindex  = ifindex;
    socket_address.sll_protocol = htons(eth_type);

    if (-1 == bind(s, (struct sockaddr*)&socket_address, sizeof(socket_address)))
    {
        close(s);
        return -1;
    }

    return s;
}
```

Note that a socket can be bound only to 1 ethertype. Therefore, we need to use 2 sockets per interface instead of just one.

### DIFF
--- a/framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
+++ b/framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
@@ -21,66 +21,6 @@
 
 namespace mapf {
 
-// helper class to manage socket filter programs (Berkely Socket Filter)
-//
-// Note: I chose to use here the classic BPF implementation. Once all targets move to kernel version >= 3.19
-// this code and the Ieee1905SocketFilter class could be migrated to the extended BPF (eBPF) system - see bpf(2)
-// http://man7.org/linux/man-pages/man2/bpf.2.html
-class Ieee1905SocketFilter {
-public:
-    // create a filter that accepts packets that match the basic transport requirements - see below.
-    // The two addresses are here to specify the AL MAC address and the interface's hardware address
-    Ieee1905SocketFilter(const uint8_t *addr0 = NULL, const uint8_t *addr1 = NULL)
-    {
-        static const uint8_t ieee1905_multicast_address[ETH_ALEN] = {
-            0x01, 0x80, 0xc2, 0x00, 0x00, 0x13}; // 01:80:c2:00:00:13
-
-        // use the IEEE1905 Multicast Address as default value (it is passed by the filter anyway)
-        if (!addr0)
-            addr0 = ieee1905_multicast_address;
-
-        if (!addr1)
-            addr1 = ieee1905_multicast_address;
-
-        filter[4].k = (uint32_t)addr0[2] << 24 | (uint32_t)addr0[3] << 16 |
-                      (uint32_t)addr0[4] << 8 | (uint32_t)addr0[5];
-        filter[6].k = (uint32_t)addr0[0] << 8 | (uint32_t)addr0[1];
-
-        filter[7].k = (uint32_t)addr1[2] << 24 | (uint32_t)addr1[3] << 16 |
-                      (uint32_t)addr1[4] << 8 | (uint32_t)addr1[5];
-        filter[9].k = (uint32_t)addr1[0] << 8 | (uint32_t)addr1[1];
-
-        if (!addr0 && !addr1) {
-            MAPF_WARN("at least one address should be specified for socket filter.");
-        }
-    }
-
-    const struct sock_fprog &sock_fprog() const { return fprog; }
-
-private:
-    struct sock_filter filter[17] = {
-        // This BPF is designed to accepts the following packets:
-        // - IEEE1905 multicast packets (with IEEE1905 Multicast Address set as destination address)
-        // - LLDP multicast packets (with LLDP Multicast Address as destination address)
-        // - IEEE1905 unicast packets (with either this devices' AL MAC address or the interface's HW address set as destination address)
-        //
-        // generated using: tcpdump -dd '(ether proto 0x893a and (ether dst 01:80:c2:00:00:13 or ether dst 11:22:33:44:55:66 or ether dst 77:88:99:aa:bb:cc)) or (ether proto 0x88cc and ether dst 01:80:c2:00:00:0e)'
-        // the two dummy addresses in this filter 11:22... and 77:88... will be replaced in runtime with the AL MAC address and the interface's HW address
-        //
-        {0x28, 0, 0, 0x0000000c}, {0x15, 0, 8, 0x0000893a}, {0x20, 0, 0, 0x00000002},
-        {0x15, 9, 0, 0xc2000013}, {0x15, 0, 2, 0x33445566}, // 4: replace with AL MAC Addr [2..5]
-        {0x28, 0, 0, 0x00000000}, {0x15, 8, 9, 0x00001122}, // 6: replace with AL MAC Addr [0..1]
-        {0x15, 0, 8, 0x99aabbcc},                           // 7: replace with IF MAC Addr [2..5]
-        {0x28, 0, 0, 0x00000000}, {0x15, 5, 6, 0x00007788}, // 9: replace with IF MAC Addr [0..1]
-        {0x15, 0, 5, 0x000088cc}, {0x20, 0, 0, 0x00000002}, {0x15, 0, 3, 0xc200000e},
-        {0x28, 0, 0, 0x00000000}, {0x15, 0, 1, 0x00000180}, {0x6, 0, 0, 0x0000ffff},
-        {0x6, 0, 0, 0x00000000},
-    };
-
-    struct sock_fprog fprog = {.len    = sizeof(filter) / sizeof(struct sock_filter),
-                               .filter = filter};
-};
-
 void Ieee1905Transport::update_network_interfaces(
     std::map<unsigned int, NetworkInterface> updated_network_interfaces)
 {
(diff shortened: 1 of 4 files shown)


tomereli:
[Replace BPF filter in transport](https://trello.com/c/MMvXUWzL/55-replace-bpf-filter-in-transport)

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -137,50 +81,62 @@ void Ieee1905Transport::update_network_interfaces(
...
-    // and send packets with the Ethernet header included in the buffer. Please consider changing
-    // implementation to use SOCK_DGRAM protocol (without L2 header handling)
-
+    //TODO: Fixup code for SOCK_RAW -> SOCK_DGRAM transition
     // open packet raw socket - see man packet(7) https://linux.die.net/man/7/packet
     int sockfd;
-    if ((sockfd = socket(AF_PACKET, SOCK_RAW, htons(ETH_P_ALL))) < 0) {
-        MAPF_ERR("cannot open raw socket \"" << strerror(errno) << "\" (" << errno << ").");
-        return false;
+    if ((sockfd = socket(AF_PACKET, SOCK_DGRAM, htons(ether_type))) < 0) {
```
arnout: We still want SOCK_RAW, otherwise the Ethernet header is stripped off and we parse that.

Note that it's not strictly needed (we can get source and destination address through ancillary data) but it's a lot simpler that way.

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -137,50 +81,62 @@ void Ieee1905Transport::update_network_interfaces(
...
     }
 
-    // the interface can be used by other processes
+    // the interface should be instantly reusable, e.g. after a crash
     int optval = 1;
     if (setsockopt(sockfd, SOL_SOCKET, SO_REUSEADDR, &optval, sizeof(optval)) < 0) {
-        MAPF_ERR("cannot set socket option SO_REUSEADDR \"" << strerror(errno) << "\" (" << errno
-                                                            << ").");
+        MAPF_ERR("cannot set socket option SO_REUSEADDR (errno: " << errno << " ["
+                                                                  << strerror(errno) << "])");
```
arnout, line 102: This is an unrelated change. Please do such changes in a separate commit. (nitpick)

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -137,50 +81,62 @@ void Ieee1905Transport::update_network_interfaces(
...
+        return -1;
+    }
+    return sockfd;
+}
+
+bool Ieee1905Transport::open_interface_socket(unsigned int if_index)
+{
+    MAPF_DBG("opening raw socket on interface " << if_index << ".");
+
+    if (network_interfaces_[if_index].ieee1905_socket != -1) {
```
arnout, line 128: This should be in two separate conditions.

though actually, the condition is redundant: close(-1) will return an error, but we ignore its errors anyway.

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -137,50 +81,62 @@ void Ieee1905Transport::update_network_interfaces(
...
+        network_interfaces_[if_index].ieee1905_socket = -1;
+        network_interfaces_[if_index].lldp_socket     = -1;
     }
 
-    network_interfaces_[if_index].fd = sockfd;
+    int ieee1905_socket = open_packet_socket(if_index, ETHERTYPE_IEEE1905);
+    int lldp_socket     = open_packet_socket(if_index, ETHERTYPE_LLDP);
 
+    if (ieee1905_socket < 0 || lldp_socket < 0) {
+        // if only the second call failed, the first socket still needs to be close()d
```
arnout, line 139: And vice-versa. Again, since close(-1) is OK, just close both.

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -137,50 +81,62 @@ void Ieee1905Transport::update_network_interfaces(
...
+    if (network_interfaces_[if_index].ieee1905_socket != -1) {
+        close(network_interfaces_[if_index].ieee1905_socket);
+        close(network_interfaces_[if_index].lldp_socket);
+        network_interfaces_[if_index].ieee1905_socket = -1;
+        network_interfaces_[if_index].lldp_socket     = -1;
     }
 
-    network_interfaces_[if_index].fd = sockfd;
+    int ieee1905_socket = open_packet_socket(if_index, ETHERTYPE_IEEE1905);
+    int lldp_socket     = open_packet_socket(if_index, ETHERTYPE_LLDP);
```
arnout, line 136: There is no need for temporary variables, we can write directly to network_interfaces_[if_index].ieee1905_socket.

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -230,18 +174,23 @@ void Ieee1905Transport::handle_interface_status_change(unsigned int if_index, bo
 
     MAPF_DBG("interface " << if_index << " is now " << (is_active ? "active" : "inactive") << ".");
 
-    if (!is_active && network_interfaces_[if_index].fd >= 0) {
-        poller_.Remove(network_interfaces_[if_index].fd);
-        close(network_interfaces_[if_index].fd);
-        network_interfaces_[if_index].fd = -1;
+    if (!is_active && network_interfaces_[if_index].ieee1905_socket >= 0) {
+        poller_.Remove(network_interfaces_[if_index].ieee1905_socket);
+        poller_.Remove(network_interfaces_[if_index].lldp_socket);
```
arnout, line 184: Condition should be duplicated for lldp_socket

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -230,18 +174,23 @@ void Ieee1905Transport::handle_interface_status_change(unsigned int if_index, bo
...
-    if (is_active && network_interfaces_[if_index].fd < 0) {
+    if (is_active && network_interfaces_[if_index].ieee1905_socket < 0) {
         if (!open_interface_socket(if_index)) {
             MAPF_ERR("cannot open network interface " << if_index << ".");
         }
-        if (network_interfaces_[if_index].fd >= 0)
-            poller_.Add(network_interfaces_[if_index].fd);
+        if (network_interfaces_[if_index].ieee1905_socket >= 0) {
+            poller_.Add(network_interfaces_[if_index].ieee1905_socket);
+            poller_.Add(network_interfaces_[if_index].lldp_socket);
```
arnout, line 197: Here as well, condition must be duplicated.

Yes, it's true that when ieee1905_socket opening failed you also set lldp_socket to -1, but checking them separately is more defensive.

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -230,18 +174,23 @@ void Ieee1905Transport::handle_interface_status_change(unsigned int if_index, bo
...
+        poller_.Remove(network_interfaces_[if_index].lldp_socket);
+        close(network_interfaces_[if_index].ieee1905_socket);
+        close(network_interfaces_[if_index].lldp_socket);
+        network_interfaces_[if_index].ieee1905_socket = -1;
+        network_interfaces_[if_index].lldp_socket     = -1;
     }
 
-    if (is_active && network_interfaces_[if_index].fd < 0) {
+    if (is_active && network_interfaces_[if_index].ieee1905_socket < 0) {
         if (!open_interface_socket(if_index)) {
```
arnout, line 192: Since this function now opens two sockets, perhaps it should be renamed.

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -350,7 +299,8 @@ bool Ieee1905Transport::send_packet_to_network_interface(unsigned int if_index,
 
     packet.header = {.iov_base = &eh, .iov_len = sizeof(eh)};
 
-    int fd             = network_interfaces_[if_index].fd;
+    int fd = packet.ether_type == ETHERTYPE_LLDP ? network_interfaces_[if_index].lldp_socket
```
arnout, line 307: Maybe this should be a switch and should error out if it is neither.

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -83,10 +83,14 @@ void Ieee1905Transport::update_network_interfaces(
 
 int Ieee1905Transport::open_packet_socket(unsigned int if_index, uint16_t ether_type)
 {
+    // Note to developer: The current implementation uses AF_PACKET socket with SOCK_RAW protocol
+    // which means we receive
+    // and send packets with the Ethernet header included in the buffer. Please consider changing
+    // implementation to use SOCK_DGRAM protocol (without L2 header handling)
```
arnout: I see no reason to switch to SOCK_DGRAM.

PR #648, opened 2020-01-07
Feature/wave api rdkb adaptations

morantr:
Add adaptations to compile prplmesh on RDKB tree.


controller/src/beerocks/bml/bml.cpp
```diff
@@ -556,3 +556,77 @@ int bml_channel_selection(BML_CTX ctx, const char *al_mac, const char *ruid)
...
 }
+
+int bml_set_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int enable)
+{
+    // Validate input parameters
+    if (!ctx) {
+        return (-BML_RET_INVALID_ARGS);
+	}
+
+	// TODO: call suit bml api
```
tomereli: call suitable (nitpick)

controller/src/beerocks/bml/bml_defs.h
```diff
@@ -549,6 +553,45 @@ struct BML_STATS_ITER {
     struct BML_STATS *(*get_node)();
 };
 
+struct BML_DCS_NEIGHBOR_AP {
+    //char  ap_Radio[64];
+    //The Radio that detected the neighboring WiFi SSID.
+    char ap_SSID[64];
```
tomereli, line 559: Please use static const definitions instead of hard coded numbers

controller/src/beerocks/bml/bml.cpp
```diff
@@ -556,3 +556,77 @@ int bml_channel_selection(BML_CTX ctx, const char *al_mac, const char *ruid)
...
+                                       int interval_time, unsigned int *channel_pool,
+                                       int channel_pool_size)
+{
+    // Validate input parameters
+    if (!ctx) {
+        return (-BML_RET_INVALID_ARGS);
+	}
+
+	// TODO: call suit bml api
+	return 0;
```
tomereli: I think that we should return failure until these stubs are implemented. This goes to all the stubs below and above this one. and also add LOG.

tomereli (changes requested):
Last commit can't be merged - it was done by @arnout  and @rmelotte for the CI, we need to find a different solution.

morantr:
@tomereli There is no point changing what I did. Itay should change it very soon.

tomereli:
> @tomereli There is no point changing what I did. Itay should change it very soon.

Not sure which comment you mean but the changes requested are mainly due to the last commit. @rmelotte added the soft link and **this is why your PR failed CI.**

controller/src/beerocks/bml/bml.cpp
tomereli: This MUST return failure, I don't care if @itayx will change it soon - cause until he does, the API should not report success when it's not implemented..

controller/src/beerocks/bml/bml_defs.h
tomereli: This should also be changed now cause I doubt that @itayx already changed it in his branches.

rmelotte:
> Not sure which comment you mean but the changes requested are mainly due to the last commit. @rmelotte added the soft link and this is probably why your PR failed CI.

The soft-link was there to make it possible to use any docker image regardless of it's default entrypoint and not break the builds for anyone.
The plan was to remove it after the required changes made their way to all the development branches (it should be the case now).
The reason it's still failing is because the images on dockerhub haven't been updated.
However I think it would be better to just override the entrypoint whenever we run the image (from the build script for instance).
Also if we decided to update the images on dockerhub at some point, we might just as well remove the entrypoint entirely, since it makes things easier for the CI.

I'll make a patch and see if I didn't miss something.

controller/src/beerocks/bml/bml_defs.h
itayx: Actually I did

tomereli: So OK :)

tomereli (approved):

rmelotte:
The last 2 commits were intended to be squashed together because removing the soft-link without any other changes won't work (unless the images on dockerhub are updated to reflect this specific change, but I don't think it's worth it to do that).

PR #651, opened 2020-01-08
Bugfix/cmdu tx load fix2

### LINKED ISSUE #622: UCC listener - dev_send_1905 hangs

tomereli:
he CmduMessageTx::load() API is used only for certification mode for
sending a prefilled CMDU when instructed to by the testbed UCC.

It got broken somewhere when we moved to ClassList and the current
implementation is now broken since load is only adding the first class
(CmduHeader) then on Finalize EOM TLV is added in the middle of the
CMDU overriding the actual buffer contents.

Instead of the whole certification buffer hack, we simply add another `CmduMessageTx` for certification only - `cert_cmdu_tx` and `cert_tx_buffer` to the beerocks socket thread base class, which is then passed as reference to the UCC listener constructor.

This allows us, when in certification mode, to use the certification cmdu instead of the regular one when sending CMDUs which needs to be resent on UCC application trigger, instead of saving the buffer and trying to load the CMDU.

While doing so, we're also fixing the affected flows to always send the CMDU regardless of certification, so in fact for the certification flow the same message will be sent twice.

This way, we can get rid of the if-else-if in the ucc listener and simple test which checks if the cmdu is finalized and send it as is, removing the need for create_cmdu API which is also removed.

Fixes #622


common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
```diff
@@ -69,6 +70,7 @@ class socket_thread : public thread_base {
 
     uint8_t rx_buffer[message::MESSAGE_BUFFER_LENGTH];
     uint8_t tx_buffer[message::MESSAGE_BUFFER_LENGTH];
+    uint8_t cert_tx_buffer[message::MESSAGE_BUFFER_LENGTH];
```
morantr, line 73: It doesn't seem right to define it here. This is only relevant to the controller, and maybe will be needed to agent certification as well.
I think that it will be more correct to define it on the controller/backhaul since they are the only ones that use it, or at least define it on the BTL which the controller and the backhaul are derived from.

Anyway I would have defined only a pointer to the buffer and cert_tx_buffer and allocated or runtime id the ucc_listener is up (port!=0) or certification_mode=1.

framework/tlvf/src/include/tlvf/CmduMessageTx.h
```diff
@@ -25,7 +25,7 @@ class CmduMessageTx : public CmduMessage {
     std::shared_ptr<cCmduHeader> load();
     std::shared_ptr<tlvVendorSpecific> add_vs_tlv(tlvVendorSpecific::eVendorOUI voui);
     template <class T> std::shared_ptr<T> addClass() { return msg.addClass<T>(); }
-
+    void reset() { msg.reset(false); }
```
morantr, line 28: What you wrote in the commit message is not true: 
We do not send the cmdu twice. If the we are on certification, we hold the cmdu without sending it and waiting for the ucc_listener to send it, otherwise, we send it.
At this point in the review I can't see how you use it yet, but the assumption on the commit message that we send the cmdu twice, makes me think that this commit should be dropped.

common/beerocks/bcl/include/bcl/beerocks_ucc_listener.h
```diff
@@ -106,7 +106,7 @@ class beerocks_ucc_listener : public socket_thread {
     friend class tlvPrefilledData;
 
     // Variables
-    ieee1905_1::CmduMessageTx &m_cmdu_tx;
+    ieee1905_1::CmduMessageTx &m_cmdu;
```
arnout: Why do you change the name of m_cmdu_tx? It's still a CmduMessageTx, and although here we never use an _rx message I still think it makes sense to make it explicit.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -802,29 +802,13 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
...
-            reply_ucc(eWfaCaStatus::ERROR, err_internal);
-            break;
-        }
-        ieee1905_1::CmduMessageTx external_cmdu_tx(
-            certification_tx_buffer.get() + sizeof(beerocks::message::sUdsHeader),
-            message::MESSAGE_BUFFER_LENGTH - sizeof(beerocks::message::sUdsHeader));
-        if (create_cmdu(external_cmdu_tx, ieee1905_1::eMessageType(message_type), err_string)) {
-            // Success
-            if (!send_cmdu_to_destination(external_cmdu_tx, dest_alid)) {
+        if (m_cmdu.is_finalized()) {
```
arnout: Not strictly related to this PR, but I think we should also check here if the message type is the same as the one we're expected to send (and BTW also that tlv_hex_list is empty). It is conceivable that we have a prepared certification buffer but that we don't need to send it yet. 

That's not so easy though, because it means that we need two cmdu_tx here as well: one for the certification buffer, and a separate one for constructed messages.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -812,7 +812,8 @@ bool master_thread::handle_cmdu_1905_channel_preference_report(const std::string
     // to the mid sent in channel preference request message
 
     // build channel request message
-    if (!cmdu_tx.create(0, ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE)) {
+    auto &cmdu = database.setting_certification_mode() ? cert_cmdu_tx : cmdu_tx;
+    if (!cmdu.create(0, ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE)) {
```
arnout: Actually, we only need to build the certification buffer in certification mode. So we simply need to return true in non-certification mode, and always use cert_cmdu_tx.

Later, when we'll actually use this message instead of VS messages, it will be stored in the database. In that case, we don't need to create the CMDU in non-certification mode. But on the other hand, we could just do it always since we anyway won't send it. 

While I'm at it: the commit is called "channel preference report refactor" but it's actually "channel selection for certification refactor".

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -935,17 +936,7 @@ bool master_thread::handle_cmdu_1905_channel_preference_report(const std::string
...
-        if (!certification_tx_buffer) {
-            LOG(ERROR) << "certification_tx_buffer is not allocated!";
-            return false;
-        }
-        database.fill_certification_tx_buffer(cmdu_tx);
-        return true;
-    }
-
-    return son_actions::send_cmdu_to_agent(src_mac, cmdu_tx, database);
+    return son_actions::send_cmdu_to_agent(src_mac, cmdu, database);
```
arnout: This should simply not be done in certification mode, right?

And in non-certification mode, it shouldn't be done either (instead, it should be part of the channel selection task, so basically we should store the channel preferences in the database and send an event to the channel selection task).

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
tomereli: I could split it to the controller (son_master_thread) and agent (backhaul_manager_thread) but I think it is pointless and wasted effort since this is anyway a hack which we should remove eventually once we have a real channel selection task which will store all the needed parameters in the DB and then we will re-trigger channel selection and be done with it, as @arnout suggested multiple times.

So for now - I think there is no harm with keeping the certification stuff in the base class. Its even better since we'll have less code to remove once we go with the correct solution.

framework/tlvf/src/include/tlvf/CmduMessageTx.h
tomereli: This commit is necessary even if we don't send it twice - since it is used by later commit to reset the certification cmdu which is like "freeing" it so it won't be used endlessly.

Regarding the sending of the same cmdu twice - I disagree - @arnout and I discussed this and came to the conclusion that we should not behave differently in certification mode.
This means that on every flow we should proceed as usual, and in the flows where we need to send a CMDU at a later time for the UCC sniffer to catch it, we will just send it a second time.

@arnout Please correct me if I'm wrong.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1204,13 +1204,13 @@ bool master_thread::handle_cmdu_1905_link_metric_response(const std::string &src
 
     print_link_metric_map(link_metric_data);
 
-    if (database.setting_certification_mode()) {
-        return construct_combined_infra_metric();
-    }
-    return true;
+    auto &cmdu = database.setting_certification_mode() ? cert_cmdu_tx : cmdu_tx;
```
arnout: Again, in non-certification mode, we don't need to do anything with this message (except store it in the database). So always use cert_cmdu_tx.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -866,6 +850,7 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
 
         // Send back second reply
         reply_ucc(eWfaCaStatus::COMPLETE, description);
+        m_cmdu.reset();
```
arnout: The reset should also be done in case of failure.

arnout (approved):
My comments are very important conceptually, but change very little in the code. I do think the commit messages of commits "add cmdu_tx::reset" and "add cmdu_tx to constructor" need to be adapted a little to reflect that though. Basically, we only want to send those CMDUs in certification mode, and even then we don't want to send them twice.

So, for the reset, it should be something like: "Normally, the ClassList is reset automatically when a new CMDU is created with CmduMessageTx::create. For certification, however, we want to be able to reset it explicitly, to detect when the certification buffer is unset. Therefore add a public reset() member to CmduMessageTx."

For the cmdu_tx in constructor: "Up to now, we detect whether to use the certification buffer or to construct a new message by calling the get_buffer_filled_with_cmdu() virtual method. However, this requires us to be able to load a CmduMessageTx from a pre-filled buffer, which is normally not possible. This can be simplified a lot by constructing the messages directly in a CmduMessageTx which is passed to the UCC listener.

Pass the CmduMessageTx corresponding to the certification buffer in the beerocks_ucc_listener constructor. Use its is_finalized state to detect if it has been filled or not. Use the same buffer also for messages constructed directly in the UCC listener. Reset it after sending the message."

Finally, I still believe it would be better to just construct the message based on information in the database. But that's more work, and we need to progress on this issue, so let's just merge this series.

There are some important changes requested, but it's all pretty obvious so approving already.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
tomereli: I added the check so now we consider the prefilled cmdu only if its ready AND matches the message type.

common/beerocks/bcl/include/bcl/beerocks_ucc_listener.h
tomereli: Changed back

framework/tlvf/src/include/tlvf/CmduMessageTx.h
tomereli: OK so apparently I *was* wrong. Fixed this as @morantr and @arnout suggested.

PR #652, opened 2020-01-08
common: docker Add tcpdump support for tests

ghost:
I'd like to take a dump of traffic over the virtual docker interfaces
    when developing my new feature. This works best however if it's built
    into the test script so tcpdump is started and stopped at the right
    time.

### DIFF
--- a/tools/docker/tests/test_flows.sh
+++ b/tools/docker/tests/test_flows.sh
@@ -20,6 +20,28 @@ error=0
 
 bridge_name=br-lan
 
+get_host_bridge_name() {
+    docker network inspect prplMesh-net --format="br-{{ printf \"%.12s\" .Id }}"
+}
+
+start_tcpdump() {
+    test="$1"
+    if [ "$TCPDUMP" = "true" ]; then
+        # -i: interface to capture on; -w: output file
+        tcpdump -i $(get_host_bridge_name) -w test_${test}.pcap &
+        TCPDUMP_PROC=$!
+        status "started tcpdump: pid=${TCPDUMP_PROC}, output file=test_${test}.pcap"
+    fi
+}
+
+kill_tcpdump() {
+    if [ -n "$TCPDUMP_PROC" ]; then
+        status "Terminating tcpdump"
+        kill $TCPDUMP_PROC;
+    fi
+    unset TCPDUMP_PROC
+}
+
 container_ip() {
     docker exec -t "$1" ip -f inet addr show "$bridge_name" | grep -Po 'inet \K[\d.]+'
 }
@@ -555,12 +577,16 @@ test_topology() {
 test_init() {
     status "test initialization"
 
+    start_tcpdump init
+
     [ "$SKIP_INIT" = "false" ] && {
         eval ${scriptdir}/test_gw_repeater.sh -f -r "repeater1" -r "repeater2" -d 7 $redirect || {
             err "start GW+Repeater failed, abort"
             exit 1
         }
     }
+    kill_tcpdump
+
     #save bml_conn_map output into a file.
     connmap=$(mktemp)
     [ -z "$connmap" ] && { err "Failed to create temp file"; exit 1; }
@@ -589,9 +615,10 @@ test_init() {
 
 }
 usage() {
-    echo "usage: $(basename $0) [-hv]"
+    echo "usage: $(basename $0) [-hvds]"
     echo "  options:"
     echo "      -h|--help - show this help menu"
+    echo "      -t|--tcpdump - Capture the results with tcpdump (requires root (sudo) or special permission/capability management of interfaces)"
     echo "      -v|--verbose - verbosity on"
     echo "      -s|--stop-on-failure - stop on failure"
     echo "      --skip-init - skip init"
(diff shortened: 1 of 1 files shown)


tools/docker/tests/test_flows.sh
```diff
@@ -20,6 +20,15 @@ error=0
 
 bridge_name=br-lan
 
+get_host_bridge_name() {
+	ip l | grep $(docker network ls | grep prplMesh-net | cut -d' ' -f1) | perl -n -e '/\d+: ([^:]+)/ && print $1."\n"' | grep "^br-"
+}
```
rmelotte, line 25: You can get the ID you're looking for using docker network directly:
```
docker network inspect prplMesh-net --format="{{ printf \"%.12s\" .Id }}"
```
Then you just need to append the 'br-'.

tools/docker/tests/test_flows.sh
```diff
@@ -664,5 +681,8 @@ main() {
 VERBOSE=false
 SKIP_INIT=false
 STOP_ON_FAILURE=false
+DUMP=false
+DUMPFILE=-
+TCPDUMP_PROC=
```
rmelotte: You don't really need to set a default value for `TCPDUMP_PROC` since it's either unused or set by the script. Same for `DUMPFILE`.

(nitpick)

rmelotte: Also why is the default output file stdout? I would think that there are other stuff on stdout that would prevent it to be used correctly, but I didn't try.

rmelotte (approved):
There are tabs almost everywhere, please convert them to spaces.
The rest is only small comments, so approving.

Edit: the commit message is also missing _what_ changes are done and how it can be used.

tools/docker/tests/test_flows.sh
ghost: Good point, thanks! I'm still wondering why `docker network ls` doesn't have an option to show the host side name of the network interfaces.
Updated the PR.

tools/docker/tests/test_flows.sh
ghost: Ok, changed

tools/docker/tests/test_flows.sh
```diff
@@ -615,13 +625,14 @@ usage() {
...
 main() {
-    OPTS=`getopt -o 'hvs' --long help,verbose,skip-init,stop-on-failure -n 'parse-options' -- "$@"`
+    OPTS=`getopt -o 'd:hvs' --long dump:,help,verbose,skip-init,stop-on-failure -n 'parse-options' -- "$@"`
     if [ $? != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
 
     eval set -- "$OPTS"
 
     while true; do
         case "$1" in
+	    -d | --dump)	    DUMP=true; DUMPFILE=$2; shift; shift ;;
```
tomereli: looks like wrong indentation

tomereli: I prefer `TCPDUMP=true; TCPDUMP_FILE=$2` (nitpick)

tools/docker/tests/test_flows.sh
```diff
@@ -631,6 +642,12 @@ main() {
         esac
     done
     [ -z "$*" ] && set ${ALL_TESTS}
+    if [ "$DUMP" = "true" ]; then
```
tomereli: I think it is more appropriate to move this into a separate function

tools/docker/tests/test_flows.sh
```diff
@@ -631,6 +642,12 @@ main() {
...
+	    tcpdump -i $(get_host_bridge_name) -U -w $DUMPFILE &
```
tomereli: it would be useful to comment the tcpdump options used in this command (nitpick)

tools/docker/tests/test_flows.sh
```diff
@@ -664,5 +681,8 @@ main() {
 VERBOSE=false
 SKIP_INIT=false
 STOP_ON_FAILURE=false
+DUMP=false
+DUMPFILE=-
```
tomereli: Is this a typo? Why DUMPFILE=*-*?

tools/docker/tests/test_flows.sh
```diff
@@ -664,5 +681,8 @@ main() {
 VERBOSE=false
 SKIP_INIT=false
 STOP_ON_FAILURE=false
+DUMP=false
```
tomereli: I think the default should be `true`. I also think it may be useful to have a separate dump file for each test under tcpdump directory. Then you can get rid of the DUMPFILE parameter (what is it used for?).
This will be extremely helpful IMO to have a separate dump file for each test. But maybe that's just my opinion so (nitpick)

tomereli (approved):
I have quite a lot of comments but since this is already an improvement than approving. Please consider my comments before merging, especially the one about having separate dump files for each test.

tools/docker/tests/test_flows.sh
ghost: This issue was addressed in a followup commit; you're not looking at the latest version

`-` is generally interpreted as standard output, which I agree was a mistake to put as a default.

tools/docker/tests/test_flows.sh
ghost: Good idea, I can see that being quite useful. It can be difficult to tell based on the capture alone where one test stops and the next one starts, so one file per test sounds like a really good idea.

tools/docker/tests/test_flows.sh
ghost: OK

tomereli (approved):

tools/docker/tests/test_flows.sh
```diff
@@ -555,12 +566,16 @@ test_topology() {
 test_init() {
     status "test initialization"
 
+    [ "$TCPDUMP" = "true" ] && tcpdump -i $(get_host_bridge_name) -w test_init.pcap &
```
arnout: `TCPDUMP_PROC=$!` missing?

tools/docker/tests/test_flows.sh
```diff
@@ -20,6 +20,17 @@ error=0
...
 bridge_name=br-lan
 
+get_host_bridge_name() {
+    docker network inspect prplMesh-net --format="br-{{ printf \"%.12s\" .Id }}"
+}
+
+kill_tcpdump() {
+    if [ -n "$TCPDUMP_PROC" ]; then
+        status "Terminating tcpdump"
+        kill $TCPDUMP_PROC;
```
arnout, line 40: Redundant semicolon (are you a C programmer? :-)

TCPDUMP_PROC should also be unset or set to empty after the kill, so we don't kill it twice (and the second time it may already be a new process...)

arnout (approved):
Important fixes but approving already since it should be fine after that.

As commented orally: we should add tcpdump to our docker-in-docker image, and enable this in CI.

PR #657, opened 2020-01-10
controller: client_steering_task: wait for disconnect and connect

### LINKED ISSUE #664: [BUG] controller: client_steering: doesn't report success
Sometimes the client steering task doesn't report success. There are several issues here:

- it doesn't wait for both CONNECTED and DISCONNECTED events, but only reports success if CONNECTED is received, so it depends on the order of events;
- sometimes events seem to be lost.

arnout:
The steering task is considered successful when the connected event for
the STA is received. However, we were not waiting for that event, so if
the disconnected event arrives before the connected event, we would
first jump to the FINALIZE state and report that steering has failed.

Add STA_CONNECTED to the events that we're waiting for.

Add test to make sure this works.

Fixes #664

### DIFF
--- a/controller/src/beerocks/master/son_actions.cpp
+++ b/controller/src/beerocks/master/son_actions.cpp
@@ -312,21 +312,19 @@ void son_actions::handle_dead_node(std::string mac, std::string hostap_mac, db &
     if (parent_hostap_mac == hostap_mac) {
         if (mac_type == beerocks::TYPE_IRE_BACKHAUL || mac_type == beerocks::TYPE_CLIENT) {
             database.set_node_state(mac, beerocks::STATE_DISCONNECTED);
+            // Notify steering task, if any, of disconnect.
+            int steering_task = database.get_steering_task_id(mac);
+            if (tasks.is_task_running(steering_task))
+                tasks.push_event(steering_task, client_steering_task::STA_DISCONNECTED);
+
             if (database.get_node_handoff_flag(mac)) {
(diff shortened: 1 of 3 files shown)


rmelotte:
What happens if the client first connects to the next AP before disconnecting from the current one?
Will the connect event still be handled or will the steering task wait until another connect event occurs?

I think both cases should be tested by `test_flows.sh`: 
- the client first connects to the next AP, then disconnects from the current one.
- the client first disconnects from the current one, then connects to the next one. 

We need to know (as in: test in CI) what happens in both cases: will the steering task hang/fail/succeed?

arnout:
> I think both cases should be tested by `test_flows.sh`:
> 
>     * the client first connects to the next AP, then disconnects from the current one.
> 
>     * the client first disconnects from the current one, then connects to the next one.

You're completely right. I updated the test to reliably test one of the two cases, and now if tails always even after the fix :-). So that will have to be fixed as well. Marking as don't merge until that is done.

arnout:
[Client steering doesn't correctly report success](https://trello.com/c/0R3DY5N3/66-client-steering-doesnt-correctly-report-success)

morantr (approved):

tomereli (approved):

rmelotte (approved):
I created a new task to test "connect before disconnect" (#685) so that we can already merge this one.

arnout:
I rebased and re-tested each individual commit. The test now consistently fails after the first commit, still fails after the second commit, and succeeds after the third one.

PR #658, opened 2020-01-10
controller: task: simplify event_received

arnout:
event_received() removes the incoming event from pending_events and then 
calls the pure virtual handle_event() callback. However, the way the 
incoming event is removed is more than a little weird:

- first a range is created over all pending events of that type;
- we iterate over the range but break at the first iteration;
- we only erase it if the iterator matches event_type, but since the range
is already matching event_type, this will always be true.

So simplify this logic by using find() instead.

While we're at it, log a message if an event received that was not pending - it's either an unexpected event, or it should have been added to
pending_events.

*This needs to be tested with actual controller flows executing to be sure that it doesn't break anything.*

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2215,21 +2215,26 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
      * Set the number of local interfaces and fill info of each of the local interfaces, according
      * to IEEE_1905 section 6.4.5
      */
-    uint32_t speed;
-    if (network_utils::linux_iface_get_speed(m_sConfig.wire_iface, speed)) {
+
+    /**
+     * Add a LocalInterfaceInfo field for the wired interface, if any.
+     */
+    std::string local_interface_name = m_sConfig.wire_iface;
+    if (!local_interface_name.empty() &&
+        network_utils::linux_iface_is_up_and_running(local_interface_name)) {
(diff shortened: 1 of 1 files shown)


morantr (approved):

arnout:
@morantr @tomereli Should I just merge this PR, or can someone at Intel check if all controller functionality still works? We don't have any controller flow tests in CI so we really don't know if this breaks something...

tomereli:
> @morantr @tomereli Should I just merge this PR, or can someone at Intel check if all controller functionality still works? We don't have any controller flow tests in CI so we really don't know if this breaks something...

We don't have bandwidth at the moment to test this on real controller. Can this wait until we have test flows working on RAX40?

arnout:
> > @morantr @tomereli Should I just merge this PR, or can someone at Intel check if all controller functionality still works? We don't have any controller flow tests in CI so we really don't know if this breaks something...
> 
> We don't have bandwidth at the moment to test this on real controller. Can this wait until we have test flows working on RAX40?

Sure, it's just a simplification. We can keep this open until we can test it.

controller/src/beerocks/master/tasks/task.cpp
```diff
@@ -45,14 +45,11 @@ void task::response_received(std::string mac,
...
-            it = pending_events.erase(it);
-            break;
-        } else {
-            it++;
-        }
+    auto it = pending_events.find(event_type);
+    if (it == pending_events.end()) {
+        TASK_LOG(DEBUG) << "received non-pending event " << event_type;
+    } else {
+        pending_events.erase(it);
```
morantr: I think that originally the function was supposed to remove all the pending events on the range, so actually the the `break` statement is probably not in place. But I guess there is a good reason that it was added so your change seems ok.

Anyway, these commits have already merged a part of #659.

ghost:
@arnout to verify that this can be closed

PR #659, opened 2020-01-10
framework: common: remove logging.conf

arnout:
This file actually gets generated in CMAKE_BINARY_DIR but was accidentally
committed at some point (before git history). Remove it.


tomereli:
I think this can be merged without the second commit

tomereli (approved):
Until we can test it, the second commit should be removed from this PR so we can merge it. Please do that before merging.

PR #661, opened 2020-01-12
Bugfix/config data attrlist

tomereli:
The configData is the un-encrypted data stored in the encrypted settings attribute in the M2 WSC message and includes the credentials which the agent is requested to configure itself for each fronthaul and backhaul VAPs.

Currently, the config data is created and parsed using the cConfigData class which has fixed attributes order and size.

According to the WSC 2.0 specification - the version extension attribute may contain 1 or more subelements. Different controllers than prplMesh sends more than 1 element which causes our agent to fail on keywrap authentication stage since it looks for the keywrapauth attribute immediately at the end of the configData which is smaller than it actually is.

The solution is to refactor the way we create and parse the config data - the config data, similar to M1 and M2, is a list of attributes, so this PR first refactors the AttrList to be generic so it can be extended to support configData, then adds configData derived class which implements create and parse similar to the way M1 and M2 are implemented.

We can now use a class to represent the vendor extension attribute - so we also update it to include an unknown length list at the end, after the known MultiAP subelement, so it will be automatically parsed correctly on RX.

Using this, we can officially pass autoconfig stage as an agent in the 4.2.1 test case.


tomereli:
[Fix autoconfig bug with certified controller](https://trello.com/c/6ROkLj3T/62-fix-autoconfig-bug-with-certified-controller)

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
```diff
@@ -174,6 +174,15 @@ cWscVendorExtWfa:
...
     _value: WSC_VENDOR_ID_WFA_3
+  subelement_id:
+    _type: uint8_t
+    _value: 0x6
+  subelement_length:
+    _type: uint8_t
+    _value: 0x1
+  subelement_value:
+    _type: uint8_t
+    _value: TEARDOWN
```
arnout, line 185: This is not correct. cWscVendorExtWfa is the vendor extension attribute with WFA as vendor ID. The subelements are in vs_data. That's exactly what is intended to be able to have more subelements in the vendor extension. So we should either make sWscAttrVendorExtMultiAp a class and and additional vs data to it (but then there's no guarantee that the multiap subelement is going to be the first one, so still a hack), or (my preference) remove sWscAttrVendorExtMultiAp and use this cWscVendorExtWfa instead, and parse the vs_data properly. In either case, this should not be added to the cWscVendorExtWfa.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3856,7 +3856,7 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
 
     std::list<WSC::m2> m2_list;
     for (auto tlv : cmdu_rx.getClassList<ieee1905_1::tlvWsc>()) {
-        auto m2 = std::dynamic_pointer_cast<WSC::m2>(WSC::AttrList::parse(*tlv));
+        auto m2 = WSC::m2::parse(*tlv);
```
arnout, line 3858: Yes! This is how I intended it to be from the beginning!

framework/tlvf/src/include/tlvf/WSC/m1.h
```diff
@@ -16,11 +16,35 @@ namespace WSC {
...
+        uint8_t pub_key[WSC_PUBLIC_KEY_LENGTH];
+        uint16_t encr_type_flags;
+        uint16_t auth_type_flags;
+        std::string manufacturer;
+        std::string model_name;
+        std::string model_number;
+        std::string serial_number;
+        uint16_t primary_dev_type_id;
+        std::string device_name;
+        eWscRfBands bands;
```
arnout, line 35: I would have kept the common part (i.e. this bit) in the AttrList class, as well as the getter wrappers for them, and I'd have kept a protected AttrList::create() function that takes the common config. I would have passed the additional configuration as separate parameters to the (public) m1 and m2 create() functions (so the two would have a different signature).

However, after seeing the configData, I kind of like how you did it: every type of attribute list has a number of expected attributes, and you can provide all of them to the factory method (create()) in one shot, and the structure is always called `config`.

framework/tlvf/src/src/WSC/m1.cpp
```diff
@@ -12,6 +12,46 @@
...
+    auto msg_type = wsc_msg_type_from_buffer(tlv.payload(), tlv.payload_length());
+    if (msg_type != eWscMessageType::WSC_MSG_TYPE_M1) {
+        TLVF_LOG(DEBUG) << "Not M1 message! message type " << int(msg_type);
+        return nullptr;
+    }
+    auto attributes = std::make_shared<m1>(tlv.payload(), tlv.payload_length(), true);
+    if (!attributes || !attributes->init()) {
+        TLVF_LOG(ERROR) << "Failed to initialize attributes";
+        return nullptr;
+    }
```
arnout: IMO it would be much more natural to do the following:

```
auto attributes = std::make_shared<m1>(...);
attributes->init();
auto msg_type = attributes->getClass<cWscAttrMessageType>();
if (msg_type->msg_type != WSC_MSG_TYPE_M1) {
    LOG(INFO) << "Not M1, msg type is " << int(msg_type->msg_type);
    return nullptr;
}
```

That avoids the extra parsing of msg_type. Since we return nullptr, the invalid attrlist (which pretends to be an M1 but really is not) will be freed immediately, so we never expose an invalid M1 object.

framework/tlvf/src/src/WSC/m1.cpp
```diff
@@ -183,8 +223,84 @@ bool m1::init(const config &cfg)
     return true;
 }
 
-bool m1::valid_custom() const
+bool m1::valid() const
```
arnout, line 226: Is this validation really worthwhile? Do we really need to fail if an agent/controller doesn't provide a version attribute?

For the attributes that we really use, there is already implicit validation in the agent/controller code that handles the M2/M1.

Oh, hang on, we don't actually validate those things in the handler code because we call wrapper functions that dereference the attributes. However, aren't those wrappers going to trigger klockwork errors because they dereference a potentially nullptr? If they do, I think we need to add an assert in those wrapper functions. Actually, I think we need to add asserts in many places in the code :-) To be discussed...

framework/tlvf/src/src/WSC/AttrList.cpp
```diff
@@ -208,10 +208,11 @@ bool AttrList::init()
...
             }
             break;
-        // Other attributes are not expected, if so ignore them
+        // Other attributes are not expected, if so ignore them silently
         default:
-            TLVF_LOG(ERROR) << "Unknown attribute " << getNextAttrType();
-            break;
+            TLVF_LOG(DEBUG) << "Unknown attribute " << getNextAttrType()
+                            << " assuming end of AttrList";
+            return true;
```
arnout, line 215: There really is no reason why an unkown attribute should be at the end of the attrlist...

However, the WSC specification is pretty strict. There simply shouldn't be an unknown attribute. If there is, it's an error. So I think a practical approach would be to return false here. Or did you see a certification device with an unknown attribute?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3726,17 +3726,21 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
     // cConfigData constructor, but then parsing will fail since the length
     // will be calculated wrong. TLVF does not support parsing network byte order
     // without full swap, so keep this workaround for now (another future TLVF V2 feature)
-    WSC::cConfigData config_data(decrypted, datalen, true);
+    auto config_data = WSC::configData::parse(decrypted, datalen);
```
arnout, line 3729: I wonder if it is really worthwhile to add a configData class for this. We could just as well directly use AttrList, IMO. But OK, the code is there now, so why not.

arnout (approved):
As usual, a lot of comments. I think the cWscVendorExtWfa bit is pretty essential, but since it actually works with the current approach, it's OK for me to merge even if that is not changed - it's something we can rework later.

So the only thing that really should be looked at is the parse function and avoiding the use of wsc_msg_type_from_buffer. I trust that that will go well, so I'm approving already.

framework/tlvf/src/src/WSC/AttrList.cpp
tomereli: No, there should not be any unknown attributes.

The problem is that we pass a buffer size which is larger than the size of the config data. So the while loop will eventually reach the end of the buffer, which has "garbage" data. So I actually used it to find the end of the list.

framework/tlvf/src/src/WSC/m1.cpp
tomereli: The validation is exactly for this - since our wrappers are accessing attributes directly.
KW is not an issue - we will mark these as non-issues.

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
tomereli: The correct solution is to add the subelements as another AttrList IMO but I thought it would be too much work, what do you think?

framework/tlvf/src/src/WSC/m1.cpp
tomereli: Agree, will be changed.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: It is convenient to have a class for this, it was really easy to write since it is very similar to M1 and M2. Also, AttrList is abstract, so we would have needed to change that.
My original design was that AttrList will never be instantiated by itself, and for every list, there will be a derived class with simple getters, validation function, and configuration struct which are used to create and parse such lists, so this is how I went on with configData class.

framework/tlvf/src/src/WSC/m1.cpp
tomereli: I guess we can be less strict if we see that some controller don't send some attributes, but for now I think its OK to be strict - we are just starting to test with other controllers so I guess we'll find out.

framework/tlvf/yaml/tlvf/WSC/WSC_Attributes.yaml
tomereli: From our slack chat - 
> it's not the right approach...
It is possible that the vendor extension is not a WFA extension, but an actual vendor extension.
And we *will* probably encounter this in practice.
Therefore, the vendor extension contents really has to be a dynamic list of uint8_t.
So the best is probably to "manually" parse it by looking at the length and matching the type (and stopping as soon as well encounter the multiap extension - remember we don't need to parse the version2).
Parsing the subelements is really trivial, so I think it's easy enough to do it manually.
And for creating them, we can use structures that are just copied. So `vendor_ext_addr->add_vs_data(sizeof(cWscVendorExtSubElement))` and then `std::copy` into the buffer. Or just set it byte by byte.

So the bottom line is that this needs to be done correctly, but as you wrote in https://github.com/prplfoundation/prplMesh/pull/661#pullrequestreview-341749380 we can do it in a separate, followup PR, and will merge this one so we can unblock agent certification.

tomereli: Opened a task for this - #667

PR #666, opened 2020-01-13
Bugfix: last Klocwork warnings in common

Galvien:
Fix some warnings and borderline errors found by Klocwork.
The rest of the warnings are false positives and should be updated to be ignored.

### DIFF
--- a/framework/common/encryption.cpp
+++ b/framework/common/encryption.cpp
@@ -90,7 +90,7 @@ diffie_hellman::diffie_hellman() : m_dh(nullptr), m_pubkey(nullptr)
 
 diffie_hellman::~diffie_hellman()
 {
-    delete m_pubkey;
+    delete[] m_pubkey;
     DH_free(m_dh);
 }
 
--- a/common/beerocks/bcl/include/bcl/network/socket.h
+++ b/common/beerocks/bcl/include/bcl/network/socket.h
@@ -64,11 +64,7 @@ public:
     void closeSocket();
     bool isOpen();
     SOCKET getSocketFd() { return m_socket; }
-    std::string getError()
-    {
-        return m_error;
-        m_error.clear();
-    }
+    std::string getError() { return m_error; }
     std::string getPeerIP() { return m_peer_ip; }
     int getPeerPort() { return m_peer_port; }
     std::string getUdsPath() { return m_uds_path; }
--- a/framework/common/include/mapf/common/encryption.h
+++ b/framework/common/include/mapf/common/encryption.h
@@ -83,6 +83,8 @@ private:
     uint8_t *m_pubkey        = nullptr;
     unsigned m_pubkey_length = 0;
     uint8_t m_nonce[16];
+
+    diffie_hellman(const diffie_hellman &) = delete;
 };
 
 /**


common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,54 +23,71 @@ template <class T> class promise {
...
+        int err;
+        if ((err = pthread_mutex_lock(&m_mut)) != 0) {
+            LOG(ERROR) << "pthread_mutex_lock failed, error code: " << err;
+            return;
+        }
         m_value  = value;
         m_signal = 1;
-        pthread_cond_broadcast(&m_cond);
-        pthread_mutex_unlock(&m_mut);
+        if (pthread_cond_broadcast(&m_cond) != 0)
```
tomereli: ` if ((err = pthread_cond_broadcast(&m_cond)) != 0)`

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,54 +23,71 @@ template <class T> class promise {
...
+            LOG(ERROR) << "pthread_cond_broadcast failed, error code: " << err;
+        if (pthread_mutex_unlock(&m_mut) != 0)
```
tomereli: `if ((err = pthread_mutex_unlock(&m_mut)) != 0)`

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,54 +23,71 @@ template <class T> class promise {
...
             m_signal = 0;
         } else {
-            pthread_cond_wait(&m_cond, &m_mut);
+            if ((err = pthread_cond_wait(&m_cond, &m_mut)) != 0) {
+                LOG(ERROR) << "pthread_cond_wait failed, error code: " << err;
+                return;
+            }
         }
-        pthread_mutex_unlock(&m_mut);
+        if (pthread_mutex_unlock(&m_mut) != 0)
```
tomereli: `if ((err = pthread_mutex_unlock(&m_mut)) != 0)`

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,54 +23,71 @@ template <class T> class promise {
...
+            LOG(ERROR) << "pthread_mutex_lock failed, error code: " << err;
+            return false;
+        }
+        int retcode = 0;
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            if ((retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout)) != 0) {
```
tomereli: err

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,54 +23,71 @@ template <class T> class promise {
...
-        int retcode = 0;
-
         // lock the mutex and wait for the conditional variable to be set or timeout
-        pthread_mutex_lock(&m_mut);
+        int err;
+        if ((err = pthread_mutex_lock(&m_mut)) != 0) {
+            LOG(ERROR) << "pthread_mutex_lock failed, error code: " << err;
+            return false;
+        }
+        int retcode = 0;
```
tomereli: no need for 2 variables

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,54 +23,71 @@ template <class T> class promise {
...
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            if ((retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout)) != 0) {
+                LOG(ERROR) << "pthread_cond_timedwait failed, error code: " << err;
+            }
+            if (pthread_mutex_unlock(&m_mut) != 0) {
```
tomereli: forgot to add `err = `

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,54 +23,71 @@ template <class T> class promise {
...
+                LOG(ERROR) << "pthread_mutex_unlock failed, error code: " << err;
+                return false;
+            }
         }
-        pthread_mutex_unlock(&m_mut);
-
-        if (retcode != 0)
-            return false;
+        return retcode == 0;
```
tomereli:
```
if (err)
  return false;
return true;
```

framework/common/encryption.cpp
```diff
@@ -90,7 +90,6 @@ diffie_hellman::diffie_hellman() : m_dh(nullptr), m_pubkey(nullptr)
 
 diffie_hellman::~diffie_hellman()
 {
-    delete m_pubkey;
```
tomereli, line 93: @arnout please comment.

tomereli (changes requested):

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +23,70 @@ template <class T> class promise {
     // Set value for waiting threads
     void set_value(const T value)
     {
-        pthread_mutex_lock(&m_mut);
+        int err;
+        if ((err = pthread_mutex_lock(&m_mut)) != 0) {
+            LOG(ERROR) << "pthread_mutex_lock failed, error code: " << err;
+            return;
+        }
```
arnout: Possibly not for this PR, but this really should be an assert. The caller doesn't check the return value of set_value (because there isn't any), but does continue life as if it was set successfully.

The same comment applies to all changes in this commit.

framework/common/encryption.cpp
```diff
@@ -88,11 +88,7 @@ diffie_hellman::diffie_hellman() : m_dh(nullptr), m_pubkey(nullptr)
     BN_bn2bin(pub_key, m_pubkey);
 }
 
-diffie_hellman::~diffie_hellman()
-{
-    delete m_pubkey;
-    DH_free(m_dh);
-}
+diffie_hellman::~diffie_hellman() { DH_free(m_dh); }
```
arnout: This can't be right... The thing that is used inside m_dh is a BigNum, while m_pubkey is a copy of that that is allocated with our `new[]` and that is filled up with BN_bn2bin(). So the delete[] really is needed IMO...

tomereli (approved):

arnout (approved):
The second comment shows that klocwork actually has value, because both issues fixed there were real issues. Nice work!

PR #671, opened 2020-01-14
Feature/ci add certification tests job

rmelotte:
Add support for running the certification tests in CI.

It requires the CI runner to have the repository containing the certification agents and scripts cloned under `/easymesh_cert` (https://gitlab.com/prpl-foundation/easymesh_cert/easymesh_cert).

Tests to run are specified by the `TESTS_TO_RUN` variable and can either be a list of (space-separated) tests, or a file containing a list of (newline-separated) tests.

By default it runs all the controller tests (but it still has to be triggered manually).

It's currently possible to trigger the tests from the Web UI on the Gitlab mirror.

It's also possible to specify variables values from the web UI, but it has a few limitations:
- If you retry the same job, it will revert to the default variable values (meaning it will run all the controller tests at the moment, even if you specified only 1 test the first time). [Issue on Gitlab.com](https://gitlab.com/gitlab-org/gitlab/issues/31367).
- When you retry a manual job, you cannot change the variable values. [Issue on Gitlab.com](https://gitlab.com/gitlab-org/gitlab/issues/32712).
- Rerunning a manual job seems to ignore the pipeline dependencies, meaning that all the artifacts will be downloaded on the certification tests runner (I don't have a specific ticket for this one).

Bottom line: running the certification tests once per pipeline via the web UI is OK for simple cases. For more complex ones, it's probably easier to make a branch and change the variables values directly in `.gitlab-ci.yml`, as @arnout suggested.


arnout (approved):
We will also need to be able to specify TARGET_DEVICE_NAME at some point (with localhost or empty or something for the controller tests running on the laptop itself). But for now, this is sufficient.

arnout:
It would also be good to refer to the repo where the run_test_file.sh script itself is stored, because that's of course the most important one to review. Even if it is in a private repo, people who have access to that repo can use that information (and others can ask for access).

rmelotte:
At the moment the scripts use `prplmesh` to refer to prplMesh running on the laptop. We could probably rename it to `prplmesh-dummy` or something like that.
Adding other targets will require more work, since we need to make sure targets that are not part of the test do not interfere with it (either stop prplMesh on all of them as a first step, or rely on dev_reset).
It also needs to be a separate job for each target, because it requires different artifacts for each (we can use a job template similar to `.test-on-target`).

I added a link to the easymesh_cert repo.

dob71 (approved):
It would be nice to have a comment with the link to the repo containing `run_test_file.sh` right there in the yaml file.

tomereli (approved):

PR #673, opened 2020-01-14
Improvements to test_flows.sh

arnout:
I was going to add a test for combined infra metrics to test_flows.sh to help with debugging #669.
While doing that, there were a lot of small improvements I made to the script.
Since I'm now temporarily dropping the work on combined infra metrics, I'm pushin this as a separate PR so others can make use of it.

### DIFF
--- a/tools/docker/tests/test_flows.sh
+++ b/tools/docker/tests/test_flows.sh
@@ -81,6 +81,14 @@ check_log() {
     check grep -q "$@" "${rootdir}/logs/${device}/beerocks_${program}.log"
 }
 
+send_bwl_event() {
+    # Send an event to a dummy bwl
+    # $1: device to send to (gateway, repeater1, repeater2)
+    # $2: radio to send to (wlan0, wlan2)
+    # $3: event string
+    check docker exec $1 sh -c 'echo '"$3"' > /tmp/$USER/beerocks/'$2'/EVENT'
+}
+
 test_initial_ap_config() {
     status "test initial autoconfig"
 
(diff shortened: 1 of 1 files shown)


tools/docker/tests/test_flows.sh
```diff
@@ -71,22 +71,34 @@ send_CAPI_command() {
...
 }
 
+check_log() {
+    # Check for a certain message in a log file
+    # $1: device to check (gateway, repeater1, repeater2)
+    # $2: program to check (controller, agent, agent_wlan0, ...)
+    # rest: arguments to grep. -q is always added.
+    device="$1"; shift
+    program="$1"; shift
+    check grep -q "$@" "${rootdir}/logs/${device}/beerocks_${program}.log"
```
ghost, line 88: So long as no test need to look at the `transport.log` or `local_bus.log`, these don't have the `beerocks_` prefix for some reason

ghost (approved):
Looks good otherwise, so approving

tomereli (approved):

tools/docker/tests/test_flows.sh
arnout: Good point. However, the chance that we need to grep those two log files is rather small, and we're anyway converting the test flows to a python-based setup so this will no longer be relevant. So I'm going to leave it as it is now, and we can still fix it later if needed.

tools/docker/tests/test_flows.sh
```diff
@@ -81,6 +81,14 @@ check_log() {
     check grep -q "$@" "${rootdir}/logs/${device}/beerocks_${program}.log"
 }
 
+send_bwl_event() {
+    # Send an event to a dummy bwl
+    # $1: device to send to (gateway, repeater1, repeater2)
+    # $2: radio to send to (wlan0, wlan2)
+    # $3: event string
+    check docker exec $1 sh -c 'echo '"$3"' > /tmp/$USER/beerocks/'$2'/EVENT'
```
morantr, line 96: AFAIR, there is also an option to send Data event, so it is probably a good idea to add an option for that as well.
@arnout

arnout: > AFAIR, there is also an option to send Data event, so it is probably a good idea to add an option for that as well.

They use the same file, they're just distinguished by the string `EVENT foo` vs. `DATA foo`.

PR #674, opened 2020-01-14
Common: Add error checking to beerocks_promise lock functions

Galvien:
Klocwork noted that in every `pthread` function calls in
`beerocks_promise.h` the return values of the pthread functions are never checked, some platforms could cause errors while calling these functions, therefore:

1. check for non-zero return values for these issues
2. add log prints for such errors

@arnout noted that this should be a PR by itself, independent from Klocwork related PRs, so it could be discussed independently, on the subject of whether such checks shouldn't be done, as the success of `pthread` functions should be asserted.
Discuss ahead.

### DIFF
--- a/common/beerocks/bcl/include/bcl/beerocks_promise.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_promise.h
@@ -9,6 +9,8 @@
 #ifndef _BEEROCKS_PROMISE_H_
 #define _BEEROCKS_PROMISE_H_
 
+#include "../../common/include/mapf/common/err.h"
+#include <easylogging++.h>
 #include <pthread.h>
 #include <sys/time.h>
 
@@ -22,25 +24,30 @@ public:
     // Set value for waiting threads
     void set_value(const T value)
     {
-        pthread_mutex_lock(&m_mut);
+        int err = pthread_mutex_lock(&m_mut);
+        mapf_assert(err == 0);
         m_value  = value;
         m_signal = 1;
-        pthread_cond_broadcast(&m_cond);
-        pthread_mutex_unlock(&m_mut);
+        err      = pthread_cond_broadcast(&m_cond);
+        mapf_assert(err == 0);
+        err = pthread_mutex_unlock(&m_mut);
+        mapf_assert(err == 0);
     }
-
     // Wait for a value to be set
     void wait()
     {
         // lock the mutex and wait for the conditional variable to be set
-        pthread_mutex_lock(&m_mut);
+        int err = pthread_mutex_lock(&m_mut);
+        mapf_assert(err == 0);
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            pthread_cond_wait(&m_cond, &m_mut);
+            err = pthread_cond_wait(&m_cond, &m_mut);
+            mapf_assert(err == 0);
         }
-        pthread_mutex_unlock(&m_mut);
+        err = pthread_mutex_unlock(&m_mut);
+        mapf_assert(err == 0);
     }
 
     // Wait for a value to be set with timeout
@@ -48,27 +55,27 @@ public:
     {
         struct timeval now;
         struct timespec timeout;
-
         // set the absolut timeout
         gettimeofday(&now, NULL);
         timeout.tv_sec  = now.tv_sec + (timeout_ms / 1000);
         timeout.tv_nsec = (now.tv_usec + (timeout_ms % 1000) * 1000) * 1000;
 
-        int retcode = 0;
-
         // lock the mutex and wait for the conditional variable to be set or timeout
-        pthread_mutex_lock(&m_mut);
+        int err      = pthread_mutex_lock(&m_mut);
+        int err_wait = 0;
+        mapf_assert(err == 0);
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            if ((err_wait = pthread_cond_timedwait(&m_cond, &m_mut, &timeout)) != 0) {
+                LOG(ERROR) << "pthread_cond_timedwait failed, error code: " << err;
+            }
+            err = pthread_mutex_unlock(&m_mut);
+            mapf_assert(err == 0);
         }
-        pthread_mutex_unlock(&m_mut);
-
-        if (retcode != 0)
+        if (err_wait)
             return false;
-
         return true;
     }
 


tomereli (approved):
I think that regardless we should go ahead and merge this. Discussions can follow

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +23,70 @@ template <class T> class promise {
...
+        if ((err = pthread_mutex_lock(&m_mut)) != 0) {
+            LOG(ERROR) << "pthread_mutex_lock failed, error code: " << err;
+            return;
+        }
         m_value  = value;
         m_signal = 1;
-        pthread_cond_broadcast(&m_cond);
-        pthread_mutex_unlock(&m_mut);
+        if ((err = pthread_cond_broadcast(&m_cond)) != 0)
+            LOG(ERROR) << "pthread_cond_broadcast failed, error code: " << err;
```
morantr: Is there a point to do the next check if this one fails?

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +23,70 @@ template <class T> class promise {
...
+            LOG(ERROR) << "pthread_mutex_lock failed, error code: " << err;
+            return false;
+        }
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            if ((err = pthread_cond_timedwait(&m_cond, &m_mut, &timeout)) != 0) {
+                LOG(ERROR) << "pthread_cond_timedwait failed, error code: " << err;
```
morantr, line 72: Is there a point to do the next check if this one fails?

morantr (approved):

common/beerocks/bcl/include/bcl/beerocks_promise.h
arnout: Yes, the mutex must still be unlocked if it fails.

But actually, this is the reason why I think it should be asserts: if this fails, we have broken invariants and there is no way to recover. Most likely something surprising will go wrong at some (potentially distant) point in the future. IMO not asserting here is even a security issue, because it might be used by an attacker to trigger specific, undesired behaviour.

arnout:
[Clean all KW issues on UGW/RDKB](https://trello.com/c/ZeIOJ2LG/26-clean-all-kw-issues-on-ugw-rdkb)

common/beerocks/bcl/include/bcl/beerocks_promise.h
tomereli: So what's blocking us from adding asserts?

CoralMalachi: I'll add it now.

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -23,38 +24,30 @@ template <class T> class promise {
...
+        int err = pthread_mutex_lock(&m_mut);
+        assert(err == 0);
         m_value  = value;
         m_signal = 1;
-        if ((err = pthread_cond_broadcast(&m_cond)) != 0)
-            LOG(ERROR) << "pthread_cond_broadcast failed, error code: " << err;
-        if ((err = pthread_mutex_unlock(&m_mut)) != 0)
-            LOG(ERROR) << "pthread_mutex_unlock failed, error code: " << err;
+        err      = pthread_cond_broadcast(&m_cond);
+        assert(err == 0);
```
arnout: Please use mapf_assert from mapf/common/err.h.

It does something different than assert at the moment, but with a wrapper it's easier to change to standard assert (and use LOG for logging).

(nitpick)

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +24,56 @@ template <class T> class promise {
...
-        pthread_mutex_lock(&m_mut);
+        int err = pthread_mutex_lock(&m_mut);
+        assert(err == 0);
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            err = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            assert(err == 0);
```
arnout: This one is actually wrong: it should stay like it was originally:

```
retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
```

The whole point of `wait_for` is that it can timeout, and here you assert on timeout.

arnout (changes requested):

common/beerocks/bcl/include/bcl/beerocks_promise.h
arnout: Actually, not such a nitpick, as you can see in https://gitlab.com/prpl-foundation/prplMesh/-/jobs/411168392

With -DNDEBUG (i.e., when building for release), `assert` is defined as empty, so gcc thinks err is not used.

ghost (approved):

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +24,57 @@ template <class T> class promise {
...
-        pthread_mutex_lock(&m_mut);
+        int err = pthread_mutex_lock(&m_mut);
+        mapf_assert(err == 0);
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            if ((err = pthread_cond_timedwait(&m_cond, &m_mut, &timeout)) != 0) {
+                LOG(ERROR) << "pthread_cond_timedwait failed, error code: " << err;
```
ghost, line 72: (nitpick) it may be useful to identify the failure at a glance, so adding `strerror(err)` to the log is definitely useful, in my experience (tracking down the correct version of errno.h and cross-referencing error numbers can be time consuming)

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +24,57 @@ template <class T> class promise {
...
+            }
+            err = pthread_mutex_unlock(&m_mut);
+            mapf_assert(err == 0);
         }
-        pthread_mutex_unlock(&m_mut);
-
-        if (retcode != 0)
+        if (err)
```
arnout: If we get here, `err` is always going to be 0. We need a different variable to store the result of `pthread_cond_timedwait`

arnout (changes requested):
Still not OK I'm afraid.

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +24,58 @@ template <class T> class promise {
...
+        int err = pthread_mutex_lock(&m_mut);
+        int err_wait = 0;
+        mapf_assert(err == 0);
         // If signaled before the wait, no need to wait
         if (m_signal == 1) {
             m_signal = 0;
         } else {
-            retcode = pthread_cond_timedwait(&m_cond, &m_mut, &timeout);
+            if ((err_wait = pthread_cond_timedwait(&m_cond, &m_mut, &timeout)) != 0) {
+                LOG(ERROR) << "pthread_cond_timedwait failed, error code: " << err;
```
arnout, line 72: If err_wait == ETIMEDOUT we don't really need to log an error.

Actually, if err_wait != ETIMEDOUT we might be better off asserting...

common/beerocks/bcl/include/bcl/beerocks_promise.h
```diff
@@ -22,53 +24,58 @@ template <class T> class promise {
...
+            }
+            err = pthread_mutex_unlock(&m_mut);
+            mapf_assert(err == 0);
         }
-        pthread_mutex_unlock(&m_mut);
-
-        if (retcode != 0)
+        if (err_wait)
```
arnout, line 77: It's a bit weird that in the condition above you do `if (err_wait != 0)` and here it's `if (err_wait)`... We should converge on a convention about comparisons in conditions.

arnout (approved):
Really nitpicks now so I'm merging.

PR #676, opened 2020-01-14
add deploy script to tools

CoralMalachi:
Add a deploy script to make the operation of updating the current prplMesh version that running on the RAX40 (or any other board) much 
easier.

currently, each change in the prplMesh source code needs to be updated on the board itself so we can test it. Therefore, add the following script, which accepts the details of the desired board you want to update 
the prplMesh version on.

currently, the way we do it is with a "proxy" between the local machine and the desired board. An option to do the same process without a proxy will be added soon.

### DIFF
--- a/tools/deploy_ipk.sh
+++ b/tools/deploy_ipk.sh
@@ -35,8 +35,10 @@ deploy() {
 # we don't want opkg to stay locked with a previous failed invocation.
 # when using this, make sure no one is using opkg in the meantime!
 pgrep opkg | xargs kill -SIGINT
-# remove any previously installed prplmesh:
-opkg remove prplmesh prplmesh-dwpal prplmesh-nl80211
+# remove any previously installed prplmesh. Use force-depends in case
+# there are packages depending on prplmesh which will cause opkg remove
+# to fail without this:
+opkg remove --force-depends prplmesh prplmesh-dwpal prplmesh-nl80211
 # currently opkg remove does not remove everything from /opt/prplmesh:
 rm -rf /opt/prplmesh
 opkg install -V2 "$DEST_FOLDER/$IPK_FILENAME"
@@ -77,6 +79,6 @@ main() {
     deploy "$@"
 }
 
-SSH_OPTIONS="\"-oBatchMode yes\""
+SSH_OPTIONS="\"-oBatchMode yes\" \"-oStrictHostKeyChecking no\""
 
 main "$@"
--- a/tools/deploy_ipk.sh
+++ b/tools/deploy_ipk.sh
@@ -1,16 +1,22 @@
 #!/bin/sh -e
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2019 Raphaël Mélotte
+# Copyright (c) 2020 Tomer Eliyahu (Intel)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
 
 usage() {
-    echo "usage: $(basename "$0") [-h]"
+    echo "usage: $(basename "$0") <user>@<target> <ipk> [-h]"
     echo "  options:"
     echo "      -h|--help - show this help menu"
     echo "      --certification-mode - enable certification-mode on the target"
+    echo "      --proxy - use ssh proxy <user>@<proxy>"
 }
 
 deploy() {
-    SSH_OPTIONS="-oBatchMode=yes"
-
-    if ! ssh "$SSH_OPTIONS" "$1" /bin/true ; then
+    if ! eval ssh "$SSH_OPTIONS" "$1" /bin/true ; then
         echo "Error: $1 unreachable via ssh"
         exit 1
     fi
@@ -21,11 +27,11 @@ deploy() {
     DEST_FOLDER=/tmp/prplmesh_ipks
 
     echo "Removing previous ipks"
-    ssh "$TARGET" "rm -rf \"$DEST_FOLDER\" ; mkdir -p \"$DEST_FOLDER\""
+    eval ssh "$SSH_OPTIONS" "$TARGET" \""rm -rf \"$DEST_FOLDER\" ; mkdir -p \"$DEST_FOLDER\"\""
     echo "Copying $IPK to $TARGET:$DEST_FOLDER/$IPK_FILENAME"
-    scp "$IPK" "$TARGET:$DEST_FOLDER/$IPK_FILENAME"
+    eval scp "$SSH_OPTIONS" "$IPK" "$TARGET:$DEST_FOLDER/$IPK_FILENAME"
 
-    ssh "$TARGET" <<EOF
+    eval ssh "$SSH_OPTIONS" "$TARGET" <<EOF
 # we don't want opkg to stay locked with a previous failed invocation.
 # when using this, make sure no one is using opkg in the meantime!
 pgrep opkg | xargs kill -SIGINT
@@ -38,16 +44,18 @@ EOF
 
     if [ "$CERTIFICATION_MODE" = true ] ; then
         echo "Certification mode will be enabled on the target"
-        ssh "$TARGET" "uci set prplmesh.config.certification_mode=1 && uci commit"
+        eval ssh "$SSH_OPTIONS" "$TARGET" "uci set prplmesh.config.certification_mode=1 && uci commit"
         echo "Certification mode enabled on the target."
     fi
     echo "Done"
 }
 
 main() {
-        OPTS=$(getopt -o 'h' --long help,certification-mode -n 'parse-options' -- "$@")
-
-    if [ "$?" != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
+    if ! OPTS=$(getopt -o 'h' --long help,certification-mode,proxy: -n 'parse-options' -- "$@"); then
+        err "Failed parsing options." >&2
+        usage
+        exit 1
+    fi
 
     eval set -- "$OPTS"
 
@@ -57,6 +65,10 @@ main() {
                 CERTIFICATION_MODE=true
                 shift
                 ;;
+            --proxy)
+                SSH_OPTIONS="$SSH_OPTIONS \"-oProxyJump \"$2\"\""
+                shift 2
+                ;;
             -- ) shift; break ;;
             * ) err "unsupported argument $1"; usage; exit 1 ;;
         esac
@@ -64,5 +76,7 @@ main() {
 
     deploy "$@"
 }
+
+SSH_OPTIONS="\"-oBatchMode yes\""
 
 main "$@"
--- a/tools/deploy_ipk.sh
+++ b/tools/deploy_ipk.sh
@@ -54,7 +54,7 @@ EOF
 
 main() {
     if ! OPTS=$(getopt -o 'h' --long help,certification-mode,proxy: -n 'parse-options' -- "$@"); then
-        err "Failed parsing options." >&2
+        echo "Failed parsing options." >&2
         usage
         exit 1
     fi
--- a/tools/scripts/deploy.sh
+++ b/tools/scripts/deploy.sh
@@ -1,0 +1,106 @@
+#!/bin/sh
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2020 Tomer Eliyahu (Intel)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+scriptdir="$(cd "${0%/*}" && pwd)"
+rootdir="${scriptdir%/*/*}"
+
+dbg() {
+    [ "$VERBOSE" = "true" ] && echo "$@"
+}
+
+err() {
+    printf "\033[1;31m%s\n\033[0m" "$@"
+}
+
+success() {
+    printf "\033[1;32m%s\n\033[0m" "$@"
+}
+
+run() {
+    dbg "$*"
+    "$@" || exit $?
+}
+
+
+usage() {
+    echo "usage: $(basename "$0") [-hvdb] [--target <user@target>] [--proxy <user@proxy>]"
+}
+
+main() {
+
+    if ! OPTS=$(getopt -o 'hvdb:' --long help,verbose,deploy,build-dir:,target:,proxy: -n 'parse-options' -- "$@"); then
+        err "Failed parsing options." >&2
+        usage
+        exit 1 
+    fi
+
+    eval set -- "$OPTS"
+
+    while true; do
+        case "$1" in
+            -v | --verbose)       VERBOSE=true; shift ;;
+            -h | --help)          usage; exit 0 ;;
+            -d | --deploy)        DEPLOY=true; shift ;;
+            -b | --build-dir)     BUILD_DIR="$2"; shift; shift ;;
+            --proxy)              SSH_OPTIONS="$SSH_OPTIONS \"-oProxyJump $2\""; shift 2 ;;
+            --target)             TARGET="$2"; shift 2 ;;
+            --keep-conf)          KEEP_CONF=true; shift ;;
+            -- ) shift; break ;;
+            * ) err "unsupported argument $1"; usage; exit 1 ;;
+        esac
+    done
+
+    echo "Start"
+    tmpdir=$(mktemp -d)
+    mkdir -p "${tmpdir}"/opt/prplmesh
+    mkdir -p "${tmpdir}"/usr
+
+    cp -r "${BUILD_DIR}"/install/bin "${tmpdir}"/opt/prplmesh
+    cp -r "${BUILD_DIR}"/install/config "${tmpdir}"/opt/prplmesh
+    cp -r "${BUILD_DIR}"/install/lib "${tmpdir}"/usr/
+    cp -r "${BUILD_DIR}"/install/scripts "${tmpdir}"/opt/prplmesh
+    cp -r "${BUILD_DIR}"/install/share "${tmpdir}"/opt/prplmesh
+
+    dbg "create prplmesh.tar.gz..."
+    tar czvf "${BUILD_DIR}"/prplmesh.tar.gz -C "${tmpdir}"/ .
+    rm -rf "${tmpdir}"
+
+    [ "$DEPLOY" = "true" ] && {
+
+        echo "local md5sum"
+        run md5sum "${BUILD_DIR}"/prplmesh.tar.gz
+
+        echo "scp ${BUILD_DIR}/prplmesh.tar.gz to ${TARGET}"
+        eval scp "${SSH_OPTIONS} ${BUILD_DIR}/prplmesh.tar.gz ${TARGET}:/tmp/"
+
+        if [ "$KEEP_CONF" = "true" ]; then
+            echo "backaup /opt/prplmesh/beerocks_agent.conf on target"
+            eval ssh "${SSH_OPTIONS} ${TARGET} cp /opt/prplmesh/config/beerocks_agent.conf /tmp/"
+        fi
+
+        echo "untar /tmp/prplmesh.tar.gz on target"
+        eval ssh "${SSH_OPTIONS} ${TARGET} tar -C / -xzvf /tmp/prplmesh.tar.gz"
+
+        if [ "$KEEP_CONF" = "true" ]; then
+            echo "restore /opt/prplmesh/beerocks_agent.conf on target"
+            eval ssh "${SSH_OPTIONS} ${TARGET} mv /tmp/beerocks_agent.conf /opt/prplmesh/config/"
+        fi
+    
+    }
+
+    success "Done! Run /etc/init.d/prplmesh.sh restart on target to apply new settings."
+}
+
+VERBOSE=false
+BUILD_DIR=${rootdir}/build
+DEPLOY=false
+KEEP_CONF=false
+TARGET=
+SSH_OPTIONS="\"-oBatchMode yes\" \"-oStrictHostKeyChecking no\""
+
+main "$@"


tools/scripts/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+
+            # echo "Deploy prplmesh completed, on target run /etc/init.d/prplmesh.sh restart for changes to take effect."
+        fi
+    }
+
+    success "Done! Run /etc/init.d/prplmesh.sh restart on target to apply new settings."
+}
+
+VERBOSE=false
+BUILD_DIR="prplMesh/build"
```
tomereli: Now we can change the path to a relative one since it is part of the repository

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+}
+
+run() {
+    dbg "$*"
+    "$@" || exit $?
+}
+
+
+usage() {
+    echo "usage: $(basename $0) [-hvsmdCD]"
```
tomereli: Update usage

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+            echo "backaup /opt/prplmesh/beerocks_agent.conf on target"
+            run sshpass -p ${PROXY_PASS} ssh ${SSH_OPTIONS} ${PROXY_USER}@${PROXY_IP} "sshpass -p ${TARGET_PASS} ssh ${SSH_OPTIONS} ${TARGET_USER}@${TARGET_IP} cp /opt/prplmesh/config/beerocks_agent.conf /tmp/"
+
+            echo "untar /tmp/prplmesh.tar.gz on target"
+            run sshpass -p ${PROXY_PASS} ssh ${SSH_OPTIONS} ${PROXY_USER}@${PROXY_IP} "sshpass -p ${TARGET_PASS} ssh ${SSH_OPTIONS} ${TARGET_USER}@${TARGET_IP} tar -C / -xzvf /tmp/prplmesh.tar.gz"
+
+            echo "restore /opt/prplmesh/beerocks_agent.conf on target"
+            run sshpass -p ${PROXY_PASS} ssh ${SSH_OPTIONS} ${PROXY_USER}@${PROXY_IP} "sshpass -p ${TARGET_PASS} ssh ${SSH_OPTIONS} ${TARGET_USER}@${TARGET_IP} cp /tmp/beerocks_agent.conf /opt/prplmesh/config/"
+    
+        # else
```
tomereli: Need to also add error checking

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,117 @@
...
+            # sshpass -p ${TARGET_PASS} scp -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no ${BUILD_DIR}/prplmesh.tar.gz ${TARGET_USER}@${TARGET_IP}:/tmp/
+
+            # echo "untar prplmesh in target"
+            # ssh -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no root@192.168.1.1 'tar -xzvf /tmp/prplmesh.tar.gz -C /'
+
+            # echo "Deploy prplmesh completed, on target run /etc/init.d/prplmesh.sh restart for changes to take effect."
+        fi
+    }
+
+    success "Done! Run /etc/init.d/prplmesh.sh restart on target to apply new settings."
```
tomereli, line 96: Change to `/opt/prplmesh/scripts/prplmesh_utils.sh restart` so it will work for rdkb also

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/sh
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2019 Tomer Eliyahu (Intel)
```
tomereli: Add yourself to the copyright 😁

tomereli (commented):
We should probably convert this script to python and add it to maptools deploy sometime in the future.

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,128 @@
...
+            * ) err "unsupported argument $1"; usage; exit 1 ;;
+        esac
+    done
+
+    echo "Start"
+    tmpdir=$(mktemp -d)
+    mkdir -p ${tmpdir}/opt/prplmesh
+    mkdir -p ${tmpdir}/usr
+
+    cp -r ${BUILD_DIR}/out/* ${tmpdir}/opt/prplmesh
```
morantr: AFAIR the correct files are placed on the install dir, and on the "out" dir the files are irrelevant. Did we fix it?

arnout: The only reason why the out dir isn't removed yet is because supposedly there were scripts like this one that still used it. I'd be more than happy to remove it!

tools/scripts/deploy.sh
```diff
@@ -6,19 +6,19 @@
 # See LICENSE file for more details.
 ###############################################################
 
-scriptdir="$(cd "${0%/*}"; pwd)"
+scriptdir="$(cd "${0%/*}" || exit; pwd)"
```
rmelotte: Just a quick note: the `|| exit` change is enough to make shellcheck happy, but it doesn't handle the error in any way: if `cd` fails `scriptdir` will still be empty, which can still lead to the issues pointed out by [SC2164](https://github.com/koalaman/shellcheck/wiki/SC2164).
However, I don't think we need to handle these kind of issues either, so if the point is only to silence shellcheck, it's easier to just change `;` to `&&`, so:
```
$(cd "${0%/*}" && pwd)
```
The result will be the same: exit if cd fails, otherwise run pwd.

Otherwise it's good that we already address the shellcheck issues we can address, so that we can run it in CI as a required check at some point, so thanks for taking the time to make those changes :-)

tools/scripts/deploy.sh
```diff
@@ -28,13 +28,14 @@ run() {
...
-    echo "usage: $(basename $0) [-hvsmdCD]"
+    echo "usage: $(basename "$0") [-hvsmdCD]"
 }
 
 main() {
-    OPTS=`getopt -o 'hvdb:i:' --long verbose,help,deploy,build-dir:,target-ip:,target-user:,target-pass:,proxy-ip:,proxy-user:,proxy-pass: -n 'parse-options' -- "$@"`
 
-    if [ $? != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
+    if ! OPTS=$(getopt -o 'hvdb:i:' --long verbose,help,deploy,build-dir:,target-ip:,target-user:,target-pass:,proxy-ip:,proxy-user:,proxy-pass: -n 'parse-options' -- "$@") ; 
+    then err "Failed parsing options." >&2 ; usage; exit 1 ;
```
rmelotte: The semicolon at the end of the lines are not needed, and there are also extra spaces at the end of the lines.
Also why squashing everything on one line?

(nitpick)

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,132 @@
...
+    [ "$DEPLOY" = "true" ] && {
+
+        echo "local md5sum"
+        run md5sum "${BUILD_DIR}"/prplmesh.tar.gz
+
+        if [ -n "$PROXY_IP" ] && [ -n "$PROXY_USER" ] && [ -n "$PROXY_PASS" ]; then
+
+            echo "scp ${BUILD_DIR}/prplmesh.tar.gz to ${PROXY_USER}@${PROXY_IP}"
+            run sshpass -p "${PROXY_PASS}" scp "${SSH_OPTIONS}" "${BUILD_DIR}"/prplmesh.tar.gz "${PROXY_USER}"@"${PROXY_IP}":/home/"${PROXY_USER}"/
+
```
rmelotte, line 80: You don't need to copy anything on the proxy. You can use ProxyJump for SSH as well as for scp commands to copy it directly to the target (at least with OpenSSH 7.3+; I have no idea about sshpass but it looks like a wrapper over the ssh commands).
With SSH it means using `-J`, with scp I think you have to use `-oProxyjump=<target>`.

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,132 @@
...
+            echo "backaup /opt/prplmesh/beerocks_agent.conf on target"
+            run sshpass -p "${PROXY_PASS}" ssh "${SSH_OPTIONS}" "${PROXY_USER}"@"${PROXY_IP}" "sshpass -p ${TARGET_PASS} ssh ${SSH_OPTIONS} ${TARGET_USER}@${TARGET_IP} cp /opt/prplmesh/config/beerocks_agent.conf /tmp/"
+
+            echo "untar /tmp/prplmesh.tar.gz on target"
+            run sshpass -p "${PROXY_PASS}" ssh "${SSH_OPTIONS}" "${PROXY_USER}"@"${PROXY_IP}" "sshpass -p ${TARGET_PASS} ssh ${SSH_OPTIONS} ${TARGET_USER}@${TARGET_IP} tar -C / -xzvf /tmp/prplmesh.tar.gz"
+
+            echo "restore /opt/prplmesh/beerocks_agent.conf on target"
+            run sshpass -p "${PROXY_PASS}" ssh "${SSH_OPTIONS}" "${PROXY_USER}"@"${PROXY_IP}" "sshpass -p ${TARGET_PASS} ssh ${SSH_OPTIONS} ${TARGET_USER}@${TARGET_IP} cp /tmp/beerocks_agent.conf /opt/prplmesh/config/"
+    
+        elif [ -n "$TARGET_IP" ] && [ -n "$TARGET_USER" ] && [ -n "$TARGET_PASS" ]; then
```
rmelotte: What if you don't use a password for the target (and use ssh keys instead)? It looks like the script won't do anything if you don't provide a password.

tools/scripts/deploy.sh
tomereli: That's a great feedback, thank you! Changed to use ProxyJump.

tools/scripts/deploy.sh
tomereli: Completely removed the password and use of sshpass.
The user is expected to set up ssh keys via `ssh-keygen; ssh-copy-id <proxy>`, otherwise he will be prompt for entering the proxy password.

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,103 @@
...
+    fi
+
+    eval set -- "$OPTS"
+
+    while true; do
+        case "$1" in
+            -v | --verbose)       VERBOSE=true; shift ;;
+            -h | --help)          usage; exit 0 ;;
+            -d | --deploy)        DEPLOY=true; shift ;;
+            -b | --build-dir)     BUILD_DIR="$2"; shift; shift ;;
```
arnout, line 49: This is only used for two things:

- The install directory, i.e. `${BUILD_DIR}/install`.
- The location of the temporary tarball.

I think it would be more appropriate to instead pass the install directory, and use an actual temporary file (i.e., a `$(mktemp --tmpdir prplmesh-XXXX.tar.gz)`) for the tarball location. But that would of course mean we have to delete it as well. Hm, I'm not sure...

Anyway, can be changed later.

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,103 @@
...
+            --proxy)              SSH_OPTIONS="$SSH_OPTIONS \"-oProxyJump $2\""; shift 2 ;;
```
arnout, line 50: I believe the commit message says this will be added later, but in fact it has already been added here, right?

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,103 @@
...
+    echo "Start"
+    tmpdir=$(mktemp -d)
+    mkdir -p "${tmpdir}"/opt/prplmesh
+    mkdir -p "${tmpdir}"/usr
+
+    cp -r "${BUILD_DIR}"/install/bin "${tmpdir}"/opt/prplmesh
+    cp -r "${BUILD_DIR}"/install/config "${tmpdir}"/opt/prplmesh
+    cp -r "${BUILD_DIR}"/install/lib "${tmpdir}"/usr/
+    cp -r "${BUILD_DIR}"/install/scripts "${tmpdir}"/opt/prplmesh
+    cp -r "${BUILD_DIR}"/install/share "${tmpdir}"/opt/prplmesh
```
arnout, line 67: OK, so if I understand correctly, this script only works if you don't build prplMesh in the normal way (i.e. `-DCMAKE_INSTALL_PREFIX=/opt/prplmesh` and `DESTDIR=build/install` at installation time), but you build it with incorrect `-DCMAKE_INSTALL_PREFIX=build/install`, right? Do we really want to add to that mess?

Or you could say this is a helper to fix things: first add this script, teach people to use it, and then we can update maptools or the docker or whatever is used to do the right thing?

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,103 @@
...
+    tar czvf "${BUILD_DIR}"/prplmesh.tar.gz -C "${tmpdir}"/ .
+    rm -rf "${tmpdir}"
+
+    [ "$DEPLOY" = "true" ] && {
+
+        echo "local md5sum"
+        run md5sum "${BUILD_DIR}"/prplmesh.tar.gz
+
+        echo "scp ${BUILD_DIR}/prplmesh.tar.gz to ${TARGET}"
+        run sh -c "scp ${SSH_OPTIONS} ${BUILD_DIR}/prplmesh.tar.gz ${TARGET}:/tmp/"
```
arnout: Why is the additional sh -c needed? If it is for second-expansion of SSH_OPTIONS, I think `eval` would be more appropriate. (nitpick)

tools/scripts/deploy.sh
```diff
@@ -0,0 +1,103 @@
...
+
+        echo "remote md5sum"
+        run sh -c "ssh ${SSH_OPTIONS} ${TARGET} md5sum /tmp/prplmesh.tar.gz"
```
arnout: Why is this needed? scp *will* return an error if the copy was not successful, so the md5sum will never be different. And you anyway don't check it, just report it, so what's the point?

tools/scripts/deploy.sh
arnout: This is marked as resolved but the change hasn't been done...

arnout (approved):

tools/scripts/deploy.sh
tomereli: The reason I didn't put the tar in a random location is because of the `--deploy` flag - you can use the script only to create `prplmesh.tar.gz`, so it helps if it is in a known location :)

tools/scripts/deploy.sh
tomereli: Right. Sooner than I thought :)

tools/scripts/deploy.sh
tomereli: As long as we have `maptools.py build` command, people will be using it, and since I wanted this script to be used right away, I chose it this way.
So your comment is correct - this is a helper to fix things, and we can update maptools later to do the right thing.

tools/scripts/deploy.sh
tomereli: I'll try and see if `eval` also works. Never understood when to use which one...

tools/scripts/deploy.sh
tomereli: good point, we can remove this then.

morantr (approved):

tools/deploy_ipk.sh
```diff
@@ -1,16 +1,22 @@
...
     echo "      -h|--help - show this help menu"
     echo "      --certification-mode - enable certification-mode on the target"
+    echo "      --proxy - use ssh proxy <user>@<proxy>"
 }
 
 deploy() {
-    SSH_OPTIONS="-oBatchMode=yes"
-
-    if ! ssh "$SSH_OPTIONS" "$1" /bin/true ; then
+    if ! eval ssh "$SSH_OPTIONS" "$1" /bin/true ; then
```
rmelotte, line 19: I often avoid using eval because I find it hard to get right. 
I know BatchMode was already in a variable so it's tempting to add more options to it, but it was a bad idea from the beginning.
How about using a function that always uses BatchMode, and that uses the proxy option depending if the option was used or not?

tools/deploy_ipk.sh
```diff
@@ -21,33 +27,35 @@ deploy() {
     DEST_FOLDER=/tmp/prplmesh_ipks
 
     echo "Removing previous ipks"
-    ssh "$TARGET" "rm -rf \"$DEST_FOLDER\" ; mkdir -p \"$DEST_FOLDER\""
+    eval ssh "$SSH_OPTIONS" "$TARGET" \""rm -rf $DEST_FOLDER ; mkdir -p $DEST_FOLDER\""
```
rmelotte: You removed the quoting around the folder names, so if they contain spaces, stars, or other special characters things will break.
Also since you added eval, you probably need another layer of quoting for SSH_OPTIONS and TARGET as well to get the same behavior as before (not using eval is easier!).

Since we probably won't use spaces in folder names, I let you decide if you want to fix it or not.

tools/deploy_ipk.sh
```diff
@@ -21,33 +27,35 @@ deploy() {
...
+    eval ssh "$SSH_OPTIONS" "$TARGET" <<EOF
 # we don't want opkg to stay locked with a previous failed invocation.
 # when using this, make sure no one is using opkg in the meantime!
 pgrep opkg | xargs kill -SIGINT
 # remove any previously installed prplmesh:
 opkg remove prplmesh prplmesh-dwpal prplmesh-nl80211
 # currently opkg remove does not remove everything from /opt/prplmesh:
 rm -rf /opt/prplmesh
-opkg install -V2 "$DEST_FOLDER/$IPK_FILENAME"
+opkg install --force-reinstall -V2 "$DEST_FOLDER/$IPK_FILENAME"
```
rmelotte: Are you really sure this is needed?
There is the 'opkg remove' right before, and the script can't get here if it failed (unless opkg remove doesn't report failures in its exit code?)

tools/deploy_ipk.sh
```diff
@@ -21,33 +27,35 @@ deploy() {
...
     fi
     echo "Done"
 }
 
 main() {
-        OPTS=$(getopt -o 'h' --long help,certification-mode -n 'parse-options' -- "$@")
-
-    if [ "$?" != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
+    if ! OPTS=$(getopt -o 'h' --long help,certification-mode,proxy: -n 'parse-options' -- "$@"); then
+        err "Failed parsing options." >&2
```
rmelotte: Only see it now, but if `err` always writes errors, couldn't it write to stderr itself? (it's a nitpick and not specific  to this PR of course).

tools/deploy_ipk.sh
```diff
@@ -57,6 +65,10 @@ main() {
                 CERTIFICATION_MODE=true
                 shift
                 ;;
+            --proxy)
+                SSH_OPTIONS="$SSH_OPTIONS \"-oProxyJump $2\""
```
rmelotte: `$2` should also be quoted.

rmelotte (approved):
I didn't repeat the comment about `eval` everywhere, but it's true for the whole file (both of them). If you can drop it easily, I would do it.

Other than that, I don't see any big issue, so approving.

tools/deploy_ipk.sh
tomereli: I used `eval` based on a comment by @arnout about not using `sh -c`. I'm not sure what you mean with this proposal - can you provide a short example?

tools/deploy_ipk.sh
tomereli: I really prefer not to use `eval` but i'm not sure how, see my comment in https://github.com/prplfoundation/prplMesh/pull/676#discussion_r378348035

tools/deploy_ipk.sh
tomereli: I think I get it now - when I built prplmesh, it was in `bin/packages/mips_24kc_nomips16/iwlwav/prplmesh-dwpal_1.4.0-d78175d4-1.4.0_mips_24kc_nomips16.ipk` so `basename` would produce `prplmesh-dwpal_1.4.0-d78175d4-1.4.0_mips_24kc_nomips16.ipk` and not `prplmesh-dwpal` which is used to remove prplmesh in line 39 - I assume this is the reason that `opkg remove` didn't work for me. I wonder though how it works (if it works) in CI - perhaps it is different since it is built with SDK and not as part of the full firmware?

tools/deploy_ipk.sh
tomereli: It certainly should, I'll try to remember to do that in a separate commit.

tools/deploy_ipk.sh
tomereli: All these quotes are giving me a headache :)

tools/deploy_ipk.sh
tomereli: So I was wrong. It is indeed `prplmesh-dwpal`, but removal fails none the less:
```
root@OpenWrt:~# opkg remove prplmesh-dwpal
No packages removed.
Collected errors:
 * print_dependents_warning: Package prplmesh-dwpal is depended upon by packages:
 * print_dependents_warning:    ltq-wlan-wave_6x-uci
 * print_dependents_warning: These might cease to work if package prplmesh-dwpal is removed.

 * print_dependents_warning: Force removal of this package with --force-depends.
 * print_dependents_warning: Force removal of this package and its dependents
 * print_dependents_warning: with --force-removal-of-dependent-packages.
root@OpenWrt:~#
```
So I guess that this we should definitely use either `--force-depends` for the `opkg remove` or `--force-reinstall` for the `opkg install`.

tools/deploy_ipk.sh
tomereli: Brought the quoting back

tools/deploy_ipk.sh
tomereli: OK so I fully understand this now - in CI, there is no dependencies on prplmesh since it is built using SDK, while on RAX40 latest firmware, prplmesh is already present and since it is selected by the wlan_wave_6x-uci package, opkg marks this as a dependency.

I added a comment and the `--force-depends` flag to the `opkg remove` to resolve this.

tools/deploy_ipk.sh
tomereli: actually err is undefined in this script, so I replaced it with echo.

tools/deploy_ipk.sh
tomereli: Still brought them back.

tools/deploy_ipk.sh
tomereli: I decided this has been waiting for far too long. We also need it for UGW since we have no other way to deploy to target there (need the deploy_ipk.sh script), so I'll merge this and keep this comment open so we can address this in a new PR.

rmelotte: I don't know any clean way to dynamically build a list of options in POSIX sh.
It's either eval, or arrays, but arrays are not POSIX except for one exception: `$@`.
This means you could use it to pass the correct options to a new function for example.

Either you have to play with the arguments passed to the script to shape the array in such a way that it can directly be passed to ssh/scp (e.g. remove the arguments you know and give everything else to ssh), or you have to call a function with all the options you're interested in, some being empty, and filter out the empty ones in the function.

Maybe we could also require bash for this script, since it's meant to be used on development platforms, not on the targets directly.

tools/deploy_ipk.sh
rmelotte: Oh, right. I forgot about that. Either the fact that `ltq-wlan-wave_6x-uci` depends on prplMesh is a bug we should fix, or we should remove it as well when we remove prplMesh (probably `--force-removal-of-dependent-packages` does it).

rmelotte:
`eval` did break `deploy_ipk.sh` after all: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/435514917

I'll just push a quick fix for now, but I still think we should get rid of `eval` if possible.

Issue #575, opened 2019-12-16
[TASK] Move platform manager to a separate process


tomereli:
## Description

Access to the platform DB and operations should only be done by a single entity (process) since it reduces the attack surface by limiting the number of processes that need access to privileged platform operations. For example, a bug in the agent will not give access to bridge control assuming correct access rights are set for each process.

In addition, all controller and agent's configuration should be moved to the platform DB (#574, #568) in order to remove target dependencies from prplMesh build system.

However, the platform manager is currently a thread in the main agent, and moving all the configuration parameters to the platform DB can't be done since the controller (and local agents) need to know the configuration on init.

In order to meet the above - The platform manager thread needs to be converted to a process, and renamed to "platform service". It should be the first process that is started so that every component can request the platform DB, trigger platform operations, and register to platform notifications.

Eventually, the platform service should (if built without `BPL_UDS` - similar to `BTL_UDS`) communicate via the local bus, similarly to the transport service, but with a `platform.` prefix.

It should expose API to get the current platform DB, and publish notifications to all subcscribers whenever there are DB updates. It should also verify the identity of any process with which it interacts (**TBD decide how exactly**).

**Subtasks**
- [ ] Map out the controller and agent configuration options that should move to the platform DB
- [ ] Replicate the controller and agent configuration in the platform DB, for example:
`uci show prplmesh.config.controller` shows the controller config
`uci show prplmesh.config.agent` shows the agent config
Debug options can stay in configuration files.
- [ ] Add platform service process which starts at init (copy of the platform manager thread) but still does not interact with prplmesh. The platform service should have a CMAKE option if to build in UDS or Local bus modes. At this point only UDS is used.
- [ ] Add UDS socket listener and register to the platform service at controller and agent init.
- [ ] Use the platform service instead of the platform manager thread
- [ ] Delete the platform manager thread

In a new PR, add local bus mode, in which the platform manager connects to the local bus and defines its own set of messages which are passed on the bus. The platform service subscribes only to `platform.tx.*` messages, which are sent from any component in prplMesh, and replies with `platform.rx.*`. Define the messages, and the steps to accomplish that.

## Exit criteria

- All operating related configuration (non debug) is moved to platform DB
- The platform service builds in UDS mode and starts first
- The agent and controller request the DB status from the platform service on boot
- On DB updates, the platform service broadcasts a notification (in UDS mode - to all the open sockets) for any DB update or platform status change.

tomereli:
[...]

PR #677, opened 2020-01-15
Feature/move sta ifaces from beerocks agent conf to platform db

adam1985d:
The following changes are contained in this PR:
- add generic BPL API to read prplmesh radio param from UCI
- add BPL APIs to read sta_iface from BPL (the UGW utilizes the API above)
- change sta_iface reads from beerocks_agent.conf to read using BPL
- platform_mgr is now responsible to read the sta_ifaces and update the config struct
- configurations that depend on sta_iface have moved to a later stage (after BPL is initialized and sta_ifaces are read from BPL)
- sta_iface was removed from beerocks_agent.conf and the sta_iface was removed from the config_file::read_slave_config_file().

### DIFF
--- a/agent/src/beerocks/slave/beerocks_slave_main.cpp
+++ b/agent/src/beerocks/slave/beerocks_slave_main.cpp
@@ -297,14 +297,6 @@ static int run_beerocks_slave(beerocks::config_file::sConfigSlave &beerocks_slav
     std::string pid_file_path =
         beerocks_slave_conf.temp_path + "pid/" + base_slave_name; // for file touching
 
-    // Threads
-    std::set<std::string> slave_sta_ifaces;
-    for (int slave_num = 0; slave_num < beerocks::IRE_MAX_SLAVES; slave_num++) {
-        if (!beerocks_slave_conf.sta_iface[slave_num].empty()) {
-            slave_sta_ifaces.insert(beerocks_slave_conf.sta_iface[slave_num]);
-        }
-    }
-
     std::set<std::string> slave_ap_ifaces;
     for (int slave_num = 0; slave_num < beerocks::IRE_MAX_SLAVES; slave_num++) {
         if (!beerocks_slave_conf.hostap_iface[slave_num].empty()) {
@@ -318,6 +310,14 @@ static int run_beerocks_slave(beerocks::config_file::sConfigSlave &beerocks_slav
     if (platform_mgr.init()) {
         // read the number of failures allowed before stopping agent from platform configuration
         int stop_on_failure_attempts = beerocks::bpl::cfg_get_stop_on_failure_attempts();
+
+        // The platform manager updates the beerocks_slave_conf.sta_iface in the init stage
+        std::set<std::string> slave_sta_ifaces;
+        for (int slave_num = 0; slave_num < beerocks::IRE_MAX_SLAVES; slave_num++) {
+            if (!beerocks_slave_conf.sta_iface[slave_num].empty()) {
+                slave_sta_ifaces.insert(beerocks_slave_conf.sta_iface[slave_num]);
+            }
+        }
 
         beerocks::backhaul_manager backhaul_mgr(beerocks_slave_conf, slave_ap_ifaces,
                                                 slave_sta_ifaces, stop_on_failure_attempts);
--- a/agent/src/beerocks/slave/beerocks_slave_main.cpp
+++ b/agent/src/beerocks/slave/beerocks_slave_main.cpp
@@ -125,6 +125,25 @@ static bool parse_arguments(int argc, char *argv[])
     return true;
 }
 
+static std::string get_sta_iface_from_hostap_iface(const std::string &hostap_iface)
+{
+    // read the sta_iface from bpl and verify it is available
+    char sta_iface_str[BPL_IFNAME_LEN];
+    std::string sta_iface;
+
+    if (beerocks::bpl::cfg_get_sta_iface(hostap_iface.c_str(), sta_iface_str) < 0) {
+        LOG(ERROR) << "failed to read sta_iface for slave ";
+        return std::string();
+    } else {
+        sta_iface = std::string(sta_iface_str);
+        if (!beerocks::net::network_utils::linux_iface_exists(sta_iface)) {
+            LOG(DEBUG) << "sta iface " << sta_iface << " does not exist, clearing it from config";
+            sta_iface.clear();
+        }
+    }
+    return sta_iface;
+}
+
 static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_slave_conf,
                                   son::slave_thread::sSlaveConfig &son_slave_conf, int slave_num)
 {
@@ -149,8 +168,9 @@ static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_
     son_slave_conf.hostap_iface = beerocks_slave_conf.hostap_iface[slave_num];
     son_slave_conf.hostap_ant_gain =
         beerocks::string_utils::stoi(beerocks_slave_conf.hostap_ant_gain[slave_num]);
-    son_slave_conf.radio_identifier        = beerocks_slave_conf.radio_identifier[slave_num];
-    son_slave_conf.backhaul_wireless_iface = beerocks_slave_conf.sta_iface[slave_num];
+    son_slave_conf.radio_identifier = beerocks_slave_conf.radio_identifier[slave_num];
+    son_slave_conf.backhaul_wireless_iface =
+        get_sta_iface_from_hostap_iface(son_slave_conf.hostap_iface);
     son_slave_conf.backhaul_wireless_iface_filter_low =
         beerocks::string_utils::stoi(beerocks_slave_conf.sta_iface_filter_low[slave_num]);
     son_slave_conf.backhaul_wireless_iface_type = son_slave_conf.hostap_iface_type;
@@ -482,14 +502,6 @@ int main(int argc, char *argv[])
                 LOG(DEBUG) << "hostap iface " << beerocks_slave_conf.hostap_iface[slave_num]
                            << " does not exist, clearing it from config";
                 beerocks_slave_conf.hostap_iface[slave_num].clear();
-            }
-        }
-        if (!beerocks_slave_conf.sta_iface[slave_num].empty()) {
-            if (!beerocks::net::network_utils::linux_iface_exists(
-                    beerocks_slave_conf.sta_iface[slave_num])) {
-                LOG(DEBUG) << "sta iface " << beerocks_slave_conf.sta_iface[slave_num]
-                           << " does not exist, clearing it from config";
-                beerocks_slave_conf.sta_iface[slave_num].clear();
             }
         }
     }
--- a/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
+++ b/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
@@ -330,21 +330,6 @@ main_thread::main_thread(config_file::sConfigSlave config_, logging &logger_)
         ap_ifaces.insert(config.hostap_iface[j]);
         i++;
     }
-    for (int j = 0; j < IRE_MAX_SLAVES && i < BPL_NUM_OF_INTERFACES; ++j) {
-        if (config.sta_iface[j].empty())
-            continue;
-
-        auto status = std::make_shared<slave_iface_status>(i, beerocks::eRadioStatus::INVALID, now);
-        if (!status) {
-            LOG(ERROR) << "failed to allocate shared pointer, i=" << i;
-            continue;
-        }
-
-        string_utils::copy_string(bpl_iface_status.ifname[i], config.sta_iface[j].c_str(),
-                                  BPL_IFNAME_LEN);
-        bpl_iface_status_map[config.sta_iface[j]] = status;
-        i++;
-    }
 
     auto status = std::make_shared<slave_iface_status>(i, beerocks::eRadioStatus::INVALID, now);
     if (!status) {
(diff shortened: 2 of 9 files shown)


adam1985d:
[move sta_iface from beerocks_agent.conf to bpl](https://trello.com/c/UEflUZjg/49-move-staiface-from-beerocksagentconf-to-bpl)

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
```diff
@@ -49,6 +49,22 @@ int cfg_get_prplmesh_param(const std::string &param, char *buf, size_t buf_len)
...
+int cfg_get_prplmesh_radio_param(int radio_id, const std::string &radio_param, char *buf,
+                                 size_t buf_len)
+{
+    char path[MAX_UCI_BUF_LEN] = {0};
+
+    if (buf_len > MAX_UCI_BUF_LEN) {
+        buf_len = MAX_UCI_BUF_LEN;
+    }
+
+    if (snprintf_s(path, MAX_UCI_BUF_LEN, "prplmesh.radio%d.%s", radio_id, radio_param.c_str()) <=
```
tomereli, line 61: I personally don't like adding '{' for one liners, @morantr thinks otherwise, but whatever you choose please be consistent.
(nitpick)

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -456,6 +456,17 @@ int cfg_notify_iface_status(const BPL_INTERFACE_STATUS_NOTIFICATION *status_noti
...
+/**
+ * Returns the STA interface for the specified radio id.
+ *
+ * @param [in] iface Interface name for the requested parameters.
+ * @param [out] sta_iface name of STA interface (up to 32 bytes in length).
+ *
+ * @return 0 Success.
+ * @return -1 Error.
+ */
+int cfg_get_sta_iface(const char iface[BPL_IFNAME_LEN], char sta_iface[BPL_IFNAME_LEN]);
```
tomereli, line 468: I think this should have been using explicit size for strings:
```
int cfg_get_sta_iface(const char *iface, size_t iface_length, char *sta_iface, size_t sta_iface_length);
```

Since this is not the case in the whole BPL UCI implementation - (nitpick)

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -220,5 +220,16 @@ int cfg_notify_iface_status(const BPL_INTERFACE_STATUS_NOTIFICATION *status_noti
 
 int cfg_get_administrator_credentials(char pass[BPL_PASS_LEN]) { return RETURN_ERR; }
 
+int cfg_get_sta_iface(const char iface[BPL_IFNAME_LEN], char sta_iface[BPL_IFNAME_LEN])
+{
+    if (iface == NULL || sta_iface == NULL) {
+        return RETURN_ERR;
```
tomereli, line 227: Why not add log(ERR) here?

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -220,5 +220,16 @@ int cfg_notify_iface_status(const BPL_INTERFACE_STATUS_NOTIFICATION *status_noti
...
+    }
+
+    // return empty STA interface name
+    sta_iface[0] = '\0';
+    return RETURN_OK;
+}
+
 } // namespace bpl
 } // namespace beerocks
```
tomereli: Can you add empty line here before end of file?

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -1664,5 +1664,20 @@ bool main_thread::init_dhcp_monitor()
     return true;
 }
 
+std::string main_thread::get_sta_iface(int slave_num)
```
tomereli: If you are not accessing other class parameters here, better make this static instead of class private.

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -149,8 +149,20 @@ static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_
     son_slave_conf.hostap_iface = beerocks_slave_conf.hostap_iface[slave_num];
     son_slave_conf.hostap_ant_gain =
         beerocks::string_utils::stoi(beerocks_slave_conf.hostap_ant_gain[slave_num]);
-    son_slave_conf.radio_identifier        = beerocks_slave_conf.radio_identifier[slave_num];
-    son_slave_conf.backhaul_wireless_iface = beerocks_slave_conf.sta_iface[slave_num];
+    son_slave_conf.radio_identifier = beerocks_slave_conf.radio_identifier[slave_num];
+    // read the sta_iface from bpl and verify it is available
+    char sta_iface_str[BPL_IFNAME_LEN];
```
tomereli: Please move this to a function to keep the main small. The function should be generic if possible since some of the other parameters here will also move to BPL as well.

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -311,6 +303,15 @@ static int run_beerocks_slave(beerocks::config_file::sConfigSlave &beerocks_slav
         // read the number of failures allowed before stopping agent from platform configuration
         int stop_on_failure_attempts = beerocks::bpl::cfg_get_stop_on_failure_attempts();
 
+        // The platform manager updates the beerocks_slave_conf.sta_iface in the init stage
+        std::set<std::string> slave_sta_ifaces;
+        for (int slave_num = 0; slave_num < beerocks::IRE_MAX_SLAVES; slave_num++) {
+
```
tomereli: Remove extra line break

tomereli (approved):
commit 1d400004d37117534df42b0a022915e574bf1a4c - commit message has some typos and its duplicated. Please fix that.

Other than that, excellent work!!! Great split to commits. Commit messages can be slightly improved.
You will need to also create corresponding PRs in UGW and prplWrt feed.
Please update the PR description with links to the UGW PR which adds the STA iface to /etc/config/prplmesh in the feed (its OK that only Intel can access the link).

Please also do a similar change here - https://git.prpl.dev/intel/iwlwav/blob/iwlwav-19.07/prplmesh/files/etc/config/prplmesh - this is the prplWrt used for RAX40. This can be done in a separate, followup PR.

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
adam1985d: added `{` and `}` as I agree with @morantr on this

adam1985d (commented):
> commit [1d40000](https://github.com/prplfoundation/prplMesh/commit/1d400004d37117534df42b0a022915e574bf1a4c) - commit message has some typos and its duplicated. Please fix that.
**done**

> 
> Other than that, excellent work!!! Great split to commits. Commit messages can be slightly improved.
> You will need to also create corresponding PRs in UGW and prplWrt feed.
> Please update the PR description with links to the UGW PR which adds the STA iface to /etc/config/prplmesh in the feed (its OK that only Intel can access the link).

**will be updated - after the review is complete and I will have the correct hash ID**
> 
> Please also do a similar change here - https://git.prpl.dev/intel/iwlwav/blob/iwlwav-19.07/prplmesh/files/etc/config/prplmesh - this is the prplWrt used for RAX40. This can be done in a separate, follow-up PR.
**there is no need to change the /etc/config/prplmesh file since the default configuration is GW only and sta_iface disabled from prplmesh perspective**

framework/platform/bpl/include/bpl/bpl_cfg.h
adam1985d: skipped for now

framework/platform/bpl/linux/bpl_cfg.cpp
adam1985d: added

framework/platform/bpl/linux/bpl_cfg.cpp
adam1985d: added

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
adam1985d: but I am accessing the `config.hostap_iface[]` which is a class parameter.

I changed the API to get hostap_iface and the API is now static.

I changed the corresponding use in the following commit titled "platform_manager - add sta_ifaces to iface_status_map in init"

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: ok - moved to a separate static function

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: done

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
```diff
@@ -49,6 +49,23 @@ int cfg_get_prplmesh_param(const std::string &param, char *buf, size_t buf_len)
...
```
arnout, line 61: Not for this PR, but: why is this file using snprintf_s instead of a sstream?

Even more unrelated: I think it would be convenient if we would import the [fmt library](https://github.com/fmtlib/fmt) and use the C++20 formatter.

framework/platform/bpl/include/bpl/bpl_cfg.h
arnout: > I think this should have been using explicit size for strings:

Actually it should have used std::string.

> Since this is not the case in the whole BPL UCI implementation - (nitpick)

This point remains of course.

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -456,6 +456,17 @@ int cfg_notify_iface_status(const BPL_INTERFACE_STATUS_NOTIFICATION *status_noti
...
 int cfg_get_administrator_credentials(char pass[BPL_USER_PASS_LEN]);
 
+/**
+ * Returns the STA interface for the specified radio id.
+ *
+ * @param [in] iface Interface name for the requested parameters.
+ * @param [out] sta_iface name of STA interface (up to 32 bytes in length).
+ *
+ * @return 0 Success.
+ * @return -1 Error.
```
arnout: It could be clarified here that it also returns -1 if there is no STA interface configured on that radio. (nitpick)

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -281,6 +281,21 @@ std::string extern_query_db(std::string parameter)
     return ret;
 }
 
+static std::string get_sta_iface(const std::string &hostap_iface)
+{
+    char sta_iface_str[BPL_IFNAME_LEN];
+    if (bpl::cfg_get_sta_iface(hostap_iface.c_str(), sta_iface_str) < 0) {
+        LOG(ERROR) << "failed to read sta_iface for slave ";
```
arnout: I don't think this is really an error - it's perfectly fine for a radio not to have a STA configured, and the only way to know whether or not it has one configured is by calling this function. (nitpick)

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -125,6 +125,25 @@ static bool parse_arguments(int argc, char *argv[])
...
+    if (beerocks::bpl::cfg_get_sta_iface(hostap_iface.c_str(), sta_iface_str) < 0) {
+        LOG(ERROR) << "failed to read sta_iface for slave ";
+        return std::string();
+    } else {
+        sta_iface = std::string(sta_iface_str);
+        if (!beerocks::net::network_utils::linux_iface_exists(sta_iface)) {
+            LOG(DEBUG) << "sta iface " << sta_iface << " does not exist, clearing it from config";
+            sta_iface.clear();
+        }
```
arnout, line 142: I think this should move into bpl::cfg_get_sta_iface. It is possible that the platform completely removes the interface when the STA is disabled, so linux_iface_exists will fail if the STA link is not up. So for those platforms, the bpl should take care of validating the correctness in a different way. Also, for example, for dummy bpl we don't really care if the interface exists.

Obviously this can be done in a future PR. I don't even think it's necessary to make an issue for it - we'll handle it if the problem presents itself (i.e. when doing test_flows with wireless backhaul).

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -149,8 +168,9 @@ static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_
     son_slave_conf.hostap_iface = beerocks_slave_conf.hostap_iface[slave_num];
     son_slave_conf.hostap_ant_gain =
         beerocks::string_utils::stoi(beerocks_slave_conf.hostap_ant_gain[slave_num]);
-    son_slave_conf.radio_identifier        = beerocks_slave_conf.radio_identifier[slave_num];
-    son_slave_conf.backhaul_wireless_iface = beerocks_slave_conf.sta_iface[slave_num];
+    son_slave_conf.radio_identifier = beerocks_slave_conf.radio_identifier[slave_num];
+    son_slave_conf.backhaul_wireless_iface =
+        get_sta_iface_from_hostap_iface(son_slave_conf.hostap_iface);
```
arnout, line 173: Hm, I'm not 100% sure if this is OK. bpl supposedly should only be used from the platform_manager thread (which BTW is going to become a separate process - cfr. #575). It is still OK-ish to do it from the main process, but I'm a bit concerned about adding calls to the son_slave - currently there aren't any.

So I think this should instead be part of the wlan_settings in platform_settings and sent by the platform manager in fill_platform_settings.

Clearly this is something that we can handle later.

arnout (approved):
Nice PR. I have comments but nothing blocking merge.

I especially like that the commits are ordered so that nothing ever breaks, and that you added doxygen documentation.

arnout:
I fixed my nitpicks and rebased on master.

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
tomereli: This will be a great addition! I always wondered why c++ fails to have something like this as part of the standard library.

PR #678, opened 2020-01-15
tools: prplmesh-builder: add clang-format

rmelotte:
The CI now uses clang-format from the prplmesh-builder image.

The image on Dockerhub already included clang-format, but the change
to the Dockerfile wasn't included in this repository.

Add clang-format to the prplmesh-builder Dockerfile to match the image
on Dockerhub.

### DIFF
--- a/tools/docker/builder/Dockerfile
+++ b/tools/docker/builder/Dockerfile
@@ -10,28 +10,29 @@ FROM $image
 
 # Update Software repository
 RUN apt-get update && apt-get install -y \
-	curl \
-	gcc \
-	cmake \
-	ninja-build \
-	binutils \
-	git \
-	autoconf \
-	autogen \
-	libtool \
-	pkg-config \
-	libreadline-dev \
-	libncurses-dev \
-	libssl-dev \
-	libjson-c-dev \
-	libnl-genl-3-dev \
-	libssl-dev \
-	libzmq3-dev \
-	python \
-	python-yaml \
-	python3 \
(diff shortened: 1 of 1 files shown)


tools/docker/builder/Dockerfile
```diff
@@ -12,7 +12,8 @@ FROM $image
 RUN apt-get update && apt-get install -y \
 	curl \
 	gcc \
-	cmake \
+        cmake \
```
tomereli: What's up with the indentation here?

tomereli (approved):

tools/docker/builder/Dockerfile
rmelotte: This time it's the other way around: the file is using tabs, and I used spaces in this commit.
In this case it probably makes more sense to keep on using tabs (or convert them all to spaces, but not mixing them).

tools/docker/builder/Dockerfile
```diff
@@ -12,7 +12,8 @@ FROM $image
...
+        clang-format \
```
arnout: I realize it's already not alphabetically ordered, but please put clang-format before cmake :-)

For indentation: I think it would be better to convert the file to (4) spaces like we use in the rest of the project. But only do that if you have time for it :-)

arnout (approved):

rmelotte:
Rebased on master, replaced tabs with 4 spaces, sorted the packages alphabetically.

PR #680, opened 2020-01-15
Hotfix/findnl libs

kantera800:
In UGW, simply adding nl libraries to the list of libraries doesn't
work since the libraries are not in standard locations.
Therefore use find_package which creates an exported target and use that.

### DIFF
--- a/cmake/Findnl-3.cmake
+++ b/cmake/Findnl-3.cmake
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# Copyright (c) 2020 Tomer Eliyahu (Intel Corporation)
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_LIBRARY "libnl-3.so")
+find_path(NL3_INCLUDE_DIRS netlink/netlink.h PATH_SUFFIXES libnl3)
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(nl-3 DEFAULT_MSG
+    NL3_LIBRARY
+    NL3_INCLUDE_DIRS
+)
+
+if (nl-3_FOUND)
+    add_library(nl-3 UNKNOWN IMPORTED)
+
+    # Includes
+    set_target_properties(nl-3 PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${NL3_INCLUDE_DIRS}/"
+    )
+
+    # Library
+    set_target_properties(nl-3 PROPERTIES
+        IMPORTED_LINK_INTERFACE_LANGUAGES "C"
+        IMPORTED_LOCATION "${NL3_LIBRARY}"
+    )
+
+endif()
--- a/cmake/Findnl-genl-3.cmake
+++ b/cmake/Findnl-genl-3.cmake
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# Copyright (c) 2020 Tomer Eliyahu (Intel Corporation)
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_GENL_LIBRARY "libnl-genl-3.so")
+find_path(NL3_INCLUDE_DIRS netlink/genl/genl.h PATH_SUFFIXES libnl3)
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(nl-genl-3 DEFAULT_MSG
+    NL3_GENL_LIBRARY
+    NL3_INCLUDE_DIRS
+)
+
+if (nl-genl-3_FOUND)
+    add_library(nl-genl-3 UNKNOWN IMPORTED)
+
+    # Includes
+    set_target_properties(nl-genl-3 PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${NL3_INCLUDE_DIRS}/"
+    )
+
+    # Library
+    set_target_properties(nl-genl-3 PROPERTIES
+        IMPORTED_LINK_INTERFACE_LANGUAGES "C"
+        IMPORTED_LOCATION "${NL3_GENL_LIBRARY}"
+    )
+
+endif()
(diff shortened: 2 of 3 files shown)


tomereli (approved):

cmake/Findnl-3.cmake
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# Copyright (c) 2019 Intel Corporation
```
arnout: 2019?

Just to be sure: this is really (c) Intel, you didn't copy it from somewhere on the interwebs?

cmake/Findnl-3.cmake
```diff
@@ -0,0 +1,31 @@
...
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_LIBRARY "libnl-3.so")
+find_path(NL3_INCLUDE_DIRS libnl3/netlink/netlink.h)
```
kantera800: I think its better to use PATH_SUFFIXES and NAMS instead of the entire path 

find_path(NL3_INCLUDE_DIRS 
    NAMES netlink.h 
    PATH_SUFFIXES libnl3
)

cmake/Findnl-3.cmake
kantera800: change to 2020

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -54,6 +54,8 @@ if(BWL_TYPE STREQUAL "DWPAL")
     )
 
     find_package(dwpal REQUIRED)
+    find_package(nl-3 REQUIRED)
```
arnout, line 56: libnl3 should be removed from include_directories, above. The target_link_libraries should take care of it (except it doesn't because Findnl-3.cmake is wrong, see my other comment).

cmake/Findnl-3.cmake
arnout: Exactly. PATH_SUFFIXES makes sure that it is set to /usr/include/libnl3, like it should be instead of /usr/include.

NAMES is not needed since we look for only one. And the file name should include the netlink part, because we do `#include <netlink/netlink.h>`. So:

```
find_path(NL3_INCLUDE_DIRS netlink/netlink.h PATH_SUFFIXES libnl3)
```

cmake/Findnl-genl-3.cmake
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# Copyright (c) 2019 Intel Corporation
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_GENL_LIBRARY "libnl-genl-3.so")
+find_path(NL3_INCLUDE_DIRS libnl3/netlink/genl/genl.h)
+message(NL3_INCLUDE_DIRS=${NL3_INCLUDE_DIRS})
```
kantera800: remove

cmake/Findnl-genl-3.cmake
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# Copyright (c) 2019 Intel Corporation
```
arnout: Also copyright year.

cmake/Findnl-genl-3.cmake
```diff
@@ -0,0 +1,32 @@
...
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_GENL_LIBRARY "libnl-genl-3.so")
+find_path(NL3_INCLUDE_DIRS libnl3/netlink/genl/genl.h)
```
arnout: Same comment: use SUFFIX.

cmake/Findnl-genl-3.cmake
```diff
@@ -0,0 +1,32 @@
...
+message(NL3_INCLUDE_DIRS=${NL3_INCLUDE_DIRS})
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(nl-genl-3 DEFAULT_MSG
+    NL3_GENL_LIBRARY
+    NL3_INCLUDE_DIRS
+)
+
+if (nl-genl-3_FOUND)
+    add_library(nl-genl-3 UNKNOWN IMPORTED)
```
arnout, line 18: I believe nl-genl-3 depends on nl-3. I'm not entirely sure if we need to encode that though.

arnout (approved):
Small comments, easily fixed, so please fix them.

arnout:
It would make sense to also use it in all other cases where we use nl-3, but that can be done later.

cmake/Findnl-3.cmake
tomereli: Not copied from anywhere. It is our style of Find modules :)

cmake/Findnl-3.cmake
```diff
@@ -0,0 +1,31 @@
...
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(nl-3 DEFAULT_MSG
+    NL3_LIBRARY
+    NL3_INCLUDE_DIRS
+)
+
+if (nl-3_FOUND)
+    add_library(nl-3 UNKNOWN IMPORTED)
+
+    # Includes
+    set_target_properties(nl-3 PROPERTIES
```
arnout, line 21: This could be `target_include_directories(nl-3 INTERFACE ${NL3_INCLUDE_DIRS})` I think. (nitpick)

cmake/Findnl-3.cmake
```diff
@@ -0,0 +1,31 @@
...
+        INTERFACE_INCLUDE_DIRECTORIES "${NL3_INCLUDE_DIRS}/"
```
arnout, line 22: The trailing slash combines with the plural of NL3_INCLUDE_DIRS sounds fishy... (nitpick)

cmake/Findnl-genl-3.cmake
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# Copyright (c) 2020 Tomer Eliyahu (Intel Corporation)
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_GENL_LIBRARY "libnl-genl-3.so")
+find_path(NL3_INCLUDE_DIRS netlink/genl/genl.h PATH_SUFFIXES libnl3)
```
arnout, line 9: Should be NL3_GENL_INCLUDE_DIRS (nitpick)

arnout (approved):

PR #682, opened 2020-01-15
[TASK] flow test refactoring

ghost:
## Description

Currently, all flows being tested are in a single shell script, with calls to docker to invoke certain commands on emulated agents. 
The plan is to make this more modular by rewriting the test flows in python:
* Make it easier to add new tests
* Put each test in its own file
* Make it extensible so we can run the same tests on actual hardware as well (by abstracting away the target platform)
* Allow test hierarchies (controller/agent/algorithm test seperation, for instance)

## Exit criteria
* A simple python test framework exists
* All existing tests are migrated to this framework
* All tests pass
* All functionality is preserved (i.e. logs, tcpdump captures, ...)

### DIFF
--- a/tools/docker/tests/flows/__init__.py
+++ b/tools/docker/tests/flows/__init__.py
@@ -1,0 +1,112 @@
+#!/usr/bin/python3
+import os
+import unittest
+import re
+from enum import Enum
+import sys
+
+class DeviceType(Enum):
+    gateway = 1
+    agent = 2
+    repeater1 = 3
+    repeater2 = 4
+
+class LogType(Enum):
+    beerocks_agent = 1
+    beerocks_agent_wlan0 = 2
+    beerocks_agent_wlan2 = 3
+    beerocks_monitor_wlan0 = 4
+    beerocks_monitor_wlan2 = 5
+    beerocks_controller = 6
+    transport = 7
+    local_bus = 8
+
+class BaseTest(unittest.TestCase):
+
+    def send_controller_command(self, command):
+        testSystem.send_command(Device.gateway, command)
+
+    def send_agent_command(self, command):
+        testSystem.send_command(Device.agent, command)
+
+    def send_repeater1_command(self, command):
+        testSystem.send_command(Device.repeater1, command)
+
+    def send_repeater2_command(self, command):
+        testSystem.send_command(Device.repeater2, command)
+
+    def find_in_logs(self, device:DeviceType, log:LogType, text:str, ignore_case:bool = True, regexp:bool = False):
+        logs = testSystem.get_log(device, log)
+        if regexp:
+            flags = re.IGNORECASE if ignore_case else 0
+            result = re.search(text, logs, flags)
+        elif ignore_case:
+            result = text.lower() in logs.lower()
+        else:
+            result = test in logs
+        self.debug("Looking for '%s' in %s log on %s ... %s" %
+                  (text, log.name, device.name, "FOUND" if result else "NOT FOUND"))
+
+        return bool(result)
+    
+    def status(self, text):
+        print("\033[1;35m%s\033[0m" % text)
+
+    def error(self, text):
+        print("\033[1;31m%s\033[0m\n" % text)
+
+    def success(self, text):
+        print("\033[1;32m%s\033[0m\n" % text)
+
+    def debug(self, text):
+        if testSystem.verbose:
+            print(text)
+
+class BaseConfig:
+    def __getitem__(self, key):
+        if key in self.values:
+            return self.values[key]
+        raise KeyError("No key %s in config!" % key)
+
+class TestSystem:
+
+    def __init__(self, verbose=False):
+        self.verbose = verbose
+
+    def get_config(self, device:DeviceType) -> BaseConfig:
+        if device == DeviceType.gateway:
+            return self.gateway_config
+        if device == DeviceType.agent:
+            return self.agent_config
+        raise KeyError("No such device!")
+    
+    def get_ip(self, device:DeviceType) -> str:
+        raise NotImplementedError("This should be implemented in the subclass!")
+
+    def get_port(self, device:DeviceType) -> int:
+        raise NotImplementedError("This should be implemented in the subclass!")
+
+    def get_log(self, device:DeviceType, log:LogType):
+        raise NotImplementedError("This should be implemented in the subclass!")
+
+    def send_command(self, device:DeviceType, command:str, wait:bool = True) -> str:
+        with UCCSocket(self.get_ip(device), self.get_port(device)) as sock:
+            sock.send_cmd(command)
+            if wait:
+                return sock.get_reply()
+            return ""
+
+testSystem = None
+testFiles = [f[0:-3] for f in os.listdir(os.path.dirname(os.path.abspath(__file__))) if 'test' in f and f.endswith('.py')]
+
+def load_tests(loader, standard_tests, pattern):
+    
+    suite = unittest.TestSuite()
+    for f in testFiles:
+        testModule = getattr(sys.modules[__name__], f)
+        tests = loader.loadTestsFromModule(testModule)
+        suite.addTests(tests)
+    return suite
+
+__all__ = ['DeviceType', 'LogType', 'BaseTest', 'BaseConfig', 'TestSystem', 'testSystem', 'load_tests'] + testFiles
+
--- a/tools/docker/tests/flows/test_ap_config_renew.py
+++ b/tools/docker/tests/flows/test_ap_config_renew.py
@@ -1,0 +1,56 @@
+from . import BaseTest, LogType, DeviceType
+import time
+"""
+Original shell test:
+    status "test initial autoconfig"
+
+    # Regression test: MAC address should be case insensitive
+    MAC_AGENT1=$(echo $mac_agent1 | tr a-z A-Z)
+    # Configure the controller and send renew
+    send_CAPI_command gateway "DEV_RESET_DEFAULT" $redirect
+    send_CAPI_command gateway "DEV_SET_CONFIG,bss_info1,$MAC_AGENT1 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,bss_info2,$mac_agent1 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0" $redirect
+    gw_mac_without_colons="$(printf $mac_gateway | tr -d :)"
+    send_CAPI_command gateway "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x000A,tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,0x${gw_mac_without_colons},tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{0x00},tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{0x00}}" $redirect
+
+    # Wait a bit for the renew to complete
+    sleep 3
+
+    check docker exec repeater1 sh -c \
+        'grep -i -q "Received credentials for ssid: Multi-AP-24G-1 .* bss_type: 2" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+    check docker exec repeater1 sh -c \
+        'grep -i -q "Received credentials for ssid: Multi-AP-24G-2 .* bss_type: 1" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+    check docker exec repeater1 sh -c \
+        'grep -i -q "ssid: .* teardown" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+
+    send_CAPI_command repeater1 "dev_get_parameter,program,map,ruid,0x000000000000,ssid,Multi-AP-24G-1,parameter,macaddr"
+    check [ "$capi_command_reply" = "status,COMPLETE,macaddr,$mac_agent1_wlan0" ];
+"""
+
+
+class test_initial_ap_config(BaseTest):
+    def test_initial_ap_config(self):
+        MAC_AGENT1="00:11:22:33:44:55"  # TODO
+        gw_mac_without_colons="".join([c for c in "AA:BB:CC:DD:EE:FF" if c != ":"])
+        
+        self.status("test initial autoconfig")
+        self.send_controller_command("DEV_RESET_DEFAULT")
+        self.send_controller_command("DEV_SET_CONFIG,bss_info1,{MAC_AGENT1} 8x Multi-AP-24G-1 " + \
+                                     "0x0020 0x0008 maprocks1 0 1,bss_info2,$mac_agent1 8x Multi-AP-24G-2 " + \
+                                     "0x0020 0x0008 maprocks2 1 0".format(MAC_AGENT1=MAC_AGENT1))
+        self.send_controller_command("DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x000A," + \
+                                     "tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,0x{gw_mac_without_colons}," + \
+                                     "tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{{0x00}}," + \
+                                     "tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{{0x00}}" \
+                                    .format(gw_mac_without_colons=gw_mac_without_colons))
+        time.sleep(3)
+        
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
+                                    "Received credentials for ssid: Multi-AP-24G-1 .* bss_type: 2", regexp=True))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
+                                      "Received credentials for ssid: Multi-AP-24G-2 .* bss_type: 1", regexp=True))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan2,
+                                       "ssid: .* teardown", regexp=True))
+        reply = self.send_repeater1_command("dev_get_parameter,program,map,ruid,0x000000000000,ssid,Multi-AP-24G-1,parameter,macaddr")
+
+        # Not done with this test yet, so make sure it fails
+        self.assertTrue(False)
--- a/tools/docker/tests/flows/test_initial_ap_config.py
+++ b/tools/docker/tests/flows/test_initial_ap_config.py
@@ -1,0 +1,18 @@
+from . import BaseTest, LogType, DeviceType
+
+class test_initial_ap_config(BaseTest):
+    def test_initial_ap_config(self):
+        self.status("test initial autoconfig")
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
+                                      "WSC Global authentication success"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan2,
+                                      "WSC Global authentication success"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
+                                       "KWA (Key Wrap Auth) success"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan2,
+                                      "KWA (Key Wrap Auth) success"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
+                                      "Controller configuration (WSC M2 Encrypted Settings)"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan2,
+                                      "Controller configuration (WSC M2 Encrypted Settings)"))
+        self.assertTrue(False)
--- a/tools/docker/tests/run_tests.py
+++ b/tools/docker/tests/run_tests.py
@@ -1,0 +1,53 @@
+import send_CAPI_command
+import unittest
+import os.path
+import os
+import sys
+import flows
+from flows import *
+from enum import Enum
+
+class StaticConfig(flows.BaseConfig):
+    base_path = "build/install/config"
+
+    def __init__(self, file_name):
+        mypath = os.path.dirname(os.path.abspath(__file__))
+        self.rootdir = os.path.abspath(os.path.join(mypath, "../../.."))
+        configfile = os.path.join(self.rootdir, self.base_path, file_name)
+        lines = []
+        with open(configfile, "r") as f:
+            lines = f.readlines()
+        values = [(k.strip(), v.strip()) for k,v in [line.split("=") for line in lines if "=" in line]]
+        self.values = dict(values)
+
+class LocalTestSystem(flows.TestSystem):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.gateway_config = StaticConfig("beerocks_controller.conf")
+        self.agent_config = StaticConfig("beerocks_controller.conf")
+
+    def get_ip(self, device:DeviceType):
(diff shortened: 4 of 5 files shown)


ghost:
[Split test-flows.sh into agent, controller and algorithm tests](https://trello.com/c/1YHCwLBz/59-split-test-flowssh-into-agent-controller-and-algorithm-tests)

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/python3
```
arnout, line 1: Please use `#!/usr/bin/env python3` so that people who have done a local install of python3 can still work.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+import unittest
+import re
+from enum import Enum
+import sys
+
+class DeviceType(Enum):
+    gateway = 1
+    agent = 2
+    repeater1 = 3
+    repeater2 = 4
```
arnout, line 12: This is a bit inconsistent... We have 3 devices: gateway, repeater1, repeater2. On the repeaters we have only agent, and on the gateway we have agent and controller.

So, either this should be `gateway, repeater1, repeater2`, or `controller, agent_gw, agent_repeater1, agent_repeater2`.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/python3
+import os
+import unittest
+import re
+from enum import Enum
+import sys
+
+class DeviceType(Enum):
```
arnout, line 8: I think it's nicer to use [the functional API](https://docs.python.org/3/library/enum.html#functional-api):

```
DeviceType = Enum('DeviceType', (`gateway', 'agent',  ...))
```

AFAIU the values are not relevant anyway.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+    agent = 2
+    repeater1 = 3
+    repeater2 = 4
+
+class LogType(Enum):
+    beerocks_agent = 1
+    beerocks_agent_wlan0 = 2
+    beerocks_agent_wlan2 = 3
+    beerocks_monitor_wlan0 = 4
+    beerocks_monitor_wlan2 = 5
```
arnout, line 19: Might be nice to make abstraction of the wlan0/wlan2 bit too. So LogType could be beerocks_agent, beerocks_agent_, beerocks_monitor_, ... and if it ends with _, you need an addition argument to specify the radio. Every (agent) device would need a list of radios.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+    def send_agent_command(self, command):
+        testSystem.send_command(Device.agent, command)
+
+    def send_repeater1_command(self, command):
+        testSystem.send_command(Device.repeater1, command)
+
+    def send_repeater2_command(self, command):
+        testSystem.send_command(Device.repeater2, command)
+
+    def find_in_logs(self, device:DeviceType, log:LogType, text:str, ignore_case:bool = True, regexp:bool = False):
```
arnout, line 38: Space after :

Would be nice to flake8 on the file (which we should do in CI, but currently it gives a lot of errors).

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+        logs = testSystem.get_log(device, log)
+        if regexp:
+            flags = re.IGNORECASE if ignore_case else 0
```
arnout, line 41: I think we should avoid ignore_case. In test_flows.sh we don't need it I believe, so there's no reason to add it now.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def send_repeater1_command(self, command):
+        testSystem.send_command(Device.repeater1, command)
+
+    def send_repeater2_command(self, command):
+        testSystem.send_command(Device.repeater2, command)
+
+    def find_in_logs(self, device:DeviceType, log:LogType, text:str, ignore_case:bool = True, regexp:bool = False):
+        logs = testSystem.get_log(device, log)
+        if regexp:
```
arnout, line 40: Is there a reason to support anything else than regular expressions?

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+        return bool(result)
+    
+    def status(self, text):
+        print("\033[1;35m%s\033[0m" % text)
+
+    def error(self, text):
+        print("\033[1;31m%s\033[0m\n" % text)
+
+    def success(self, text):
+        print("\033[1;32m%s\033[0m\n" % text)
```
arnout, line 59: These really shouldn't be used, instead the unittest assertFoo methods should be used. And unittest already prints something when a test starts and finishes IIRC.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+    beerocks_agent_wlan0 = 2
+    beerocks_agent_wlan2 = 3
+    beerocks_monitor_wlan0 = 4
+    beerocks_monitor_wlan2 = 5
+    beerocks_controller = 6
+    transport = 7
+    local_bus = 8
+
+class BaseTest(unittest.TestCase):
+
```
arnout, line 25: Can we add a docstring please? Maybe not needed for all classes and functions, but the main ones should be documented IMO.

But with all my comments below, I don't think there's a need to have a BaseTest class.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+    def get_ip(self, device:DeviceType) -> str:
+        raise NotImplementedError("This should be implemented in the subclass!")
+
+    def get_port(self, device:DeviceType) -> int:
+        raise NotImplementedError("This should be implemented in the subclass!")
+
+    def get_log(self, device:DeviceType, log:LogType):
+        raise NotImplementedError("This should be implemented in the subclass!")
+
+    def send_command(self, device:DeviceType, command:str, wait:bool = True) -> str:
```
arnout, line 92: Again: currently we always wait for the reply, and I don't really see a reason not to wait for it. So let's just remove this until it's needed.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+        with UCCSocket(self.get_ip(device), self.get_port(device)) as sock:
```
arnout, line 93: I think it's better to keep the UCC socket open instead of reconnecting for every command.

I think this is also the only place where get_ip and get_port are used. So if we move the creation of the UCCSocket to the subclass, it can handle that stuff.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+            sock.send_cmd(command)
+            if wait:
+                return sock.get_reply()
+            return ""
+
+testSystem = None
+testFiles = [f[0:-3] for f in os.listdir(os.path.dirname(os.path.abspath(__file__))) if 'test' in f and f.endswith('.py')]
```
arnout, line 100: IMO this is a little too much for a oneliner, so I'd put the listdir results in a variable.

That said: I believe unittest has a function to handle this, no?

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def __init__(self, verbose=False):
+        self.verbose = verbose
+
+    def get_config(self, device:DeviceType) -> BaseConfig:
+        if device == DeviceType.gateway:
+            return self.gateway_config
+        if device == DeviceType.agent:
+            return self.agent_config
+        raise KeyError("No such device!")
```
arnout, line 81: It's not very clear to me if TestSystem is supposed to be a singleton, or if there is an instance per device (where device is gateway, repeater1, repeater2, so not agent).

What I *think* we should have is:

- TestSystem which is a singleton and which abstracts the testsetup. It has a list of Device. One of them is the gateway.
- Device which represents a single device, i.e. gateway, repeater1, repeater2. Device has a bunch of properties like AL-MAC address, list of radios, whether it has a controller. It can also have a bunch of methods like deploy, restart, check_status, send_ucc_command, get_config, check_log. These can have arguments to specify the program it applies to (cfr. LogFile, but it applies to more than just the log file) and sometimes the radio.

I don't think we really need DeviceType. There's a list of devices, the first one is the gateway, the rest are repeaters.

We don't need to implement all of that now, of course, but the basic structure would be nice.

tools/docker/tests/flows/__init__.py
arnout: This stuff is not (or should not) be related to docker any more. So I think it would be more appropriate to create a top-level `tests` directory and put it there.  run_tests.py does still belong to the docker/tests directory. That makes it a bit difficult to import the flows package, but easy to solve by making a symlink.

Others may have a different opinion about that though...

tools/docker/tests/run_tests.py
```diff
@@ -0,0 +1,53 @@
+import send_CAPI_command
+import unittest
+import os.path
+import os
+import sys
+import flows
+from flows import *
```
arnout, line 7: No need to important twice...

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def debug(self, text):
+        if testSystem.verbose:
+            print(text)
+
+class BaseConfig:
+    def __getitem__(self, key):
+        if key in self.values:
+            return self.values[key]
+        raise KeyError("No key %s in config!" % key)
```
arnout, line 69: Isn't it simpler to just derive from dict?

```
class BaseConfig(dict):
    pass
```

There's also a frozendict or something like that, so that values have to be set in derived class __init__ and can't change afterwards.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+    beerocks_monitor_wlan0 = 4
+    beerocks_monitor_wlan2 = 5
+    beerocks_controller = 6
+    transport = 7
+    local_bus = 8
+
+class BaseTest(unittest.TestCase):
+
+    def send_controller_command(self, command):
+        testSystem.send_command(Device.gateway, command)
```
arnout, line 27: I don't think these wrappers add value. You can just as well call this directly from the test.

tools/docker/tests/flows/__init__.py
arnout: I think it makes more sense to implement this on the device directly. So again, I don't think this wrapper adds a lot of value.

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def error(self, text):
+        print("\033[1;31m%s\033[0m\n" % text)
+
+    def success(self, text):
+        print("\033[1;32m%s\033[0m\n" % text)
+
+    def debug(self, text):
+        if testSystem.verbose:
+            print(text)
```
arnout, line 63: Again, I think we can use unittest infra for this: just always print, and if the unittest 'buffer' option is set, it will only be output in case of failure.

tools/docker/tests/flows/test_ap_config_renew.py
```diff
@@ -0,0 +1,56 @@
+from . import BaseTest, LogType, DeviceType
+import time
+"""
+Original shell test:
```
arnout, line 4: I don't think it's very useful to repeat that here. Well, actually for the review it is :-) but when it's going to be merged, it looses its purpose.

So I propose to keep it here, but add another commit that removes it again.

In a similar vein, I think it would be useful to split this PR into a first commit that adds the infrastructure (`flows.__init__.py` and `run_tests.py`), and then two commits that add the two tests. The people who will need to write tests just need to review those latter commits.

tools/docker/tests/flows/__init__.py
arnout, line 92: This is sending a UCC command, so it should be called `send_ucc_command`.

We'll also need a `send_bml_command` at some point :-)

tools/docker/tests/flows/test_ap_config_renew.py
```diff
@@ -0,0 +1,56 @@
...
+    def test_initial_ap_config(self):
+        MAC_AGENT1="00:11:22:33:44:55"  # TODO
+        gw_mac_without_colons="".join([c for c in "AA:BB:CC:DD:EE:FF" if c != ":"])
+        
+        self.status("test initial autoconfig")
+        self.send_controller_command("DEV_RESET_DEFAULT")
+        self.send_controller_command("DEV_SET_CONFIG,bss_info1,{MAC_AGENT1} 8x Multi-AP-24G-1 " + \
+                                     "0x0020 0x0008 maprocks1 0 1,bss_info2,$mac_agent1 8x Multi-AP-24G-2 " + \
+                                     "0x0020 0x0008 maprocks2 1 0".format(MAC_AGENT1=MAC_AGENT1))
+        self.send_controller_command("DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x000A," + \
```
arnout, line 40: `$mac_agent1` can't be right...

tools/docker/tests/flows/__init__.py
arnout: Did you forget to import UCCSocket from send_CAPI_command?

tools/docker/tests/flows/__init__.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def get_log(self, device:DeviceType, log:LogType):
+        raise NotImplementedError("This should be implemented in the subclass!")
+
+    def send_command(self, device:DeviceType, command:str, wait:bool = True) -> str:
+        with UCCSocket(self.get_ip(device), self.get_port(device)) as sock:
+            sock.send_cmd(command)
+            if wait:
+                return sock.get_reply()
+            return ""
```
arnout, line 97: I think we should assert here (which means it has to move to BaseTest after all). Also send_cmd should probably be checked with an assert.

tools/docker/tests/flows/test_ap_config_renew.py
```diff
@@ -0,0 +1,56 @@
...
+                                     "tlv_type1,0x01,tlv_length1,0x0006,tlv_value1,0x{gw_mac_without_colons}," + \
+                                     "tlv_type2,0x0F,tlv_length2,0x0001,tlv_value2,{{0x00}}," + \
+                                     "tlv_type3,0x10,tlv_length3,0x0001,tlv_value3,{{0x00}}" \
+                                    .format(gw_mac_without_colons=gw_mac_without_colons))
+        time.sleep(3)
+        
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
```
arnout, line 47: find_in_logs should also assert.

So it should probably be called assertInLogs.

So, I think we should have a BaseTest which has assertInLogs and assertUCCCommand, which calls into the Device functions and asserts on the result (and also returns it).

Bottom line, I think this should become something like:

```
        self.assertInLogs(testSystem.device[1], LogType.beerocks_agent_wlan0,
                                   "Received credentials for ssid: Multi-AP-24G-1 .* bss_type: 2")
```

tools/docker/tests/run_tests.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def get_ip(self, device:DeviceType):
+        docker_name = "gateway" if device == DeviceType.gateway else "agent"
+        command = "docker exec -t %s ip -f inet addr show br_lan | grep -Po 'inet \K[\d.]+'" % docker_name
+        with os.popen(command) as proc:
+            return proc.readline().strip()
+        raise RuntimeError("Command execution failed! Command=" + command)
+
+    def get_port(self, device:DeviceType):
+        return int(get_config(device)['ucc_listener_port'])
```
arnout, line 38: Since this is the only place where get_config is used, I don't think it makes sense to add all that infrastructure for it.

In general, this kind of information should anyway be accessed through either:

- a UCC command;
- a BML command;
- some other command that is executed on the Device (e.g. to get radio MAC address).

tools/docker/tests/flows/test_initial_ap_config.py
```diff
@@ -0,0 +1,18 @@
...
+                                      "WSC Global authentication success"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
+                                       "KWA (Key Wrap Auth) success"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan2,
+                                      "KWA (Key Wrap Auth) success"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan0,
+                                      "Controller configuration (WSC M2 Encrypted Settings)"))
+        self.assertTrue(self.find_in_logs(DeviceType.repeater1, LogType.beerocks_agent_wlan2,
+                                      "Controller configuration (WSC M2 Encrypted Settings)"))
+        self.assertTrue(False)
```
arnout, line 18: This test is complete now, no?

tools/docker/tests/flows/test_ap_config_renew.py
```diff
@@ -0,0 +1,56 @@
...
+        'grep -i -q "ssid: .* teardown" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+
+    send_CAPI_command repeater1 "dev_get_parameter,program,map,ruid,0x000000000000,ssid,Multi-AP-24G-1,parameter,macaddr"
+    check [ "$capi_command_reply" = "status,COMPLETE,macaddr,$mac_agent1_wlan0" ];
+"""
+
+
+class test_initial_ap_config(BaseTest):
+    def test_initial_ap_config(self):
+        MAC_AGENT1="00:11:22:33:44:55"  # TODO
```
arnout, line 32: We should get this stuff from testSystem. So in my proposal it would be

```
MAC_AGENT1 = testSystem.devices[1].al_mac_addr
```

BTW, the all-caps name is inherited from the shell script, but here it should be snake case.

tools/docker/tests/test_flows.sh
```diff
@@ -72,23 +72,23 @@ send_CAPI_command() {
...
-        'grep -i -q "KWA (Key Wrap Auth) success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec repeater1 sh -c \
-        'grep -i -q "KWA (Key Wrap Auth) success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
-    check docker exec repeater1 sh -c \
-        'grep -i -q "Controller configuration (WSC M2 Encrypted Settings)" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
-    check docker exec repeater1 sh -c \
-        'grep -i -q "Controller configuration (WSC M2 Encrypted Settings)" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
-
-    return $check_error
+#    status "test initial autoconfig"
```
arnout, line 75: Don't comment it out, just remove.

tools/docker/tests/test_flows.sh
```diff
@@ -72,23 +72,23 @@ send_CAPI_command() {
...
+#         'grep -i -q "WSC Global authentication success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+#     check docker exec repeater1 sh -c \
+#         'grep -i -q "KWA (Key Wrap Auth) success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+#     check docker exec repeater1 sh -c \
+#         'grep -i -q "KWA (Key Wrap Auth) success" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+#     check docker exec repeater1 sh -c \
+#         'grep -i -q "Controller configuration (WSC M2 Encrypted Settings)" /tmp/$USER/beerocks/logs/beerocks_agent_wlan0.log'
+#     check docker exec repeater1 sh -c \
+#         'grep -i -q "Controller configuration (WSC M2 Encrypted Settings)" /tmp/$USER/beerocks/logs/beerocks_agent_wlan2.log'
+    python3 ${rootdir}/tools/docker/tests/run_tests.py
```
arnout, line 90: This will run all tests. You should pass an argument to run one specific test (as a transient measure until all tests are converted and we can just call run_tests.py directly).

tools/docker/tests/run_tests.py
```diff
@@ -0,0 +1,53 @@
...
+            return ''.join(log_file.readlines())
+
+    def cleanup(self):
+        pass
+  
+if __name__ == '__main__':
+    flows.testSystem = LocalTestSystem(verbose=True)
+    rv = unittest.main()
+    flows.testSystem.cleanup()
+    sys.exit(rv)
```
arnout, line 53: I like how you set things up here: the generic flows package, and then individual run_tests.py scripts for the different test setups. It is a bit non-intuitive though, so it should be described somewhere. Perhaps a README.md in the flows directory?

arnout (commented):
I have many comments, but nothing really major or blocking - even though some of the comments are pretty fundamental.

In general, I think it's a bit over-engineered, i.e. it covers more than what we actually need.
I'd prefer with something as small as possible that does what we need it to do.

It is unfortunate that you converted to issue into a PR, since now we when we merge the PR (because part1 can be merged already) we loose the issue...
So I propose that you duplicate this into a new issue, and update the PR description to say that it's part1 of that issue.

The commit message has a couple of problems:

- it's not in common, so it shouldn't start with "common";
- should be wrapped at 72 columns (AFAIU vim does this by default);
- refer to an issue with the full URL.

So a corrected commit message would look like this:
```
tools: test_flows: refactor test_flows.sh into python (part 1)

Currently, all flows being tested are in a single shell script, with 
calls to docker to invoke certain commands on emulated agents. The plan
is to make this more modular by rewriting the test flows in python:

*    Make it easier to add new tests
*    Put each test in its own file
*    Make it extensible so we can run the same tests on actual hardware
     as well (by abstracting away the target platform)
*    Allow test hierarchies (controller/agent/algorithm test seperation, for
     instance)

See https://github.com/prplfoundation/prplMesh/issue/<issuename>

Signed-off-by: Frederik Van Bogaert <frederik.vanbogaert@essensium.com>
```

That said, this commit message describes the whole project, not just this one commit.
It's good that it describes the big picture, but it should also say how this single commit fits into it.

So for now I'm not going to approve, nor request changes. The minimum that needs to be done is to make the test pass :-) . But since you may be making bigger changes based on my request, I can't approve yet.

tools/docker/tests/flows/__init__.py
ghost: For the hashbang line: OK, good point
For the tests part: I've actually already done this, just not pushed the commit yet, because I'm still fixing some of the stuff that breaks and I don't feel comfortable pushing commits that break functionality. So it will be part of the merge eventually, but I may just split the work into multiple parts, of which the move of the tests to a top level directory would be one of those.

tools/docker/tests/flows/__init__.py
ghost: Thanks. I haven't worked with Enums in python before (previous real python experience was on version 2.5, which IIRC didn't have them), so I wasn't aware of this API

tools/docker/tests/flows/__init__.py
ghost: The point of the wrapper (and baseclass) is to make the tests look as close to the bash tests as possible, with as few references to other modules or classes as possible.
Just writing `self.send_repeater1_command("...")` and `self.find_in_logs` is easier than writing `from .. import testSystem;` ... `testSystem.send_command(repeater1, "...");` ... `testSystem.find_in_logs("..."); ` IMO. 
Even though that would be a bit cleaner, architecture-wise.

tools/docker/tests/flows/__init__.py
ghost: I guess so. I'll remove it

tools/docker/tests/flows/__init__.py
ghost: Yes! if you search for a string containing a . or other special characters, you would need to remember to escape them if we'd always use regexes.
Again, the aim is to remove complexity from the tests themselves.

tools/docker/tests/flows/__init__.py
ghost: That's true, I guess. I'll remove it

PR #684, opened 2020-01-15
Bugfix/get actual ssid2

### LINKED ISSUE #672: [BUG] TEST-4.4.1 failed - APOperBSSTLV do not contain the correct SSID
**Describe the bug** -
After running test 4.4.1 on the testbed, the following message was received:

image

It turned out that the APOperBSS TLV (which is part of the Topology response CMDU) does not include the real SSID (Multi-AP-24G-1) :

image

## Steps to reproduce

run 4.4.1 test with Qualcomm as the controller and prplMesh as agent
## Expected behavior

The APOperBSS  TLV inside the Topology response CMDU should contain the actual SSID

tomereli:
The topology response contains an AP Operational BSS TLV which contains
a radio info list, where each radio contains a radio bss list which
contains the radui bssid and bssid.

The current implementation used dummy values, which causes a failure in
the agent certification since the UCC compares these values to the
reported mac addresses and ssids recieved via dev_get_parameter.

This commit adds implementation in the backhaul manager to fetch this
data from the VAP list map and prepare the AP Operational BSS TLV with
correct data as received from BWL.

Fixes #672

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1694,6 +1694,8 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
 
             m_agent_ucc_listener->update_vaps_list(network_utils::mac_to_string(msg->ruid()),
                                                    msg->params());
+
+            m_vaps_map[msg->ruid()] = msg->params();
         }
         break;
     }
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1809,15 +1809,22 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
         LOG(ERROR) << "addClass wfa_map::tlvApOperationalBSS failed, mid=" << std::hex << (int)mid;
         return false;
     }
-    //add dummy data
-    auto radio_list     = tlvApOperationalBSS->create_radio_list();
-    auto radio_bss_list = radio_list->create_radio_bss_list();
-    radio_bss_list->set_ssid("wlan_0");
-    radio_list->add_radio_bss_list(radio_bss_list);
-    radio_bss_list = radio_list->create_radio_bss_list();
-    radio_bss_list->set_ssid("wlan_2");
-    radio_list->add_radio_bss_list(radio_bss_list);
-    tlvApOperationalBSS->add_radio_list(radio_list);
+
+    for (const auto &vaps_list : m_vaps_map) {
+        auto radio_list         = tlvApOperationalBSS->create_radio_list();
+        radio_list->radio_uid() = vaps_list.first;
+        for (const auto &vap : vaps_list.second.vaps) {
+            if (vap.mac == network_utils::ZERO_MAC)
+                continue;
+            auto radio_bss_list           = radio_list->create_radio_bss_list();
+            radio_bss_list->radio_bssid() = vap.mac;
+            auto ssid =
+                std::string(vap.ssid, strnlen(vap.ssid, beerocks::message::WIFI_SSID_MAX_LENGTH));
+            radio_bss_list->set_ssid(ssid);
+            radio_list->add_radio_bss_list(radio_bss_list);
+        }
+        tlvApOperationalBSS->add_radio_list(radio_list);
+    }
 
     auto tlvAssociatedClients = cmdu_tx.addClass<wfa_map::tlvAssociatedClients>();
     if (!tlvAssociatedClients) {
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -228,6 +228,8 @@ private:
     std::string bssid_bridge_mac;
 
     std::unique_ptr<beerocks::agent_ucc_listener> m_agent_ucc_listener;
+    // vaps map of ruid (key) to VapsList (value)
+    std::unordered_map<sMacAddr, beerocks_message::sVapsList> m_vaps_map;
 
     /*
  * State Machines


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -228,6 +228,7 @@ class backhaul_manager : public btl::transport_socket_thread {
     std::string bssid_bridge_mac;
 
     std::unique_ptr<beerocks::agent_ucc_listener> m_agent_ucc_listener;
+    std::unordered_map<std::string, beerocks_message::sVapsList> m_vaps_map;
```
arnout: Why is the key a std::string instead of sMacAddr? Is there a reason beyond that it was done wrong for the UCC listener, so we want to make the same mistake again?

I'm sorry if I sound a bit pedantic, but it's annoying me more and more that people keep on using strings for storing MAC addresses, and I had hoped that @tomereli would be one to give a good example...

arnout: Also, please add a doxygen comment, in particular specifying what the key and value are.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1809,15 +1809,22 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
-    radio_bss_list->set_ssid("wlan_0");
-    radio_list->add_radio_bss_list(radio_bss_list);
-    radio_bss_list = radio_list->create_radio_bss_list();
-    radio_bss_list->set_ssid("wlan_2");
-    radio_list->add_radio_bss_list(radio_bss_list);
-    tlvApOperationalBSS->add_radio_list(radio_list);
+
+    for (auto vaps_list : m_vaps_map) {
+        auto radio_list         = tlvApOperationalBSS->create_radio_list();
+        radio_list->radio_uid() = network_utils::mac_from_string(vaps_list.first);
```
arnout: Imagine that, it needs to be converted to a sMacAddr again...

arnout (approved):
Second commit message has a bunch of small spelling/grammar errors, here's a corrected version:

```
The topology response contains an AP Operational BSS TLV which contains
a radio info list, where each radio contains a radio bss list which
contains the VAP ssid and bssid.

The current implementation used dummy values, which causes a failure in
the agent certification since the UCC compares these values to the
reported SSIDs and BSSIDs received via dev_get_parameter.

This commit adds an implementation in the backhaul manager to fetch this
data from the VAP list map and prepare the AP Operational BSS TLV with
correct data as received from BWL.
```

Even though my comment is definitely *NOT* a nitpick :-) I'm approving already since it can easily be fixed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1809,15 +1809,22 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
-    auto radio_list     = tlvApOperationalBSS->create_radio_list();
-    auto radio_bss_list = radio_list->create_radio_bss_list();
-    radio_bss_list->set_ssid("wlan_0");
-    radio_list->add_radio_bss_list(radio_bss_list);
-    radio_bss_list = radio_list->create_radio_bss_list();
-    radio_bss_list->set_ssid("wlan_2");
-    radio_list->add_radio_bss_list(radio_bss_list);
-    tlvApOperationalBSS->add_radio_list(radio_list);
+
+    for (auto vaps_list : m_vaps_map) {
```
rmelotte: Is there any reason to not use a const ref?

(nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1809,15 +1809,22 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+        auto radio_list         = tlvApOperationalBSS->create_radio_list();
+        radio_list->radio_uid() = network_utils::mac_from_string(vaps_list.first);
+        for (auto vap : vaps_list.second.vaps) {
```
rmelotte: Same here, why couldn't vap be a const ref? 

(nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1809,15 +1809,22 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+            if (vap.mac == network_utils::ZERO_MAC)
+                continue;
```
rmelotte, line 1818: For single-lines if, maybe we should agree on either always omitting the braces, or always use them. At the moment it looks like it's up to the developer's preference, which means we may end up with both styles mixed.
If so we should add it to `CODINGSTYLE.md`. clang-format may even be able to detect and/or fix it.

(not for this PR)

rmelotte (approved):
Only nitpicks, so approving.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: https://doodle.com/poll/wwyfaymx84chhe3k

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
tomereli: You don't sound pedantic at all - you are perfectly right, fixed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: Unfortunately, [clang-format doesn't support checking this... clang-tidy does support it](https://stackoverflow.com/questions/26111162/can-clang-format-add-braces-to-single-line-if-statements-etc), but that has to be called with all the relevant -I and -D options, so separately in each build type.

PR #687, opened 2020-01-16
Add operator<< for sMacAddr

arnout:
We very often need to log a MAC address. Instead of calling 
`network_utils::mac_to_string` all the time, it's much easier to define an 
additional operator<< for elpp that handles sMacAddr.

This PR adds that operator, and also replaces all the instances of `network_utils::mac_to_string` that have become redundant.

This is a spin-off of #431. They are kind of independent though: the first commit is shared between the two PRs, the others are not. When one is merged, the other
will have to be rebased.


tomereli (approved):

arnout:
[sMacAddr output](https://trello.com/c/0TqOFkOj/72-smacaddr-output)

PR #643, opened 2020-01-06
[DCS Patch 2/5] port dynamic channel selection from rdkb infrastructure

itayx:
> The Dynamic Channel Selection as a feature can provide information on the supported channels which can be utilized by the management system to provide better availability and user experience.
#562

The Dynamic Channel Selection feature is divided into several pull request, this is the second out of 5
start review from 9db41a9

### High Level Infastructure (Controller & yaml): Support for Channel Scan & DCS:
Changes in this PR include the following:
- beerocks defines: enums for Channel Scan & DCS
- yaml: beerocks_message_common: Add channel scan structures
- yaml: beerocks_message_control: Add channel scan & DCS CMDUs
- controller: node: Add channel scan structures
- controller: db: Add channel scan DB methods

PR #689, opened 2020-01-19
Hotfix/channel scan trigger

itayx:
Hot fix for channel scan
the function channel_scan_trigger in mon_wlan_hal_dwpal
needs to be inherited from mon_wlan_hal in order to be accessible from monitor_thread

This PR needs to merged after #643

### DIFF
--- a/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
@@ -150,6 +150,13 @@ bool mon_wlan_hal_dummy::sta_link_measurements_11k_request(const std::string &st
     return true;
 }
 
+bool mon_wlan_hal_dummy::channel_scan_trigger(int dwell_time_msec,
+                                              const std::vector<unsigned int> &channel_pool)
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED";
+    return false;
+}
+
 bool mon_wlan_hal_dummy::process_dummy_data(parsed_obj_map_t &parsed_obj)
 {
     char *tmp_str;
--- a/common/beerocks/bwl/nl80211/mon_wlan_hal_nl80211.cpp
+++ b/common/beerocks/bwl/nl80211/mon_wlan_hal_nl80211.cpp
@@ -385,6 +385,13 @@ bool mon_wlan_hal_nl80211::sta_link_measurements_11k_request(const std::string &
     return true;
 }
 
+bool mon_wlan_hal_nl80211::channel_scan_trigger(int dwell_time_msec,
+                                                const std::vector<unsigned int> &channel_pool)
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED!";
+    return false;
+}
+
 bool mon_wlan_hal_nl80211::process_nl80211_event(parsed_obj_map_t &parsed_obj)
 {
     // Filter out empty events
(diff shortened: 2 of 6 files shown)


common/beerocks/bwl/nl80211/mon_wlan_hal_nl80211.cpp
```diff
@@ -385,6 +385,13 @@ bool mon_wlan_hal_nl80211::sta_link_measurements_11k_request(const std::string &
     return true;
 }
 
+bool mon_wlan_hal_nl80211::channel_scan_trigger(int dwell_time_msec,
+                                                const std::vector<unsigned int> &channel_pool)
+{
+    LOG(TRACE) << __func__;
```
kantera800: add  " - NOT IMPLEMENTED!";
return false

kantera800:
These shouldn't be any dependency on other PRs
Please rebase it to master.
also, no reason for hotfix to be draft :)

arnout:
> These shouldn't be any dependency on other PRs

Actually, it's fine to have a PR that depends on other PRs. But indeed, that forces the PRs to be merged in that particular order, and that's probably not the intention here.

itayx:
> > These shouldn't be any dependency on other PRs
> 
> Actually, it's fine to have a PR that depends on other PRs. But indeed, that forces the PRs to be merged in that particular order, and that's probably not the intention here.

Actually it was
I didn't want to rebase the 2nd PR to the hotfix, and It's only relevant to the 4th PR
I wanted to keep the PR order and flow of commits as neat as possible
since the 2nd PR should be merged soon, I didn't believe it was an issue

common/beerocks/bwl/nl80211/mon_wlan_hal_nl80211.cpp
itayx: Changed, but If i'm not mistaken, all other "not implemented" methods still return true

arnout:
[hotfix channel scan trigger](https://trello.com/c/JPrd3HMX/81-hotfix-channel-scan-trigger)

kantera800 (approved):

tomereli (approved):

arnout (approved):

arnout:
Sign-off-by is missing so I can't merge.

Also it needs to be rebased on master.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -150,6 +150,13 @@ bool mon_wlan_hal_dummy::sta_link_measurements_11k_request(const std::string &st
     return true;
 }
 
+bool mon_wlan_hal_dummy::channel_scan_trigger(int dwell_time_msec,
+                                              const std::vector<unsigned int> &channel_pool)
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED";
```
ghost, line 156: nitpick: `__PRETTY_FUNCTION__` is generally more useful in C++ (https://gcc.gnu.org/onlinedocs/gcc/Function-Names.html)
Although non-standard.

kantera800: In my opinion its "pretty" only for  development stage.
Not sure that it will give any added value to a QA engineer while debugging.
Also its significantly increased log file size.

ghost: It might be, depending on how many subclasses implement the method, and how many function overrides there are for the method (and templates, namespaces etc).
If there's more than one version of the function, I would want to know which one was triggered, even as a QA engineer while debugging. In fact, especially then.

PR #690, opened 2020-01-20
cmake: Disable auto-roll on RDKB

morantr:
RDKB uses syslog-ng which has an internal mechanism to perform log roll.
When the auto roll definition is set on the CMake, we use Beerocks internal,
roll mechanism.
If it is set on RDKB, it causes the logs to be empty after the
automatic beerocks roll.
Disable the Beerocks auto-roll on RDKB to fix it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,7 @@ elseif (TARGET_PLATFORM STREQUAL "rdkb")
     # Disable file and enable syslog logging
     set(BEEROCKS_LOG_FILES_ENABLED "false")
     set(BEEROCKS_LOG_SYSLOG_ENABLED "true")
+    set(BEEROCKS_LOG_FILES_AUTO_ROLL "false")
     set(BEEROCKS_LOG_FILES_PATH "/rdklogs/logs")
     set(BEEROCKS_LOG_FILES_SUFFIX ".txt.0")
     set(CMAKE_SKIP_RPATH TRUE)


tomereli (approved):

vitalybu (approved):

CMakeLists.txt
```diff
@@ -136,6 +136,7 @@ elseif (TARGET_PLATFORM STREQUAL "rdkb")
     # Disable file and enable syslog logging
     set(BEEROCKS_LOG_FILES_ENABLED "false")
     set(BEEROCKS_LOG_SYSLOG_ENABLED "true")
+    set(BEEROCKS_LOG_FILES_AUTO_ROLL "false")
```
arnout, line 139: I think I'm missing something here... If log files are disabled and instead we're logging to syslog, what is auto roll still doing? Actually the LOG_FILES_PATH and LOG_FILES_SUFFIX changes have the same issue.

arnout (commented):
I wonder if it wouldn't be better to remove the autoroll feature entirely. On a real platform, we should always log to syslog. Re-implementing the system's log rotation mechanism, whatever it is, doesn't make a lot of sense IMO.

CMakeLists.txt
morantr: The beerocks log performs a different roll then the Syslog.
The syslog roll done by writing to the beginning of the file. If you ask me it is less convenient.
Beerocks roll put all the logs in an archive and create new logs files. At most, only the latest 3 archives are saved.

PR #694, opened 2020-01-20
[DCS Patch 4/5] port dynamic channel selection from rdkb flow

itayx:
> The Dynamic Channel Selection as a feature can provide information on the supported channels which can be utilized by the management system to provide better availability and user experience.
#562

The Dynamic Channel Selection feature is divided into several pull request, this is the forth out of 5
### Dynamic channel selection message flow & actions
Changes in this PR include the following:
- agent: monitor_thread: proccess nl events
- agent: monitor_thread: handle channel scan events
- agent: monitor_thread: handle channel scan trigger CMDU
- controller & agent: master & slave: DCS channel scan
- controller: son_managment: DCS channel scan stubs
- controller: son_managment: DCS channel scan
- controller: bml: DCS channel scan setup
- controller: bml: DCS channel scan, set/get enable
- controller: bml: DCS channel scan, set/get params
- controller: bml: DCS channel scan, start single scan
- controller: bml: DCS channel scan, get results

**Before merging this PR please merge hotfix #695**

### DIFF
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -1409,13 +1409,50 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
 
     } break;
     case Event::Channel_Scan_Triggered: {
+        auto notification = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_TRIGGERED_NOTIFICATION>(cmdu_tx);
+        if (!notification) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_TRIGGERED_NOTIFICATION msg";
+            return false;
+        }
+
+        message_com::send_cmdu(slave_socket, cmdu_tx);
     } break;
     case Event::Channel_Scan_New_Results_Ready:
     case Event::Channel_Scan_Dump_Result: {
+        auto notification = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION>(cmdu_tx);
+        if (!notification) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION msg";
+            return false;
+        }
+
+        // If event == Channel_Scan_New_Results_Ready do nothing since is_dump's default is 0
+        if (event == Event::Channel_Scan_Dump_Result) {
+            notification->is_dump() = 1;
+        }
+
+        message_com::send_cmdu(slave_socket, cmdu_tx);
     } break;
     case Event::Channel_Scan_Finished: {
+        auto notification = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_FINISHED_NOTIFICATION>(cmdu_tx);
+        if (!notification) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_FINISHED_NOTIFICATION msg";
+            return false;
+        }
+
+        message_com::send_cmdu(slave_socket, cmdu_tx);
     } break;
     case Event::Channel_Scan_Abort: {
+        auto notification = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_ABORT_NOTIFICATION>(cmdu_tx);
+        if (!notification) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_ABORT_NOTIFICATION msg";
+            return false;
+        }
+
+        message_com::send_cmdu(slave_socket, cmdu_tx);
     } break;
 
     // Unhandled events
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -15,6 +15,7 @@
 #include <beerocks/tlvf/beerocks_message.h>
 
 #include <cmath>
+#include <vector>
 
 using namespace beerocks;
 using namespace net;
@@ -380,7 +381,6 @@ void monitor_thread::after_select(bool timeout)
                     stop_monitor_thread();
                     return;
                 }
-                LOG(DEBUG) << "proccess nl events finished ";
                 clear_ready(mon_hal_nl_events);
             }
         }
@@ -1143,6 +1143,52 @@ bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
                    << "mac=" << mac_str;
 
         mon_wlan_hal->sta_link_measurements_11k_request(mac_str);
+        break;
+    }
+    case beerocks_message::ACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST: {
+        auto request =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST>();
+        if (!request) {
+            LOG(ERROR) << "addClass cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST failed";
+            return false;
+        }
+
+        auto radio_mac         = request->scan_params().radio_mac;
+        auto dwell_time_ms     = request->scan_params().dwell_time_ms;
+        auto channel_pool      = request->scan_params().channel_pool;
+        auto channel_pool_size = request->scan_params().channel_pool_size;
+        auto channel_pool_vector =
+            std::vector<unsigned int>(channel_pool, channel_pool + channel_pool_size);
+        std::string channels;
+
+        //loop for priting the channal pool
+        for (int index = 0; index < int(channel_pool_size); index++) {
+            channels += ((index != 0) ? "," : "") + std::to_string(channel_pool[index]);
+        }
+
+        //debug print incoming information:
+        LOG(DEBUG) << std::endl
+                   << "scan_params:" << std::endl
+                   << "radio_mac=" << radio_mac << std::endl
+                   << "dwell_time_ms=" << dwell_time_ms << std::endl
+                   << "channel_pool_size=" << int(channel_pool_size) << std::endl
+                   << "channel_pool=" << channels;
+
+        auto response_out = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE>(
+            cmdu_tx, beerocks_header->id());
+        if (!response_out) {
+            LOG(ERROR)
+                << "Failed building cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE message!";
+            return false;
+        }
+
+        bool result = mon_wlan_hal->channel_scan_trigger(int(dwell_time_ms), channel_pool_vector);
+        LOG_IF(!result, ERROR) << "channel_scan_trigger Failed";
+
+        response_out->success() = (result) ? 1 : 0;
+        message_com::send_cmdu(slave_socket, cmdu_tx);
         break;
     }
     default: {
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -79,10 +79,10 @@ void monitor_thread::stop_monitor_thread()
     } else {
 
         LOG(ERROR) << "disconnecting monitor_thread sockets";
-        if (mon_hal_ext_events || slave_socket)
-            LOG(DEBUG) << "stop_monitor_thread()";
-
+        LOG_IF(mon_hal_ext_events || mon_hal_nl_events || slave_socket, DEBUG)
+            << "stop_monitor_thread()";
         if (mon_hal_ext_events) {
+            LOG(DEBUG) << "stopping mon_hal_ext_events!";
             mon_rssi.stop();
             mon_stats.stop();
 #ifdef BEEROCKS_RDKB
@@ -96,12 +96,21 @@ void monitor_thread::stop_monitor_thread()
         }
 
         if (mon_hal_int_events) {
+            LOG(DEBUG) << "stopping mon_hal_int_events!";
             remove_socket(mon_hal_int_events);
             delete mon_hal_int_events;
             mon_hal_int_events = nullptr;
         }
 
+        if (mon_hal_nl_events) {
+            LOG(DEBUG) << "stopping mon_hal_nl_events!";
+            remove_socket(mon_hal_nl_events);
+            delete mon_hal_nl_events;
+            mon_hal_nl_events = nullptr;
+        }
+
         if (slave_socket) {
+            LOG(DEBUG) << "stopping slave_socket!";
             remove_socket(slave_socket);
             slave_socket->closeSocket();
             delete slave_socket;
@@ -133,6 +142,11 @@ bool monitor_thread::socket_disconnected(Socket *sd)
     } else if (mon_hal_int_events && (sd == mon_hal_int_events)) {
         LOG(ERROR) << "mon_hal_int_events socket disconnected!";
         thread_last_error_code = MONITOR_THREAD_ERROR_HAL_DISCONNECTED;
+        stop_monitor_thread();
+        return false;
+    } else if (mon_hal_nl_events && (sd == mon_hal_nl_events)) {
+        LOG(ERROR) << "mon_hal_nl_events socket disconnected!";
+        thread_last_error_code = MONITOR_THREAD_ERROR_NL_EVENTS_SOCKET_DISCONNECTED;
         stop_monitor_thread();
         return false;
     }
@@ -262,6 +276,17 @@ void monitor_thread::after_select(bool timeout)
                 return;
             }
 
+            int nl_events_fd = mon_wlan_hal->get_nl_events_fd();
+            if (nl_events_fd > 0) {
+                mon_hal_nl_events = new Socket(nl_events_fd);
+                add_socket(mon_hal_nl_events);
+                LOG(DEBUG) << "nl socket created for FD #" << nl_events_fd;
+            } else {
+                LOG(ERROR) << "Couldn't get NL socket ";
+                mon_hal_nl_events      = nullptr;
+                thread_last_error_code = MONITOR_THREAD_ERROR_NL_ATTACH_FAIL;
+            }
+
             // start local monitors //
             LOG(TRACE) << "mon_stats.start()";
             if (!mon_stats.start(&mon_db, slave_socket)) {
@@ -343,6 +368,20 @@ void monitor_thread::after_select(bool timeout)
                 thread_last_error_code = MONITOR_THREAD_ERROR_REPORT_PROCESS_FAIL;
                 stop_monitor_thread();
                 return;
+            }
+        }
+
+        // Process nl events
+        if (mon_hal_nl_events) {
+            if (read_ready(mon_hal_nl_events)) {
+                if (!mon_wlan_hal->process_nl_events()) {
+                    LOG(ERROR) << "process_nl_events() failed!";
+                    thread_last_error_code = MONITOR_THREAD_ERROR_NL_REPORT_PROCESS_FAIL;
+                    stop_monitor_thread();
+                    return;
+                }
+                LOG(DEBUG) << "proccess nl events finished ";
+                clear_ready(mon_hal_nl_events);
             }
         }
 
@@ -1368,6 +1407,15 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
 
         update_vaps_in_db();
 
+    } break;
+    case Event::Channel_Scan_Triggered: {
+    } break;
+    case Event::Channel_Scan_New_Results_Ready:
+    case Event::Channel_Scan_Dump_Result: {
+    } break;
+    case Event::Channel_Scan_Finished: {
+    } break;
+    case Event::Channel_Scan_Abort: {
     } break;
 
     // Unhandled events
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -1123,6 +1123,30 @@ bool slave_thread::handle_cmdu_control_message(Socket *sd,
 
         break;
     }
+    case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST: {
+        LOG(TRACE) << "ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST";
+        auto request_in =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST>();
+        if (!request_in) {
+            LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST failed";
+            return false;
+        }
+
+        auto request_out = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST>(cmdu_tx);
+        if (!request_out) {
+            LOG(ERROR)
+                << "Failed building cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST message!";
+            return false;
+        }
+
+        request_out->scan_params() = request_in->scan_params();
+
+        LOG(DEBUG) << "send cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST";
+        message_com::send_cmdu(monitor_socket, cmdu_tx);
+        break;
+    }
     default: {
         LOG(ERROR) << "Unknown CONTROL message, action_op: " << int(beerocks_header->action_op());
         return false;
@@ -2791,6 +2815,105 @@ bool slave_thread::handle_cmdu_monitor_message(Socket *sd,
             return false;
         }
         notification_out->params() = notification_in->params();
+        send_cmdu_to_controller(cmdu_tx);
+        break;
+    }
+    case beerocks_message::ACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE: {
+        auto response_in =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE>();
+        if (!response_in) {
+            LOG(ERROR) << "addClass cACTION_MONITOR_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE failed";
+            return false;
+        }
+
+        auto response_out = message_com::create_vs_message<
+            beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE>(cmdu_tx);
+        if (!response_out) {
+            LOG(ERROR) << "Failed building cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE";
+            return false;
+        }
+
+        response_out->success() = response_in->success();
+        send_cmdu_to_controller(cmdu_tx);
+        break;
+    }
+    case beerocks_message::ACTION_MONITOR_CHANNEL_SCAN_TRIGGERED_NOTIFICATION: {
+        auto notification_in =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_TRIGGERED_NOTIFICATION>();
+        if (!notification_in) {
+            LOG(ERROR) << "addClass cACTION_MONITOR_CHANNEL_SCAN_TRIGGERED_NOTIFICATION failed";
+            return false;
+        }
+
+        auto notification_out = message_com::create_vs_message<
+            beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION>(cmdu_tx);
+        if (!notification_out) {
+            LOG(ERROR) << "Failed building cACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION !";
+            return false;
+        }
+
+        send_cmdu_to_controller(cmdu_tx);
+        break;
+    }
+    case beerocks_message::ACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION: {
+        auto notification_in =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION>();
+        if (!notification_in) {
+            LOG(ERROR) << "addClass cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION failed";
+            return false;
+        }
+
+        auto notification_out = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION>(cmdu_tx);
+        if (!notification_out) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION !";
+            return false;
+        }
+
+        notification_out->scan_results() = notification_in->scan_results();
+        notification_out->is_dump()      = notification_in->is_dump();
+
+        send_cmdu_to_controller(cmdu_tx);
+        break;
+    }
+    case beerocks_message::ACTION_MONITOR_CHANNEL_SCAN_FINISHED_NOTIFICATION: {
+        auto notification_in =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_FINISHED_NOTIFICATION>();
+        if (!notification_in) {
+            LOG(ERROR) << "addClass cACTION_MONITOR_CHANNEL_SCAN_FINISHED_NOTIFICATION failed";
+            return false;
+        }
+
+        auto notification_out = message_com::create_vs_message<
+            beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_FINISHED_NOTIFICATION>(cmdu_tx);
+        if (!notification_out) {
+            LOG(ERROR) << "Failed building cACTION_CONTROL_CHANNEL_SCAN_FINISHED_NOTIFICATION !";
+            return false;
+        }
+
+        send_cmdu_to_controller(cmdu_tx);
+        break;
+    }
+    case beerocks_message::ACTION_MONITOR_CHANNEL_SCAN_ABORT_NOTIFICATION: {
+        auto notification_in =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_ABORT_NOTIFICATION>();
+        if (!notification_in) {
+            LOG(ERROR) << "addClass cACTION_MONITOR_CHANNEL_SCAN_ABORT_NOTIFICATION failed";
+            return false;
+        }
+
+        auto notification_out = message_com::create_vs_message<
+            beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION>(cmdu_tx);
+        if (!notification_out) {
+            LOG(ERROR) << "Failed building cACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION !";
+            return false;
+        }
+
         send_cmdu_to_controller(cmdu_tx);
         break;
     }
--- a/controller/src/beerocks/bml/bml.cpp
+++ b/controller/src/beerocks/bml/bml.cpp
@@ -564,19 +564,21 @@ int bml_set_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int e
         return (-BML_RET_INVALID_ARGS);
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    auto pBML = static_cast<bml_internal *>(ctx);
+    return pBML->set_dcs_continuous_scan_enable(
+        network_utils::mac_from_string(std::string(radio_mac)), enable);
 }
 
-int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int *output_enable)
+int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int *enable)
 {
     // Validate input parameters
     if (!ctx) {
         return (-BML_RET_INVALID_ARGS);
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    auto pBML = static_cast<bml_internal *>(ctx);
+    return pBML->get_dcs_continuous_scan_enable(
+        network_utils::mac_from_string(std::string(radio_mac)), *enable);
 }
 
 int bml_set_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int dwell_time,
--- a/controller/src/beerocks/bml/bml.cpp
+++ b/controller/src/beerocks/bml/bml.cpp
@@ -633,6 +633,7 @@ int bml_start_dcs_single_scan(BML_CTX ctx, const char *radio_mac, int dwell_time
         return (-BML_RET_INVALID_ARGS);
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    auto pBML = static_cast<bml_internal *>(ctx);
+    return pBML->start_dcs_single_scan(network_utils::mac_from_string(std::string(radio_mac)),
+                                       dwell_time, channel_pool, channel_pool_size);
 }
--- a/controller/src/beerocks/bml/bml.cpp
+++ b/controller/src/beerocks/bml/bml.cpp
@@ -590,8 +590,10 @@ int bml_set_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int d
         return (-BML_RET_INVALID_ARGS);
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    auto pBML = static_cast<bml_internal *>(ctx);
+    return pBML->set_dcs_continuous_scan_params(
+        network_utils::mac_from_string(std::string(radio_mac)), dwell_time, interval_time,
+        channel_pool, channel_pool_size);
 }
 
 int bml_get_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int *output_dwell_time,
@@ -603,8 +605,10 @@ int bml_get_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int *
         return (-BML_RET_INVALID_ARGS);
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    auto pBML = static_cast<bml_internal *>(ctx);
+    return pBML->get_dcs_continuous_scan_params(
+        network_utils::mac_from_string(std::string(radio_mac)), output_dwell_time,
+        output_interval_time, output_channel_pool, output_channel_pool_size);
 }
 
 int bml_get_dcs_scan_results(BML_CTX ctx, const char *radio_mac,
--- a/controller/src/beerocks/bml/bml.cpp
+++ b/controller/src/beerocks/bml/bml.cpp
@@ -612,7 +612,7 @@ int bml_get_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int *
 }
 
 int bml_get_dcs_scan_results(BML_CTX ctx, const char *radio_mac,
-                             struct BML_DCS_NEIGHBOR_AP **output_results,
+                             struct BML_NEIGHBOR_AP **output_results,
                              unsigned int *output_results_size, unsigned char *output_result_status,
                              bool is_single_scan)
 {
@@ -621,8 +621,10 @@ int bml_get_dcs_scan_results(BML_CTX ctx, const char *radio_mac,
         return (-BML_RET_INVALID_ARGS);
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    auto pBML = static_cast<bml_internal *>(ctx);
+    return pBML->get_dcs_scan_results(network_utils::mac_from_string(std::string(radio_mac)),
+                                      output_results, *output_results_size, *output_results_size,
+                                      *output_result_status, is_single_scan);
 }
 
 int bml_start_dcs_single_scan(BML_CTX ctx, const char *radio_mac, int dwell_time,
--- a/controller/src/beerocks/bml/bml.h
+++ b/controller/src/beerocks/bml/bml.h
@@ -554,7 +554,7 @@ int bml_set_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int e
  *
  * @return BML_RET_OK on success.
  */
-int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int *output_enable);
+int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int *enable);
 
 /**
  * set DCS continuous scan params.
@@ -601,7 +601,7 @@ int bml_get_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int *
  * @return BML_RET_OK on success.
  */
 int bml_get_dcs_scan_results(BML_CTX ctx, const char *radio_mac,
-                             struct BML_DCS_NEIGHBOR_AP **output_results,
+                             struct BML_NEIGHBOR_AP **output_results,
                              unsigned int *output_results_size, unsigned char *output_result_status,
                              bool is_single_scan);
 
--- a/controller/src/beerocks/bml/internal/bml_internal.cpp
+++ b/controller/src/beerocks/bml/internal/bml_internal.cpp
@@ -27,8 +27,12 @@ INITIALIZE_EASYLOGGINGPP
 ////////////////////////// Local Module Definitions //////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-#define SELECT_TIMEOUT (500)    // 500 milliseconds
-#define RESPONSE_TIMEOUT (5000) // 5 seconds
+//These timeout definitions used throughout the code.
+//Usually used for setting wait periods for promises,
+//and sleeping until a partial or complete response arrives.
+#define SELECT_TIMEOUT (500)             // equivelent of 0.5 seconds
+#define RESPONSE_TIMEOUT (5000)          // equivelent of 5 seconds
+#define DELAYED_RESPONSE_TIMEOUT (30000) // equivelent of 30 seconds
 
 //////////////////////////////////////////////////////////////////////////////
 //////////////////////// Static Members Initialization ///////////////////////
@@ -883,6 +887,18 @@ int bml_internal::process_cmdu_header(std::shared_ptr<beerocks_header> beerocks_
             }
             *m_pvaps_list_size = i;
 
+        } break;
+        case beerocks_message::ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_ENABLE_RESPONSE: {
+        } break;
+        case beerocks_message::ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_ENABLE_RESPONSE: {
+        } break;
+        case beerocks_message::ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_RESPONSE: {
+        } break;
+        case beerocks_message::ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_RESPONSE: {
+        } break;
+        case beerocks_message::ACTION_BML_CHANNEL_SCAN_GET_RESULTS_RESPONSE: {
+        } break;
+        case beerocks_message::ACTION_BML_CHANNEL_SCAN_START_SCAN_RESPONSE: {
         } break;
         default: {
             LOG(WARNING) << "unhandled header BML action type 0x" << std::hex
--- a/controller/src/beerocks/bml/internal/bml_internal.cpp
+++ b/controller/src/beerocks/bml/internal/bml_internal.cpp
@@ -920,9 +920,42 @@ int bml_internal::process_cmdu_header(std::shared_ptr<beerocks_header> beerocks_
                              << " response, but no one is waiting...";
             }
         } break;
-        case beerocks_message::ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_RESPONSE: {
+                case beerocks_message::ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_RESPONSE: {
+            auto response = beerocks_header->addClass<
+                beerocks_message::cACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_RESPONSE>();
+            if (!response) {
+                LOG(ERROR)
+                    << "addClass cACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_RESPONSE failed";
+                return BML_RET_OP_FAILED;
+            }
+
+            //Signal any waiting threads
+            if (m_prmChannelScanParamsGet) {
+                if (m_scan_params != nullptr) {
+                    *m_scan_params = response->params();
+                    m_prmChannelScanParamsGet->set_value(0);
+                }
+                m_prmChannelScanParamsGet = nullptr;
+            } else {
+                LOG(WARNING) << "Received ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_RESPONSE"
+                             << " response, but no one is waiting...";
+            }
         } break;
         case beerocks_message::ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_RESPONSE: {
+            auto response = beerocks_header->addClass<
+                beerocks_message::cACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_RESPONSE>();
+            if (!response) {
+                LOG(ERROR)
+                    << "addClass cACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_RESPONSE failed";
+                return BML_RET_OP_FAILED;
+            }
+
+            //Signal any waiting threads
+            if (!wake_up(beerocks_message::ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_REQUEST,
+                         response->op_error_code())) {
+                LOG(WARNING) << "Received ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_RESPONSE"
+                             << " response, but no one is waiting...";
+            }
         } break;
         case beerocks_message::ACTION_BML_CHANNEL_SCAN_GET_RESULTS_RESPONSE: {
         } break;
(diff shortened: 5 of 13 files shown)


agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -79,7 +79,7 @@ void monitor_thread::stop_monitor_thread()
     } else {
 
         LOG(ERROR) << "disconnecting monitor_thread sockets";
-        if (mon_hal_ext_events || slave_socket)
+        if (mon_hal_ext_events || mon_hal_nl_events || slave_socket)
```
vitalybu: We always use `{}` with `if` statements (even one-liners)

vitalybu: What is the purpose of this `if` statement anyway?

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -101,6 +101,13 @@ void monitor_thread::stop_monitor_thread()
             mon_hal_int_events = nullptr;
         }
 
+        if (mon_hal_nl_events) {
+            remove_socket(mon_hal_nl_events);
+            delete mon_hal_nl_events;
+            mon_hal_nl_events = nullptr;
+            LOG(ERROR) << "nl socket remove";
```
vitalybu: Remove this log

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -262,6 +274,21 @@ void monitor_thread::after_select(bool timeout)
                 return;
             }
 
+            int nl_events_fd = mon_wlan_hal->get_nl_events_fd();
+            if (nl_events_fd > 0) {
+                mon_hal_nl_events = new Socket(nl_events_fd);
+                add_socket(mon_hal_nl_events);
+                LOG(DEBUG) << "nl socket created ";
```
vitalybu: Maybe it's a good idea to print the FD of the newly created socket... Can be useful in future logging

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -262,6 +274,21 @@ void monitor_thread::after_select(bool timeout)
...
+            } else if (nl_events_fd == 0) {
```
vitalybu: I don't think it's a valid case.. Can be removed and handled by the `else` statement

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -346,6 +373,26 @@ void monitor_thread::after_select(bool timeout)
             }
         }
 
+        // Process nl events
+        if (mon_hal_nl_events) {
```
vitalybu, line 376: `mon_hal_nl_events` is actually a Socket instance, so perhaps `mon_hal_nl_socket` is a better name

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -346,6 +373,26 @@ void monitor_thread::after_select(bool timeout)
...
+            if (read_ready(mon_hal_nl_events)) {
+                if (!mon_wlan_hal->process_nl_events()) {
+                    LOG(ERROR) << "process_nl_events() failed!";
+                    thread_last_error_code = MONITOR_THREAD_ERROR_NL_REPORT_PROCESS_FAIL;
+                    stop_monitor_thread();
+                    return;
+                }
+                LOG(DEBUG) << "proccess nl events finished ";
```
vitalybu: Remove

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -346,6 +373,26 @@ void monitor_thread::after_select(bool timeout)
...
+                clear_ready(mon_hal_nl_events);
+            }
+        } else {
+            // There is no socket for external events
+            LOG(ERROR) << "no socket exists for mon_hal_nl_events";
+            thread_last_error_code = MONITOR_THREAD_ERROR_NL_ATTACH_FAIL;
+            stop_monitor_thread();
+            return;
+        }
```
vitalybu: Can this ever happen?
`stop_monitor_thread()` is called if the FD is not valid..

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        if (notification == nullptr) {
+            LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION failed";
+            return false;
+        }
+
+        auto radio_mac = notification->radio_mac();
+
+        dynamic_channel_selection_task::sScanEvent new_event;
+        new_event.radio_mac = radio_mac;
+        tasks.push_event(database.get_dynamic_channel_selection_task_id(radio_mac),
```
vitalybu: What if there's no DCS task for this MAC?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        if (notification == nullptr) {
+            LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_RESULTS_NOTIFICATION failed";
+            return false;
+        }
+
+        auto radio_mac = notification->radio_mac();
+
+        //send results to dcs task
+        if (database.get_channel_scan_in_progress(radio_mac, false) ||
+            database.get_channel_scan_in_progress(radio_mac, true)) {
```
vitalybu, line 3380: Is it possible to change the API to allow checking both conditions in a single call?

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -147,7 +148,7 @@ bool monitor_thread::socket_disconnected(Socket *sd)
         thread_last_error_code = MONITOR_THREAD_ERROR_HAL_DISCONNECTED;
         stop_monitor_thread();
         return false;
-    }
+     }
```
vitalybu: Remove the extra space

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1153,6 +1154,52 @@ bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
...
+        }
+
+        auto radio_mac         = request->scan_params().radio_mac;
+        auto dwell_time_ms     = request->scan_params().dwell_time_ms;
+        auto channel_pool      = request->scan_params().channel_pool;
+        auto channel_pool_size = request->scan_params().channel_pool_size;
+        auto channel_pool_vector =
+            std::vector<unsigned int>(channel_pool, channel_pool + channel_pool_size);
+        std::string channels;
+        for (int8_t index = 0; index < channel_pool_size; index++) {
```
vitalybu: Use regular `int` instead of `int8_t`
Also, add a newline separation between variable declaration and code segments

vitalybu: This loop is for debug purposes only? If so, please add a relevant comment

controller/src/beerocks/master/son_management.cpp
```diff
@@ -1803,26 +1803,355 @@ void son_management::handle_bml_message(Socket *sd,
...
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_BML_CHANNEL_SCAN_GET_RESULTS_REQUEST>();
+        if (!request) {
+            LOG(ERROR) << "addClass cACTION_BML_CHANNEL_SCAN_GET_RESULTS_REQUEST failed";
+            break;
+        }
+
+        // Get information from request
+        auto radio_mac      = request->radio_mac();
+        auto is_single_scan = request->scan_mode() == 1;
```
vitalybu, line 1966: There's no constant or enum for the scan modes?
Are there any more modes or is it just 0 and 1?

controller/src/beerocks/master/son_management.cpp
```diff
@@ -1803,26 +1803,355 @@ void son_management::handle_bml_message(Socket *sd,
...
+        auto scan_results      = database.get_channel_scan_results(radio_mac, is_single_scan);
+        auto scan_results_size = scan_results.size();
+
+        LOG(DEBUG) << "scan_results received for hostap_mac= " << radio_mac << std::endl
+                   << "scan_results_size= " << scan_results_size;
+
+        //Results avaliablity check
+        if (scan_results_size == 0) {
+            LOG(DEBUG) << "no scan results are available";
+            //op_error_code = eChannelScanOpErrCode::CHANNEL_SCAN_OP_SCAN_RESULTS_EMPTY;
```
vitalybu: Remove unused code

vitalybu (dismissed):
Reviewed till: `controller: bml: DCS channel scan setup`

controller/src/beerocks/master/son_management.cpp
itayx: Moved the whole segment to fix an issue with sending CHANNEL_SCAN_RESULTS_EMPTY

controller/src/beerocks/master/son_management.cpp
itayx: just 0 & 1 since we don't send booleans over CMDU

itayx:
This PR is dependent on hotfix #695

controller/src/beerocks/master/son_management.cpp
tomereli: So define an enum in the yaml and use it instead

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: To log that at least one of the sockets is about to be disconnected
removed the first "if" but added LOG(DEBUG) << "stopping <socket name>" for each individual socket

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: removed

arnout:
[[DCS Patch 4/5]Port dynamic channel selection](https://trello.com/c/pV5WJCd8/78-dcs-patch-4-5port-dynamic-channel-selection)

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Added a print

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: removed

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Removed

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: since `mon_hal_ext_events` uses the same naming convention, I don't think we should refactor the naming here unless we change that one as well

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: @kantera800
What do you think?
`get_nl_events_fd()` description says  it returns -1 on error but has no mention of other cases

controller/src/beerocks/master/son_master_thread.cpp
itayx: I could add another api called `get_any_channel_scan_in_progress(const sMacAddr&)` but I think that would be redundant as this will be the only place it will be used

controller/src/beerocks/master/son_master_thread.cpp
itayx: Added a check and an error log in case it doesn't find one

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: removed

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Changed the `uint8_t` to `int`.
Added a comment explaining this loop is only for the print.
And made the whole loop thinner by using a ternary operator instead of the if statement
`channels += ((index != 0) ? "," : "") + std::to_string(int(channel_pool[index]));`

agent/src/beerocks/monitor/monitor_thread.cpp
tomereli: > We always use `{}` with `if` statements (even one-liners)

You are right, unfortunately, according to the majority :)
https://doodle.com/poll/wwyfaymx84chhe3k

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -135,6 +142,11 @@ bool monitor_thread::socket_disconnected(Socket *sd)
         thread_last_error_code = MONITOR_THREAD_ERROR_HAL_DISCONNECTED;
         stop_monitor_thread();
         return false;
+    } else if (mon_hal_nl_events && (sd == mon_hal_nl_events)) {
+        LOG(ERROR) << "mon_hal_nl_events socket disconnected!";
+        thread_last_error_code = MONITOR_THREAD_ERROR_HAL_DISCONNECTED;
```
tomereli: This is incorrect, should be `thread_last_error_code = MONITOR_THREAD_ERROR_NL_EVENTS_SOCKET_DISCONNECTED`;

agent/src/beerocks/monitor/monitor_thread.cpp
tomereli: You're both right, but I guess we should keep the naming convention change to a separate PR.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2788,6 +2812,105 @@ bool slave_thread::handle_cmdu_monitor_message(Socket *sd,
...
+
+        auto notification_out = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION>(cmdu_tx);
+        if (!notification_out) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION !";
+            return false;
+        }
+
+        notification_out->scan_results() = notification_in->scan_results();
+        notification_out->is_dump()      = notification_in->is_dump();
```
tomereli, line 2877: What is the `is_dump()` parameter? Can it be part of the `scan_results()`?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
     }
+    case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE: {
+        LOG(TRACE) << "ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE for mac " << hostap_mac;
+        break;
+    }
+    case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION: {
+        LOG(TRACE) << "DCS_task, sending SCAN_TRIGGERED for mac " << hostap_mac;
+        auto notification =
+            beerocks_header->addClass<beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION>();
+        if (notification == nullptr) {
```
tomereli: `if (!notification)` to avoid KW issues.
Have you run KW on this PR? Please do so.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        }
+
+        auto radio_mac = notification->radio_mac();
+
+        dynamic_channel_selection_task::sScanEvent new_event;
+        new_event.radio_mac = radio_mac;
+
+        tasks.push_event(database.get_dynamic_channel_selection_task_id(radio_mac),
+                         (int)dynamic_channel_selection_task::eEvent::SCAN_ABORTED,
+                         (void *)&new_event);
```
rmelotte: No C-style cast, use a more appropriate type of cast (probably a static_cast).

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        tasks.push_event(database.get_dynamic_channel_selection_task_id(radio_mac),
+                         (int)dynamic_channel_selection_task::eEvent::SCAN_FINISHED,
+                         (void *)&new_event);
+        break;
+    }
+    case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION: {
+        LOG(DEBUG) << "DCS_task, sending SCAN_ABORTED for mac " << hostap_mac;
+        auto notification =
+            beerocks_header->addClass<beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION>();
+        if (notification == nullptr) {
```
rmelotte: this should be `if (!notification) {`.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        }
+
+        auto radio_mac = notification->radio_mac();
+
+        dynamic_channel_selection_task::sScanEvent new_event;
+        new_event.radio_mac = radio_mac;
+
+        tasks.push_event(database.get_dynamic_channel_selection_task_id(radio_mac),
+                         (int)dynamic_channel_selection_task::eEvent::SCAN_FINISHED,
+                         (void *)&new_event);
```
rmelotte: avoid c-style casts.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+    case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION: {
+        LOG(TRACE) << "DCS_task, sending SCAN_TRIGGERED for mac " << hostap_mac;
+        auto notification =
+            beerocks_header->addClass<beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION>();
+        if (notification == nullptr) {
+            LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION failed";
+            return false;
+        }
+
+        auto radio_mac = notification->radio_mac();
```
tomereli, line 3351: Why do you need an extra variable definition? Just use:
```
dynamic_channel_selection_task::sScanEvent new_event;
new_event.radio_mac = notification->radio_mac();
tasks.push_event(database.get_dynamic_channel_selection_task_id(notification->radio_mac()),
     (int)dynamic_channel_selection_task::eEvent::SCAN_TRIGGERED,
     (void *)&new_event);
```
(nitpick)

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1106,6 +1152,52 @@ bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
...
+            if (index != 0) {
+                channels += ",";
+            }
+            channels += std::to_string(int(channel_pool[index]));
+        }
+
+        //debug print incoming information:
+        LOG(DEBUG) << std::endl
+                   << "scan_params:" << std::endl
+                   << "radio_mac=" << network_utils::mac_to_string(radio_mac) << std::endl
```
rmelotte: No need to use mac_to_string, the << operator should already be defined for `sMacAddr`.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: So at least provide an explanation in the form of a comment before the if condition - when I read this code I don't understand how `false` and `true` differ.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3323,6 +3324,103 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        LOG(DEBUG) << "DCS_task, sending SCAN_ABORTED for mac " << hostap_mac;
+        auto notification =
+            beerocks_header->addClass<beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION>();
+        if (notification == nullptr) {
+            LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION failed";
+            return false;
+        }
+
+        auto radio_mac = notification->radio_mac();
+
```
tomereli: remove extra line break (here and everywhere else)

controller/src/beerocks/master/son_management.cpp
```diff
@@ -1800,6 +1801,34 @@ void son_management::handle_bml_message(Socket *sd,
         son_actions::send_cmdu_to_agent(agent_mac, cmdu_tx, database);
         break;
     }
+    case beerocks_message::ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_REQUEST: {
+        LOG(TRACE) << "ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_REQUEST";
```
tomereli, line 1805: Remove extra space

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1417,13 +1417,50 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
...
     case Event::Channel_Scan_Dump_Result: {
+        auto notification = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION>(cmdu_tx);
+        if (!notification) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION msg";
+            return false;
+        }
+
+        // If event == Channel_Scan_New_Results_Ready do nothing since is_dump's default is 0
+        if (event == Event::Channel_Scan_Dump_Result) {
```
tomereli, line 1477: I think a one-liner is enough here: `notification->is_dump() = (event == Event::Channel_Scan_Dump_Result)` (nitpick)

agent/src/beerocks/monitor/monitor_thread.cpp
tomereli: Also - please use `LOG_IF` instead of `if` when only printing logs based on condition.

controller/src/beerocks/bml/bml.h
```diff
@@ -601,7 +601,7 @@ int bml_get_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int *
  * @return BML_RET_OK on success.
  */
 int bml_get_dcs_scan_results(BML_CTX ctx, const char *radio_mac,
-                             struct BML_DCS_NEIGHBOR_AP **output_results,
+                             struct BML_NEIGHBOR_AP **output_results,
```
tomereli, line 604: I now understand the change - it replaces the results structure as we discussed.
This is a good example where a commit message should provide much more context, so the reviewer can understand why this change was done.

All the commit messages in this PR (and the other PRs) are in the format of "Do this, do that" - this is not in par with the prplMesh contribution guide.

Therefore, please change *all* commit messages in this PR (and future PRs) to abide by the prplMesh standard defined in https://github.com/prplfoundation/prplMesh/blob/master/CONTRIBUTING.md#commits.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -101,6 +102,13 @@ void monitor_thread::stop_monitor_thread()
             mon_hal_int_events = nullptr;
         }
 
+        if (mon_hal_nl_events) {
+            remove_socket(mon_hal_nl_events);
+            delete mon_hal_nl_events;
```
rmelotte, line 109: By looking at the rest of the file I wonder why we have to resort to explicitly calling delete here, though that's not related to this PR.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -27,8 +27,9 @@ INITIALIZE_EASYLOGGINGPP
 ////////////////////////// Local Module Definitions //////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-#define SELECT_TIMEOUT (500)    // 500 milliseconds
-#define RESPONSE_TIMEOUT (5000) // 5 seconds
+#define SELECT_TIMEOUT (500)             // 500 milliseconds
+#define RESPONSE_TIMEOUT (5000)          // 5 seconds
+#define DELAYED_RESPONSE_TIMEOUT (30000) // 30 seconds
```
tomereli: What is this used for? Please add a more elaborate comment (do this for all the timeouts while you're at it).

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1106,6 +1152,52 @@ bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
...
+        auto channel_pool      = request->scan_params().channel_pool;
+        auto channel_pool_size = request->scan_params().channel_pool_size;
+        auto channel_pool_vector =
+            std::vector<unsigned int>(channel_pool, channel_pool + channel_pool_size);
+        std::string channels;
+        for (int8_t index = 0; index < channel_pool_size; index++) {
+            if (index != 0) {
+                channels += ",";
+            }
+            channels += std::to_string(int(channel_pool[index]));
```
rmelotte: Why does it need to be converted to an int before making the std::string ?

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1219,6 +1219,53 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
     return (iRet);
 }
 
+int bml_internal::set_dcs_continuous_scan_enable(const std::string &mac, int enable)
+{
+    //CMDU message
```
tomereli: The comment is meaningless - all are CMDU messages.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1219,6 +1219,53 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+    auto request = message_com::create_vs_message<
+        beerocks_message::cACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_ENABLE_REQUEST>(cmdu_tx);
+
+    if (!request) {
+        LOG(ERROR)
+            << "Failed building cACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_ENABLE_REQUEST message!";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    request->radio_mac() = network_utils::mac_from_string(mac);
```
tomereli: Please, no `mac_from_string` - use sMacAddr in the YAML instead of string unless you have a very good reason not to.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1219,6 +1219,53 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+    request->isEnable()  = enable;
+
+    int result = 0;
+    if (send_bml_cmdu(result, request->get_action_op()) != BML_RET_OK) {
+        LOG(ERROR) << "Send ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_ENABLE_REQUEST failed";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    if (result > 0) {
```
tomereli: Can result be negative?

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1106,6 +1152,52 @@ bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
...
+        }
+
+        auto radio_mac         = request->scan_params().radio_mac;
+        auto dwell_time_ms     = request->scan_params().dwell_time_ms;
+        auto channel_pool      = request->scan_params().channel_pool;
+        auto channel_pool_size = request->scan_params().channel_pool_size;
+        auto channel_pool_vector =
+            std::vector<unsigned int>(channel_pool, channel_pool + channel_pool_size);
+        std::string channels;
+        for (int8_t index = 0; index < channel_pool_size; index++) {
```
rmelotte: `channel_pool_size` is a `uint8_t`, `index` should probably also be a uint8_t.

Unrelated to this PR:
channel_pool_size should also probably be of a size type since it's a size.

controller/src/beerocks/bml/bml.cpp
```diff
@@ -564,8 +564,8 @@ int bml_set_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int e
         return (-BML_RET_INVALID_ARGS);
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    bml_internal *pBML = (bml_internal *)ctx;
```
rmelotte: Don't use c-style casts (see other comments).

controller/src/beerocks/bml/bml.cpp
```diff
@@ -575,8 +575,8 @@ int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int *
...
```
rmelotte: Don't use c-style casts (see other comments).

controller/src/beerocks/bml/bml.cpp
```diff
@@ -588,8 +588,9 @@ int bml_set_dcs_continuous_scan_params(BML_CTX ctx, const char *radio_mac, int d
...
```
rmelotte: Don't use c-style casts (see other comments).

controller/src/beerocks/bml/bml.cpp
```diff
@@ -615,8 +617,9 @@ int bml_get_dcs_scan_results(BML_CTX ctx, const char *radio_mac,
...
```
rmelotte: Don't use c-style casts (see other comments).

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+    auto request = message_com::create_vs_message<
+        beerocks_message::cACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_ENABLE_REQUEST>(cmdu_tx);
+
+    if (!request) {
+        LOG(ERROR)
+            << "Failed building ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_ENABLE_REQUEST message!";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    request->radio_mac() = network_utils::mac_from_string(mac);
```
rmelotte: Don't use string for mac addresses, use sMacAddr. It would avoid the need to convert it back here.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+
+    return send_bml_cmdu(*output_enable, request->get_action_op());
+}
+
+int bml_internal::set_dcs_continuous_scan_params(const std::string &mac, int dwell_time,
```
rmelotte: Don't use a string for `mac` if possible (use sMacAddr).

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+{
+    if (dwell_time == BML_CHANNEL_SCAN_INVALID_PARAM &&
+        interval_time == BML_CHANNEL_SCAN_INVALID_PARAM &&
+        channel_pool_size == BML_CHANNEL_SCAN_INVALID_PARAM) {
+        LOG(ERROR) << "Function is called, but no data is being set!";
+        return (-BML_RET_INVALID_DATA);
+    }
+
+    // If the socket is not valid, attempt to re-establish the connection
+    if (m_sockMaster == nullptr) {
```
rmelotte: `if (! m_sockMaster) {`

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+            << "Failed building cACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_REQUEST message!";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    request->radio_mac()                = network_utils::mac_from_string(mac);
+    request->params().dwell_time_ms     = dwell_time;
+    request->params().interval_time_sec = interval_time;
+    request->params().channel_pool_size = channel_pool_size;
+    if (channel_pool_size > 0 && channel_pool_size <= BML_CHANNEL_SCAN_MAX_CHANNEL_POOL_SIZE &&
+        channel_pool != nullptr) {
```
rmelotte: compare pointers with (! pointer), not `==nullptr`.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+    return BML_RET_OK;
+}
+
+int bml_internal::get_dcs_continuous_scan_params(const std::string &mac, int *output_dwell_time,
+                                                 int *output_interval_time,
+                                                 unsigned int *output_channel_pool,
+                                                 int *output_channel_pool_size)
+{
+    if (output_dwell_time == nullptr && output_interval_time == nullptr &&
+        output_channel_pool_size == nullptr) {
```
rmelotte: compare pointers with (! pointer), not `==nullptr`.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+        LOG(ERROR) << "Function is called, but no data is being requested!";
+        return (-BML_RET_INVALID_DATA);
+    }
+
+    if (m_sockPlatform == nullptr && !connect_to_platform()) {
```
rmelotte: compare pointers with (! pointer), not `==nullptr`.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+    auto request = message_com::create_vs_message<
+        beerocks_message::cACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_REQUEST>(cmdu_tx);
+
+    if (!request) {
+        LOG(ERROR)
+            << "Failed building ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_REQUEST message!";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    request->radio_mac() = network_utils::mac_from_string(mac);
```
rmelotte: mac should already be a sMacAddr so that you don't need to convert it.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+        iRet = -BML_RET_TIMEOUT;
+    }
+
+    // Clear the scan params member
+    m_scan_params = nullptr;
+
+    // Clear the promise holder
+    m_prmChannelScanParamsGet = nullptr;
+
+    if (output_dwell_time != nullptr) {
```
rmelotte: compare pointers with (! pointer), not `==nullptr`.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+        *output_dwell_time = ScanParams.dwell_time_ms;
+    }
+    if (output_interval_time != nullptr) {
```
rmelotte: compare pointers with (! pointer), not `==nullptr`.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+        *output_interval_time = ScanParams.interval_time_sec;
+    }
+    if (output_channel_pool_size != nullptr && output_channel_pool != nullptr) {
```
rmelotte: compare pointers with (! pointer), not `==nullptr`.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1206,6 +1251,309 @@ int bml_internal::bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps
...
+
+    return (iRet);
+}
+
+int bml_internal::get_dcs_scan_results(const std::string &mac, BML_NEIGHBOR_AP **output_results,
+                                       unsigned int *output_results_size,
+                                       const unsigned int max_results_size,
+                                       uint8_t *output_result_status, bool is_single_scan)
+{
+    if (m_sockPlatform == nullptr && !connect_to_platform()) {
```
rmelotte: compare pointers with (! pointer), not `==nullptr`.

controller/src/beerocks/bml/internal/bml_internal.h
```diff
@@ -152,6 +152,9 @@ class bml_internal : public beerocks::socket_thread {
     int bml_set_vap_list_credentials(const BML_VAP_INFO *vaps, const uint8_t vaps_num);
     int bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps_num);
 
+    //set and get channel scan enable flag
+    int set_dcs_continuous_scan_enable(const std::string &mac, int enable);
+    int get_dcs_continuous_scan_enable(const std::string &mac, int *output_enable);
```
tomereli: Please change to `int get_dcs_continuous_scan_enable(const std::string &mac, int *enable);` and add doxygen marking `int *enable` as `[OUT]` in the doxygen comment.

controller/src/beerocks/bml/internal/bml_internal.h
```diff
@@ -152,6 +152,25 @@ class bml_internal : public beerocks::socket_thread {
...
+
+    //set and get channel scan params
+    int set_dcs_continuous_scan_params(const std::string &mac, int dwell_time, int interval_time,
```
rmelotte: Don't use std::string for mac, use sMacAddr.

controller/src/beerocks/bml/internal/bml_internal.h
```diff
@@ -152,6 +152,25 @@ class bml_internal : public beerocks::socket_thread {
...
+                                       unsigned int *channel_pool, int channel_pool_size);
+    int get_dcs_continuous_scan_params(const std::string &mac, int *output_dwell_time,
```
rmelotte: Don't use std::string for mac, use sMacAddr.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+{
+    if (dwell_time == BML_CHANNEL_SCAN_INVALID_PARAM &&
+        interval_time == BML_CHANNEL_SCAN_INVALID_PARAM &&
+        channel_pool_size == BML_CHANNEL_SCAN_INVALID_PARAM) {
+        LOG(ERROR) << "Function is called, but no data is being set!";
+        return (-BML_RET_INVALID_DATA);
+    }
+
+    // If the socket is not valid, attempt to re-establish the connection
+    if (m_sockMaster == nullptr) {
```
tomereli: `if (!m_sockMaster)`

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+        int iRet = connect_to_master();
+        if (iRet != BML_RET_OK) {
+            return iRet;
+        }
+    }
+
+    //CMDU message
```
tomereli: meaningless comment

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+            << "Failed building cACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_REQUEST message!";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    request->radio_mac()                = network_utils::mac_from_string(mac);
+    request->params().dwell_time_ms     = dwell_time;
+    request->params().interval_time_sec = interval_time;
+    request->params().channel_pool_size = channel_pool_size;
+    if (channel_pool_size > 0 && channel_pool_size <= BML_CHANNEL_SCAN_MAX_CHANNEL_POOL_SIZE &&
+        channel_pool != nullptr) {
```
tomereli: `!channel_pool`

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+        std::copy_n(channel_pool, channel_pool_size, request->params().channel_pool);
+    }
+
+    int result = 0;
+    if (send_bml_cmdu(result, request->get_action_op()) != BML_RET_OK) {
+        LOG(ERROR) << "Send ACTION_BML_CHANNEL_SCAN_SET_CONTINUOUS_PARAMS_REQUEST failed";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    if (result > 0) {
```
tomereli: Again - can result be negative? what is considered as success? shouldn't this be simple boolean?

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+    auto request = message_com::create_vs_message<
+        beerocks_message::cACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_REQUEST>(cmdu_tx);
+
+    if (!request) {
+        LOG(ERROR)
+            << "Failed building ACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_REQUEST message!";
+        return (-BML_RET_OP_FAILED);
+    }
+
+    request->radio_mac() = network_utils::mac_from_string(mac);
```
tomereli: no mac_from_string please

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+        iRet = -BML_RET_TIMEOUT;
+    }
+
+    // Clear the scan params member
+    m_scan_params = nullptr;
+
+    // Clear the promise holder
+    m_prmChannelScanParamsGet = nullptr;
+
+    if (output_dwell_time != nullptr) {
```
tomereli: `if (output_dwell_time)`

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+        *output_dwell_time = ScanParams.dwell_time_ms;
+    }
+    if (output_interval_time != nullptr) {
```
tomereli: same here

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+        *output_interval_time = ScanParams.interval_time_sec;
+    }
+    if (output_channel_pool_size != nullptr && output_channel_pool != nullptr) {
```
tomereli: same here

rmelotte (commented):
Did a quick code review, lot of the comments are the same. Note that I only blindly reviewed the _code_ itself, not the overall changes, because I don't know enough about it to review it. For this reason I'm only adding this as a comment, not approving nor requesting changes (please take the comments into account though).

Most were:
- There were many uses of `ptr == nullptr` to compare pointers, you should use `! ptr` directly. There were a lot of these, I didn't add a comment for all of them.

- Please also remove all C style-casts and use a more appropriate one (probably static_cast in all cases).

- Also don't use std::string for mac addresses, use sMacAddr instead.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1401,6 +1401,44 @@ int bml_internal::get_dcs_continuous_scan_params(const std::string &mac, int *ou
     return (iRet);
 }
 
+int bml_internal::start_dcs_single_scan(const std::string &mac, int dwell_time_ms,
+                                        int channel_pool_size, unsigned int *channel_pool)
+{
+    LOG(DEBUG) << "start_single_channel_scan";
+    //CMDU message
```
tomereli: remove comment

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -64,6 +64,38 @@ static void config_logger(const std::string log_file = std::string())
 
 #endif // BEEROCKS_DEBUG
 
+static void translate_channel_scan_results(const beerocks_message::sChannelScanResults &res_in,
```
tomereli, line 70: Why not change to return `bool` if it can fail?

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1401,6 +1433,89 @@ int bml_internal::get_dcs_continuous_scan_params(const std::string &mac, int *ou
     return (iRet);
 }
 
+int bml_internal::get_dcs_scan_results(const std::string &mac, BML_NEIGHBOR_AP **output_results,
+                                       unsigned int *output_results_size,
+                                       const unsigned int max_results_size,
+                                       uint8_t *output_result_status, bool is_single_scan)
+{
+    if (m_sockPlatform == nullptr && !connect_to_platform()) {
```
tomereli: Also, please separate to 2 separate if conditions.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1401,6 +1433,89 @@ int bml_internal::get_dcs_continuous_scan_params(const std::string &mac, int *ou
...
+    // Initialize the promise for receiving the response
+    beerocks::promise<int> prmChannelScanResultsGet;
+    m_prmChannelScanResultsGet    = &prmChannelScanResultsGet;
+    int iOpTimeout                = DELAYED_RESPONSE_TIMEOUT; // Default timeout
+    auto scan_results             = std::list<beerocks_message::sChannelScanResults>();
+    uint32_t scan_results_size    = 0;
+    uint32_t scan_results_maxsize = uint32_t(max_results_size);
+
+    m_scan_results         = &scan_results;
+    m_scan_results_size    = &scan_results_size;
```
tomereli: I don't understand why we need both `m_scan_results_size` and `m_scan_results_maxsize`

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1401,6 +1433,89 @@ int bml_internal::get_dcs_continuous_scan_params(const std::string &mac, int *ou
...
+    m_scan_results_maxsize = &scan_results_maxsize;
+    m_scan_results_status  = output_result_status;
+
+    //CMDU message
```
tomereli: remove comment

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1401,6 +1433,89 @@ int bml_internal::get_dcs_continuous_scan_params(const std::string &mac, int *ou
...
+    iRet = prmChannelScanResultsGet.get_value();
+    if (iRet != int(eChannelScanOpErrCode::CHANNEL_SCAN_OP_SUCCESS)) {
+        LOG(ERROR) << "Results returned with error code:" << iRet << ". Aborting!";
+        return iRet;
+    }
+
+    //output_results_size will be set to the number of actual returning results
+    *output_results_size = 0;
+    for (auto res : scan_results) {
+        auto out = &((*output_results)[*output_results_size]);
```
tomereli: `auto &`

controller/src/beerocks/bml/internal/bml_internal.h
```diff
@@ -240,15 +246,19 @@ class bml_internal : public beerocks::socket_thread {
...
+    beerocks_message::sDeviceInfo *m_device_info                     = nullptr;
+    beerocks_message::sWifiCredentials *m_wifi_credentials           = nullptr;
+    beerocks_message::sAdminCredentials *m_admin_credentials         = nullptr;
+    beerocks_message::sVersions *m_master_slave_versions             = nullptr;
+    beerocks_message::sRestrictedChannels *m_Restricted_channels     = nullptr;
+    beerocks_message::sChannelScanRequestParams *m_scan_params       = nullptr;
+    std::list<beerocks_message::sChannelScanResults> *m_scan_results = nullptr;
+    uint8_t *m_scan_results_status                                   = nullptr;
+    uint32_t *m_scan_results_size                                    = nullptr;
+    uint32_t *m_scan_results_maxsize                                 = nullptr;
```
tomereli: Please add a comment explaining each member.

tomereli (dismissed):

agent/src/beerocks/monitor/monitor_thread.cpp
kantera800: Let's remove it

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: This is a preparation for the 5th PR
When `event == Event::Channel_Scan_Dump_Result` not only do we set the `is_dump() = 1`
but also copy the data received with the event to the notification

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Removed & moved the `mon_hal_nl_events = nullptr` to the (now only) `else` statement

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Changed

agent/src/beerocks/slave/son_slave_thread.cpp
itayx: No, `is_dump()` is a flag that is only used in the "Get Scan Results" messages
while `scan_results()` is a structure called `sChannelScanResults` which is used throughout the code

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Changed it to `LOG_IF`

controller/src/beerocks/master/son_master_thread.cpp
itayx: Fixed

controller/src/beerocks/master/son_master_thread.cpp
itayx: Fixed

controller/src/beerocks/master/son_master_thread.cpp
itayx: Fixed

controller/src/beerocks/master/son_master_thread.cpp
itayx: Added a comment explaining it

controller/src/beerocks/master/son_master_thread.cpp
itayx: Changed

controller/src/beerocks/master/son_management.cpp
itayx: It was removed in the next commit

itayx: It was removed in the next commit

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Changed

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Fixed

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Changed the for loop to
`for (int index = 0; index < int(channel_pool_size); index++) {`
that way it is not dependent on what type is the `channel_pool_size`

And regarding the `channel_pool_size`, it needs to be a basic type for the CMDUs

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: They are used as timeouts for promises set on responses to requests
and are mostly used like the following: `<promise>.wait_for(<timeout>)`

I've added a comment explaining this

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: bml_internal is called by the bml which is called by either the bml cli or by C libraries
so the only place I could use the `network_utils::mac_from_string()` would be within the bml_internal

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Fixed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Fixed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: All it does is copy so it's can't really fail unless it fails to allocate memory

controller/src/beerocks/master/son_management.cpp
tomereli: nitpick anyway..

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Fixed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Removed all `//CMDU message`

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Removed all `//CMDU message`

controller/src/beerocks/bml/internal/bml_internal.h
itayx: Added

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Removed all `//CMDU message`

agent/src/beerocks/monitor/monitor_thread.cpp
kantera800: in case platform dosnt support netlink (dummy mode for example) we shouldn't call stop_monitor_thread() in cased FD is invalid.
@itayx  didn't you add a fixup that removes it ?

itayx: yeah, I commented saying I did remove it

kantera800: was not clear :)

controller/src/beerocks/bml/bml.cpp
itayx: Changed

controller/src/beerocks/bml/bml.cpp
itayx: Changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: No, changed it to `!= int(eChannelScanOpErrCode::CHANNEL_SCAN_OP_SUCCESS)`

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: The feature must use the mac_from_string at least once since the bml gets the mac address as a string buffer (char[]) in the bml functions
So I moved the mac_from_string to the bml

controller/src/beerocks/bml/bml.cpp
itayx: Changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: [See](https://github.com/prplfoundation/prplMesh/pull/694/commits/3104d970c25b24adcf45be9e062c5d8b5a68a96e#r369162085)

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: removed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed to check the result against the corresponding enum

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: [See](https://github.com/prplfoundation/prplMesh/pull/694/commits/3104d970c25b24adcf45be9e062c5d8b5a68a96e#r369162085)

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: [See](https://github.com/prplfoundation/prplMesh/pull/694/commits/3104d970c25b24adcf45be9e062c5d8b5a68a96e#r369162085)

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: changed

controller/src/beerocks/bml/internal/bml_internal.h
itayx: [See](https://github.com/prplfoundation/prplMesh/pull/694/commits/3104d970c25b24adcf45be9e062c5d8b5a68a96e#r369162085)

controller/src/beerocks/bml/internal/bml_internal.h
itayx: [See](https://github.com/prplfoundation/prplMesh/pull/694/commits/3104d970c25b24adcf45be9e062c5d8b5a68a96e#r369162085)

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Fixed

tomereli:
Channel selection test fails with this pr

tomereli:
Please fix the channel selection test, then rebase squash and we'll do the followup review.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -346,6 +373,24 @@ void monitor_thread::after_select(bool timeout)
...
+            if (read_ready(mon_hal_nl_events)) {
+                if (!mon_wlan_hal->process_nl_events()) {
+                    LOG(ERROR) << "process_nl_events() failed!";
+                    thread_last_error_code = MONITOR_THREAD_ERROR_NL_REPORT_PROCESS_FAIL;
+                    stop_monitor_thread();
+                    return;
+                }
+                clear_ready(mon_hal_nl_events);
+            }
+        } else {
```
kantera800: @itay please  remove this  else completely.
We are entering here each after_select  no need for this print at all.

controller/src/beerocks/bml/internal/bml_internal.h
vitalybu: Why not reference?

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1266,6 +1266,141 @@ int bml_internal::get_dcs_continuous_scan_enable(const std::string &mac, int *ou
...
+int bml_internal::set_dcs_continuous_scan_params(const std::string &mac, int dwell_time,
+                                                 int interval_time, unsigned int *channel_pool,
+                                                 int channel_pool_size)
+{
+    if (dwell_time == BML_CHANNEL_SCAN_INVALID_PARAM &&
+        interval_time == BML_CHANNEL_SCAN_INVALID_PARAM &&
+        channel_pool_size == BML_CHANNEL_SCAN_INVALID_PARAM) {
+        LOG(ERROR) << "Function is called, but no data is being set!";
+        return (-BML_RET_INVALID_DATA);
+    }
```
vitalybu, line 1444: What if only one of the parameters has invalid value?
Shouldn't this statement use OR conditions?

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -64,6 +64,38 @@ static void config_logger(const std::string log_file = std::string())
...
 
+static void translate_channel_scan_results(const beerocks_message::sChannelScanResults &res_in,
+                                           BML_NEIGHBOR_AP *res_out)
+{
+    if (!res_out) {
+        LOG(ERROR) << "Failed translating results, res_out==NULL";
+        return;
+    }
+
+    string_utils::copy_string(res_out->ap_SSID, res_in.ssid, 36);
```
vitalybu: Use a constant for the SSID size

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: No, this handles the case if no parameters is being requested by the caller
If all params are invalid it means that there is no need to continue with the function
yet, even if one is valid  it will still continue but return only that value

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: removed

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Changed

controller/src/beerocks/bml/internal/bml_internal.cpp
tomereli: OK

controller/src/beerocks/bml/bml.h
tomereli: ?

controller/src/beerocks/bml/bml.cpp
```diff
@@ -564,8 +564,9 @@ int bml_set_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int e
...
     }
 
-    // TODO: call suitable bml api
-    return BML_RET_OP_FAILED;
+    auto pBML = static_cast<bml_internal *>(ctx);
+    return pBML->set_dcs_continuous_scan_enable(
+        network_utils::mac_from_string(std::string(radio_mac)), enable);
 }
 
 int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int *output_enable)
```
tomereli: Change to `int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int *enable)`

Or even better as Vitaly suggested - why not use reference?

controller/src/beerocks/bml/internal/bml_internal.h
```diff
@@ -152,6 +152,9 @@ class bml_internal : public beerocks::socket_thread {
     int bml_set_vap_list_credentials(const BML_VAP_INFO *vaps, const uint8_t vaps_num);
     int bml_get_vap_list_credentials(BML_VAP_INFO *vaps, uint8_t &vaps_num);
 
+    //set and get channel scan enable flag
+    int set_dcs_continuous_scan_enable(const sMacAddr &mac, int enable);
+    int get_dcs_continuous_scan_enable(const sMacAddr &mac, int *enable);
```
tomereli: I prefer `int bml_get_dcs_continuous_scan_enable(BML_CTX ctx, const char *radio_mac, int &enable)`

controller/src/beerocks/bml/internal/bml_internal.cpp
tomereli: Any reason this conversation is unresolved?

controller/src/beerocks/bml/internal/bml_internal.cpp
tomereli: I meant to pass the output result as a reference.

Change the declaration of `translate_channel_scan_results` to accept `BML_NEIGHBOR_AP &res_out` instead of `BML_NEIGHBOR_AP *res_out` and set `auto &out = (*output_results)[*output_results_size];`

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -67,6 +67,39 @@ static void config_logger(const std::string log_file = std::string())
 
 #endif // BEEROCKS_DEBUG
 
+static void translate_channel_scan_results(const beerocks_message::sChannelScanResults &res_in,
```
tomereli, line 70: Change to `static void translate_channel_scan_results(const beerocks_message::sChannelScanResults &res_in, BML_NEIGHBOR_AP &res_out)`

Please always use reference whenever possible.

tomereli (changes requested):
Only small comments only but since the commit messages where not changed yet I'm not approving.
Please fix the new comments, update commit messages appropriately, rebase + squash and ask for a final re-review.

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: @kantera800
I'm not sure myself if this is needed

controller/src/beerocks/bml/bml.cpp
itayx: In the BML I can't use a reference, since I'm not sure how well it will be imported in C
I can change the `bml_internal` to use a reference instead

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Oh.. my bad
Fixed and added a comment explaining the whole `(*output_results)[*output_results_size]` deal so it would be more understandable

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Forgot to resolve the conversation

agent/src/beerocks/monitor/monitor_thread.cpp
kantera800: from looking at the code remove_socket just isn't strong enough

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Changed

controller/src/beerocks/bml/internal/bml_internal.h
itayx: the BML cannot use a reference since it's used as a C API
but the bml_internal uses reference

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: Changed

tomereli (approved):

vitalybu (approved):

PR #695, opened 2020-01-20
Hotfix/channel scan no results

itayx:
The CHANNEL_SCAN_RESULTS_EMPTY needs to be moved
from the error enum to the status enum as sending an empty results list
is not an error but a valid status of the current scan response.


tomereli (approved):

ghost (approved):

rmelotte:
The commit message looks wrong. It says it's "moving an element from the error enum to the status enums", but both enums contain "CHANNEL_SCAN*_SUCCESS" so neither of them look like an error enum.
The message should be updated to say why it belong more to one enum than the other.

I don't understand what the 2 separate enums are used for, so I can't comment on the relevance of the change.

Note: the PR has been merged while I was redacting, so the commit message cannot be updated anymore.

PR #696, opened 2020-01-20
Hotfix/fix start monitor with empty ip

morantr:
Currently, when a client connects, an association handling task is triggered
on topology notification. The task may reach to a point when it needs to
send START_MONITORING_REQUEST, before the client ipv4 has been updated
(from DHCP notification), and send the request with previous and not updated IP.
This causes a bug in the measurements, since the arp to the client
which used to measure the client will be sent to the wrong IP and the
measurement will fail.

Discovered by QA on RDKB, when the client connected and the statistics polling task was disabled,
a measurement was supposed to be performed by the START_MONITORING_REQUEST, but the RSSI in the CLI remained "-127"

### DIFF
--- a/controller/src/beerocks/master/son_actions.cpp
+++ b/controller/src/beerocks/master/son_actions.cpp
@@ -312,6 +312,10 @@ void son_actions::handle_dead_node(std::string mac, std::string hostap_mac, db &
     if (parent_hostap_mac == hostap_mac) {
         if (mac_type == beerocks::TYPE_IRE_BACKHAUL || mac_type == beerocks::TYPE_CLIENT) {
             database.set_node_state(mac, beerocks::STATE_DISCONNECTED);
+
+            // Clear node ipv4
+            database.set_node_ipv4(mac);
+
             // Notify steering task, if any, of disconnect.
             int steering_task = database.get_steering_task_id(mac);
             if (tasks.is_task_running(steering_task))
--- a/controller/src/beerocks/master/tasks/association_handling_task.cpp
+++ b/controller/src/beerocks/master/tasks/association_handling_task.cpp
@@ -41,8 +41,13 @@ void association_handling_task::work()
 {
     switch (state) {
     case START: {
+        // If this task already has been created by another event, let it finish and finish the new
+        // instance of it.
         int prev_task_id = database.get_association_handling_task_id(sta_mac);
-        tasks.kill_task(prev_task_id);
+        if (tasks.is_task_running(prev_task_id)) {
+            finish();
+            return;
+        }
         database.assign_association_handling_task_id(sta_mac, id);
 
         original_parent_mac = database.get_node_parent(sta_mac);
--- a/controller/src/beerocks/master/tasks/association_handling_task.cpp
+++ b/controller/src/beerocks/master/tasks/association_handling_task.cpp
@@ -28,6 +28,7 @@ using namespace son;
 #define BEACON_MEASURE_REQ_TIME_SPAN 3000
 #define BEACON_MEASURE_MAX_ATTEMPTS 3
 #define REQUEST_RSSI_MEASUREMENT_DELAY 5000
+#define GET_CLIENT_IP_TIMEOUT_SEC 3
 
 association_handling_task::association_handling_task(db &database_,
                                                      ieee1905_1::CmduMessageTx &cmdu_tx_,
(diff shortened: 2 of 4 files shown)


tomereli:
On which platform does it happen? Please explain how this was found in the PR description

morantr:
> On which platform does it happen? Please explain how this was found in the PR description

Updated the description.

controller/src/beerocks/master/son_actions.cpp
```diff
@@ -312,6 +312,10 @@ void son_actions::handle_dead_node(std::string mac, std::string hostap_mac, db &
     if (parent_hostap_mac == hostap_mac) {
         if (mac_type == beerocks::TYPE_IRE_BACKHAUL || mac_type == beerocks::TYPE_CLIENT) {
             database.set_node_state(mac, beerocks::STATE_DISCONNECTED);
+
+            // Clear node ipv4
+            database.set_node_ipv4(mac);
```
adam1985d, line 317: nitpick: wouldn't a call to an API  "clear_node_ipv4" be more readable to what is performed here?

adam1985d (approved):
good PR - nice work

controller/src/beerocks/master/son_actions.cpp
morantr: maybe, I just didn't want to add an additional function.

morantr:
> Wouldn't it make more sense to do this check-in the callers that create the task, and remove it here? I believe I have seen that pattern for some other task.

No, it will create code duplication, and also I don't think it is a good idea that the task creator should deal with it when it can be managed internally.

arnout (approved):
Excellent PR with good commit splitting and messages. Two remarks are more about checking that I understand things correctly than things that need to be fixed.

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -28,6 +28,7 @@ using namespace son;
 #define BEACON_MEASURE_REQ_TIME_SPAN 3000
 #define BEACON_MEASURE_MAX_ATTEMPTS 3
 #define REQUEST_RSSI_MEASUREMENT_DELAY 5000
+#define GET_IP_CLIENT_TIMEOUT_SEC 3
```
tomereli: `#define GET_CLIENT_IP_TIMEOUT_SEC` is more appropriate IMO

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -70,18 +71,39 @@ void association_handling_task::work()
...
+            new_hostap_mac = database.get_node_parent(sta_mac);
             if (new_hostap_mac != original_parent_mac ||
                 database.get_node_state(sta_mac) != beerocks::STATE_CONNECTED) {
                 TASK_LOG(DEBUG) << "sta " << sta_mac << " is no longer connected to "
                                 << original_parent_mac << " finishing task";
                 finish();
                 return;
             }
-            std::string ipv4 = database.get_node_ipv4(sta_mac);
+
```
tomereli, line 90: I think that it will be much more readable like this:

```
bool send_start_monitoring_request = false;
...
if (!ipv4.empty()) {
    send_start_monitoring_request = true;
}

if (send_start_monitoring_request) {
    ...
} else {
    // handle timeouts
}
```

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -70,18 +71,39 @@ void association_handling_task::work()
...
+            // Wait until ipv4 will be received on DHCP event or platform arp monitor notification
+            if (ipv4.empty()) {
+                if (std::chrono::steady_clock::now() >
+                    (task_started_timestamp + std::chrono::seconds(GET_IP_CLIENT_TIMEOUT_SEC))) {
+                    TASK_LOG(ERROR) << "ipv4 hasn't been received on client " << sta_mac
+                                    << ", skip start monitoring";
+                    send_start_monitoring_request = false;
+                } else {
+                    break;
```
tomereli, line 102: If you still decide to keep the current implementation (I hope you won't) - please at least add a comment here that explains the break is causing the state machine to stay in the same state.

tomereli (approved):
I don't want to be a party pooper so approving, but please fix my comment in the last commit. I really think it will be much more readable this way.

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: changed

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: I disagree with that one since IMO, we should first handle errors and then the rest of the code. This makes the code more readable.

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: add a comment

tomereli, line 102: Still no comment

tomereli (approved):

Issue #665, opened 2020-01-13
[TASK] DCS cleanups


tomereli:
# Cleanups
> Fix all remaining review comments which were marked as such which can be cleaned up later.
> Each review comment should be tagged in this issue.
### common/beerocks/bwl/CMakeLists.txt:62
- `list(APPEND BWL_LIBS dwpal nl-3 nl-genl-3)`
- Perhaps it's better to create a FindNL cmake file and use the standard cmake find_package(). Here's an example: https://github.com/nasa/channel-emulator/blob/master/cmake/Modules/FindLibNL.cmake
- Fixed in #680 
### common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:311
- ` if (!m_dwpal_nl_ctx) {`
- The dwpal netlink context is created in the Init state.
- On failure, we transition the FSM to the Detach state.
- In what scenario do we get to the Attach state without a valid context? this statement is redundant.
- RAX40 has a much older driver version at the moment so the attach will fail
- We mustn't fail the FSM on that! Just log an warning and proceed without NL. Similar to what we do when we lack permissions to open raw socket - we simply don't open it.
- Still, this should be removed
### Remove / Alter redundant logs
- common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp:467
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:594
- common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp:284
### Small patches
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:311 This should be removed
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:593 `if (!vendor_data) { ... }`
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:616 `if (!out_buffer) { ... }`
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:609 Function should return ssize_t (signed size_t).
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:626 Either initialize data_size to -1, or simply return -1 here to indicate a failure.
### Remove USE_LIBSAFEC from other files
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h:23
### No additional comment
- common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:675
[...]

PR #699, opened 2020-01-21
Feature/add bml cli commands to dynamic channel selection bml apis

adam1985d:
This PR adds CLI BML commands to the beerocks_cli for the following DCS BML APIS:
bml_set_dcs_continuous_scan_enable
bml_get_dcs_continuous_scan_enable
bml_set_dcs_continuous_scan_params
bml_get_dcs_continuous_scan_params
bml_start_dcs_single_scan
bml_get_dcs_scan_results

The commands naming are aligned to the called APIs without the 'bml_' initial:
set_dcs_continuous_scan_enable
get_dcs_continuous_scan_enable
set_dcs_continuous_scan_params
get_dcs_continuous_scan_params
start_dcs_single_scan
get_dcs_scan_results

The command's description is available in the beerocks_cli help menu

### DIFF
--- a/controller/src/beerocks/cli/beerocks_cli_bml.cpp
+++ b/controller/src/beerocks/cli/beerocks_cli_bml.cpp
@@ -507,6 +507,42 @@ void cli_bml::setFunctionsMapAndArray()
                        static_cast<pFunction>(&cli_bml::bml_rdkb_steering_client_measure_caller), 3,
                        3, INT_ARG, STRING_ARG, STRING_ARG);
 #endif
+    insertCommandToMap(
+        "bml_set_dcs_continuous_scan_enable", "<mac> <1 or 0>",
+        "Enable/Disable (1 or 0) the Dynamic-Channel-Selection for the given AP mac.",
+        static_cast<pFunction>(&cli_bml::set_dcs_continuous_scan_enable_caller), 2, 2, STRING_ARG,
+        INT_ARG);
+    insertCommandToMap("bml_get_dcs_continuous_scan_enable", "<mac>",
+                       "Get Dynamic-Channel-Selection enable configuration for the given AP mac.",
+                       static_cast<pFunction>(&cli_bml::get_dcs_continuous_scan_enable_caller), 1,
+                       1, STRING_ARG);
+    insertCommandToMap("bml_set_dcs_continuous_scan_params", "<mac> [<params>]",
+                       "Set the continuous scan params for the given AP mac: 'params' = "
+                       "(dwell_time, interval_time, channel_pool"
+                       ")=value. Params description: dwell_time - dwell time in "
+                       "milliseconds, interval_time - interval time in seconds,"
+                       " channel_pool - channels separated by commas.",
+                       static_cast<pFunction>(&cli_bml::set_dcs_continuous_scan_params_caller), 2,
+                       4, STRING_ARG, STRING_ARG, STRING_ARG, STRING_ARG);
+    insertCommandToMap("bml_get_dcs_continuous_scan_params", "<mac>",
+                       "Get Dynamic-Channel-Selection params for the given AP mac: dwell_time - "
+                       "dwell time in milliseconds, interval_time - interval time in seconds,"
+                       " channel_pool - channels seperated by commas.",
+                       static_cast<pFunction>(&cli_bml::get_dcs_continuous_scan_params_caller), 1,
+                       1, STRING_ARG);
+    insertCommandToMap(
+        "bml_start_dcs_single_scan", "<mac> <dwell_time_msec> <channel_pool>",
+        "Start a single scan, for the given AP mac, with the following scan params:"
+        " dwell_time - dwell time in milliseconds, channel_pool - channels seperated by commas.",
+        static_cast<pFunction>(&cli_bml::start_dcs_single_scan_caller), 3, 3, STRING_ARG, INT_ARG,
+        STRING_ARG);
+    insertCommandToMap(
+        "bml_get_dcs_scan_results", "<mac> <max-results-size> [<is-single-scan>]",
+        "Get Dynamic-Channel-Selection scan results for the given AP mac:"
+        " max-results-size - maximal size of the returned results,"
+        " is-single-scan - 0 for continuous-scan results (default), 1 for single scan.",
+        static_cast<pFunction>(&cli_bml::get_dcs_scan_results_caller), 2, 3, STRING_ARG, INT_ARG,
+        INT_ARG);
     //bool insertCommandToMap(std::string command, std::string help_args, std::string help,  pFunction funcPtr, uint8_t minNumOfArgs, uint8_t maxNumOfArgs,
 }
 
@@ -1212,7 +1248,7 @@ int cli_bml::get_dcs_continuous_scan_params_caller(int numOfArgs)
  */
 int cli_bml::start_dcs_single_scan_caller(int numOfArgs)
 {
-    if (numOfArgs == 4) {
+    if (numOfArgs == 3) {
         return start_dcs_single_scan(args.stringArgs[0], args.intArgs[1], args.stringArgs[2]);
     }
     return -1;
--- a/controller/src/beerocks/cli/beerocks_cli_bml.cpp
+++ b/controller/src/beerocks/cli/beerocks_cli_bml.cpp
@@ -1123,6 +1123,119 @@ int cli_bml::bml_rdkb_steering_client_measure_caller(int numOfArgs)
 }
 #endif //BEEROCKS_RDKB
 
+/**
+ * caller function for set_dcs_continuous_scan_enable
+ *
+ * @param [in] numOfArgs Num of received arguments
+ *
+ * @return 0 on success.
+ */
+int cli_bml::set_dcs_continuous_scan_enable_caller(int numOfArgs)
+{
+    if (numOfArgs == 2) {
+        return set_dcs_continuous_scan_enable(args.stringArgs[0], args.intArgs[1]);
+    }
+    return -1;
+}
+
+/**
+ * caller function for get_dcs_continuous_scan_enable
+ *
+ * @param [in] numOfArgs Num of received arguments
+ *
+ * @return 0 on success.
+ */
+int cli_bml::get_dcs_continuous_scan_enable_caller(int numOfArgs)
+{
+    if (numOfArgs == 1) {
+        return get_dcs_continuous_scan_enable(args.stringArgs[0]);
+    }
+    return -1;
+}
+
+/**
+ * caller function for set_dcs_continuous_scan_params
+ *
+ * @param [in] numOfArgs Num of received arguments
+ *
+ * @return 0 on success.
+ */
+int cli_bml::set_dcs_continuous_scan_params_caller(int numOfArgs)
+{
+    std::string radio_mac(network_utils::WILD_MAC_STRING);
+    int32_t dwell_time    = BML_CHANNEL_SCAN_INVALID_PARAM;
+    int32_t interval_time = BML_CHANNEL_SCAN_INVALID_PARAM;
+    std::string channel_pool;
+
+    std::string::size_type pos;
+    //[radio_mac=<radio_mac> dwell_time=<dwell_time> interval_time='<interval_time>']"
+    for (int i = 1; i < numOfArgs; i++) { //first optional arg
+        if ((pos = args.stringArgs[i].find("radio_mac=")) != std::string::npos) {
+            radio_mac = args.stringArgs[i].substr(pos + sizeof("radio_mac"));
+        } else if ((pos = args.stringArgs[i].find("dwell_time=")) != std::string::npos) {
+            dwell_time = string_utils::stoi(args.stringArgs[i].substr(pos + sizeof("dwell_time")));
+        } else if ((pos = args.stringArgs[i].find("interval_time=")) != std::string::npos) {
+            interval_time =
+                string_utils::stoi(args.stringArgs[i].substr(pos + sizeof("interval_time")));
+        } else if ((pos = args.stringArgs[i].find("channel_pool=")) != std::string::npos) {
+            channel_pool = args.stringArgs[i].substr(pos + sizeof("channel_pool"));
+        }
+    }
+
+    if (numOfArgs > 1) {
+        return set_dcs_continuous_scan_params(radio_mac, dwell_time, interval_time, channel_pool);
+    }
+    return -1;
+}
+
+/**
+ * caller function for get_dcs_continuous_scan_params
+ *
+ * @param [in] numOfArgs Num of received arguments
+ *
+ * @return 0 on success.
+ */
+int cli_bml::get_dcs_continuous_scan_params_caller(int numOfArgs)
+{
+    if (numOfArgs == 1) {
+        return get_dcs_continuous_scan_params(args.stringArgs[0]);
+    }
+    return -1;
+}
+
+/**
+ * caller function for start_dcs_single_scan
+ *
+ * @param [in] numOfArgs Num of received arguments
+ *
+ * @return 0 on success.
+ */
+int cli_bml::start_dcs_single_scan_caller(int numOfArgs)
+{
+    if (numOfArgs == 4) {
+        return start_dcs_single_scan(args.stringArgs[0], args.intArgs[1], args.stringArgs[2]);
+    }
+    return -1;
+}
+
+/**
+ * caller function for get_dcs_scan_results
+ *
+ * @param [in] numOfArgs Num of received arguments
+ *
+ * @return 0 on success.
+ */
+int cli_bml::get_dcs_scan_results_caller(int numOfArgs)
+{
+    if (numOfArgs == 2) {
+        return get_dcs_scan_results(args.stringArgs[0], args.intArgs[1]);
+    } else if (numOfArgs == 3) {
+        bool single_scan = (args.intArgs[2] == 1);
+        return get_dcs_scan_results(args.stringArgs[0], args.intArgs[1], single_scan);
+    }
+    return -1;
+}
+
 //
 // Functions
 //
--- a/controller/src/beerocks/cli/beerocks_cli_bml.cpp
+++ b/controller/src/beerocks/cli/beerocks_cli_bml.cpp
@@ -1661,3 +1661,102 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
     return 0;
 }
 #endif //BEEROCKS_RDKB
+
+/**
+ * Enables or disables beerocks DCS continuous scans.
+ *
+ * @param [in] radio_mac Radio MAC of selected radio
+ * @param [in] enable Value of 1 to enable or 0 to disable.
+ *
+ * @return 0 on success.
+ */
+int cli_bml::set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t enable)
+{
+    std::cout << "set_dcs_continuous_scan_enable" << std::endl;
+
+    return 0;
+}
+
+/**
+ * get DCS continuous scans param. Value is printed to the console.
+ *
+ * @param [in] radio_mac Radio MAC of selected radio
+ *
+ * @return 0 on success.
+ */
+int cli_bml::get_dcs_continuous_scan_enable(const std::string &radio_mac)
+{
+    std::cout << __func__ << ", mac=" << radio_mac << std::endl;
+
+    return 0;
+}
+
+/**
+ * set DCS continuous scan params.
+ *
+ * @param [in] radio_mac Radio MAC of selected radio
+ * @param [in] dwell_time Set the dwell time in milliseconds.
+ * @param [in] interval_time Set the interval time in seconds.
+ * @param [in] channel_pool Set the channel pool for the DCS.
+ *
+ * @return 0 on success.
+ */
+int cli_bml::set_dcs_continuous_scan_params(const std::string &radio_mac, int32_t dwell_time,
+                                            int32_t interval_time, const std::string &channel_pool)
+{
+    std::cout << __func__ << ", mac=" << radio_mac << ", dwell_time=" << dwell_time
+              << ", interval_time=" << interval_time << ", channel_pool=" << channel_pool
+              << std::endl;
+
+    return 0;
+}
+
+/**
+ * get DCS continuous scan params. Values are printed to the console.
+ *
+ * @param [in] radio_mac Radio MAC of selected radio
+ *
+ * @return 0 on success.
+ */
+int cli_bml::get_dcs_continuous_scan_params(const std::string &radio_mac)
+{
+    std::cout << __func__ << ", mac=" << radio_mac << std::endl;
+
+    return 0;
+}
+
+/**
+ * Start a single DCS scan with parameters.
+ *
+ * @param [in] radio_mac radio MAC of selected radio
+ * @param [in] dwell_time Set the dwell time in milliseconds.
+ * @param [in] channel_pool Set the channel pool for the DCS.
+ *
+ * @return 0 on success.
+ */
+int cli_bml::start_dcs_single_scan(const std::string &radio_mac, int32_t dwell_time,
+                                   const std::string &channel_pool)
+{
+    std::cout << "start_dcs_single_scan, mac=" << radio_mac << ", dwell_time=" << dwell_time
+              << ", channel_pool=" << channel_pool << std::endl;
+
+    return 0;
+}
+
+/**
+ * get DCS continuous scan params.
+ *
+ * @param [in] radio_mac radio MAC of selected radio
+ * @param [in] max_results_size Max number of the returned results
+ * @param [in] is_single_scan Flag indicating if the results belong to a single scan or not
+ * 
+ * @return 0 on success.
+ */
+int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_results_size,
+                                  bool is_single_scan)
+{
+    std::cout << __func__ << ", mac=" << radio_mac << ", max_results_size=" << max_results_size
+              << ", is_single_scan=" << string_utils::bool_str(is_single_scan) << std::endl;
+
+    return 0;
+}
--- a/controller/src/beerocks/cli/beerocks_cli_bml.cpp
+++ b/controller/src/beerocks/cli/beerocks_cli_bml.cpp
@@ -24,6 +24,17 @@ using namespace net;
 /////////////////////////// Local Module Functions ///////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
+static const std::string string_from_int_array(void *arr, size_t arr_max_size)
+{
+    std::stringstream ss;
+    if (arr) {
+        for (size_t i = 0; i < arr_max_size; i++) {
+            ss << ((i != 0) ? ", " : "") << int(static_cast<int *>(arr)[i]);
+        }
+    }
+    return ss.str();
+}
+
 static void fill_conn_map_node(
     std::unordered_multimap<std::string, std::shared_ptr<cli_bml::conn_map_node_t>> &conn_map_nodes,
     struct BML_NODE *node)
@@ -1217,7 +1228,6 @@ int cli_bml::set_dcs_continuous_scan_params_caller(int numOfArgs)
             channel_pool = args.stringArgs[i].substr(pos + sizeof("channel_pool"));
         }
     }
-
     if (numOfArgs > 1) {
         return set_dcs_continuous_scan_params(radio_mac, dwell_time, interval_time, channel_pool);
     }
@@ -1821,7 +1831,11 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
  */
 int cli_bml::set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t enable)
 {
-    std::cout << "set_dcs_continuous_scan_enable" << std::endl;
+    std::cout << __func__ << ", mac=" << radio_mac << ", enable=" << enable << std::endl;
+
+    int ret = bml_set_dcs_continuous_scan_enable(ctx, radio_mac.c_str(), enable);
+
+    printBmlReturnVals("bml_set_dcs_continuous_scan_enable", ret);
 
     return 0;
 }
@@ -1836,6 +1850,15 @@ int cli_bml::set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t
 int cli_bml::get_dcs_continuous_scan_enable(const std::string &radio_mac)
 {
     std::cout << __func__ << ", mac=" << radio_mac << std::endl;
+
+    int enable = -1;
+    int ret    = bml_get_dcs_continuous_scan_enable(ctx, radio_mac.c_str(), &enable);
+
+    if (ret == BML_RET_OK) {
+        std::cout << "dcs-continuous-scan-enable=" << string_utils::bool_str(enable) << std::endl;
+    }
+
+    printBmlReturnVals("bml_get_dcs_continuous_scan_enable", ret);
 
     return 0;
 }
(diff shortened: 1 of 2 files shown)


adam1985d:
[add cli to all DCS BMLs](https://trello.com/c/GrioG4Rj/80-add-cli-to-all-dcs-bmls)

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1661,49 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
     return 0;
 }
 #endif //BEEROCKS_RDKB
+
+int cli_bml::set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t enable)
```
morantr, line 1832: why the `enable` is `int8` and not `bool`?

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1661,49 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
...
+
+int cli_bml::get_dcs_continuous_scan_params(const std::string &radio_mac)
+{
+    std::cout << "get_dcs_continuous_scan_params" << std::endl;
+
+    return 0;
+}
+
+int cli_bml::start_dcs_single_scan(const std::string &radio_mac, int dwell_time,
+                                   int channel_pool_size, const std::string &channel_pool)
```
morantr: I prefer using types that imply the number of bits that are used: uint32_t/int32_t.

apply this on all new stubs.

controller/src/beerocks/cli/beerocks_cli_bml.h
```diff
@@ -190,6 +190,18 @@ class cli_bml : public cli {
     int steering_client_measure(uint32_t steeringGroupIndex, const std::string &str_bssid,
                                 const std::string &str_client_mac);
 #endif
+    int set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t enable);
+    int get_dcs_continuous_scan_enable(const std::string &radio_mac);
+    int set_dcs_continuous_scan_params(const std::string &radio_mac,
+                                       int dwell_time                  = BML_DCS_INVALID_PARAM,
+                                       int interval_time               = BML_DCS_INVALID_PARAM,
+                                       const std::string &channel_pool = std::string(),
+                                       int channel_pool_size           = BML_DCS_INVALID_PARAM);
```
morantr: I don't know think it is a good idea to define too many optional parameters, especially when they share common types.
I think a better idea will be to define a struct with default parameters and give it an argument.
(nitpick)

morantr: There is no reason to add here `channel_pool_size` since it is a user interface. The function should automatically figure how many channels there are in the user input.

controller/src/beerocks/cli/beerocks_cli_bml.h
```diff
@@ -190,6 +190,18 @@ class cli_bml : public cli {
     int steering_client_measure(uint32_t steeringGroupIndex, const std::string &str_bssid,
                                 const std::string &str_client_mac);
 #endif
+    int set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t enable);
+    int get_dcs_continuous_scan_enable(const std::string &radio_mac);
+    int set_dcs_continuous_scan_params(const std::string &radio_mac,
+                                       int dwell_time                  = BML_DCS_INVALID_PARAM,
+                                       int interval_time               = BML_DCS_INVALID_PARAM,
+                                       const std::string &channel_pool = std::string(),
```
morantr: Why are you passing `channel_pool` as a string and not a vector or array? The caller function should parse it and convert it to vector/array.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1661,49 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
...
+{
+    std::cout << "start_dcs_single_scan" << std::endl;
+
+    return 0;
+}
+
+int cli_bml::get_dcs_scan_results(const std::string &radio_mac, unsigned int max_results_size,
+                                  int8_t is_single_scan)
```
morantr: why the` is_single_scan` is int8 and not bool?

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -507,6 +507,46 @@ void cli_bml::setFunctionsMapAndArray()
...
+        "Enable/Disable (1 or 0) the Dynamic-Channel-Selection for the given AP mac.",
+        static_cast<pFunction>(&cli_bml::set_dcs_continuous_scan_enable_caller), 2, 2, STRING_ARG,
+        INT_ARG);
+    insertCommandToMap("bml_get_dcs_continuous_scan_enable", "<mac>",
+                       "Get Dynamic-Channel-Selection enable configuration for the given AP mac.",
+                       static_cast<pFunction>(&cli_bml::get_dcs_continuous_scan_enable_caller), 1,
+                       1, STRING_ARG);
+    insertCommandToMap(
+        "bml_set_dcs_continuous_scan_params",
+        "<mac> <dwell_time_msec> [<interval_time>] [<channel_pool>] [<channel_pool_size>]",
```
morantr: You can define it like this only if you set all the optional arguments altogether. This is not the case here, so what you need to do is to explicitly write to what argument do you mean:
`<mac> <dwell_time_msec> [ interval_time=<interval_time>] [channel_pool=<channel_pool>]`

on the caller function you will need to parse it, and here convert all the argument types to `STRING_ARG`.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1661,49 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
     return 0;
 }
 #endif //BEEROCKS_RDKB
+
+int cli_bml::set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t enable)
+{
+    std::cout << "set_dcs_continuous_scan_enable" << std::endl;
```
tomereli: Probably better to also print the radio mac and the enable.
Can we use `bool enable` instead of `int8_t`?

Same comments for the rest of the APIs

controller/src/beerocks/cli/beerocks_cli_bml.h
```diff
@@ -190,6 +190,18 @@ class cli_bml : public cli {
     int steering_client_measure(uint32_t steeringGroupIndex, const std::string &str_bssid,
                                 const std::string &str_client_mac);
 #endif
+    int set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t enable);
```
tomereli, line 199: Please add doxygen to all new APIs either here or in the implementation file.
A rule of thumb - every new function comes with doxygen comments.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1123,6 +1123,67 @@ int cli_bml::bml_rdkb_steering_client_measure_caller(int numOfArgs)
...
+}
+
+int cli_bml::set_dcs_continuous_scan_params_caller(int numOfArgs)
+{
+    if (numOfArgs == 2) {
+        return set_dcs_continuous_scan_params(args.stringArgs[0], args.intArgs[1]);
+    } else if (numOfArgs == 3) {
+        return set_dcs_continuous_scan_params(args.stringArgs[0], args.intArgs[1], args.intArgs[2]);
+    } else if (numOfArgs == 4) {
+        std::cout << "Invalid number of arguments, set of channel_pool must also provide "
```
tomereli: Why don't we print such an error in other failure cases?

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1806,5 +1913,63 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, unsigned int max
 {
     std::cout << "get_dcs_scan_results" << std::endl;
 
+    if (max_results_size == 0) {
+        std::cout << __func__ << "invalid input, max_results_size==0" << std::endl;
+        return -1;
+    }
+
+    struct BML_DCS_NEIGHBOR_AP *results = nullptr;
```
tomereli: No need for `struct` keyword - its CPP.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1806,5 +1913,63 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, unsigned int max
...
+
+    results =
+        (struct BML_DCS_NEIGHBOR_AP *)calloc(max_results_size, sizeof(struct BML_DCS_NEIGHBOR_AP));
+    if (results == nullptr) {
```
tomereli: Did you run KW on this? please refrain from using `if (ptr == nullptr)` and instead use `if (!ptr)` to avoid KW issues.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1806,5 +1913,63 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, unsigned int max
...
+    bool single_scan           = (is_single_scan == 1);
+    int ret = bml_get_dcs_scan_results(ctx, radio_mac.c_str(), &results, &results_count, &status,
+                                       single_scan);
+
+    if (ret == BML_RET_OK) {
+        if (results_count > max_results_size) {
+            std::cout << __func__ << "ERROR: results_count(" << results_count << ")"
+                      << " > max_results_size(" << max_results_size << ")" << std::endl;
+        } else if (status == 0 && results_count > 0) {
+            //todo: replace the temporary debug print with actual prints of the new results structure
```
tomereli: I'm not sure I follow - what is temporary with the debug print? Aren't the results already in the new structure?

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1806,5 +1913,63 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, unsigned int max
...
 
+    if (max_results_size == 0) {
+        std::cout << __func__ << "invalid input, max_results_size==0" << std::endl;
+        return -1;
+    }
+
+    struct BML_DCS_NEIGHBOR_AP *results = nullptr;
+
+    results =
+        (struct BML_DCS_NEIGHBOR_AP *)calloc(max_results_size, sizeof(struct BML_DCS_NEIGHBOR_AP));
```
tomereli: There is absolutely no good reason to use `calloc` or `alloc` in c++, and especially in this function where you simply print the returned results and free them right after.
Simply declare the results on the stack:

`BML_DCS_NEIGHBOR_AP results[max_results_size];`

Also - why is the `max_results_size` argument even needed? This can also be statically declared in the function.

tomereli (changes requested):

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1767,13 +1767,27 @@ int cli_bml::set_dcs_continuous_scan_enable(const std::string &radio_mac, int8_t
...
 
 int cli_bml::get_dcs_continuous_scan_enable(const std::string &radio_mac)
 {
     std::cout << "get_dcs_continuous_scan_enable" << std::endl;
 
+    int enable = -1;
+    int ret    = bml_get_dcs_continuous_scan_enable(ctx, radio_mac.c_str(), &enable);
+
+    if (ret == BML_RET_OK) {
+        std::cout << "dcs-continuous-scan-enable=" << ((enable == 1) ? "True" : "False")
```
morantr: string_utils::bool_str(enable);

controller/src/beerocks/cli/beerocks_cli_bml.cpp
morantr: Why didn't you declare it on the stack???
Anyway, this is C allocation. The CPP way is with `new`/`delete` and the better option will be to use smart pointers.

morantr (changes requested):

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: this is a special case where 2 params must be provided together or omitted together

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: 1. bool instead of uint8_t:
- the BML API uses int - and the CML CLI provides an int - so I don't see a value in int->bool->int conversion
- in the 'get_dcs_scan_results()' I will move the int->bool translation to the caller function and update the signature
2. print the input values:
done

controller/src/beerocks/cli/beerocks_cli_bml.h
adam1985d: I added inside the CPP file (not to clatter the class header)

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: removed the struct keyword

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: no - since the PR that changes the structure is not available yet
(PRs 4/5): 
https://github.com/prplfoundation/prplMesh/pull/694

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: the max_results_size is to simulate the max-results-size limit also used in <wave-api> in RDKB (currently set as 200)
when DCS supports ordering of the results by RSSI or another ranking, it is possible to the get only the X best results even if more are available

I replaced the <calloc> with a stack declaration.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: the BML API uses int - and the CML CLI provides an int - so I don't see a value in int->bool->int conversion

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: done

controller/src/beerocks/cli/beerocks_cli_bml.h
adam1985d: might be, but will not be changed in this PR

adam1985d: discussed this with @tomereli  and he agrees with you. will be removed

controller/src/beerocks/cli/beerocks_cli_bml.h
adam1985d: none of the caller functions in the BML CLI perform any parsing. they are all a wrapper for getting the information from the CLI command to the cli_bml command.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: was changed to bool

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: ok - replaced the implementation to what you suggested (used the beerocks_cli_socket::client_beacon_11k_req as reference)

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: wasn't familiar with it - very useful - fixed, thanks

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: @tomereli already mentioned it - I fixed it and am using a user-defined-sized array declared on the stack.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: the code was removed as part of refactoring due to @morantr comment

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: removed the <calloc> entirely - and will run KW

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -507,6 +507,46 @@ void cli_bml::setFunctionsMapAndArray()
...
+    insertCommandToMap("bml_get_dcs_continuous_scan_enable", "<mac>",
+                       "Get Dynamic-Channel-Selection enable configuration for the given AP mac.",
+                       static_cast<pFunction>(&cli_bml::get_dcs_continuous_scan_enable_caller), 1,
+                       1, STRING_ARG);
+    insertCommandToMap(
+        "bml_set_dcs_continuous_scan_params",
+        "<mac> <dwell_time_msec> [<interval_time>] [<channel_pool>] [<channel_pool_size>]",
+        "Set the continuous scan params for the given AP mac: dwell_time - dwell time in "
+        "miliseconds, interval_time - interval time in seconds,"
+        " channel_pool - channels seperated by commas (set of channel_pool must also provide "
```
mariomaz: typo: should be "separated"

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -507,6 +507,46 @@ void cli_bml::setFunctionsMapAndArray()
...
+        INT_ARG);
+    insertCommandToMap("bml_get_dcs_continuous_scan_enable", "<mac>",
+                       "Get Dynamic-Channel-Selection enable configuration for the given AP mac.",
+                       static_cast<pFunction>(&cli_bml::get_dcs_continuous_scan_enable_caller), 1,
+                       1, STRING_ARG);
+    insertCommandToMap(
+        "bml_set_dcs_continuous_scan_params",
+        "<mac> <dwell_time_msec> [<interval_time>] [<channel_pool>] [<channel_pool_size>]",
+        "Set the continuous scan params for the given AP mac: dwell_time - dwell time in "
+        "miliseconds, interval_time - interval time in seconds,"
```
mariomaz: typo: "milliseconds"
This typo is repeated in other commands

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -507,6 +507,46 @@ void cli_bml::setFunctionsMapAndArray()
...
+        INT_ARG, INT_ARG, STRING_ARG, INT_ARG);
+    insertCommandToMap("bml_get_dcs_continuous_scan_params", "<mac>",
+                       "Get Dynamic-Channel-Selection params for the given AP mac: dwell_time - "
+                       "dwell time in miliseconds, interval_time - interval time in seconds,"
+                       " channel_pool - channels seperated by commas, channel_pool_size - size of "
+                       "the channel_pool.",
+                       static_cast<pFunction>(&cli_bml::get_dcs_continuous_scan_params_caller), 1,
+                       1, STRING_ARG);
+    insertCommandToMap(
+        "bml_start_dcs_single_scan", "<mac> <dwell_time_msec> <channel_pool_size> <channel_pool>",
```
mariomaz: For the sake of consistency, channel_pool_size and channel_pool should be given in the same order in all commands.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
mariomaz: IMO, provided APIs should be always as clear as possible and thus enable should be bool in this case

controller/src/beerocks/cli/beerocks_cli_bml.h
tomereli: I agree with Moran here - we should do the conversion in the BML CLI and get rid of the size of channel pool.

controller/src/beerocks/cli/beerocks_cli_bml.h
tomereli: Agree - the channel_pool should be a vector, you can parse it in the caller.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1762,214 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
...
+                                                 channel_pool_arr, channel_pool_size);
+    }
+    printBmlReturnVals("bml_set_dcs_continuous_scan_params", ret);
+
+    return 0;
+}
+
+int cli_bml::get_dcs_continuous_scan_params(const std::string &radio_mac)
+{
+    std::cout << "get_dcs_continuous_scan_params" << std::endl;
```
mariomaz: You should probably be using __func__ instead of a literal string to avoid copy & paste errors. This way maintenance is easier (if method name changes, there is no need to change anything else). 
This is something to be considered in all the methods in this file.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1762,214 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
...
+
+    int dwell_time = -1, interval_time = -1, channel_pool_size = BML_DCS_MAX_CHANNEL_POOL_SIZE;
+    unsigned int channel_pool[BML_DCS_MAX_CHANNEL_POOL_SIZE] = {0};
+
+    int ret = bml_get_dcs_continuous_scan_params(ctx, radio_mac.c_str(), &dwell_time,
+                                                 &interval_time, channel_pool, &channel_pool_size);
+
+    if (ret == BML_RET_OK) {
+        std::cout << "dcs-continuous-scan-params for mac:" << std::endl
```
mariomaz: MAC address is not being printed

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1762,214 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
...
+    return 0;
+}
+
+int cli_bml::start_dcs_single_scan(const std::string &radio_mac, int dwell_time,
+                                   int channel_pool_size, const std::string &channel_pool)
+{
+    std::cout << "start_dcs_single_scan" << std::endl;
+
+    if (dwell_time <= 0) {
+        std::cout << __func__ << "invalid input: dwell_time(" << dwell_time << ") <= 0"
```
mariomaz: Method name is being printed together with next literal string

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1661,3 +1762,214 @@ int cli_bml::steering_client_measure(uint32_t steeringGroupIndex, const std::str
...
+
+    auto channels      = string_utils::str_split(channel_pool, ',');
+    auto channels_size = channels.size();
+
+    if (channels_size > BML_DCS_MAX_CHANNEL_POOL_SIZE) {
+        std::cout << "size of channel_pool is too big. size=" << channels_size << std::endl;
+        return -1;
+    }
+
+    if (channels_size > 0 && channels_size != (size_t)channel_pool_size) {
```
mariomaz: IMO, parentheses in conditionals, even if they are not strictly required, help writing code that is clear and correct

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1123,6 +1163,67 @@ int cli_bml::bml_rdkb_steering_client_measure_caller(int numOfArgs)
...
+{
+    if (numOfArgs == 2) {
+        return set_dcs_continuous_scan_enable(args.stringArgs[0], args.intArgs[1]);
+    }
+    return -1;
+}
+
+int cli_bml::get_dcs_continuous_scan_enable_caller(int numOfArgs)
+{
+    if (numOfArgs == 1) {
```
mariomaz, line 1197: I've been using Yoda conditions for a long time now that these lines make me feel uncomfortable. 
Some static code analyzers would also complain.

mariomaz (approved):

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: fixed - thanks

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: fixed - thanks

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: I felt the same itch - but this was aligned to the signature of DCS BML APIS.
we decided to remove the channel_pool_size completely (and parse it from the string of channel_pool) - so will be fixed

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: fixed

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: fixed

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: fixed

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: this is not maintained across the code - but if you find it more readable, I don't mind adding it

fixed.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: I am familiar with this approach - but this is not maintained in the prpl.

if @tomereli and @arnout  want to add this requirement to prplmesh guidelines it is OK, but a short search of the following regex. <[a-z,A-Z] == [0-9]> produced 375 results in 93 files

will not be handled in this PR, but we can set a task for this change across the code if decided

controller/src/beerocks/cli/beerocks_cli_bml.cpp
tomereli: OK

controller/src/beerocks/cli/beerocks_cli_bml.cpp
tomereli: The best would be to change the DCS WLAN APIs as well - is it feasible?

controller/src/beerocks/cli/beerocks_cli_bml.cpp
tomereli: Fix this, we will.
image

controller/src/beerocks/cli/beerocks_cli_bml.cpp
mariomaz: This was not something to be addressed in this PR but to be discussed if code writing rules are ever revisited.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: it might be - you can create a JIRA to check this at a later time - shouldn't block this PR

tomereli (approved):

controller/src/beerocks/cli/beerocks_cli_bml.h
itayx: Removed `channel_pool_size` throughout

controller/src/beerocks/cli/beerocks_cli_bml.h
itayx: Removed `channel_pool_size` throughout

adam1985d: fixed - removed the channel pool size from user input params. calculated from channel_pool input.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
adam1985d: update - removed the channel_pool_size from user input

morantr (approved):

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -24,6 +24,17 @@ using namespace net;
 /////////////////////////// Local Module Functions ///////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
+static const std::string string_from_int_array(void *arr, size_t arr_max_size)
```
tomereli, line 27: The name of this function implies it creates a string from an array of integers. So it should be 
```
static const std::string string_from_int_array(int arr[], size_t len);
```
Also - please add a doxygen comment explaining this function.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -2021,28 +2032,40 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_res
...
-                          << "  mode=" << res.ap_Mode << std::endl
-                          << "  channel=" << (int)res.ap_Channel << std::endl
-                          << "  signal_strength=" << (int)res.ap_SignalStrength << std::endl
-                          << "  security_mode_enabled=" << res.ap_SecurityModeEnabled << std::endl
-                          << "  encryption_mode=" << res.ap_EncryptionMode << std::endl
+                          << "  mode=" << int(res.ap_Mode) << std::endl
+                          << "  channel=" << int(res.ap_Channel) << std::endl
+                          << "  signal_strength=" << int(res.ap_SignalStrength) << std::endl
+                          << "  security_mode_enabled="
+                          << string_from_int_array(res.ap_SecurityModeEnabled,
```
tomereli, line 2044: I'm think that security mode should be a single value - I'm dazzled why you chose array. Please make it a single value unless there's a very good reason why you need an array. Then we can get rid of this weird `string_from_int_array` function.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -2021,28 +2032,40 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_res
...
+                                                   BML_CHANNEL_SCAN_ENUM_LIST_SIZE)
+                          << std::endl
+                          << "  encryption_mode="
+                          << string_from_int_array(res.ap_EncryptionMode,
```
tomereli, line 2048: Same here - there is a single encryption mode set at a given time. I don't understand why it is an array.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -2021,28 +2032,40 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_res
             std::cout << __func__ << "ERROR: results_count(" << results_count << ")"
                       << " > max_results_size(" << max_results_size << ")" << std::endl;
         } else if ((status == 0) && (results_count > 0)) {
-            //todo: replace the temporary debug print with actual prints of the new results structure
             for (size_t i = 0; i < results_count; i++) {
                 auto res = results[i];
                 std::cout << "result[" << i << "]:" << std::endl
```
tomereli, line 2037: Instead of this whole big print, why not add `inline std::ostream &operator<<(std::ostream &os, const BML_NEIGHBOR_AP &neigh);` to `struct BML_NEIGHBOR_AP` in `bml_defs.h` then you can replace this with:
```
for (size_t i = 0; i < results_count; i++) {
    std::cout << "result[" << i << "]:" << std::endl << results[i] << std::endl;
}
```

tomereli (approved):
I added some more comments, I think you should fix them but if this is urgent it can be done as part of #665 , so approving again.

tomereli:
Finally, please rebase and squash before merge.

adam1985d:
fixed all code review comments and passed all tests

PR #706, opened 2020-01-21
controller: fix encrypted_settings length bug

### LINKED ISSUE #681: [BUG] Controller certification tests fail on master
19/23 tests failed on master (`556f54dc1feb5d7b7ba85f924b71436fbbd08393`).

The logs are available here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/403667226 .

The first noticable thing is that beerocks_agent.log contains a lot of these:
```
ERROR 16:39:14:753 <139724723128064> tlvAssociatedClients.cpp[225] --> TLVF: Requested index is greater than the number of available entries
```

tomereli:
Commit a35edf81532bc5d081e1ad26bc41719b482aa26d converted the son master
thread to use configData class based on AttrList for creating the config
data in the M2 message.

This commit introduced a bug which caused the plaintextlen to be larger
than the actual size of the config data contents, causing a huge
encrypted settings of over 1000 bytes to be add to the CMDU.

Due to this , the CMDU gets fragmented. As we know from agent
tests, the testbed devices don't necessarily work correctly with
fragmented CMDUs, which explains the tests failure.

Fix the bug by using the getMessageLength() instead of the wrong
getMessageBufLength() for getting the config data actual length.

Fixes #681

### DIFF
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -453,8 +453,7 @@ bool master_thread::autoconfig_wsc_add_m2_encrypted_settings(WSC::m2::config &m2
     // attribute type, 2 bytes attribute length, 8 bytes data), but check it anyway
     // to be on the safe side. Then, we add keywrapauth at its end.
     uint8_t *plaintext = config_data.getMessageBuff();
-    int plaintextlen =
-        config_data.getMessageBuffLength() + sizeof(WSC::sWscAttrKeyWrapAuthenticator);
+    int plaintextlen   = config_data.getMessageLength() + sizeof(WSC::sWscAttrKeyWrapAuthenticator);
     WSC::sWscAttrKeyWrapAuthenticator *keywrapauth =
         reinterpret_cast<WSC::sWscAttrKeyWrapAuthenticator *>(
             &plaintext[config_data.getMessageLength()]);


rmelotte (approved):
I started the controller certification tests job with these changes: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/410101110
It looks like it solves the main issue we had.
The CI job will prevent you from merging it (you'd have to wait for all the certification tests to finish) so if you want to merge it as soon as you have the 2 required review, you can use the "cancel" button on the job page.
I'll anyway re-run the tests separately here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/410166616

arnout (approved):
Good catch.

However, it means that plaintextlen was larger than the buffer. How come there was no error due to this buffer overflow? I thought tlv factory was fully protected against buffer overflows now?

tomereli:
> Good catch.
> 
> However, it means that plaintextlen was larger than the buffer. How come there was no error due to this buffer overflow? I thought tlv factory was fully protected against buffer overflows now?

Not related to tlvf, we get the buffer pointer to the config_data created by the TLVF and then we do a buffer overflow in aes_encrypt. Actually, there is no buffer overflow really since we define the ciphertext array with the huge and wrong size of the plaintext. Then reading from out of bound memory is not a buffer overflow..

PR #708, opened 2020-01-21
.gitlab-ci.yml: split certification job between manual and scheduled

rmelotte:
I initially thought Gitlab schedules would trigger manual jobs, but it turns out it's not the case, hence this change.
Marking as dont-merge because this time I want to wait for the scheduled pipelines to actually trigger the job (next schedule is at 8PM CET).

--- 
The certification job was using "when: manual" to make sure it doesn't
run on every pipeline, and to make it possible to run it manually.

It was also intended to be run with Gitlab schedules. However, it
turns out the manual jobs are not run automatically when a pipeline is
started through Gitlab schedules.

Split run-certification-tests between a job that run only on
schedules (so that it actually gets run when a scheduled pipeline is
triggered) and another job which can only be run manually (to keep the
possibility of running the certification tests at any time through the
web interface).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -135,7 +135,7 @@ test-on-netgear-rax40:
     TARGET_DEVICE_NAME: netgear-rax40-1
   needs: ["build-for-netgear-rax40"]
 
-run-certification-tests:
+.run-certification-tests:
   stage: test
   variables:
     TESTS_TO_RUN: "tools/docker/tests/certification/all_controller_tests.txt"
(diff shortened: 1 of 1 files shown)


rmelotte:
Ready to review/merge.
Nighlty pipeline was here: https://gitlab.com/prpl-foundation/prplMesh/pipelines/111254813 .

tomereli (approved):

arnout (approved):
Good call.

Is it possible to remove the default value of TESTS_TO_RUN from the template, so that you're forced to give it when triggered manually? That would be a way to make sure that the manual invocation is not done for all tests unless that was really intended.

rmelotte:
@arnout It would make sense yes. If we do that the manual job will either fail, or run the intended tests.
I think variables can also be specified from the schedules page, so the variable could also be removed for the scheduled job.
I'll do that and add a comment with the variable name to use.

PR #711, opened 2020-01-21
Fix optimal_path_dummy

arnout:
optimal_path_dummy is broken on master, but this isn't caught by CI because it doesn't properly return error.

This PR is meant to fix it. It also fixes additional problems introduced by #431 (in the assumption that that one gets merged first).


ghost:
@arnout #431 was merged. Does that mean that this one can be merged as well?

arnout:
We'll remove the optimal_path_dummy test instead.

A new task #1086 was created to revive this test.

PR #715, opened 2020-01-22
Feature/ci alpine based prplmesh builder

rmelotte:
We need another prplmesh-builder image to run some of the static analysis tools.

This PR adds an alpine-based image, because:
- it uses another libc, which adds variety in the builds we do.
- new package version are usually released pretty quickly after an upstream release (and we need recent versions of cppcheck for instance).

Note that no static analysis is added in this PR, only the alpine-based image and a new CI job to build with it.


tools/docker/builder/alpine/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    ncurses-dev \
+    ninja \
+    openssl-dev \
+    pkgconf \
+    py2-pip \
+    python2 \
+    readline-dev \
+    && apk add --update --no-cache --repository=http://dl-cdn.alpinelinux.org/alpine/edge/main \
+    clang-extra-tools
+# cland-extra-tools needs to be fetched from the edge repository for now
```
tomereli: typo cland --> clang

tomereli (approved):
Why do we need our own builder docker image then? Isn't it redundant now?

tools/docker/builder/alpine/Dockerfile
rmelotte: Fixed.

rmelotte:
@tomereli In a sense yes, we don't strictly need the 2. However it's useful to check that it can be built on other distributions as well (especially Alpine since it uses a different libc).
The main goal of this PR is to add the image's Dockerfile, not the build job. I added the build job because it's for "free" now that we have the image.

.gitlab-ci.yml
```diff
@@ -28,11 +28,14 @@ build-in-docker:
     # Check that the AutoGenerated files are correct by checking git status
     - git update-index -q --refresh
     - git diff-index --exit-code HEAD
-
   artifacts:
     paths:
       - build/install
 
+build-in-docker-alpine:
```
arnout, line 35: Naming of these things is a bit weird... But OK for now.

.gitlab-ci.yml
```diff
@@ -28,11 +28,14 @@ build-in-docker:
...
+  extends: build-in-docker
+  image: prplfoundationinc/prplmesh-builder:alpine-3.11.3
```
arnout, line 37: It would be nice to build the docker image as part of the pipeline as well. But that's again something we can add later.

arnout (approved):
Would have been nice if the second commit included the explanation you gave to @tomereli as to why we want to do this.

But I'm going to merge now, it's not important enough to bother.

PR #722, opened 2020-01-22
Update CONTRIBUTING.md

CoralMalachi:

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -124,3 +124,10 @@ For very long sentences, subsentences may also be placed on a separate line.
 
 Non-inline blocks (lists, preformatted, ...) must be explicitly on a separate paragraph.
 I.e., leave a blank line before and after the block.
+
+### Klocwork 
+
+We use **Klocwork** as our static code analyzer.
+One of the most common issues that the Klocwork complaints about is comparing the return value of a function to nullptr instead of checking for non-zero return value.
+
+So to avoid creating more of these kinds of Klocwork issues, please always prefer `if (!ptr)` instead of `if (ptr == nullptr)` when the return type of the function is a smart pointer.


ghost (changes requested):
Please actually take the time to explain what Klocwork is, rather than just saying "to avoid creating KW issues". New contributors reading this file won't know what you mean by "KW".
BTW, we should probably also add that to our glossary page on the wiki if it's not on there already

CONTRIBUTING.md
```diff
@@ -40,6 +40,12 @@ Code should adhere to the [coding style](#coding-style).
 Note that everybody follows these same guidelines.
 Core contributors, occasional contributors and maintainers are all alike.
 
+### Warnings 
+
+We use **Klocwork** as our static code analyzer. One of the most common issues that the Klocwork complaints about is comparing the return value of a function to nullptr instead of checking for non-zero return value.
```
ghost:
```suggestion
We use **Klocwork** as our static code analyzer. One of the most common issues that the Klocwork complains about is comparing the return value of a function to nullptr instead of checking for non-zero return value.
```
(nitpick)

CONTRIBUTING.md
```diff
@@ -40,6 +40,12 @@ Code should adhere to the [coding style](#coding-style).
...
+
+So to avoid creating more of this klocwork issue, please always prefer `if (!ptr)` instead of `if(ptr == nullptr)`.
```
ghost:
```suggestion
So to avoid creating more of these kinds of Klocwork issues, please always prefer `if (!ptr)` instead of `if(ptr == nullptr)`.
```
(nitpick)

ghost (approved):

ghost:
> Please actually take the time to explain what Klocwork is, rather than just saying "to avoid creating KW issues". New contributors reading this file won't know what you mean by "KW".
> BTW, we should probably also add that to our glossary page on the wiki if it's not on there already

Added to the wiki: https://github.com/prplfoundation/prplMesh/wiki/Glossary#project-specific-terminology

CONTRIBUTING.md
arnout: Another nitpick: our coding style says to use one line per sentence for markdown files. That gives the smallest diffs when you change a few words.

CONTRIBUTING.md
arnout: In addition, please use correct spacing:

```suggestion
So to avoid creating more of these kinds of Klocwork issues, please always prefer `if (!ptr)` instead of `if (ptr == nullptr)`.
```

arnout (changes requested):
Incredible how much bikeshedding is possible on two sentences :-)

In addition to my inline comments...

1. This should go into CODING_SYTLE.md, not CONTRIBUTING.md. Put it as a new subsection `#### Klocwork` of the Coding patterns section.
2. The commit message doesn't say much. Should be something like: "CODING_STYLE.md: add klocwork constraints" - i.e., be more accurate about what is done. The why should be there as well, but in this case it's implied.

CONTRIBUTING.md
morantr: This is relevant only for smart pointers. Regular pointers can still be compared to `nullptr`. I think it is worth to mention it.

ghost (approved):

tomereli (approved):

morantr (approved):

arnout (approved):

PR #724, opened 2020-01-22
tlvf: Add default values to configData::config

tomereli:
In case the agent checks these values, make sure they are valid even
when we set teardown bit and they should be ignored.


tomereli:
Not needed.

PR #726, opened 2020-01-22
common: tests: move tests to top level directory

ghost:


arnout (commented):
I'm not sure why splitting the moves off in a separate commit is good. It's more difficult to review and it makes it un-bisectable, and it's also more confusing when looking at the history.

Also, I don't think it's appropriate for the tests-runner Dockerfile to move to the top level - it should stay in support/docker - though maybe better in support/docker/tests-runner.

.gitlab-ci.yml
```diff
@@ -49,9 +49,10 @@ run-tests:
   stage: test
   image: $CI_REGISTRY_IMAGE/tests-runner:18.09.7-dind
   script:
-    - cd tools/docker
+    - pushd tools/docker
```
rmelotte: `pushd` doesn't work because this jobs uses busybox's sh. I don't think it's needed though.
If possible I think `image-pull.sh` should be called from the root dir directly. I don't think we need to keep `image-build.sh`.

rmelotte (commented):
Only one small comment for now.
Also since we removed Travis entirely, we could remove `.travis.yml` so that we don't have to update it everytime.

.gitlab-ci.yml
```diff
@@ -141,7 +142,7 @@ test-on-netgear-rax40:
 .run-certification-tests:
   stage: test
   variables:
-    TESTS_TO_RUN: "tools/docker/tests/certification/all_controller_tests.txt"
+    TESTS_TO_RUN: "tests/certification/all_controller_tests.txt"
```
rmelotte: As we now provide this variable from the Gitlab schedules ([PR](https://github.com/prplfoundation/prplMesh/pull/730) not yet merged), we will need to update the value there.
When you merge this, please change it on Gitlab or notify me otherwise the nightly tests will fail.

.gitlab-ci.yml
```diff
@@ -54,7 +54,8 @@ run-tests:
   script:
     - cd tools/docker
     # Try to pull from dockerhub, build the container images if that fails
-    - ./image-pull.sh || ./image-build.sh
+    - ./image-pull.sh
+    - cd ../..
```
rmelotte: As I mentioned in my previous comment, please use `tools/docker/image-pull.sh` directly to avoid changing dirs around.

.travis.yml
```diff
@@ -21,4 +21,5 @@ script:
   - git diff-index --exit-code HEAD
   # Tests
   - ./run.sh --entrypoint "$PWD/../../build/install/bin/tests/tlvf_test"
+  - cd ..
```
rmelotte: `.travis.yml` can be removed.

rmelotte (approved):
Minor comments but please take them into account.

I also think the 2 commits should be squashed, because if you take them separately they break the tests.
Approving so that you don't have to wait for a re-review after you made the changes.

CONTRIBUTING.md
```diff
@@ -280,7 +280,7 @@ That means the following conditions must hold.
 * At least one maintainer has reviewed and approvied.
 * Code builds on Ubuntu, with `MSGLIB=zmq` and `BWL_TYPE=DUMMY`. Automatic with Travis CI.
 * Unit tests run successfully. Automatic with Travis CI.
-* For each new flow being added, `tools/docker/tests/test_flows.sh` and `.travis.yaml` must be updated accordingly to test the new flow.
+* For each new flow being added, `tests/test_flows.sh` and `.travis.yaml` must be updated accordingly to test the new flow.
```
arnout: This reference to .travis.yml should have been removed as well.

arnout (approved):
The removal of .travis.yml should have been a separate commit.

I'll do that and merge.

PR #727, opened 2020-01-22
Dev/dwpal update vap cfg

### LINKED ISSUE #716: BWL DWPAL update VAP configuration

dob71:
The testing has been completed. Please proceed with the review.

The code review changes have been pushed.
The comments marked resolved are addressed in the code, the ones still unresolved have my replies/answers/explanations.

Pushed more code review changes and posted a few answers.
Please mark the conversations resolved if the answers are satisfactory.

Fixes #716

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,12 @@ _GO_KW
 .kwlp
 .kwps
 kw_reports
+# ignore artifacts for building w/ openwrt USE_SOURCE_DIR
+.built
+.built_check
+.configured_*
+.pkgdir/
+.prepared
+.quilt_checked
+.source_dir
+ipkg-*/
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -16,7 +16,6 @@
 #include <bcl/network/network_utils.h>
 #include <bcl/son/son_wireless_utils.h>
 #include <easylogging++.h>
-
 #include <math.h>
 
 #ifdef USE_LIBSAFEC
@@ -416,6 +415,297 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
     return msg_buff;
 }
 
+///
+/// @brief Finds value by key in a string vector representing hostapd config
+///        (head or VAP) and returns value found in the supplied string object.
+///
+static bool hostapd_config_get_value(const std::vector<std::string> &hostapd_config,
+                                     const std::string &key, std::string &value)
+{
+    std::string key_eq(key + "=");
+    auto it_str = std::find_if(hostapd_config.begin(), hostapd_config.end(),
+                               [&key_eq](std::string str) -> bool {
+                                   return (str.compare(0, key_eq.length(), key_eq) == 0);
+                               });
+    if (it_str == hostapd_config.end()) {
+        return false;
+    }
+    value.assign(*it_str, key_eq.length(), std::string::npos);
+    return true;
+}
+
+///
+/// @brief Finds a value by key in a string vector representing hostapd config
+///        (head or VAP) and replaces it with the supplied one. If the value
+///        does not exist, it's added. If the value is empty string it's deleted.
+///
+static void hostapd_config_set_value(std::vector<std::string> &hostapd_config,
+                                     const std::string &key, const std::string &value)
+{
+    std::string key_eq(key + "=");
+    auto it_str = std::find_if(hostapd_config.begin(), hostapd_config.end(),
+                               [&key_eq](std::string str) -> bool {
+                                   return (str.compare(0, key_eq.length(), key_eq) == 0);
+                               });
+    // Delete the key-value if found
+    if (it_str != hostapd_config.end()) {
+        hostapd_config.erase(it_str);
+    }
+    // If the new value is provided add the key back with that new value
+    if (value.length() != 0) {
+        hostapd_config.push_back(key_eq + value);
+    }
+    return;
+}
+
+static void hostapd_config_set_value(std::vector<std::string> &hostapd_config,
+                                     const std::string &key, const int value)
+{
+    return hostapd_config_set_value(hostapd_config, key, std::to_string(value));
+}
+
+///
+/// @brief Loads hostapd configuration from specified file
+///
+static bool
+load_hostapd_config_file(const std::string &fname, std::vector<std::string> &hostapd_config_head,
+                         std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    std::ifstream ifs(fname);
+    std::string line;
+
+    bool parsing_vaps = false;
+    std::string cur_vap;
+    while (getline(ifs, line)) {
+        // Skip empty lines
+        if (beerocks::string_utils::is_empty(line)) {
+            continue;
+        }
+        // Check if the string belongs to a VAP config part and
+        // capture which one.
+        const std::string bss_eq("bss=");
+        if (line.compare(0, bss_eq.length(), bss_eq) == 0) {
+            parsing_vaps = true;
+            cur_vap.assign(line, bss_eq.length(), std::string::npos);
+        }
+        // If not a VAP line store it in the header part of the config,
+        // otherwise add to the currently being parsed VAP storage.
+        if (!parsing_vaps) {
+            hostapd_config_head.push_back(line);
+        } else {
+            std::vector<std::string> &hostapd_config_vap = hostapd_config_vaps[cur_vap];
+            hostapd_config_vap.push_back(line);
+        }
+    }
+
+    // Log something if there was an error during the read
+    if (ifs.bad()) {
+        LOG(ERROR) << "Unable to read " << fname << ": " << strerror(errno);
+    }
+
+    // If we've got to parsing VAPs and no read errors, assume all is good
+    return parsing_vaps && !ifs.bad();
+}
+
+///
+/// @brief Figures out hostapd config file name by the interface name and loads
+///        its content using load_hostapd_config_file()
+///        Warning: the content of the hostapd_config_head and hostapd_config_vaps
+///        is destroyed.
+///
+static bool
+load_hostapd_config(const std::string &radio_iface_name, std::string &fname,
+                    std::vector<std::string> &hostapd_config_head,
+                    std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    bool loaded                                = false;
+    std::vector<std::string> hostapd_cfg_names = {"/var/run/hostapd-phy0.conf",
+                                                  "/var/run/hostapd-phy1.conf"};
+
+    for (const auto &try_fname : hostapd_cfg_names) {
+        LOG(DEBUG) << "Trying to load " << try_fname << "...";
+
+        hostapd_config_head.clear();
+        hostapd_config_vaps.clear();
+        if (!beerocks::os_utils::file_exists(try_fname)) {
+            continue;
+        }
+        if (!load_hostapd_config_file(try_fname, hostapd_config_head, hostapd_config_vaps)) {
+            LOG(ERROR) << "Failed to load hostapd cofig file: " << try_fname;
+            continue;
+        }
+        // See if it's the right one
+        std::string ifname;
+        if (!hostapd_config_get_value(hostapd_config_head, "interface", ifname)) {
+            LOG(ERROR) << "No interface value in " << try_fname;
+            continue;
+        }
+        if (ifname != radio_iface_name) {
+            LOG(DEBUG) << "File " << try_fname << " interface " << radio_iface_name << " no match";
+            continue;
+        }
+        LOG(DEBUG) << "Loaded " << try_fname << " for interface " << radio_iface_name;
+
+        loaded = true;
+        fname.assign(try_fname);
+        break;
+    }
+    if (!loaded) {
+        hostapd_config_head.clear();
+        hostapd_config_vaps.clear();
+    }
+
+    return loaded;
+}
+
+///
+/// @brief Saves hostapd configuration to specified file
+///
+static bool
+save_hostapd_config_file(const std::string &fname, std::vector<std::string> &hostapd_config_head,
+                         std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    std::ofstream out(fname, std::ofstream::out | std::ofstream::trunc);
+
+    for (const auto &line : hostapd_config_head) {
+        out << line << "\n";
+    }
+    for (auto &it : hostapd_config_vaps) {
+        out << "\n"; // add empty line for readability
+        const std::vector<std::string> &hostapd_config_vap = it.second;
+        for (auto &line : hostapd_config_vap) {
+            out << line << "\n";
+        }
+    }
+
+    // Log something if there was an error during the write
+    out.close();
+    if (out.fail()) {
+        LOG(ERROR) << "Unable to write " << fname << ": " << strerror(errno);
+    }
+
+    return !out.fail();
+}
+
+static bool
+update_vap_credentials_configure_wpa(const std::string &vap_if,
+                                     std::vector<std::string> &vap_hostapd_config,
+                                     const son::wireless_utils::sBssInfoConf &bss_info_conf)
+{
+    // Hostapd "wpa" field.
+    // This field is a bit field that can be used to enable WPA (IEEE 802.11i/D3.0)
+    // and/or WPA2 (full IEEE 802.11i/RSN):
+    // bit0 = WPA
+    // bit1 = IEEE 802.11i/RSN (WPA2) (dot11RSNAEnabled)
+    int wpa = 0;
+
+    // Set of accepted key management algorithms (WPA-PSK, WPA-EAP, or both). The
+    // entries are separated with a space. WPA-PSK-SHA256 and WPA-EAP-SHA256 can be
+    // added to enable SHA256-based stronger algorithms.
+    // WPA-PSK = WPA-Personal / WPA2-Personal
+    std::string wpa_key_mgmt(""); // Empty -> delete from hostapd config
+
+    // (dot11RSNAConfigPairwiseCiphersTable)
+    // Pairwise cipher for WPA (v1) (default: TKIP)
+    //  wpa_pairwise=TKIP CCMP
+    // Pairwise cipher for RSN/WPA2 (default: use wpa_pairwise value)
+    //  rsn_pairwise=CCMP
+    std::string wpa_pairwise(""); // Empty -> delete from hostapd config
+
+    // WPA pre-shared keys for WPA-PSK. This can be either entered as a 256-bit
+    // secret in hex format (64 hex digits), wpa_psk, or as an ASCII passphrase
+    // (8..63 characters), wpa_passphrase.
+    std::string wpa_passphrase("");
+    std::string wpa_psk("");
+
+    // ieee80211w: Whether management frame protection (MFP) is enabled
+    // 0 = disabled (default)
+    // 1 = optional
+    // 2 = required
+    std::string ieee80211w("");
+
+    // This parameter can be used to disable caching of PMKSA created through EAP
+    // authentication. RSN preauthentication may still end up using PMKSA caching if
+    // it is enabled (rsn_preauth=1).
+    // 0 = PMKSA caching enabled (default)
+    // 1 = PMKSA caching disabled
+    std::string disable_pmksa_caching("");
+
+    // Opportunistic Key Caching (aka Proactive Key Caching)
+    // Allow PMK cache to be shared opportunistically among configured interfaces
+    // and BSSes (i.e., all configurations within a single hostapd process).
+    // 0 = disabled (default)
+    // 1 = enabled
+    std::string okc("");
+
+    // This parameter can be used to disable retransmission of EAPOL-Key frames that
+    // are used to install keys (EAPOL-Key message 3/4 and group message 1/2). This
+    // is similar to setting wpa_group_update_count=1 and
+    std::string wpa_disable_eapol_key_retries("");
+
+    // EasyMesh R1 only allows Open and WPA2 PSK auth&encryption methods.
+    // Quote: A Multi-AP Controller shall set the Authentication Type attribute
+    //        in M2 to indicate WPA2-Personal or Open System Authentication.
+    // bss_info_conf.authentication_type is a bitfield, but we are not going
+    // to accept any combinations due to the above limitation.
+    if (bss_info_conf.authentication_type == WSC::eWscAuth::WSC_AUTH_OPEN) {
+        wpa = 0x0;
+        if (bss_info_conf.encryption_type != WSC::eWscEncr::WSC_ENCR_NONE) {
+            LOG(ERROR) << "Autoconfiguration: " << vap_if << " encryption set on open VAP";
+            return false;
+        }
+        if (bss_info_conf.network_key.length() > 0) {
+            LOG(ERROR) << "Autoconfiguration: " << vap_if << " network key set for open VAP";
+            return false;
+        }
+    } else if (bss_info_conf.authentication_type == WSC::eWscAuth::WSC_AUTH_WPA2PSK) {
+        wpa = 0x2;
+        wpa_key_mgmt.assign("WPA-PSK");
+        // Cipher must include AES for WPA2, TKIP is optional
+        if ((uint16_t(bss_info_conf.encryption_type) & uint16_t(WSC::eWscEncr::WSC_ENCR_AES)) ==
+            0) {
+            LOG(ERROR) << "Autoconfiguration:  " << vap_if << " CCMP(AES) is required for WPA2";
+            return false;
+        }
+        if ((uint16_t(bss_info_conf.encryption_type) & uint16_t(WSC::eWscEncr::WSC_ENCR_TKIP)) !=
+            0) {
+            wpa_pairwise.assign("TKIP CCMP");
+        } else {
+            wpa_pairwise.assign("CCMP");
+        }
+        if (bss_info_conf.network_key.length() < 8 || bss_info_conf.network_key.length() > 64) {
+            LOG(ERROR) << "Autoconfiguration: " << vap_if << " invalid network key length "
+                       << bss_info_conf.network_key.length();
+            return false;
+        }
+        if (bss_info_conf.network_key.length() < 64) {
+            wpa_passphrase.assign(bss_info_conf.network_key);
+        } else {
+            wpa_psk.assign(bss_info_conf.network_key);
+        }
+        ieee80211w.assign("0");
+        disable_pmksa_caching.assign("1");
+        okc.assign("0");
+        wpa_disable_eapol_key_retries.assign("0");
+    } else {
+        LOG(ERROR) << "Autoconfiguration: " << vap_if << " invalid authentication type";
+        return false;
+    }
+
+    hostapd_config_set_value(vap_hostapd_config, "wpa", wpa);
+    hostapd_config_set_value(vap_hostapd_config, "okc", okc);
+    hostapd_config_set_value(vap_hostapd_config, "wpa_key_mgmt", wpa_key_mgmt);
+    hostapd_config_set_value(vap_hostapd_config, "wpa_pairwise", wpa_pairwise);
+    hostapd_config_set_value(vap_hostapd_config, "wpa_psk", wpa_psk);
+    hostapd_config_set_value(vap_hostapd_config, "ieee80211w", ieee80211w);
+    hostapd_config_set_value(vap_hostapd_config, "wpa_passphrase", wpa_passphrase);
+    hostapd_config_set_value(vap_hostapd_config, "disable_pmksa_caching", disable_pmksa_caching);
+    hostapd_config_set_value(vap_hostapd_config, "wpa_disable_eapol_key_retries",
+                             wpa_disable_eapol_key_retries);
+
+    return true;
+}
+
 //////////////////////////////////////////////////////////////////////////////
 /////////////////////////////// Implementation ///////////////////////////////
 //////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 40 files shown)


.gitignore
```diff
@@ -6,3 +6,11 @@ _GO_KW
 .kwlp
 .kwps
 kw_reports
+.built
```
arnout, line 10: It would be nice to add a comment above that says the following are openwrt build artefacts (nitpick)

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -170,11 +169,17 @@ int cfg_get_wifi_params(const char iface[BPL_IFNAME_LEN], struct BPL_WLAN_PARAMS
...
     }
 
-    retVal |= cfg_uci_get_wireless_bool(TYPE_RADIO, index, "disabled", &disabled);
-    if (!retVal) {
-        wlan_params->enabled = !disabled;
-    }
-    retVal |= cfg_get_channel(index, &wlan_params->channel);
+    // The UCI "disabled" setting is optional, defaults to false if not present
+    bool disabled = false;
+    cfg_uci_get_wireless_bool(TYPE_RADIO, index, "disabled", &disabled);
```
arnout, line 174: It would still be nice to have in INFO message that the default is used. (nitpick)

tools/docker/tests/send_CAPI_command.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, host: str, port: int, timeout : int = 10):
         self.host = host
         self.port = port
         self.timeout = timeout
-        self.conn: socket.socket
+        self.conn = socket.socket
```
arnout: This change is wrong.

The point of the statement is to declare `self.conn` as being of type socket.socket, but to not set it yet (it will only be set in the `__enter__` context manager). So to "fix" this issue, the statement should be removed entirely.

It took me some time to find out why this change is needed. The explanation in the commit message could be better. E.g.: "Type hints are supported since Python 3.5, but type hints for variables are only supported since Python 3.6. Therefore, remove the type hint for UCCSocket.conn."

framework/tlvf/tlvf.py
```diff
@@ -1151,10 +1151,6 @@ def addClassVarLenMethods(self, obj_meta, param_type, param_name, param_meta, pa
             lines_cpp.append( "%sTLVF_LOG(ERROR) << \"Out of order allocation for variable length list %s, abort!\";" %(self.getIndentation(2), param_name) )
             lines_cpp.append( "%sreturn false;" %self.getIndentation(2))
             lines_cpp.append( "%s}" %self.getIndentation(1))
-            lines_cpp.append( "%sif (count == 0) {" % (self.getIndentation(1)) )
-            lines_cpp.append( '%sTLVF_LOG(WARNING) << "can\'t allocate 0 bytes";' %  self.getIndentation(2) )
-            lines_cpp.append( "%sreturn false;" % self.getIndentation(2))
-            lines_cpp.append( "%s}" % self.getIndentation(1) )
```
arnout, line 1157: I don't see why this should be removed. Indeed, for the specific case of a string, it's duplicated because we already do this check in the set_xxx call and we normally don't call alloc_xxx directly. However, for fields of a different type, we do call alloc_xxx manually and allocating 0 elements is not supposed to happen.

framework/tlvf/tlvf.py
```diff
@@ -1009,8 +1009,8 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
 
                 lines_h.append( "bool set_%s(const char buffer[], size_t size);" % (param_name) )
                 lines_cpp.append( "bool %s::set_%s(const char str[], size_t size) {" % (obj_meta.name, param_name) )
-                lines_cpp.append( "%sif (str == nullptr || size == 0) {" % self.getIndentation(1))
-                lines_cpp.append( '%sTLVF_LOG(WARNING) << "set_%s received an empty string.";' %  (self.getIndentation(2), param_name) )
+                lines_cpp.append( "%sif (str == nullptr) {" % self.getIndentation(1))
+                lines_cpp.append( '%sTLVF_LOG(WARNING) << "set_%s received a null pointer.";' %  (self.getIndentation(2), param_name) )
```
arnout, line 1013: I think instead of continuing with the alloc_xxx, we should just add a separate condition that returns true in case size == 0.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
     return msg_buff;
 }
 
+// Checks if file exists
+static bool file_exists(const std::string &fname)
+{
+    struct stat st;
+    return (stat(fname.c_str(), &st) == 0);
+}
```
arnout: It would be better to put such a function in beerocks_utils.h (or utils.h) so it can be used more in general.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+// and returns value found in the supplied string object.
+static bool hostapd_config_get_value(const std::vector<std::string> &hostapd_config,
+                                     const std::string &key, std::string &value)
+{
+    std::string key_eq(key + "=");
+    auto it_str =
+        std::find_if(hostapd_config.begin(), hostapd_config.end(), [&](std::string str) -> bool {
+            return (str.compare(0, key_eq.length(), key_eq) == 0);
+        });
+    if (it_str >= hostapd_config.end()) {
```
arnout: Normally we compare with `== foo.end()`, not `>=`, because not all iterator types have a `>=` operator. Only a random-access iterator has this operator. Since hostapd_config is a vector, it happens to be a random-access iterator, but there's no reason at all to rely on that.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+        return false;
+    }
+    value.assign(*it_str, key_eq.length(), std::string::npos);
```
arnout, line 433: Can't we just return value (and empty string if not found)?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+    std::string key_eq(key + "=");
+    auto it_str =
+        std::find_if(hostapd_config.begin(), hostapd_config.end(), [&](std::string str) -> bool {
+            return (str.compare(0, key_eq.length(), key_eq) == 0);
+        });
+    // Delete the key if value is empty
+    if (value.length() == 0) {
+        if (it_str >= hostapd_config.end()) {
+            hostapd_config.erase(it_str);
+        }
```
arnout: That can't be right... I think you reverted the condition here.

Also, `==` instead of `>=`.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+        // We are good if it is already not present
+        return true;
+    }
+    // If the key is not found add it
+    if (it_str >= hostapd_config.end() && value.length() > 0) {
```
arnout: `==`.

`value.length()` was already checked above, so don't do it again here.

Also I prefer to have an explicit else for the value condition instead of relying on the return. return is good when checking preconditions or error situations, but when dealing with different "options", it's easier to read (IMO) if it's with explicit condition trees.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+        hostapd_config.push_back(key_eq + value);
+        return true;
+    }
+    // Replace the value
+    it_str->resize(key_eq.length() + value.length());
+    it_str->replace(key_eq.length(), std::string::npos, value);
+    return true;
+}
+static bool hostapd_config_set_value(std::vector<std::string> &hostapd_config,
```
arnout: Empty line before the function. I thought clang-format would do that.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+        return true;
+    }
+    // If the key is not found add it
+    if (it_str >= hostapd_config.end() && value.length() > 0) {
+        hostapd_config.push_back(key_eq + value);
+        return true;
+    }
+    // Replace the value
+    it_str->resize(key_eq.length() + value.length());
+    it_str->replace(key_eq.length(), std::string::npos, value);
```
arnout: This (= the whole function after the it_str declaration) can be simplified to:

```suggestion
if (it_str != hostapd_config.end()) {
    hostapd_config.erase(it_str);
}
if (value) {
    hostapd_config.push_back(key_eq + value);
}
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+        std::string ifname;
+        if (!hostapd_config_get_value(hostapd_config_head, "interface", ifname)) {
+            LOG(ERROR) << "No interface value in " << try_fname;
+            continue;
+        }
+        if (ifname.compare(radio_iface_name) != 0) {
+            LOG(DEBUG) << "File " << try_fname << " interface " << radio_iface_name << " no match";
+            continue;
+        }
+        LOG(DEBUG) << "Loaded " << try_fname << " for interface " << radio_iface_name;
```
arnout: This part is not so nice. However, it will anyway eventually be removed and replaced with a config file fragment that we generate ourselves, so I guess it can stay.

Maybe, however, we should do this search for the correct filename in the attach() function, and save the filename in a member variable.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+    bool loaded                                = false;
+    std::vector<std::string> hostapd_cfg_names = {"/var/run/hostapd-phy0.conf",
+                                                  "/var/run/hostapd-phy1.conf"};
+
+    for (auto &try_fname : hostapd_cfg_names) {
+        LOG(DEBUG) << "Trying to load " << try_fname << "...";
+
+        hostapd_config_head.clear();
+        hostapd_config_vaps.clear();
+        if (!file_exists(try_fname)) {
```
arnout: Is this useful? Can't we just fail on the load, below?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+    it_str->replace(key_eq.length(), std::string::npos, value);
+    return true;
+}
+static bool hostapd_config_set_value(std::vector<std::string> &hostapd_config,
+                                     const std::string &key, const int value)
+{
+    return hostapd_config_set_value(hostapd_config, key, std::to_string(value));
+}
+
+// Loads hostapd configuration from specidfied file
```
arnout: specified (typo)

BTW, use doxygen formatting for this documentation. Full details including arguments is not needed maybe, but even for a single-line docstring it's useful if it's doxygen-formatted.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+    bool parsing_vaps = false;
+    std::string cur_vap;
+    while (getline(ifs, line)) {
+        // Skip empty lines
+        if (line.find_first_not_of(' ') == std::string::npos) {
+            continue;
+        }
+        // Check if the string belongs to a VAP config part and
+        // capture which one.
+        std::string bss_eq("bss=");
```
arnout: Maybe make this one const

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+// Loads hostapd configuration from specidfied file
+static bool
+load_hostapd_config_file(const std::string &fname, std::vector<std::string> &hostapd_config_head,
+                         std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    std::ifstream ifs(fname);
+    std::string line;
+
+    bool parsing_vaps = false;
+    std::string cur_vap;
```
arnout, line 478: So instead of checking for file existence explicitly, you can just add here:
```
if (!ifs.good()) [
    return false;
}
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+    }
+    if (!loaded) {
+        hostapd_config_head.clear();
+        hostapd_config_vaps.clear();
+    }
+
+    return loaded;
+}
+
+// Saves hostapd configuration to specidfied file
```
arnout: Typo again, + doxygen

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+    // and/or WPA2 (full IEEE 802.11i/RSN):
+    // bit0 = WPA
+    // bit1 = IEEE 802.11i/RSN (WPA2) (dot11RSNAEnabled)
+    int wpa = 0;
+
+    // Set of accepted key management algorithms (WPA-PSK, WPA-EAP, or both). The
+    // entries are separated with a space. WPA-PSK-SHA256 and WPA-EAP-SHA256 can be
+    // added to enable SHA256-based stronger algorithms.
+    // WPA-PSK = WPA-Personal / WPA2-Personal
+    std::string wpa_key_mgmt(""); // Empty -> delete from hostapd config
```
arnout, line 606: Great that you added those descriptive comments! :1st_place_medal:

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+    // to accept any combinations due to the above limitation.
+    if (bss_info_conf.authentication_type == WSC::eWscAuth::WSC_AUTH_OPEN) {
+        wpa = 0x0;
+        if (bss_info_conf.encryption_type != WSC::eWscEncr::WSC_ENCR_NONE) {
+            LOG(ERROR) << "Autoconfiguration: " << vap_if << " encryption set on open VAP";
+            return false;
+        }
+        if (bss_info_conf.network_key.length() > 0) {
+            LOG(ERROR) << "Autoconfiguration: " << vap_if << " network key set for open VAP";
+            return false;
```
arnout, line 659: I don't know if we really should return false here. It is possible that some bad implementation does send a network key even for an open network.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+static bool file_exists(const std::string &fname)
+{
+    struct stat st;
+    return (stat(fname.c_str(), &st) == 0);
+}
+
+// Finds value by key in a string vector representing hostapd config (head or VAP)
+// and returns value found in the supplied string object.
+static bool hostapd_config_get_value(const std::vector<std::string> &hostapd_config,
+                                     const std::string &key, std::string &value)
```
arnout, line 423: I think it would be worth adding a small private class for hostapd_config that has this get_value as a member function and also set_value() and a load() function that does the getline calls. (nitpick)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+        return false;
+    }
+    value.assign(*it_str, key_eq.length(), std::string::npos);
+    return true;
+}
+
+// Finds a value by key in a string vector representing hostapd config (head or VAP)
+// and replaces it with the supplied one. If the value does not exist, it's added.
+// If the value is empty string it's deleted.
+static bool hostapd_config_set_value(std::vector<std::string> &hostapd_config,
```
arnout: This function always returns true, so maybe don't return anything?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -647,8 +945,175 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    // At the moment we are taking a shortcut to make this work for certification setup.
+    // We are updating the hostapd .conf files with the info from the bss_info_conf_list
+    // and sending RELOAD command to hostapd making it update the VAPs.
+    // With that approach there are following limitations:
+    // - all the VAPs are expected to be always present in the config, but have
+    //   start_disabled=1 option if not active
+    // - any configuration change causing the hostapd config files to be re-genrated
+    //   and the hostapd processes restarted wil kill the purplMesh configuration
+
+    // The hostapd config is stored here. The items order in the header and VAPs sections
```
arnout, line 951: Why does the order have to be preserved?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -647,8 +945,175 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    // Go through the bss_info_conf_list and change the hostapd config accordingly
+    for (const auto &bss_info_conf : bss_info_conf_list) {
+        // If a Multi-AP Agent receives an AP-Autoconfiguration WSC message containing an M2 with
+        // a Multi-AP Extension subelement with bit 4 (Tear Down bit) of the subelement value set
+        // to one (see Table 4), it shall tear down all currently operating BSS(s) on the radio
+        // indicated by the Radio Unique Identifier, and shall ignore the other attributes in the M2.
+        if ((bss_info_conf.bss_type & WSC::eWscVendorExtSubelementBssType::TEARDOWN) != 0) {
+            LOG(DEBUG) << "Autoconfiguration: received teardown, continuing";
+            continue;
+        }
```
arnout: Actually I think it would be more appropriate to handle this in son_slave_thread, and just send an empty bss_info_conf_list if the teardown bit was set. There is already special handling for teardown anyway.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -647,8 +945,175 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+        LOG(DEBUG) << "Autoconfiguration for ssid: " << bss_info_conf.ssid
+                   << " auth_type: " << auth_type << " encr_type: " << enc_type
+                   << " network_key: " << bss_info_conf.network_key << " bss_type: " << bss_type;
+
+        // We cannot configure STAs yet
+        if ((bss_info_conf.bss_type & WSC::eWscVendorExtSubelementBssType::BACKHAUL_STA) != 0) {
+            LOG(ERROR) << "Autoconfiguration: STAs are not yet supported";
+            return false;
+        }
```
arnout: This is also something that is more appropriate to handle in the son_slave_thread.

Note BTW that the spec doesn't even say what this bit would mean when it is in a WSC M2 message. It's only meaningful during WPS.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -647,8 +945,175 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+
+        // We only can use AP entries, skip STAs if ended up on one
+        std::string entry_mode;
+        while (hostapd_vap_iterator != hostapd_config_vaps.end() &&
+               hostapd_config_get_value(hostapd_vap_iterator->second, "mode", entry_mode) &&
+               entry_mode.compare("ap") != 0) {
```
arnout: Doesn't `entry_mode == "ap"` work?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -647,8 +945,175 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+        // Enable Multi-AP functionality
+        // 0 = disabled (default)
+        // 1 = AP support backhaul BSS
+        // 2 = AP support fronthaul BSS
+        // 3 = AP supports both backhaul BSS and fronthaul BSS
+        int multi_ap = 0;
+        multi_ap |=
+            ((bss_info_conf.bss_type & WSC::eWscVendorExtSubelementBssType::BACKHAUL_BSS) != 0)
+                ? 0x01
+                : 0;
```
arnout: IMHO it's easier to read/understand as
```
if (bss_info_conf.bss_type & WSC::eWscVendorExtSubelementBssType::BACKHAUL_BSS) != 0) {
    multi_ap |= 0x01;
}
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -647,8 +945,175 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+
+    // Start by disabling all the existent VAPs
+    for (auto &it : hostapd_config_vaps) {
+        const std::string &vap_if                    = it.first;
+        std::vector<std::string> &vap_hostapd_config = it.second;
+
+        if (!hostapd_config_set_value(vap_hostapd_config, "start_disabled", "1")) {
+            LOG(ERROR) << "Autoconfiguration: cannot disable VAP " << vap_if << ", ignoring";
+        }
+    }
```
arnout: In the code below, we skip the non-ap VAPs, which IMO is a good thing. But that means we have to skip them here as well.

So I think a better approach maybe is to move this *after* the bss_info loop, and just disable all the remaining VAPs (still skipping the non-ap VAPs).

This is a nitpick since it will anyway be reworked later. Maybe add a TODO to remember to do this correctly when it is reworked.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -647,8 +945,175 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+            LOG(ERROR) << "Autoconfiguration: \"" << cmd << "\" command to hostapd has failed!";
+            // Keep going and try to complete what we can
+            continue;
+        }
+        ++vap_ok_count;
+    }
+
+    LOG(INFO) << "Autoconfiguration: completed successfully for " << vap_ok_count << " out of "
+              << hostapd_config_vaps.size() << " available hostapd BSS sections";
+    return (vap_ok_count == hostapd_config_vaps.size());
```
arnout: We don't ned the OK count, just success, so I'd use `bool success = true;` instead. Also the final INFO message is redundant since we already report ERROR on each failure.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -157,8 +157,8 @@ int cfg_get_device_info(BPL_DEVICE_INFO *device_info) { return 0; }
 
 int cfg_get_wifi_params(const char iface[BPL_IFNAME_LEN], struct BPL_WLAN_PARAMS *wlan_params)
 {
-    int retVal    = 0;
-    int index     = 0;
+    int retVal = 0;
+    int index  = 0;
```
arnout, line 161: This change should have been part of the commit that updates this function.

arnout (approved):
As usual for a big PR, I have many comments, but nothing controversial, so I'm approving already. For the nitpick comments, you can choose to implement them or not. For the important ones, you really need to implement them, but then you can merge yourself after CI passes. Until them, I'm marking as Don't Merge.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
morantr: `beerocks_os_utils` is the right place.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+}
+
+// Finds value by key in a string vector representing hostapd config (head or VAP)
+// and returns value found in the supplied string object.
+static bool hostapd_config_get_value(const std::vector<std::string> &hostapd_config,
+                                     const std::string &key, std::string &value)
+{
+    std::string key_eq(key + "=");
+    auto it_str =
+        std::find_if(hostapd_config.begin(), hostapd_config.end(), [&](std::string str) -> bool {
```
morantr: There is no need to pass everything as a reference to the lambda function. Please move only `&key_eq`.

morantr (changes requested):
I didn't finish the review. This is what I got so far.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
     return msg_buff;
 }
 
+// Checks if file exists
```
morantr: We use Doxygen to document function. Please convert functions description to Doxygen.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+
+// Finds a value by key in a string vector representing hostapd config (head or VAP)
+// and replaces it with the supplied one. If the value does not exist, it's added.
+// If the value is empty string it's deleted.
+static bool hostapd_config_set_value(std::vector<std::string> &hostapd_config,
+                                     const std::string &key, const std::string &value)
+{
+    std::string key_eq(key + "=");
+    auto it_str =
+        std::find_if(hostapd_config.begin(), hostapd_config.end(), [&](std::string str) -> bool {
```
morantr: Here also, please move as reference only `key_eq`.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
morantr: Agree with @arnout , `it_str != hostapd_config.end()` means that the element exist so it can be deleted.
What you did is equal to `it_str == hostapd_config.end()` which means the oppisite.
Please change it to `!=`.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+                         std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    std::ifstream ifs(fname);
+    std::string line;
+
+    bool parsing_vaps = false;
+    std::string cur_vap;
+    while (getline(ifs, line)) {
+        // Skip empty lines
+        if (line.find_first_not_of(' ') == std::string::npos) {
```
morantr: On `beerocks_string_utils.cpp` you can find the following definition: `std::string WHITESPACE_CHARS(" \t\n\r\f\v")`.
I think it will be a good idea to create a function there that does what you do here, only with `WHITESPACE_CHARS`.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+static bool
+load_hostapd_config(const std::string &radio_iface_name, std::string &fname,
+                    std::vector<std::string> &hostapd_config_head,
+                    std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    bool loaded                                = false;
+    std::vector<std::string> hostapd_cfg_names = {"/var/run/hostapd-phy0.conf",
+                                                  "/var/run/hostapd-phy1.conf"};
+
+    for (auto &try_fname : hostapd_cfg_names) {
```
morantr: const auto &try_fname

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +417,303 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+}
+
+// Saves hostapd configuration to specidfied file
+static bool
+save_hostapd_config_file(const std::string &fname, std::vector<std::string> &hostapd_config_head,
+                         std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    std::ofstream out(fname, std::ofstream::out | std::ofstream::trunc);
+
+    for (auto &line : hostapd_config_head) {
```
morantr: const auto &
also on the lines above.

morantr (dismissed):

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
dob71: I'll add INFO message for the "channel" setting (that one is a bit awkward),
"disabled" option is not present (normally) and only added by the UI when the interface is to be disabled (for this one message would just be noise).

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: I wanted the success message showing how many VAPs were processed (since that shows how many are available in the hostapd config to use by prplMesh), therefore the counter. This is a rare operation and the info message is quite useful here to confirm that auto-configuration has completed. I can change to debug if it really feels like it's polluting the logs.

framework/tlvf/tlvf.py
dob71: Why handling 0 should be a special case requiring adding extra code for variable length values?
If the parameter should not accept 0 then it has to be enforced for that specific parameter, but not generally.
I.e. the check belongs in that set_xxx call you mentioned or that manually written code calling alloc_xxx. Such explicit verification seems more appropriate than doing it behind the scenes. 
That said, I'm new to this code and if things are done that way here I'll change it.

framework/tlvf/tlvf.py
dob71: See the other comment in this file. Let me know if you still think it's better to just add a check above line 1022 and return true prematurely. I'll do it in a separate pull request then since this change will trigger tons of changes in the auto-generated committed files which probably will make this pull request harder to deal with.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: It's more versatile to return bool here and potentially allow to deal with keys set to empty values.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: I'm trying to preserve the order of parameters whenever possible. Are you certain that there are no ordering requirements in hostapd config besides the bss key-value pairs being on top of the VAP sections? I can just remove/add key-values if that's the case.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: I thought about that, it should be available in m_radio_info, but decided not to spend time on that for exactly the reason that this is a temporary solution.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: That's to avoid the error message if the radio is off.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: see the answer in the previous comment about the file presence check

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: The spec says it has to be 0 bytes long. If it sends a key string for open VAP that implementation is probably doing something wrong and potentially insecure.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: Is it going to survive the rework of this code? Does it make sense to do it now or is it better to get this code in to have auto-configuration working for the certification tests?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: I tried to put it in a map first, everything got messed up and didn't work, 
so decided it's more efficient to try preserving the order than to risk something going wrong,
it also makes tracking the changes easier and the config is more readable.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: Yes, it's not covered in the spec, but I have a suspicion it will be needed unless the expectation is that people should run around pushing buttons every time the backhaul SSID changes. If it ends up to be needed then its place is here, if not then in the son_slave_thread. I'd prefer to avoid moving things around unless it's useful. Do you have access to R2 or 3 of the spec (I only have R1)?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: Probably, I just prefer to stay away from operators overloading. If it's something that I should be doing in this code please let me know and I'll search/replace "compare()" to "==".

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: It seems like this has to be done, otherwise even experiment with the wireless backhauls will not be possible till this code is gone.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
dob71: I'll try not to forget about this when rebasing...

framework/tlvf/tlvf.py
arnout: Yes, good point.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: Yes, please use `==` for comparing strings. It's the idiomatic way.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: OK fair enough.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: You mean, distinguish between an empty value and not set? That's a good point indeed.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: There are no ordering requirements in the hostapd.conf file.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: No, indeed, it's OK to go in as it is.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: You do indeed need separate maps for the head and the VAP sections, and you do need to preserve the order of the VAP sections. But the lines within a section are unordered.

Anyway, preserving the order is definitely not a problem. Maybe extend the comment with something about making the config more readable and diffable.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: I have access to R2, and they haven't clarified this.

Using this to implement backhaul STA is a very liberal interpretation of the spec I think. The problem is that the backhaul STA is not tied to a particular radio, while the autoconfig messages are tied to a radio. If you have a radio dedicated to backhaul STA, it wouldn't even do any autoconfig (because you don't want VAPs on it).

Anyway, if it is used for backhaul STA configuration, then this is *definitely* not the place to handle it. STA handling is in sta_wlan_hal, and it's not talking to hostapd, it's talking to wpa_supplicant. And it's done in the backhaul manager, not in the slave.

arnout: Note that this is a nitpick.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: Wireless backhaul are not VAPs. It's completely separate.

A wireless backhaul (usually) does count towards the maximum number of BSSes that can be configured on a radio, but we can only handle that by reserving a BSS slot for the bSTA.

common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_backhaul.cpp
```diff
@@ -41,8 +41,8 @@ char* cACTION_BACKHAUL_REGISTER_REQUEST::sta_iface(size_t length) {
 
 bool cACTION_BACKHAUL_REGISTER_REQUEST::set_sta_iface(const std::string& str) { return set_sta_iface(str.c_str(), str.size()); }
 bool cACTION_BACKHAUL_REGISTER_REQUEST::set_sta_iface(const char str[], size_t size) {
-    if (str == nullptr || size == 0) {
-        TLVF_LOG(WARNING) << "set_sta_iface received an empty string.";
+    if (str == nullptr) {
```
rmelotte, line 44: Since you anyway change this line (and all the others), you could have gone for `if(!str)` but ok...

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -179,7 +179,9 @@ int cfg_get_wifi_params(const char iface[BPL_IFNAME_LEN], struct BPL_WLAN_PARAMS
     // fail when wifi still works fine, so default to "auto" (0) and if
     // can't get the channel from UCI just move on.
     wlan_params->channel = 0;
-    cfg_get_channel(index, &wlan_params->channel);
+    if (!cfg_get_channel(index, &wlan_params->channel)) {
+        MAPF_INFO("UCI: radio" << index << ": channel is not configured assuming auto\n");
```
rmelotte, line 183: I don't think the `\n` is needed.

rmelotte (approved):
I don't see any big issue, please don't forget to rebase and clean up the commits.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: Ok, at the moment wpa_supplicant is not even built into the RAX40 image, and its possible to configure a virtual interface into a client mode, which results in the hostapd configuration "mode" for for that interface changing. Perhaps it's something specific to the Intel driver (i.e. when the supplicant functionality is built-in and the external supplicant is not required). 
I don't know the details. I'll just change the messaging to say we cannot configure STA here and leave it in place. If the situation changes in the future the message will go away, if not it won't hurt to have it here.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
dob71: I had the same hunch, but given that all other messages there are \n terminated decided that if not for correctness then for uniformity it has to be there :-)

common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_backhaul.cpp
dob71: This was intentional. It's not compact, but a bit easier to read, also the search for "str == nullptr" showed that it is more common for this codebase.

tomereli: See https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md#klocwork - KW complains about this, so we decided to opt for `if (!ptr)` over `if (`ptr == nullptr)`.
We don't run KW on the autogenerated code yet so its not an issue, however if we can fix some things along the way then we certainly should.

tomereli (changes requested):
.

tomereli (approved):
Please rebase-squash-merge
Note that you added regular commits for code review fixups - normally you should use --fixup=<commit hash> then you can autosquash using `git rebase -i --autosquash origin/master`.
Now since you didn't, you need to do this manually with interactive rebase.

tomereli:
I'm keeping the don't merge tag since I'm now testing with this in the testbed - it doesn't always work, still debugging. I will request changes once I get something conclusive, for the meanwhile - don't merge please.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1002,17 +995,17 @@ bool ap_wlan_hal_dwpal::update_vap_credentials(
...
+            LOG(ERROR) << "Autoconfiguration: STA cannot be configured here";
             return false;
         }
 
         // We only can use AP entries, skip STAs if ended up on one
         std::string entry_mode;
         while (hostapd_vap_iterator != hostapd_config_vaps.end() &&
                hostapd_config_get_value(hostapd_vap_iterator->second, "mode", entry_mode) &&
-               entry_mode.compare("ap") != 0) {
+               entry_mode == "ap") {
```
tomereli: This looks like a wrong comparison - didn't you mean `entry_mode != "ap"` ?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1065,7 +1058,7 @@ bool ap_wlan_hal_dwpal::update_vap_credentials(
         // Skip STAs
         std::string entry_mode;
         if (hostapd_config_get_value(vap_hostapd_config, "mode", entry_mode) &&
-            entry_mode.compare("ap") != 0) {
+            entry_mode == "ap") {
```
tomereli: Same here

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1090,7 +1083,7 @@ bool ap_wlan_hal_dwpal::update_vap_credentials(
         // Skip STAs
         std::string entry_mode;
         if (hostapd_config_get_value(it.second, "mode", entry_mode) &&
-            entry_mode.compare("ap") != 0) {
+            entry_mode == "ap") {
```
tomereli: and here

tomereli (changes requested):
Also - there is a bug in loading the hostapd conf file - the correct file is not always loaded. I'm still debugging why:

>DEBUG 12:29:11:127 <1990487312> ap_wlan_hal_dwpal.cpp[940] --> TOMER update vap credentials
DEBUG 12:29:11:128 <1991855088> son_slave_thread.cpp[4396] --> Received LINK_METRIC_QUERY_MESSAGE, mid=6d31
DEBUG 12:29:11:130 <1990487312> ap_wlan_hal_dwpal.cpp[526] --> Trying to load /var/run/hostapd-phy0.conf...
DEBUG 12:29:11:134 <1990487312> ap_wlan_hal_dwpal.cpp[544] --> File /var/run/hostapd-phy0.conf interface wlan0 no match
DEBUG 12:29:11:135 <1990487312> ap_wlan_hal_dwpal.cpp[526] --> Trying to load /var/run/hostapd-phy1.conf...
DEBUG 12:29:11:138 <1990487312> ap_wlan_hal_dwpal.cpp[547] --> Loaded /var/run/hostapd-phy1.conf for interface wlan0
DEBUG 12:29:11:140 <1990487312> ap_wlan_hal_dwpal.cpp[965] --> Loaded hostapd config /var/run/hostapd-phy1.conffor iface wlan0

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -416,6 +415,297 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
...
+            LOG(ERROR) << "Failed to load hostapd cofig file: " << try_fname;
+            continue;
+        }
+        // See if it's the right one
+        std::string ifname;
+        if (!hostapd_config_get_value(hostapd_config_head, "interface", ifname)) {
+            LOG(ERROR) << "No interface value in " << try_fname;
+            continue;
+        }
+        if (ifname == radio_iface_name) {
```
tomereli: This is a bug - condition should be `if (ifname == radio_iface_name)`

dob71:
> Please rebase-squash-merge
> Note that you added regular commits for code review fixups - normally you should use --fixup= then you can autosquash using `git rebase -i --autosquash origin/master`.
> Now since you didn't, you need to do this manually with interactive rebase.

I was asked to put specific changes into specific commits. The changes will be rearranged into 3 commits (the stuff unrelated to autoconfig will go to the other 2, autoconfiguration to the 3rd one) after the reviews are done.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
dob71: Yes, good catch, I have not tested after changing the function calls to compare operators. Did not want to spend time on that since the review was still not finished.

tomereli (approved):

PR #729, opened 2020-01-23
docker: multiple users support

tomereli:
Add multiple users support for docker scripts so that multiple users on
the same server can work in parallel running test_flows.sh.

This requires basically renaming all containers to include the username
of the current user.
In addition, we also add the current user prefix to the docker network
which is used for the test, for obvious reasons.

### DIFF
--- a/tools/docker/run.sh
+++ b/tools/docker/run.sh
@@ -24,6 +24,7 @@ usage() {
     echo "      -p|--port - port to expose on the container"
     echo "      -I|--image - docker network to which to attach the container"
     echo "      -N|--network - docker network to which to attach the container"
+    echo "      -u|--unique-id - unique id to add as suffix to container and network names"
     echo "      --entrypoint - use a different entrypoint for the container"
 }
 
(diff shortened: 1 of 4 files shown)


tools/docker/run.sh
```diff
@@ -42,7 +42,7 @@ generate_container_random_ip() {
 }
 
 gateway_netid_length() {
-    docker network inspect prplMesh-net --format "{{(index .IPAM.Config 0).Subnet}}" | sed -rn 's/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}(\/[0-9]{2})$/\1/p'
+    docker network inspect prplMesh-net-${SUDO_USER:-${USER}} --format "{{(index .IPAM.Config 0).Subnet}}" | sed -rn 's/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}(\/[0-9]{2})$/\1/p'
```
arnout: instead of the username, it makes more sense to pass in a unique ID as an argument (which could default to `${SUDO_USER:-${USER}}`. With that, we can get rid of the docker-in-docker approach and instead run it directly in a shell runner. That would speed up our run-tests CI job significantly because the cache can be used (now it re-downloads the images every time).

If adding the argument is too much work, please at least make it a variable e.g. CONTAINER_NAME_SUFFIX.

arnout (approved):
I made an important remark but can be fixed easily. And if not, that's fine too, we can fix it later.

rmelotte (approved):
I agree with Arnout, using an argument/variable would save us time when we get rid of the docker-in-docker approach.

Also if the network/container names are made unique, we really have to ensure that every script that creates a container/network tries to delete it at the end of the test (it's not that important if the username is used).

tools/docker/run.sh
tomereli: I should probably already know this - but what do you mean by the docker-in-docker approach?

rmelotte: Because the tests cannot be run in parallel on the same host at the moment, the CI runs the prplmesh-runner containers in a docker container (a new docker daemon is run inside a docker container). It works, but it's a bit overkill and has a few downsides like not keeping the cache and images so we would like to use something else (either a shell runner, or keeping the docker runner but with a single daemon).

tomereli: Understood.

tomereli: Done

PR #730, opened 2020-01-23
.gitlab-ci.yml: remove TESTS_TO_RUN

rmelotte:
Due to a number of issues with setting variables manually using
Gitlab's web UI, setting TESTS_TO_RUN to run all the controller tests
can accidentally trigger more tests to be run than what was intended.
For example: when a job is retried all manually set variables are
ignored and take back their default value.

Remove TESTS_TO_RUN to force the user to specify the tests to run. If
TESTS_TO_RUN is not set for any reason, the test will fail.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -141,7 +141,7 @@ test-on-netgear-rax40:
 .run-certification-tests:
   stage: test
   variables:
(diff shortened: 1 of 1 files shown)


rmelotte:
Schedules still work: https://gitlab.com/prpl-foundation/prplMesh/pipelines/111728977 .

ghost (approved):

tomereli (approved):

PR #740, opened 2020-01-24
Feature/rdkb toolchain docker builder

tomereli:
In order to build Intel RDKB, we need to use xenial Ubuntu at
the moment, so add a Dockerfile to allow building RDKB toolchain from a
container.

The entrypoint script includes sourcing puma_setup so the next command
should be bitbake.

The build.sh script maps uses the current user id and group id as the
container group id and user id so newly created files will have the
same permissions as in the host.

The RDKB root folder is the only argument for the script (positional)
and is also mapped into the container.

Finally, since during RDKB build there is a need to clone via git
repositories, also map the user HOME directory so we are using the same
.gitconfig as the user does which should fix any git proxy issues.


vitalybu (approved):
Looks good to me

tools/docker/builder/rdkb/toolchain/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+echo RDKB_ROOT=$RDKB_ROOT
+cd $RDKB_ROOT/setup
+
+if [ -d "${RDKB_ROOT}/atom_rdkbos/build" ]; then 
+    . puma_setup -o 
+else
+    . puma_setup -c configs/p7_atom_rdkbos_dev-debug.config
```
morantr: we use `configs/p7_atom_rdkbos_wave600_dev-debug.config`

morantr (approved):

tools/docker/builder/rdkb/toolchain/entrypoint.sh
tomereli: Thanks, I'll change it then. Good catch!

PR #747, opened 2020-01-27
[Patch 1/2] Feature/client capability response [Test 4.6.2]

CoralMalachi:
### Description
According to Multi-AP Specification, If a Multi-AP Agent receives a
Client Capability Query message, then within one second it shall respond
with a Client Capability Report message.

If the STA specified in the Client Capability Query message is not
associated with any of the BSS operated by the Multi-AP Agent (an error
scenario), the Multi-AP Agent shall set the result code in the Client
Capability Report TLV to 0x01  and include an Error Code TLV.

### Changes
Therefore, the following changes were made:

- Create a Client Capability Report message CMDU.

- Add the Client Info TLV.

- Add a Client Capability Report TLV.

- Check that the Multi-AP Agent has the BSSID that was specified in the Client Info TLV, and that the client is associated with that BSS. If not, this is an error scenario, so set the result code in the Client
Capability Report TLV to 0x01, and include an Error Code TLV with the reason code set to 0x02.

### Testbed status
This branch was tested on testbed while using Qualcomm as CTT Controller, Marvell STA as client and prplMesh as MAUT.
Currently, there is an issue with the Marvell STA, which can not connect to the prplMesh agent.
Since the client was not connected to the agent, the Client Capability Report contained an Error Code TLV as it should.

_Status update: test 4.6.2 **passed** the test on the testbed._

### Next steps

- [x] After solving the issue with the Marvell STA, run the test again and verify that in case of success the Client Capability Report does not contain an Error Code TLV.

- [ ] Client Capability Report TLV  should include the frame body of the most recently received (Re)Association Request frame from this client (if it is an error scenario this field is omitted).
This part will be done in a separate PR.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -39,6 +39,9 @@
 #include <tlvf/wfa_map/tlvApOperationalBSS.h>
 #include <tlvf/wfa_map/tlvApRadioBasicCapabilities.h>
 #include <tlvf/wfa_map/tlvAssociatedClients.h>
+#include <tlvf/wfa_map/tlvClientCapabilityReport.h>
+#include <tlvf/wfa_map/tlvClientInfo.h>
+#include <tlvf/wfa_map/tlvErrorCode.h>
 #include <tlvf/wfa_map/tlvHigherLayerData.h>
 #include <tlvf/wfa_map/tlvSearchedService.h>
 #include <tlvf/wfa_map/tlvSupportedService.h>
@@ -1805,13 +1808,85 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     case ieee1905_1::eMessageType::COMBINED_INFRASTRUCTURE_METRICS_MESSAGE: {
         return handle_1905_combined_infrastructure_metrics(cmdu_rx, src_mac);
     }
-
+    case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE:
+        return handle_client_capability_query(cmdu_rx, src_mac);
     default: {
         // TODO add a warning once all vendor specific flows are replaced with EasyMesh
         // flows, since we won't expect a 1905 message not handled in this function
         return false;
     }
     }
+}
+
+bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                      const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received CLIENT_CAPABILITY_QUERY_MESSAGE , mid=" << std::dec << int(mid);
+
+    auto client_info_tlv_r = cmdu_rx.getClass<wfa_map::tlvClientInfo>();
+    if (!client_info_tlv_r) {
+        LOG(ERROR) << "getClass wfa_map::tlvClientInfo failed";
+        return false;
+    }
+
+    //Check if it is an error scenario - if the STA specified in the Client Capability Query message is not associated
+    //with any of the BSS operated by the Multi-AP Agent [ though the TLV does contain a BSSID, the specification
+    // says that we should answer if the client is associated with any BSS on this agent.]
+    bool associated_client_found = false;
+    for (const auto &slave : m_radio_info_map) {
+        auto associated_clients_map = slave.second.associated_clients_map;
+        for (const auto &vap : associated_clients_map) {
+            if (vap.second.find(client_info_tlv_r->client_mac()) != vap.second.end()) {
+                associated_client_found = true;
+                break;
+            }
+        }
+    }
+
+    // send CLIENT_CAPABILITY_REPORT_MESSAGE back to the controller
+    if (!cmdu_tx.create(mid, ieee1905_1::eMessageType::CLIENT_CAPABILITY_REPORT_MESSAGE)) {
+        LOG(ERROR) << "cmdu creation of type CLIENT_CAPABILITY_REPORT_MESSAGE, has failed";
+        return false;
+    }
+
+    auto client_info_tlv_t = cmdu_tx.addClass<wfa_map::tlvClientInfo>();
+    if (!client_info_tlv_t) {
+        LOG(ERROR) << "addClass wfa_map::tlvClientInfo has failed";
+        return false;
+    }
+    client_info_tlv_t->bssid()      = client_info_tlv_r->bssid();
+    client_info_tlv_t->client_mac() = client_info_tlv_r->client_mac();
+
+    auto client_capability_report_tlv = cmdu_tx.addClass<wfa_map::tlvClientCapabilityReport>();
+    if (!client_capability_report_tlv) {
+        LOG(ERROR) << "addClass wfa_map::tlvClientCapabilityReport has failed";
+        return false;
+    }
+
+    //if it is an error scenario, set Success status to 0x01 = Failure and do nothing after it.
+    if (associated_client_found) {
+        client_capability_report_tlv->result_code() = wfa_map::tlvClientCapabilityReport::SUCCESS;
+        LOG(DEBUG) << "Result Code: SUCCESS";
+        //TODO: Add frame body of the most recently received (Re)Association Request frame from this client
+    } else {
+        client_capability_report_tlv->result_code() = wfa_map::tlvClientCapabilityReport::FAILURE;
+
+        LOG(DEBUG) << "Result Code: FAILURE";
+        LOG(DEBUG) << "STA specified in the Client Capability Query message is not associated with "
+                      "any of the BSS operated by the Multi-AP Agent ";
+        //Add an Error Code TLV
+        auto error_code_tlv = cmdu_tx.addClass<wfa_map::tlvErrorCode>();
+        if (!error_code_tlv) {
+            LOG(ERROR) << "addClass wfa_map::tlvErrorCode has failed";
+            return false;
+        }
+        error_code_tlv->reason_code() =
+            wfa_map::tlvErrorCode::STA_NOT_ASSOCIATED_WITH_ANY_BSS_OPERATED_BY_THE_AGENT;
+        error_code_tlv->sta_mac() = client_info_tlv_r->client_mac();
+    }
+    LOG(DEBUG) << "Send a CLIENT_CAPABILITY_REPORT_MESSAGE back to controller";
+    return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
 }
 
 bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmdu_rx,
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -385,8 +385,6 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         return handle_client_steering_request(sd, cmdu_rx);
     case ieee1905_1::eMessageType::ACK_MESSAGE:
         return handle_ack_message(sd, cmdu_rx);
-    case ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE:
-        return handle_client_capability_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE:
         return handle_multi_ap_policy_config_request(sd, cmdu_rx);
     default:
@@ -4195,13 +4193,6 @@ bool slave_thread::parse_non_intel_join_response(Socket *sd)
 
     slave_state = STATE_UPDATE_MONITOR_SON_CONFIG;
 
-    return true;
-}
-
-bool slave_thread::handle_client_capability_query(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
-{
-    const auto mid = cmdu_rx.getMessageId();
-    LOG(DEBUG) << "Received CLIENT_CAPABILITY_QUERY_MESSAGE , mid=" << std::dec << int(mid);
     return true;
 }
 
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -48,6 +48,8 @@
 #include <tlvf/wfa_map/tlvChannelPreference.h>
 #include <tlvf/wfa_map/tlvChannelSelectionResponse.h>
 #include <tlvf/wfa_map/tlvClientAssociationEvent.h>
+#include <tlvf/wfa_map/tlvClientCapabilityReport.h>
+#include <tlvf/wfa_map/tlvClientInfo.h>
 #include <tlvf/wfa_map/tlvHigherLayerData.h>
 #include <tlvf/wfa_map/tlvOperatingChannelReport.h>
 #include <tlvf/wfa_map/tlvRadioOperationRestriction.h>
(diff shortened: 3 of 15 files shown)


tomereli (approved):
Some comments I'd like your feedback on before merge BUT nothing serious so approving.

common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -136,4 +136,11 @@ inline el::base::MessageBuilder &operator<<(el::base::MessageBuilder &log, const
     return log << beerocks::net::network_utils::mac_to_string(addr);
 }
 
+inline bool operator<(const sMacAddr &lmac, const sMacAddr &rmac)
+{
+    auto s1 = beerocks::net::network_utils::mac_to_string(lmac);
+    auto s2 = beerocks::net::network_utils::mac_to_string(rmac);
+    return (s1.compare(s2));
+}
```
arnout: This is only needed because you're using a std::set or std::map somewhere, which you shouldn't do.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2115,6 +2115,16 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         std::string client_mac = network_utils::mac_to_string(notification_in->params().mac);
         LOG(INFO) << "client associated sta_mac=" << client_mac;
 
+        auto it = m_vaps_clients_map.find(notification_in->params().bssid);
+        if (it == m_vaps_clients_map.end()) {
+            std::set<sMacAddr> clients;
```
arnout: std::unordered_set

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1933,6 +1933,12 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         std::string client_mac = network_utils::mac_to_string(notification_in->params().mac);
         LOG(INFO) << "client disconnected sta_mac=" << client_mac;
 
+        //update the map to remove this client from its vap
+        auto it = m_vaps_clients_map.find(notification_in->params().bssid);
+        if (it != m_vaps_clients_map.end()) {
```
arnout: We might want to throw a warning if the vap is not found...

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4251,7 +4254,67 @@ bool slave_thread::handle_client_capability_query(Socket *sd, ieee1905_1::CmduMe
...
+    //Check if it is an error scenario - if the STA specified in the Client Capability Query message is not associated
+    //with any of the BSS operated by the Multi-AP Agent (an error scenario)
+    bool error_scenario = 1;
+    auto it             = m_vaps_clients_map.find(client_info_tlv_r->bssid());
+    if (it != m_vaps_clients_map.end()) {
+        if (it->second.count(client_info_tlv_r->client_mac())) {
+            //only if the client exist in one of the BSS operated by the Multi-AP Agent this is not an error scenario
+            error_scenario = 0;
+        }
+    }
```
arnout: This is not accurate... If the client is associated with a BSS on another radio in the same agent, both radios will receive the query, but it is not associated with this radio. We shouldn't treat that as an error.

Also, the bssid given in the client info should be treated as just a hint. The spec says "If the STA specified in the Client Capability Query message is not associated with any of the BSS operated by the Multi-AP Agent". It doesn't say "if it is not associated with the BSS given by the client info TLV". A likely scenario is that a client associates and immediately disassociates and moves to another band, so it's possible that the controller sends the query still using the old BSS (which it got from the topology notification).

Bottom line, I think the best approach is to again cache everything on the backhaul_manager and generate the message there.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4251,7 +4254,67 @@ bool slave_thread::handle_client_capability_query(Socket *sd, ieee1905_1::CmduMe
...
+        LOG(ERROR) << "cmdu creation of type CLIENT_CAPABILITY_REPORT_MESSAGE, has failed";
+        return false;
+    }
+
+    auto client_info_tlv_t = cmdu_tx.addClass<wfa_map::tlvClientInfo>();
+    if (!client_info_tlv_t) {
+        LOG(ERROR) << "addClass wfa_map::tlvClientInfo has failed";
+        return false;
+    }
+    client_info_tlv_t->bssid()      = client_info_tlv_r->bssid();
```
arnout: bssid may have changed already...

For all we know, stupid controllers may just leave it at 00:00:00:00:00:00.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -972,6 +974,14 @@ bool master_thread::handle_cmdu_1905_steering_completed_message(const std::strin
     return son_actions::send_cmdu_to_agent(src_mac, cmdu_tx, database);
 }
 
+bool master_thread::handle_cmdu_1905_client_capability_report_message(
+    const std::string &src_mac, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received CLIENT_CAPABILITY_REPORT_MESSAGE, mid=" << std::hex << int(mid);
```
arnout: Please also log the details so we can check it in test_flows.

arnout (changes requested):
I unfortunately have some fundamental architectural comments:

> Bottom line, I think the best approach is to again cache everything on the backhaul_manager and generate the message there.

To be verified with @tomereli if he agrees.

agent/src/beerocks/slave/son_slave_thread.cpp
CoralMalachi: Thanks for pointing it, I'll change it

tomereli:
> I unfortunately have some fundamental architectural comments:
> 
> > Bottom line, I think the best approach is to again cache everything on the backhaul_manager and generate the message there.
> 
> To be verified with @tomereli if he agrees.

I completely agree and I think I also suggested that in my review, didn't want to block merge on this but on second thought I think you are right, we should spend the extra effort and start using the VAPs map for everything - clients, client capabilities, etc.

controller/src/beerocks/master/son_master_thread.cpp
CoralMalachi: Client Capability Report TLV should include the frame body of the most recently received (Re)Association Request frame from this client (if it is an error scenario this field is omitted).
This part will be done in a separate PR.
So, for now the log message can contain if this is an error scenario or not, and I'll add a TODO comment to add the details later.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -960,6 +963,27 @@ bool master_thread::handle_cmdu_1905_steering_completed_message(const std::strin
...
+        LOG(ERROR) << "getClass wfa_map::tlvClientCapabilityReport has failed";
+        return false;
+    }
+
+    std::string result_code =
+        (client_capability_report_tlv->result_code() == wfa_map::tlvClientCapabilityReport::SUCCESS)
+            ? "SUCCESS"
+            : "FAILURE";
+
+    //TODO: log the details so it can be checked in the test_flows
```
arnout, line 988: :1st_place_medal:

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1814,6 +1818,76 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
...
+    LOG(DEBUG) << "Received CLIENT_CAPABILITY_QUERY_MESSAGE , mid=" << std::dec << int(mid);
+
+    auto client_info_tlv_r = cmdu_rx.getClass<wfa_map::tlvClientInfo>();
+    if (!client_info_tlv_r) {
+        LOG(ERROR) << "getClass wfa_map::tlvClientInfo failed";
+        return false;
+    }
+
+    //Check if it is an error scenario - if the STA specified in the Client Capability Query message is not associated
+    //with any of the BSS operated by the Multi-AP Agent (an error scenario)
```
arnout: It may be useful to explicitly mention that even though client_info_tlv_r does contain a BSSID, the specification says that we should answer if the client is associated with *any* BSS on this agent.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1814,6 +1818,76 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
...
+    bool error_scenario = 1;
```
arnout: Use `true`/`false` for booleans, not 1 and 0.

Also, instead of `error_scenario`, I think it makes more sense to call it e.g. `associated_client_found` (and inverse the logic, obviously).

(nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1814,6 +1818,76 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
...
+    if (!client_capability_report_tlv) {
+        LOG(ERROR) << "addClass wfa_map::tlvClientCapabilityReport has failed";
+        return false;
+    }
+
+    //if it is an error scenario, set Success status to 0x01 = Failure and do nothing after it.
+    if (!error_scenario) {
+        client_capability_report_tlv->result_code() = wfa_map::tlvClientCapabilityReport::SUCCESS;
+        LOG(DEBUG) << "Result Code: SUCCESS";
+        //TODO: Add frame body of the most recently received (Re)Association Request frame from this client
```
arnout, line 1871: So, this...

We have two options here. Either we send a message to the slave that forwards it to the ap_manager that converts it to a bwl call and generates a response message to the slave that gets forwarded to the backhaul manager that generates the response message (with the same MID). Or, we add the ReAssociation Request frame to the CLIENT_ASSOCIATED_NOTIFICATION and cache it in the backhaul manager. The latter would also require us to resend the notification every time a new reassociation request is received by the AP.

If we choose the first option, we have to take into account a possible race condition: by the time the message reaches hostapd, the client may already have left the BSS.

@vitalybu @morantr @tomereli your opinions?

Note that for this PR it doesn't matter much. If it is cached, we'll just need to update the loop in the beginning that finds the associated client to not just set a flag, but to set (a pointer to) the association request frame.

tests/test_flows.sh
```diff
@@ -211,15 +211,31 @@ test_channel_selection() {
 }
 test_client_capability_query() { 
     status "test client capability"
+    sta_mac=00:00:00:11:00:22
 
     check_error=0
-    send_CAPI_1905 ${GATEWAY} $mac_agent1 0x8009 "tlv_type,0x90,tlv_length,\
-0x000C,tlv_value,{$mac_agent1_wlan0 0x000000110022}"
+    check send_CAPI_command ${GATEWAY} "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8009,tlv_type,0x90,tlv_length,\
```
arnout: Please use send_CAPI_1905 function.

tests/test_flows.sh
```diff
@@ -211,15 +211,31 @@ test_channel_selection() {
...
+0x000C,tlv_value,{$mac_agent1_wlan0 0x000000110022}" $redirect
     sleep 1
     dbg "Confirming client capability query has been received on agent"
     # check that both radio agents received it,in the future we'll add a check to verify which radio the query was intended for.
-    check_log ${REPEATER1} agent_wlan0 "CLIENT_CAPABILITY_QUERY_MESSAGE"
-    check_log ${REPEATER1} agent_wlan2 "CLIENT_CAPABILITY_QUERY_MESSAGE"
+    check_log ${REPEATER1} agent "CLIENT_CAPABILITY_QUERY_MESSAGE"
+
+    dbg "Confirming that Client Capability Report message contains an Error Code TLV"
+    check_log ${REPEATER1} agent "Result Code: FAILURE"
```
arnout: It would be better to check this on the controller instead of the agent. The commit that adds the handler on the controller could very easily parse out the result code and report it.

Also, to make sure the correct log message is checked, add the client MAC and BSSID to the output.

tests/test_flows.sh
```diff
@@ -211,15 +211,31 @@ test_channel_selection() {
...
+
+    dbg "Connect dummy STA to wlan0"
+    send_bwl_event ${REPEATER1} wlan0 "EVENT AP-STA-CONNECTED ${sta_mac}"
+
+    check send_CAPI_command ${GATEWAY} "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8009,tlv_type,0x90,tlv_length,\
```
arnout: Again, send_CAPI_1905 please.

tests/test_flows.sh
```diff
@@ -211,15 +211,31 @@ test_channel_selection() {
...
+0x000C,tlv_value,{$mac_agent1_wlan0 0x000000110022}" $redirect
+    sleep 1
+
+    check_log ${REPEATER1} agent "Result Code: SUCCESS"
```
arnout: Same here, better to check it on the controller.

tests/test_flows.sh
```diff
@@ -211,15 +211,31 @@ test_channel_selection() {
 }
 test_client_capability_query() { 
     status "test client capability"
+    sta_mac=00:00:00:11:00:22
 
     check_error=0
-    send_CAPI_1905 ${GATEWAY} $mac_agent1 0x8009 "tlv_type,0x90,tlv_length,\
-0x000C,tlv_value,{$mac_agent1_wlan0 0x000000110022}"
+    check send_CAPI_command ${GATEWAY} "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8009,tlv_type,0x90,tlv_length,\
+0x000C,tlv_value,{$mac_agent1_wlan0 0x000000110022}" $redirect
```
arnout: Please use a *different* sta_mac here than the one for the success scenario. This helps to make sure we parse out the correct log message. Cfr. #822

tests/test_flows.sh
```diff
@@ -211,15 +211,31 @@ test_channel_selection() {
...
+    check_log ${REPEATER1} agent "CLIENT_CAPABILITY_QUERY_MESSAGE"
+
+    dbg "Confirming that Client Capability Report message contains an Error Code TLV"
+    check_log ${REPEATER1} agent "Result Code: FAILURE"
+
+    dbg "Connect dummy STA to wlan0"
+    send_bwl_event ${REPEATER1} wlan0 "EVENT AP-STA-CONNECTED ${sta_mac}"
+
+    check send_CAPI_command ${GATEWAY} "DEV_SEND_1905,DestALid,$mac_agent1,MessageTypeValue,0x8009,tlv_type,0x90,tlv_length,\
+0x000C,tlv_value,{$mac_agent1_wlan0 0x000000110022}" $redirect
```
arnout: You can actually use `${sta_mac}` here - our CAPI parser accepts it.

arnout (approved):
*Very* good splitting in commits and good commit messages. :1st_place_medal: 

Of course I have a few nitpicks, but we can go through with merging this.

There is one major question remaining about how it should be implemented.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: Agree.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: I prefer option (2) - caching. I think it is cleaner.
However, the race condition may occur in both options since the client may have left also by the time the message with the updated ReAssociation frame is received in the BH manager

CoralMalachi: I also prefer the second option.

Issue #746, opened 2020-01-27
[TASK] Run agents tests nightly in CI


rmelotte:
## Description

As soon as a new agent test is passing on the rax40, it could already be added to the nigthly schedule to be able to spot any regression. 

This task depends on #741 .

## Exit criteria

- [ ] Test at least 4.4.1 with the rax40 manually.

- [ ] Once the test is passing, add a nightly schedule with agent tests, only 4.4.1 at first.

ghost:
Blocked by #748 and #741

rmelotte:
Done, the schedule name is `Nightly certification tests - agent - rax40`: https://gitlab.com/prpl-foundation/prplMesh/pipeline_schedules

PR #748, opened 2020-01-27
.gitlab-ci.yml: add support for other DUTs

### LINKED ISSUE #502: [TASK] Allow running any combination of tests and uploading their result
## Description

We need to be able to easily run any (combination of) tests, and make it easy to upload them (as part of Gitlab artifacts for example). This task consist of adding the necessary scripting to:

- be able to either run a chosen subset of tests, or all the test that are part of the certification
- restart every agent as well as prplmesh between each test
- separate the tests results between failed and passed tests to make it easy to tell them apart
- group the test logs with all the logs that can be gathered (pcap files, sigma agents logs, prplmesh logs, etc) into different subfolders to make them easy to upload
- make it easy to add a new DUT
- make it easy to choose/change the DUT to use for the tests

rmelotte:
Add support for other devices by using the new '-d' option of
run_test_file.sh.

The currently accepted devices are:
"prplmesh": prplmesh with dummy BWL.
"prplmesh-rax40": prplmesh on the rax40 with DWPAL BWL.

By default, "prplmesh" is used.

To be able to use the current prplMesh version on the rax40, it also
needs to be deployed first, so add a call to tools/deploy.sh before
running the tests.

close #502 .

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -162,14 +162,22 @@ test-on-netgear-rax40:
   variables:
     # TESTS_TO_RUN needs to be set by the user (or the pipeline schedule)
     GIT_CLONE_PATH: "/builds/prpl-foundation/prplMesh/"
+    # device to test with: prplmesh for dummy bwl, prplmesh-rax40 for dwpal on rax40
+    DEVICE_UNDER_TEST: prplmesh
   script:
-      - /easymesh_cert/run_test_file.sh -o logs $TESTS_TO_RUN
+      - |
+        if [ "$DEVICE_UNDER_TEST" != prplmesh ] ; then
+            echo "Deploying to $DEVICE_UNDER_TEST"
+            tools/deploy_ipk.sh $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/*.ipk
+        fi
+      - /easymesh_cert/run_test_file.sh -o logs -d $DEVICE_UNDER_TEST $TESTS_TO_RUN
   artifacts:
     paths:
       - logs
     when: always
   needs:
     - job: build-in-docker
+    - job: build-for-netgear-rax40
   tags:
     - certs-tests
   timeout: 24h


tomereli (approved):

.gitlab-ci.yml
```diff
@@ -153,14 +153,17 @@ test-on-netgear-rax40:
   variables:
     # TESTS_TO_RUN needs to be set by the user (or the pipeline schedule)
     GIT_CLONE_PATH: "/builds/prpl-foundation/prplMesh/"
+    DEVICE_UNDER_TEST: prplmesh
```
arnout, line 166: Maybe add a comment about this variable as well.

.gitlab-ci.yml
```diff
@@ -153,14 +153,17 @@ test-on-netgear-rax40:
...
   script:
-      - /easymesh_cert/run_test_file.sh -o logs $TESTS_TO_RUN
+      - tools/deploy_ipk.sh prplmesh-rax40 tools/docker/builder/openwrt/*.ipk
```
arnout: Shouldnt this be
```suggestion
      - tools/deploy_ipk.sh $DEVICE_UNDER_TEST tools/docker/builder/openwrt/*.ipk
```

.gitlab-ci.yml
```diff
@@ -153,14 +153,17 @@ test-on-netgear-rax40:
...
+      - /easymesh_cert/run_test_file.sh -o logs -d $DEVICE_UNDER_TEST $TESTS_TO_RUN
   artifacts:
     paths:
       - logs
     when: always
   needs:
     - job: build-in-docker
+    - job: build-for-netgear-rax40
```
arnout, line 180: I'm thinking it would be better to have a separate job per device. Something like:

```
.run-certification-tests:
  stage: test
  variables:
    # TESTS_TO_RUN needs to be set by the user (or the pipeline schedule)
    # DEVICE_UNDER_TEST needs to be set by the user or pipeline schedule
    GIT_CLONE_PATH: "/builds/prpl-foundation/prplMesh/"
  script:
      - tools/deploy_ipk.sh $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/*.ipk
      - /easymesh_cert/run_test_file.sh -o logs -d $DEVICE_UNDER_TEST $TESTS_TO_RUN
  artifacts:
    paths:
      - logs
    when: always
  tags:
    - certs-tests
  timeout: 24h
  rules:
    - if: '$TESTS_TO_RUN =~ /^$/'
       when: manual
    - if: '$SOME_VARIABLE_SET_BY_SCHEDULE = ON'

certification-tests-ubuntu:
  extends: .run-certification-tests
  variables:
    DEVICE_UNDER_TEST: prplmesh
  needs:
    - job: build-in-docker

certification-tests-rax40:
  extends: .run-certification-tests
  variables:
    DEVICE_UNDER_TEST: netgear-rax40
  needs:
    - job: build-for-netgear-rax40
```

It would be even better if we could do `job: build-for-$DEVICE_UNDER_TEST` but I don't know if that's possible.

The rules syntax is also a bit fuzzy so I'm not sure if it will work like that.

arnout (approved):

.gitlab-ci.yml
rmelotte: I really love the idea of combining the scheduled and manual jobs!
Looking for TESTS_TO_RUN value is genius. I believe the schedules set it _before_ the rules are evaluated, while when we want to run them manually the rules have already been evaluated (in other words, it should do what we want).

It also makes sense to have 2 separate jobs for the 2 different devices, but as it is, I think it will always run the tests for both devices when 1 schedule is run, so we miss the possibility to run only for one DUT on schedule.

We need to find a way for the schedules to specify a list of DUTs to run with while keeping 1 manual job per device like here.

.gitlab-ci.yml
rmelotte: No, because if the DEVICE_UNDER_TEST is 'prplmesh', we don't need to deploy anything.
I agree we could use a conditional and deploy to DEVICE_UNDER_TEST in all other cases though.

.gitlab-ci.yml
rmelotte: I tried something on `dev/ci-multiple-duts-merge-scheduled-manual`, but for some reason Gitlab marks the pipeline as "blocked" when there are manual, not triggered, jobs and no longer as passing: https://gitlab.com/prpl-foundation/prplMesh/pipelines/113248562 

Also if we want to split the certification tests between smaller groups, I'm starting to think that it would be simpler to just generate a yaml.

I'll make the other suggested change and keep these in a separate issue so that we can already merge this one and unblock #746 .

.gitlab-ci.yml

.gitlab-ci.yml
rmelotte: Done.

.gitlab-ci.yml
rmelotte: Opened https://github.com/prplfoundation/prplMesh/issues/759 to track this.

PR #749, opened 2020-01-27
gitlab-ci.yml: create a persistent docker image tag on master

### LINKED ISSUE #744: [TASK] CI: add a single docker image tag for each target on master
## Description

Since we now remove the unique tag automatically in CI, docker removes the last image entirely (since the image has no other tags).
By adding a single (non-unique) tag per target if tests pass in CI on master, the latest image would be kept.

## Exit criteria

- [ ] If build pass in CI on master, tag the image with a non-unique tag (one tag per target).

rmelotte:
Since commit 36c676aafd6fd5cf0a8b76e930c0d430f6df357a, we're always
deleting the unique docker image tag we create. Because no other tags
use the same image, it actually also gets deleted from the cache.
Because of this, the next pipeline will have to rebuild the entirety
of the last stage.

It was not a problem until now because the shell runner we had did
keep tags from back when we didn't delete them automatically. Now that
we want to add other runners however, the issue shows up.

If we're on the master branch, add a docker image tag per target
device, so that at least one image is kept in cache (the latest one
from master).

close #744

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -87,6 +87,13 @@ run-tests:
     - cd tools/docker/builder/openwrt
     - ./build.sh -v -d "$TARGET_DEVICE" -t "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee build.log | grep '^make\[[12]\]\|^Step\|^ --->' --color=never --line-buffered
   after_script:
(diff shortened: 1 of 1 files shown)


tomereli (approved):

arnout (approved):
IIU, however, it means that if you change something in the Dockerfile, the builds on the corresponding PR are not going to use the cache at all until it's merged in master, right?

It would be nicer if we could tell docker to keep images cached for some time (couple of days)...

PR #751, opened 2020-01-28
easylogging++: bug - access attempt to released memory

adam1985d:
On the UGW platform, a SIGSEGV event is sent to the APM application
that utilizes the easylogging++ library. The event is sent due to an
access attempt to a released memory. The released memory is a pointer
to the "LogStreamsReferenceMap m_logStreamsReference" which is defined
statically in RegisteredLoggers class.

The RegisteredLoggers object is released in UGW before the release of
other Class objects that make use of the pointer, and thus - the pointer
is accessed after the map is already released.

The issue may be seen in other platforms and is dependent on the difference in object release order.

After this change, the map is defined as a shared-pointer in the
RegisteredLoggers class and is set to a shared-pointer of type
LogStreamsReferenceMap in the RegisteredLoggers CTOR. Even if the
RegisteredLoggers object is released, the map resource will be
available to the rest of the objects until all are destructed.

### DIFF
--- a/framework/external/easylogging/easylogging++.cc
+++ b/framework/external/easylogging/easylogging++.cc
@@ -603,7 +603,7 @@ void LogBuilder::convertToColoredOutput(base::type::string_t* logLine, Level lev
 
 // Logger
 
-Logger::Logger(const std::string& id, base::LogStreamsReferenceMap* logStreamsReference) :
+Logger::Logger(const std::string& id, base::LogStreamsReferenceMapPtr logStreamsReference) :
   m_id(id),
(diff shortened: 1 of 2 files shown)


tomereli (approved):

morantr (approved):

arnout:
Could you also make a PR upstream at https://github.com/amrayn/easyloggingpp ?

adam1985d:
> Could you also make a PR upstream at https://github.com/amrayn/easyloggingpp ?

done:
https://github.com/amrayn/easyloggingpp/pull/739

PR #752, opened 2020-01-28
Agent UCC listener: support BSSID option

CoralMalachi:
### Description

currently, test 4.6.2 keeps on failing on step 9, which is the topology response message, with the following error message:

image

The reason for this error is that our dev_get_parameter does not support a `bssid` as an option.
Therefore, since the `bssid` is equal to the MAC address, add an or logic to support this option exactly the same way we do with the `macaddr` option.

### Testbed status
passed this step.

### DIFF
--- a/agent/src/beerocks/slave/agent_ucc_listener.cpp
+++ b/agent/src/beerocks/slave/agent_ucc_listener.cpp
@@ -109,7 +109,7 @@ void agent_ucc_listener::update_vaps_list(std::string ruid, beerocks_message::sV
 bool agent_ucc_listener::handle_dev_get_param(std::unordered_map<std::string, std::string> &params,
                                               std::string &value)
 {
-    if (params["parameter"] == "macaddr") {
+    if (params["parameter"] == "macaddr" || params["parameter"] == "bssid") {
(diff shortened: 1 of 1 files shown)


tomereli (approved):

rmelotte:
The explanations should be added to the commit message.

Also why are we using "mac" for the bssid? Shouldn't we rename it to bssid?
I don't understand why there is a distinction between the two here. Is the UCC requesting macaddr in some places and bssid in others when it actually wants the same thing?

arnout:
> Also why are we using "mac" for the bssid? Shouldn't we rename it to bssid?

This is in fact a workaround. What should be done is to look through the vaps list for a vap matching the requested ssid, and returning the bssid of that vap.

As a quick fix, however, we can just take the mac address of the radio, because in practice we use that as the bssid of the first vap (and most configurations have just one vap).

So in a final solution the code paths for macaddr and for bssid would be completely different.

rmelotte (approved):
I thought it was approved and merged already, I just realized it's not the case.

rmelotte:
I updated the commit message with the PR description and @arnout 's comments.
Will merge as soon as CI has passed.

PR #753, opened 2020-01-28
Switch RAX40 build to use GCC8.3

tomereli:
Switch to build RAX40 with GCC8.3.0 instead of default GCC7.4.0, as done
in prplWrt forks in the following commits:
https://git.prpl.dev/prplmesh/feed-intel/commit/9cf99c890264f64b3f1b5b2ed38fa817b9ced37a
https://git.prpl.dev/prplmesh/iwlwav/commit/80be7244f6977f07b647982e7bf438ba9f5cf810

For that, add a build argument SDK_CONFIG, sdk_default_config and
sdk_gcc8_config which are used as follows:
- RAX40 build is using GCC8
- Other targets are using default, which is GCC7 so no change is needed
in the file.

The Dockerfile simply copies the relevant config file to the image, and
then uses it when building the SDK.


tomereli:
https://git.prpl.dev/prplmesh/feed-intel/commit/9cf99c890264f64b3f1b5b2ed38fa817b9ced37a

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -44,12 +44,17 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
 
 WORKDIR /home/openwrt/openwrt
 
+COPY --chown=openwrt:openwrt configs/* /home/openwrt/
```
rmelotte: nack.
Avoid using globs at all cost when using COPY, because any change in the copied file will trigger a new build from here.
If you have any file you don't want lying in `configs`, it's also going to be included in the image.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -44,12 +44,18 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
...
+
 RUN if [ -n "$INTEL_FEED" ] ; then \
 	echo "src-git intel $INTEL_FEED" >> feeds.conf.default \
 	&& scripts/feeds update intel \
 	&& scripts/feeds install intel_mips ; \
+    fi ;
+    
+RUN if [ -e /home/openwrt/${TARGET}_config ]; then \
+        cp /home/openwrt/${TARGET}_config .config ;\
```
rmelotte: After thinking about it, I think it might be more useful to split it between different SDK configuration rather than between targets.
What I mean is that we could have a file say sdk_default which would be the configuration we used until now, and another one called sdk_gcc8 for example, so that any target could easily be switched between different GCC versions (it could even be integrated later as an option in the builds script, although that's not mandatory).

If you anyway want to go for the per-target approach, you still need an argument to decide which file to copy, so you could remove the if statement entirely (and copy the corresponding config file in any case).

rmelotte (changes requested):

tools/docker/builder/openwrt/Dockerfile
tomereli: Done.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -44,12 +46,16 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
...
-	&& scripts/feeds update intel \
-	&& scripts/feeds install intel_mips ; \
-    fi ; \
-    echo "CONFIG_SDK=y" > .config \
+        echo "src-git intel $INTEL_FEED" >> feeds.conf.default \
+        && scripts/feeds update intel \
+        && scripts/feeds install intel_mips ; \
+    fi ;\
+    cp /home/openwrt/sdk_config .config \
+    && cat .config \
```
rmelotte: If we're going to add this `cat .config`, I think it would be most useful right after the `make defconfig`, because then you know if some options were unused. The content of the copied one is easier to guess. 

(nitpick)

rmelotte (approved):
Looks good!
We just have to wait for the CI to finish, and to double check GCC 8 was actually used.

Also, I noticed GCC 9 is already available in OpenWrt, maybe we could build another target with it? What do you think?

tools/docker/builder/openwrt/Dockerfile
tomereli: but if we move it to after the `defconfig` won't it include *all* the configuration options? I wanted to avoid that and only show the options we set by default.

tomereli:
> Looks good!
> We just have to wait for the CI to finish, and to double check GCC 8 was actually used.
> 
> Also, I noticed GCC 9 is already available in OpenWrt, maybe we could build another target with it? What do you think?

GCC8 was used and its the reason why RAX40 failed for me locally when I built that. Now I added a commit which updates the iwlwav hash to the one which fixed GCC8.3 issues :)

Regarding GCC9 - sounds good, do we have it in prplWrt? If so, lets use it for Turris Omnia build. Then we cover all of them.

tools/docker/builder/openwrt/Dockerfile
rmelotte: It will. It allows to check if the options you specified were actually used (they won't be kept in the file if they were unused for any reason).
The reasoning is that you already know what the defaults are for a specific prplMesh hash (by looking at the content of the file in the repository for that hash), but it's harder to know what the resulting SDK configuration will be (it's not kept in the image).
It's not that important though, it's mostly useful when debugging the build of the image itself.

rmelotte:
Ready for review.

rmelotte:
CI now passes. 
However:
- pushing to the branch or merging to master will require building the full image again (~80 minutes on Mind2 runner).
- Runners `prplmesh-ci-1-intel` and `prplmesh-ci-1-intel-bash` are offline for some time, so only 1 OpenWrt build at a time is possible at the moment.

tomereli:
Can we merge it now?

rmelotte:
We can merge it at any time yes (I can't merge it because it needs 1 more review).

Issue #756, opened 2020-01-29
[BUG] Disable UCC Listener on Agent by UCI configuration


morantr:
## Description

UCC Listeners use port 8001 and 8002 to communicate between the UCC and prplMesh.
UGW uses these ports for internal operation and when the UCC listener tries to use this port as well, it causes bus errors.

To solve it, we need to disable certification mode on UGW on the UCI.
The problem is that certification mode defined in the UCI, is not used in order to run the UCC Listener on the Agent, what is used is the UCC Listener port which defined on the Controller/Agent configuration file.

## Solution Options
1. Remove "Certification mode" from the database, and replace it with UCC Listener Port.
2. Make the UCC Listener run on the Agent only if the  "Certification mode" is enabled on the UCI database.

Personally, I prefer solution 1, since "Certification mode" flag is redundant, and it is enough to read the Port number, so If the port is 0, then certification mode is off.

tomereli:
Intel JIRA - https://jira-chd.intel.com/browse/LMTSYS-2688

tomereli:
Solution is to do this in both controller and agent.

## Controller 
Create and start the UCC listener only after receiving platform settings *if* certification mode enabled

## Agent
Add platform settings to backhaul manager and start the listener only after receiving platform settings *if* certification mode enabled

Please split this issue to 2 PRs - one for the agent, one for the controller - start with the controller.

tomereli:
Fixed.

PR #757, opened 2020-01-29
controller: Run UCC Listener if certification mode is set

morantr:
Currently, the UCC Listener has been initiated on the son_master init function.
This has caused to the UCC Listener to bind to the port from the configuration
file.
The problem is that on some platforms, the port that we are using is used by other
applications, and interfere with them.

Move the UCC Listener initiation to slave_join. It will start only once
if not has been started before.

#756

### DIFF
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -142,15 +142,6 @@ bool master_thread::init()
         LOG(DEBUG) << "Health check is DISABLED!";
     }
 
-    if (database.config.ucc_listener_port != 0) {
-        m_controller_ucc_listener =
-            std::make_unique<controller_ucc_listener>(database, cert_cmdu_tx);
-        if (m_controller_ucc_listener && !m_controller_ucc_listener->start("ucc_listener")) {
-            LOG(FATAL) << "failed start controller_ucc_listener";
-            return false;
-        }
-    }
-
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


tomereli (approved):

vitalybu (approved):
Approving, although this commit doesn't really solve the described problem.
Now the UCC listener will start later, but can still bind to a used port...
Also, if the UCC listener fails to start, the slave_join process is aborted (unless that's the desired behavior for certification mode).

tomereli:
> Approving, although this commit doesn't really solve the described problem.
> Now the UCC listener will start later, but can still bind to a used port...
> Also, if the UCC listener fails to start, the slave_join process is aborted (unless that's the desired behavior for certification mode).

It will be accompanied with a change in the UGW feed - https://gts-chd.intel.com/projects/SW_WAVE/repos/feed_wlan_6x/pull-requests/210/overview

Edit - the change in the UGW feed is to change `prplmesh.config.certification_mode=0` in the under `files/etc/config/prplmesh`. This means that UGW only is not setting certification mode by default.
The same will be done in the RDKB yocto recipe - @morantr please make sure this happens.

tomereli:
Merged since it passed all tests and part of the controller certification tests. Since it is enough to pass a single test in the testbed to verify this change is OK, and we don't have 2 hours to wait for it, then I merged it.

arnout:
I'm sorry, but this is not OK. As @vitalybu notes, this change doesn't really solve the issue described in the commit message. So either the commit message is wrong, or you're just avoiding the problem and it will bite us again later.

If the problem is indeed a conflicting port, then just change the port number! It is read from the config file so you can easily change it locally, or for all I care you can change it for everybody (slightly more tricky because the config file in the certification setup also has to be updated, but not that big a deal).

tomereli:
> I'm sorry, but this is not OK. As @vitalybu notes, this change doesn't really solve the issue described in the commit message. So either the commit message is wrong, or you're just avoiding the problem and it will bite us again later.
> 

You are absolutely right in one thing - the commit message is wrong. It should have said something like "There is currently a bug that even if certification mode is disabled, the UCC listener is still brought up, which is both a security risk and may cause ports clashing with other critical applications. Therefore, this commit changes the boot sequence of the controller to only start the UCC listener after it has received the platform settings and certification mode is indeed enabled in /etc/config/prplmesh"

> If the problem is indeed a conflicting port, then just change the port number! It is read from the config file so you can easily change it locally, or for all, I care you can change it for everybody (slightly more tricky because the config file in the certification setup also has to be updated, but not that big a deal).

As I explained above, this is not true. The conflicting port is just a symptom of a much more critical issue which is the fact we are running certification mode in a real product. Eventually, we should disable certification mode completely somehow, but this anyway should have been done in a separate PR.

Bottom line - there was a review for this PR by both Vitaly and myself. We both didn't pay enough attention for the commit description, will do in the future.

PR #758, opened 2020-01-29
Hotfix/agent run ucc listener on backhaul register handler

morantr:
Currently, the UCC Listener has been initiated on the backhaul_manager  init function.
This has caused to the UCC Listener to bind to the port from the configuration
file.
The problem is that on some platforms, the port that we are using is used by other
applications, and interfere with them.

Move the UCC Listener initiation to son_slave register handler.
It will start only once if not has been started before.

#756

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -124,16 +124,6 @@ bool backhaul_manager::init()
         return false;
     }
 
-    if (m_sConfig.ucc_listener_port != 0) {
-        m_agent_ucc_listener = std::make_unique<agent_ucc_listener>(
-            *this, m_sConfig.ucc_listener_port, m_sConfig.vendor, m_sConfig.model,
-            m_sConfig.bridge_iface, cert_cmdu_tx);
-        if (m_agent_ucc_listener && !m_agent_ucc_listener->start("ucc_listener")) {
-            LOG(ERROR) << "failed start agent_ucc_listener";
-            return false;
-        }
-    }
-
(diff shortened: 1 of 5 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1467,6 +1457,16 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
 
+        if (!m_agent_ucc_listener && request->certification_mode() && m_sConfig.ucc_listener_port != 0) {
+            m_agent_ucc_listener = std::make_unique<agent_ucc_listener>(
+                *this, m_sConfig.ucc_listener_port, m_sConfig.vendor, m_sConfig.model,
+                m_sConfig.bridge_iface, cert_cmdu_tx);
+            if (m_agent_ucc_listener && !m_agent_ucc_listener->start("ucc_listener")) {
+                LOG(ERROR) << "failed start agent_ucc_listener";
+                return false;
+            }
+        }
```
vitalybu, line 1468: What problem is solved by starting the UCC listener from `handle_slave_backhaul_message` instead of `init`?

vitalybu (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: That on the init function we don't have certification mode enable flag from UCI.
So if we run UCC Listener we interfere with other applications on the platform that are using that port.

tomereli (approved):

PR #765, opened 2020-01-30
platform_manager:work_queue - add delay between timer checks

### LINKED ISSUE #764: [BUG] beerocks_agent: beerocks agent takes 48% CPU load if stuck in stage of trying to read wifi settings from BPL
**Describe the bug** -
In the case where the platform manager fails to read wifi settings, there is configured a wait time between retries - but there is no delay between checks if the wait-period has passed.
This results in a check if the wait-period has passed every 1 millisecond, causing a CPU usage of ~48% as seen in the attached screen-shot:
beerocks_agent high cpu load

Note that once the platform manages to successfully read the wifi settings, and prplmesh reaches the operational state, the CPU load is back to ~0.
- Observed on the UGW platform (Axepoint), but reproducible on any platform.

## Steps to reproduce
reproduced on the latest UGW version after each factory-reset, due to continuous failure to read wifi settings (will be handled in a separate PR and will be updated here if the investigation leads to the  prplmesh as the source of the issue)

adam1985d:
Before the change, in the case where the platform manager fails to read
wifi settings, there is configured a wait time between retries - but
there is no delay between checks if the wait-period has passed.
This results in a check if the wait-period has passed every one
millisecond, causing a CPU usage of ~48%.
Note that once the platform manages to successfully read the wifi settings,
and prplmesh reaches the operational state, the CPU load is
back to ~0.

After the change, on failure to read the wifi-settings, sleep for the configured delay between reads is performed and even if beerocks_agent is stuck in the stage of reading wifi settings, the CPU load is ~0%.

Fixes #764

### DIFF
--- a/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
+++ b/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
@@ -782,26 +782,24 @@ bool main_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
                 return;
             }
 
-            std::chrono::steady_clock::time_point retry_timer = std::chrono::steady_clock::now();
-            uint8_t retry_cnt                                 = 0;
-
+            uint8_t retry_cnt          = 0;
             register_response->valid() = 0;
             do {
-                if (std::chrono::steady_clock::now() > retry_timer) {
-                    LOG(TRACE) << "Trying to read settings of iface:" << strIfaceName
-                               << ", attempt=" << int(retry_cnt);
-                    if (fill_platform_settings(strIfaceName, register_response,
-                                               platform_common_conf, bpl_iface_wlan_params_map,
-                                               sd)) {
-                        register_response->valid() = 1;
-                    } else {
-                        LOG(INFO) << "Reading settings of iface:" << strIfaceName
-                                  << ", attempt=" << int(retry_cnt) << " has failed!";
-                        if (++retry_cnt == PLATFORM_READ_CONF_MAX_ATTEMPTS)
-                            break;
-                        retry_timer = std::chrono::steady_clock::now() +
-                                      std::chrono::seconds(PLATFORM_READ_CONF_RETRY_SEC);
-                    }
+                LOG(TRACE) << "Trying to read settings of iface:" << strIfaceName
+                           << ", attempt=" << int(retry_cnt);
+                if (fill_platform_settings(strIfaceName, register_response, platform_common_conf,
+                                           bpl_iface_wlan_params_map, sd)) {
+                    register_response->valid() = 1;
+                } else {
+                    LOG(INFO) << "Reading settings of iface:" << strIfaceName
+                              << ", attempt=" << int(retry_cnt) << " has failed!";
+
+                    // reached max number of retries
+                    if (++retry_cnt == PLATFORM_READ_CONF_MAX_ATTEMPTS)
+                        break;
+
+                    // sleep and retry
+                    std::this_thread::sleep_for(std::chrono::seconds(PLATFORM_READ_CONF_RETRY_SEC));
                 }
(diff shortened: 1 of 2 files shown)


agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -802,6 +802,8 @@ bool main_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
                         retry_timer = std::chrono::steady_clock::now() +
                                       std::chrono::seconds(PLATFORM_READ_CONF_RETRY_SEC);
                     }
+                } else {
+                    usleep(PLATFORM_READ_CONF_SLEEP_USEC);
```
vitalybu: There's no need to add a new sleep value.
On line 802 there's a `retry_timer` based on the current time and the `PLATFORM_READ_CONF_RETRY_SEC` define.
You can simply [sleep_until](https://en.cppreference.com/w/cpp/thread/sleep_until) the specified time before retrying.

I also suggest reducing the timeout value to 5 seconds (10 is quite a large value).

vitalybu (changes requested):

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
adam1985d: I considered this approach but wasn't sure if the possible delay of 5 seconds in the shutdown (when performed) is something we wish to avoid.

Fixed:
- wait on `PLATFORM_READ_CONF_RETRY_SEC` as suggested
- reduce it to 5 seconds.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -803,7 +803,7 @@ bool main_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
                                       std::chrono::seconds(PLATFORM_READ_CONF_RETRY_SEC);
                     }
                 } else {
-                    usleep(PLATFORM_READ_CONF_SLEEP_USEC);
+                    sleep(PLATFORM_READ_CONF_RETRY_SEC);
```
vitalybu: Dude... You're still not using the `retry_timer` variable.
Anyway, this entire loop can be simplified:
```
register_response->valid() = 0;
do {
	LOG(TRACE) << "Trying to read settings of iface:" << strIfaceName
			   << ", attempt=" << int(retry_cnt);

	if (fill_platform_settings(strIfaceName, register_response,
							   platform_common_conf, bpl_iface_wlan_params_map,
							   sd)) {
		register_response->valid() = 1;
	} else {
		LOG(INFO) << "Reading settings of iface:" << strIfaceName
				  << ", attempt=" << int(retry_cnt) << " has failed!";
		
		// Reached the maximal number of retries, stop trying...
		if (++retry_cnt == PLATFORM_READ_CONF_MAX_ATTEMPTS)
			break;

		// Sleep and retry
		std::this_thread::sleep_for(std::chrono::seconds(PLATFORM_READ_CONF_RETRY_SEC));
	}
	
} while (!register_response->valid() && !should_stop);
```

vitalybu (changes requested):

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
adam1985d: ok - refactored the loop to what you suggested

vitalybu (approved):
Looks good. Cleanup the PR (rebase) and update the description for other reviewers.

rmelotte (approved):
Please don't forget to squash before merging.

adam1985d:
> Looks good. Cleanup the PR (rebase) and update the description for other reviewers.

done

PR #766, opened 2020-01-30
Hotfix/deploy enable certification mode

rmelotte:
With commit dd9d7fe6c8fc4f2f8ce1e1527b2186fb83d75e79, the UCC listener
won't run on the controller unless certification_mode is set to 1.

Since we don't want the certification mode to default to 1, it makes
sense to be able to enable it on the target right after deploying (to
be able to run the certification tests from CI for example).

Update the deploy script to have a '--certification-mode' option to enable certification mode on the target.
Use the new option in CI.


tomereli (approved):
I basically already reviewed this in another PR and approved it, so please take my comments from there :)

Issue #741, opened 2020-01-27
[TASK] Use a single port for the UCC listener


rmelotte:
## Description

Currently the agent listens on port 8002, and the controller on port 8001.

For certification, it makes things harder, as we need 2 different DUTs: one for prplmesh running as an agent (for MAUT tests) and one for prplMesh running as a controller (for MCUT tests).

## Exit criteria

prplMesh always listens on port 8001.
If it's run in agent mode, the agent listens on port 8001 (MAUT tests).
If it's run in agent+controller mode, the controller listens on port 8001. (MCUT tests).

PR #768, opened 2020-01-30
Feature/ucc listener single port

rmelotte:
Use the same port for the UCC listener on both agent and controller.

Also include the changes from https://github.com/prplfoundation/prplMesh/pull/766 , because they are both required to run agent tests on rax40.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -39,6 +39,7 @@
 
 // BPL Error Codes
 #include <bpl/bpl_err.h>
+#include <bpl/bpl_cfg.h>
 
 using namespace beerocks::net;
 
@@ -1458,7 +1459,7 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
         m_sConfig.slave_iface_socket[soc->sta_iface] = soc;
 
         if (!m_agent_ucc_listener && request->certification_mode() &&
-            m_sConfig.ucc_listener_port != 0) {
+            m_sConfig.ucc_listener_port != 0 && !bpl::cfg_is_master()) {
             m_agent_ucc_listener = std::make_unique<agent_ucc_listener>(
                 *this, m_sConfig.ucc_listener_port, m_sConfig.vendor, m_sConfig.model,
                 m_sConfig.bridge_iface, cert_cmdu_tx);
--- a/controller/config/beerocks_controller.conf.in
+++ b/controller/config/beerocks_controller.conf.in
@@ -10,7 +10,7 @@ temp_path=@TMP_PATH@
 
 vendor=Intel
 model=prplMesh
-ucc_listener_port=8001 # 0 - disabled
+ucc_listener_port=8002 # 0 - disabled
 
 # Features:
 #   DFS reentry feature:


rmelotte:
Close #741 .

tools/deploy_ipk.sh
```diff
@@ -28,4 +36,33 @@ rm -rf /opt/prplmesh
...
+main() {
+        OPTS=$(getopt -o 'h' --long help,certification-mode -n 'parse-options' -- "$@")
+
+    if [ "$?" != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
+
+    eval set -- "$OPTS"
+
+    while true; do
+        case "$1" in
+            --certification-mode)
```
tomereli: why not add also short option?

tools/deploy_ipk.sh
```diff
@@ -1,23 +1,31 @@
...
+    echo "      -h|--help - show this help menu"
+    echo "      --certification-mode - enable certification-mode on the target"
+}
 
-if ! ssh "$SSH_OPTIONS" "$1" /bin/true ; then
-    echo "Error: $1 unreachable via ssh"
-    exit 1
-fi
+deploy() {
+    SSH_OPTIONS="-oBatchMode=yes"
```
tomereli: I don't see SSH_OPTIONS used below. BTW what does `BatchMode` do?

tools/deploy_ipk.sh
```diff
@@ -1,23 +1,31 @@
...
-echo "Copying $IPK to $TARGET:$DEST_FOLDER/$IPK_FILENAME"
-scp "$IPK" "$TARGET:$DEST_FOLDER/$IPK_FILENAME"
+    TARGET="$1"
+    IPK="$2"
+    IPK_FILENAME="$(basename "$2")"
+    DEST_FOLDER=/tmp/prplmesh_ipks
 
-ssh "$TARGET" <<EOF
+    echo "Removing previous ipks"
+    ssh "$TARGET" "rm -rf \"$DEST_FOLDER\" ; mkdir -p \"$DEST_FOLDER\""
```
tomereli: No SSHOPTIONS here?

tools/deploy_ipk.sh
```diff
@@ -1,23 +1,31 @@
 #!/bin/sh -e
 
-SSH_OPTIONS="-oBatchMode=yes"
+usage() {
```
tomereli: Regarding the commit message - I agree that in a product we should disable certification mode completely, and better to do that in compile time (we can simply make the certification getter/setter a NOP).
But for the time being, I think that RAX40 should have certification_mode enabled by default.

.gitlab-ci.yml
```diff
@@ -168,7 +168,7 @@ test-on-netgear-rax40:
       - |
         if [ "$DEVICE_UNDER_TEST" != prplmesh ] ; then
             echo "Deploying to $DEVICE_UNDER_TEST"
-            tools/deploy_ipk.sh $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/*.ipk
+            tools/deploy_ipk.sh --certification-mode $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/*.ipk
```
tomereli: Now this commit message makes it clear why we need this flag - I would prefer to make it clear in the PR description and the commit which initially adds this flag.

tomereli (approved):
Mainly nitpicks so approving. Please update the PR description, it says nothing about the new certification mode flag.

tools/deploy_ipk.sh
rmelotte: I can convert it to a short option if you want. 
I find having both a short option and a long option for everything is more work (and in the end we mess up the options by forgetting to update one or the other), and doesn't bring much advantages (we end up always using only one of the 2).

tools/deploy_ipk.sh
rmelotte: `BatchMode` doesn't ask for a password (it fails if you need a password).
It's used in the very first check, it's not strictly needed to use it afterwards (although it doesn't hurt).

tools/deploy_ipk.sh
rmelotte: We can definitely enable it when deploying on our targets, but enabling it by default from the prplWrt feed looks wrong to me.
In any case I think it doesn't hurt to have and use this option in CI, so if you agree let's keep it and we can discuss the default value in prplWrt later.

tomereli: Agree

rmelotte:
Rebased on master.

PR #770, opened 2020-01-30
Feature/ci add cppcheck common

rmelotte:
Add a CI job to run cppcheck on common, save the results as an artifact and output a summary.

Only issues of severity 'warning' and 'error' are enforced for now.

Fix the issues of severity 'warning' and 'error', and some of severity style.


tools/docker/static-analysis/cppcheck.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+scriptdir="$(cd "${0%/*}"; pwd)"
+rootdir="${scriptdir%/*/*/*}"
+
+. "${rootdir}/tools/docker/functions.sh"
+
+OUTPUT_FILE="$rootdir/cppcheck_results.txt"
```
tomereli, line 8: why not add the output in an 'out' directory in this folder? (nitpick)

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -0,0 +1,89 @@
...
+    cppcheck --template='{file}:{line}:{column}: {severity}: {message} [{id}]\n{code}' --force --error-exitcode=1 --enable=warning,style,information -q $INCLUDES "$@" 2>&1 | tee "$OUTPUT_FILE"
+}
+
+colorize_severity() {
+    local severity msg
+    severity="$1"
+    msg="$2"
+    case "$severity" in
+        error)
+            printf '\033[1;31m%s\033[0m\n' "$msg"
```
tomereli, line 45: Why not include `functions.sh` and use `err` `info` and friends?

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -0,0 +1,89 @@
...
+            status=1
+        else
+            colorize_severity "ok" "$nb_issues issues with severity $s"
+        fi
+    done
+    return $status
+}
+
+main() {
+    if ! command -v cppcheck > /dev/null ; then
```
tomereli, line 94: Please add cppcheck to the README.md in the prerequisites step.

.gitlab-ci.yml
```diff
@@ -5,6 +5,16 @@
   tags:
     - docker
 
+cppcheck:
+  image: prplfoundationinc/prplmesh-builder:alpine-3.11.3
+  stage: build
+  script:
+    - tools/docker/static-analysis/cppcheck.sh common
```
tomereli, line 14: Since eventually, we will run it on all folders - why not just make cppcheck.sh accept no argument and internally only run the check on common for now?

.gitlab-ci.yml
```diff
@@ -5,6 +5,16 @@
   tags:
     - docker
 
+cppcheck:
+  image: prplfoundationinc/prplmesh-builder:alpine-3.11.3
```
tomereli, line 10: Why are we using the apline builder for cppcheck? This should be mentioned in the commit message.

.gitlab-ci.yml
tomereli: Also, why do we run it only on common? why not run it on all folders and be done with it? Can you explain the reason in the commit message?

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_1905_vs.h
```diff
@@ -30,7 +30,7 @@ class tlvVsClientAssociationEvent : public BaseClass
 {
     public:
         tlvVsClientAssociationEvent(uint8_t* buff, size_t buff_len, bool parse = false);
-        tlvVsClientAssociationEvent(std::shared_ptr<BaseClass> base, bool parse = false);
+        explicit tlvVsClientAssociationEvent(std::shared_ptr<BaseClass> base, bool parse = false);
```
tomereli, line 33: The commit message is not very clear.

tomereli (approved):
Mostly nitpicks, so approving.

tools/docker/static-analysis/cppcheck.sh
rmelotte: We could, although we don't have such a folder for anything else at the moment.
In our case the scrupt is mostly meant to be used in a docker container though, because the results highly depend on the `cppcheck` version.

tools/docker/static-analysis/cppcheck.sh
rmelotte: Because there is 1 color per severity type. `functions.sh` only has `err` and `success`. I didn't add them to `functions.sh` because I don't see any other use for them at the moment.

tools/docker/static-analysis/cppcheck.sh
rmelotte: This comment makes me realize I misunderstood your similar comment on a previous PR.

`cppcheck` is not a pre-requisite to build prplMesh, so to me it doesn't belong to that list (the same is true for `uuidgen`, it's only needed if you want to build an ipk using the docker script).
Also, as the results really depend on the cppcheck version, I don't see much use in running it with the Ubuntu version.

What we could do if there is interest for it, is to add a section on how to start a docker container running cppcheck the same way the CI does.

.gitlab-ci.yml
rmelotte: I wanted to keep the usage of `cppcheck.sh` flexible, to be able to call it on any folder namely (you can even give options to `cppcheck.sh`, they will be passed to cppcheck).

In CI we can only run it on `common` for now because the other folders still have many issues of severity 'warning' and 'error' to solve first. The job is made to fail if any of those issues remain.

.gitlab-ci.yml
rmelotte: Done in a fixup (we use alpine because the version in Ubuntu 18.04 is too old).

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_1905_vs.h
rmelotte: Ok, I will explain the issue in more detail when I rebase.

.gitlab-ci.yml
tomereli: So you can add an optional parameter which is the directory to check. If none given, default should be to check all.

tools/docker/static-analysis/cppcheck.sh
tomereli: This would be great - then people tasked with fixing cppcheck issues can use it instead of relying on CI.

rmelotte:
Updated to use a compilation database instead, and rebased on master.
Because we don't want to display the issues for folders other than common (until the issues for other folders are fixed), there remain a lot of " * [unmatchedSuppression]" which we would have to ignore.

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
+
+scriptdir="$(cd "${0%/*}"; pwd)"
+rootdir="${scriptdir%/*/*/*}"
+
+. "${rootdir}/tools/docker/functions.sh"
+
+OUTPUT_FILE="$rootdir/cppcheck_results.txt"
+INCLUDES="\
+-I$rootdir/framework/platform/bpl/include \
```
arnout: (nitpick) I'd prefer the contents of the variable to be indented.

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -0,0 +1,96 @@
...
+-I$rootdir/common/beerocks/bwl/include"
+
+usage() {
+    echo "usage: $(basename "$0") <source> [source]"
+    echo "run cpp check on <source> and output number of issues per severity"
+}
+
+run_cppcheck() {
+    # --force is needed to check more than 12 ifdefs
+    cppcheck --template='{file}:{line}:{column}: {severity}: {message} [{id}]\n{code}' --force --error-exitcode=1 --enable=warning,style,information -q $INCLUDES "$@" 2>&1 | tee "$OUTPUT_FILE"
```
arnout: (nitpick) Line is too long.

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -0,0 +1,96 @@
...
+    severities="error warning style performance portability"
+    status=0
+    for s in $severities; do
+        nb_issues="$(grep -Ec "^[^ ]+ $s:" "$OUTPUT_FILE")"
+        if [ $nb_issues -gt 0 ] ; then
+            colorize_severity "$s" "$nb_issues issues with severity $s"
+            case "$s" in
+                # report the following severities in exit code:
+                error|warning)
+                    status=$((status + 1))
```
arnout: Note that if we have exactly 256 errors+warnings, this will overflow back to 0. Not important enough to fix IMO though.

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -0,0 +1,96 @@
...
+                    ;;
```
arnout, line 82: I think it would be good to also show these on the output, so you don't have to hunt down the artefacts to fix issues. Something as simple as the grep above without the -c should be sufficient.

However, that should probably be done before the summary...

Ideal output would be:
```
WARNING
WARNING
WARNING
ERROR
ERROR
ERROR
3 issues with severity error
3 issues with severity warning
0 issues with severity style
```

i.e., first all warnings, then all errors, then the summary. This makes sure the errors are clearly visible at the end.

(nitpick)

.gitlab-ci.yml
```diff
@@ -5,6 +5,28 @@
   tags:
     - docker
 
+.with-compilation-database:
```
arnout: This compilation database has a little disadvantage: it limits the cppcheck to the "linux" headers and sources. So only the dummy bwl and bpl are checked.

This can be fixed later by adding different variants (but that's more work because we need stubs for the different libraries that are used by bwl/bpl).

tools/docker/builder/Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN apt-get update && apt-get install -y \
     binutils \
     clang-format \
     cmake \
+    cppcheck \
```
arnout: We actually don't use it in this image anymore, so I guess this commit is not needed.

Do we have a Dockerfile for the Alpine image?

common/beerocks/bcl/source/beerocks_thread_base.cpp
```diff
@@ -38,7 +38,7 @@ void thread_base::join()
 void thread_base::stop(bool block)
 {
     should_stop = true;
-    before_stop();
+    thread_base::before_stop();
```
arnout, line 45: I don't think this is correct... This is declared empty in thread_base, so if the intention was to call the one from thread_base, it would just have been left out.

But indeed, the code is in fact broken, because when called from the destructor, the derived class's before_stop is no longer available!

So, I think the proper fix is to:
- duplicate the code below to the destructor instead of calling stop();
- add a destructor to async_work_queue that calls its own before_stop().

This is not a nitpick!

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport_messages.h
```diff
@@ -304,10 +302,6 @@ class InterfaceConfigurationRequestMessage : public Message {
         }
     }
 
-    virtual const std::string topic_prefix() const { return kTopicPrefix; }
-
-    virtual const std::string topic() const { return topic_prefix(); }
```
arnout, line 309: Can't this one stay in this class?

arnout (approved):
Mostly nitpicks, except for one bug that you introduced. But you can fix that before merging.

This one will require full rebase before merging, since new cppcheck failures may have been introduced.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_1905_vs.h
ghost, line 33: @tomereli Do you know why the autogenerated code is even in git?
Usually, autogenerated code should not be put under version control. There's no point and it can cause conflicts in some cases (in our case, it just clutters up reviews that change the TLVs)

tools/docker/static-analysis/cppcheck.sh
ghost: I think it is important. Glibc docs even explicitly warn against this:
```
Warning: Don’t try to use the number of errors as the exit status. This is actually not very useful; a parent process would generally not care how many errors occurred. Worse than that, it does not work, because the status value is truncated to eight bits. Thus, if the program tried to report 256 errors, the parent would receive a report of 0 errors—that is, success.

For the same reason, it does not work to use the value of errno as the exit status—these can exceed 255. 
```
https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html

.gitlab-ci.yml
rmelotte: In that case, maybe it would be better to keep using the hardcoded list of includes in the meantime? What do you think?

tools/docker/builder/Dockerfile
rmelotte: Indeed it's a mistake. The alpine image is already there and already contains cppcheck (`tools/docker/builder/alpine/Dockerfile`).

tools/docker/static-analysis/cppcheck.sh
rmelotte: Actually the exact value of the exit code is not used, only the fact that it's non-zero, so I'll just change it to `status=1`.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_1905_vs.h
arnout: If everyone new to the project asks this question, maybe there's something wrong with the answer :-)

The answer is:

-----

By putting the AutoGenerated source in revision control, it is easier for someone to browse the code after a simple checkout or even online, without building it. Even for review it is sometimes easier, e.g. when tlvf is changed it's more relevant to review the generated code than the python code.

And to make sure the AutoGenerated code in git is always consistent with the yaml source, we have CI that verifies it.

-----

I also wouldn't have chosen to put the generated code in git, but I don't feel so strongly about it that I'm going to fight the decision. I believe @vitalybu is the only real proponent.

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport_messages.h
rmelotte: Indeed. Done.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_1905_vs.h
tomereli: What Arnout said is the answer and I fully agree with it. If you want we can put it up for a vote again..

common/beerocks/bcl/source/beerocks_thread_base.cpp
tomereli: Is this resolved?

tomereli (approved):
There is one major comment by Arnout which is not addressed. Other than that LGTM

.gitlab-ci.yml
rmelotte: I dropped it and used the static list of includes for now.

tools/docker/static-analysis/cppcheck.sh
rmelotte: You mean you want the issues to be sorted by severity level?
In that case I can wait until cppcheck is done, then output them sorted.
If I keep the current behavior and add the warnings/errors at the end, they will be output 2 times...

common/beerocks/bcl/source/beerocks_thread_base.cpp
rmelotte: Yes, done. Nice catch BTW!

tools/docker/static-analysis/cppcheck.sh
arnout: Never mind, it was absolutely a "nice-to-have". Please merge ASAP.

rmelotte:
Rebased on master, fixed some of the nitpicks.

tools/docker/static-analysis/cppcheck.sh
rmelotte: Added to the wiki: https://github.com/prplfoundation/prplMesh/wiki/Testing-in-CI

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_1905_vs.h
vitalybu: I actually was against it at first as well, but the reasons Arnout mentioned eventually convinced me.

arnout: Hang on, so who actually wants the generated code in git? IIUC @tomereli, @vitalybu, @arnout and @fvbogaert are all saying that they would prefer the generated code to *not* be in git, but that they see the arguments and are not opposed to it either. So who is left as a proponent? @morantr ? @rmelotte ?

Just to be clear, some advantages of *not* having the generated code in git:

- There is no risk that the generated code is inconsistent with the yaml files.
- It makes it clearer that this is generated code that shouldn't be modified.
- It removes some commits ("update autogenerated") and makes others easier to review.
- We can remove the check from CI.

rmelotte: I don't like having the generated code in the repository either.
However in this case, the way to generate it is nothing "standard" or easy to read or check for errors, so I also agree for reviews it can help to have it included.

tomereli: @arnout - The main disadvantage of removing the autogenerated files from git is reviewing changes to tlvf.py. Unfortunately, tlvf.py is not a very readable python module, so I think that a followup commit which shows the change is an advantage, not a disadvantage.
I assume @morantr and @LiorAmram would agree on this.

RanRegev: M2c for this discussion -
Some PDFs which are the output of a complicated 'compilation' process are kept in source control. E.g. https://github.com/wg21il/Papers/tree/P2022/master/P2022/generated
These files are the "final product" and are the subject for review by humans.

Intermediate code, of all kinds, that is meant to be consumed mainly by other code, IMO should be kept locally, not in a source control, the reasons @arnout popinted.

The main issue that I see, which legitimate the use of git for generated code is human review.
So maybe we are using an incorrect tool to solve the problem.
Maybe we should address the review-auto-generated-code problem differently.
E.g. Supply a really easy-to-use generator.

tomereli: OK @RanRegev, which easy-to-use generator do you have in mind?

RanRegev: @tomereli, 
1. I suggested that sometimes we choose the wrong tool to solve a real problem.
If we agree that the main reason to save auto generated files in git is to enable humans review them - then maybe it is the wrong tool.
2. As for your question:
a. the [documentation](https://github.com/prplfoundation/prplMesh/blob/master/framework/tlvf/README.md) does not help. It says that there are 2 argument needed, but it calls for at least 3. I spent 10 minutes on it and I still don't know how to use tlvf.py and compile even a single file. Yes, maybe I'm a slow catcher and hard to understand, but 10 minutes are too much.
b. I understand the that _final_ requirement is that this script runs via the build system: bring other headers with it, compile many files together, have debug flags, and more. But this use of the script doesn't mean that it has to be complicated from the first place. For simple review it will be enough to have `tlvf.py <src-yaml>` - and that it is. No configuration file, not even an output file (stdout is great).
On top of this simplicity we can add as much complexity as we want. 
Having broad ranges of opportunities to compile tlv may, maybe, will remove the need store products in git.

tomereli: > @tomereli,
> 
> 1. I suggested that sometimes we choose the wrong tool to solve a real problem.
>    If we agree that the main reason to save auto generated files in git is to enable humans review them - then maybe it is the wrong tool.

I still don't understand your point - what is the problem you are referring to and what is the correct tool in your opinion we should use?

> 2. As for your question:
>    a. the [documentation](https://github.com/prplfoundation/prplMesh/blob/master/framework/tlvf/README.md) does not help. It says that there are 2 argument needed, but it calls for at least 3. I spent 10 minutes on it and I still don't know how to use tlvf.py and compile even a single file. Yes, maybe I'm a slow catcher and hard to understand, but 10 minutes are too much.

The tlvf generates (not compiles) multiple files, there is no support to generate a single file. We didn't hear any complaints till now about TLVF from other developers. Perhaps if you struggle with TLVF than ask someone from your team to guide you, almost everyone in the community knows how to add a new class / structure to TLVF. Raising complaints or arguments that TLVF is hard to use is pointless - we have plans for TLVFv2 but its a huge effort and since TLVFv1 just works, there is no plan to replace it anytime soon. Its not like we had a wide array of tools to select from and we chose TLVF - there was none.

>    b. I understand the that _final_ requirement is that this script runs via the build system: bring other headers with it, compile many files together, have debug flags, and more. But this use of the script doesn't mean that it has to be complicated from the first place. For simple review it will be enough to have `tlvf.py <src-yaml>` - and that it is. No configuration file, not even an output file (stdout is great).
>    On top of this simplicity we can add as much complexity as we want.
>    Having broad ranges of opportunities to compile tlv may, maybe, will remove the need store products in git.

Even if this was supported by TLVF, it wouldn't help to review TLVF changes. IMO nothing we change in TLVF will make any difference - unless we trust TLVF as a complete black box, then we can just use the YAML files and ignore the autogenerated files altogether. We are actually pretty close to this now, TLVF supports almost everything we need.

Regarding your suggestion for having tlvf option which accepts a single yaml file - this is not possible since yaml files may refer to other yaml files, and in most cases they do. Even it would work (and it wouldn't) - it won't make reviewing code easier, since it will cause developers to run tlvf to generate code instead of just browsing it, which beats the purpose IMO.

RanRegev: Fine.
Again - I'm not complaining on tlvf.
The question in this discussion is *why do we keep products in git?*
I see lots of drawbacks doing so, and I am not sure git is the correct solution to the problem raised when it was chosen. _I might be wrong_ by not understanding the problem from the first place.
Currently I don't work with tlvs, change them or have anything with them. When I will the _last_ thing that I'll do is to suggest a change in the mechanism. And that I will do *only after* I have tried everything with all the knowledge I can get.
Suggesting changing without understanding the problem is a _huge_ mistake. I'm not doing that.
The point here is not the tlv - but saving compilation (a.k.a generated) results.

PR #771, opened 2020-01-30
Hotfix/bpl uci remove swpal dependency

morantr:
Dependency to swpal, shouldn't have ever been.
swpal has changed and it caused prplMesh to break.
Fix it by removing swpal dependency, and unite common RDKB and UGW code.

Since uci_wrapper.h from swpal is no longer included, we need to copy
the latest changes from this file.
Align changes on UCI converter macros to latest swpal changes.

### DIFF
--- a/framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
@@ -9,20 +9,18 @@
 #ifndef BPL_CFG_UCI_H_
 #define BPL_CFG_UCI_H_
 
-#ifdef BEEROCKS_UGW
+#include <cstring>
 
-#include <stdint.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-
+#ifdef USE_LIBSAFEC
 #define restrict __restrict
 #include <libsafec/safe_str_lib.h>
 #undef snprintf_s
 #define snprintf_s snprintf
-
-#ifndef u_int_32
-#define u_int_32 unsigned int
+#elif USE_SLIBC
+#include <slibc/stdio.h>
+#include <slibc/string.h>
+#else
+#error "No safe C library defined, define either USE_LIBSAFEC or USE_SLIBC"
 #endif
 
 #ifndef _cplusplus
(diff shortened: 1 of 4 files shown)


yaniv-yitzhak (approved):

yaniv-yitzhak (approved):

framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
```diff
@@ -21,8 +21,11 @@
 #undef snprintf_s
 #define snprintf_s snprintf
 
-#ifndef u_int_32
-#define u_int_32 unsigned int
+#elif BEEROCKS_RDKB
```
tomereli: Please use the macros @arnout added instead of explicitly checking for BEEROCKS_RDKB

morantr: Which macro did he add?

tomereli:
```
#ifdef USE_LIBSAFEC
#define restrict __restrict
#include <libsafec/safe_str_lib.h>
#elif USE_SLIBC
#include <slibc/string.h>
#else
#error "No safe C library defined, define either USE_LIBSAFEC or USE_SLIBC"
#endif
```

morantr: I fixed it, bu rax40 doesn't compile now.
`slibc/stdio.h: No such file or directory`

rmelotte:
(just so you don't also need to download the artifacts I tried to get the relevant part from the logs):

In CI `beerocks_agent_wlan2.log` contains:
```
ERROR 20:32:53:937 <1994381984> bpl_cfg.cpp[286] --> cfg_get_sta_iface: Failed to get radio index from iface

ERROR 20:32:53:938 <1994381984> beerocks_slave_main.cpp[135] --> failed to read sta_iface for slave 
WARNING 20:32:53:938 <1994381984> beerocks_string_utils.cpp[112] --> string_utils::stoi(), cannot convert "" to int64_t, caller: beerocks_string_utils.h[75]
```

Issue #725, opened 2020-01-22
[TASK] Implement EasyMesh channel selection


arnout:
## Description

In slave_thread::handle_channel_selection_request(), send a message to the ap manager to perform the actual channel switch.

## Exit criteria

-  slave_thread::handle_channel_selection_request() sends a message to ap manager thread.
- This message contains all information from the channel selection request, including the power limit.
- ap manager thread takes into account this information to perform channel selection.

## Detailed description

This is needed for certification. Certification sends a request with a single channel enabled (= all channels except one disabled). Certification also sends a power limit. Current Intel VS implementation doesn't have a power limit so that needs to be added.

After channel switch an operating channel report must be sent. This is part of #369.

morantr:
https://github.com/prplfoundation/prplMesh/issues/62

PR #772, opened 2020-01-31
Feature/easy mesh agent channel selection

### LINKED ISSUE #369: [TASK] Implement Operating Channel Report for nl80211
## Description

When the agent receives a Channel Selection Request, after the channel switch has completed, it should send an Operating Channel Report.

## Exit criteria

- BWL event for channel switch exists in nl80211.
- BWL event for channel switch exists in dummy.
- This event triggers Operating Channel Report.
- Controller accepts and acks the Operating Channel Report.
- Controller updates database with new information.
- A test flow exists that validates all of the above.

YuriyMasechko:
Agent part of channel selection request. According to design attached in #62 channel selection request only restricted channels. Agent should check if channel switch is required and if yes if there is at least one non dfs supported channel that is not present in channel selection request. Tx power limit could be present in channel selection request. Only stab bwl functions were added for tx power limit.

Fixes #369 #725

### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -52,6 +52,20 @@ static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_
     }
 }
 
+static int8_t get_tx_power(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
+{
+    auto radio_channels = ap_wlan_hal->get_radio_info().supported_channels;
+    uint8_t channel     = ap_wlan_hal->get_radio_info().channel;
+
+    for (uint8_t i = 0;
+         i < beerocks::message::SUPPORTED_CHANNELS_LENGTH && i < radio_channels.size(); i++) {
+        if (radio_channels[i].channel == channel)
+            return radio_channels[i].tx_pow;
+    }
+
+    return 0;
+}
+
 static std::string
 get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
 {
@@ -589,7 +603,8 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
                 return false;
             }
 
-            notification->cs_params().channel = ap_wlan_hal->get_radio_info().channel;
+            notification->cs_params().channel  = ap_wlan_hal->get_radio_info().channel;
+            notification->cs_params().tx_power = get_tx_power(ap_wlan_hal);
             notification->cs_params().bandwidth =
                 uint8_t(beerocks::utils::convert_bandwidth_to_enum(
                     ap_wlan_hal->get_radio_info().bandwidth));
@@ -1046,6 +1061,7 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
             msg_params = &notification->cs_params();
         }
 
+        msg_params->tx_power  = get_tx_power(ap_wlan_hal);
         msg_params->channel   = ap_wlan_hal->get_radio_info().channel;
         msg_params->bandwidth = uint8_t(
             beerocks::utils::convert_bandwidth_to_enum(ap_wlan_hal->get_radio_info().bandwidth));
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -586,7 +586,8 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             LOG(DEBUG) << "Start ACS";
 
         // Set AP channel
-        if (!ap_wlan_hal->switch_channel(request->cs_params().channel,
+        if (ap_wlan_hal->get_radio_info().channel != request->cs_params().channel &&
+            !ap_wlan_hal->switch_channel(request->cs_params().channel,
                                          request->cs_params().bandwidth,
                                          request->cs_params().vht_center_frequency)) { //error
             std::string error("Failed to set AP channel!");
@@ -614,7 +615,12 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
                 ap_wlan_hal->get_radio_info().vht_center_freq;
             notification->cs_params().switch_reason = beerocks::CH_SWITCH_REASON_UNKNOWN;
             message_com::send_cmdu(slave_socket, cmdu_tx);
+            return false;
         }
+        //TODO: in case only tx power received need to send notification to slave as there
+        // is no event for this
+        if (request->tx_limit_valid())
+            ap_wlan_hal->set_tx_power_limit(request->tx_limit());
         break;
     }
     case beerocks_message::ACTION_APMANAGER_HOSTAP_SET_NEIGHBOR_11K_REQUEST: {
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -1757,6 +1757,8 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             LOG(ERROR) << "addClass ACTION_APMANAGER_HOSTAP_ACS_NOTIFICATION failed";
             return false;
         }
+        hostap_cs_params = notification_in->cs_params();
+
         auto notification_out = message_com::create_vs_message<
             beerocks_message::cACTION_CONTROL_HOSTAP_ACS_NOTIFICATION>(cmdu_tx,
                                                                        beerocks_header->id());
@@ -1782,6 +1784,8 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             LOG(ERROR) << "addClass cACTION_APMANAGER_HOSTAP_CSA_ERROR_NOTIFICATION failed";
             return false;
         }
+        hostap_cs_params = notification_in->cs_params();
+
         auto notification_out = message_com::create_vs_message<
             beerocks_message::cACTION_CONTROL_HOSTAP_CSA_NOTIFICATION>(cmdu_tx,
                                                                        beerocks_header->id());
@@ -1803,6 +1807,8 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             LOG(ERROR) << "addClass cACTION_APMANAGER_HOSTAP_CSA_ERROR_NOTIFICATION failed";
             return false;
         }
+        hostap_cs_params = notification_in->cs_params();
+
         auto notification_out = message_com::create_vs_message<
             beerocks_message::cACTION_CONTROL_HOSTAP_CSA_ERROR_NOTIFICATION>(cmdu_tx,
                                                                              beerocks_header->id());
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4387,10 +4387,12 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
     const auto mid = cmdu_rx.getMessageId();
     LOG(DEBUG) << "Received CHANNEL_SELECTION_REQUEST_MESSAGE, mid=" << std::dec << int(mid);
 
-    uint8_t selected_channel         = 0;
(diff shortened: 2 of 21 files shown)


common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -71,6 +71,8 @@ cACTION_APMANAGER_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION:
 cACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START:
   _type: class
   cs_params: sApChannelSwitch
+  tx_limit: int8_t
+  tx_limit_received: int8_t
```
arnout: This should be uint8_t, and it's boolean so add `# bool` at the end.

Also, I think `tx_limit_valid` is a slightly better name.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1135,6 +1135,12 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
     return true;
 }
 
+bool ap_wlan_hal_dwpal::set_tx_power_limit(int tx_pow_limit)
+{
+    LOG(DEBUG) << "set_tx_power_limit(): missing function implementation";
```
arnout, line 1591: I think this should be ERROR rather than DEBUG. (nitpick)

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
```diff
@@ -318,6 +318,13 @@ class ap_wlan_hal : public virtual base_wlan_hal {
      */
     virtual bool read_supported_channels() = 0;
 
+    /*!
+     * Set Transmit Power Limit 
+     *
+     * @return true on success or false on error.
+     */
+    virtual bool set_tx_power_limit(int tx_pow_limit) = 0;
```
arnout, line 328: Please indicate the unit used (either in the variable name or in the doxygen comment).

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -52,6 +52,20 @@ static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_
     }
 }
 
+static int8_t get_tx_power(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
```
arnout, line 55: Maybe add a doxygen comment, and mention that this needs to be preceded by either `read_acs_report` or `read_supported_channels` to make sure the radio info is up to date.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -586,7 +586,8 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             LOG(DEBUG) << "Start ACS";
 
         // Set AP channel
-        if (!ap_wlan_hal->switch_channel(request->cs_params().channel,
+        if (ap_wlan_hal->get_radio_info().channel != request->cs_params().channel &&
```
arnout, line 589: Shouldn't we also check bandwidth and center frequency?

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -205,6 +205,7 @@ class slave_thread : public beerocks::socket_thread {
     sSlaveBackhaulParams backhaul_params;
     beerocks_message::sNodeHostap hostap_params;
     beerocks_message::sApChannelSwitch hostap_cs_params;
+    std::list<wireless_utils::sChannelPreference> channel_preferences;
```
arnout: This is only used inside handle_channel_selection_request, so it can be a local variable there.

Also, unless it frequently needs insertion in the middle of the list, you should use a std::vector and not a std::list [SL.con.2](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rsl-vector).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4496,35 +4499,75 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
...
-                // TODO: check that the data is parsed properly after fixing the following bug:
-                // Since sFlags is defined after dynamic list cPreferenceOperatingClasses it cause data override
-                // on the first channel on the list and sFlags itself.
-                // See: https://github.com/prplfoundation/prplMesh/issues/8
+                beerocks::message::sWifiChannel wifi_channel;
+                wifi_channel.channel = *channel;
+                channels_list.push_back(wifi_channel);
             }
             LOG(DEBUG) << ss.str();
+            auto pref = beerocks::message::sPreference({operating_class, preference, reason_code});
```
arnout:
```suggestion
            auto pref = beerocks::message::sPreference({operating_class, preference, uint8_t(reason_code)});
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4496,35 +4499,75 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
...
+            channel_preferences.push_back({pref, channels_list});
```
arnout, line 4447: I think this entire algorithm can be simplified by first constructing preferences as a map of <sWifiChannel, sPreference> based on the hostap_params.supported_channels, and then in the parsing loop update those preferences. To make that easy, we'd need an `operator==` on sWifiChannel that checks only channel and bandwidth.

After parsing the preference TLVs, we end up with a preference list and can use that to do the actual channel selection.
- If the list is empty, it's an error and we don't switch channels.
- If the current channel is still in the list, it's not an error but we still don't switch channels.
- Else we choose a maximum-preference non-dfs channel.

Maybe this is a bit too big of a change, so consider it nitpick.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli, line 589: I'm really not sure about this one - from what I can remember, we need to trigger ACS by calling switch channel.
@vitalybu @morantr please comment.
@YuriyMasechko Is this commit absolutely needed? It looks like an optimization.

tests/test_flows.sh
```diff
@@ -181,8 +181,9 @@ test_channel_selection() {
     
     dbg "Confirming 1905.1 Ack Message request was received on agent"
     # TODO: When creating handler for the ACK message on the agent, replace lookup of this string
-    check_log ${REPEATER1} agent_wlan0 "ACK_MESSAGE"
-    check_log ${REPEATER1} agent_wlan2 "ACK_MESSAGE"
+    # TODO: currently controller sends empty channel selection request, so no switch is performed
+    # check_log ${REPEATER1} agent_wlan0 "ACK_MESSAGE"
+    # check_log ${REPEATER1} agent_wlan2 "ACK_MESSAGE"
```
arnout, line 209: Actually, this was incorrect anyway. Indeed, the agent should send a Channel Selection Response so that's what we have to test on the controller.

agent/src/beerocks/slave/ap_manager_thread.cpp
YuriyMasechko: It's optimization for case if only tx power limit was received and no channel switch is required. In such case we just path current channel to ap manager.

arnout (approved):
Lots of small improvements again. But overall looks good.

As usual I really would like more tests in test_flows. So test that the channel switch actually happened. But I understand that that is more work, because it requires a new event in the dummy bwl.

agent/src/beerocks/slave/ap_manager_thread.cpp
vitalybu: @arnout is right. All arguments should be checked for same values before deciding to avoid the channel switch.

YuriyMasechko: Bandwidth is checked in slave with operating channel check (same op channel or not). Center frequency is not a part of TLVF.

agent/src/beerocks/slave/son_slave_thread.h
YuriyMasechko: It is used once for now. But later it should be used for channel switch is case of radar for example.

agent/src/beerocks/slave/son_slave_thread.cpp
YuriyMasechko: As I remember from previous commits, we are trying to get rid of additional operators like operator==.

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
YuriyMasechko: Done.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
YuriyMasechko: All prints for non implemented functions inside this file are going to DEBUG.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: > As I remember from previous commits, we are trying to get rid of additional operators like operator==.

Absolutely not. What we have avoided is to add an `operator <=` which is only needed to be able to use `std::map` instead of `std::unordered_map`.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
YuriyMasechko: Done.

agent/src/beerocks/slave/son_slave_thread.h
YuriyMasechko: Changed to vector.

agent/src/beerocks/slave/son_slave_thread.cpp

tests/test_flows.sh
tomereli: So can we remove the TODO?

tomereli (changes requested):
LGTM but since we must remain passive in Not-Multi-AP mode, changes requested.
Note that in this mode from now on UTAF will fail if hostapd configuration is modified by prplmesh.

YuriyMasechko:
> LGTM but since we must remain passive in Not-Multi-AP mode, changes requested.
> Note that in this mode from now on UTAF will fail if hostapd configuration is modified by prplmesh.

Done.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -52,6 +52,20 @@ static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_
     }
 }
 
+static int8_t get_tx_power(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
+{
+    auto radio_channels = ap_wlan_hal->get_radio_info().supported_channels;
+    uint8_t channel     = ap_wlan_hal->get_radio_info().channel;
+
+    for (uint i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH && i < radio_channels.size();
```
vitalybu: We don't usually use `uint`. 
Any specific reason you decided to not use `uint8_t` here?

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -614,7 +615,12 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
                 ap_wlan_hal->get_radio_info().vht_center_freq;
             notification->cs_params().switch_reason = beerocks::CH_SWITCH_REASON_UNKNOWN;
             message_com::send_cmdu(slave_socket, cmdu_tx);
+            return false;
         }
+        //TODO: in case only tx power received need to send notification to slave as there
+        // is no event for this
+        if (request->tx_limit_valid())
+            ap_wlan_hal->set_tx_power_limit(request->tx_limit());
```
vitalybu, line 623: We do we allow setting the tx power using a channel switch message without providing channel information?

YuriyMasechko: According to Multi-AP spec channel preference TLV could be missing, so only TX power will be received.

agent/src/beerocks/slave/ap_manager_thread.cpp
YuriyMasechko: Changed to uint8_t. It was copy past from upper function.

tomereli (approved):

Issue #760, opened 2020-01-29
[TASK] CI: keep all the docker images for some time


rmelotte:
## Description

Follow-up from https://github.com/prplfoundation/prplMesh/pull/749#pullrequestreview-349275483 .
Since `c271a28abf16e7a78f2b0c4fb2b5cfda30239064`, the image from the master branch is saved with a separate tag, which is already a huge improvement from not saving the image at all.

However, when we make changes to the OpenWrt Dockerfile on other branches, the resulting images will always be deleted and rebuilt on every push, until they are finally pushed to master.

Worse, the branches lagging behind master and using an old Dockerfile risk having to rebuild the whole image at every push.

One way to solve it would be:
- The tags are always saved for some time, no matter what the branch name is (resulting in the image being kept).
- The tags are deleted if no container used them for the last X days (resulting in the image being deleted when the last tag using it is deleted).

However, this approach requires a way to know when an image last had a container created with it, which might not be possible.

## Exit criteria

- The images are always kept for some configurable time before being deleted, no matter what the branch name is.

rmelotte:
Knowing when an image was last used by a container doesn't seem to be supported at the moment: https://github.com/moby/moby/issues/4237 .


However I just came across this when inspecting the images, it might be useful:
```
        "Metadata": {
            "LastTagTime": "2020-01-16T10:08:11.388310467Z"
        }
```

rmelotte:
Since https://github.com/prplfoundation/prplMesh/pull/902, tags are not removed at all anymore.

To remove old tags, something like this can be used periodically on the runner:
```
#!/bin/sh
max_days=30
repository="prplmesh-"
# get all the image ids with the corresponding repository (no matter what the tag is):
image_ids="$(docker images --filter=reference="$repository:*" -q | sort | uniq)"
if [ ! "$?" ] ; then
    echo No image found, nothing to do.
    exit 0
fi
for i in $image_ids ; do
    # last time image was used:
    image_time="$(docker inspect "$i" --format='{{ .Metadata.LastTagTime.Unix }}')"
    # time elapsed since last tags (in seconds):
    time_since_last_tag="$(( $(date +%s) - image_time ))"
    if [ "$time_since_last_tag" -gt $((3600*24*max_days)) ] ; then
        echo "Image with id $i is older than $max_days days and will be deleted."
        # get all the tags associated to the image id:
        tags_to_delete="$(docker image inspect "$i" --format='{{join .RepoTags " "}}')"
[...]

PR #774, opened 2020-01-31
.gitlab-ci.yml: openwrt: only delete images after some time

rmelotte:
Only keeping an image for the master branch is already an improvement
from not keeping any image at all, but is still a problem. When we
update the Dockerfile on another branch, it has to potentially be
fully rebuilt at every push.

Worse, when a change to the Dockerfile is merged into master, all the
other branches lagging behind master could have their image rebuilt
every time.

Always keep the unique tag that is created by the .build-for-openwrt
job.

At the end of the job, look for images that haven't had a tag created
for them for more than a pre-defined number of days. If such images
are found, remove them by removing all their associated tags.

With this change, images will be deleted 30 days after they were last
used in CI (an image is "used in CI" as soon as a pipeline builds and
tags it).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -86,14 +86,30 @@ run-tests:
     - mkdir -p "build/$TARGET_DEVICE"
     - tools/docker/builder/openwrt/build.sh -v -d "$TARGET_DEVICE" -t "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee "build/$TARGET_DEVICE/openwrt-build.log" | grep -E '^[[:blank:]]?make\[[12]\]|^Step|^ --->' --color=never --line-buffered
   after_script:
-    # if we're on master, tag the docker image with a persistant tag:
     - |
-      if [ $CI_COMMIT_REF_NAME == "master" ] ; then
-          echo "Keeping tag prplmesh-builder-$TARGET_DEVICE"
-          docker tag prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID prplmesh-builder-$TARGET_DEVICE
+      max_days=30
+      repository="prplmesh-builder-$TARGET_DEVICE"
+      # get all the image ids with the corresponding repository (no matter what the tag is):
+      image_ids="$(docker images --filter=reference="$repository:*" -q | sort | uniq)"
+      if [ ! "$?" ] ; then
+        echo No image found, nothing to do.
+        exit 0
       fi
-    # always delete the unique tag:
-    - docker rmi "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID"
+      for i in $image_ids ; do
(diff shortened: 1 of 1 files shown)


rmelotte:
Closes #760

tomereli (approved):

arnout (approved):
It's not clear to me how you're supposed to launch a manual test now. Do you need to go through the triggers and do a curl command? Or is there some webby way to launch a pipeline with variables.

Can you document this in the wiki or maybe in CONTRIBUTING.md?

rmelotte:
I didn't try, but there should be a way to do it through the API with curl indeed, it might become useful at some point.
There is also a webby way through this page https://gitlab.com/prpl-foundation/prplMesh/pipelines/new

I'll update the wiki: https://github.com/prplfoundation/prplMesh/wiki/Testing-in-CI

.gitlab-ci.yml
```diff
@@ -86,14 +86,30 @@ run-tests:
     - mkdir -p "build/$TARGET_DEVICE"
     - tools/docker/builder/openwrt/build.sh -v -d "$TARGET_DEVICE" -t "prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID" 2>&1 | tee "build/$TARGET_DEVICE/openwrt-build.log" | grep -E '^[[:blank:]]?make\[[12]\]|^Step|^ --->' --color=never --line-buffered
   after_script:
-    # if we're on master, tag the docker image with a persistant tag:
     - |
-      if [ $CI_COMMIT_REF_NAME == "master" ] ; then
-          echo "Keeping tag prplmesh-builder-$TARGET_DEVICE"
-          docker tag prplmesh-builder-$TARGET_DEVICE:$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_ID prplmesh-builder-$TARGET_DEVICE
+      max_days=30
+      repository="prplmesh-builder-$TARGET_DEVICE"
```
arnout, line 91: By using TARGET_DEVICE here, we will not automatically clean up after a device has been removed from CI (or renamed). So I think it would be better to use * here (if docker's filter supports that). Actually, maybe remove all old tags that start with prplmesh.

arnout (approved):
To be honest, I think this is more appropriate as a timer on the build server (or cron job for the people running Devuan - you know who you are @fvbogaert :-). Although there is of course an advantage to be able to do this without requiring admin access on every build server.

Failing that, I think it would be more appropriate as a separate job instead of an after_script. Maybe even a job that is only triggered on schedule - it makes no sense to do this 1000s of times per day.

.gitlab-ci.yml
rmelotte: Good catch, indeed, removing a target from CI becomes a problem (although a smaller problem).
I'll switch to `prplmesh-*` to include the prplmesh-builder images and prplmesh-tests-runner as well.
You're also right on the fact that it doesn't need to run every time. Given that it already takes a few seconds to run (even when there is no image to delete) and that we currently control every runner assigned to the project, I'm going to keep this on the runners.

I'll later add instructions to the wiki on how to use it.

rmelotte:
Closing. I'll open another PR which removes the image deletion entirely, and do the deletions directly on the runners.

PR #776, opened 2020-02-02
[TASK] Certification mode compile time flag

tomereli:
Currently, certification mode can be enabled/disabled in runtime via UCI. This is a security breach in a real product since it opens the controller and agent to an attack via CAPI command socket which is not protected.

This is not an issue in a side version used for certification, but must not be allowed in a real product.

Therefore - add a compile-time flag - `CERTIFICATION`, default `ON`.
If `OFF`, we always return false in `cfg_get_certification_mode()`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,7 @@ endif()
 
 option (BUILD_AGENT "Build EasyMesh agent" ON)
 option (BUILD_CONTROLLER "Build EasyMesh controller" ON)
+option (CERTIFICATION "EasyMesh certification support" ON)
 
 ## Generic checks and defaults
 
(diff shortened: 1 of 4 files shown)


morantr (approved):
The question is if it defined as a compilation flag, why do we need it defined also in the database?
IMO, the compilation flag is enough.

tomereli:
> The question is if it defined as a compilation flag, why do we need it defined also in the database?
> IMO, the compilation flag is enough.

Actually, based on your feedback, I now think this is not needed at all, since the platform should protect the UCI DB from attackers. So I'll close this PR.

tomereli:
Thinking about this PR again - it is not needed. The certification mode can be disabled in a product and the platform should protect the UCI DB.

ghost:
Note: you misspelled certification in the CMakeLists.txt, so I don't think this feature will actually work
```
add_definitions(-DCERTIFICAITON)
```

tomereli:
> Note: you misspelled certification in the CMakeLists.txt, so I don't think this feature will actually work
> 
> ```
> add_definitions(-DCERTIFICAITON)
> ```

I spent a lot of time finding this typo only to realize the whole change is not needed :)

PR #779, opened 2020-02-02
Hotfix/rdkb toolchain builder

tomereli:
By default, the user and group id given by useradd is 1000, which is why
this docker image will work only for the first user in the server.

Fix this by adding the user and group id in the entrypoint script by
adding a new user and running the bitbake command with su.

In addition, rename build.sh to bitbake.sh, beautify and optimize the
Dockerfile, and run shellcheck on the entripoint and bitbake scripts.

Now we can alias bitbake to call build.sh bitbake "$@" and run it as
follows:
bitbake rdk-generic-broadband-image


tools/docker/builder/rdkb/toolchain/build.sh
```diff
@@ -11,23 +11,51 @@ rootdir="${scriptdir%/*/*/*/*/*}"
...
+            -- ) shift; break ;;
+            * ) err "unsupported argument $1"; usage; exit 1 ;;
+        esac
+    done
+
+    build_image
     # Default docker arguments
     docker_args="\
-     --workdir=$PWD --user=${SUDO_UID:-$(id -u)}:${SUDO_GID:-$(id -g)} \
-     -e USER=${SUDO_USER:-${USER}} -e HOME=${HOME} -v ${HOME}:${HOME} \
```
rmelotte, line 26: Did the script really used to mount the whole "$HOME" inside the container? Why not just the folder you need?

rmelotte (commented):
Building a docker image for every different user doesn't look like a convenient approach.
Why do you need to have the same user as the host user inside the container?
If you really need to, you could create the users at runtime instead, to avoid having a different image for every user.

Not specific to this PR: The Dockerfile uses a lot of unnecessary layers, which will increase the image size a lot.
See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get for details about the specific apt-get case.

rmelotte (approved):
Approving already since it's needed urgently.
If using "$HOME" as a volume is really doing what I think it does, it's not going to work in Gitlab CI, but we can anyway refactor the Dockerfile file when/if we use it in CI.

tools/docker/builder/rdkb/toolchain/build.sh
tomereli: Fixed to use a single image.

tomereli:
No rush to merge this. Please do a thorough review.

tools/docker/builder/rdkb/toolchain/Dockerfile
```diff
@@ -8,37 +8,20 @@
...
-RUN apt-get install -y bzr cvs gawk flex gettext zip unzip connect-proxy
-
-# Build support libraries
-RUN apt-get install -y libncurses5-dev libssl-dev libxml-parser-perl \
-  zlib1g-dev libjson-c-dev repo diffstat texinfo chrpath wget cpio
-
-RUN apt-get install openssl
-# Python/Java
-RUN apt-get install -y openjdk-8-jdk python3 python2.7-dev python-yaml
+RUN apt-get update --fix-missing && \
```
rmelotte: The more I see it, the less I like it.
Does `apt-get update` fail systematically or is the `fix-missing` only there for some cases where it fails?
In case `update` fails for whatever reason (an incorrectly configured source or whatever), I think you would rather want the build to fail and try again later, rather than going on building with potential issues with some packages.

tools/docker/builder/rdkb/toolchain/Dockerfile
```diff
@@ -8,37 +8,20 @@
...
+    apt-get install -y build-essential git cmake gcc-multilib \
+    subversion autoconf autogen libtool pkg-config \
+    bzr cvs gawk flex gettext zip unzip connect-proxy \
+    libncurses5-dev libssl-dev libxml-parser-perl \
+    zlib1g-dev libjson-c-dev repo diffstat texinfo chrpath wget cpio \
+    openssl python3 python2.7-dev python-yaml locales \
```
rmelotte: It would be great to have each package on a single line, and to sort them alphabetically if you can (it's really a nitpick though).

Something like this:
```
RUN apt-get update \
    && apt-get install -y \
    autoconf \
    autogen \
    build-essential \
    bzr \
    chrpath \
    cmake \
    connect-proxy \
    cpio \
    cvs \
    diffstat \
    flex \
    gawk \
    gcc-multilib \
    gettext \
    git \
    libjson-c-dev \
    libncurses5-dev \
    libssl-dev \
    libtool \
    libxml-parser-perl \
    locales \
    openssl \
    pkg-config \
    python-yaml \
    python2.7-dev \
    python3 \
    repo \
    subversion \
    texinfo \
    unzip \
    wget \
    zip \
    zlib1g-dev \
    && rm -rf /var/lib/apt/lists/*
```

tools/docker/builder/rdkb/toolchain/bitbake.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/sh
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2020 Tomer Eliyahu (Intel)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+scriptdir="$(cd "${0%/*}"; pwd)"
+rootdir="${scriptdir%/*/*/*/*/*}"
```
rmelotte: See @morantr 's changes on a similar script: https://github.com/prplfoundation/prplMesh/pull/676/commits/7a7cdc768de19afd4e77fdb4c5d9f9af07fabb00
If you can run shellcheck and fix what it reports it would be great (also a nitpick, I don't see any problematic one except VERBOSE not being used).

tools/docker/builder/rdkb/toolchain/entrypoint.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
+
 echo RDKB_ROOT=$RDKB_ROOT
+echo USER=$USER
+echo USER_ID=$USER_ID
+echo GROUP_ID=$GROUP_ID
+
+groupadd -g ${GROUP_ID} $USER
+useradd -l -u ${USER_ID} -g ${USER} ${USER}
```
rmelotte: After thinking about it again, you only need the correct UID/GID at runtime, but you could already create a user from the Dockerfile (no need for an argument, you could always use `rdkb` as the user for example).
Then you only need to change the UID/GID when running the container.

tools/docker/builder/rdkb/toolchain/entrypoint.sh
```diff
@@ -1,5 +1,13 @@
...
+su ${USER} <<EOSU
 cd $RDKB_ROOT/setup
```
rmelotte: Maybe check if `cd` was successful or not, to leave early if $RDKB_ROOT didn't exist/had a wrong value.
If you want to keep it as a oneliner, you can use something like `|| { echo "Failed to cd to ${RDKB_ROOT}/setup ; exit 1"}` (or if you prefer the longer version, `if ! cd .... ; then` also works).

tools/docker/builder/rdkb/toolchain/bitbake.sh
```diff
@@ -0,0 +1,65 @@
...
+    # Default docker arguments
+    docker_args="\
+     --workdir=$PWD \
+     -e HOME=${HOME} \
+     -e RDKB_ROOT=${RDKB_ROOT}
+     -e USER=${SUDO_USER:-${USER}} \
+     -e USER_ID=${SUDO_UID:-$(id -u)} \
+     -e GROUP_ID=${SUDO_GID:-$(id -g)} \
+     -v ${HOME}:${HOME} \
+     -v ${RDKB_ROOT}:${RDKB_ROOT} \
```
rmelotte: Do you really need both $RDKB_ROOT and the whole $HOME to be bind-mounted?
I don't see $HOME used directly, is it used by some of the commands run later?

tools/docker/builder/rdkb/toolchain/bitbake.sh
```diff
@@ -0,0 +1,65 @@
...
+     "
+
+    run docker container run -it --rm ${docker_args} rdkb-builder "$@"
+}
+
+RDKB_ROOT=$PWD
```
rmelotte: Please add quotes around "$PWD".

tools/docker/builder/rdkb/toolchain/bitbake.sh
```diff
@@ -0,0 +1,65 @@
...
+
+    OPTS=`getopt -o 'hir:v' --long help,image,rdkb-path:,verbose -n 'parse-options' -- "$@"`
+
+    if [ $? != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
+
+    eval set -- "$OPTS"
+
+    while true; do
+        case "$1" in
+            -h | --help)               usage; exit 0; shift ;;
```
rmelotte: shift after exit 0 won't do anything (nitpick).

tools/docker/builder/rdkb/toolchain/bitbake.sh
```diff
@@ -0,0 +1,65 @@
...
+            -i | --image)              build_image; exit 0; shift ;;
```
rmelotte: Same, shift is useless.
Also calling the function from here doesn't look like a good idea. If build_image ever need an option to be parsed first, the results of the script would depend on the order or the options (`-i --foo` wouldn't be the same as `--foo -i`).

In the case of `help` it looks ok though.

tools/docker/builder/rdkb/toolchain/bitbake.sh
```diff
@@ -0,0 +1,65 @@
...
+            -r | --rdkb-root)          RDKB_ROOT="$2"; shift; shift ;;
```
rmelotte: `shift` accepts an argument (the number of args to shift), you can use `shift 2` instead of `shift; shift`

tools/docker/builder/rdkb/toolchain/bitbake.sh
```diff
@@ -0,0 +1,65 @@
...
+            -v | --verbose)            VERBOSE=true; shift ;;
+            -- ) shift; break ;;
+            * ) err "unsupported argument $1"; usage; exit 1 ;;
+        esac
+    done
+
+    build_image
+    # Default docker arguments
+    docker_args="\
+     --workdir=$PWD \
```
rmelotte: if $PWD contains spaces, it breaks (the same is true for all the other options inside `docker_args`).
There is no easy way to avoid it without using quotes/eval, which is why having command args inside a string can be considered an anti-pattern.
You could use a bash array instead, but the easiest is probably to give the args to the docker command directly, since `docker_args` is not used elsewhere.

rmelotte (approved):
It all started with small nitpicks only, but you might want to fix some of them after all.

tools/docker/builder/rdkb/toolchain/Dockerfile
```diff
@@ -35,7 +35,13 @@ ENV LANG=en_US.UTF-8
...
 RUN rm -rf /var/lib/apt/lists/*
 
-RUN useradd -ms /bin/bash rdkb
+ARG USER_ID
+ARG GROUP_ID
+
+RUN if [ ${USER_ID:-0} -ne 0 ] && [ ${GROUP_ID:-0} -ne 0 ]; then \
+    groupadd -g ${GROUP_ID} rdkb &&\
+    useradd -l -u ${USER_ID} -g rdkb rdkb\
+;fi
```
vitalybu: What's up with the semicolon before the `fi`?

tools/docker/builder/rdkb/toolchain/Dockerfile
```diff
@@ -8,7 +8,7 @@
 FROM ubuntu:xenial
 
 # Update Software repository
-RUN apt-get update
+RUN apt-get update --fix-missing &&
```
vitalybu:
```suggestion
RUN apt-get update --fix-missing
```

What occasions? Never experienced an apt-get failure that required this...

tools/docker/builder/rdkb/toolchain/bitbake.sh
tomereli: .gitconfig for example. We need the home directory mounted for working with the real user instead of duplicating the home folder.

tools/docker/builder/rdkb/toolchain/entrypoint.sh
tomereli: But isn't it nicer using the actual user? So if the user wants to debug the build all he needs to do is to override the entrypoint.

tools/docker/builder/rdkb/toolchain/bitbake.sh
tomereli: Done and installed on prplmesh-s1 - @morantr and @rmelotte - great suggestion. I also installed a vscode extension now for it, very useful.

tools/docker/builder/rdkb/toolchain/Dockerfile
tomereli: Removed.

PR #669, opened 2020-02-03
[BUG] Certification test 5.7.1 fails

tomereli:
Controller certification tests 5.7.1 fails on fefec5f.

Logs can be found here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/401365002/artifacts/browse/logs/FAIL/

PR #784, opened 2020-02-04
Test combined infra metrics

arnout:
A number of fixes are required along the way.

### DIFF
--- a/tests/test_flows.sh
+++ b/tests/test_flows.sh
@@ -71,11 +71,17 @@ send_bml_command() {
 send_CAPI_command() {
     # send a CAPI command to a container.
     ip="$(container_ip "$1")"
+    [ -z "$ip" ] && return 1
     port="$(container_CAPI_port "$1")"
+    [ -z "$port" ] && return 1
     dbg "Sending to $ip:$port. Command: $2"
-    capi_command_result=$("${rootdir}/tests/send_CAPI_command.py" "$ip" "$port" "$2" | tr  -d '\r')
-    dbg "Result:\n$capi_command_result"
-    capi_command_reply=$(echo -e "$capi_command_result" | awk 'NR==2')
+    capi_command_result=$(mktemp)
+    [ -z "$capi_command_result" ] && { err "Failed to create temp file"; return 1; }
+    trap "rm -f $capi_command_result" EXIT
+    "${rootdir}/tests/send_CAPI_command.py" "$ip" "$port" "$2" > "$capi_command_result" || return $?
+    dbg "Result:"
+    [ "$VERBOSE" = true ] && cat "$capi_command_result"
+    capi_command_reply=$(tr  -d '\r' < "$capi_command_result" | awk 'NR==2')
 }
 
 check_log() {
--- a/tests/test_flows.sh
+++ b/tests/test_flows.sh
@@ -231,19 +231,46 @@ test_combined_infra_metrics() {
     status "test combined infrastructure metrics"
     check_error=0
     dbg "Send AP Metrics query message to agent 1"
-    mac_agent1_wlan0_hex=0x$(echo $mac_agent1_wlan0 | tr -d :)
-    check send_CAPI_1905 ${GATEWAY} $mac_agent1 0x800B "tlv_type,0x93,tlv_length,0x0007,tlv_value,{0x01 $mac_agent1_wlan0_hex}"
-    sleep 1
-    check_log ${REPEATER1} agent_wlan0 "AP_METRICS_QUERY_MESSAGE"
-    check_log ${GATEWAY} controller "AP_METRIC_RESPONSE_MESSAGE"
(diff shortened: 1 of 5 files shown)


tests/test_flows.sh
```diff
@@ -84,6 +84,19 @@ send_CAPI_command() {
...
 
+send_CAPI_1905() {
+    container="$1"; shift
+    dest="$1"; shift
+    messagetype="$1"; shift
+    command="DEV_SEND_1905,DestALid,${dest},MessageTypeValue,${messagetype}"
+    for arg; do
+        command="$command,$arg"
+    done
+    capi_command_mid=""
```
tomereli, line 95: Why is this needed? you set this 2 lines below

tests/test_flows.sh
```diff
@@ -207,35 +220,63 @@ test_client_capability_query() {
...
     sleep 1
     dbg "Confirming ap capability query has been received on agent"
     check_log ${REPEATER1} agent_wlan0 "AP_CAPABILITY_QUERY_MESSAGE"
     return $check_error
 }
+
 test_combined_infra_metrics() {
-    err "combined_infra_metrics not implemented yet."
-    return 0
+    status "test combined infra metrics"
```
tomereli: This has to be in a separate commit

tomereli (changes requested):
Looking good so far, but the test MUST make sure the agents actually sends the link metrics and check that the controller adds these as TLVs to the combined infrastructure message. This is the bug that is fixed by #669 so we have to be able to catch these kind of bugs in the test.

tomereli:
> Looking good so far, but the test MUST make sure the agents actually sends the link metrics and check that the controller adds these as TLVs to the combined infrastructure message. This is the bug that is fixed by #669 so we have to be able to catch these kind of bugs in the test.

@YuriyMasechko I don't see that this is taken care of by this PR. The problem is that when the UCC requests to send an empty combined infrastructure metrics like you did here:
```
# Trigger combined infra metrics
dbg "Send Combined infrastructure metrics message to agent 1"
check send_CAPI_1905 ${GATEWAY} $mac_agent1 0x8013
```

The UCC listener should send the preset certification CMDU. If there is no such CMDU (which was the bug solved in my other PR), then the UCC listener will send an *empty* 1905 CMDU with message type 0x8013 - the test failed in the certification since the CMDU was empty, while it should have contained TLVs.

In this case, I expect the test to simulate the same thing and actually verify that the received CMDU in the agent contains the TLVs we expect.
This is why I suggested to add dummy link metrics for now in the agent, so we can verify that the combined infrastructure metrics include these same values and is not just an empty CMDU which is a bug.

tomereli (changes requested):
> Looking good so far, but the test MUST make sure the agents actually sends the link metrics and check that the controller adds these as TLVs to the combined infrastructure message. This is the bug that is fixed by #669 so we have to be able to catch these kind of bugs in the test.

@YuriyMasechko I don't see that this is taken care of by this PR. The problem is that when the UCC requests to send an empty combined infrastructure metrics like you did here:
```
# Trigger combined infra metrics
dbg "Send Combined infrastructure metrics message to agent 1"
check send_CAPI_1905 ${GATEWAY} $mac_agent1 0x8013
```

The UCC listener should send the preset certification CMDU. If there is no such CMDU (which was the bug solved in my other PR), then the UCC listener will send an *empty* 1905 CMDU with message type 0x8013 - the test failed in the certification since the CMDU was empty, while it should have contained TLVs.

In this case, I expect the test to simulate the same thing and actually verify that the received CMDU in the agent contains the TLVs we expect.
This is why I suggested to add dummy link metrics for now in the agent, so we can verify that the combined infrastructure metrics include these same values and is not just an empty CMDU which is a bug.

YuriyMasechko:
> > Looking good so far, but the test MUST make sure the agents actually sends the link metrics and check that the controller adds these as TLVs to the combined infrastructure message. This is the bug that is fixed by #669 so we have to be able to catch these kind of bugs in the test.
> 
> @YuriyMasechko I don't see that this is taken care of by this PR. The problem is that when the UCC requests to send an empty combined infrastructure metrics like you did here:
> 
> ```
> # Trigger combined infra metrics
> dbg "Send Combined infrastructure metrics message to agent 1"
> check send_CAPI_1905 ${GATEWAY} $mac_agent1 0x8013
> ```
> 
> The UCC listener should send the preset certification CMDU. If there is no such CMDU (which was the bug solved in my other PR), then the UCC listener will send an _empty_ 1905 CMDU with message type 0x8013 - the test failed in the certification since the CMDU was empty, while it should have contained TLVs.
> 
> In this case, I expect the test to simulate the same thing and actually verify that the received CMDU in the agent contains the TLVs we expect.
> This is why I suggested to add dummy link metrics for now in the agent, so we can verify that the combined infrastructure metrics include these same values and is not just an empty CMDU which is a bug.

Added check for Transmitter/Receiver link metrics TLV inside handle_1905_combined_infrastructure_metrics() and test_flows.

tomereli (approved):

tests/test_flows.sh
arnout: (For future reference) It is needed because of the `return $?` in the following line, to make sure the mid is not set if the command fails.

PR #786, opened 2020-02-04
Feature/add associated clients tlv to topology response

### LINKED ISSUE #704: [TASK] Add Associated Clients TLV to Topology Response message
## Description

Add Associated Clients TLV to Topology Response message.

## Exit criteria

Topology Response message contains Associated Clients TLV with real data.

## Detailed description

For a detailed description, see task #525 (this task has been divided into smaller tasks to be managed more easily).

mariomaz:
[TASK] Add Associated Clients TLV to Topology Response message #704


common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
```diff
@@ -59,6 +59,9 @@ cACTION_BACKHAUL_ENABLE:
   wireless_iface_type: uint8_t  
   mem_only_psk: uint8_t
   backhaul_preferred_radio_band: uint8_t
+  supported_channels_list:
```
arnout, line 62: The commit message mentions that ruid is removed as well. However, that was already done in an earlier commit, so the commit message should be updated.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1824,10 +1824,13 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
         return false;
     }
 
-    for (const auto &vaps_list : m_vaps_map) {
+    for (const auto &radio_info : m_radio_info_map) {
+        auto ruid      = radio_info.first;
```
arnout, line 1953: (nitpick) for clarity it would be better to call this radio_mac.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1824,10 +1824,13 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+        auto vaps_list = radio_info.second.vaps_list;
+
         auto radio_list         = tlvApOperationalBSS->create_radio_list();
-        radio_list->radio_uid() = vaps_list.first;
-        for (const auto &vap : vaps_list.second.vaps) {
+        radio_list->radio_uid() = ruid;
```
arnout, line 1957: Note that radio_uid has to stay radio_uid since that's the name used by the MultiAP Spec.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -228,8 +228,23 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+     * @brief Information gathered about a radio.
+     *
+     * Radio information is obtained from messages sent by slave threads and is used to build
+     * the TLVs to include in notification messages or responses to CDMU query messages.
+     */
+    struct sRadioInfo {
+        beerocks_message::sVapsList vaps_list; /**< List of VAPs in radio. */
+        std::array<beerocks::message::sWifiChannel, beerocks::message::SUPPORTED_CHANNELS_LENGTH>
+            supported_channels; /**< Array of supported channels in radio. */
+    };
```
arnout: Excellent documentation, and excellent choice of std::array as the underlying type. :1st_place_medal:

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1493,7 +1493,24 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+
+        auto tuple_supported_channels = request->supported_channels_list(0);
+        if (!std::get<0>(tuple_supported_channels)) {
+            LOG(ERROR) << "access to supported channels list failed!";
+            return false;
+        }
+
+        auto channels = &std::get<1>(tuple_supported_channels);
+
+        std::array<beerocks::message::sWifiChannel, beerocks::message::SUPPORTED_CHANNELS_LENGTH>
```
arnout: Why this intermediate array rather than writing immediately to `m_radio_info_map[request->iface_mac()].supported_channels` ?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1493,7 +1493,24 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+            supported_channels;
+        for (size_t i = 0; i < supported_channels.size(); i++) {
+            supported_channels[i] = channels[i];
+        }
```
arnout: Can't we use std::copy?
```
    std::copy(channels, beerocks::message::SUPPORTED_CHANNELS_LENGTH,
              m_radio_info_map[request->iface_mac()].supported_channels);
```

agent/src/beerocks/slave/tlvf_utils.cpp
```diff
@@ -0,0 +1,70 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 Intel Corporation
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "tlvf_utils.h"
```
arnout, line 9: Nitpick: I think this would fit better in bcl than here.

agent/src/beerocks/slave/tlvf_utils.cpp
```diff
@@ -0,0 +1,70 @@
...
+
+#include <bcl/son/son_wireless_utils.h>
+#include <easylogging++.h>
+
+#include <tlvf/wfa_map/tlvApRadioBasicCapabilities.h>
+
+using namespace beerocks;
+
+bool tlvf_utils::add_ap_radio_basic_capabilities(
+    ieee1905_1::CmduMessageTx &cmdu_tx, const sMacAddr &ruid,
```
arnout, line 19: Again, radio_mac would be better than ruid (nitpick).

agent/src/beerocks/slave/tlvf_utils.cpp
```diff
@@ -0,0 +1,70 @@
...
+        operationClassesInfo->operating_class() = op_class;
+        operationClassesInfo->maximum_transmit_power_dbm() =
+            son::wireless_utils::get_operating_class_max_tx_power(supported_channels.data(),
+                                                                  op_class);
+
+        std::vector<uint8_t> non_oper_channels;
+        non_oper_channels = son::wireless_utils::get_operating_class_non_oper_channels(
+            supported_channels.data(), op_class);
+        // Create list of statically non-oper channels
+        if (!non_oper_channels.empty()) {
```
arnout: IIRC somebody made a change that makes it possible to allocate a list of size 0, so this condition would no longer be needed.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3347,8 +3348,14 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             return false;
         }
 
-        if (!add_radio_basic_capabilities()) {
-            LOG(ERROR) << "Failed adding WSC M1 TLV";
+        std::array<beerocks::message::sWifiChannel, beerocks::message::SUPPORTED_CHANNELS_LENGTH>
+            supported_channels;
```
arnout, line 3435: It would be better to change hostap_params.supported_channels into an array (or vector) as well, but that can be done later.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -241,6 +273,9 @@ class backhaul_manager : public btl::transport_socket_thread {
         beerocks_message::sVapsList vaps_list; /**< List of VAPs in radio. */
         std::array<beerocks::message::sWifiChannel, beerocks::message::SUPPORTED_CHANNELS_LENGTH>
             supported_channels; /**< Array of supported channels in radio. */
+        std::unordered_map<sMacAddr,
+                           std::unordered_set<associated_client_t, associated_client_hash>>
+            associated_clients_map; /**< Associated clients grouped by BSSID. */
```
arnout: Since the MAC address here has to be one of the VAPs in vaps_list, it would make sense to do something similar to the radioInfo, i.e. make a struct that contains bssid, ssid and associated clients. Note that it wouldn't be a map then, because we need to be able to index it both based on BSSID and on SSID. So a vector and searching with find() is more appropriate.

However, that's probably a bit too big of a change, so let's leave that for some later refactoring.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1722,6 +1722,60 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+            LOG(ERROR) << "Failed building ACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION message!";
+            return false;
+        }
+
+        // Create a new tuple to contain client association information
+        auto associated_client =
+            std::make_tuple(msg->client_mac(), std::chrono::steady_clock::now());
+
+        // If exists, remove current client association information for given client
+        auto &&associated_clients =
```
arnout: Why the `&&` reference? A normal reference seems more appropriate to me. But I'm not so familiar with the semantics of `&&` references (I don't even know what they're called :-))

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1722,6 +1722,60 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+            m_radio_info_map[msg->iface_mac()].associated_clients_map[msg->bssid()];
+        auto it =
+            std::find(associated_clients.begin(), associated_clients.end(), associated_client);
+        if (it != associated_clients.end()) {
+            associated_clients.erase(it);
```
arnout: I'm not sure that this is the correct thing to do. AFAIU it is possible for a client to be associated in multiple BSSIDs. And in any case, when it disassociates, a disconnect message should be received from it.

If it is necessary to be able to find the one BSSID with which a station is associated, we should probably find another solution for that.

That said: probably better keep it as is now and fix it if there really is a problem.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1722,6 +1722,60 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+
+        // Create a new tuple to contain client association information
+        auto associated_client =
+            std::make_tuple(msg->client_mac(), std::chrono::steady_clock::now());
+
+        // If exists, remove current client association information for given client
+        auto &&associated_clients =
+            m_radio_info_map[msg->iface_mac()].associated_clients_map[msg->bssid()];
+        auto it =
+            std::find(associated_clients.begin(), associated_clients.end(), associated_client);
```
arnout: Doesn't this check both sides of the tuple? How can the timestamp ever be identical?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1722,6 +1722,60 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION>();
+        if (!msg) {
+            LOG(ERROR) << "Failed building ACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION message!";
+            return false;
+        }
+
+        // Create a new tuple to contain client association information
+        auto associated_client =
+            std::make_tuple(msg->client_mac(), std::chrono::steady_clock::now());
```
arnout: I think the timestamp should be marked as early as possible, i.e. it should be part of the message at least. Preferably it would already be set in bwl - but that would be a bigger change so not necessary here.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1722,6 +1722,60 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+
+        // Create a tuple to search for client association information
+        auto associated_client =
+            std::make_tuple(msg->client_mac(), std::chrono::steady_clock::time_point());
+
+        // If exists, remove client association information for disconnected client
+        auto &&associated_clients =
+            m_radio_info_map[msg->iface_mac()].associated_clients_map[msg->bssid()];
+        auto it =
+            std::find(associated_clients.begin(), associated_clients.end(), associated_client);
```
arnout: Since the time_point is definitely different from the one that was inserted, this will never find anything.

I think we need a helper function find_associated_client to find the client. Or it has to be a map client_mac -> time_point.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1954,23 +1952,71 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
 
-    auto tlvAssociatedClients = cmdu_tx.addClass<wfa_map::tlvAssociatedClients>();
-    if (!tlvAssociatedClients) {
-        LOG(ERROR) << "addClass wfa_map::tlvAssociatedClients failed, mid=" << std::hex << (int)mid;
-        return false;
+    // The Multi-AP Agent shall include an Associated Clients TLV in the message if there is at
+    // least one 802.11 client directly associated with any of the BSS(s) that is operated by the
+    // Multi-AP Agent
+    bool shall_include_associated_clients_tlv = false;
+    for (auto &&radio_info_entry : m_radio_info_map) {
```
arnout: Again, why `&&`? `const auto &radio_info_entry` seems more appropriate.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1954,23 +1952,71 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+        auto associated_clients_map = radio_info_entry.second.associated_clients_map;
+        for (auto &&associated_clients_entry : associated_clients_map) {
```
arnout: Same here.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2001,13 +2001,17 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
                         auto association_time = std::get<1>(associated_client);
 
-                        auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(
-                            now - association_time);
+                        auto elapsed =
+                            std::chrono::duration_cast<std::chrono::seconds>(now - association_time)
+                                .count();
+                        if (elapsed > UINT16_MAX) {
+                            elapsed = UINT16_MAX;
+                        }
```
arnout, line 2025: For completeness, we should also handle the `elapsed < 0` case.

arnout (approved):
I have a *lot* of comments, mostly nitpicks but a few important ones.

Please fix at least the important ones, rebase on master, wait for CI to finish and merge right away. This really needs to be tested close to master so don't wait too long after the rebase before merging.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Ok

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
mariomaz: Thanks

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/tlvf_utils.cpp
mariomaz: Let's see how this file evolves before moving it. Since it is a nitpick comment, I'm not going to move it for the moment.

agent/src/beerocks/slave/tlvf_utils.cpp
mariomaz: This parameter is copied to a TLV field named radio_uid, so should not be renamed.
Description is "Radio unique identifier of the radio for which capabilities are reported."

agent/src/beerocks/slave/tlvf_utils.cpp
mariomaz: Removed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
mariomaz: Ok. Not changed now

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Removed

agent/src/beerocks/slave/son_slave_thread.cpp
mariomaz: Ok, later

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: This is a C++11 rvalue reference. 
AFAIK, auto&& is the same as const auto&.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: There is a list of associated clients per BSSID. BSSID value is being received as part of the message and it is used to obtain the list of associated clients to that BSSID. So AFAIK there shouldn't be any problem.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: associated_clients is an std::unordered_set so std::find works on the hash value of the key. 
associated_client_hash has been defined so the hash value is computed as the hash value of the MAC address of the associated client. The timestamp is not even being considered to compare the tuples.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Ok, not changed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: AFAIK they are equivalent and && is preferred over const auto&. 
I'm sorry but don't know how to explain it. That's something I learned once and use whenever I have to iterate over a map. I can change it if you prefer.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Sorry, I was miserably wrong. And it was over complicated so I've also simplified it

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: You are right! Fixed with a map client_mac -> time_point

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: Thanks for the info.

I read all of [this article](http://thbecker.net/articles/rvalue_references/section_01.html) to understand what rvalue references are about. As far as I can see, it makes little sense to declare a variable as an rvalue reference. There's a [stackoverflow answer](https://stackoverflow.com/questions/18766324/is-it-useless-to-declare-a-local-variable-as-rvalue-reference-e-g-t-r-move) that says the same but I trust it a little less. 

As far as I can see, `const auto&` and `auto&&` are indeed more or less equivalent. However, the `const auto&` pattern is used all over the code, and is (IMO) easier to understand for people who are not utter C++11 experts. Also our CONTRIBUTING.md contains an example that uses that pattern. So I tend to prefer the `const auto&` variant even if they are equivalent.

arnout:
The `auto &&` thing is a nitpick so I removed the don't merge tag. We still need a second review however.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: I think you missed my point.

When you receive a message that a client has associated with a BSS, you remove that client from any previous BSS it was associated with. My point is that:
- maybe it actually still is associated with that BSS;
- in any case, when it does disassociate, we'll get a disassociation message and it will be removed then.

In other words, it is not needed and maybe incorrect.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1754,6 +1758,9 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
         return handle_1905_discovery_query(cmdu_rx, src_mac);
         //LOG(INFO) << "I got the Topology Query message!";
```
tomereli: This can be removed now (nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1769,6 +1776,39 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
...
+
+    // Capability bitmask is set to 0 because neither unassociated STA link metrics
+    // reporting or agent-initiated RCPI-based steering are supported
+
+    for (const auto &radio_info : m_radio_info_map) {
+        auto radio_mac          = radio_info.first;
+        auto supported_channels = radio_info.second.supported_channels;
+
+        if (!tlvf_utils::add_ap_radio_basic_capabilities(cmdu_tx, radio_mac, supported_channels)) {
+            return false;
```
tomereli, line 1844: Shouldn't we add an error message here?

framework/tlvf/tlvf_conf.yaml
```diff
@@ -28,6 +28,9 @@ include_yaml_path: {
   "tlvf/wfa_map/tlvApMetricQuery.yaml",
   "tlvf/wfa_map/tlvClientAssociationEvent.yaml",
   "tlvf/wfa_map/tlvClientAssociationControlRequest.yaml",
+  "tlvf/wfa_map/tlvApHtCapabilities.yaml",
```
tomereli, line 31: The commit message should mention which TLVs are added (nitpick)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1328,6 +1331,15 @@ bool master_thread::handle_cmdu_1905_ap_metric_response(const std::string &src_m
     return true;
 }
 
+bool master_thread::handle_cmdu_1905_ap_capability_report(const std::string &src_mac,
+                                                          ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received AP_CAPABILITY_REPORT_MESSAGE, mid=" << std::dec << int(mid);
```
tomereli, line 1338: We should also print the capabilities here. Then simulate the capabilities in the agent dummy implementation, and then make sure that test_flows.sh chech that the capability report was generated. But that can be in a different PR, so (nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -231,6 +231,22 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+     * Associated client information consists of:
+     * - The MAC address of the 802.11 client that associates to a BSS.
+     * - Timestamp of the 802.11 client's last association to this Multi-AP device.
+     *
+     * Associated client information is gathered from
+     * ACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION events received from slave threads.
+     *
+     * Associated client information is later used to fill in the Associated Clients TLV
+     * in the Topology Response message.
+     */
```
tomereli, line 245: It's really great that you write such a comprehensive description of the new container.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: I agree with @arnout - so please change this to `const auto &`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: I really prefer that you change it to `const auto &`. Pretty please?:)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1940,23 +1938,75 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+
+        // Get current time to compute elapsed time since last client association
+        auto now = std::chrono::steady_clock::now();
+
+        // Fill in Associated Clients TLV
+        for (auto &&radio_info_entry : m_radio_info_map) {
+            auto associated_clients_map = radio_info_entry.second.associated_clients_map;
+
+            // Associated clients map contains sets of associated clients grouped by BSSID
+            for (auto &&associated_clients_entry : associated_clients_map) {
```
tomereli: same here

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1940,23 +1938,75 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+            LOG(ERROR) << "addClass wfa_map::tlvAssociatedClients failed, mid=" << std::hex
+                       << (int)mid;
+            return false;
+        }
+
+        // Get current time to compute elapsed time since last client association
+        auto now = std::chrono::steady_clock::now();
+
+        // Fill in Associated Clients TLV
+        for (auto &&radio_info_entry : m_radio_info_map) {
```
tomereli: same here

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1940,23 +1938,75 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
...
+                auto associated_clients = associated_clients_entry.second;
+
+                if (associated_clients.size() > 0) {
+                    auto bss_list = tlvAssociatedClients->create_bss_list();
+
+                    bss_list->bssid() = bssid;
+
+                    // Information for each associated client includes its MAC address and
+                    // the timestamp of its last association.
+                    for (auto &&associated_client : associated_clients) {
```
tomereli: and here

tomereli (approved):
Approving already cause my comments are pretty simple can can be fixed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Error messages are already being written to log inside the add_ap_radio_basic_capabilities method, so there is no need to do it here back again.

framework/tlvf/tlvf_conf.yaml
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
mariomaz: Thanks

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

controller/src/beerocks/master/son_master_thread.cpp
mariomaz: I agree with you. All tests in test_flows.sh should be extended to test the contents of the messages and not only if a particular message name appears on the log. Tests should also check if the message has been sent or received and the number of times.

PR #789, opened 2020-02-05
Hotfix/fix client invalid rssi

morantr:
On commit 873bf9e I've implemented
a mechanism that blocks the start monitoring request for 3 seconds until
the client IP is received.
This solution wan not good enough since on a real-time system.
I encountered situations that the IP has been received after more than 15 seconds,
and in such a case, the monitoring was not started. This has caused
the statistics polling task not to work as well since the monitoring request
is inserting the client to the monitor DB, and without it, the statistics polling task
measurement will not work.

This PR reverts the commit that blocks the monitoring request, and add messages that update 
the IP on the monitor database independently.

Currently, when sending arp with an empty IP address, the client that receives
the request will ignore it since the empty IP string is translated to 0.0.0.0.
Replace the empty IP string to a wildcard IP to make sure that the client will
respond to the arp request. This will solve the ARP measurement issue that occurs before
the client IP has been updated on the monitor database. 

Another issue resolved on this PR is when a client is connected to the AP before prplMesh
has brought up, the controller
receives the auto-generated client connected notification and send start monitoring
request before the monitor is attached. The outcome is that the monitor ignores
the request and the client is not being monitored.
If a client is not monitored it cannot be measured and therefore could not be steered
by optimal path task.

To resolve this issue, add START_MONITORING_RESPONSE message. The association
with success flag so the association handling task could increment its state machine
only after receiving a response with success.

The auto-generated code will be added to the follow-up commit.


common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -898,7 +898,16 @@ bool network_utils::arp_send(const std::string &iface, const std::string &dst_ip
     struct sockaddr_ll sock;
     uint8_t packet_buffer[128];
 
-    uint32_t dst_ip_uint = network_utils::uint_ipv4_from_string(dst_ip);
+    // If the destination IP is empty, there is no point sending the arp, therefore replace the 
+    // with broadcast IP, so all clients will receive it and answer, but since the request is
+    // being sent to a specific mac address, then only the requested client will answer.
+    uint32_t dst_ip_uint; 
+    if (dst_ip.empty()) {
+        dst_ip_uint = network_utils::uint_ipv4_from_string("255.255.255.255");
```
vitalybu:
```suggestion
        dst_ip_uint = 0xFFFFFFFF; // Broadcast IP address
```

vitalybu: We're using a broadcast address (it's not wildcard). Please update the commit message.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_action.h
```diff
@@ -261,7 +257,13 @@ enum eActionOp_MONITOR: uint8_t {
...
     ACTION_MONITOR_CHANNEL_SCAN_ABORT_NOTIFICATION = 0x42,
     ACTION_MONITOR_CHANNEL_SCAN_FINISHED_NOTIFICATION = 0x43,
-    ACTION_MONITOR_ENUM_END = 0x44,
+    ACTION_MONITOR_STEERING_CLIENT_SET_GROUP_REQUEST = 0x50,
+    ACTION_MONITOR_STEERING_CLIENT_SET_GROUP_RESPONSE = 0x51,
+    ACTION_MONITOR_STEERING_CLIENT_SET_REQUEST = 0x52,
+    ACTION_MONITOR_STEERING_CLIENT_SET_RESPONSE = 0x53,
+    ACTION_MONITOR_STEERING_EVENT_CLIENT_ACTIVITY_NOTIFICATION = 0x54,
+    ACTION_MONITOR_STEERING_EVENT_SNR_XING_NOTIFICATION = 0x55,
+    ACTION_MONITOR_ENUM_END = 0x56,
```
vitalybu, line 268: Probably should mention in the commit message why you decided to reorder the opcodes...

vitalybu (changes requested):
Stopped at "controller: Send client new IP notification".
Will continue later.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -902,11 +902,12 @@ bool network_utils::arp_send(const std::string &iface, const std::string &dst_ip
...
-        dst_ip_uint = network_utils::uint_ipv4_from_string("255.255.255.255");
-    } else {
-       dst_ip_uint = network_utils::uint_ipv4_from_string(dst_ip);
-    }
+    dst_ip_uint = network_utils::uint_ipv4_from_string("255.255.255.255");
+    // if (dst_ip.empty()) {
+    //     dst_ip_uint = network_utils::uint_ipv4_from_string("255.255.255.255");
+    // } else {
+    //    dst_ip_uint = network_utils::uint_ipv4_from_string(dst_ip);
+    // }
```
vitalybu: Don't forget to revert it if all tests will pass

controller/src/beerocks/master/tasks/statistics_polling_task.cpp
```diff
@@ -27,6 +27,7 @@ statistics_polling_task::statistics_polling_task(db &database_, ieee1905_1::Cmdu
 
 void statistics_polling_task::work()
 {
+    return;
```
vitalybu: Don't forget to revert it if all tests will pass

vitalybu (changes requested):

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -898,7 +898,16 @@ bool network_utils::arp_send(const std::string &iface, const std::string &dst_ip
     struct sockaddr_ll sock;
     uint8_t packet_buffer[128];
 
-    uint32_t dst_ip_uint = network_utils::uint_ipv4_from_string(dst_ip);
+    // If the destination IP is empty, there is no point sending the arp, therefore replace the
```
tomereli, line 901: I just don't understand the comment "there is no point sending the arp" - you still do send the arp and there is a very good point. I think the comment should be updated to reflect the reason for why the destination IP may be empty.
Also please explain in the comment which clients were tested and answers broadcast ARPs.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1082,8 +1082,7 @@ bool ap_wlan_hal_dwpal::update_vap_credentials(
     for (const auto &it : hostapd_config_vaps) {
         // Skip STAs
         std::string entry_mode;
-        if (hostapd_config_get_value(it.second, "mode", entry_mode) &&
-            entry_mode != "ap") {
+        if (hostapd_config_get_value(it.second, "mode", entry_mode) && entry_mode != "ap") {
```
tomereli: How is this related to the change?

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -28,7 +28,6 @@ using namespace son;
 #define BEACON_MEASURE_REQ_TIME_SPAN 3000
 #define BEACON_MEASURE_MAX_ATTEMPTS 3
 #define REQUEST_RSSI_MEASUREMENT_DELAY 5000
-#define GET_CLIENT_IP_TIMEOUT_SEC 3
```
tomereli, line 31: I think it makes more sense to actually do `git revert <commit>` then `git commit --amend` to update with the (great!) explanation you added in the commit message. This way, the commit headline will include the hash of the reverted commit so its easier for searching in the git log.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -76,6 +76,7 @@ eActionOp_CONTROL:
   ACTION_CONTROL_CLIENT_RX_RSSI_MEASUREMENT_START_NOTIFICATION: 104
   ACTION_CONTROL_CLIENT_RX_RSSI_MEASUREMENT_NOTIFICATION: 105
   ACTION_CONTROL_CLIENT_NO_RESPONSE_NOTIFICATION: 106
+  ACTION_CONTROL_CLIENT_NEW_IP_ADDRESS_NOTIFICATION: 107
```
tomereli:
```suggestion
  ACTION_CONTROL_CLIENT_DHCP_NOTIFICATION: 107
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -275,19 +276,14 @@ eActionOp_MONITOR:
...
   ACTION_MONITOR_CLIENT_LINK_MEASUREMENTS_11K_RESPONSE: 35
   ACTION_MONITOR_CLIENT_RX_RSSI_MEASUREMENT_CMD_RESPONSE: 36
-  ACTION_MONITOR_STEERING_CLIENT_SET_GROUP_REQUEST: 37
-  ACTION_MONITOR_STEERING_CLIENT_SET_GROUP_RESPONSE: 38
-  ACTION_MONITOR_STEERING_CLIENT_SET_REQUEST: 39
-  ACTION_MONITOR_STEERING_CLIENT_SET_RESPONSE: 40
-  ACTION_MONITOR_STEERING_EVENT_CLIENT_ACTIVITY_NOTIFICATION: 41
-  ACTION_MONITOR_STEERING_EVENT_SNR_XING_NOTIFICATION: 42
+  ACTION_MONITOR_CLIENT_NO_ACTIVITY_NOTIFICATION: 37
+  ACTION_MONITOR_CLIENT_NEW_IP_ADDRESS_NOTIFICATION: 38
```
tomereli:
```suggestion
  ACTION_MONITOR_CLIENT_DHCP_NOTIFICATION: 38
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_control.yaml
```diff
@@ -252,6 +252,10 @@ cACTION_CONTROL_CLIENT_NO_RESPONSE_NOTIFICATION:
   _type: class
   mac: sMacAddr 
 
+cACTION_CONTROL_CLIENT_NEW_IP_ADDRESS_NOTIFICATION:
```
tomereli, line 260:
```suggestion
cACTION_CONTROL_CLIENT_DHCP_NOTIFICATION:
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_monitor.yaml
```diff
@@ -145,6 +145,11 @@ cACTION_MONITOR_CLIENT_LINK_MEASUREMENTS_11K_RESPONSE:
   _type: class
   params: sLinkMeasurementsResponse11k 
 
+cACTION_MONITOR_CLIENT_NEW_IP_ADDRESS_NOTIFICATION:
```
tomereli, line 152:
```suggestion
cACTION_MONITOR_CLIENT_DHCP_NOTIFICATION:
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -2794,16 +2794,16 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
         break;
     }
     case beerocks_message::ACTION_CONTROL_CLIENT_DHCP_COMPLETE_NOTIFICATION: {
-        auto notification =
+        auto notification_in =
             beerocks_header
                 ->addClass<beerocks_message::cACTION_CONTROL_CLIENT_DHCP_COMPLETE_NOTIFICATION>();
-        if (notification == nullptr) {
+        if (!notification_in) {
```
tomereli, line 2794: Should mention these fixes in the commit message (nitpick)

tomereli (approved):
Mainly nitpick comments so approving. Nice one!

common/beerocks/bcl/source/network/network_utils.cpp
morantr: There is no point sending the arp since any client which will receive it will ignore it because the request has ip of 0.0.0.0.
The reason that the IP is empty is irrelevant for this commit. This commit makes the function robust.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
morantr: It doesn't. the Clang-format changed it.

controller/src/beerocks/master/son_master_thread.cpp
morantr: I changed the variable name from `notification` to 'notification_in` since I had to add new tlv `notification_out`. I seem weird to me to mention it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -38,8 +38,8 @@
 #include <tlvf/wfa_map/tlvSupportedService.h>
 
 // BPL Error Codes
-#include <bpl/bpl_err.h>
 #include <bpl/bpl_cfg.h>
+#include <bpl/bpl_err.h>
```
arnout: I guess this is a clang-format effect... weird...

common/beerocks/bcl/source/network/network_utils.cpp
arnout: I believe the real issue is that with IP address 0.0.0.0, the client will ignore it - as mentioned in the commit message.

But actually, this makes me think: why do we ever need an IP address? AFAIU the only reason we have these IP addresses is to be able to do this ARP, but the IP addresses are non-standard. So if we can simply use the broadcast address, we really don't ever need an IP, and we can always use the broadcast address.

Probably not for this PR though.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -898,7 +898,16 @@ bool network_utils::arp_send(const std::string &iface, const std::string &dst_ip
...
+    // with broadcast IP, so all clients will receive it and answer, but since the request is
+    // being sent to a specific mac address, then only the requested client will answer.
+    uint32_t dst_ip_uint;
+    if (dst_ip.empty()) {
+        dst_ip_uint = 0xFFFFFFFF; // "255.255.255.255" equivalent
```
arnout, line 906: Maybe it's not needed, but I have the habit of putting UL or at least U after the constant. On a 32-bit system this may be undefined behaviour, I think.

common/beerocks/bcl/source/network/network_utils.cpp
vitalybu: I agree with @tomereli's first point - The comment is indeed not clear enough.
You can write something similar to: "If the STA's IP address is unknown, instead of sending an empty value (0.0.0.0), a broadcast address is used (255.255.255.255)".

As for the second part, I don't see any reason to list any clients at this point. 
The new implementation is much better than the old one, even if not all clients respond to broadcast ARP requests.
It was tested already with several devices, but in the following couple of weeks will be tested with many more.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
vitalybu: I think that "NEW_IP_ADDRESS" is good enough.
Today the only supported IP source is indeed a DHCP server, but maybe in the future we may allow setting the IP address via BML.

vitalybu (approved):

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -2820,16 +2820,30 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+                notification_out->mac()  = notification_in->mac();
+                notification_out->ipv4() = notification_in->ipv4();
 
-            if (!database.is_node_wireless(client_mac)) {
+                auto clients_bssid_radio_mac = database.get_node_parent(client_mac);
+                if (clients_bssid_radio_mac.empty()) {
+                    LOG(WARNING) << "Client does not have a valid parent hostap on the database";
+                    return true;
+                }
+                son_actions::send_cmdu_to_agent(src_mac, cmdu_tx, database,
```
arnout, line 2837: I am clearly missing something since this is weird... The master receives a DHCP notification from the agent, and it sends back exactly the same info to the agent...

arnout (approved):
Maybe not for this PR, but I'm thinking that with this change, it would be better / more consistent to remove the ipv4 from cACTION_CONTROL_CLIENT_START_MONITORING_REQUEST and cACTION_CONTROL_CLIENT_RX_RSSI_MEASUREMENT_REQUEST. Either the monitor already has received the DHCP update or it hasn't, but that shouldn't be blocking in either case.

controller/src/beerocks/master/son_master_thread.cpp
morantr: I agree it seems weird, but we have to do it like this since the DHCP event is getting to the son_slave which controls the backhaul manager which is not necessarily the son_slave the client is connected to. The controller knows to which son_slave the client is connected to so it can start the client monitoring on the right monitor thread.
Besides that, it is more convenient that this functionality is centralized on the controller from DEBUG POV.

morantr:
Add a few more commits.
The new commits start with the commit:
"tlvf: Fix enum typo" facd076570171e762fb01661189c75faf135aa61

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -70,13 +70,14 @@ eActionOp_CONTROL:
...
 
   ACTION_CONTROL_CLIENT_START_MONITORING_REQUEST: 100
-  ACTION_CONTROL_CLIENT_STOP_MONITORING_REQUEST: 101
-  ACTION_CONTROL_CLIENT_RX_RSSI_MEASUREMENT_REQUEST: 102
-  ACTION_CONTROL_CLIENT_RX_RSSI_MEASUREMENT_RESPONSE: 103
-  ACTION_CONTROL_CLIENT_RX_RSSI_MEASUREMENT_START_NOTIFICATION: 104
-  ACTION_CONTROL_CLIENT_RX_RSSI_MEASUREMENT_NOTIFICATION: 105
-  ACTION_CONTROL_CLIENT_NO_RESPONSE_NOTIFICATION: 106
-  ACTION_CONTROL_CLIENT_NEW_IP_ADDRESS_NOTIFICATION: 107
+  ACTION_CONTROL_CLIENT_START_MONITORING_RESPONSE: 101
```
vitalybu, line 73: > If not, the client will be disconnected.

Wait.. What? We don't want to disconnect already connected clients!
If the monitor is not ready yet, or does not respond within some (pre-defined) timeout, the message should be sent again.

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -29,6 +29,8 @@ using namespace son;
 #define BEACON_MEASURE_MAX_ATTEMPTS 3
 #define REQUEST_RSSI_MEASUREMENT_MAX_ATTEMPTS 5
 #define REQUEST_RSSI_MEASUREMENT_DELAY 5000
+#define START_MONITORING_RESPONSE_TIMEOUT_MSEC 3000
+#define START_MONITORING_MAX_ATTEMPTS (60 * START_MONITORING_RESPONSE_TIMEOUT_MSEC) // 3 minutes
```
vitalybu, line 33: I think that 3 minutes is too long... 30 seconds should be enough.

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -256,6 +253,45 @@ void association_handling_task::handle_response(std::string mac,
...
+                << "addClass failed for cACTION_CONTROL_CLIENT_START_MONITORING_RESPONSE";
+            return;
+        }
+
+        if (!response->success()) {
+            ++attempts;
+            if (attempts >= max_attempts) {
+                TASK_LOG(ERROR) << "state START_RSSI_MONITORING reached maximum attempts="
+                                << attempts << " aborting task, and disconnecting client!";
+                son_actions::disconnect_client(database, cmdu_tx, sta_mac, original_parent_mac,
```
vitalybu: That's quite a drastic move...
The problem is probably not with the station, but with the monitor (not ready yet?).
@tomereli, @arnout, what do you guys think?

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -256,6 +253,45 @@ void association_handling_task::handle_response(std::string mac,
...
+                ->addClass<beerocks_message::cACTION_CONTROL_CLIENT_START_MONITORING_RESPONSE>();
+        if (!response) {
+            TASK_LOG(ERROR)
+                << "addClass failed for cACTION_CONTROL_CLIENT_START_MONITORING_RESPONSE";
+            return;
+        }
+
+        if (!response->success()) {
+            ++attempts;
+            if (attempts >= max_attempts) {
```
vitalybu: What's wrong with: `if (++attempts >= max_attempts) {`?

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -256,6 +253,45 @@ void association_handling_task::handle_response(std::string mac,
...
+                TASK_LOG(ERROR) << "state START_RSSI_MONITORING reached maximum attempts="
```
vitalybu:
```suggestion
                TASK_LOG(ERROR) << "state START_RSSI_MONITORING reached maximum retries = "
```

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -364,6 +400,20 @@ void association_handling_task::handle_responses_timeout(
     ++attempts;
 
     switch (state) {
+    case START_RSSI_MONITORING: {
+        TASK_LOG(DEBUG) << "response for rx rssi measurement from " << original_parent_mac
+                        << " for sta " << sta_mac << " timed out! attempts=" << attempts;
```
vitalybu, line 410:
```suggestion
                        << " for sta " << sta_mac << " timed out! retries = " << attempts;
```

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -364,6 +400,20 @@ void association_handling_task::handle_responses_timeout(
...
+        if (attempts >= max_attempts) {
+            TASK_LOG(ERROR) << "state START_RSSI_MONITORING reached maximum attempts=" << attempts
```
vitalybu:
```suggestion
            TASK_LOG(ERROR) << "state START_RSSI_MONITORING reached maximum retries = " << attempts
```

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -364,6 +400,20 @@ void association_handling_task::handle_responses_timeout(
...
+                                           eDisconnect_Type_Disassoc, 0);
+            finish();
+        } else {
+            state = REQUEST_RSSI_MEASUREMENT;
+        }
+        break;
+    }
     case REQUEST_RSSI_MEASUREMENT: {
         TASK_LOG(DEBUG) << "response for rx rssi measurement from " << original_parent_mac
                         << " for sta " << sta_mac << " timed out! attempts=" << attempts;
```
vitalybu, line 422:
```suggestion
                        << " for sta " << sta_mac << " timed out! retries = " << attempts;
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
vitalybu: Only by reviewing the following commits, I saw that you actually implemented a retries mechanism.
This should probably be mentioned in the commit message... Still think that disconnecting the station is too harsh.
Left a separate comment there.

vitalybu (changes requested):

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
tomereli: That's what I thought, planned to leave a comment tomorrow. We can't disconnect a connected client because of our internal logic! Where did the "passive" in passive mode go?

controller/src/beerocks/master/tasks/association_handling_task.cpp
tomereli: https://github.com/prplfoundation/prplMesh/pull/789#discussion_r376810230

tomereli (changes requested):
We just can't disconnect a client which was connected before prplmesh started. Actually, we can't disconnect a client ever unless it's part of active steering, and also in this case it's temporary.
This is the opposite of passive.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
vitalybu: So unless anyone objects, let's just log a warning and keep the client connected.
If we'll get additional reports of un-monitored stations in the future, we'll re-think this solution.

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: The maximum waiting for a single response is 3 seconds. 3 minutes is the maximum timeout for repeated requests which is a reasonable time for the wifi to be up.
I think I'll keep it.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
morantr: My thought was that leaving a client unmonitored is bad because it means that the client won't be measured and therefore couldn't be steered. But Ok I'll remove the disconnect and replace it with warning print.

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: I'll remove the disconnect call.
https://github.com/prplfoundation/prplMesh/pull/789#discussion_r377014434

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: will change

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: Not a critic one. I'll prefer staying with "attempts" since it is the variable name

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: https://github.com/prplfoundation/prplMesh/pull/789#discussion_r377016150

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: https://github.com/prplfoundation/prplMesh/pull/789#discussion_r377016150

controller/src/beerocks/master/tasks/association_handling_task.cpp
morantr: https://github.com/prplfoundation/prplMesh/pull/789#discussion_r377016150

vitalybu (approved):

tomereli (approved):

PR #799, opened 2020-02-10
Bugfix - non EasyMesh mode VAP credentials update/tear down

### LINKED ISSUE #797: [BUG] autoconfig: teardown / update VAPs in non-easymesh mode
Currently, the son slave is updating VAP credentials regardless of the operating mode.
This means that prplMesh is *always* tearing down all radios, which is fine in EasyMesh mode BUT not in a product.

BTW - I suggest renaming the "proprietary-mesh" mode to "legacy" or something like that.
We can decide whatever we want since we don't use anything other than the `Multi-AP-Controller-And-Agent` mode. 

## Steps to reproduce

Simply get to operational in UGW/RDKB.
Use hostapd_cli to see that all VAPs are in start_disabled mode, or simply read `/var/run/hostapd-phy0.conf` and see that `start_disabled=1` for all VAPs.

tomereli:
In EasyMesh mode, as part of autoconfiguration, the radio agent sends a
message to the AP manager to update VAP credentials which results with
an updated credentials or a full radio teardown.
In non EasyMesh mode, this should never happen - so add a check for that
and only send the message to the ap manager when in EasyMesh management
mode.

Fixes #797.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4050,7 +4050,22 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
         }
     }
 
-    message_com::send_cmdu(ap_manager_socket, cmdu_tx);
+    ///////////////////////////////////////////////////////////////////
+    // TODO https://github.com/prplfoundation/prplMesh/issues/797
+    //
+    // Short term solution
+    // In non-EasyMesh mode, never modify hostapd configuration
+    // and in this case VAPs credentials
+    //
+    // Long term solution
+    // All EasyMesh VAPs will be stored in the platform DB.
+    // All other VAPs are manual, AKA should not be modified by prplMesh
+    ////////////////////////////////////////////////////////////////////
+    if (platform_settings.management_mode != BPL_MGMT_MODE_NOT_MULTIAP) {
+        message_com::send_cmdu(ap_manager_socket, cmdu_tx);
+    } else {
+        LOG(WARNING) << "non-EasyMesh mode - skip updating VAP credentials";
+    }
 
     // Notify backhaul manager that onboarding has finished (certification flow)
     auto onboarding_finished_notification = message_com::create_vs_message<
--- a/framework/platform/bpl/include/bpl/bpl_cfg.h
+++ b/framework/platform/bpl/include/bpl/bpl_cfg.h
@@ -51,6 +51,12 @@ namespace bpl {
 #define BPL_OPER_MODE_WDS_REPEATER 3
 #define BPL_OPER_MODE_L2NAT_CLIENT 4
 
+/* Platform Management Mode */
+#define BPL_MGMT_MODE_MULTIAP_CONTROLLER_AGENT 0 /* EasyMesh controller and agent */
+#define BPL_MGMT_MODE_MULTIAP_CONTROLLER 1       /* EasyMesh controller */
+#define BPL_MGMT_MODE_MULTIAP_AGENT 2            /* EasyMesh agent */
+#define BPL_MGMT_MODE_NOT_MULTIAP 3              /* Non EasyMesh */
+
 /* Platform Certification Mode */
 #define BPL_CERTIFICATION_MODE_OFF 0
 #define BPL_CERTIFICATION_MODE_ON 1
@@ -259,6 +265,18 @@ int cfg_is_master();
  * @return -1 Error.
  */
 int cfg_get_operating_mode();
+
+/**
+ * Returns the current management mode configuration.
+ *
+ * @return valid possibilities:
+ *   BPL_MGMT_MODE_MULTIAP_CONTROLLER_AGENT,
+ *   BPL_MGMT_MODE_MULTIAP_CONTROLLER,
+ *   BPL_MGMT_MODE_MULTIAP_AGENT,
+ *   BPL_MGMT_MODE_NOT_MULTIAP
+ * @return -1 Error.
+ */
+int cfg_get_management_mode();
 
 /**
  * Returns certification mode value.
--- a/framework/platform/bpl/linux/bpl_cfg.cpp
+++ b/framework/platform/bpl/linux/bpl_cfg.cpp
@@ -83,20 +83,18 @@ int cfg_is_enabled() { return 1; }
 
 int cfg_is_master()
 {
-    std::string mode_str;
-    if (cfg_get_param("management_mode=", mode_str) < 0) {
-        MAPF_ERR("cfg_is_master: Failed to read ManagementMode");
+    switch (cfg_get_management_mode()) {
+    case BPL_MGMT_MODE_MULTIAP_CONTROLLER_AGENT:
+        return 1;
+    case BPL_MGMT_MODE_MULTIAP_CONTROLLER:
+        return 1;
+    case BPL_MGMT_MODE_MULTIAP_AGENT:
+        return 0;
+    case BPL_MGMT_MODE_NOT_MULTIAP:
+        return (cfg_get_operating_mode() == BPL_OPER_MODE_GATEWAY) ? 1 : 0;
+    default:
         return RETURN_ERR;
     }
-
-    if (mode_str == "Multi-AP-Controller-and-Agent" || mode_str == "Multi-AP-Controller")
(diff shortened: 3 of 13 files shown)


framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -51,6 +51,13 @@ namespace bpl {
 #define BPL_OPER_MODE_WDS_REPEATER 3
 #define BPL_OPER_MODE_L2NAT_CLIENT 4
 
+/* Platform Management Mode */
+#define BPL_MGMT_MODE_MULTIAP_CONTROLLER_AGENT 0 /* EasyMesh controller and agent */
+#define BPL_MGMT_MODE_MULTIAP_CONTROLLER 1       /* EasyMesh controller */
+#define BPL_MGMT_MODE_MULTIAP_AGENT 2            /* EasyMesh agent */
+#define BPL_MGMT_MODE_NOT_MULTIAP                                                                  \
+    3 /* Non EasyMesh /*
```
arnout: Why is this split over two lines?

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -34,25 +34,18 @@ int cfg_is_enabled()
...
+    case BPL_MGMT_MODE_MULTIAP_CONTROLLER_AGENT:
+        return 1;
+    case BPL_MGMT_MODE_MULTIAP_CONTROLLER:
+        return 1;
+    case BPL_MGMT_MODE_MULTIAP_AGENT:
+        return 0;
+    case BPL_MGMT_MODE_NOT_MULTIAP:
+        return (cfg_get_operating_mode() == BPL_OPER_MODE_GATEWAY) ? 1 : 0;
+    default:
+        return -1;
```
arnout, line 47: I think it would be better to move this to the base class and make it non-virtual.

Oh, never mind, we don't have that for bpl...

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -83,20 +83,18 @@ int cfg_is_enabled() { return 1; }
...
+    case BPL_MGMT_MODE_MULTIAP_CONTROLLER:
+        return 1;
+    case BPL_MGMT_MODE_MULTIAP_AGENT:
         return 0;
-
-    MAPF_ERR("cfg_is_master: Unexpected management_mode " << mode_str);
-
-    return RETURN_ERR;
+    case BPL_MGMT_MODE_NOT_MULTIAP:
+        return (cfg_get_operating_mode() == BPL_OPER_MODE_GATEWAY) ? 1 : 0;
```
arnout, line 94: Hm, it seems that in MULTIAP mode, the operating mode is redundant then... Actually, AFAICS, the operating mode is only used to determine if we're a gateway or not. So maybe that should be encoded here as well?

Anyway can be handled later.

framework/platform/bpl/include/bpl/bpl_cfg.h
morantr: The last line is messed up.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4050,7 +4050,21 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
         }
     }
 
-    message_com::send_cmdu(ap_manager_socket, cmdu_tx);
+    ///////////////////////////////////////////////////////////////////
+    // TODO https://github.com/prplfoundation/prplMesh/issues/797
+    //
+    // Short term solution
+    // In None EasyMesh mode, never modify hostapd configuration
```
arnout:
```suggestion
    // In non-EasyMesh mode, never modify hostapd configuration
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4050,7 +4050,21 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
+    // and in this case VAPs credentials
+    //
+    // Long term solution
+    // All EasyMesh VAPs will be stored in the platform DB.
+    // All other VAPs are manual, AKA should not be modified by prplMesh
+    ////////////////////////////////////////////////////////////////////
+    if (platform_settings.management_mode != BPL_MGMT_MODE_NOT_MULTIAP) {
+        LOG(INFO) << "EasyMesh mode - really update VAP credentials";
```
arnout: I would rather have a message in the non-EasyMesh case... And put it at WARNING level.

Ideally even only issue the message when the SSIDs are different from the already-configured SSIDs.

arnout (approved):

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -51,6 +51,13 @@ namespace bpl {
 #define BPL_OPER_MODE_WDS_REPEATER 3
 #define BPL_OPER_MODE_L2NAT_CLIENT 4
 
+/* Platform Management Mode */
+#define BPL_MGMT_MODE_MULTIAP_CONTROLLER_AGENT 0 /* EasyMesh controller and agent */
+#define BPL_MGMT_MODE_MULTIAP_CONTROLLER 1       /* EasyMesh controller */
+#define BPL_MGMT_MODE_MULTIAP_AGENT 2            /* EasyMesh agent */
```
morantr, line 57: I wonder why the macro uses the term "multiap" and the comment us "eazymesh".
I think that is a good opportunity to replace it.

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -51,6 +51,13 @@ namespace bpl {
...
+#define BPL_MGMT_MODE_NOT_MULTIAP                                                                  \
```
morantr: It seems weird to mee to put it in the same category of what runs on the platform because basically, the controller and the agent do run on this configuration so this is confusing a little bit.
I think it will be more suitable to give it a new category (passive mode?). But that just me, let see what others will say...

framework/platform/bpl/include/bpl/bpl_cfg.h
tomereli: I would have done this as part of this change but decided not to since it will require updating prplWrt, UGW and RDKB to the new names, and will break everything until it does.
MultiAP is also correct IMO.
I can fix comment to use Multi-AP terminology.

framework/platform/bpl/include/bpl/bpl_cfg.h
tomereli: It has to be in the same category, AKA Management mode, I don't want to introduce yet another category just for this.

framework/platform/bpl/include/bpl/bpl_cfg.h
tomereli: because I accidentally used `/*` instead of `*/` to close the comment.. fixed.

framework/platform/bpl/linux/bpl_cfg.cpp
tomereli: Agree - there is a lot of redundancy here, I actually don't think we need all the modes, and indeed - it can be handled later.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Added a warning.

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -99,6 +99,29 @@ int cfg_is_master()
     return RETURN_ERR;
 }
 
+int cfg_get_management_mode()
+{
+    int retVal = 0;
```
vitalybu: Shouldn't this be initialized to `-1`?
Otherwise this function will return `0 == BPL_MGMT_MODE_MULTIAP_CONTROLLER_AGENT` for an unexpected mode.
```suggestion
    int retVal = -1;
```

vitalybu (changes requested):

framework/platform/bpl/linux/bpl_cfg.cpp
tomereli: Nice catch!

vitalybu (approved):

PR #800, opened 2020-02-10
CONTRIBUTING.md: add a comment about the "MergeWhenReady" tag

rmelotte:
Added a comment about the usage of the "[merge when ready](https://github.com/apps/mergewhenready)" feature.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -271,6 +271,9 @@ Example use cases:
 
 If a PR is marked "don't merge" and it becomes ready (i.e. you did the necessary fixups and rebased on master), please remember to remove the tag again.
 
+When your PR is ready to be merged (you have the required number of reviews and you addressed all the review comments), you don't need to wait for the CI to finish.
+You can add the "MergeWhenReady" tag, and Github will automatically merge the pull request once the CI has finished and is passing.
+
 ### Definition of done
 
 Before a pull request can be merged, it must be considered "Done".


CONTRIBUTING.md
```diff
@@ -271,6 +271,9 @@ Example use cases:
 
 If a PR is marked "don't merge" and it becomes ready (i.e. you did the necessary fixups and rebased on master), please remember to remove the tag again.
 
+When your PR is ready to be merged (you have the required number of reviews and you addressed all the review comments), you don't need to wait for the CI to finish.
+You can add the "MergeWhenReady" tag, and Github will automatically merge the pull request once the CI has finished and is passing.
```
arnout, line 275: Actually, my idea was to remove the use of the "don't merge" tag completely, and instead rely on the MergeWhenReady tag to mark it as ready to merge. So it's a bit more invasive change to our workflow.

arnout (changes requested):

tomereli:
Seems that this feature is not working..

rmelotte:
[This issue](https://github.com/fossapps/MergeWhenReady/issues/2) mentions the tag has to be added /after/ the PR has the required review (which is accidentally the workflow I had in mind and that s described in the PR).
We can try again by adding the tag after the reviews.

rmelotte:
If it still doesn't work we could also try this one, it looks more actively maintained, and also more flexible:
https://github.com/palantir/bulldozer#deployment

morantr (approved):

arnout:
> If it still doesn't work we could also try this one, it looks more actively maintained, and also more flexible:
> https://github.com/palantir/bulldozer#deployment

Although it looks interesting, it seems to require us to set up a server that can be reached by github... That's a bit too much IMO.

@tomereli found the [merge-me action](https://github.com/marketplace/actions/merge-me). That looks promising. I'll give it a go.

BTW I tried another PR with the workflow: get two approvals, push a new version, while building add the tag -> still didn't work.

arnout:
Looking at the merge-me action: it also seems pretty heavyweight. I don't know if it's really worth the trouble...

arnout:
Mergify is now enabled, and has a different workflow (no MergeWhenReady tag, it's automatic). So closing this PR.

PR #802, opened 2020-02-10
controller: ucc listener add dev_get_param support

### LINKED ISSUE #798: [TASK] Support agent and controller UCC commands in parallel
## Description

Commit f68d51717f235f4a6be5ffdd3a104e72aa90cd96 in #768 unified the UCC listener ports for controller and agent certification.
This means that we are not able to send controller and agent UCC commands at the same run, which is a problem for unit testing - we want to be able to run test 5.4.2 for example in gateway mode (single device) - this is needed for UTAF.
The problem in this case is that we send controller commands to set up credentials, then trigger autoconfig renew. Then we need to send an agent command `dev_get_parameter` to see that the VAPs were actually configured.

## Exit criteria
Both controller and agent UCC commands can be served in parallel

tomereli:
Add dev_get_param support for getting the mac address for a VAP.
This can be used in tests to verify VAP credentials were set as
expected.

Fixes #798

### DIFF
--- a/controller/src/beerocks/master/controller_ucc_listener.cpp
+++ b/controller/src/beerocks/master/controller_ucc_listener.cpp
@@ -49,6 +49,31 @@ void controller_ucc_listener::clear_configuration() { m_database.clear_bss_info_
 bool controller_ucc_listener::handle_dev_get_param(
     std::unordered_map<std::string, std::string> &params, std::string &value)
 {
+    if (params["parameter"] == "macaddr" || params["parameter"] == "bssid") {
+        if (params.find("ruid") == params.end()) {
+            value = "missing ruid";
+            return false;
+        }
+        if (params.find("ssid") == params.end()) {
+            value = "missing ssid";
+            return false;
+        }
+        auto ruid = net::network_utils::mac_to_string(std::stoull(params["ruid"], nullptr, 16));
+        auto ssid = params["ssid"];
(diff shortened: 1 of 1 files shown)


morantr (approved):

controller/src/beerocks/master/controller_ucc_listener.cpp
```diff
@@ -49,6 +49,31 @@ void controller_ucc_listener::clear_configuration() { m_database.clear_bss_info_
 bool controller_ucc_listener::handle_dev_get_param(
     std::unordered_map<std::string, std::string> &params, std::string &value)
```
adam1985d, line 50: any reason not to use const for `std::unordered_map<std::string, std::string> &params`?
nitpick

adam1985d (approved):
looks good

tomereli:
It doesn't seem that the "merge when ready" app is really working..

arnout (approved):
I hate it that this is so similar to the one in the agent, but not the same so it can't be refactored... But can't be helped.

controller/src/beerocks/master/controller_ucc_listener.cpp
arnout: Good point, but a separate issue.

rmelotte (approved):

PR #804, opened 2020-02-10
bpl: read wireless configuration by interface name instead of index

### LINKED ISSUE #574: [TASK] Move slave radio enumeration to bpl
## Description

Move the configuration of which slaves to start (wlan0, wlan1, wlan2, ...) to BPL instead of getting it from a configuration file.

- non hostap_iface configuration is moved to #803
- backhaul configuration is moved to #803

## Exit criteria

- bpl_cfg has functions to get slave config;
- hostap_iface is removed from the config file.
- when this task is completed, OpenWrt must be updated as well as described in #787

## Detailed description

Semantically, the configuration of which radios to use in prplMesh and which interface to use for backhaul should be part of the platform (i.e. retrieved from UCI), not in the internal configuration file. CFR. #542 

This is made more difficult because the slaves and the backhaul manager are started before a connection is made with the platform manager. Therefore, first, the code flow has to be refactored to start the platform manager first, wait for it to become online, and only then start the slaves and backhaul manager.

### sub-tasks ###
phase 1:
- [x] add hostap_iface param to the /etc/config/prplmesh
- [x] provide infrastructure to get the hostap_iface from the prplmesh config for a given radio/slave
- [x] provide API to get all hostap-interfaces (slave_agent_main requires to know all the configured hostap-interfaces of the configured slaves before starting the slaves)
- [x] propagate the read hostap_iface information read to the son_slave, platform_manger and backhaul_manager
- [x] remove the hostap_iface from the beerocks_config
- [x] add a similar PR to prplmeshOpenWrt
phase 2:
- [x] move all radio/agent configuration from beerocks_agent.conf to the prplmesh config file
phase 3:
- [x] move backhaul configuration from beerocks_agent.conf to the prplmesh config file

adam1985d:
This PR breaks the dependency of BPL in wireless radio section naming in the UCI config.
The new implementation looks-up the interface name in the wireless config.
If the param is requested from the main radio - a search of the param value is done in the `wireless.<device>.param` where <device> is the device name found in the section where the interface name was found.
```
e.g.
lookup **great_iface_name**:
wireless.**adam**.ifname=**great_iface_name**->
wireless.**adam**.device=**adam_device** ->
read: wireless.**adam_device**.<param>
```
for VAPs, the param is searched in the current section, but again - the section is located using the interface name:
```
lookup **great_iface_name**:
wireless.**adam**.ifname=**great_iface_name**->
read: wireless.**adam**.<param>
```

this PR is rebased to the following PR:
https://github.com/prplfoundation/prplMesh/pull/785

Fixes #574

### DIFF
--- a/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
+++ b/agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
@@ -99,16 +99,9 @@ static bool fill_platform_settings(
     /* update message */
     msg->wlan_settings().band_enabled = params.enabled;
     msg->wlan_settings().channel      = params.channel;
-    string_utils::copy_string(msg->wlan_settings().ssid, params.ssid,
-                              beerocks::message::WIFI_SSID_MAX_LENGTH);
-    string_utils::copy_string(msg->wlan_settings().pass, params.passphrase,
-                              beerocks::message::WIFI_PASS_MAX_LENGTH);
-    string_utils::copy_string(msg->wlan_settings().security_type, params.security,
-                              beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH);
 
     LOG(DEBUG) << "wlan settings:"
-               << " ssid=" << msg->wlan_settings().ssid
-               << " sec=" << msg->wlan_settings().security_type << " pass=***"
+               << " band_enabled=" << string_utils::bool_str(msg->wlan_settings().band_enabled)
                << " channel=" << int(msg->wlan_settings().channel);
 
     // initialize wlan params cache
@@ -122,12 +115,6 @@ static bool fill_platform_settings(
 
     params_ptr->band_enabled = params.enabled;
     params_ptr->channel      = params.channel;
-    string_utils::copy_string(params_ptr->ssid, params.ssid,
-                              beerocks::message::WIFI_SSID_MAX_LENGTH);
-    string_utils::copy_string(params_ptr->pass, params.passphrase,
-                              beerocks::message::WIFI_PASS_MAX_LENGTH);
-    string_utils::copy_string(params_ptr->security_type, params.security,
-                              beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH);
 
     iface_wlan_params_map[iface_name] = params_ptr;
 
@@ -615,28 +602,6 @@ bool main_thread::wlan_params_changed_check()
             LOG(DEBUG) << "channel changed";
             wlan_params_changed = true;
         }
-        if (std::string(elm.second->ssid).compare(std::string(params.ssid))) {
-            LOG(DEBUG) << "ssid changed, cached=" << std::string(elm.second->ssid) << ", new="
-                       << std::string(params.ssid, beerocks::message::WIFI_SSID_MAX_LENGTH);
-            string_utils::copy_string(elm.second->ssid, params.ssid,
-                                      beerocks::message::WIFI_SSID_MAX_LENGTH);
-            wlan_params_changed = true;
-        }
-        if (std::string(elm.second->pass).compare(std::string(params.passphrase))) {
-            LOG(DEBUG) << "pass changed";
-            string_utils::copy_string(elm.second->pass, params.passphrase,
-                                      beerocks::message::WIFI_PASS_MAX_LENGTH);
-            wlan_params_changed = true;
-        }
-        if (std::string(elm.second->security_type).compare(std::string(params.security))) {
-            LOG(DEBUG) << "security_type changed, cached=" << std::string(elm.second->security_type)
-                       << ", new="
-                       << std::string(params.security,
-                                      beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH);
-            string_utils::copy_string(elm.second->security_type, params.security,
-                                      beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH);
-            wlan_params_changed = true;
-        }
 
         if (wlan_params_changed) {
             any_slave_changed = true;
@@ -649,13 +614,6 @@ bool main_thread::wlan_params_changed_check()
 
             notification->wlan_settings().band_enabled = elm.second->band_enabled;
             notification->wlan_settings().channel      = elm.second->channel;
-            string_utils::copy_string(notification->wlan_settings().ssid, elm.second->ssid,
-                                      beerocks::message::WIFI_SSID_MAX_LENGTH);
-            string_utils::copy_string(notification->wlan_settings().pass, elm.second->pass,
-                                      beerocks::message::WIFI_PASS_MAX_LENGTH);
-            string_utils::copy_string(notification->wlan_settings().security_type,
-                                      elm.second->security_type,
-                                      beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH);
 
             Socket *sd = get_slave_socket_from_hostap_iface_name(elm.first);
             if (!sd) {
--- a/framework/platform/bpl/include/bpl/bpl_cfg.h
+++ b/framework/platform/bpl/include/bpl/bpl_cfg.h
@@ -220,15 +220,6 @@ struct BPL_WLAN_PARAMS {
 
     /* Wi-Fi Channel (0 for ACS) */
     int channel;
-
-    /* Wi-Fi SSID */
-    char ssid[BPL_SSID_LEN];
-
-    /* Wi-Fi KeyPassphrase */
-    char passphrase[BPL_PASS_LEN];
-
-    /* Wi-Fi Securirt Mode */
-    char security[BPL_SEC_LEN];
 };
 
 /**
--- a/framework/platform/bpl/uci/cfg/bpl_cfg.cpp
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg.cpp
@@ -201,21 +201,6 @@ int cfg_get_wifi_params(const char iface[BPL_IFNAME_LEN], struct BPL_WLAN_PARAMS
         MAPF_INFO("UCI: radio" << index << ": channel is not configured assuming auto\n");
     }
 
-    char ssid[MAX_UCI_BUF_LEN] = {0}, security[MAX_UCI_BUF_LEN] = {0},
-         passphrase[MAX_UCI_BUF_LEN] = {0};
-    retVal |= cfg_uci_get_wireless(TYPE_VAP, index, "ssid", ssid);
-    retVal |= cfg_uci_get_wireless(TYPE_VAP, index, "wav_security_mode", security);
-    std::string mode = std::string(security);
-    if (mode == BPL_WLAN_SEC_WEP64_STR || mode == BPL_WLAN_SEC_WEP128_STR) {
-        retVal |= cfg_get_wep_key(index, -1, passphrase);
-    } else if (mode != BPL_WLAN_SEC_NONE_STR) {
-        retVal |= cfg_uci_get_wireless(TYPE_VAP, index, "key", passphrase);
-    }
-
-    utils::copy_string(wlan_params->ssid, ssid, BPL_SSID_LEN);
-    utils::copy_string(wlan_params->security, security, BPL_SEC_LEN);
-    utils::copy_string(wlan_params->passphrase, passphrase, BPL_PASS_LEN);
-
     return retVal;
 }
 
--- a/framework/platform/bpl/uci/cfg/bpl_cfg.cpp
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg.cpp
@@ -175,21 +175,15 @@ int cfg_get_device_info(BPL_DEVICE_INFO *device_info) { return 0; }
 
 int cfg_get_wifi_params(const char iface[BPL_IFNAME_LEN], struct BPL_WLAN_PARAMS *wlan_params)
 {
-    int retVal = 0;
-    int index  = 0;
-
     if (!iface || !wlan_params) {
+        MAPF_ERR("cfg_get_wifi_params: invalid input: iface = "
+                 << intptr_t(iface) << " wlan_params = " << intptr_t(wlan_params));
         return RETURN_ERR;
-    }
-
-    retVal = cfg_get_index_from_interface(iface, &index);
-    if (retVal) {
-        return retVal;
     }
 
     // The UCI "disabled" setting is optional, defaults to false if not present
     bool disabled = false;
-    cfg_uci_get_wireless_bool(TYPE_RADIO, index, "disabled", &disabled);
+    cfg_uci_get_wireless_bool(TYPE_RADIO, iface, "disabled", &disabled);
     wlan_params->enabled = !disabled;
 
     // The UCI "channel" setting is not documented as optional, but for Intel
@@ -197,11 +191,11 @@ int cfg_get_wifi_params(const char iface[BPL_IFNAME_LEN], struct BPL_WLAN_PARAMS
     // fail when wifi still works fine, so default to "auto" (0) and if
     // can't get the channel from UCI just move on.
     wlan_params->channel = 0;
-    if (!cfg_get_channel(index, &wlan_params->channel)) {
-        MAPF_INFO("UCI: radio" << index << ": channel is not configured assuming auto\n");
+    if (!cfg_get_channel(iface, &wlan_params->channel)) {
+        MAPF_INFO("UCI: interface " << iface << ": channel is not configured assuming auto\n");
     }
 
-    return retVal;
+    return RETURN_OK;
 }
 
 int cfg_get_backhaul_params(int *max_vaps, int *network_enabled, int *preferred_radio_band)
@@ -291,6 +285,7 @@ int cfg_get_sta_iface(const char iface[BPL_IFNAME_LEN], char sta_iface[BPL_IFNAM
         return RETURN_ERR;
     }
 
+    //TODO: remove dependency in wireless section naming in UCI #801
     int index = -1;
     if (cfg_get_index_from_interface(iface, &index) == RETURN_ERR) {
         MAPF_ERR("cfg_get_sta_iface: Failed to get radio index from iface");
--- a/framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
@@ -82,16 +82,19 @@ int cfg_get_prplmesh_param_int(const std::string &param, int *buf)
     return RETURN_OK;
 }
 
-int cfg_get_channel(int index, int *channel)
+int cfg_get_channel(const std::string &interface_name, int *channel)
 {
     if (!channel) {
         return RETURN_ERR;
     }
 
     char channel_num[MAX_UCI_BUF_LEN] = {0};
-    int status = cfg_uci_get_wireless(TYPE_RADIO, index, "channel", channel_num);
-    if (status == RETURN_ERR)
+    char ifname[BPL_IFNAME_LEN]       = {0};
+
+    utils::copy_string(ifname, interface_name.c_str(), BPL_IFNAME_LEN);
+    if (cfg_uci_get_wireless_from_ifname(TYPE_RADIO, ifname, "channel", channel_num) != RETURN_OK) {
         return RETURN_ERR;
+    }
 
     std::string channel_str(channel_num);
     if (!channel_str.compare("auto")) {
@@ -103,7 +106,7 @@ int cfg_get_channel(int index, int *channel)
     return RETURN_OK;
 }
 
-int cfg_get_wep_key(int index, int keyIndex, char *key)
+int cfg_get_wep_key(const std::string &interface_name, int keyIndex, char *key)
 {
     /*TODO: implement using d/s-pal apis*/
     return RETURN_OK;
--- a/framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
@@ -7,6 +7,9 @@
  */
 
 #include "bpl_cfg_uci.h"
+#include "../../common/utils/utils.h"
+
+#include <string>
 
 extern "C" {
 #include <uci.h>
@@ -160,6 +163,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
     }
 
     return status;
+}
+
+int cfg_uci_get_wireless_from_ifname(enum paramType type, const char *interface_name,
+                                     const char param[], char *value)
+{
+    struct uci_context *ctx = uci_alloc_context();
+    if (!ctx) {
+        ERROR("%s, uci alloc context failed!\n", __FUNCTION__);
+        return RETURN_ERR;
+    }
+
+    char lookup_str[MAX_UCI_BUF_LEN] = "wireless";
+    struct uci_ptr ptr;
+    if ((uci_lookup_ptr(ctx, &ptr, lookup_str, true) != UCI_OK)) {
+        ERROR("%s, uci lookup failed!\n", __FUNCTION__);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
+
+    if (!ptr.p) {
+        ERROR("%s, returned pointer is null\n", __FUNCTION__);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
+
+    bool is_section_found = false;
+    struct uci_package *p = ptr.p;
+    struct uci_element *e = nullptr;
+    struct uci_element *n = nullptr;
+    struct uci_section *s = nullptr;
+    // Iterate over all wireless sections in the UCI DB
+    uci_foreach_element(&p->sections, e)
+    {
+        s = uci_to_section(e);
+
+        if (strncmp(s->type, "wifi-iface", MAX_UCI_BUF_LEN))
+            continue;
+
+        // Iterate over all the options in the section
+        uci_foreach_element(&s->options, n)
+        {
+            struct uci_option *o = uci_to_option(n);
+
+            if (o->type != UCI_TYPE_STRING)
+                continue;
+
+            if (strncmp(n->name, "ifname", MAX_UCI_BUF_LEN))
+                continue;
+
+            if (strncmp(interface_name, o->v.string, MAX_UCI_BUF_LEN))
+                continue;
+
+            // We reached the section containing the requested ifname
+            is_section_found = true;
+            break;
+        }
+
+        if (is_section_found) {
+            break;
+        }
+    }
+
+    //if interface not found in wireless
+    if (!is_section_found) {
+        ERROR("%s, interface(%s) not found", __FUNCTION__, interface_name);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
+
+    if (type == TYPE_RADIO) {
+        // create path to the param in the device: wireless.<device>.param
+        bool device_option_exist = false;
+        std::string path_str;
+
+        uci_foreach_element(&s->options, n)
+        {
+            struct uci_option *o = uci_to_option(n);
+
+            if (strncmp(n->name, "device", MAX_UCI_BUF_LEN) == 0 && o->type == UCI_TYPE_STRING) {
+                path_str =
+                    std::string("wireless." + std::string(o->v.string) + "." + std::string(param));
+                device_option_exist = true;
+                break;
+            }
+        }
+        uci_free_context(ctx);
+
+        if (device_option_exist) {
+            // radio not found
+            ERROR("%s device option not found\n", __func__);
+            return RETURN_ERR;
+        }
+
+        char path[MAX_UCI_BUF_LEN] = "";
+        utils::copy_string(path, path_str.c_str(), MAX_UCI_BUF_LEN);
+
+        if (cfg_uci_get(path, value, MAX_UCI_BUF_LEN) != RETURN_OK) {
+            ERROR("%s option N/A. path=%s\n", __func__, path);
+        }
+
+        return RETURN_OK;
+
+    } else { //* type == TYPE_VAP *
+        // read the value from the selected section
+        // get param in the selected section
+        uci_foreach_element(&s->options, n)
+        {
+            struct uci_option *o = uci_to_option(n);
+            // if param is found in options
+            if (strncmp(n->name, param, MAX_UCI_BUF_LEN) == 0 && o->type == UCI_TYPE_STRING) {
+                strncpy_s(value, MAX_UCI_BUF_LEN, o->v.string, MAX_UCI_BUF_LEN - 1);
+                uci_free_context(ctx);
+                return RETURN_OK;
+            }
+        }
+
+        // param not found in option
+        ERROR("%s, interface(%s) found but param(%s) isn't configured\n", __FUNCTION__,
+              interface_name, param);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
 }
 
 int cfg_uci_get_wireless_radio_idx(const char *interfaceName, int *radio_index)
--- a/framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
@@ -69,12 +69,13 @@ int cfg_uci_get(char *path, char *value, size_t length)
     return RETURN_OK;
 }
 
-int cfg_uci_get_wireless_int(enum paramType type, int index, const char param[], int *value)
+int cfg_uci_get_wireless_int(enum paramType type, const char *interface_name, const char param[],
+                             int *value)
 {
     int status;
     char val[MAX_UCI_BUF_LEN] = "";
 
-    status = cfg_uci_get_wireless(type, index, param, val);
+    status = cfg_uci_get_wireless_from_ifname(type, interface_name, param, val);
     if (status == RETURN_ERR)
         return RETURN_ERR;
 
@@ -439,12 +440,12 @@ int cfg_uci_get_wireless_idx(char *interfaceName, int *rpc_index)
     return RETURN_ERR;
 }
 
-int cfg_uci_get_wireless_bool(enum paramType type, int index, const char param[], bool *value)
+int cfg_uci_get_wireless_bool(enum paramType type, const char *interface_name, const char param[],
+                              bool *value)
 {
-    int status;
     int res;
 
-    status = cfg_uci_get_wireless_int(type, index, param, &res);
+    int status = cfg_uci_get_wireless_int(type, interface_name, param, &res);
     if (status == RETURN_ERR)
         return RETURN_ERR;
 
--- a/framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
@@ -91,6 +91,9 @@ int cfg_uci_get_radio_param(int index, const char param[], char *value, size_t b
 int cfg_uci_get_wireless_radio_idx(const char *interfaceName, int *radio_index);
 int cfg_uci_get_radio_param_ulong(int index, const char param[], unsigned long *value);
 
+int cfg_uci_get_wireless_from_ifname(enum paramType type, const char *interface_name,
+                                     const char param[], char *value);
+
 } // namespace bpl
 } // namespace beerocks
 
--- a/framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
@@ -84,7 +84,8 @@ namespace bpl {
 int cfg_uci_get_wireless_idx(char *interfaceName, int *rpc_index);
 int cfg_uci_get(char *path, char *value, size_t length);
 int cfg_uci_get_wireless(enum paramType type, int index, const char param[], char *value);
-int cfg_uci_get_wireless_bool(enum paramType type, int index, const char param[], bool *value);
+int cfg_uci_get_wireless_bool(enum paramType type, const char *interface_name, const char param[],
+                              bool *value);
 
 int cfg_uci_get_radio_param_int(int index, const char param[], int *value);
 int cfg_uci_get_radio_param(int index, const char param[], char *value, size_t buf_len);
--- a/common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_common.h
+++ b/common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_common.h
@@ -114,9 +114,6 @@ typedef struct sPlatformSettings {
 typedef struct sWlanSettings {
     uint8_t band_enabled;
     uint8_t channel;
-    char ssid[beerocks::message::WIFI_SSID_MAX_LENGTH];
-    char pass[beerocks::message::WIFI_PASS_MAX_LENGTH];
-    char security_type[beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH];
     void struct_swap(){
     }
     void struct_init(){
--- a/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
+++ b/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
@@ -105,15 +105,6 @@ sWlanSettings:
   band_enabled:
     _type: uint8_t
   channel: uint8_t 
-  ssid:
-    _type: char    
-    _length: [ "beerocks::message::WIFI_SSID_MAX_LENGTH" ]
-  pass:
-    _type: char    
-    _length: [ "beerocks::message::WIFI_PASS_MAX_LENGTH" ]
-  security_type:
-    _type: char    
-    _length: [ "beerocks::message::WIFI_SECURITY_TYPE_MAX_LENGTH" ]
 
 sApSetRestrictedFailsafe:
   _type: struct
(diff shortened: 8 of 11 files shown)


common/beerocks/bcl/include/bcl/beerocks_async_work_queue.h
```diff
@@ -20,7 +20,7 @@ namespace beerocks {
 
 class async_work_queue : public thread_base {
 public:
-    async_work_queue()  = default;
+    async_work_queue() = default;
```
vitalybu: All non feed related files should be removed from this commit, as they are clang related issues, and were already fixed in the master.

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -498,6 +498,26 @@ int main(int argc, char *argv[])
...
 
+    beerocks::bpl::BPL_WLAN_IFACE interfaces[beerocks::IRE_MAX_SLAVES] = {0};
+    int num_of_interfaces                                              = beerocks::IRE_MAX_SLAVES;
+    if (beerocks::bpl::cfg_get_all_prplmesh_wifi_interfaces(interfaces, &num_of_interfaces)) {
+        std::cout << "failed to read interfaces map" << std::endl;
+        return 1;
+    }
+
+    //create unordered_map of interfaces
+    std::unordered_map<int, std::string> interfaces_map;
```
vitalybu: If the map is empty at the end of this loop, maybe there's no reason to continue...

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -182,26 +182,35 @@ static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_
...
+        if (hostap_iface.empty()) {
+            continue;
+        }
+        // check if slave is still running
+        std::string base_name = std::string(BEEROCKS_AGENT) + "_" + hostap_iface;
+        std::string temp_path(beerocks_slave_temp_path);
+        if (!beerocks::os_utils::is_pid_running(temp_path, base_name)) {
+            //start new slave process
+            std::string file_name = "./" + std::string(BEEROCKS_AGENT);
+            if (access(file_name.c_str(), F_OK) == -1) //file does not exist in current location
```
vitalybu:
```suggestion
            if (access(file_name.c_str(), F_OK) == -1) { //file does not exist in current location
```

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -182,26 +182,35 @@ static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_
...
+            {
```
vitalybu:
```suggestion
```

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -552,24 +550,31 @@ int main(int argc, char *argv[])
...
-            {
-                file_name = BEEROCKS_BIN_PATH + std::string(BEEROCKS_AGENT);
-            }
-            std::string cmd = file_name + " -i " + beerocks_slave_conf.hostap_iface[slave_num];
-            beerocks::SYSTEM_CALL(cmd, 0, true);
+        //start new slave process
+        LOG(INFO) << "Starting slave for iface '" << hostap_iface << "'";
+
+        std::string file_name = "./" + std::string(BEEROCKS_AGENT);
+        if (access(file_name.c_str(), F_OK) == -1) //file does not exist in current location
```
vitalybu:
```suggestion
        if (access(file_name.c_str(), F_OK) == -1) { //file does not exist in current location
```

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -552,24 +550,31 @@ int main(int argc, char *argv[])
...
+        {
```
vitalybu:
```suggestion
```

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -214,15 +214,6 @@ struct BPL_WLAN_PARAMS {
...
     int channel;
-
-    /* Wi-Fi SSID */
-    char ssid[BPL_SSID_LEN];
-
-    /* Wi-Fi KeyPassphrase */
-    char passphrase[BPL_PASS_LEN];
-
-    /* Wi-Fi Securirt Mode */
-    char security[BPL_SEC_LEN];
```
vitalybu, line 231: Are you sure it's the right thing to do? We won't need it in the future?
I would leave it for now anyway, as it has nothing to do with the current PR.

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
```diff
@@ -99,16 +99,9 @@ static bool fill_platform_settings(
     /* update message */
     msg->wlan_settings().band_enabled = params.enabled;
     msg->wlan_settings().channel      = params.channel;
-    string_utils::copy_string(msg->wlan_settings().ssid, params.ssid,
```
vitalybu, line 102: Some as the previous commit.
I'm not sure this information won't be used in the future, and it's not part of the current PR anyway.
Should be reverted.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+                                     char *value)
+{
+    bool is_section_found = false;
+    struct uci_ptr ptr;
+    struct uci_context *ctx;
+    struct uci_package *p = NULL;
+    struct uci_element *e;
+    struct uci_section *s;
+    struct uci_element *n;
+    char lookup_str[MAX_UCI_BUF_LEN] = "wireless";
```
vitalybu: We're in a C++ world, why not define the variables where you need them?
Especially if you name them as `p`, `e`, `s` & `n`...

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+    ctx = uci_alloc_context();
+    if (!ctx)
+        return RETURN_ERR;
```
vitalybu: For example, I would define the `ctx` here: `uci_context *ctx = uci_alloc_context();`
But if you don't, you can at least allocate and validate on the same line: `if (!(ctx = uci_alloc_context())) {`
Lastly, always surround `if` statements with `{ }`, even one-liners, as at some point someone can replace your code with a multi-line macro...

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+    if ((uci_lookup_ptr(ctx, &ptr, lookup_str, true) != UCI_OK)) {
```
vitalybu, line 180: Here I would define the `ptr` and `lookup_str` variables:
```suggestion
    struct uci_ptr ptr;
    const char lookup_str[MAX_UCI_BUF_LEN] = "wireless";
    if ((uci_lookup_ptr(ctx, &ptr, lookup_str, true) != UCI_OK)) {
```

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+        ERROR("%s, uci lookup failed!\n", __FUNCTION__);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
+
+    p = ptr.p;
```
vitalybu: Probably should make sure that `ptr` and `ptr.p` are valid.
```C++
if (!ptr || !ptr.p) { ... }
```

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+    uci_foreach_element(&p->sections, e)
```
vitalybu, line 198: This entire code block requires some commenting.
For example - `// Iterate over all "wireless." sections in the UCI database`

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+    if (type == TYPE_RADIO) {
+        //get radio name from .device param value
+        char deviceName[MAX_UCI_BUF_LEN];
+        deviceName[0] = '\0';
+        uci_foreach_element(&s->options, n)
+        {
+            struct uci_option *o = uci_to_option(n);
+
+            if (strncmp(n->name, "device", MAX_UCI_BUF_LEN) == 0 && o->type == UCI_TYPE_STRING) {
```
vitalybu, line 245: Better check the type first, for two reasons:
1. Native type comparison is far more efficient than string comparison.
2. If the type is not `UCI_TYPE_STRING`, your code will `strncmp()` some unknown type.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+        }
+    }
+
+    if (type == TYPE_RADIO) {
+        //get radio name from .device param value
+        char deviceName[MAX_UCI_BUF_LEN];
+        deviceName[0] = '\0';
+        uci_foreach_element(&s->options, n)
+        {
+            struct uci_option *o = uci_to_option(n);
```
vitalybu, line 243: Can `o` be `NULL`?

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,128 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+        if (is_section_found) {
+            break;
+        }
+    }
+
+    if (type == TYPE_RADIO) {
+        //get radio name from .device param value
+        char deviceName[MAX_UCI_BUF_LEN];
+        deviceName[0] = '\0';
```
vitalybu: Why not initialize the entire array to `{}` or `{0}` on definition (one line above)?

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -157,33 +157,30 @@ int cfg_get_device_info(BPL_DEVICE_INFO *device_info) { return 0; }
 
 int cfg_get_wifi_params(const char iface[BPL_IFNAME_LEN], struct BPL_WLAN_PARAMS *wlan_params)
 {
-    int retVal = 0;
-    int index  = 0;
-
     if (!iface || !wlan_params) {
+        MAPF_ERR("cfg_get_wifi_params: invalid input: iface=" << intptr_t(iface) << " wlan_params="
```
vitalybu:
```suggestion
        MAPF_ERR("cfg_get_wifi_params: invalid input: iface = " << intptr_t(iface) << " wlan_params = "
```

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -157,33 +157,30 @@ int cfg_get_device_info(BPL_DEVICE_INFO *device_info) { return 0; }
...
+                                                              << intptr_t(wlan_params) << "\n");
```
vitalybu:
```suggestion
                                                              << intptr_t(wlan_params));
```

vitalybu (dismissed):

common/beerocks/bcl/include/bcl/beerocks_async_work_queue.h
adam1985d: I will rebase again to master

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: would you consider this case as error or return 0?

framework/platform/bpl/include/bpl/bpl_cfg.h
adam1985d: this information isn't used at all - although read periodically.
it is checked for change and the agent is alerted, but the agent ignores this information and only handles a change in `band_enabled`.

there is a task to remove the rest of the settings read from wireless config from there and read using NL message or similar.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: ok - changed the order of checks

agent/src/beerocks/slave/beerocks_slave_main.cpp
vitalybu: What's the point to continue if there are no interfaces?

vitalybu: Log an INFO and return 0

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: it is an offset calculation back inside the current line, so I think no.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: added more comments

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: fixed

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: fixed

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: fixed

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: fixed

framework/platform/bpl/include/bpl/bpl_cfg.h
vitalybu: Reading this information in a different way is fine, but still it feels like this change should not be part of this PR.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+    char lookup_str[MAX_UCI_BUF_LEN] = "wireless";
+    struct uci_ptr ptr;
+    if ((uci_lookup_ptr(ctx, &ptr, lookup_str, true) != UCI_OK)) {
+        ERROR("%s, uci lookup failed!\n", __FUNCTION__);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
+
+    if (!ptr.p) {
```
vitalybu, line 186: `ptr` should be initialized to `nullptr` and checked for validity here (mentioned this in one of the previous comments)

adam1985d: the naming here is a bit confusing.
although named ptr, it is actually a statically defined struct with set of pointers and other information on the data:
```
struct uci_ptr
{
	enum uci_type target;
	enum {
		UCI_LOOKUP_DONE =     (1 << 0),
		UCI_LOOKUP_COMPLETE = (1 << 1),
		UCI_LOOKUP_EXTENDED = (1 << 2),
	} flags;

	struct uci_package *p;
	struct uci_section *s;
	struct uci_option *o;
	struct uci_element *last;

	const char *package;
	const char *section;
	const char *option;
	const char *value;
};
```

agent/src/beerocks/slave/beerocks_slave_main.cpp
adam1985d: fixed

agent/src/beerocks/slave/platform_manager/platform_manager_thread.cpp
tomereli: Since currently it is not used I don't think there's a problem removing it.
Also I don't see why prplmesh should ever need to read the SSID from BPL. Same for the other deleted configuration parameters.
@arnout what do you think?

framework/platform/bpl/include/bpl/bpl_cfg.h
tomereli: As I recall, there was another reason apart from the fact this is not used which is why @adam1985d removed these - perhaps it should be explained in the commit message.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
     return status;
 }
 
+int cfg_uci_get_wireless_from_ifname(enum paramType type, char *interface_name, const char param[],
```
tomereli: In general, the commit message should not use the word "break" unless its doing something bad :)
I think a better way to phrase this is "This API removes the unnecessary dependency ..."

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+                                     char *value)
+{
+    struct uci_context *ctx = uci_alloc_context();
+    if (!ctx) {
+        ERROR("%s, uci alloc context failed!\n", __FUNCTION__);
+        return RETURN_ERR;
+    }
+
+    char lookup_str[MAX_UCI_BUF_LEN] = "wireless";
+    struct uci_ptr ptr;
```
tomereli, line 179: I don't think there is a need to use the `struct` keyword since this is CPP (nitpick)

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+        //get from wireless.radio?.param the param wanted
+        int status = snprintf_s(path, MAX_UCI_BUF_LEN, "wireless.%s.%s", deviceName, param);
+        if (status <= 0) {
+            ERROR("%s failed snprintf. device=%s param=%s\n", __func__, deviceName, param);
+            return RETURN_ERR;
+        }
+
+        status = cfg_uci_get(path, value, MAX_UCI_BUF_LEN);
+        if (status == RETURN_OK) {
+            // DEBUG("%s index=%d %s=%s\n", __func__, index, param, value);
```
tomereli: Removed dead code

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -175,33 +175,30 @@ int cfg_get_device_info(BPL_DEVICE_INFO *device_info) { return 0; }
...
-    if (retVal) {
-        return retVal;
-    }
+    char ifname[BPL_IFNAME_LEN] = {0};
+    utils::copy_string(ifname, iface, BPL_IFNAME_LEN);
 
     // The UCI "disabled" setting is optional, defaults to false if not present
     bool disabled = false;
-    cfg_uci_get_wireless_bool(TYPE_RADIO, index, "disabled", &disabled);
+    cfg_uci_get_wireless_bool(TYPE_RADIO, ifname, "disabled", &disabled);
```
tomereli: Not OK - this can fail, I know this wasn't tested before but it should definitely be tested now since `ifname` is not guaranteed to be in `/etc/config/wireless`.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -175,33 +175,30 @@ int cfg_get_device_info(BPL_DEVICE_INFO *device_info) { return 0; }
...
     // can't get the channel from UCI just move on.
     wlan_params->channel = 0;
-    if (!cfg_get_channel(index, &wlan_params->channel)) {
-        MAPF_INFO("UCI: radio" << index << ": channel is not configured assuming auto\n");
+    if (!cfg_get_channel(ifname, &wlan_params->channel)) {
+        MAPF_INFO("UCI: interface " << ifname << ": channel is not configured assuming auto\n");
     }
 
-    return retVal;
+    return RETURN_OK;
```
tomereli, line 198: We should not always return success here - depending on the actual operation success/failure.

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
```diff
@@ -82,14 +82,17 @@ int cfg_get_prplmesh_param_int(const std::string &param, int *buf)
...
         return RETURN_ERR;
     }
 
     char channel_num[MAX_UCI_BUF_LEN] = {0};
-    int status = cfg_uci_get_wireless(TYPE_RADIO, index, "channel", channel_num);
+    char ifname[BPL_IFNAME_LEN]       = {0};
+
+    utils::copy_string(ifname, interface_name.c_str(), BPL_IFNAME_LEN);
+    int status = cfg_uci_get_wireless_from_ifname(TYPE_RADIO, ifname, "channel", channel_num);
     if (status == RETURN_ERR)
```
tomereli: This is silly - just change it to;
```
if (cfg_uci_get_wireless_from_ifname(TYPE_RADIO, ifname, "channel", channel_num) != RETURN_OK) {
    return RETURN_ERR;
}
```

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
```diff
@@ -103,7 +106,7 @@ int cfg_get_channel(int index, int *channel)
     return RETURN_OK;
 }
 
-int cfg_get_wep_key(int index, int keyIndex, char *key)
+int cfg_get_wep_key(const std::string &interface_name, int keyIndex, char *key)
```
tomereli, line 109: I think this whole function can and should be removed. WEP is obsolete and we don't support it.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -66,12 +66,13 @@ int cfg_uci_get(char *path, char *value, size_t length)
...
 
-int cfg_uci_get_wireless_int(enum paramType type, int index, const char param[], int *value)
+int cfg_uci_get_wireless_int(enum paramType type, char *interface_name, const char param[],
+                             int *value)
 {
     int status;
     char val[MAX_UCI_BUF_LEN] = "";
 
-    status = cfg_uci_get_wireless(type, index, param, val);
+    status = cfg_uci_get_wireless_from_ifname(type, interface_name, param, val);
```
tomereli, line 78: Similar here

tomereli (changes requested):

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
arnout: I think it's needed because the type is declared within an `extern "C"`.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+            is_section_found = true;
+            break;
+        }
+
+        if (is_section_found) {
+            break;
+        }
+    }
+
+    if (type == TYPE_RADIO) {
```
arnout, line 236: I really don't like the pattern
```
if (arg == 1) {
   ...
   return x;
}

// arg == 2
...
return y;
```

I.e., rely on the return for something that is a choice. The pattern is excellent for handling errors, but not for handling two alternative code paths IMO.

In this case, I much prefer either:
```
if (arg == 1) {
   ... foo = ...
} else { // arg == 2
   ... foo = ...
}
return foo;
```

or even

```
switch (arg) {
case 1:
   ...
   return x;
case 2:
   ...
   return y;
default:
   ERROR
   return z;
}
```

(nitpick)

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+    struct uci_ptr ptr;
+    if ((uci_lookup_ptr(ctx, &ptr, lookup_str, true) != UCI_OK)) {
+        ERROR("%s, uci lookup failed!\n", __FUNCTION__);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
+
+    if (!ptr.p) {
+        ERROR("%s, returned pointer is null\n", __FUNCTION__);
+        return RETURN_ERR;
```
arnout, line 189: Missing uci_free_context

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+    if (type == TYPE_RADIO) {
+        //get radio name from .device param value
+        char deviceName[MAX_UCI_BUF_LEN] = {0};
+        uci_foreach_element(&s->options, n)
+        {
+            struct uci_option *o = uci_to_option(n);
+
+            if (strncmp(n->name, "device", MAX_UCI_BUF_LEN) == 0 && o->type == UCI_TYPE_STRING) {
+                strncpy_s(deviceName, MAX_UCI_BUF_LEN, o->v.string, MAX_UCI_BUF_LEN - 1);
```
arnout: You could have constructed path here directly, but OK.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+        if (deviceName[0] == '\0') {
+            // radio not found
+            ERROR("%s deviceName is mandatory param and not configured in %s\n", __func__,
+                  lookup_str);
+            return RETURN_ERR;
+        }
+
+        char path[MAX_UCI_BUF_LEN] = "";
+        //get from wireless.radio?.param the param wanted
+        int status = snprintf_s(path, MAX_UCI_BUF_LEN, "wireless.%s.%s", deviceName, param);
```
arnout: Is there any particular reason to use snprintf_s rather than std::string?

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+        if (is_section_found) {
+            break;
+        }
+    }
+
+    if (type == TYPE_RADIO) {
+        //get radio name from .device param value
+        char deviceName[MAX_UCI_BUF_LEN] = {0};
+        uci_foreach_element(&s->options, n)
```
arnout, line 241: Is this still valid if is_section_found is false?

framework/platform/bpl/include/bpl/bpl_cfg.h
adam1985d: the reason to remove was that we are not using the code.
the reason why we were looking at this at all was that due to the difference in the wireless configuration between UGW, GLInet, RAX40, etc.

we could have changed those wifi settings to optional - but that would just hide the fact we aren't really reading it.
we could have invested in fixing the wireless configuration in GLInet and Turris Omnia - but the decision was not to invest in this at this point for several reasons:
1. the data read is not really used by the prplmesh
2. there is a task to remove the prplmesh access to wireless configuration and instead use BWL or some other method to retrieve the information.

regarding the commit message, I don't think that the platform wireless configuration difference is really beneficial information here since if we wanted to address it we would have fixed it.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+        ERROR("%s, interface(%s) found but param(%s) isn't configured\n", __FUNCTION__,
+              interface_name, param);
+        uci_free_context(ctx);
+        return RETURN_ERR;
+    }
+
+    //interface not found in wireless
+    ERROR("%s, interface(%s) not found", __FUNCTION__, interface_name);
+    uci_free_context(ctx);
+    return RETURN_ERR;
```
arnout: So I think this bit should move up before the type == RADIO condition.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +162,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
...
+
+    //* type == TYPE_VAP *
+    // read the value from the selected section
+    if (is_section_found) {
+        // get param in the selected section
+        uci_foreach_element(&s->options, n)
+        {
+            struct uci_option *o = uci_to_option(n);
+            // if param is found in options
+            if (strncmp(n->name, param, MAX_UCI_BUF_LEN) == 0 && o->type == UCI_TYPE_STRING) {
```
arnout, line 276: Not specific for this commit, because it's done like this all over this file, but:

Why don't we use the uci helper functions? Like uci_parse_section, or for this specific one, uci_lookup_option?

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -175,33 +175,30 @@ int cfg_get_device_info(BPL_DEVICE_INFO *device_info) { return 0; }
...
+                 << intptr_t(iface) << " wlan_params = " << intptr_t(wlan_params));
         return RETURN_ERR;
     }
 
-    retVal = cfg_get_index_from_interface(iface, &index);
-    if (retVal) {
-        return retVal;
-    }
+    char ifname[BPL_IFNAME_LEN] = {0};
+    utils::copy_string(ifname, iface, BPL_IFNAME_LEN);
```
arnout: Why is this copied instead of just using `iface.c_str()`?

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
arnout: ... but it was already like that, so it should be fixed in a separate commit.

And as I explained, I think it actually is OK.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
arnout: Well... If it fails, it can be because the section is missing (a real error), or because the option is not set in UCI (default false, as explained in the comment above). So this *is* actually OK IMO.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
```diff
@@ -91,6 +91,9 @@ int cfg_uci_get_radio_param(int index, const char param[], char *value, size_t b
 int cfg_uci_get_wireless_radio_idx(const char *interfaceName, int *radio_index);
 int cfg_uci_get_radio_param_ulong(int index, const char param[], unsigned long *value);
 
+int cfg_uci_get_wireless_from_ifname(enum paramType type, char *interface_name, const char param[],
```
arnout: interface_name and param should be const.

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
```diff
@@ -82,14 +82,17 @@ int cfg_get_prplmesh_param_int(const std::string &param, int *buf)
...
     if (!channel) {
         return RETURN_ERR;
     }
 
     char channel_num[MAX_UCI_BUF_LEN] = {0};
-    int status = cfg_uci_get_wireless(TYPE_RADIO, index, "channel", channel_num);
+    char ifname[BPL_IFNAME_LEN]       = {0};
+
+    utils::copy_string(ifname, interface_name.c_str(), BPL_IFNAME_LEN);
+    int status = cfg_uci_get_wireless_from_ifname(TYPE_RADIO, ifname, "channel", channel_num);
```
arnout: Again, why copy ifname?

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
arnout: But lets not make this PR block on that...

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -66,12 +66,13 @@ int cfg_uci_get(char *path, char *value, size_t length)
     return RETURN_OK;
 }
 
-int cfg_uci_get_wireless_int(enum paramType type, int index, const char param[], int *value)
+int cfg_uci_get_wireless_int(enum paramType type, char *interface_name, const char param[],
```
arnout: interface_name should be const

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -440,12 +441,12 @@ int cfg_uci_get_wireless_idx(char *interfaceName, int *rpc_index)
     return RETURN_ERR;
 }
 
-int cfg_uci_get_wireless_bool(enum paramType type, int index, const char param[], bool *value)
+int cfg_uci_get_wireless_bool(enum paramType type, char *interface_name, const char param[],
```
arnout: interface_name should be const

arnout (approved):
Lots of changes requested, but can be merged after fixing those.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
tomereli: Right, I get it.

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
tomereli: Right, forgot to mark this as (nitpick)

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
tomereli: Kudos on giving such an elaborate comment even though you mark it as (nitpick)

tomereli (approved):

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
adam1985d: I agree - already fixed it locally - will push the commit after fixing all the comments

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
adam1985d: ok - we can remove it as part of #805

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
```diff
@@ -162,6 +163,132 @@ int cfg_uci_get_wireless(enum paramType type, int index, const char param[], cha
     return status;
 }
 
+int cfg_uci_get_wireless_from_ifname(enum paramType type, char *interface_name, const char param[],
+                                     char *value)
+{
+    struct uci_context *ctx = uci_alloc_context();
+    if (!ctx) {
+        ERROR("%s, uci alloc context failed!\n", __FUNCTION__);
+        return RETURN_ERR;
```
rmelotte, line 175: ~~Shouldn't there also be a call to uci_free_context here?~~
Edit: nevermind

rmelotte (approved):

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: the battle between the forces of indentation and forces of if-else continues :-)
I will change it to if-else

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: actually no - a bad example of copy-paste.
changed to use std::string

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: changed to do exactly that. thanks

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: you are correct - added a check after the loop. if not found - return false

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: done as a response to the previous comment. thanks

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: I was using the existing implementation as a reference and wasn't aware of the `uci_lookup_option`
since the access to wireless config is going to be removed in #805 I don't want to invest in refactoring it, but took notice for any later change in this area and in general. thanks

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
adam1985d: The BPL UCI APIs provide a C interface.
although currently, we use it by CPP files only, @vitalybu suggested leaving the interface in C style for that purpose.
The BPL UCI APIs implementation requires non-const char arrays as the UCI lib indicates the provided strings might be changed inside and thus must not be const.

framework/platform/bpl/uci/cfg/bpl_cfg_uci.h
adam1985d: changed to const

framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
adam1985d: removed once const is used for interface name

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

framework/platform/bpl/uci/cfg/bpl_cfg_uci.cpp
adam1985d: fixed

tomereli:
Added don't merge tag cause Adam needs to push some more code review fixes. @adam1985d please merge once you are done (and remove the tag).

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
arnout: Indeed, uci_lookup_ptr modifies its path argument. But it would be more appropriate to do the copy in uci_cfg_get() then. Anyway, something for a separate PR.

BTW, I believe cfg_uci_get_wireless() is no longer used and can be removed.

PR #806, opened 2020-02-11
Make sure topology response contains up-to-date VAP list

### LINKED ISSUE #791: [TASK] Update VAP list after setting new VAPs
After setting new VAPs from the M2, i.e. when the AP manager gets an ACTION_APMANAGER_WIFI_CREDENTIALS_UPDATE_REQUEST, the AP manger should send back the updated VAP list.

This is required to handle the teardown test (4.4.3) correctly.

arnout:
Fixes #791.

In certification tests, it turns out that the teardown test 4.4.3 fails because the AP Operational BSS TLV is not empty.

First I added a test that the VAP list is updated to test_flows.sh
This uncovered a number of issues.

The first problem is that the VAP list doesn't get updated in dummy when a BSS is torn down.
However, this doesn't affect certification.
It does affect test_flows, so it needed to be fixed.

The second problem is that the VAP list doesn't get updated if there is no UCC listener.
Again, this doesn't affect certification, because in certification mode there *is* a UCC listener.

Finally, the AP Operational BSS TLV contains BSSes even for the VAPs that are torn down.
That is simply fixed by excluding the VAPs with empty SSID.
Unfortunately, I didn't find a way to check this in test_flows.
I use bml_conn_map in test_flows, but that uses a vendor specific message rather than Topology Response.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1707,19 +1707,17 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
     case beerocks_message::ACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION: {
         LOG(DEBUG) << "ACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION received from iface "
                    << soc->hostap_iface;
-        if (m_agent_ucc_listener) {
-            auto msg = beerocks_header->addClass<
-                beerocks_message::cACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION>();
-            if (!msg) {
-                LOG(ERROR)
-                    << "Failed building ACTION_BACKHAUL_DL_RSSI_REPORT_NOTIFICATION message!";
-                return false;
-            }
+        auto msg = beerocks_header->addClass<
+            beerocks_message::cACTION_BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION>();
+        if (!msg) {
+            LOG(ERROR) << "Failed parsing BACKHAUL_HOSTAP_VAPS_LIST_UPDATE_NOTIFICATION message!";
+            return false;
+        }
 
+        m_radio_info_map[msg->ruid()].vaps_list = msg->params();
+        if (m_agent_ucc_listener) {
             m_agent_ucc_listener->update_vaps_list(network_utils::mac_to_string(msg->ruid()),
                                                    msg->params());
-
-            m_radio_info_map[msg->ruid()].vaps_list = msg->params();
         }
         break;
     }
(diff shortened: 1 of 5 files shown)


arnout:
Certification test 4.4.3 on RAX40: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/433089358

common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
```diff
@@ -75,6 +75,8 @@ class base_wlan_hal_dummy : public virtual base_wlan_hal,
     bool dummy_send_cmd(const std::string &cmd, char **reply); // for external process
     bool dummy_send_cmd(const std::string &cmd);
 
+    static const int predefined_vaps_num = 4;
```
tomereli, line 78: Just so I don't forget - we should make sure we are using this value in the M1, and not hard coded in the agent.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1836,6 +1836,8 @@ bool backhaul_manager::handle_1905_discovery_query(ieee1905_1::CmduMessageRx &cm
         for (const auto &vap : vaps_list.second.vaps) {
             if (vap.mac == network_utils::ZERO_MAC)
                 continue;
+            if (vap.ssid[0] == '\0')
```
tomereli, line 1959: Is it guaranteed that empty ssid means that there is no beacon?
Why not instead add another member to the VAP structure which states the VAP state?
Yes, this would mean we have to implement this in all BWL backends, but isn't it cleaner?

tomereli (approved):
One important comment regarding how to identify that a VAP is tore down - otherwise LGTM. Approving since I guess you thought of the option I suggested, but I'd still like to hear the reason for choosing empty SSID and not just adding a new field to the VAP structure in BWL.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
arnout: "We should" as in "we should add a bwl function to report the max number of VAPs", right? Because currently this is indeed hard coded in the agent. -> #810

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: Yeah, I didn't want to change too much...

The situation is a bit schizophrenic. The bwl interface uses a map vap_index->vap_info, so it kind of assumes that it will only contain "valid" vaps. But then the messages use sVapsList which has a fixed size IFACE_TOTAL_VAPS, which kind of assumes a pre-allocated number of VAPs. And the bwl implementations also all have a fixed number of VAPs they support, bounded by IFACE_VAP_ID_MAX (which is based on IFACE_TOTAL_VAPS).

In my book, it's a mess :-) So I didn't want to touch it too much.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
tomereli: Added to the sprint backlog and to the first milestone.

dob71: Maybe max VAPs supported by the radio could be useful as a parameter coming from the folders emulating the radios rather than some fixed value. I don't know if there is any value in this to test different limits, just a thought.

dob71 (approved):

common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
arnout: Good one! Updated #810 with that idea.

PR #807, opened 2020-02-11
Hotfix/fix dcs flow issues

itayx:
all DCS code has been ported #562 

DCS Validation has been started on both RDKB/UGW platforms.

This PR is used to fix small issues throughout the codebase in the DCS feature's flow


tomereli:
> **DO NOT REVIEW**
Next time, you can open a draft PR if you don't want people to review it just yet:)

kantera800 (commented):
better 1st commit  headline would be:
son_slave: fix cmdu typo in monitor message

better 2st commit  headline would be:
get mac from hostap_mac in control message
also i would expect to know why reading radio_mac from notification is not correct.

tomereli (approved):

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3359,7 +3359,8 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
             return false;
         }
 
-        auto radio_mac = notification->radio_mac();
+        //get the mac from hostap_mac
+        auto radio_mac = network_utils::mac_from_string(hostap_mac);
```
arnout, line 3368: This sounds wrong... "hostap" is normally used to indicate the VAP, not the radio.

But looking at the definition of hostap_mac, it is in fact wrongly named, because it takes the radio_mac from the beerocks header. So it's OK.

That said, a much better change would be to (in one commit) change the beginning of the function to

```
    auto radio_mac = beerocks_header->actionhdr()->radio_mac();
    auto hostap_mac = network_utils::mac_to_string(radio_mac);
```

and (optionally) in a second commit remove most instances of hostap_mac (in the logging, you can just use an sMacAddr directly now).

Also, the commit message should explain why you do this. The commit message just says "because we can". I thought it would be because you're going to remove the radio_mac from cACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION, but you don't do that.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3359,7 +3359,8 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
             return false;
         }
 
-        auto radio_mac = notification->radio_mac();
+        //get the mac from hostap_mac
```
arnout, line 3367: This comment is useless. It basically says "get the mac from the mac".

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3382,7 +3383,8 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        auto radio_mac = network_utils::mac_from_string(hostap_mac);
```
arnout, line 3392: Same comment.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3422,10 +3424,13 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
             return false;
         }
 
+        //get the mac from hostap_mac
+        auto radio_mac = network_utils::mac_from_string(hostap_mac);
```
arnout, line 3433: Same comment.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3440,10 +3445,13 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
```
arnout, line 3454: Same comment.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1510,9 +1510,9 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
 
         // If event == Channel_Scan_New_Results_Ready do nothing since is_dump's default is 0
         if (event == Event::Channel_Scan_Dump_Result) {
-            auto &msg = *(static_cast<bwl::sCHANNEL_SCAN_RESULTS_NOTIFICATION *>(data));
+            auto msg = static_cast<bwl::sCHANNEL_SCAN_RESULTS_NOTIFICATION *>(data);
 
-            auto &in_result  = msg.channel_scan_results;
+            auto &in_result  = msg->channel_scan_results;
```
arnout, line 1519: Although the code indeed looks better like this, it does nothing to fix the issue mentioned in the commit message. in_result is still the same reference as it was before, so it will become invalid as soon as the lifetime of data ends.

I also don't see any problem with the lifetime of in_result. It is only used within this event handler, and no references are taken anywhere.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1216,7 +1216,8 @@ int cli_bml::set_dcs_continuous_scan_params_caller(int numOfArgs)
 
     std::string::size_type pos;
     //[radio_mac=<radio_mac> dwell_time=<dwell_time> interval_time='<interval_time>']"
-    for (int i = 1; i < numOfArgs; i++) { //first optional arg
+    for (int i = 0; i < numOfArgs; i++) { //first optional arg
+        std::cout << "args[" << i << "]: " << args.stringArgs[i] << std::endl;;
```
arnout: Do we really want to leave this debug thing in here?

arnout (approved):
Please fix up the minor comments I made and merge. Take good care with the "fix bug in dereferencing scan results event" because I don't think it fixes anything, so either there was never any bug, or the bug is still there.

arnout:
Another little nitpick: @itayx you often use "crush" instead of "crash". Also some commit messages have wrong capitalisation or weird wrapping.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: No.
It was an oversight that it stayed, removed due to it being unneeded

controller/src/beerocks/master/son_master_thread.cpp
itayx: Since this change would effect more then just the DCS feature,
I'm going to create a new issue for it (assigned to me), and I'll add a Hotfix to solve this once I'm finished with my current issues

PR #808, opened 2020-02-11
Hotfix/fix dcs get results

itayx:
**To be merged after #807**

all DCS code has been ported #562 
DCS Validation has been started on both RDKB/UGW platforms.

This PR is used to fix an issue with the DCS feature's "get results" flow
This PR contains:
- a fix to the bml message results_size variable
- changes to the get results flow in the son_management
- extended logs in bml_internal
- fix casting error in bml_internal
- fix validation error in bml_internal

### DIFF
--- a/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
@@ -384,11 +384,15 @@ static bool translate_nl_data_to_bwl_results(sChannelScanResults &results,
         results.beacon_period_ms = (unsigned int)nla_get_u16(bss[NL80211_BSS_BEACON_INTERVAL]);
     }
 
-    // get signal strength
+    // get signal strength, signal strength units not specified, scaled to 0-100
     if (bss[NL80211_BSS_SIGNAL_UNSPEC]) {
-        results.signal_strength_dBm = (nla_get_u8(bss[NL80211_BSS_SIGNAL_UNSPEC])) / 100;
+        //signal strength of the probe response/beacon in unspecified units, scaled to 0..100 <u8>
+        results.signal_strength_dBm =
+            int32_t(nla_get_u8(bss[NL80211_BSS_SIGNAL_UNSPEC]));
     } else if (bss[NL80211_BSS_SIGNAL_MBM]) {
-        results.signal_strength_dBm = (nla_get_u32(bss[NL80211_BSS_SIGNAL_MBM])) / 100;
+        //signal strength of probe response/beacon in mBm (100 * dBm) <s32>
+        results.signal_strength_dBm =
+            int32_t(nla_get_u32(bss[NL80211_BSS_SIGNAL_MBM])) / 100;
     }
 
     //get information elements from information-elements-buffer or from beacon
--- a/controller/src/beerocks/bml/bml_defs.h
+++ b/controller/src/beerocks/bml/bml_defs.h
@@ -555,40 +555,40 @@ struct BML_STATS_ITER {
 };
 
 struct BML_NEIGHBOR_AP {
-    char ap_SSID[BML_SSID_MAX_LENGTH];
     //The current service set identifier in use by the neighboring WiFi SSID. The value MAY be empty for hidden SSIDs.
-    unsigned char ap_BSSID[BML_MAC_ADDR_LEN];
+    char ap_SSID[BML_SSID_MAX_LENGTH];
     //The BSSID used for the neighboring WiFi SSID.
-    uint8_t ap_Mode;
+    uint8_t ap_BSSID[BML_MAC_ADDR_LEN];
     //The mode the neighboring WiFi radio is operating in. Enumerate
-    uint32_t ap_Channel;
+    uint8_t ap_Mode;
     //The current radio channel used by the neighboring WiFi radio.
-    int32_t ap_SignalStrength;
+    uint32_t ap_Channel;
     //An indicator of radio signal strength (RSSI) of the neighboring WiFi radio measured in dBm, as an average of the last 100 packets received.
+    int32_t ap_SignalStrength;
+    //The type of encryption the neighboring WiFi SSID advertises. Enumerate List.
     uint8_t ap_SecurityModeEnabled[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
     //The type of encryption the neighboring WiFi SSID advertises. Enumerate List.
     uint8_t ap_EncryptionMode[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
-    //The type of encryption the neighboring WiFi SSID advertises. Enumerate List.
-    uint8_t ap_OperatingFrequencyBand;
     //Indicates the frequency band at which the radio this SSID instance is operating. Enumerate
-    uint8_t ap_SupportedStandards[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
+    uint8_t ap_OperatingFrequencyBand;
     //List items indicate which IEEE 802.11 standards thisResultinstance can support simultaneously, in the frequency band specified byOperatingFrequencyBand. Enumerate List
-    uint8_t ap_OperatingStandards;
+    uint8_t ap_SupportedStandards[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
     //Indicates which IEEE 802.11 standard that is detected for this Result. Enumerate
-    uint8_t ap_OperatingChannelBandwidth;
+    uint8_t ap_OperatingStandards;
     //Indicates the bandwidth at which the channel is operating. Enumerate
-    uint32_t ap_BeaconPeriod;
+    uint8_t ap_OperatingChannelBandwidth;
     //Time interval (inms) between transmitting beacons.
-    int32_t ap_Noise;
+    uint32_t ap_BeaconPeriod;
     //Indicator of average noise strength (indBm) received from the neighboring WiFi radio.
-    uint32_t ap_BasicDataTransferRates[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
+    int32_t ap_Noise;
     //Basic data transmit rates (in Kbps) for the SSID.
-    uint32_t ap_SupportedDataTransferRates[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
+    uint32_t ap_BasicDataTransferRates[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
     //Data transmit rates (in Kbps) for unicast frames at which the SSID will permit a station to connect.
-    uint32_t ap_DTIMPeriod;
+    uint32_t ap_SupportedDataTransferRates[BML_CHANNEL_SCAN_ENUM_LIST_SIZE];
     //The number of beacon intervals that elapse between transmission of Beacon frames containing a TIM element whose DTIM count field is 0. This value is transmitted in the DTIM Period field of beacon frames. [802.11-2012]
-    uint32_t ap_ChannelUtilization;
+    uint32_t ap_DTIMPeriod;
     //Indicates the fraction of the time AP senses that the channel is in use by the neighboring AP for transmissions.
+    uint32_t ap_ChannelUtilization;
 };
 
 /****************************************************************************/
--- a/controller/src/beerocks/bml/internal/bml_internal.cpp
+++ b/controller/src/beerocks/bml/internal/bml_internal.cpp
@@ -997,20 +997,20 @@ int bml_internal::process_cmdu_header(std::shared_ptr<beerocks_header> beerocks_
             if (m_prmChannelScanResultsGet) {
 
                 if (m_scan_results && m_scan_results_maxsize && m_scan_results_status) {
-
-                    LOG(DEBUG) << "Receiving Response";
-
                     uint8_t op_error_code  = response->op_error_code();
                     *m_scan_results_status = response->result_status();
                     auto scan_results_size = response->results_size();
                     uint8_t last           = response->last();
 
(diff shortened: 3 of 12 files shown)


tomereli (approved):
Most comments are easy to fix, so approving. Great split to commits and good commit messages!

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -2020,12 +2020,12 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_res
...
     }
 
-    BML_NEIGHBOR_AP results[max_results_size] = {0};
+    BML_NEIGHBOR_AP results[max_results_size];
 
     uint8_t status             = 0;
     unsigned int results_count = max_results_size;
     int ret                    = bml_get_dcs_scan_results(ctx, radio_mac.c_str(),
-                                       reinterpret_cast<BML_NEIGHBOR_AP **>(&results),
+                                       reinterpret_cast<BML_NEIGHBOR_AP *>(results),
```
arnout: The reinterpret_cast can be removed now I think. It's due to this cast that we didn't notice the issue before.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -70,7 +70,8 @@ static void config_logger(const std::string log_file = std::string())
 static void translate_channel_scan_results(const beerocks_message::sChannelScanResults &res_in,
                                            BML_NEIGHBOR_AP &res_out)
 {
-    string_utils::copy_string(res_out.ap_SSID, res_in.ssid,
+    string_utils::copy_string(reinterpret_cast<char *>(res_out.ap_SSID),
+                              reinterpret_cast<const char *>(res_in.ssid),
```
arnout: This is (again) a bogus change that does not solve the issue that is claimed in the commit message.

res_out.ap_SSID is already of type char*, so adding a cast does nothing.

res_in.ssid is also a char*, so it's essentially just a const_cast, and unnecessary because the const is added.

Also, if a cast would be required, it should be a static_cast, not a reinterpret_cast, because the latter is simply a C-style cast in another jacket.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1574,7 +1574,7 @@ int bml_internal::get_dcs_scan_results(const sMacAddr &mac, BML_NEIGHBOR_AP *res
                                        const unsigned int max_results_size, uint8_t &result_status,
                                        bool is_single_scan)
 {
-    if (!results || !results_size || !result_status) {
+    if (!results || max_results_size == 0) {
```
arnout, line 1574: Actually, the `!results` bit is also not entirely correct: that is a check of the input parameter validity (similar to the `!ctx` check in bml_get_dcs_scan_results).

So better would be to have a separate check for results and for max_results_size.

(nitpick)

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -382,9 +382,13 @@ static bool translate_nl_data_to_bwl_results(sChannelScanResults &results,
 
     // get signal strength
     if (bss[NL80211_BSS_SIGNAL_UNSPEC]) {
-        results.signal_strength_dBm = (nla_get_u8(bss[NL80211_BSS_SIGNAL_UNSPEC])) / 100;
+        //signal strength of the probe response/beacon in unspecified units, scaled to 0..100 <u8>
+        results.signal_strength_dBm =
+            static_cast<int32_t>(nla_get_u8(bss[NL80211_BSS_SIGNAL_UNSPEC]));
```
arnout: The static cast does nothing here, it is implied by the type change from u8 to int32.

Also, if the value is in the range 0..100, it means we can get something up to 100dBm... Is that a sane range? It's unspecified anyway so it probably doesn't matter, but still...

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -382,9 +382,13 @@ static bool translate_nl_data_to_bwl_results(sChannelScanResults &results,
...
     } else if (bss[NL80211_BSS_SIGNAL_MBM]) {
-        results.signal_strength_dBm = (nla_get_u32(bss[NL80211_BSS_SIGNAL_MBM])) / 100;
+        //signal strength of probe response/beacon in mBm (100 * dBm) <s32>
+        results.signal_strength_dBm =
+            static_cast<int32_t>(nla_get_u32(bss[NL80211_BSS_SIGNAL_MBM])) / 100;
```
arnout: I didn't know static cast could do that! Excellent.

controller/src/beerocks/bml/bml_defs.h
```diff
@@ -555,40 +555,40 @@ struct BML_STATS_ITER {
 };
 
 struct BML_NEIGHBOR_AP {
-    char ap_SSID[BML_SSID_MAX_LENGTH];
     //The current service set identifier in use by the neighboring WiFi SSID. The value MAY be empty for hidden SSIDs.
-    unsigned char ap_BSSID[BML_MAC_ADDR_LEN];
+    char ap_SSID[BML_SSID_MAX_LENGTH];
     //The BSSID used for the neighboring WiFi SSID.
-    uint8_t ap_Mode;
+    uint8_t ap_BSSID[BML_MAC_ADDR_LEN];
```
arnout, line 561: It would have been nicer if the two changes here were in separate commits, because now the unsigned char -> uint8_t change is a bit lost between the rest.

Also, while changing the comments, you could also have converted them to doxygen (just add an additional `/`, so `///`)

Both nitpicks.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -2078,3 +2067,25 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_res
...
+const std::string cli_bml::string_from_int_array(uint8_t *arr, size_t arr_max_size)
+{
+    std::stringstream ss;
+    if (arr) {
+        for (size_t i = 0; i < arr_max_size; i++) {
+            ss << ((i != 0) ? ", " : "") << static_cast<int>(arr[i]);
+        }
+    }
+    return ss.str();
+}
```
arnout: Please implement with a template instead of copying the same code 3 times.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -2078,3 +2067,25 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_res
 
     return 0;
 }
+
+const std::string cli_bml::string_from_int_array(uint8_t *arr, size_t arr_max_size)
+{
+    std::stringstream ss;
+    if (arr) {
+        for (size_t i = 0; i < arr_max_size; i++) {
+            ss << ((i != 0) ? ", " : "") << static_cast<int>(arr[i]);
```
arnout: In this case, we want `int(arr[i])` instead of static_cast, I think. Or better yet, `unsigned int(arr[i])`.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -2038,22 +2038,22 @@ int cli_bml::get_dcs_scan_results(const std::string &radio_mac, uint32_t max_res
...
-                          << "  operating_standards=" << res.ap_OperatingStandards << std::endl
-                          << "  operating_channel_bandwidth=" << res.ap_OperatingChannelBandwidth
-                          << std::endl
-                          << "  beacon_period=" << (int)res.ap_BeaconPeriod << std::endl
-                          << "  noise=" << (int)res.ap_Noise << std::endl
-                          << "  basic_data_transfer_rates="
+                          << "  operating_standards=" << int(res.ap_OperatingStandards) << std::endl
+                          << "  operating_channel_bandwidth="
+                          << int(res.ap_OperatingChannelBandwidth) << std::endl
+                          << "  beacon_period_ms=" << (int)res.ap_BeaconPeriod << std::endl
```
arnout: You could have converted the casts into initializers as well (nitpick).

arnout (commented):
I have a few important comments that warrant a re-review. However, I might not do the re-review, so just commenting rather than changes requested.

arnout:
> Please open an issue for the TLVF bug, and update the commit message to reflect that this change is a workaround but is harmless since uint8 serves the needs just fine.

tlvVarList.yaml actually has this:
```
```

But apparently there's no issue for it yet.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: Changed to `static_cast`, since the type of `results` is `BML_NEIGHBOR_AP[max_results_size]` while the requested type is `BML_NEIGHBOR_AP*`. And while the two are identical, I believe it is better to cast the type ourselves

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: You're right, the segmentation error solved in an earlier commit, it's an oversight that this commit was included so I'm removing the entire commit

controller/src/beerocks/bml/internal/bml_internal.cpp
itayx: I understand and agree with your point that `!results` is an  input parameter validity check
But I think that `max_results_size == 0` could also be classified as a parameter validity check since it check the validity of the parameter `max_results_size`

Since the parameter `max_results_size` is defined as a `const unsigned int` is cannot have a signed value, but can still be set as 0, which is invalid for the "get results" flow since it would mean that the client has requested 0 results
therefore the error `"Function is called, but no data is being requested!"`

controller/src/beerocks/cli/beerocks_cli_bml.cpp
arnout: No, the two really are identical. Casting would give the wrong impression.

arnout: Let me qualify that... By adding an explicit static cast, you hide a possible problem where the type of either `bml_get_dcs_scan_results` or `results` changes. True, the number of casts that are allowed with static_cast is *very* limited, but there are still a few possible, and it is very unlikely that they are what we want there. So removing the static cast may avoid bugs in the future (or rather, make sure they are captured by the compiler).

controller/src/beerocks/bml/internal/bml_internal.cpp
arnout: OK fair enough.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: Changed to use template

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: Changed

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
kantera800: I wasn't sure about it either but this is what the nl80211 api provides.
https://lxr.openwrt.org/source/iwinfo/api/nl80211.h#L3407

never saw this type in the logs but it doesn't mean it anything.
I suggest to leave it and maybe add a warning message .
some thing like 
"signal strength units not specified, scaled to 0-100"

kantera800:
> > Please open an issue for the TLVF bug, and update the commit message to reflect that this change is a workaround but is harmless since uint8 serves the needs just fine.
> 
> tlvVarList.yaml actually has this:
> 
> ```
> # TODO tlvf currently doesn't support list length field of more than 1 byte
> ```
> 
> But apparently there's no issue for it yet.

https://github.com/prplfoundation/prplMesh/issues/836

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
arnout: Good idea.

itayx: Added the warning as a comment
and removed the unneeded static casts, but retained the explicit cast to `int32_t` using a functional cast

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: changed to functional cast instead

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: Removed the cast all together
It builds with no issues, before when we used `BML_NEIGHBOR_AP **` without a cast the build failed, but when we changed the API call to `BML_NEIGHBOR_AP *` we don't need an explicit cast

controller/src/beerocks/cli/beerocks_cli_bml.h
```diff
@@ -205,6 +205,8 @@ class cli_bml : public cli {
                               const std::string &channel_pool);
     int get_dcs_scan_results(const std::string &radio_mac, uint32_t max_results_size,
                              bool is_single_scan = false);
+    template <typename T>
+    const std::string string_from_int_array(T *arr, size_t arr_max_size);
```
arnout, line 209: Why does it need to be exported now? in other words, why is it extern instead of static and why is it declared in the header file? It is not explained in the commit message.

arnout (approved):
Looks good, but needs a manual rebase on master.

PR #817, opened 2020-02-12
Feature/add dcs scan dump

itayx:
Add "Dwell time = 0" feature to allow for the DCS task to request a scan dump.
A scan dump consists of the BWL "dumping" all available cached results
#818


common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -157,6 +157,12 @@ bool mon_wlan_hal_dummy::channel_scan_trigger(int dwell_time_msec,
     return false;
 }
 
+bool mon_wlan_hal_dummy::channel_scan_dump_results()
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED";
+    return false;
```
vitalybu, line 163: Shouldn't this method return `true` for the dummy bwl?
Otherwise the DCS logic will always fail on Linux.
Maybe it's better to return true here and provide an empty result-set.

common/beerocks/bwl/nl80211/mon_wlan_hal_nl80211.cpp
```diff
@@ -392,6 +392,12 @@ bool mon_wlan_hal_nl80211::channel_scan_trigger(int dwell_time_msec,
     return false;
 }
 
+bool mon_wlan_hal_nl80211::channel_scan_dump_results()
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED";
+    return false;
```
vitalybu, line 398: Same comment as for the dummy bwl

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1226,6 +1226,42 @@ bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
...
+        message_com::send_cmdu(slave_socket, cmdu_tx);
+
+        auto notification = message_com::create_vs_message<
+           beerocks_message::cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION>(cmdu_tx);
+        if (!notification) {
+            LOG(ERROR) << "Failed building cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION msg";
+            return false;
+        }
+
+        //Sending the cACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION without
```
vitalybu, line 1264: I don't understand this comment

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1171,6 +1171,28 @@ bool slave_thread::handle_cmdu_control_message(Socket *sd,
...
+            beerocks_message::cACTION_MONITOR_CHANNEL_DUMP_SCAN_RESULTS_REQUEST>(cmdu_tx);
+        if (!request_out) {
+            LOG(ERROR)
+                << "Failed building cACTION_MONITOR_CHANNEL_DUMP_SCAN_RESULTS_REQUEST message!";
+            return false;
+        }
+
+        LOG(DEBUG) << "send cACTION_MONITOR_CHANNEL_DUMP_SCAN_RESULTS_REQUEST";
+        message_com::send_cmdu(monitor_socket, cmdu_tx);
+        break;
```
vitalybu: This kind of code is used widely to forward messages between the different components.
Perhaps we should create a dedicated method for forwarding messages. Obviously it shouldn't be part of this PR.
@arnout @tomereli what do you guys think?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -26,7 +26,69 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
 {
     m_fsm_state = eState::INIT;
 }
+beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_dump()
+{
+    // When a dump is requested, do not set the scan_in_progress flag
+    // simply send the request as is, and wait for the results ready event
+
```
vitalybu, line 33: Remove the extra newline

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -26,7 +26,69 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+        beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST>(cmdu_tx);
+    if (request == nullptr) {
+        LOG(ERROR) << "Failed building message cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST";
+        return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
+    }
+
+    int32_t dwell_time_msec =
+        database.get_channel_scan_dwell_time_msec(m_radio_mac, m_is_single_scan);
+    if (dwell_time_msec < 0) {
+        LOG(ERROR) << "invalid dwell_time < 0";
```
vitalybu:
```suggestion
        LOG(ERROR) << "Invalid dwell_time = " << dwell_time_msec;
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -26,7 +26,69 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+        return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
+    }
+
+    //get current channel pool from DB
+    auto &curr_channel_pool = database.get_channel_scan_pool(m_radio_mac, m_is_single_scan);
+    if (curr_channel_pool.empty()) {
+        LOG(ERROR) << "empty channel pool is not supported. please set channel pool for mac="
+                   << m_radio_mac;
```
vitalybu, line 70: A better error message would be:
```C++
LOG(ERROR) << "Unable to trigger dynamic channel scan without a pre-defined channels pool. Please define the requested channels for radio: " << m_radio_mac;
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -26,7 +26,69 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+        return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
+    }
+
+    if (curr_channel_pool.size() > beerocks::message::SUPPORTED_CHANNELS_LENGTH) {
+        LOG(ERROR) << "channel_pool is too big";
```
vitalybu, line 75:
```suggestion
        LOG(ERROR) << "DCS supported channel list is too large: " << curr_channel_pool.size() << "/" << beerocks::message::SUPPORTED_CHANNELS_LENGTH;
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -58,63 +120,40 @@ void dynamic_channel_selection_task::work()
         LOG(TRACE) << "TRIGGER_SCAN, mac=" << m_radio_mac << ", scan_type is "
                    << ((m_is_single_scan) ? "single-scan" : "continuous-scan");
 
-        // When a scan is requested send the scan parameters Channel pool & Dwell time
-        // Before sending the request set the scan_in_progress flag to true
-        // So another scan would not trigger on the same radio simultaneously
-
+        // Before sending any request, set the scan_in_progress flag to true
+        // So another scan request would not launch on the same radio simultaneously
         database.set_channel_scan_in_progress(m_radio_mac, true, m_is_single_scan);
```
vitalybu, line 125: Should you first check that this flag is not already `true`?

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
itayx: Other not implemented functions also return false,
I'm not sure if it's a good thing to break consistency

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: This is how we "trick" the task into moving to the next current state
Incoming: ACTION_MONITOR_CHANNEL_DUMP_SCAN_RESULTS_REQUEST
Outgoing:
- ACTION_MONITOR_CHANNEL_DUMP_SCAN_RESULTS_RESPONSE
- ACTION_MONITOR_CHANNEL_SCAN_RESULTS_NOTIFICATION

Most requests only return a single response, but this message sends one response and one notification immediately

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: You cannot reach this state if the flag is already `true`

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Removed

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Changed

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Agree, Changed

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Changed

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Well, with the current tlvf, forwarding messages is not so simple, and in this case it even changes class (cACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_REQUEST -> cACTION_MONITOR_CHANNEL_DUMP_SCAN_RESULTS_REQUEST). So there are a few possibilities here.

- Ignore the issue. Many of the copies will disappear by themselves because we will use EasyMesh messages from the slave onwards instead of VS messages.
- Stop creating separate classes for each connection, and just use cACTION_CHANNEL_DUMP_SCAN_RESULTS_REQUEST; add a forwarding function to the slave.
- Make sure that all the details are in a substructure, so the forwarding is simply a copy of the substructure even when forwarding. We may have to add an assignment operator to tlvf to make sure this works.
- Allow all threads to connect to the bus directly, so no UDS between components is needed any more. This removes the need for forwarding entirely.

The latter has my personal preference as the long-term solution, but is possibly the most work. It also requires every platform to support a full transport layer, so we probably really need to replace zmq with something else then.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
kantera800: dummy mode currently doesn't have DCS  tests so its better to leave the false .

common/beerocks/bwl/nl80211/mon_wlan_hal_nl80211.cpp
kantera800: I think the false is correct since we don't have any DCS implementation in nl80211 bwl.

itayx: Just like Alex said, since there is no implementation I think returning false would be the correct way

agent/src/beerocks/monitor/monitor_thread.cpp
vitalybu: Why do we need a RESPONSE and a NOTIFICATION message?
Can't we use only a NOTIFICATION message?

When sending a response to a request, it'll contain the ID of the request.
When sending async notifications, the ID will remain 0.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
vitalybu: I don't see the change

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1964,7 +1964,7 @@ int cli_bml::start_dcs_single_scan(const std::string &radio_mac, int32_t dwell_t
     std::cout << "start_dcs_single_scan, mac=" << radio_mac << ", dwell_time=" << dwell_time
               << ", channel_pool=" << channel_pool << std::endl;
 
-    if (dwell_time <= 0) {
+    if (dwell_time < 0) {
         std::cout << __func__ << ", invalid input: dwell_time(" << dwell_time << ") <= 0"
```
vitalybu:
```suggestion
        std::cout << __func__ << ", invalid input: dwell_time(" << dwell_time << ") < 0"
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2236,8 +2236,8 @@ bool db::set_channel_scan_dwell_time_msec(const sMacAddr &mac, int dwell_time_ms
     LOG(DEBUG) << (single_scan ? "single" : "continuous")
                << ", dwell time msec = " << dwell_time_msec;
 
-    if (dwell_time_msec <= 0) {
-        LOG(ERROR) << "Only >0 dwell time is supported!";
+    if (dwell_time_msec < 0) {
+        LOG(ERROR) << "Only >=0 dwell time is supported!";
```
vitalybu:
```suggestion
        LOG(ERROR) << "Invalid dwell time: " << dwell_time_msec << ". Only positive values are supported!";
```

agent/src/beerocks/monitor/monitor_thread.cpp
itayx: Since the response is part of the request-response flow and the notification has nothing to do it .
The notification is sent when the driver sends a `scan_triggered` NL event
But when using `dwell time = 0` the we never request a scan from the driver so the Nl event will never be sent

so to "trick" the FSM in the dcs' task to work as if a scan was triggered, we send the notification back manually while not actually triggering the bml channel scan

controller/src/beerocks/master/db/db.cpp

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: Changed

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -104,12 +104,14 @@ eActionOp_CONTROL:
...
   ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST : 140
   ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE : 141
-  ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION : 142
-  ACTION_CONTROL_CHANNEL_SCAN_RESULTS_NOTIFICATION : 143
-  ACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION : 144
-  ACTION_CONTROL_CHANNEL_SCAN_FINISHED_NOTIFICATION : 145
-
-  ACTION_CONTROL_ENUM_END: 146
+  ACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_REQUEST: 142
+  ACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_RESPONSE: 143
```
arnout: It's weird that you add the two new messages in the middle instead of at the end (i.e. the two new ones would be numbered 146 and 147).

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_control.yaml
```diff
@@ -368,6 +368,12 @@ cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE:
   _type: class
   success: uint8_t
 
+cACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_REQUEST:
+  _type: class
+
+cACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_RESPONSE:
+  _type: class
+  success: uint8_t
```
arnout: Is this message even useful? I'd expect the reply to be either a cACTION_CONTROL_CHANNEL_SCAN_RESULTS_NOTIFICATION or something like cACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -104,12 +104,14 @@ eActionOp_CONTROL:
 
   ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST : 140
   ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE : 141
-  ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION : 142
-  ACTION_CONTROL_CHANNEL_SCAN_RESULTS_NOTIFICATION : 143
-  ACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION : 144
-  ACTION_CONTROL_CHANNEL_SCAN_FINISHED_NOTIFICATION : 145
-
-  ACTION_CONTROL_ENUM_END: 146
+  ACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_REQUEST: 142
```
arnout: It would be nicer to make them more similar to the other message names, i.e. `CHANNEL_SCAN_DUMP_RESULTS_REQUEST`.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_control.yaml
```diff
@@ -368,6 +368,12 @@ cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE:
   _type: class
   success: uint8_t
 
+cACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_REQUEST:
```
arnout: The commit message that introduces this is fairly confusing. It speaks of "dwell time" but that's pretty unrelated to these new messages. IIUC, the idea is to use the a dwell time of 0 in the user interface (i.e. BML) to indicate that you want to use cached results instead of doing a scan. But these two new messages are just about using cached results. Therefore, I'd reword the commit message as follows:

```
 tlfv: Add dump results request and response

The dynamic channel scan will switch channels to scan for beacons
from neighboring APs. Sometimes, we may want to get scan results
without switching channels, e.g. cached results or results from scanning
the channel that the AP is operating on.

Add two new messages, CHANNEL_DUMP_SCAN_RESULTS_REQUEST
and CHANNEL_DUMP_SCAN_RESULTS_RESPONSE, to get scan results
without switching channels. The new messages are added both to controller
and monitor.
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3349,6 +3349,10 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
         LOG(TRACE) << "ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE for mac " << hostap_mac;
         break;
     }
+    case beerocks_message::ACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_RESPONSE: {
+        LOG(TRACE) << "ACTION_CONTROL_CHANNEL_DUMP_SCAN_RESULTS_RESPONSE for mac " << hostap_mac;
+        break;
+    }
```
arnout: As I wrote: the response isn't used for anything, so there is no need to add it.

There is of course the symmetry of having a request/response for everything. However, the symmetry is already broken a bit through the different notification messages. And for sure, I don't think it warrants adding roughly 100 lines of code just for symmetry.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -26,7 +26,69 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
+    auto request = beerocks::message_com::create_vs_message<
+        beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST>(cmdu_tx);
+    if (request == nullptr) {
+        LOG(ERROR) << "Failed building message cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST";
+        return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
+    }
+
+    int32_t dwell_time_msec =
+        database.get_channel_scan_dwell_time_msec(m_radio_mac, m_is_single_scan);
+    if (dwell_time_msec < 0) {
```
arnout: This should be `<= 0` IMO.

arnout (approved):
None of my comments are essential in practice. Since this anyway requires a manual rebase, I'm approving as is, and you can take the comments into account as you wish.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
itayx: I wanted to add them in relation to where they would be used in the scan flow
And since the `scan dump request` is an alternative version of the `scan trigger request` I thought it was best to add them after the scan trigger request (and response)

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_control.yaml
itayx: You're right,
changed the commit message

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_control.yaml
itayx: You're right the message isn't in use other then to align with the CMDU symmetry
It also has a side benefit when it comes to logs, as a confirmation that the request was sent correctly and that the dump request operation finished successfully
Note: This does not mean that the dump is successful, just that the request operation was

agent/src/beerocks/slave/son_slave_thread.cpp
itayx: I think the best course would be to use the second option mentioned by @arnout, using a non-separate messages for forward-only message would reduce the code added, all that would be needed is a static function in the slave, this method should only be used if the messages do not have any additional function in the slave.

controller/src/beerocks/master/son_master_thread.cpp
itayx: I partly agree, while I do think that adding code for symmetry's sake is not a good reason, I have used the additional logs that the response prints to validate that the request has gone through and has performed correctly so I still see a use for the request/response symmetry, even if the only reason is logging

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: I understand why, but keep in mind that the only way to reach this code is if the dwell time `!= 0`
Changed to `<= 0` anyway

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3383,6 +3383,10 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
         LOG(TRACE) << "ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE for mac " << hostap_mac;
         break;
     }
+    case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE: {
+        LOG(TRACE) << "ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE for mac " << hostap_mac;
```
morantr, line 3387: "... the response is only used as confirmation for the request".
I would expect to see the confirmation. Sending an event to the pending caller, or something similar.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -26,7 +26,69 @@ dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
...
     m_fsm_state = eState::INIT;
 }
+beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_dump()
+{
+    // When a dump is requested, do not set the scan_in_progress flag
+    // simply send the request as is, and wait for the results ready event
+
+    auto request = beerocks::message_com::create_vs_message<
+        beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_REQUEST>(cmdu_tx);
+    if (request == nullptr) {
```
morantr: For new code, better use `!share_ptr` than `shared_ptr == nullptr`.
You repeat it several times in this commit.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1953,9 +1953,9 @@ int cli_bml::start_dcs_single_scan(const std::string &radio_mac, int32_t dwell_t
     std::cout << "start_dcs_single_scan, mac=" << radio_mac << ", dwell_time=" << dwell_time
               << ", channel_pool=" << channel_pool << std::endl;
 
-    if (dwell_time <= 0) {
-        std::cout << __func__ << ", invalid input: dwell_time(" << dwell_time << ") <= 0"
-                  << std::endl;
+    if (dwell_time < 0) {
+        std::cout << __func__ << ", invalid input: dwell_time(" << dwell_time
+                  << ") < 0. Only positive values are supported!" << std::endl;
```
morantr: This commit (67a5cb412e2f938b4334e7fc5fd9ed6dba300c10: beerocks_cli_bml: start_dcs_single_scan: remove dwell time 0 varifica…) headline is too long and exeeds 72 characters.

Also, the print would look like this:
"invalid input: dwell_time(-3) < 0. Only positive values are supported!"

I think it would look better like this:
"invalid input, dwell_time='-3', Only positive values are supported!
(nitpick)

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2236,8 +2236,9 @@ bool db::set_channel_scan_dwell_time_msec(const sMacAddr &mac, int dwell_time_ms
     LOG(DEBUG) << (single_scan ? "single" : "continuous")
                << ", dwell time msec = " << dwell_time_msec;
 
-    if (dwell_time_msec <= 0) {
-        LOG(ERROR) << "Only >0 dwell time is supported!";
+    if (dwell_time_msec < 0) {
+        LOG(ERROR) << "Invalid dwell time: " << dwell_time_msec
+                   << ". Only positive values (>=0) are supported!";
```
morantr: There is no need to mention `(>=0)` since it is the definition of `positive number`.

morantr (approved):
Small comments. I approved and labeled with "don't merge" so you will have a chance to change it before it automatically merged.

controller/src/beerocks/master/son_master_thread.cpp
itayx: The confirmation is due to the Log prints,
It's a way to print in the controller log that the action sent from the controller was successful

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Changed

controller/src/beerocks/master/db/db.cpp
itayx: I wanted to remove the ambiguity with `positive number` and a `0` value

controller/src/beerocks/cli/beerocks_cli_bml.cpp
itayx: I agree it looks better without the `<0`
Changed

PR #819, opened 2020-02-12
WSC: proper padding of public key

### LINKED ISSUE #560: [BUG] agent: occasional slave reset after join attempt
Occasionally, a slave resets when it receives M2 (i.e. during onboarding or after autoconfig renew).

## Steps to reproduce

Unfortunately, there is no reliable reproducer.

## Target

Ubuntu in docker with test_flows.sh.

## Analysis

During one reproduction, I was able to add some logging which led me to determine the following:

- slave reset is due to authentication failure of M2;
- It is the "authenticator" of the entire M2 that fails (so not the keywrap authenticator);
- M1 and M2 used to calculate the authenticator are the same.

That leaves the following possible reasons for authentication failure:

- authkey is different;
- M1 and M2 are the same but the buffer over which authenticator is calculated is not (i.e. something wrong when copying);
- length of the buffer is different;
- comparison is wrong;
- algorithm to compute authenticator is wrong.

arnout:
Fixes #560.

The first commit adds more debug output that helps to analyse the issue.
It allowed me to identify the problem: the public key is sometimes shorter than 192 bytes.
Checking the WSC specification, it turns out that we must left-pad it with zeroes in that case.
We did right-padding.

The second commit forces the public key generation in this corner case, by re-generating the key until it's shorter than 192 bytes.

The third commit fixes the issue.

Since we don't want the changes of the first and definitely not the second commit in master, the following two commits revert them.
I could of course have rebased them away, but I think it's useful to have this in the history.
That way, if we ever need it again, we can just cherry-pick it.
Note that because the final rebase done by github will generate new commit hashes, I removed the commit hash references from the revert commit messages.

The next commit removes some more debug information that was added months ago to debug this and other issues with WSC.
Since WSC seems to be stable now, we can remove that as well.

The final commit is just something I had lying around and never pushed.
It speeds up my clang-format runs with a factor of two.


framework/common/encryption.cpp
```diff
@@ -134,6 +135,14 @@ bool diffie_hellman::compute_key(uint8_t *key, unsigned &key_length, const uint8
     return true;
 }
 
+void diffie_hellman::copy_pubkey(uint8_t *dest) const
```
vitalybu: That's not really diffie hellman related logic.
Perhaps it can be moved into a generic "copy with padding" (left or right) function in one of the utility classes.

tomereli: Agree, but I think its rare that we'll need such a function, so I guess this should be a (nitpick)

framework/common/encryption.cpp
```diff
@@ -134,6 +135,14 @@ bool diffie_hellman::compute_key(uint8_t *key, unsigned &key_length, const uint8
...
+{
+    int padding = max_pubkey_length - m_pubkey_length;
+    mapf_assert(padding >= 0);
```
tomereli: This is not really doing anything in release builds, I think it is more appropriate to at least add a LOG(FATAL) here.

framework/common/include/mapf/common/encryption.h
```diff
@@ -71,6 +79,11 @@ class diffie_hellman {
      */
     unsigned nonce_length() const { return sizeof(m_nonce); }
 
+    /**
+     * @brief Maximum length of pubkey().
+     */
+    static const unsigned max_pubkey_length = 192;
```
tomereli, line 77: I'm confused - It looks like the pubkey length is exactly 192 with the new logic which is added. Shouldn't we change the terminology to always treat the pubkey as 192 bytes? (nitpick)

.clang-ignore
```diff
@@ -1,3 +1,5 @@
 framework/external/easylogging/
 framework/tlvf/AutoGenerated
 common/beerocks/tlvf/AutoGenerated
+^build.*
```
tomereli, line 4: OHHHH this is why clang-format became so slow!

tomereli (approved):
I really like the way this is split to commits which show how the issue is reproduced, debug, and fixed. It can teach our young padawans a lot on how to debug complex issues!

@CoralMalachi @LiorAmram @Galvien

framework/common/encryption.cpp
arnout: The idea was that this would be the only way to access the pubkey, i.e. the pubkey() method would be removed. Annoyingly, though, it is (or rather was) used for debug output.

framework/common/encryption.cpp
arnout: That should be part of mapf_assert then.

framework/common/include/mapf/common/encryption.h
arnout: The pubkey can be smaller (i.e. pubkey_length() can be smaller). But copy_pubkey always writes to a 192-byte buffer.

framework/common/encryption.cpp
arnout: So, If I'd make an external function `copy_pubkey(const diffie_helman &dh, uint8_t *dest)`, you would be happy?

framework/common/encryption.cpp
```diff
@@ -134,6 +135,14 @@ bool diffie_hellman::compute_key(uint8_t *key, unsigned &key_length, const uint8
...
+    std::fill_n(dest, padding, 0);
```
rmelotte: +1 for std::fill instead of memset.

rmelotte (approved):
I was on the point of commenting on the second commit, before reading the rest.... :-)

Excellent!

framework/common/encryption.cpp
dob71: Agree on rare/limited use, so IMO should not be  a blocker for merging.

dob71 (approved):

framework/common/encryption.cpp
arnout: Moved to global instead of diffie_hellman member.

PR #823, opened 2020-02-12
Bugfix/4.4.3

### LINKED ISSUE #360: [TASK] Pass flow 4.4.3 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.4.3 in the certification testbed.

# Exit criteria

### LINKED ISSUE #362: 4.4.3 AP configuration BSS Tear Down
The prplMesh agent on Turris Omnia with OpenWrt needs to pass the certification test case 4.4.3 as defined by the Wi-Fi CERTIFIED EasyMesh (tm) Test Plan.

- [ ] #361 Test flow for 4.4.3 AP configuration BSS Tear Down
- [x] #791 Update VAP list after VAP config
- [ ] #360 Pass flow 4.4.3 test in testbed

tomereli:
agent: refactor M2 teardown handling

According to thethe MultiAP specification:
"If a Multi-AP Controller receives a WSC TLV containing an M1, then it
shall respond within one second with either 1) one or more M2s for each
BSS to be configured on the radio identified in the AP Radio Identifier
TLV or 2) one M2 with bit 4 (Tear down bit) set to 1 in the Multi-AP
Extension subelement to indicate that zero BSS are to be configured on
the radio identified in the AP Radio Identifier TLV per section 17.1.3."

In practice, it seems that some controllers ignore this and send
multiple M2 TLVs for the same RUID with teardown bit set.

In addition, some controllers simply sends an empty M2 (with invalid attributes)
instead of explicitly setting teardown.

Therefore, refactor the M2 teardown handling in the agent as follows:
- Invalid M2 TLV - teardown
- Teardown bit set but more than one M2 - teardown

Also update test_flows.sh accordingly since the debug messages have
changed.

Fixes #360 


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3965,9 +3965,16 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
...
     }
+    // Update VAP configuration
+    config.auth_type   = config_data->auth_type();
+    config.encr_type   = config_data->encr_type();
+    config.bssid       = config_data->bssid();
+    config.network_key = config_data->network_key();
+    config.ssid        = config_data->ssid();
+    config.bss_type    = static_cast<WSC::eWscVendorExtSubelementBssType>(config_data->bss_type());
     // Swap to network byte order for KWA HMAC calculation
+    // from this point config data is not readable!
```
arnout, line 3972: Good idea to add this!

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4104,13 +4091,7 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
     }
 
-    auto request = message_com::create_vs_message<
-        beerocks_message::cACTION_APMANAGER_WIFI_CREDENTIALS_UPDATE_REQUEST>(cmdu_tx);
-    if (!request) {
-        LOG(ERROR) << "Failed building message!";
-        return false;
-    }
-
+    std::list<WSC::configData::config> configs;
```
arnout: std::list should only be used for lists where you insert in the middle relatively frequently. Use std::vector [SL.con.2](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rsl-vector)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3965,9 +3965,16 @@ bool slave_thread::autoconfig_wsc_parse_m2_encrypted_settings(
         LOG(ERROR) << "invalid config data length";
         return false;
     }
+    // Update VAP configuration
+    config.auth_type   = config_data->auth_type();
+    config.encr_type   = config_data->encr_type();
+    config.bssid       = config_data->bssid();
+    config.network_key = config_data->network_key();
+    config.ssid        = config_data->ssid();
+    config.bss_type    = static_cast<WSC::eWscVendorExtSubelementBssType>(config_data->bss_type());
```
arnout, line 3970: I would have liked this refactoring to be in a separate commit. (nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4121,32 +4102,49 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
-        LOG(DEBUG) << "Controller configuration (WSC M2 Encrypted Settings)";
-        LOG(DEBUG) << "     Manufacturer: " << m2.manufacturer()
-                   << ", ssid: " << credentials->ssid_str()
-                   << ", bssid: " << credentials->bssid_attr().data
-                   << ", authentication_type: " << std::hex
-                   << int(credentials->authentication_type_attr().data)
-                   << ", encryption_type: " << int(credentials->encryption_type_attr().data)
-                   << (fronthaul ? " fronthaul" : "") << (backhaul ? " backhaul" : "")
-                   << (teardown ? " teardown" : "");
+        WSC::configData::config config_;
```
arnout: Why `config_` and not just `config`?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4121,32 +4102,49 @@ bool slave_thread::handle_autoconfiguration_wsc(Socket *sd, ieee1905_1::CmduMess
...
+        if (!autoconfig_wsc_parse_m2_encrypted_settings(m2, authkey, keywrapkey, config_)) {
+            LOG(ERROR) << "Invalid config data, tear down radio";
```
arnout: I think we'll need to revisit this at some point. From a security perspective, if authentication fails, we should do nothing instead of tearing down. If config data is invalid it's a bit uncertain what we should do.

tests/test_flows.sh
```diff
@@ -123,8 +123,6 @@ test_initial_ap_config() {
     check_log ${REPEATER1} agent_wlan2 "WSC Global authentication success"
     check_log ${REPEATER1} agent_wlan0 "KWA (Key Wrap Auth) success"
     check_log ${REPEATER1} agent_wlan2 "KWA (Key Wrap Auth) success"
-    check_log ${REPEATER1} agent_wlan0 "Controller configuration (WSC M2 Encrypted Settings)"
-    check_log ${REPEATER1} agent_wlan2 "Controller configuration (WSC M2 Encrypted Settings)"
```
arnout, line 127: Just removing them is not nice. Can't we check something here?

arnout (approved):

rmelotte (approved):
:+1:

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: You are absolutely right - I rather do this right now than postpone it.
I changed the way we parse config data so that there is a difference between encryption/authentication failures - in which we ignore the config data, and empty config_data which is also invalid according to Table 20 in the WSC specification but Marvell controller sends it none the less.

tests/test_flows.sh
tomereli: Fixed and brought it back.

PR #826, opened 2020-02-13
tools: deploy_ipk.sh: fix missing quotes when using eval

rmelotte:
Since 96c75adedf824dbab4dd0abd9ccc10b3473f94d4 eval is used to expand
the dynamically built SSH_OPTIONS.

Without an additional level of quoting, '&&' ends the arguments to the
ssh command, and "uci commit" gets executed locally.

Add another level of quoting around the uci commands so that they are
fed to ssh as a single argument.

### DIFF
--- a/tools/deploy_ipk.sh
+++ b/tools/deploy_ipk.sh
@@ -46,7 +46,7 @@ EOF
 
     if [ "$CERTIFICATION_MODE" = true ] ; then
         echo "Certification mode will be enabled on the target"
-        eval ssh "$SSH_OPTIONS" "$TARGET" "uci set prplmesh.config.certification_mode=1 && uci commit"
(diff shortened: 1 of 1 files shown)


tomereli (approved):

arnout (approved):
It would be nicer if there was a function to do the ssh.

Alternatively, the quoting and eval could be avoided entirely by using `-oBatchMode=yes` etc., and removing the quoting around `$SSH_OPTIONS`.

But this works so let's merge it.

rmelotte:
See also my comments in [#676](https://github.com/prplfoundation/prplMesh/pull/676#discussion_r378700877)
Maybe unquoting SSH_OPTIONS would be a compromise since we use very few of them. For some of the options it wouldn't work though...

PR #829, opened 2020-02-13
github/workflows: add merge when ready workflow

arnout:
Using [PR merge bot](https://github.com/marketplace/actions/pr-merge-bot), automatically merge a pull request when the
checks pass and the MergeWhenReady tag is set.

### DIFF
--- a/.github/workflows/merge_when_ready.yml
+++ b/.github/workflows/merge_when_ready.yml
@@ -1,6 +1,8 @@
 name: Merge Bot
 
 on:
+  check_suite:
+    types: [completed]
   pull_request:
     types:
       - labeled # MergeWhenReady added
@@ -27,3 +29,4 @@ jobs:
         checks_enabled: true
         reviewers: true
         method: rebase
+        test: true # make the bot comment instead of taking action
--- a/.github/workflows/merge_when_ready.yml
+++ b/.github/workflows/merge_when_ready.yml
@@ -3,16 +3,6 @@ name: Merge Bot
 on:
   check_suite:
     types: [completed]
-  pull_request:
-    types:
-      - labeled # MergeWhenReady added
-      - review_request_removed # Blocking review removed
-      - review_requested
-      - unlabeled # don't merge label removed
-  pull_request_review:
-    types:
-      - dismissed # Blocking review removed
-      - submitted # Approving review added
   status: # CI
 
 jobs:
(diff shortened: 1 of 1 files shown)


arnout:
The output from merge bot is not very promising:
```
Run squalrus/merge-bot@v0.2.0
  with:
    GITHUB_TOKEN: ***
    labels: MergeWhenReady
    blocking-labels: don't merge
    checks: true
    method: rebase
    test: false
    reviewers: true
test mode: false
required reviewers: true
required labels: ["MergeWhenReady"]
blocking labels: ["don't merge"]
merge method: rebase
##[error]Not Found
##[error]Node run failed with exit code 1
```

I expect the action to fail, but not with "Not found"...

.github/workflows/merge-when-ready.yml
```diff
@@ -0,0 +1,30 @@
...
+      - submitted # Approving review added
+  status:
+
+jobs:
+  merge:
+    runs-on: ubuntu-latest
+    name: Merge when ready
+    steps:
+    - name: Integration check
+      uses: squalrus/merge-bot@v0.2.0
```
rmelotte: Why "v0.2.0"? Is it intentional? It looks like the latest one is "v0.4.0"

arnout: He, I just copy-pasted from the example... Will fix.

rmelotte (commented):
[Just adding a review comment to test the bot.]

rmelotte:
It looks like checks do not include the reviews.

arnout:
> It looks like checks do not include the reviews.

Darn. I assumed that the `checks: true` part would cover the reviews.

I intentionally turned the reviews off because according to the documentation says "This enforces a reviewer mode where there cannot be any pending reviews". This basically means that *all* code owners must review.

I'll try again with reviews and when we get two accepts we can check if it works.

rmelotte:
Yes the behavior with the reviews doesn't exactly match what we used to do.
It also seems that if it has run once on one commit, it won't run again when the other checks completed (which defeats the purpose here, since we want it to run when the CI has finished).

It's not perfect, but if we could at least make it run one time /after/ all the other checks completed, we could already improve our current workflow:
- when you are certain that the PR is ready to be merged, rebase and add a tag.
- with the tag, the merge bot runs once CI has finished and merge it if it passed.

If I understand the way it works correctly, it wouldn't work if you retry a failed pipeline for example.

rmelotte (approved):

arnout:
Since it doesn't get triggered when CI finishes, I don't think this is useful. So I'm closing this PR.

Issue #782, opened 2020-02-03
Send standard NL80211 commands using DWPAL


ghost:
From the description of the problem by @mariomaz :

In order to obtain AP VT/VHT/HE capabilities using Netlink and DWPAL we should be sending command NL80211_CMD_GET_WIPHY (as in ap_wlan_hal_nl80211::read_supported_channels).

However, this is something we cannot do as method base_wlan_hal_dwpal::dwpal_nl_cmd_get can be used to send NL80211_CMD_VENDOR command only. This method is internally calling dwpal_driver_nl_cmd_send(), using NL80211_CMD_VENDOR and given sub-command. We should then create another but similar method to call dwpal_driver_nl_cmd_send() with NL80211_CMD_GET_WIPHY instead. However, dwpal_driver_nl_cmd_send() implementation in prplWrt only supports NL80211_CMD_VENDOR command:

 ```c

        if (nl80211Command != NL80211_CMD_VENDOR /*0x67*/)
        {
                 console_printf("%s; non supported command (0x%x); currently we support ONLY NL80211_CMD_VENDOR (0x67) ==> Abort!\n", __FUNCTION__, (unsigned int)nl80211Command);
                 return DWPAL_FAILURE;
        }
```

tomereli:
Answer is that it is OK to add implementation in DWPAL. Once it is there, we should port it back to Intel so we can get it in the next prplWrt upgrade.

tomereli:
@mariomaz please implement what you need to support this in DWPAL, create a PR to our gitlab fork which updates dwpal_6x-uci-1.0.0.1.tar.bz2 in https://git.prpl.dev/prplmesh/iwlwav/blob/iwlwav-19.07/dl/dwpal_6x-uci-1.0.0.1.tar.bz2 and https://git.prpl.dev/prplmesh/iwlwav/blob/iwlwav-19.07/dwpal-6x-uci/Makefile.
Then also create a patch and send it in response to the mail from Gur Sannikov, He will port it to Intel.

kantera800:
@mariomaz you are using an older dwpal version than INTEL's repositories so apparently you would need to implement it.

dwpal_driver_nl_msg_get() already supports  non-vendor commands.

I checked the code and you can use NL80211_CMD_GET_SCAN implementation as reference (see dwpal_driver_nl_scan_dump).

you have 2 options:
[...]

PR #831, opened 2020-02-14
Feature/collect link metrics

### LINKED ISSUE #390: [TASK] Link metric collection for Ethernet links
## Description

The agent should collect link metrics for Ethernet links.
- [ ] Specific neighbor at wired backhaul link (i.e. controller)
- [ ] Invalid neighbor and other neighbors (requires topology database to be available)
- [ ] All neighbors (requires topology database to be available)

## Exit criteria

- A mechanism exists to collect link metrics from the Ethernet interfaces registered in the agent.
- LinkMetricsResponse TLVs are filled in with the correct data.
- Test flow with dummy link metrics.

## Detailed description

To be decided how to collect link metrics. We may need to do it separately for wired and wireless links. Probably the cleanest is to have a separate abstraction library for it, even for wireless. Somehow, it should allow for different implementations for different interfaces. Link metric collection for WiFi links will be treated in a sibling task.

It's not very clear how this is going to be implemented for Ethernet links, because there is no way to know which part of the traffic goes to which 1905 neighbour - we only have the total traffic per port. I guess we'll just assume there are no non-1905 bridges/switches on the Ethernet link.

Finally, it has to be decided who is responsible for the link metric collection. Is it the agent or the topology manager?

mariomaz:
See "[TASK] Link metric collection for Ethernet links #390" for a detailed description of the task. 

Currently we only support collecting link metrics for Ethernet interfaces and it is done partially, because we need topology information to know which are the neighbors of the node. This is not available yet. 

Collecting metrics for WiFi interfaces requires adding support for standard commands in DWPAL. See "Send standard NL80211 commands using DWPAL #782" for more information.

For tests_flows.sh to pass, "Fix several problems with the Device Information TLV #825" should be merged first. 

I preferred to create a draft-PR so you can verify the design of the solution and give early feedback. Also to be particularly reviewed is the way MAC address of requested specific neighbor is compared against MAC address of the controller.

### DIFF
--- a/agent/src/beerocks/slave/CMakeLists.txt
+++ b/agent/src/beerocks/slave/CMakeLists.txt
@@ -6,6 +6,9 @@ set(MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
 
 file(GLOB beerocks_agent_sources 
     ${MODULE_PATH}/backhaul_manager/*.c*
+    # This code should be moved to BPL 
+    # [TASK] Move link metric related classes to BPL #910
+    ${MODULE_PATH}/link_metrics/*.c*
     ${MODULE_PATH}/platform_manager/*.c*
     ${MODULE_PATH}/*.c*
 )
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2215,21 +2215,26 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
      * Set the number of local interfaces and fill info of each of the local interfaces, according
      * to IEEE_1905 section 6.4.5
      */
-    uint32_t speed;
-    if (network_utils::linux_iface_get_speed(m_sConfig.wire_iface, speed)) {
+
+    /**
+     * Add a LocalInterfaceInfo field for the wired interface, if any.
+     */
+    std::string local_interface_name = m_sConfig.wire_iface;
+    if (!local_interface_name.empty() &&
+        network_utils::linux_iface_is_up_and_running(local_interface_name)) {
+        ieee1905_1::eMediaType media_type = ieee1905_1::eMediaType::UNKNONWN_MEDIA;
+        if (!get_media_type(local_interface_name, ieee1905_1::eMediaTypeGroup::IEEE_802_3,
+                            media_type)) {
+            LOG(ERROR) << "Unable to compute media type for interface " << local_interface_name;
+            return false;
+        }
+
         std::shared_ptr<ieee1905_1::cLocalInterfaceInfo> localInterfaceInfo =
             tlvDeviceInformation->create_local_interface_list();
 
-        ieee1905_1::eMediaType media_type = ieee1905_1::eMediaType::UNKNONWN_MEDIA;
-        if (SPEED_100 == speed) {
-            media_type = ieee1905_1::eMediaType::IEEE_802_3U_FAST_ETHERNET;
-        } else if (SPEED_1000 <= speed) {
-            media_type = ieee1905_1::eMediaType::IEEE_802_3AB_GIGABIT_ETHERNET;
-        }
-
         // default to zero mac if get_mac fails.
         std::string wire_iface_mac = network_utils::ZERO_MAC_STRING;
-        network_utils::linux_iface_get_mac(m_sConfig.wire_iface, wire_iface_mac);
+        network_utils::linux_iface_get_mac(local_interface_name, wire_iface_mac);
         localInterfaceInfo->mac()               = network_utils::mac_from_string(wire_iface_mac);
         localInterfaceInfo->media_type()        = media_type;
         localInterfaceInfo->media_info_length() = 0;
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -8,6 +8,7 @@
 
 #include "backhaul_manager_thread.h"
 
+#include "../link_metrics/ieee802_3_link_metrics_collector.h"
 #include "../tlvf_utils.h"
 
 #include <bcl/beerocks_utils.h>
@@ -30,6 +31,8 @@
 #include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvDeviceInformation.h>
 #include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
+#include <tlvf/ieee_1905_1/tlvLinkMetricQuery.h>
+#include <tlvf/ieee_1905_1/tlvLinkMetricResultCode.h>
 #include <tlvf/ieee_1905_1/tlvMacAddress.h>
 #include <tlvf/ieee_1905_1/tlvReceiverLinkMetric.h>
 #include <tlvf/ieee_1905_1/tlvSearchedRole.h>
@@ -77,6 +80,227 @@ namespace beerocks {
 //////////////////////////////////////////////////////////////////////////////
 
 const char *backhaul_manager::s_arrStates[] = {FOREACH_STATE(GENERATE_STRING)};
+
+/**
+ * @brief Gets media type group for given interface.
+ *
+ * @param[in] interface_name Name of the local interface.
+ * @param[in, out] media_type_group The media type group of the connecting interface.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_media_type_group(const std::string &interface_name,
+                                 ieee1905_1::eMediaTypeGroup &media_type_group)
+{
+    // TODO: Currently, only Ethernet is supported
+    // When dealing with WiFi interfaces in task #792, use NL80211_CMD_GET_INTERFACE command to
+    // find out if media type group is WiFi
+    media_type_group = ieee1905_1::eMediaTypeGroup::IEEE_802_3;
+
+    return true;
+}
+
+/**
+ * @brief Gets media type for given interface.
+ *
+ * The mechanism to use to obtain media type depends on the media type group:
+ * Ethernet, WiFi, MoCA, etc.
+ *
+ * @param[in] interface_name Name of the local interface.
+ * @param[in] media_type_group The media type group of the connecting interface.
+ * @param[in, out] media_type The underlying network technology of the connecting interface.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_media_type(const std::string &interface_name,
+                           ieee1905_1::eMediaTypeGroup media_type_group,
+                           ieee1905_1::eMediaType &media_type)
+{
+    bool result = false;
+
+    if (ieee1905_1::eMediaTypeGroup::IEEE_802_3 == media_type_group) {
+        uint32_t speed;
+        if (network_utils::linux_iface_get_speed(interface_name, speed)) {
+            if (SPEED_100 == speed) {
+                media_type = ieee1905_1::eMediaType::IEEE_802_3U_FAST_ETHERNET;
+            } else if (SPEED_1000 <= speed) {
+                media_type = ieee1905_1::eMediaType::IEEE_802_3AB_GIGABIT_ETHERNET;
+            } else {
+                media_type = ieee1905_1::eMediaType::UNKNONWN_MEDIA;
+            }
+
+            result = true;
+        }
+    } else if (ieee1905_1::eMediaTypeGroup::IEEE_802_11 == media_type_group) {
+        // TODO: Not supported yet
+        LOG(ERROR) << "IEEE_802_11 media is not supported yet";
+    } else if (ieee1905_1::eMediaTypeGroup::IEEE_1901 == media_type_group) {
+        // TODO: Not supported yet
+        LOG(ERROR) << "IEEE_1901 media is not supported yet";
+    } else if (ieee1905_1::eMediaTypeGroup::MoCA == media_type_group) {
+        // TODO: Not supported yet
+        LOG(ERROR) << "MoCA media is not supported yet";
+    } else {
+        media_type = ieee1905_1::eMediaType::UNKNONWN_MEDIA;
+        result     = true;
+    }
+
+    return result;
+}
+
+/**
+ * @brief Adds link metrics to response message.
+ *
+ * Gets link metrics of the interface between the reporter AL MAC and the neighbor AL MAC.
+ * Creates a Transmitter Link Metric TLV or a Receiver Link Metric TLV or both and adds them to
+ * the response message.
+ *
+ * @param[in] local_interface_name Name of the local interface.
+ * @param[in] media_type The underlying network technology of the connecting interface.
+ * @param[in] reporter_al_mac 1905.1 AL MAC address of the device that transmits the response message.
+ * @param[in] neighbor_al_mac 1905.1 AL MAC address of a neighbor of the receiving device.
+ * @param[in] link_metrics_type The link metrics type requested: TX, RX or both.
+ * @param[in, out] cmdu_tx CMDU response message where TLVs must be added to.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool add_link_metrics(const std::string &local_interface_name,
+                             ieee1905_1::eMediaType media_type, const sMacAddr &reporter_al_mac,
+                             const sMacAddr &neighbor_al_mac,
+                             ieee1905_1::eLinkMetricsType link_metrics_type,
+                             ieee1905_1::CmduMessageTx &cmdu_tx)
+{
+    /**
+     * Metrics information associated to the link between the local interface and the neighbor's interface.
+     */
+    sLinkMetrics link_metrics;
+
+    /**
+     * Link metrics collector to use to get link metrics in the connecting interface.
+     */
+    std::unique_ptr<link_metrics_collector> collector;
+
+    /**
+     * Get a link metrics collector suitable for the underlying network technology of the
+     * connecting interface.
+     * Collector choice depends on bits 15 to 8 of media type.
+     */
+    uint8_t media_type_msb = media_type >> 8;
+    if (0 == media_type_msb) {
+        collector = std::make_unique<ieee802_3_link_metrics_collector>();
+    } else if (1 == media_type_msb) {
+        // TODO: Create a collector for wireless interfaces
+        //collector = std::make_unique<ieee802_11_link_metrics_collector>();
+        LOG(ERROR) << "Unsupported media type: " << std::hex << (int)media_type;
+    } else {
+        LOG(ERROR) << "Unsupported media type: " << std::hex << (int)media_type;
+        return false;
+    }
+
+    /**
+     * Get link metrics information
+     */
+    if (collector &&
+        collector->get_link_metrics(local_interface_name, neighbor_al_mac, link_metrics)) {
+
+        /**
+         * Get MAC address for local interface.
+         */
+        std::string mac;
+        if (!network_utils::linux_iface_get_mac(local_interface_name, mac)) {
+            LOG(ERROR) << "Failed getting MAC address for interface: " << local_interface_name;
+            return false;
+        }
+        sMacAddr local_interface_mac = network_utils::mac_from_string(mac);
+
+        /**
+         * Add Transmitter Link Metric TLV if specifically requested or both requested
+         */
+        if ((ieee1905_1::eLinkMetricsType::TX_LINK_METRICS_ONLY == link_metrics_type) ||
+            (ieee1905_1::eLinkMetricsType::BOTH_TX_AND_RX_LINK_METRICS == link_metrics_type)) {
+            auto tlvTransmitterLinkMetric =
+                cmdu_tx.addClass<ieee1905_1::tlvTransmitterLinkMetric>();
+            if (!tlvTransmitterLinkMetric) {
+                LOG(ERROR) << "addClass ieee1905_1::tlvTransmitterLinkMetric failed";
+                return false;
+            }
+
+            tlvTransmitterLinkMetric->reporter_al_mac() = reporter_al_mac;
+            tlvTransmitterLinkMetric->neighbor_al_mac() = neighbor_al_mac;
+
+            if (!tlvTransmitterLinkMetric->alloc_interface_pair_info()) {
+                LOG(ERROR) << "alloc_interface_pair_info failed";
+                return false;
+            }
+            auto interface_pair_info = tlvTransmitterLinkMetric->interface_pair_info(0);
+            if (!std::get<0>(interface_pair_info)) {
+                LOG(ERROR) << "Failed accessing interface_pair_info";
+                return false;
+            }
+            auto interfacePairInfo             = std::get<1>(interface_pair_info);
+            interfacePairInfo.rc_interface_mac = local_interface_mac;
+            // TODO: This is not correct... We actually have to get this from the topology
+            // discovery message, which will give us the neighbor interface MAC.
+            interfacePairInfo.neighbor_interface_mac    = neighbor_al_mac;
+            interfacePairInfo.link_metric_info.intfType = media_type;
+            // TODO
+            //Indicates whether or not the 1905.1 link includes one or more IEEE 802.1 bridges
+            //eIEEE802_1BridgeFlag IEEE802_1BridgeFlag;
+            interfacePairInfo.link_metric_info.packet_errors =
+                link_metrics.transmitter.packet_errors;
+            interfacePairInfo.link_metric_info.transmitted_packets =
+                link_metrics.transmitter.transmitted_packets;
+            interfacePairInfo.link_metric_info.mac_throughput_capacity =
+                std::min(link_metrics.transmitter.mac_throughput_capacity_mbps,
+                         static_cast<uint32_t>(UINT16_MAX));
+            interfacePairInfo.link_metric_info.link_availability =
+                link_metrics.transmitter.link_availability;
+            interfacePairInfo.link_metric_info.phy_rate =
+                std::min(link_metrics.transmitter.phy_rate_mbps, static_cast<uint32_t>(UINT16_MAX));
+        }
+
+        /**
+         * Add Receiver Link Metric TLV if specifically requested or both requested
+         */
+        if ((ieee1905_1::eLinkMetricsType::RX_LINK_METRICS_ONLY == link_metrics_type) ||
+            (ieee1905_1::eLinkMetricsType::BOTH_TX_AND_RX_LINK_METRICS == link_metrics_type)) {
+            auto tlvReceiverLinkMetric = cmdu_tx.addClass<ieee1905_1::tlvReceiverLinkMetric>();
+            if (!tlvReceiverLinkMetric) {
+                LOG(ERROR) << "addClass ieee1905_1::tlvReceiverLinkMetric failed";
+                return false;
+            }
+
+            tlvReceiverLinkMetric->reporter_al_mac() = reporter_al_mac;
+            tlvReceiverLinkMetric->neighbor_al_mac() = neighbor_al_mac;
+
+            if (!tlvReceiverLinkMetric->alloc_interface_pair_info()) {
+                LOG(ERROR) << "alloc_interface_pair_info failed";
+                return false;
+            }
+            auto interface_pair_info = tlvReceiverLinkMetric->interface_pair_info(0);
+            if (!std::get<0>(interface_pair_info)) {
+                LOG(ERROR) << "Failed accessing interface_pair_info";
+                return false;
+            }
+            auto interfacePairInfo             = std::get<1>(interface_pair_info);
+            interfacePairInfo.rc_interface_mac = local_interface_mac;
+            // TODO: This is not correct... We actually have to get this from the topology
+            // discovery message, which will give us the neighbor interface MAC.
+            interfacePairInfo.neighbor_interface_mac         = neighbor_al_mac;
+            interfacePairInfo.link_metric_info.intfType      = media_type;
+            interfacePairInfo.link_metric_info.packet_errors = link_metrics.receiver.packet_errors;
+            interfacePairInfo.link_metric_info.packets_received =
+                link_metrics.receiver.packets_received;
+            interfacePairInfo.link_metric_info.rssi_db = link_metrics.receiver.rssi;
+        }
+    } else {
+        LOG(ERROR) << "Unable to get link link_metrics for interface " << local_interface_name
+                   << " and neighbor " << network_utils::mac_to_string(neighbor_al_mac);
+        return false;
+    }
+
+    return true;
+}
 
 //////////////////////////////////////////////////////////////////////////////
 /////////////////////////////// Implementation ///////////////////////////////
@@ -407,7 +631,7 @@ bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
             // Read the IP addresses of the bridge interface
             if (network_utils::get_iface_info(iface_info, strIface) != 0) {
                 LOG(ERROR) << "Failed reading addresses for: " << strIface;
-                return (false);
+                return false;
             }
 
             notification->params().gw_ipv4 = network_utils::ipv4_from_string(bridge_info.ip_gw);
@@ -880,7 +1104,7 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
     }
     default: {
         LOG(ERROR) << "Undefined state: " << int(m_eFSMState);
-        return (false);
+        return false;
     }
     }
 
@@ -1358,7 +1582,7 @@ bool backhaul_manager::backhaul_fsm_wireless(bool &skip_select)
     }
     default: {
         LOG(ERROR) << "backhaul_fsm_wireless() Invalid state: " << int(m_eFSMState);
-        return (false);
+        return false;
     }
     }
     return (true);
@@ -1837,6 +2061,9 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     case ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE: {
         return handle_1905_higher_layer_data_message(cmdu_rx, src_mac);
     }
+    case ieee1905_1::eMessageType::LINK_METRIC_QUERY_MESSAGE: {
+        return handle_1905_link_metric_query(cmdu_rx, src_mac);
+    }
     case ieee1905_1::eMessageType::COMBINED_INFRASTRUCTURE_METRICS_MESSAGE: {
         return handle_1905_combined_infrastructure_metrics(cmdu_rx, src_mac);
     }
@@ -2262,6 +2489,154 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
         return false;
     }
     LOG(DEBUG) << "sending ACK message to the originator, mid=" << std::hex << int(mid);
+    return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
+}
+
+bool backhaul_manager::handle_1905_link_metric_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                     const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received LINK_METRIC_QUERY_MESSAGE, mid=" << std::hex << int(mid);
+
+    /**
+     * The IEEE 1905.1 standard says about the Link Metric Query TLV and the neighbor type octet
+     * that "If the value is 0, then the EUI48 field is not present; if the value is 1, then the
+     * EUI-48 field shall be present."
+     *
+     * However, optional fields are not currently supported by TLVF.
+     *
+     * As a workaround, instead of defining a tlvLinkMetricQuery TLV with an optional field, we
+     * have defined two different TLVs: tlvLinkMetricQuery with the optional field and
+     * tlvLinkMetricQueryAllNeighbors without it. Application must check which of both TLVs has
+     * been received inside the message.
+     */
+    std::shared_ptr<ieee1905_1::tlvLinkMetricQueryAllNeighbors> tlvLinkMetricQueryAllNeighbors;
+    std::shared_ptr<ieee1905_1::tlvLinkMetricQuery> tlvLinkMetricQuery;
+
+    tlvLinkMetricQueryAllNeighbors = cmdu_rx.getClass<ieee1905_1::tlvLinkMetricQueryAllNeighbors>();
+    if (!tlvLinkMetricQueryAllNeighbors) {
+        tlvLinkMetricQuery = cmdu_rx.getClass<ieee1905_1::tlvLinkMetricQuery>();
+        if (!tlvLinkMetricQuery) {
+            LOG(ERROR) << "getClass ieee1905_1::tlvLinkMetricQueryAllNeighbors and "
+                          "ieee1905_1::tlvLinkMetricQuery failed";
+            return false;
+        }
+    }
+
+    /**
+     * 1905.1 AL MAC address of the device that transmits the response message.
+     */
+    sMacAddr reporter_al_mac = network_utils::mac_from_string(bridge_info.mac);
+
+    /**
+     * 1905.1 AL MAC address of a neighbor of the receiving device.
+     * Query can specify a particular neighbor device or all neighbor devices.
+     */
+    sMacAddr neighbor_al_mac = network_utils::ZERO_MAC;
+
+    /**
+     * Obtain link metrics for either all neighbors or a specific neighbor
+     */
+    ieee1905_1::eLinkMetricNeighborType neighbor_type;
+
+    /**
+     * The link metrics type requested: TX, RX or both
+     */
+    ieee1905_1::eLinkMetricsType link_metrics_type;
+
+    if (tlvLinkMetricQuery) {
+        /**
+    	   * If tlvLinkMetricQuery has been included in message, we will be permissive enough to
+    	   * allow it specify ALL_NEIGHBORS and if so, then we will just ignore the field
+    	   * containing the MAC address of neighbor.
+    	   */
+        neighbor_type     = tlvLinkMetricQuery->neighbor_type();
+        neighbor_al_mac   = tlvLinkMetricQuery->mac_al_1905_device();
+        link_metrics_type = tlvLinkMetricQuery->link_metrics_type();
+    } else {
+        neighbor_type = tlvLinkMetricQueryAllNeighbors->neighbor_type();
+        if (ieee1905_1::eLinkMetricNeighborType::ALL_NEIGHBORS != neighbor_type) {
+            LOG(ERROR) << "Unexpected neighbor type: " << std::hex << int(neighbor_type);
+            return false;
+        }
+        link_metrics_type = tlvLinkMetricQueryAllNeighbors->link_metrics_type();
+    }
+
+    /**
+     * Set alias flag to true if link metrics for a specific neighbor have been requested
+     */
+    bool specific_neighbor =
+        ieee1905_1::eLinkMetricNeighborType::SPECIFIC_NEIGHBOR == neighbor_type;
+
+    /**
+     * Create response message
+     */
+    auto cmdu_tx_header =
+        cmdu_tx.create(mid, ieee1905_1::eMessageType::LINK_METRIC_RESPONSE_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "Failed creating LINK_METRIC_RESPONSE_MESSAGE header! mid=" << std::hex
+                   << (int)mid;
+        return false;
+    }
+
+    /**
+     * Get the list of neighbor links from the topology database.
+     * Neighbor links are grouped by the interface that connects to them.
+     */
+    std::map<std::string, std::vector<sMacAddr>> neighbor_links_map;
+    if (!get_neighbor_links(neighbor_al_mac, neighbor_links_map)) {
+        LOG(ERROR) << "Failed to get the list of neighbor links";
+        return false;
+    }
+
+    /**
+     * If the specified neighbor 1905.1 AL ID does not identify a neighbor of the receiving 1905.1
+     * AL, then a link metric ResultCode TLV (see Table 6-21) with a value set to “invalid
+     * neighbor” shall be included in this message.
+     */
+    bool invalid_neighbor = specific_neighbor && neighbor_links_map.empty();
+    if (invalid_neighbor) {
+        auto tlvLinkMetricResultCode = cmdu_tx.addClass<ieee1905_1::tlvLinkMetricResultCode>();
+        if (!tlvLinkMetricResultCode) {
+            LOG(ERROR) << "addClass ieee1905_1::tlvLinkMetricResultCode failed, mid=" << std::hex
+                       << (int)mid;
+            return false;
+        }
+
+        tlvLinkMetricResultCode->value() = ieee1905_1::tlvLinkMetricResultCode::INVALID_NEIGHBOR;
+    } else {
+
+        /**
+         * Report link metrics for the link with specific neighbor or for all neighbors, as
+         * obtained from topology database
+         */
+        for (const auto &entry : neighbor_links_map) {
+            std::string local_interface_name = entry.first;
+            std::vector<sMacAddr> neighbors  = entry.second;
+
+            ieee1905_1::eMediaTypeGroup media_type_group = ieee1905_1::eMediaTypeGroup::UNKNOWN;
+            if (!get_media_type_group(local_interface_name, media_type_group)) {
+                LOG(ERROR) << "Unable to compute media type group for interface "
+                           << local_interface_name;
+                return false;
+            }
+
+            ieee1905_1::eMediaType media_type = ieee1905_1::eMediaType::UNKNONWN_MEDIA;
+            if (!get_media_type(local_interface_name, media_type_group, media_type)) {
+                LOG(ERROR) << "Unable to compute media type for interface " << local_interface_name;
+                return false;
+            }
+
+            for (const auto &neighbor : neighbors) {
+                if (!add_link_metrics(local_interface_name, media_type, reporter_al_mac, neighbor,
+                                      link_metrics_type, cmdu_tx)) {
+                    return false;
+                }
+            }
+        }
+    }
+
+    LOG(DEBUG) << "Sending LINK_METRIC_RESPONSE_MESSAGE, mid: " << std::hex << (int)mid;
     return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
 }
 
@@ -2932,6 +3307,23 @@ std::shared_ptr<bwl::sta_wlan_hal> backhaul_manager::get_wireless_hal(std::strin
     }
 
     return slave_sk->second->sta_wlan_hal;
+}
+
+bool backhaul_manager::get_neighbor_links(
+    const sMacAddr &neighbor_mac_filter,
+    std::map<std::string, std::vector<sMacAddr>> &neighbor_links_map)
+{
+
+    // TODO: Topology Database is required to implement this method.
+    // Since we don't have it yet, method will return a single neighbor, the controller (which is
+    // BTW incorrect for the second repeater if we have a chain topology, because then the
+    // controller is not a neighbor).
+    sMacAddr neighbor = network_utils::mac_from_string(controller_bridge_mac);
+    if ((neighbor_mac_filter == network_utils::ZERO_MAC) || (neighbor_mac_filter == neighbor)) {
+        neighbor_links_map[m_sConfig.wire_iface].push_back(neighbor);
+    }
+
+    return true;
 }
 
 } // namespace beerocks
--- a/agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
+++ b/agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
@@ -1,0 +1,277 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "ieee802_3_link_metrics_collector.h"
+
+#include <bcl/network/network_utils.h>
+
+#include <easylogging++.h>
+
+#include <errno.h>
+#include <linux/netlink.h>
+#include <linux/rtnetlink.h>
+#include <net/if.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <sys/socket.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+// SPEED values
+#include <linux/ethtool.h>
+
+#define IFLIST_REPLY_BUFFER 8192
+
+namespace beerocks {
+
+/**
+ * @brief Netlink request type.
+ */
+struct nl_req_t {
+    /**
+     * Netlink message
+     */
+    struct nlmsghdr hdr;
+
+    /**
+     * "general form of address family dependent" message, i.e. how to tell which AF we are
+     * interested in. */
+    struct rtgenmsg gen;
+};
+
+/**
+ * @brief Gets link metrics of an Ethernet network interface.
+ *
+ * Gets link metrics for given Ethernet network interface from the specified Netlink message of
+ * type RTM_NEWLINK.
+ *
+ * @param[in] h Pointer to the Netlink message containing the data.
+ * @param[in] local_interface_name Name of the Ethernet network interface.
+ * @param[in, out] link_metris Link metrics structure with read values.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_link_metrics(const struct nlmsghdr *h, const std::string &local_interface_name,
+                             sLinkMetrics &link_metrics)
+{
+    bool result = false;
+
+    struct ifinfomsg *iface = static_cast<ifinfomsg *>(NLMSG_DATA(h));
+
+    size_t length = 0;
+    if (h->nlmsg_len > NLMSG_LENGTH(sizeof(*iface))) {
+        length = h->nlmsg_len - NLMSG_LENGTH(sizeof(*iface));
+    }
+
+    /**
+     * Loop over all attributes of the RTM_NEWLINK message
+     */
+    for (struct rtattr *attribute = IFLA_RTA(iface); RTA_OK(attribute, length);
+         attribute                = RTA_NEXT(attribute, length)) {
+        switch (attribute->rta_type) {
+        case IFLA_IFNAME:
+            /**
+             * This message contains the stats for the interface we are interested in
+             */
+            if (0 == std::strncmp(local_interface_name.c_str(), (char *)RTA_DATA(attribute),
+                                  local_interface_name.length() + 1)) {
+                result = true;
+            }
+            break;
+        case IFLA_STATS:
+            if (result) {
+                struct rtnl_link_stats *stats = (struct rtnl_link_stats *)RTA_DATA(attribute);
+
+                /**
+                 * Get interface speed into PHY rate.
+                 */
+                uint32_t phy_rate_mbps = UINT32_MAX;
+                beerocks::net::network_utils::linux_iface_get_speed(local_interface_name,
+                                                                    phy_rate_mbps);
+
+                link_metrics.transmitter.packet_errors       = stats->tx_errors;
+                link_metrics.transmitter.transmitted_packets = stats->tx_packets;
+                /**
+                 * Note: The MAC throughput capacity is a function of the physical data rate and
+                 * of the MAC overhead. We could somehow compute such overhead or, for simplicity,
+                 * set the MAC throughput as a percentage of the physical data rate.
+                 * For Ethernet, we can estimate the overhead: 7 bytes preamble, 1 byte SFD, 14
+                 * bytes header, 4 bytes CRC and 12 bytes of interpacket gap on a 1500 byte
+                 * payload. So 1500/1538.
+                 * (see https://en.wikipedia.org/wiki/Ethernet_frame)
+                 */
+                const float layer2_payload_size = 1500;
+                const float layer1_total_size   = 1538;
+                link_metrics.transmitter.mac_throughput_capacity_mbps =
+                    phy_rate_mbps * (layer2_payload_size / layer1_total_size);
+                // Note: For simplicity, link availability is set to "100% of the time"
+                link_metrics.transmitter.link_availability = 100;
+                link_metrics.transmitter.phy_rate_mbps     = phy_rate_mbps;
+
+                link_metrics.receiver.packet_errors    = stats->rx_errors;
+                link_metrics.receiver.packets_received = stats->rx_packets;
+                link_metrics.receiver.rssi             = UINT8_MAX;
+            }
+            break;
+        }
+    }
+
+    return result;
+}
+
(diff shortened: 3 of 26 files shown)


agent/src/beerocks/slave/link_metrics/link_metrics.h
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Estimated number of packets transmitted by the Transmitter of the link on the same
+     * measurement period used to estimate tx_packet_errors.
+     */
+    uint32_t transmitted_packets;
+
+    /**
+     * The maximum MAC throughput of the link estimated at the transmitter and expressed in Mb/s.
+     */
+    uint16_t mac_throughput_capacity;
```
arnout: It is better to put the unit explicitly in the name, i.e. `mac_throughput_capacity_mbps`.

Also, is it a good idea to limit to 64Gbps? 10Gbps links are a reality in commercial switches, so 64Gbps is not that far off... The spec indeed puts it as a uint16_t, but I think it's better to use a generic unsigned or uint32_t here and do saturation when converting to the TLV.

agent/src/beerocks/slave/link_metrics/link_metrics.h
```diff
@@ -0,0 +1,123 @@
...
+     * The estimated average percentage of time that the link is available for data transmissions.
+     */
+    uint16_t link_availability;
+
+    /**
+     * If the media type of the link is IEEE 802.3, then IEEE 1901 or MoCA 1.1, then this value is
+     * the PHY rate estimated at the transmitter of the link expressed in Mb/s; otherwise, it is
+     * set to 0xFFFF.
+     */
+    uint16_t phy_rate;
```
arnout: Same question about the uint16_t.

agent/src/beerocks/slave/link_metrics/link_metrics.h
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param[in] local_interface_name Name of the local interface.
+     * @param[in] neighbor_interface_address MAC address at the other end of the link (this MAC
+     * address belongs to a neighbor's interface.
+     * @param[out] link_metrics Link metrics information.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_link_metrics(const std::string &local_interface_name,
+                                  const sMacAddr &neighbor_interface_address,
```
arnout, line 117: I think in practice it will never be possible to distinguish between different neighbours that are connected on the same interface. This is a serious shortcoming of the 1905.1 spec, but not much we can do about it... @mariomaz do you think you can ask your colleagues who are more familiar with 1905.1 how you are supposed to deal with interfaces where there are multiple neighbours, or none?

Anyway, my point is: I think in practice we will only use the local_interface_name.

(nitpick)

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+    req.hdr.nlmsg_seq    = 1;
+    req.hdr.nlmsg_pid    = 0;
+    req.gen.rtgen_family = AF_PACKET; /*  no preferred AF, we will get *all* interfaces */
+
+    io.iov_base          = &req;
+    io.iov_len           = req.hdr.nlmsg_len;
+    rtnl_msg.msg_iov     = &io;
+    rtnl_msg.msg_iovlen  = 1;
+    rtnl_msg.msg_name    = &kernel;
+    rtnl_msg.msg_namelen = sizeof(kernel);
```
arnout: Is there any reason to open-code all this, rather than using libnl3 genl? In fact, we could move base_wlan_hal_nl80211::send_nl80211_msg to network_utils and use that function. The nl80211 part of it is only hardcoded in the genl_ctrl_resolve call.

Anyway, something that can be done later.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+     * Loop over all attributes of the NEWLINK message
+     */
+    for (struct rtattr *attribute = IFLA_RTA(iface); RTA_OK(attribute, length);
+         attribute                = RTA_NEXT(attribute, length)) {
+        switch (attribute->rta_type) {
+        case IFLA_IFNAME:
+            /**
+             * This message contains the stats for the interface we are interested in
+             */
+            if (0 == strcmp(local_interface_name.c_str(), (char *)RTA_DATA(attribute))) {
```
arnout: I believe there is a way to specify this already in the GET message, so instead of getting all interfaces, you get just one.

You would need to convert it to ifindex first though, IIRC.

(nitpick).

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+     */
+    struct nlmsghdr hdr;
+
+    /**
+     * "general form of address family dependent" message, i.e. how to tell which AF we are
+     * interested in. */
+    struct rtgenmsg gen;
+};
+
+static bool get_link_metrics(const struct nlmsghdr *h, const std::string &local_interface_name,
```
arnout, line 61: I think this function should get a slightly more specific name.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+                link_metrics.transmitter.packet_errors       = stats->tx_errors;
+                link_metrics.transmitter.transmitted_packets = stats->tx_packets;
+                /**
+                 * Note: The MAC throughput capacity is a function of the physical data rate and
+                 * of the MAC overhead. We could somehow compute such overhead or, for simplicity,
+                 * set the MAC throughput as a percentage of the physical data rate. To make
+                 * things even simpler, we will set the MAC throughput capacity with the same
+                 * value as the physical data rate, as if there was no overhead at all.
+                 */
+                link_metrics.transmitter.mac_throughput_capacity = phy_rate;
```
arnout: For Ethernet, we do know the overhead, or at least we can estimate it: 14 bytes header and 4 bytes CRC plus 12 bytes of interframe gap on a 1500 byte payload. So 1500/1530. But again, can be done later.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+                 */
+                uint16_t phy_rate = UINT16_MAX;
+                uint32_t speed    = 0;
+                if (beerocks::net::network_utils::linux_iface_get_speed(local_interface_name,
+                                                                        speed)) {
+                    if (SPEED_100 == speed) {
+                        phy_rate = 100;
+                    } else if (SPEED_1000 == speed) {
+                        phy_rate = 1000;
+                    }
```
arnout: Here, there really is no reason at all to limit to specific speeds. We can simply use the speed value directly. The only thing we have to do is to saturate at 64Gbps, but that's something which we should do while converting to TLV (as mentioned below).

In other words, I think we can pass phy_rate directly into linux_iface_get_speed.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+                }
+
+                link_metrics.transmitter.packet_errors       = stats->tx_errors;
```
arnout, line 99: I believe we have to do this right...

All the stats fields are uint32_t and wrap. tx_errors probably won't wrap, but the number of rx/tx packets will wrap if the device stays up for a couple of days.

Therefore, we have to keep the previously measured value and subtract it (wrapping subtraction, I'm not entirely sure what the standard C++ way of doing that is).

While we're at it, we should probably also keep the timestamp. In 1905.1 the time of the measurement isn't specified (which renders it completely useless), but in R2 there is a timestamp included. And to make the metrics somewhat useful even in R1, we should probably divide by the sampling time so the numbers are at least comparable with each other.

So, I think it would be better to keep an object for the link metrics for each interface.

That also means you can keep the netlink fd open. Minor optimisation.

framework/tlvf/yaml/tlvf/ieee_1905_1/eLinkMetricNeighborType.yaml
```diff
@@ -0,0 +1,10 @@
+#
+---
+_namespace: ieee1905_1
+
+eLinkMetricNeighborType:
+  _type: enum
+  _enum_storage: uint8_t
+  ALL_NEIGHBORS: 0x00
+  SPECIFIC_NEIGHBOR: 0x01
```
arnout, line 9: Moving these to a separate file wasn't really needed, but OK.

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvLinkMetricQuery.yaml
```diff
@@ -2,29 +2,30 @@
...
+
+tlvLinkMetricQueryAllNeighbors:
   _type: class
   _is_tlv_class : True
   type:
     _type: eTlvType
     _value_const: TLV_LINK_METRIC_QUERY
   length: uint16_t
-  neighbor_type: eNeighborType
+  neighbor_type: eLinkMetricNeighborType
```
arnout: You could make this
```
  neighbor_type:
    _type: eLinkMetricNeighborType
    _value_const: ALL_NEIGHBORS
```

Note that you can't do that for the other one.

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvLinkMetricQuery.yaml
```diff
@@ -2,29 +2,30 @@
...
+  link_metrics_type: eLinkMetricsType
+
+tlvLinkMetricQuerySpecificNeighbor:
```
arnout: This is not named correctly. It will also be used for the All Neighbors case for devices that do include a 00:00:00:00:00:00 MAC address.

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvLinkMetricQuery.yaml
```diff
@@ -2,29 +2,30 @@
 ---
 _namespace: ieee1905_1
 
-tlvLinkMetricQuery:
+# TLVF does not currently support optional fields and this TLV contains one.
+# As a workaround, we define two different TLVs instead of a single tlvLinkMetricQuery with an
+# optional field. Application must then check the length of received message to know if optional 
+# field is present or not and then create an instance of either of these classes.
+
+tlvLinkMetricQueryAllNeighbors:
```
arnout, line 10: Since the tlvLinkMetricQuery class is removed here, we also have to remove all instances of it in the same commit. It is used in the parse function in CmduMessageRx and in tlvf_test.

To split up the commits a bit more, you could first simply add the two new ones but keep the old one around, then update the parse function and tlvf_test, and then remove the old one.

arnout (commented):
Reviewed up to "split link metric query TLV into two TLVs".

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -74,7 +74,30 @@ std::shared_ptr<BaseClass> CmduMessageRx::parseNextTlv()
...
+         */
+        const uint16_t min_tlv_length = 2;
+        uint16_t tlv_length           = getNextTlvLength();
+
+        if ((min_tlv_length + sizeof(sMacAddr)) == tlv_length) {
+            return msg.addClass<tlvLinkMetricQuerySpecificNeighbor>();
+        } else if (min_tlv_length == tlv_length) {
+            return msg.addClass<tlvLinkMetricQueryAllNeighbors>();
+        } else {
+            return msg.addClass<tlvUnknown>();
```
arnout: Maybe there should be a warning here.

framework/tlvf/src/src/CmduMessage.cpp
```diff
@@ -32,7 +32,11 @@ uint16_t CmduMessage::getNextTlvLength() const
     if (!getCmduHeader())
         return -1;
     sTlvHeader *tlv = reinterpret_cast<sTlvHeader *>(msg.prevClass()->getBuffPtr());
-    return tlv->length;
+
+    uint16_t tlv_length = tlv->length;
+    swap_16(tlv_length);
```
arnout: This swapping is only correct in the context where you call it, i.e. in CmduMessageRx::parse(). In the one case where it is currently used, in a debug message in tlvf_test, it actually should not be swapped, since it's already swapped by the parse() that precedes it.

Therefore, I propose to move this function down into CmduMessageRx first and make it private.

And while you're at it, remove getNextTlvData and getNextTlvType(eTlvType), and move getNextTlvType() down into CmduMessageRx as well. But maybe that's a bit too much for this PR :-).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2069,6 +2267,186 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
...
+    /**
+     * The link metrics type requested: TX, RX or both
+     */
+    ieee1905_1::eLinkMetricsType link_metrics_type;
+
+    if (tlvLinkMetricQuerySpecificNeighbor) {
+        /**
+    	   * If tlvLinkMetricQuerySpecificNeighbor has been included in message, we will be
+    	   * permissive enough to allow it specify ALL_NEIGHBORS and if so, then we will just
+    	   * ignore the field containing the MAC address of neighbor.
```
arnout: :1st_place_medal:

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2069,6 +2267,186 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
...
+    /**
+     * If the specified neighbor 1905.1 AL ID does not identify a neighbor of the receiving 1905.1
+     * AL, then a link metric ResultCode TLV (see Table 6-21) with a value set to “invalid
+     * neighbor” shall be included in this message.
+     */
+    bool invalid_neighbor = false;
+    if (ieee1905_1::eLinkMetricNeighborType::SPECIFIC_NEIGHBOR == neighbor_type) {
+
+        // TODO: check if specified neighbor is invalid (topology database is required for that)
+    }
```
arnout: Nice preparation for the future! I like it!

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2069,6 +2267,186 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
...
+    if (invalid_neighbor) {
+        auto tlvLinkMetricResultCode = cmdu_tx.addClass<ieee1905_1::tlvLinkMetricResultCode>();
+        if (!tlvLinkMetricResultCode) {
+            LOG(ERROR) << "addClass ieee1905_1::tlvLinkMetricResultCode failed, mid=" << std::hex
+                       << (int)mid;
+            return false;
+        }
+
+        tlvLinkMetricResultCode->value() = ieee1905_1::tlvLinkMetricResultCode::INVALID_NEIGHBOR;
+    } else if (ieee1905_1::eLinkMetricNeighborType::SPECIFIC_NEIGHBOR == neighbor_type) {
```
arnout: (nitpick) I think it would be nicer to have a boolean "specific_neighbor" that is set above where we parse the two message types, so we can simply do `if (specific_neighbor)` here.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2069,6 +2267,186 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
...
+
+        } else {
+            //TODO this is a valid neighbor so it must be connected through a wireless interface, which is not yet supported
+        }
+
+        /**
+         * Obtain link metrics and add a Transmitter Link Metric TLV or a Receiver Link Metric TLV
+         * or both to response message.
+         */
+        if (!add_link_metrics(local_interface_name, media_type, reporter_al_mac, neighbor_al_mac,
```
arnout: I think we have to do it differently than the standard here. I don't think we'll ever support more than one neighbor on an interface. So somehow there's a one-to-one mapping between neighbors and interfaces.

So, I think we should get from the topology manager a list of (neighbor, interface) pairs, and then call add_link_metrics on it. The two al_mac addresses are there only to copy into the TLVs. In the specific neighbor case, the list is filtered to find the single instance that was requested - which also covers the invalid neighbor case.

I believe this approach will simplify things somewhat.

Concretely, since we don't have a topology manager yet, I would add a function that returns a list with a single (neighbor, interface) pair for the controller (which is BTW incorrect for the second repeater if we have a chain topology, because then the controller is not a neighbor). So that replaces the logic above.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2069,6 +2267,186 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
...
+             * If specified neighbor is the controller and it is connected through the wired backhaul
+             * interface, then interface is the wired backhaul interface.
+             */
+            local_interface_name = m_sConfig.wire_iface;
+
+            /**
+             * Compute particular Ethernet media type
+             */
+            if (!get_media_type(local_interface_name, ieee1905_1::eMediaTypeGroup::IEEE_802_3,
+                                media_type)) {
```
arnout: I don't think it makes sense to do get_media_type at this level, I'd rather do it inside a dispatcher function in the link metrics collector.

Or better yet, for each interface I would like to have an actual object instead of just a name, and that object would have a reference to the collector (and the media type). But maybe that's going a bit too far for this PR...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -74,6 +77,198 @@ namespace beerocks {
...
+                return false;
+            }
+            auto interface_pair_info = tlvTransmitterLinkMetric->interface_pair_info(0);
+            if (!std::get<0>(interface_pair_info)) {
+                LOG(ERROR) << "Failed accessing interface_pair_info";
+                return false;
+            }
+            auto interfacePairInfo                      = std::get<1>(interface_pair_info);
+            interfacePairInfo.rc_interface_mac          = local_interface_mac;
+            interfacePairInfo.neighbor_interface_mac    = neighbor_al_mac;
```
arnout, line 244: Hm... This is not correct... We actually have to get this from the topology discovery message, which will give us the neighbor interface MAC.

So for now, add a TODO.

agent/src/beerocks/slave/link_metrics/link_metrics.h
mariomaz: Done

agent/src/beerocks/slave/link_metrics/link_metrics.h
mariomaz: Done

mariomaz:
I've also replaced strcmp() banned function with std::strncmp()

agent/src/beerocks/slave/link_metrics/link_metrics.h
mariomaz: I agree with you.
I've checked our 1905.1 implementation at this point and found this TODO which also confirms your opinion:
```c
            //   TODO: In Linux there is no easy way to obtain this
            //   information. We will just report the amount of *total* packets
            //   transmitted from our local interface, no matter the
            //   destination.
            //   This information will only be correct when the local interface
            //   is connected to one single remote interface... however we
            //   better report this than nothing at all.
```
However, be aware that this is an interface which will be implemented by different collectors, not only the 802.3 collector. In the 802.11 collector it is possible to distinguish between different neighbors and thus we do need the neighbor_interface_address parameter.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: I agree with you. 

This is something that could probably be done using changes in "Feature/add support for nl80211 standard commands to BWL #900". However, be aware that such changes have been made into BWL and not in network_utils. 

Anyway, I agree with you that this is a refactoring that could be done later (I think it is not going to be so easy and quick as might seem at first glance).

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Since this is a nitpick comment, the change could probably be done as part of the previous refactoring.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: I'm not sure about what you mean by "more specific". If you please suggest a better name I will change it. However, be aware that this code will be refactored and this function will disappear.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Done

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Done

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Sorry but I cannot see the timestamp field you mentioned. Maybe I'm not looking at the right document.  Can you please clarify which are those R1 and R2 docs?
I agree with you that if we cannot include the time measurement window along with the counter, then it is useless to provide the number of packets/errors **during** such measurement period. 

I think however that, even if we could provide timestamp or elapsed time since last measure, we have no way to know if the number of packets we read with the netlink socket have wrapped (and how many times) since last call to get link metrics. Although such wrapping is quite improbable to go unnoticed, given the frequency with which calls to get link metrics will be probably issued, we cannot guarantee that wrapping won't happen. Keeping timestamps is useless if they are not provided by the kernel.

framework/tlvf/yaml/tlvf/ieee_1905_1/eLinkMetricNeighborType.yaml
mariomaz: If you declare eLinkMetricNeighborType in the same YAML file as the two TLV classes that use it then, as TLVF defines the eLinkMetricNeighborType **inside** one of the TLV C++ classes, you get a compilation error because it's undefined for the other TLV class.

Maybe there is some other workaround for this but I found moving eLinkMetricNeighborType to a different file the easiest.

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvLinkMetricQuery.yaml
mariomaz: Done

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvLinkMetricQuery.yaml
mariomaz: Renamed tlvLinkMetricQuerySpecificNeighbor with its original name tlvLinkMetricQuery because it's suitable for both use cases: all-neighbors and specific-neighbor

framework/tlvf/yaml/tlvf/ieee_1905_1/tlvLinkMetricQuery.yaml
mariomaz: With previous renaming, tlvLinkMetricQuery is being used back again.

agent/src/beerocks/slave/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
 
 file(GLOB beerocks_agent_sources 
     ${MODULE_PATH}/backhaul_manager/*.c*
+    ${MODULE_PATH}/link_metrics/*.c*
```
tomereli, line 11: This is classic BPL implementation, so I would prefer it in the BPL.
But since you say below that this is a temporary solution, and this code should be refactored pretty soon so its part of the discovery service - (nitpick)

However, please add a comment here that this should move to the discovery agent and add a link to the github relevant task.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
tomereli: strcmp is baned C function if I'm not mistaken.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
tomereli: Can you elaborate on the fact that this will disappear? You mean it will be part of the discovery service? In that case it is worth to mention that in the commit message.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+    struct rtgenmsg gen;
+};
+
+static bool get_link_metrics(const struct nlmsghdr *h, const std::string &local_interface_name,
+                             sLinkMetrics &link_metrics)
+{
+    bool result = false;
+
+    struct ifinfomsg *iface = static_cast<ifinfomsg *>(NLMSG_DATA(h));
+    int length              = h->nlmsg_len - NLMSG_LENGTH(sizeof(*iface));
```
tomereli: Any reason not to use `size_t`? Can `length` be negative?

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
tomereli: So it appears I was mistaken and apparently the list of banned C functions by Intel SDL is listed in github:
https://github.com/intel/safestringlib/wiki/SDL-List-of-Banned-Functions
We should add this to our CONTRIBUTING.md, perhaps @vitalii-komisarenko @abelog or @VladyslavTupikin can do that as good quick first PR.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * Loop over all attributes of the NEWLINK message
+     */
+    for (struct rtattr *attribute = IFLA_RTA(iface); RTA_OK(attribute, length);
+         attribute                = RTA_NEXT(attribute, length)) {
```
tomereli, line 77: Extra space

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+    struct sockaddr_nl kernel; /* the remote (kernel space) side of the communication */
+
+    struct msghdr rtnl_msg; /* generic msghdr struct for use with sendmsg */
+    struct iovec io;        /* IO vector for sendmsg */
+    struct nl_req_t req;    /* structure that describes the Netlink packet itself */
+
+    /**
+     * Netlink socket is ready for use, prepare and send request
+     */
+    memset(&rtnl_msg, 0, sizeof(rtnl_msg));
```
tomereli: Although `memset` is not banned, it is recommended to use `memset_s` or an equivalent c++ STD function such as `std::fill`. See https://github.com/intel/safestringlib/wiki/SDL-List-of-Banned-Functions

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
tomereli: Why not indeed move `base_wlan_hal_nl80211::send_nl80211_msg` to BCL? just that function, shouldn't be too hard even as a followup commit in this PR. I really prefer that we don't duplicate code when we really don't need to. Still, as you say, it can be done later, so (nitpoick)

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,246 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    return result;
+}
+
+static bool get_link_metrics(const std::string &local_interface_name, sLinkMetrics &link_metrics)
```
tomereli, line 241: I see now this is the 4th function overload with the same `get_link_metrics` name. This is fine, but can we at least add doxygen documentation to each function?

tomereli (commented):
Partial review only for now, so just comments

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Method base_wlan_hal_nl80211::send_nl80211_msg() is going to disappear after we had approved the design and merged "Feature/add support for nl80211 standard commands to BWL #900" and done a refactoring of the NL80211 version of the BWL. 

The work to be done is then to implement the get_link_metrics() method using the new classes defined in #900.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Although strcmp is not in the list of banned C functions, I've replaced it with std::strncmp() C++ function which is safer.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: This function will disappear once we do a refactor of this class to use Netlink classes defined in #900 to get link metrics.

framework/tlvf/src/src/CmduMessageRx.cpp
mariomaz: Since now we have tlvLinkMetricQuery back again, I've changed it to use tlvLinkMetricQuery when the length of the TLV does not match the length of tlvLinkMetricQueryAllNeighbors.
tlvUnknown is now used by the default switch clause only.

framework/tlvf/src/src/CmduMessage.cpp
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Very good point. 
I've included your suggestions to simplify code. However, one-to-many mapping between interfaces and neighbors is possible in WiFi so I've kept that possibility also for Ethernet (because it's simpler to code, doesn't hurt and who knows if someday someone connects more than one IEEE 1905 device to a switch).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: get_media_type() is an expensive function in terms of performance. If we take it out from where it is now, then we'll end up calling it more than once for the same interface.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

agent/src/beerocks/slave/CMakeLists.txt
mariomaz: Done.
I've created "[TASK] Move link metric related classes to BPL #910"
However, link metric collection is not meant to be part of the discovery service but it requires the topology database that will be populated by the discovery service.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: I've replaced strcmp with std::strncmp.
When I say that get_link_metrics() will disappear I mean that it's going to be refactored and coded in a different way.
I've changed the type of `length` from `int` to `size_t` and checked the values before subtracting.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: clang-format.sh is to blame

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: I've set the initial value of structs to zero on declaration and removed all calls to `memset`. 
Looks like `memset_s` is not available.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Link metrics cannot be obtained with `base_wlan_hal_nl80211::send_nl80211_msg` as is. See https://github.com/prplfoundation/prplMesh/pull/831#discussion_r386899613

When you say BCL, I guess you mean BPL. Be aware that in #900 I've implemented the classes to send NL80211 commands and all Netlink related stuff in BWL. First of all we should decide  where those classes should be. 

I've documented the `get_link_metrics` functions.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
arnout: > Sorry but I cannot see the timestamp field you mentioned.

I should have written "we should keep *a* timestamp".

> we have no way to know if the number of packets we read with the netlink socket have wrapped (and how many times) since last call to get link metrics.

Stupid me again, I thought these were 64-bit numbers, but they're only 64-bit if you get them through netlink, not if you get them from IFLA_STATS64. Which apparently requires special handling because they're unaligned...

Which brings me to the question: why don't we use rtnl for this? E.g. https://gist.github.com/beejeebus/cc4fb07472cf5a0afd41 rtnl_link takes care of the 64-bit handling and fallback to 32-bit if the 64-bit stats are not available.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
arnout: Regardless of the new classes, the question remains why you would do all the parsing manually rather than use one of the rtnl_link_get_ functions to get all info in a nice struct.

framework/tlvf/yaml/tlvf/ieee_1905_1/eLinkMetricNeighborType.yaml
arnout: I didn't realize that. OK.

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -51,6 +51,26 @@
...
+    if (!getCmduHeader())
+        return -1;
+    sTlvHeader *tlv = reinterpret_cast<sTlvHeader *>(msg.prevClass()->getBuffPtr());
+    return tlv->type;
+}
+
+uint16_t CmduMessageRx::getNextTlvLength() const
+{
+    if (!getCmduHeader())
+        return -1;
```
arnout: This would need an explicit cast to uint16_t IMO.

Or maybe even better: make the return value int.

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -96,7 +96,27 @@ std::shared_ptr<BaseClass> CmduMessageRx::parseNextTlv()
...
+         * octet that "If the value is 0, then the EUI48 field is not present; if the value is 1,
+         * then the EUI-48 field shall be present."
+         *
+         * However, optional fields are not currently supported by TLVF.
+         *
+         * As a workaround, instead of defining a tlvLinkMetricQuery TLV with an optional field,
+         * we have defined two different TLVs, one with the optional field and the other one
+         * without it. Application must then check the length of received TLV to know if optional
+         * field (MAC address of neighbor device) is present or not and then create an instance of
+         * either tlvLinkMetricQuery or tlvLinkMetricQueryAllNeighbors respectively.
```
arnout, line 111: I don't know if I said it before, but this is really an excellent explanation.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2142,6 +2364,154 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
...
+            return false;
+        }
+        link_metrics_type = tlvLinkMetricQueryAllNeighbors->link_metrics_type();
+    }
+
+    /**
+     * Set alias flag to true if link metrics for a specific neighbor have been requested
+     */
+    bool specific_neighbor =
+        ieee1905_1::eLinkMetricNeighborType::SPECIFIC_NEIGHBOR == neighbor_type;
```
arnout, line 2574: (nitpick) I would set this directly in the condition above instead of saving neighbor_type.

tests/test_flows.sh
```diff
@@ -200,6 +201,23 @@ test_client_capability_query() {
...
+    check send_CAPI_1905 ${GATEWAY} $mac_agent1 0x0005 "tlv_type,0x08,tlv_length,0x0002,tlv_value,0x00 0x02"
+    sleep 1
+    
+    dbg "Confirming link metric query has been received on agent"
+    check_log ${REPEATER1} agent "Received LINK_METRIC_QUERY_MESSAGE"
+    
+    dbg "Confirming link metric response has been received on controller"
+    check_log ${GATEWAY} controller "Received LINK_METRIC_RESPONSE_MESSAGE"
+    check_log ${GATEWAY} controller "Received TLV_TRANSMITTER_LINK_METRIC"
+    check_log ${GATEWAY} controller "Received TLV_RECEIVER_LINK_METRIC"
```
arnout: This should of course be converted to Python.

arnout (approved):
Except for doing the test in python instead of shell, all my comments are nitpicky.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
arnout: Just to be clear: since the code is there now anyway, I would merge it as is. But when later we refactor with #900, I think we should also look into using rtnl_link instead of doing all the parsing manually. However, before going into that, we should check if the libnl-tiny used in openwrt also has it.

framework/tlvf/src/src/CmduMessageRx.cpp
mariomaz: I've changed it to return UINT16_MAX

tests/test_flows.sh
mariomaz: Done

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+                /**
+                 * Note: The MAC throughput capacity is a function of the physical data rate and
+                 * of the MAC overhead. We could somehow compute such overhead or, for simplicity,
+                 * set the MAC throughput as a percentage of the physical data rate.
+                 * For Ethernet, we can estimate the overhead: 7 bytes preamble, 1 byte SFD, 14
+                 * bytes header, 4 bytes CRC and 12 bytes of interpacket gap on a 1500 byte
+                 * payload. So 1500/1538.
+                 * (see https://en.wikipedia.org/wiki/Ethernet_frame)
+                 */
+                const uint32_t layer2_payload_size = 1500;
```
tomereli: Why not define as `float` and get rid of the casting below?
```suggestion
                const float layer2_payload_size = 1500;
```
(nitpick)

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+                const uint32_t layer1_total_size   = 1538;
+                link_metrics.transmitter.mac_throughput_capacity_mbps =
+                    phy_rate_mbps * ((float)layer2_payload_size / (float)layer1_total_size);
```
tomereli: C style casting should be avoided.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
```
tomereli, line 1: It took me some time to understand that this is only for ethernet links. I guess I should know by heart that 802.3 is Ethernet, but consider changing the file name to `eth_link_metrics_collector.cpp` (nitpick)

agent/src/beerocks/slave/link_metrics/link_metrics.h
```diff
@@ -0,0 +1,123 @@
...
+     * Number of packets received at the interface during the same measurement period used to count
+     * packet_errors
+     */
+    uint32_t packets_received = 0;
+
+    /**
+     * If the media type of the link is IEEE 802.11, then this value is the estimated RSSI in dB at
+     * the receive side of the link expressed in dB; otherwise, it is set to 0xFF.
+     */
+    uint8_t rssi = 0;
```
tomereli, line 76: So why not set it to UINT8_MAX by default instead of 0? Then you don't need to set it in the ethernet link collector implementation.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+                const uint32_t layer1_total_size   = 1538;
+                link_metrics.transmitter.mac_throughput_capacity_mbps =
+                    phy_rate_mbps * ((float)layer2_payload_size / (float)layer1_total_size);
+                // Note: For simplicity, link availability is set to "100% of the time"
+                link_metrics.transmitter.link_availability = 100;
+                link_metrics.transmitter.phy_rate_mbps     = phy_rate_mbps;
+
+                link_metrics.receiver.packet_errors    = stats->rx_errors;
+                link_metrics.receiver.packets_received = stats->rx_packets;
+                link_metrics.receiver.rssi             = UINT8_MAX;
```
tomereli, line 120: This should be the default in the struct so not needed to set here (nitpick)

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+ * @param[in, out] link_metris Link metrics structure with read values.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_link_metrics(int fd, const std::string &local_interface_name,
+                             sLinkMetrics &link_metrics)
+{
+    bool result = false;
+
+    struct sockaddr_nl kernel; /* the remote (kernel space) side of the communication */
```
tomereli: I think `socket` or `sock` is a more appropriate name here.
```suggestion
    struct sockaddr_nl sock; /* the remote (kernel space) side of the communication */
```
(nitpick)

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+{
+    bool result = false;
+
+    /**
+     * Create Netlink socket for kernel/user-space communication.
+     * No need to call bind() as packets are sent only between the kernel and the originating
+     * process (no multicasting).
+     */
+    int fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+    if (fd < 0) {
```
tomereli, line 251: Get rid of unnecessary indentation:
```suggestion
    if (fd < 0) {
        LOG(ERROR) << "Failed creating Netlink socket: " << strerror(errno);
        return false;
    }
```

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+ * protocol.
+ *
+ * @param[in] local_interface_name Name of the Ethernet network interface.
+ * @param[in, out] link_metris Link metrics structure with read values.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_link_metrics(const std::string &local_interface_name, sLinkMetrics &link_metrics)
+{
+    bool result = false;
```
tomereli, line 243: Not needed, just return true of false where needed.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * Create Netlink socket for kernel/user-space communication.
+     * No need to call bind() as packets are sent only between the kernel and the originating
+     * process (no multicasting).
+     */
+    int fd = socket(AF_NETLINK, SOCK_RAW, NETLINK_ROUTE);
+    if (fd < 0) {
+        LOG(ERROR) << "Failed creating Netlink socket: " << strerror(errno);
+    } else {
```
tomereli, line 253:
```suggestion
    /**
         * Get link metrics using Netlink socket
         */
        result = get_link_metrics(fd, local_interface_name, link_metrics);

        /**
         * Clean up and finish properly
         */
        close(fd);
        return true;
```

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
```diff
@@ -0,0 +1,279 @@
...
+    return result;
+}
+
+ieee802_3_link_metrics_collector::~ieee802_3_link_metrics_collector() {}
+
+bool ieee802_3_link_metrics_collector::get_link_metrics(const std::string &local_interface_name,
+                                                        const sMacAddr &neighbor_interface_address,
+                                                        sLinkMetrics &link_metrics)
+{
+    (void)neighbor_interface_address;
```
tomereli: We discussed this before - we should add an `UNUSED` macro instead (nitpick)

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/eLinkMetricNeighborType.h
```diff
@@ -0,0 +1,30 @@
+///////////////////////////////////////
+// AUTO GENERATED FILE - DO NOT EDIT //
+///////////////////////////////////////
+
```
tomereli, line 4: This commit is misplaced as it seems - the previous commit did not change tlvf.py..

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/eMediaType.h
```diff
@@ -36,6 +36,14 @@ enum eMediaType: uint16_t {
     UNKNONWN_MEDIA = 0xffff,
 };
 
+enum eMediaTypeGroup: uint8_t {
+    IEEE_802_3 = 0x0,
```
tomereli, line 40: Adding comments (possible in `yaml` file) is welcome here

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -78,6 +81,222 @@ namespace beerocks {
...
+ * @param[in, out] media_type_group The media type group of the connecting interface.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_media_type_group(const std::string &interface_name,
+                                 ieee1905_1::eMediaTypeGroup &media_type_group)
+{
+    // TODO: Currently, only Ethernet is supported
+    // When dealing with WiFi interfaces, use NL80211 command to know if media type group is WiFi
+    media_type_group = ieee1905_1::eMediaTypeGroup::IEEE_802_3;
```
tomereli, line 98: Which NL80211 command should be used to get the media type? Can you add this to the comment and perhaps a link to the task which will add it?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -78,6 +81,222 @@ namespace beerocks {
...
+            } else if (SPEED_1000 <= speed) {
+                media_type = ieee1905_1::eMediaType::IEEE_802_3AB_GIGABIT_ETHERNET;
+            } else {
+                media_type = ieee1905_1::eMediaType::UNKNONWN_MEDIA;
+            }
+
+            result = true;
+        }
+    } else if (ieee1905_1::eMediaTypeGroup::IEEE_802_11 == media_type_group) {
+        // TODO: Not supported yet
```
tomereli, line 135: Can we add LOG prints for these cases?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -78,6 +81,222 @@ namespace beerocks {
...
+    /**
+     * Get a link metrics collector suitable for the underlying network technology of the
+     * connecting interface.
+     * Collector choice depends on bits 15 to 8 of media type.
+     */
+    uint8_t media_type_msb = media_type >> 8;
+    if (0 == media_type_msb) {
+        collector = std::make_unique<ieee802_3_link_metrics_collector>();
+    } else if (1 == media_type_msb) {
+        // TODO: Create a collector for wireless interfaces
```
tomereli, line 192: Should print LOG(ERROR) in this case for now.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -78,6 +81,222 @@ namespace beerocks {
...
+        //collector = std::make_unique<ieee802_11_link_metrics_collector>();
```
tomereli, line 193: Why not create the collector here instead of putting dead code? You can just add a stub implementation for the 802_11 collector. (nitpick)

tomereli: Now I see you dereference the collector below - we can just define the collector as nullptr, add a log print and a TODO.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -78,6 +81,222 @@ namespace beerocks {
...
+    } else {
+        LOG(ERROR) << "Unsupported media type: " << std::hex << (int)media_type;
+        return (false);
+    }
+
+    /**
+     * Get link metrics information
+     */
+    if (collector &&
```
tomereli, line 203: I think it is better to split to 2 if conditions (nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -78,6 +81,222 @@ namespace beerocks {
...
+        collector->get_link_metrics(local_interface_name, neighbor_al_mac, link_metrics)) {
+
+        /**
+         * Get local interface information to obtain its MAC address.
+         */
+        network_utils::iface_info iface_info;
+        if (network_utils::get_iface_info(iface_info, local_interface_name) != 0) {
+            LOG(ERROR) << "Failed reading interface info for: " << local_interface_name;
+            return (false);
```
tomereli:
```suggestion
            return false;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -78,6 +81,222 @@ namespace beerocks {
...
+     * Get link metrics information
+     */
+    if (collector &&
+        collector->get_link_metrics(local_interface_name, neighbor_al_mac, link_metrics)) {
+
+        /**
+         * Get local interface information to obtain its MAC address.
+         */
+        network_utils::iface_info iface_info;
+        if (network_utils::get_iface_info(iface_info, local_interface_name) != 0) {
```
tomereli: You only use the mac address, so better to use:
```suggestion
        std::string mac;
        if (!network_utils::linux_iface_get_mac(local_interface_name, mac)) {
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2265,6 +2487,154 @@ bool backhaul_manager::handle_1905_higher_layer_data_message(ieee1905_1::CmduMes
...
+     * EUI-48 field shall be present."
+     *
+     * However, optional fields are not currently supported by TLVF.
+     *
+     * As a workaround, instead of defining a tlvLinkMetricQuery TLV with an optional field, we
+     * have defined two different TLVs: tlvLinkMetricQuery with the optional field and
+     * tlvLinkMetricQueryAllNeighbors without it. Application must check which of both TLVs has
+     * been received inside the message.
+     */
+    std::shared_ptr<ieee1905_1::tlvLinkMetricQueryAllNeighbors> tlvLinkMetricQueryAllNeighbors;
```
tomereli, line 2518: Perhaps we can handle each option in separate functions? (nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2210,21 +2210,26 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
-    uint32_t speed;
-    if (network_utils::linux_iface_get_speed(m_sConfig.wire_iface, speed)) {
-        std::shared_ptr<ieee1905_1::cLocalInterfaceInfo> localInterfaceInfo =
-            tlvDeviceInformation->create_local_interface_list();
 
+    /**
+     * Add a LocalInterfaceInfo field for the wired interface, if any.
+     */
+    std::string local_interface_name = m_sConfig.wire_iface;
+    // TODO: check if interface is UP?
```
tomereli: Why not check it then?
For example:
```
 // If a wired (WAN) interface was provided, try it first, check if the interface is UP
            wan_monitor::ELinkState wired_link_state = wan_monitor::ELinkState::eInvalid;
            if (!m_sConfig.wire_iface.empty()) {
                wired_link_state = wan_mon.initialize(m_sConfig.wire_iface);
                // Failure might be due to insufficient permissions, datailed error message is being
                // printed inside.
                if (wired_link_state == wan_monitor::ELinkState::eInvalid) {
                    LOG(WARNING) << "wan_mon.initialize() failed, skip wired link establishment";
                }
            }
            if ((wired_link_state == wan_monitor::ELinkState::eUp) &&
                (selected_backhaul.empty() || selected_backhaul == DEV_SET_ETH)) {
```

tomereli (approved):
I have only nitpicks - this is an amazing work @mariomaz - well done!

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Done

mariomaz (commented):
Sorry, I pressed the wrong button

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Done

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: RSSI field must be set to UINT8_MAX in Ethernet metrics only. It's not the default value

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Done

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Single vs. Multiple exit point, that's the question.
Have you agreed on which one to use or is it up to the programmer? I haven't seen anything about this topic in `CODINGSTYLE.md`

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: `result` from `get_link_metrics(fd, ...)` is being ignored with your proposal

mariomaz: If we opt for multiple exit point we must call `close` twice.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Fixed. 
However, I've used the C++11 attribute syntax [[gnu::unused]] instead of __attribute__((unused))
This way we do not have to define the UNUSED in some common header file and include such header.

agent/src/beerocks/slave/link_metrics/link_metrics.h
mariomaz: This struct is going to be used in the WiFi link collector implementation too. UINT8_MAX is not a default value, but the value to use for ethernet links (according to standard)

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/eLinkMetricNeighborType.h
mariomaz: This is not the first time I see this. Looks like github is messing the commits but they are actually ok. 
I've cloned the repository in another directory and double-checked that the order of these commits is correct.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done (The default initialization for unique_ptr points to nothing)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Some static code analyzers would complain in such case.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Most of the code is common to both options. Separated functions would contain lots of duplicated code so I prefer to leave it as it is now.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
tomereli: I think in general we favour multiple exit points, it keeps the code indentation shorter and is what we have in most of prplmesh. I will update the `CODYINGSTULE.md` if agreed - @fvbogaert you have a chance for another poll :)

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
tomereli: If we're going for single exit point then we should go all the way and do it like in the kernel, with `goto`. @fvbogaert should we use `goto`? Another poll..

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/eLinkMetricNeighborType.h
tomereli: @morantr @CoralMalachi I think you found a way to solve this in github?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done. 
However, I've used `network_utils::linux_iface_is_up_and_running()` to check interface status because it's simpler than your proposal and has no side effects.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
arnout: We use multiple exit points everywhere. Because normally, all cleanup is automatic since we use smart pointers etc.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
arnout: Yes, that's my preferred pattern if there is any cleanup to do. However, we really should use automatic cleanup everywhere. So the better approach would be to wrap the file descriptor so it gets closed automatically by the destructor.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
arnout: Would be nice to add that to CODING_STYLE.md.

agent/src/beerocks/slave/link_metrics/ieee802_3_link_metrics_collector.cpp
mariomaz: Read in https://wireless.wiki.kernel.org/en/developers/documentation/nl80211

> libnl tiny
> OpenWrt folks created a tiny version of libnl based on a git snapshot, which only contains genl, not rtnetlink or any of the netfilter stuff, and compiles down to less than 30k in binary size. You can find it here:
> https://git.openwrt.org/?p=openwrt/openwrt.git;a=tree;f=package/libs/libnl-tiny;hb=HEAD

Issue #738, opened 2020-01-24
[EPIC] Test flow refactoring


ghost:
# Description

Currently, all flows being tested are in a single shell script, with calls to docker to invoke certain commands on emulated agents.
The plan is to make this more modular by rewriting the test flows in python:

* Make it easier to add new tests
* Put each test in its own file
* Make it extensible so we can run the same tests on actual hardware as well (by abstracting away the target platform)
* Allow test hierarchies (controller/agent/algorithm test separation, for instance)

# Exit criteria

- [x] #726 Move test flows to top level directory
- [x] #894 Create python test framework
- [x] Move existing tests to python test framework
- [x] #896 Refactor test_flows.py
- [x] #919 Investigate whether we can use board farm
- [x] #1140 Document test_flows.

Some shortcoming of current tests should be addressed implicitly:

- [x] #821 check_error should be initialized

arnout:
Moved #822 to #1074

ghost:
All subtasks of this epic are done, so closing it :-)

PR #833, opened 2020-02-15
Hotfix/ucc listener get param

tomereli:
The UCC command to get the device ALid `dev_get_parameter,program,map,parameter,ALid` is currently implemented in the UCC listener base class, thus using hard coded bridge name "br-lan".
For most devices this is fine, but in this is not the case for others, such as CGR, where the bridge interface is called "brlan0".
Since both the controller UCC listener and the agent UCC listener hold the bridge name as part of the database / member, move handling to the derived classes.
This allows running UTAF autoconfig tests on RDKB.

### DIFF
--- a/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
+++ b/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
@@ -563,22 +563,14 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
             LOG(ERROR) << "failed to send reply";
             break;
         }
-        auto parameter = params["parameter"];
-        std::transform(parameter.begin(), parameter.end(), parameter.begin(), ::tolower);
         std::string value;
-        if (parameter == "alid") {
-            if (!net::network_utils::linux_iface_get_mac("br-lan", value)) {
-                LOG(ERROR) << "failed to get br-lan mac address";
-                reply_ucc(eWfaCaStatus::ERROR, "failed to get br-lan mac address");
-                break;
-            }
-        } else if (!handle_dev_get_param(params, value)) {
-            LOG(ERROR) << "failed to get parameter " << parameter << "error: " << value;
+        if (!handle_dev_get_param(params, value)) {
+            LOG(ERROR) << "failed to get parameter " << params["parameter"] << "error: " << value;
             reply_ucc(eWfaCaStatus::ERROR, value);
             break;
         }
         // Success
-        reply_ucc(eWfaCaStatus::COMPLETE, parameter + "," + value);
+        reply_ucc(eWfaCaStatus::COMPLETE, params["parameter"] + "," + value);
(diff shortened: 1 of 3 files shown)


common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -563,22 +563,14 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
...
-            if (!net::network_utils::linux_iface_get_mac("br-lan", value)) {
-                LOG(ERROR) << "failed to get br-lan mac address";
-                reply_ucc(eWfaCaStatus::ERROR, "failed to get br-lan mac address");
-                break;
-            }
-        } else if (!handle_dev_get_param(params, value)) {
-            LOG(ERROR) << "failed to get parameter " << parameter << "error: " << value;
+        if (!handle_dev_get_param(params, value)) {
+            LOG(ERROR) << "failed to get parameter " << params["parameter"] << "error: " << value;
             reply_ucc(eWfaCaStatus::ERROR, value);
```
morantr, line 569: I don't know why this part is implemented by using access element operator `[ ]` which creates the parameter if it not exist. This is not right and definitely not the convention in this file.
First, there should be a use of `find` function tho check if the "parameter" key exists on the map, and only then continue processing it.
Second, all common operations such as lowering the case of the value should remain in the base class.
Also, you might consider passing only the "parameter" element copy to `handle_dev_get_param` instead of the whole map.

morantr (changes requested):
One comment.
Also, there is a typo on 06f5c706e608f0f45319fc3e8bb91e96134b5668 commit message:
databse -> database

morantr (approved):

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
tomereli: I don't understand this comment - we check that `params["parameter"]` exists in `parse_params` in line 557 where "parameter" is a mandatory param, so its fine to access it with `[]` operator.
Regarding your other comment - since we pass the entire parameters map to the derived class it doesn't make sense to do any conversions in the base class - some parameters might require to be upper case, and we can't assume otherwise in the base `dev_get_parameter` implementation which does nothing but call the derived implmenetation.

rmelotte (approved):
The second and third commit headlines are very "generic" (nitpick).

Also there is one more reference of `br-lan` in `test_flows.sh` maybe it should also be updated?

tomereli:
> The second and third commit headlines are very "generic" (nitpick).
>
Not sure how to explain it better:)

> Also there is one more reference of `br-lan` in `test_flows.sh` maybe it should also be updated?
This is OK - in test_flows.sh we rely on the bridge we create in the docker container so its fine to use its name directly, but should be sorted out as part of #738 .

PR #834, opened 2020-02-15
Hotfix/dwpal attach before enable

tomereli:
Since prplmesh is passive (passive mode is the only mode), there is no
reason to wait till the hostapd gets to ENABLED (which is what
wifi_ctrl_enabled == 1 means).
Actually, there is no point to wait in static channel as well, since we
simply do not rely on the hostapd status, and if we do, it is a bug, and
we should not stop the attach FSM on this.
So - simply always pring the radio info and complete the attach FSM.

As a welcome consequence, the whole prplMesh init becomes much faster as
we don't wait for CAC to complete.

As another welcome consequence of this, which is why I came around to
debug this in the first case, the controller UCC listener now starts
much sooner, since it depends on the backhaul manager slave to join
which on many occasions used to be the one waiting for CAC, resulting
with no listener for more than a minute after starting prplmesh.

### DIFF
--- a/common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
@@ -205,51 +205,8 @@ bool base_wlan_hal_dwpal::fsm_setup()
                            << ", fixed_dfs_channel=" << int(fixed_dfs_channel)
                            << ", fixed_channel=" << int(fixed_channel);
 
-                // AP is Enabled
                 LOG(DEBUG) << "wifi_ctrl_enabled  = " << int(m_radio_info.wifi_ctrl_enabled);
-
-                bool error = false;
-                if (fixed_dfs_channel) {
-                    if (m_radio_info.tx_enabled) {
-                        LOG(DEBUG) << "ap_enabled tx = " << int(m_radio_info.tx_enabled);
-                        return true;
-                    } else {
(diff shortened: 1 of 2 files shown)


common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -205,51 +205,8 @@ bool base_wlan_hal_dwpal::fsm_setup()
...
-                    }
-                }
-
-                if (error || (std::chrono::steady_clock::now() >= m_state_timeout)) {
-                    return (transition.change_destination(dwpal_fsm_state::Detach));
-                }
-
-                // Remain in the current state
-                return false;
+                return true;
```
morantr, line 209: This is wrong. You can't continue the attach process until you are enabled.
There is a reason for that. You can't assure that the hostapd readings and requests will work when you are not enabled.

morantr (changes requested):

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
tomereli: Yes, but hostapd readings which require a specific state should verify that we reached this state and not just block the entire FSM for a reading which may or may not happen.
Also, if you mark it as "changes needed" please provide explanation on which readings will get affected.

PR #837, opened 2020-02-16
Hotfix/start controller ucc listener at init

tomereli:
Currently, the controller waits for slave join and only then starts the UCC listener.
This is an unnecessary by-product of the fact we are avoiding reading the platform DB by the controller.
Since we plan to allow the controller to read the platform DB anyway, as most configuration parameters which are shared between the controller and agent are stored in the platform DB, there is no reason to keep this separation anymore.
Update the controller to link with BPL, and read the certification mode directly at init.
Then, start the UCC listener at init instead of waiting for slave join.

We can do the same for the agent, but it is not really required since the platform manager is part of the agent and is anyway the first entity to which each agent component is registering to.

### DIFF
--- a/controller/src/beerocks/master/son_master_thread.h
+++ b/controller/src/beerocks/master/son_master_thread.h
@@ -114,7 +114,7 @@ private:
 
     db &database;
     task_pool tasks;
-    std::unique_ptr<beerocks::controller_ucc_listener> m_controller_ucc_listener;
+    beerocks::controller_ucc_listener m_controller_ucc_listener;
 };
 
 } // namespace son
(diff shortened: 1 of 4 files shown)


morantr (approved):

controller/src/beerocks/master/son_master_thread.h
```diff
@@ -112,7 +112,7 @@ class master_thread : public beerocks::btl::transport_socket_thread {
 
     db &database;
     task_pool tasks;
-    std::unique_ptr<beerocks::controller_ucc_listener> m_controller_ucc_listener;
+    beerocks::controller_ucc_listener m_controller_ucc_listener;
```
rmelotte, line 117: Could you explain this additional change in the commit message?

rmelotte (approved):

PR #650, opened 2020-01-07
Feature/build statically

mariomaz:
Make SHARED/STATIC libraries an option

Listing of lib and bin directories with SHARED libraries (Release build on x86_64, stripped):

```
../release/prplmesh-install/bin:
total 5.3M
-rwxr-xr-x 1 arnout arnout 2.1M Jan  9 17:24 beerocks_agent*
-rwxr-xr-x 1 arnout arnout 417K Jan  9 17:24 beerocks_cli*
-rwxr-xr-x 1 arnout arnout 2.4M Jan  9 17:24 beerocks_controller*
-rwxr-xr-x 1 arnout arnout 389K Jan  9 17:24 beerocks_monitor*
-rwxr-xr-x 1 arnout arnout  21K Jan  9 17:24 ieee1905_transport*
-rwxr-xr-x 1 arnout arnout  21K Jan  9 17:24 local_bus*
-rwxr-xr-x 1 arnout arnout  17K Jan  9 17:24 mapfcli*
-rwxr-xr-x 1 arnout arnout  17K Jan  9 17:24 version*

../release/prplmesh-install/lib64:
total 5.4M
lrwxrwxrwx 1 arnout arnout   11 Jan  3 19:55 libbcl.so -> libbcl.so.1*
lrwxrwxrwx 1 arnout arnout   15 Jan  3 19:55 libbcl.so.1 -> libbcl.so.1.4.0*
-rwxr-xr-x 1 arnout arnout 466K Jan  9 17:24 libbcl.so.1.4.0*
lrwxrwxrwx 1 arnout arnout   11 Jan  3 19:55 libbml.so -> libbml.so.1*
lrwxrwxrwx 1 arnout arnout   15 Jan  3 19:55 libbml.so.1 -> libbml.so.1.4.0*
-rwxr-xr-x 1 arnout arnout 582K Jan  9 17:24 libbml.so.1.4.0*
lrwxrwxrwx 1 arnout arnout   11 Jan  3 19:55 libbpl.so -> libbpl.so.1*
lrwxrwxrwx 1 arnout arnout   15 Jan  3 19:55 libbpl.so.1 -> libbpl.so.1.4.0*
-rwxr-xr-x 1 arnout arnout  55K Jan  9 17:24 libbpl.so.1.4.0*
lrwxrwxrwx 1 arnout arnout   11 Jan  3 19:55 libbtl.so -> libbtl.so.1*
[...]

PR #851, opened 2020-02-19
Allow building statically

### LINKED ISSUE #469: [TASK] Add possibility for building statically

arnout:
This PR is an update of #650, which didn't work completely.

With this PR, `BUILD_SHARED_LIBS` is an option, and explicit SHARED is removed from all library definitions.

Tested by running test_flows with a statically built beerocks.
All internal libraries are static, but the executables of course still link dynamically with libc, openssl etc.
To completely link statically, the caller/integration must set `CMAKE_EXE_LINKER_FLAGS` to `-static`.


mariomaz (approved):
You are great. I'm speechless.

tomereli (approved):

PR #853, opened 2020-02-19
.mergify.yml: add mergify configuration

arnout:
Supposedly, the configuration can be tested at
https://dashboard.mergify.io/installation/6872982/simulator?repositoryName=prplMesh&pullRequestNumber=808
but I have had no success with it.

This configuration:
 - replicates the merge conditions we have now:
   - at least two approving reviews;
   - no changes requested;
   - no don't merge tag;
   - CI has finished successfully.
 - automatically rebases on master when master is updated ("strict"
   merge policy).

As far as I understand, the automatic rebase will only happen if all the
merge conditions are met. That's a bit unfortunate, since it means that
CI has to run first, then the rebase happens, and then CI has to run
again. The "smart" option of strict merges alleviates this problem
somewhat, by queueing the rebases if there is more than one.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,22 @@
+pull_request_rules:
+  - name: Automatic rebase and merge on approval
+    conditions:
+      - "#approved-reviews-by>=2" # Require at least two approving reviewers
+      - "#changes-requested-reviews-by=0" # Changes requested blocks the merge
+      - "label!=don't merge" # Don't merge label blocks the merge
+      # All status checks must pass. Since mergify itself is a status check, this is a bit
+      # recursive... So, explicitly say that there should be no unfinished (= neutral) status
+      # check, no failures obviously, and that at least two status checks are successful.
+      - "#status-neutral=0"
+      - "#status-failure=0"
+      - "#status-success>=2"
+    actions:
+      merge:
+        method: rebase # Merge with rebase
+        rebase_fallback: null # Manual intervention needed if that doesn't work
+        # If not up-to-date with master, bring it up-to-date first.
+        # Use rebase for bringing up-to-date
+        # Use "smart" to do the rebase one-by-one, in case multiple PRs need to be
+        # rebased.
+        strict: smart
+        strict_method: rebase


rmelotte (approved):

rmelotte:
At some point we also requested an approving review from at least one code owner. I don't know if it's still the case, but I guess adding Mergify doesn't make any difference in that regard.

Is the number of successful checks relevant if there can be no pending check, and no failing check?

tomereli (approved):

arnout:
> At some point we also requested an approving review from at least one code owner. I don't know if it's still the case, but I guess adding Mergify doesn't make any difference in that regard.

It is no longer the case, we now require two reviews but they can be from anybody.

It's not entirely clear if mergify takes into account the restrictions configured in github. So I don't know if "adding mergify doesn't make any difference" really is true.

> Is the number of successful checks relevant if there can be no pending check, and no failing check?

Again, I don't know if it's needed, but I think it might be needed. AFAIU, when a PR is opened or updated, it starts with 0 checks. Then the checks are added one by one in *some* order. If mergify  is the first one, and it runs very quickly, it is possible that it will see 0 successes, 0 neutral, and 0 failures, just because nothing was started yet. That's why I added the check for successes as well. I tried to explain that in the comment, but I guess I was too brief :-)

ghost (approved):

rmelotte:
> Again, I don't know if it's needed, but I think it might be needed. AFAIU, when a PR is opened or updated, it starts with 0 checks. Then the checks are added one by one in _some_ order. If mergify is the first one, and it runs very quickly, it is possible that it will see 0 successes, 0 neutral, and 0 failures, just because nothing was started yet. That's why I added the check for successes as well. I tried to explain that in the comment, but I guess I was too brief :-)

Ok. I assumed Github would create all the checks "at the same time", but if there is any possibility for it not to be the case, then yes let's keep it.

I'm trying to understand if there could be a possibility for the DCO and "Assign PR to author" checks to be enough for mergify to decide to do the merge, but from what you say it seems that they would have to both finish before the CI check is created, which seems unlikely.

I'm not sure "Assign PR to author" is always there, but if it is the case we could set the required successful checks to `>=3` just to be safe (that would be DCO, "Assign PR to author", and CI).

PR #854, opened 2020-02-19
easylogging: Remove sticky std::hex configuration and add showbase and boolalpha flags to logs

morantr:
Add to LoggingFlag on easylogging 3 new flags:
ForceDecBase - Set the log to print integer on decimal format with every log-entry.
ShowBase - Set the log to print the base of integer value according to iostream base flag.
BoolAlpha - Set the log to print the booleans as "true" or "false" instead of "1" or "0".

Set these configurations by default on all processes.
Now  the log will look like this:

LOG(DEBUG) << std::hex << "num_hex=" << 32 << ", bool=" << bool(false);
LOG(DEBUG) << "num_dec=" << 32 << ", bool=" << bool(true);

output:
```
DEBUG 16:33:21:731 <Some file> --> num_hex=0x20, bool=false
DEBUG 16:33:21:731 <Some file> --> num_dec=32, bool=true
```

#879

### DIFF
--- a/framework/external/easylogging/easylogging++.h
+++ b/framework/external/easylogging/easylogging++.h
@@ -725,8 +725,14 @@ enum class LoggingFlag : base::type::EnumType {
   AutoSpacing = 8192,
   /// @brief Preserves time format and does not convert it to sec, hour etc (performance tracking only)
   FixedTimeFormat = 16384,
-  // @brief Ignore SIGINT or crash
+  /// @brief Ignore SIGINT or crash
   IgnoreSigInt = 32768,
+  /// @brief Set the log to print integer on decimal format with every log-entry
+  ForceDecBase = 65536,
+  /// @brief Set the log to print the base of integer value according to iostream base flag
+  ShowBase = 131072,
+  /// @brief Set the log to print the booleans as "true" or "false" instead of "1" or "0"
+  BoolAlpha = 262144,
 };
 namespace base {
 /// @brief Namespace containing constants used internally.
--- a/common/beerocks/bcl/source/beerocks_logging.cpp
+++ b/common/beerocks/bcl/source/beerocks_logging.cpp
@@ -513,6 +513,9 @@ void logging::apply_settings()
     el::Loggers::addFlag(el::LoggingFlag::LogDetailedCrashReason);
     el::Loggers::addFlag(el::LoggingFlag::DisableApplicationAbortOnFatalLog);
     el::Loggers::addFlag(el::LoggingFlag::StrictLogFileSizeCheck);
+    el::Loggers::addFlag(el::LoggingFlag::ForceDecBase);
+    el::Loggers::addFlag(el::LoggingFlag::ShowBase);
+    el::Loggers::addFlag(el::LoggingFlag::BoolAlpha);
 
     auto logger = el::Loggers::getLogger("default");
     if (!logger) {
--- a/framework/external/easylogging/easylogging++.cc
+++ b/framework/external/easylogging/easylogging++.cc
@@ -2543,6 +2543,15 @@ void MessageBuilder::initialize(Logger* logger) {
   m_logger = logger;
   m_containerLogSeperator = ELPP->hasFlag(LoggingFlag::NewLineForContainer) ?
                             ELPP_LITERAL("\n    ") : ELPP_LITERAL(", ");
+  if (ELPP->hasFlag(LoggingFlag::ForceDecBase)) {
+    m_logger->stream().setf(std::ios::dec);
+  }
+  if (ELPP->hasFlag(LoggingFlag::ShowBase)) {
+    m_logger->stream().setf(std::ios::showbase);
+  }
+  if (ELPP->hasFlag(LoggingFlag::BoolAlpha)) {
+    m_logger->stream().setf(std::ios::boolalpha);
+  }
 }
 
 MessageBuilder& MessageBuilder::operator<<(const wchar_t* msg) {
--- a/tests/test_flows.sh
+++ b/tests/test_flows.sh
@@ -527,7 +527,7 @@ test_client_steering_policy() {
     check_log ${REPEATER1} agent_wlan0 "MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE"
     sleep 1
     dbg "Confirming client steering policy ack message has been received on the controller"
-    check_log ${GATEWAY} controller "ACK_MESSAGE, mid=$MID1_STR"
+    check_log ${GATEWAY} controller "ACK_MESSAGE, mid=0x$MID1_STR"
 
     return $check_error
 }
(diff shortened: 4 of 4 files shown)


tomereli (commented):
I don't think this should be merged. This has nothing to do with easylogging - this is how the standard library works and I don't think we should change it.
But I'm open to other opinions - @arnout @rmelotte @vitalybu

morantr:
> I don't think this should be merged. This has nothing to do with easylogging - this is how the standard library works and I don't think we should change it.
> But I'm open to other opinions - @arnout @rmelotte @vitalybu

It has nothing with easylogging, but as I can see it, changing it on easylogging is the only way.
Right now it is very hard to debug things. For example on optimal path print of rssi measurement looks like this:
"rssi=ffffffc4"

So unless you want to manually add std::dec after every print, this is the way to go.

vitalybu (changes requested):
I agree with @tomereli. 
We should not modify external libraries source code.
If you think it's a useful change, you can try adding into the easylogging repo.

tomereli:
> > I don't think this should be merged. This has nothing to do with easylogging - this is how the standard library works and I don't think we should change it.
> > But I'm open to other opinions - @arnout @rmelotte @vitalybu
> 
> It has nothing with easylogging, but as I can see it, changing it on easylogging is the only way.
> Right now it is very hard to debug things. For example on optimal path print of rssi measurement looks like this:
> "rssi=ffffffc4"
> 
> So unless you want to manually add std::dec after every print, this is the way to go.

You can add `std::dec` where you think it is important for debugging that the print will be decimal, it certainly isn't the case in all cases. The real solution is to add std::dec in the places where we switched to std::hex and forgot to reset it.

morantr:
> > > I don't think this should be merged. This has nothing to do with easylogging - this is how the standard library works and I don't think we should change it.
> > > But I'm open to other opinions - @arnout @rmelotte @vitalybu
> > 
> > 
> > It has nothing with easylogging, but as I can see it, changing it on easylogging is the only way.
> > Right now it is very hard to debug things. For example on optimal path print of rssi measurement looks like this:
> > "rssi=ffffffc4"
> > So unless you want to manually add std::dec after every print, this is the way to go.
> 
> You can add `std::dec` where you think it is important for debugging that the print will be decimal, it certainly isn't the case in all cases. The real solution is to add std::dec in the places where we switched to std::hex and forgot to reset it.

It will work on theory, but developers will forget to do add it after switching to hex, and the log will be impossible to read. Also, it will be very hard to map all the places you need decimal value.
Is it possible to make clang-format to add `<< std::dec` at the end of a stream that contains `std::hex`?

tomereli:
@arnout please share your opinion

tomereli:
> > > > I don't think this should be merged. This has nothing to do with easylogging - this is how the standard library works and I don't think we should change it.
> > > > But I'm open to other opinions - @arnout @rmelotte @vitalybu
> > > 
> > > 
> > > It has nothing with easylogging, but as I can see it, changing it on easylogging is the only way.
> > > Right now it is very hard to debug things. For example on optimal path print of rssi measurement looks like this:
> > > "rssi=ffffffc4"
> > > So unless you want to manually add std::dec after every print, this is the way to go.
> > 
> > 
> > You can add `std::dec` where you think it is important for debugging that the print will be decimal, it certainly isn't the case in all cases. The real solution is to add std::dec in the places where we switched to std::hex and forgot to reset it.
> 
> It will work on theory, but developers will forget to do add it after switching to hex, and the log will be impossible to read. Also, it will be very hard to map all the places you need decimal value.
> Is it possible to make clang-format to add `<< std::dec` at the end of a stream that contains `std::hex`?

It's not the issue where to put it, it's if to put this at all.
This will break cases where user wants to dump multiple LOGs with std::hex for example:
```
LOG(DEBUG) << "start dump in hex" << std::hex ;
...
LOG(DEBUG) << "val=" << val;
...
LOG(DEBUG) << "val2=" << val2;
...
LOG(DEBUG) << "End dump in hex" << std::dec;
```

arnout:
> It will work on theory, but developers will forget to do add it after switching to hex
As mentioned before, this is standard library behaviour:
```
#include <iostream>

using namespace std;
int main ()
{
    cout << "hex: " << hex << 0x40 << " still hex: " << 100 << endl;
    cout << "still hex: " << 100 << endl;
    return 0;
}
```

That said, I agree that the standard C++ behaviour is extremely annoying. So I kind of agree with this change.  Only, I think doing it in the LOG macro is not appropriate. It is IMO more appropriate to do it in `MessageBuilder::initialize()` and to explicitly set all flags with `m_logger->stream().flags(...)`. We could also add some other useful flags like showbase.

tomereli:
> > It will work on theory, but developers will forget to do add it after switching to hex
> > As mentioned before, this is standard library behaviour:
> 
> ```
> #include <iostream>
> 
> using namespace std;
> int main ()
> {
>     cout << "hex: " << hex << 0x40 << " still hex: " << 100 << endl;
>     cout << "still hex: " << 100 << endl;
>     return 0;
> }
> ```
> 
> That said, I agree that the standard C++ behaviour is extremely annoying. So I kind of agree with this change. Only, I think doing it in the LOG macro is not appropriate. It is IMO more appropriate to do it in `MessageBuilder::initialize()` and to explicitly set all flags with `m_logger->stream().flags(...)`. We could also add some other useful flags like showbase.

I'm OK with this solution. We should not block changes in easylogging since we don't know if they will be accepted, so I suggest to implement this and then upstream all our changes in a single PR. @morantr I think it is OK if we just create a task for upstreaming the easylogging changes for now, and mark it as "good first issue" :)

ghost:
I still don't really get why people think working with `operator<<` and ostreams is so much better than using C streams and `fprintf`. At least, then you always know what you're going to get ...

dammit! I'm sounding like a grumpy old man again! And I swear I'm only 31 years old ...

tomereli:
Wait, so now this means I can never print in hex?

framework/external/easylogging/easylogging++.h
```diff
@@ -725,8 +725,14 @@ enum class LoggingFlag : base::type::EnumType {
   AutoSpacing = 8192,
   /// @brief Preserves time format and does not convert it to sec, hour etc (performance tracking only)
   FixedTimeFormat = 16384,
-  // @brief Ignore SIGINT or crash
+  //  @brief Ignore SIGINT or crash
```
vitalybu: What's up with the extra space?

morantr: All other similar lines were with like this, so I aligned to the rest of the lines.
Edit:
I now realize that I was missing slash. All other comments were with 3 of them, and I added only two. Fixed.

morantr:
> What will happen now if I do `LOG(DEBUG) <<std::hex<< x;`? Will it still print X in decimal instead of hex?

It will print in hex notation until the end of the log print call.
The difference is that now we don't need to change it back to `std::dec`. It will be done automatically before every new print.
Please see the PR description. I have updated it, and it explains it.

tomereli (approved):

vitalybu:
Overall I'm OK with this change, but as we introduce more and more changes into this library, maybe we should move to CMake's [ExternalProject](https://cmake.org/cmake/help/v3.8/module/ExternalProject.html) + our own patches.

vitalybu (approved):

PR #855, opened 2020-02-19
scripts: search full command line with pgrep -f

tomereli:
prplmesh_utils.sh includes a `killall_program()` function which is used
to kill processes which were brought up by the script when stopping
prplmesh, and is using pgrep to search for the PID to kill.

Commit 574fe63 removed the -f for not
killing other applications which have the command line string of the
process we are killing in their command line - used when we tail the
logs.

This solved the issue where tail was also killed when we stopped
prplmesh, but it created a new bug, which was the reason -f was used in
the first place -
By default, pgrep only searches the process name,
which is a truncated version of the entire command, for example in
ubuntu16.04 LTS:
"tester@WcsTEst100:~/work/prplMesh$ cat /proc/6389/status
Name:   beerocks_contro"

This means that pgrep beerocks_controller simply will not work, which
results with processes not being killed as they should.

Fix this by bringing back -f flag for pgrep which will match the full
process name. Not that with the -f flag, pgreping the truncated name
will also work, so this will not break other cases.

The second commit fixes the "tail" issue by matching the process name with the full executable name we read from `/proc/$PID/exe`.

### DIFF
--- a/common/beerocks/scripts/prplmesh_utils.sh.in
+++ b/common/beerocks/scripts/prplmesh_utils.sh.in
@@ -33,7 +33,7 @@ killall_program() {
     else
         KILL_SIG=$SIG_KILL
     fi
-    for PID in $(pgrep "$PROGRAM_NAME"); do
+    for PID in $(pgrep -f "$PROGRAM_NAME"); do
         echo "kill $KILL_SIG $PID $PROGRAM_NAME";
         kill $KILL_SIG $PID > /dev/null 2>&1;
     done
--- a/common/beerocks/scripts/prplmesh_utils.sh.in
+++ b/common/beerocks/scripts/prplmesh_utils.sh.in
@@ -34,8 +34,10 @@ killall_program() {
         KILL_SIG=$SIG_KILL
     fi
     for PID in $(pgrep -f "$PROGRAM_NAME"); do
-        echo "kill $KILL_SIG $PID $PROGRAM_NAME";
-        kill $KILL_SIG $PID > /dev/null 2>&1;
+        if [ "$(basename "$(readlink /proc/"$PID"/exe)")" = "$PROGRAM_NAME" ]; then
+            echo "kill $KILL_SIG $PID $PROGRAM_NAME";
+            kill $KILL_SIG "$PID" > /dev/null 2>&1;
+        fi
     done
 }
 


ghost:
See https://github.com/prplfoundation/prplMesh/pull/839

morantr:
That was before my change and @rmelotte fix. So the tail will stop working as well as it used to.

tomereli:
> See #839

Thanks for that - this doesn't always work since the program name may be truncated, see https://unix.stackexchange.com/questions/267007/pgrep-full-match-not-work-only-part-why.

Due to this, on Ubuntu-16.04 (and I expect it could happen on other platforms), `pgrep ieee1905_transport` for example, fails to find the PID since the name is truncated in the proc filesystem. I guess it depends on the kernel version.

I added commit https://github.com/prplfoundation/prplMesh/pull/855/commits/1c75c663f4298fe97e7fe757dc044707f6fe78da which solves this completely while keeping the `-f` flag so we don't fail to find the PID in such platforms.

@rmelotte should review this of course, and I really hope we won't need to change it again :)

morantr (approved):

rmelotte (approved):
I find "fix killall_program()" again a bit too generic.
If we make another change fixing another issue with killall_program, it will be hard to distinguish the different fixes.
What about something like "killall_program(): search the full command line".
It can be harder to keep it short when describe it like that though.

It's only a nitpick and it's not that important to fix for this PR, but rather a comment for future PRs. Also others may have a different opinion on what the headline should say.

EDIT: After looking at  CONTIBUTING.md again, it specifically says to avoid using "fix" in headlines, so it looks like I'm not the only one with this opinion.

arnout (approved):
We're now doing exactly what `start-stop-daemon -K -x <path-to-program>` does. But I guess we don't have start-stop-daemon available on all platforms, so OK.

vitalybu (approved):

tomereli:
> I find "fix killall_program()" again a bit too generic.
> If we make another change fixing another issue with killall_program, it will be hard to distinguish the different fixes.
> What about something like "killall_program(): search the full command line".
> It can be harder to keep it short when describe it like that though.
> 
> It's only a nitpick and it's not that important to fix for this PR, but rather a comment for future PRs. Also others may have a different opinion on what the headline should say.
> 
> EDIT: After looking at CONTIBUTING.md again, it specifically says to avoid using "fix" in headlines, so it looks like I'm not the only one with this opinion.

I changed to `scripts: search full command line with pgrep -f` which I think is easier to search for in history and abides to the CONTRIBUTING.md.

tomereli:
> We're now doing exactly what `start-stop-daemon -K -x <path-to-program>` does. But I guess we don't have start-stop-daemon available on all platforms, so OK.

Actually, I checked in prplWrt RAX40, Turris-Omnia, UGW and RDKB - we have `start-stop-daemon` in ALL platforms. I am now confident we should change to use it, and I'll do this right away cause I found yet another bug with our killall_program! If we override the binary, killall_program will fail since `readlink <binary path>` will show it as deleted till the next time it runs the new binary.

PR #856, opened 2020-02-19
bwl: dwpal: set empty ssids when tearing down vaps

tomereli:
Currently, tear down is implemented with setting start_disabled=1 for
each tore down VAP in the hostapd conf.
The backhaul manager uses the VAPs SSIDs when building the AP operational
BSS TLV in the topology response message by testing for empty string to
determine if the VAP is enabled and should be reported or not.
Fix that by setting the SSID to empty string in the BWL DWPAL
implementation when tearing down a VAP.


arnout (approved):
(nitpick) it would maybe make sense to also clear the credentials.

morantr (approved):

PR #859, opened 2020-02-20
tests: convert test_flows.sh to python (test_flows.py)

### LINKED ISSUE #894: [TASK] Create python test framework
## Description

Currently, all flows being tested are in a single shell script, with calls to docker to invoke certain commands on emulated agents.
The plan is to make this more modular by rewriting the test flows in python:

- Make it easier to add new tests
- Put each test in its own file
- Make it extensible so we can run the same tests on actual hardware as well (by abstracting away the target platform)
- Allow test hierarchies (controller/agent/algorithm test separation, for instance)

## Exit criteria

- [ ] A simple python test framework exists
- [ ] All existing tests are migrated to this framework
- [ ] All tests pass
- [ ] All functionality is preserved (i.e. logs, tcpdump captures, ...)

## Detailed description

This is a subtask of #738 . See that epic for a description of the context.

arnout:
test_flows.py is a drop-in replacement for test_flows.sh.
This PR ports it directly without any changes or refactorings, except for a few things which are easier to do differently in Python.
E.g. searching in the log file is done directly in Python, not by calling grep.

Fixes #894

### DIFF
--- a/tests/send_CAPI_command.py
+++ b/tests/send_CAPI_command.py
@@ -3,7 +3,7 @@ import argparse
 import logging
 import socket
 from enum import Enum
-from typing import Union
+from typing import Union, Dict
 
 logger = logging.getLogger(__name__)
 
@@ -60,7 +60,7 @@ class UCCSocket:
             command += "\n"
         self.conn.send(command.encode("utf-8"))
 
-    def get_reply(self, verbose: bool = False) -> str:
+    def get_reply(self, verbose: bool = False) -> Dict[str, str]:
         """Wait until the server replies with a `CAPIReply` message other than `CAPIReply.RUNNING`.
 
         The replies from the server will be printed as they are received.
@@ -74,8 +74,10 @@ class UCCSocket:
 
         Returns
         -------
-        str
-            The reply from the device.
+        Dict[str, str]
+            A mapping of parameter -> value. The CAPI COMPLETE message is followed by
+            parameter,value pairs. These are converted to a dict and returned. If the COMPLETE
+            message has no parameters, an empty dict is returned.
         """
         data = bytearray()
         while True:
@@ -94,13 +96,15 @@ class UCCSocket:
                 elif CAPIReply.COMPLETE.value in r:
                     if verbose:
                         print(r)
-                    return r
+                    reply_value_str = r[len(CAPIReply.COMPLETE.value) + 1:].strip()
+                    reply_values = reply_value_str.split(',')
+                    return {k: v for k, v in zip(reply_values[::2], reply_values[1::2])}
                 elif CAPIReply.INVALID.value in r or CAPIReply.ERROR.value in r:
                     raise ValueError("Server replied with {}".format(r))
                 else:
                     raise ValueError("Received an unknown reply from the server:\n {}".format(r))
 
-    def cmd_reply(self, command: str, verbose: bool = False) -> str:
+    def cmd_reply(self, command: str, verbose: bool = False) -> Dict[str, str]:
         """Open the connection, send a command and wait for the reply."""
         with self:
             self.send_cmd(command)
--- a/tests/send_CAPI_command.py
+++ b/tests/send_CAPI_command.py
@@ -100,6 +100,11 @@ class UCCSocket:
                 else:
                     raise ValueError("Received an unknown reply from the server:\n {}".format(r))
 
+    def cmd_reply(self, command: str, verbose: bool = False) -> str:
+        """Open the connection, send a command and wait for the reply."""
+        with self:
+            self.send_cmd(command)
+            return self.get_reply(verbose)
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="Simulated UCC")
@@ -109,6 +114,4 @@ if __name__ == "__main__":
     args = parser.parse_args()
     socket.gethostbyname(args.host)
 
-    with UCCSocket(args.host, args.port) as sock:
-        sock.send_cmd(args.command)
-        sock.get_reply(True)
+    UCCSocket(args.host, args.port).cmd_reply(args.command, True)
--- a/tests/send_CAPI_command.py
+++ b/tests/send_CAPI_command.py
@@ -60,12 +60,17 @@ class UCCSocket:
             command += "\n"
         self.conn.send(command.encode("utf-8"))
 
-    def get_reply(self) -> str:
+    def get_reply(self, verbose: bool = False) -> str:
         """Wait until the server replies with a `CAPIReply` message other than `CAPIReply.RUNNING`.
 
         The replies from the server will be printed as they are received.
         Note that this method only returns once a `CAPIReply.COMPLETE`, `CAPIReply.INVALID`,
         or `CAPIReply.ERROR` message has been received from the server.
+
+        Parameters
+        ----------
+        verbose : bool
+            If True, print out the valid replies (RUNNING and COMPLETE) as they arrive.
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


ghost (approved):
Nice transition to python! I can't really fault it.

arnout:
Conversion is not complete yet, but it is already getting big so it would be nice to get this merged already.

tests/test_flows.py
```diff
@@ -0,0 +1,356 @@
...
+    def run_tests(self):
+        '''Run all tests as specified on the command line.'''
+        total_errors = 0
+        for test in self.opts.tests:
+            self.start_test(test)
+            self.tcpdump_start()
+            self.check_error = 0
+            try:
+                getattr(self, 'test_' + test)()
+            finally:
```
ghost, line 208: There should be a way to find out if one of the tests in self.opts.tests does not exist. `getattr` raises an `AttributeError` in that case IIRC.

```suggestion
            except AttributeError:
                self.err("test %s does not exist" % test)
                self.check_error += 1
            finally:
```

ghost: That's only a nitpick though, since the list of tests is generated automatically and shouldn't ever be wrong (`self.tests = [attr[len('test_'):] for attr in dir(self) if attr.startswith('test_')]`)

tests/test_flows.py
```diff
@@ -0,0 +1,356 @@
...
+        self.gateway_ucc.dev_send_1905(self.mac_repeater2, 0x0002)
+        time.sleep(1)
+        self.debug("Confirming topology query was received")
+        self.check_log(self.repeater2, "agent", "TOPOLOGY_QUERY_MESSAGE")
+
+        self.debug("Send Client Steering Request message for Steering Mandate to CTT Agent1")
+        self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8014,
+            (0x9B, 0x001b, "{%s 0xe0 0x0000 0x1388 0x01 {0x000000110022} 0x01 {%s 0x73 0x24}}" %
+                (self.mac_repeater1_wlan0, self.mac_repeater2_wlan0)))
+        time.sleep(1)
```
ghost, line 327: Probably better to add `time.sleep(1)` to the `dev_send_1905` command, instead of having to write it manually each time. People are going to forget to do that time and time again.
Also, I believe in python you can do `time.sleep(0.1)`, i.e. you don't need to wait for an integer number of seconds.

tests/send_CAPI_command.py
```diff
@@ -84,15 +84,53 @@ def get_reply(self) -> str:
...
+        with self:
+            self.send_cmd(command)
+            return self.get_reply(verbose)
+
+    def dev_get_parameter(self, parameter: str) -> str:
+        """Call dev_get_parameter and return the parameter, or raise ValueError if it is missing."""
+        return self.cmd_reply("dev_get_parameter,program,map,parameter,{}".format(parameter))[parameter]
+
+    def dev_send_1905(self, dest: str, message_type: int, *tlvs: Tuple[int, int, str]) -> int:
+        """Call dev_send_1905 to "dest" with CMDU type "message_type" and additional "tlvs".
```
rmelotte: About tlvs: I'm not a big fan of using custom collections instead of classes unless they're really trivial.
It makes it harder for other people to know what it's supposed to contain, what you can do with it, and so on.
Especially since you need the format_tlv function and that tlvs are probably going to be used elsewhere, I think it would make it clearer to have a TLV class.

tests/send_CAPI_command.py
```diff
@@ -60,7 +60,7 @@ def send_cmd(self, command: str) -> None:
             command += "\n"
         self.conn.send(command.encode("utf-8"))
 
-    def get_reply(self) -> str:
+    def get_reply(self, verbose: bool = False) -> Dict[str, str]:
         """Wait until the server replies with a `CAPIReply` message other than `CAPIReply.RUNNING`.
 
         The replies from the server will be printed as they are received.
```
rmelotte, line 66: Could you add "verbose" to the docstring?
Something like:
`If set to True, the replies from the server will be printed as they are received.`
It doesn't look like arguments are documented in docstrings elsewhere. Maybe it's the good time to agree on the docstring style we would use.

rmelotte: The returned type should also be documented, especially since it's a bit hard to understand what the result will be if you don't already know the syntax of the CAPI replies.

tools/docker/stop.sh
```diff
@@ -43,13 +43,18 @@ main() {
...
+        while read -r container; do
+            docker "$stop_cmd" "${container}" >/dev/null 2>&1 || true
+            if [ "$remove" = true ] ; then
+                docker rm "${container}" >/dev/null 2>&1 || true
+            fi
+        done
+
+    sed -n '/^network \(.*\)$/s//\1/p' "$containers_file" |\
+        while read -r network; do
+            docker network rm -f "${network}" >/dev/null 2>&1 || true
```
rmelotte: The `-f` option doesn't seem to be available (at least not in  19.03.4) for `network rm`.
It should also check for "$remove". I don't know what happens if you try to remove a network used by some containers, I guess it just fails.

tests/test_flows.py
```diff
@@ -0,0 +1,356 @@
...
+        return subprocess.check_output(("docker", "exec", device) + command)
+
+    def open_CAPI_socket(self, device: str, controller: bool = False) -> send_CAPI_command.UCCSocket:
+        '''Open a CAPI socket to the agent (or controller, if set) on "device".'''
+        # First, get the UCC port from the config file
+        if controller:
+            config_file_name = 'beerocks_controller.conf'
+        else:
+            config_file_name = 'beerocks_agent.conf'
+        with open(os.path.join(self.installdir, 'config', config_file_name)) as config_file:
```
rmelotte, line 134: As you said we'll need to handle it in a different manner later, I'm just commenting so that we don't forget.

rmelotte (approved):
I guess the overall architecture is going to change quite a lot to make it possible to run it with different backends?
If yes I don't see any blocker except the missing option for docker network, so approving but marking as don't merge for the bot.

tests/test_flows.py
arnout: Also, when it gets converted to unittest, this code will go away anyway, so no need to spend time on it I'd say.

tests/test_flows.py
arnout: This sleep is taken over directly from the existing test_flows.sh script, that's why I kept it.

I have added a sleep(.1) in check_log. With python, it is needed a lot more often because there's no fork/exec overhead unlike the shell script. I chose to do it in check_log instead of dev_send_1905 because my idea is to convert check_log into a wait-with-timeout, which will remove the need for sleeping again.

tests/send_CAPI_command.py
arnout: Very good point. However, I won't do that in this PR, instead I'll do it in part 2.

tests/send_CAPI_command.py
arnout: >  It doesn't look like arguments are documented in docstrings elsewhere. Maybe it's the good time to agree on the docstring style we would use.

Actually they are, in the send_cmd function. So I've taken over that style.

tools/docker/stop.sh
arnout: Well, that's exactly what the -f option does: remove any containers connected to that network. But apparently it's podman-specific, then.

tests/test_flows.py
arnout: Eh, I'm afraid I forgot what the "as I said" is...

rmelotte: Actually I don't know why I've put my comment here.
I originally wanted to comment on the `cmd_reply` function but it was not something to fix in this PR (the "as you said" was to open the each socket only once in the beginning of the test suite, and to make sure it gets closed at the end).

PR #871, opened 2020-02-24
[DRAFT] moving no_vendor_specific to UCI

vitalii-komisarenko:


tomereli:
This goes in a wrong direction IMO - why can't we just read it in the platform manager like we do for other platform db params?

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -168,7 +169,7 @@ static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_
     son_slave_conf.backhaul_wireless_iface_filter_low =
         beerocks::string_utils::stoi(beerocks_slave_conf.sta_iface_filter_low[slave_num]);
     son_slave_conf.backhaul_wireless_iface_type = son_slave_conf.hostap_iface_type;
-    son_slave_conf.no_vendor_specific           = beerocks_slave_conf.no_vendor_specific == "1";
+    // son_slave_conf.no_vendor_specific =
```
arnout, line 172: Never leave something commented-out. Just remove the code.

arnout (commented):
Not a thorough review so not approving yet.

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -534,8 +537,14 @@ int main(int argc, char *argv[])
             auto hostap_iface_elm = interfaces_map.find(slave_num);
             if ((hostap_iface_elm != interfaces_map.end()) &&
                 (g_son_slave_iface == hostap_iface_elm->second)) {
-                return run_son_slave(slave_num, beerocks_slave_conf, hostap_iface_elm->second, argc,
-                                     argv);
+                int no_vendor_specific = beerocks::bpl::cfg_get_no_vendor_specific();
```
arnout, line 540: Clarifying what @tomereli commented earlier: normally, the only component making calls to bpl should be the platform_manager_thread. Although there are already a couple of calls to bpl from this file, they really should be exceptions.

So @tomereli's proposal is to add it to the ACTION_PLATFORM_SON_SLAVE_REGISTER_RESPONSE structure and fill it in in fill_platform_settings() in platform_manager_thread. The only thing to be checked is if it then arrives on time on the slave thread (i.e. before the no_vendor_specific config is used). I think it does, but needs to be checked.

agent/src/beerocks/slave/beerocks_slave_main.cpp
arnout: Also, for completeness, the no_vendor_specific option should be removed from sSlaveConfig and moved to another structure.

vitalii-komisarenko:
Closed due to moving changes into another pull request.

PR #875, opened 2020-02-24
Bugfix: Fix AL-MAC getter inconsistencies in test_flows.sh

Galvien:
## Overview

Running `test_flows.sh` tests with the argument `-- skip-init` revealed a flaw in the mechanism for getting different bridges' AL-MAC addresses, this was due to the fact `bml_conn_map` often scrambled the order of repeaters.

As a result of this behavior, some tests would fail or show inconsistent results, for example, 
The teardown test would tear down the wrong repeater. When ran, it would try to update **repeater 1**, but repeater 1 would appear **_second_** in the `conn_map` therefore it would try to perform a teardown on **repeater 2**, obviously failing the test as repeater 1 would still hold an unmodified configuration.
This is merely one example.
## Fix

Implement a function that gets a bridge's al mac using dev_get_parameter, and the relevant docker **container's** ipv4 address, using the same protocol utilized in the UCC testbed.
This assures consistency and accuracy, and no false positives or false negatives.

### DIFF
--- a/tests/test_flows.sh
+++ b/tests/test_flows.sh
@@ -130,7 +130,7 @@ test_initial_ap_config() {
 }
 
 test_ap_config_renew() {
-    status "test initial autoconfig"
+    status "test autoconfig renew"
 
     # Regression test: MAC address should be case insensitive
     MAC_AGENT1=$(echo $mac_agent1 | tr a-z A-Z)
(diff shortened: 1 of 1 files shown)


Galvien:
@arnout I was asked to tag you to suggest you use this method when implementing the new test flows script 😀

tomereli (approved):

arnout (approved):

PR #885, opened 2020-02-25
Feature/4.8.1 client steering mandate

### LINKED ISSUE #400: [TASK] Pass flow 4.8.1 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.8.1 in the certification testbed.

# Exit criteria

### LINKED ISSUE #429: [TASK] EasyMesh SteeringBTMReport should contain source BSSID and target BSSID
Extend BWL to return source BSSID and target BSSID in BTM response, so
manager could add them to SteeringBTMReport.

rmelotte:
The source BSSID was missing from the BTM report, which made test 4.8.1 fail.

Add it to the internal BTM responses so that it can also be added to the BTM report.

Fixes #429

### DIFF
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -2276,22 +2276,25 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
 
         char MACAddress[MAC_ADDR_SIZE] = {0};
         int status_code                = 0;
+        char vap_name[beerocks::message::IFACE_NAME_LENGTH] = {0};
         size_t numOfValidArgs[4]       = {0};
         FieldsToParse fieldsToParse[]  = {
             {NULL /*opCode*/, &numOfValidArgs[0], DWPAL_STR_PARAM, NULL, 0},
-            {NULL, &numOfValidArgs[1], DWPAL_STR_PARAM, NULL, 0},
+            {(void*) vap_name, &numOfValidArgs[1], DWPAL_STR_PARAM, NULL,
+             beerocks::message::IFACE_NAME_LENGTH},
             {(void *)MACAddress, &numOfValidArgs[2], DWPAL_STR_PARAM, NULL, sizeof(MACAddress)},
             {(void *)&status_code, &numOfValidArgs[3], DWPAL_INT_PARAM, "status_code=", 0},
             /* Must be at the end */
             {NULL, NULL, DWPAL_NUM_OF_PARSING_TYPES, NULL, 0}};
 
-        if (dwpal_string_to_struct_parse(buffer, bufLen, fieldsToParse, sizeof(MACAddress)) ==
-            DWPAL_FAILURE) {
+        if (dwpal_string_to_struct_parse(buffer, bufLen, fieldsToParse,
+                                         sizeof(vap_name) + sizeof(MACAddress) + sizeof(status_code)) == DWPAL_FAILURE) {
             LOG(ERROR) << "DWPAL parse error ==> Abort";
             return false;
         }
 
         /* TEMP: Traces... */
+        LOG(DEBUG) << "numOfValidArgs[1]= " << numOfValidArgs[1] << " VAP name= " << vap_name;
         LOG(DEBUG) << "numOfValidArgs[2]= " << numOfValidArgs[2] << " MACAddress= " << MACAddress;
         LOG(DEBUG) << "numOfValidArgs[3]= " << numOfValidArgs[3]
                    << " status_code= " << int(status_code);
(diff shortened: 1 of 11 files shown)


rmelotte:
The test `MAP-4.8.1_ETH_FH24G` on rax40 is here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/449752655 (using the Marvell controller, and not Broadcom).

rmelotte:
The test 4.8.1 on rax40 is here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/449752655

rmelotte:
I took a shortcut when I said it allows us to pass 4.8.1 .
With this PR we're passing test `MAP-4.8.1_ETH_FH24G` but with a different controller than what the test plan says (until https://github.com/prplfoundation/prplMesh/issues/886 is fixed).

rmelotte:
> The reserved fields are there for alignment, which is messed up now with the bssids. I think the status code should come at the end so the others are more or less aligned. And reserved3 I would just remove, it's pretty pointless at the end of a struct.

I don't really understand... What were the 3 reserved uint8_t supposed to do before this change?
I changed it to have status_code at the end.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -2276,22 +2276,25 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
...
+             beerocks::message::IFACE_NAME_LENGTH},
             {(void *)MACAddress, &numOfValidArgs[2], DWPAL_STR_PARAM, NULL, sizeof(MACAddress)},
             {(void *)&status_code, &numOfValidArgs[3], DWPAL_INT_PARAM, "status_code=", 0},
             /* Must be at the end */
             {NULL, NULL, DWPAL_NUM_OF_PARSING_TYPES, NULL, 0}};
 
-        if (dwpal_string_to_struct_parse(buffer, bufLen, fieldsToParse, sizeof(MACAddress)) ==
-            DWPAL_FAILURE) {
+        if (dwpal_string_to_struct_parse(buffer, bufLen, fieldsToParse,
+                                         sizeof(fieldsToParse)) == DWPAL_FAILURE) {
```
morantr: Not sure why, but I have a Deja Vu when I wrote this comment.
Anyway, setting the last argument to `sizeof(fieldsToParse)` is wrong.
According to DWPAL code, the last argument should is defined by this description:
`size_t userBufLen - The user total buffer size to hold parsed info from inside fieldsToParse`
This means that it actually should be the size of `vap_name + MACAddress + status_code`.

I see wrong usage of it on all DWPAL functions.

morantr (approved):
Only one comment, but approved since it is a small issue.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: I wondered about that as well...

arnout:
> > The reserved fields are there for alignment, which is messed up now with the bssids. I think the status code should come at the end so the others are more or less aligned. And reserved3 I would just remove, it's pretty pointless at the end of a struct.
> 
> I don't really understand... What were the 3 reserved uint8_t supposed to do before this change?
> I changed it to have status_code at the end.

They were supposed to make sure the structure is aligned. However, it was already pretty bogus. sMacAddr is properly declared as packed, so it will occupy exactly 6 bytes. However, since sNodeBssSteerResponse is not declared as packed, the compiler may already insert 2 additional bytes between `mac` and `status_code`. So the only effect of the three reserved fields is to make sure the whole thing is 2-byte aligned. Which is anyway pointless, because it doesn't have any fields that require alignment. And moreover, in other places (e.g. tlvf) we're using structures in a way that assume the CPU has no alignment restrictions at all...

I think those reserved fields are simply cargo cult.

So my proposal would be to either insert a preparatory commit that removes all the reserved fields from base_wlan_hal_types.h, or to remove them only in sNodeBssSteerResponse in this commit.

arnout (approved):
I gave my review comments on the phone - sorry about that, it really sucks...

Some review comments really have to be taken into account, but I'm approving already with don't merge label.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
rmelotte: Replaced with the sum of the size of each element.

tomereli (approved):

rmelotte:
Unfortunately some of the comments disappeared with the force-push, but the only part I didn't address yet is this one:
`beerocks::net::network_utils::mac_from_string(get_radio_mac());`

The relevant comment is here: https://github.com/prplfoundation/prplMesh/pull/891#commitcomment-37577580

PR #888, opened 2020-02-26
[Patch 2/2] Feature/client capability response [Test 4.6.2]

CoralMalachi:
### Description
According to Multi-AP Specification, If a Multi-AP Agent receives a
Client Capability Query message, then within one second it shall respond
with a Client Capability Report message.

If the STA specified in the Client Capability Query message is not
associated with any of the BSS operated by the Multi-AP Agent (an error
scenario), the Multi-AP Agent shall set the result code in the Client
Capability Report TLV to 0x01 and include an Error Code TLV.
If it is not an error scenario, the Client Capability Report TLV should
include the frame body of the most recently received (Re)Association
Request frame from this client.

### Changes

- Parse the association request frame from the AP-STA-CONNECTED event.

- Save in the associated_clients_t map the frame body of the most recently 
received (Re)Association Request frame from each client.

- Add  the frame body of the most recently received (Re)Association
Request frame from the client to the Client Capability Report TLV

### Testbed status
Passed, you can check out the logs:https://gitlab.com/prpl-foundation/prplMesh/pipelines/133224675

there is a bug in the master version #1097, the logs above are before rebasing with the master which causes the bug---solved !

latest logs : https://gitlab.com/prpl-foundation/prplMesh/pipelines/135801475

### Next steps

- [x] Add an association frame in dummy mode.
- [x] Add a test to check it [dummy mode].
- [x] Currently, I'm facing a strange issue that the association frame is not delivered from son slave to backhaul manager, need to debug it. -----> Solved! an issue with the TLVF, thanks to @LiorAmram 
- [x] Solve the issue on the testbed

Signed-off-by: Coral Malachi coral.malachi@intel.com

### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1169,10 +1169,10 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
             return false;
         }
 
-        notification->params().mac          = msg->params.mac;
-        notification->params().vap_id       = msg->params.vap_id;
-        notification->params().bssid        = network_utils::mac_from_string(vap_node->second.mac);
-        notification->params().capabilities = msg->params.capabilities;
+        notification->mac()          = msg->params.mac;
+        notification->vap_id()       = msg->params.vap_id;
+        notification->bssid()        = network_utils::mac_from_string(vap_node->second.mac);
+        notification->capabilities() = msg->params.capabilities;
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
 
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1173,6 +1173,12 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
         notification->vap_id()       = msg->params.vap_id;
         notification->bssid()        = network_utils::mac_from_string(vap_node->second.mac);
         notification->capabilities() = msg->params.capabilities;
+        if (!msg->params.association_frame) {
+            LOG(DEBUG) << "no association frame";
+        } else {
+            notification->set_association_frame(msg->params.association_frame,
+                                                msg->params.association_frame_length);
+        }
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
 
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2033,8 +2033,19 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<sRadioInfo>
         }
 
         // Set client association information for associated client
-        auto &associated_clients              = soc->associated_clients_map[msg->bssid()];
-        associated_clients[msg->client_mac()] = std::chrono::steady_clock::now();
+        //remove this client from other radios
+        remove_client_from_all_radios(msg->client_mac());
+
+        auto &associated_clients = soc->associated_clients_map[msg->bssid()];
+
+        sClientInfo client_info;
+        client_info.client_mac = msg->client_mac();
+        client_info.bssid      = msg->bssid();
+        client_info.time_stamp = std::chrono::steady_clock::now();
+        client_info.asso_len   = msg->association_frame_length();
+        memcpy(client_info.assoc_req, msg->association_frame(), client_info.asso_len);
+        associated_clients[msg->client_mac()] = client_info;
+
         break;
     }
     case beerocks_message::ACTION_BACKHAUL_CLIENT_DISCONNECTED_NOTIFICATION: {
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2930,6 +2930,18 @@ bool backhaul_manager::send_slaves_enable()
     return true;
 }
 
+void backhaul_manager::remove_client_from_all_radios(sMacAddr &client_mac)
+{
+    for (const auto &slave : slaves_sockets) {
+        const auto &associated_clients_map = slave->associated_clients_map;
+        for (const auto &client : associated_clients_map) {
+            if (client.second.find(client_mac) != client.second.end()) {
+                slaves_sockets.remove(slave);
+            }
+        }
+    }
+}
+
 bool backhaul_manager::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event_ptr,
                                          std::string iface)
 {
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2145,11 +2145,14 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx 
     //with any of the BSS operated by the Multi-AP Agent [ though the TLV does contain a BSSID, the specification
     // says that we should answer if the client is associated with any BSS on this agent.]
     bool associated_client_found = false;
+    sClientInfo client_info;
     for (const auto &slave : slaves_sockets) {
         auto associated_clients_map = slave->associated_clients_map;
         for (const auto &vap : associated_clients_map) {
-            if (vap.second.find(client_info_tlv_r->client_mac()) != vap.second.end()) {
+            auto it = vap.second.find(client_info_tlv_r->client_mac());
+            if (it != vap.second.end()) {
                 associated_client_found = true;
+                client_info             = it->second;
                 break;
             }
         }
@@ -2175,18 +2178,22 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx 
         return false;
     }
 
-    //if it is an error scenario, set Success status to 0x01 = Failure and do nothing after it.
+    // if it is an error scenario, set Success status to 0x01 = Failure and do nothing after it.
     if (associated_client_found) {
         client_capability_report_tlv->result_code() = wfa_map::tlvClientCapabilityReport::SUCCESS;
         LOG(DEBUG) << "Result Code: SUCCESS";
-        //TODO: Add frame body of the most recently received (Re)Association Request frame from this client
+        // Add frame body of the most recently received (Re)Association Request frame from this client
+
+        client_capability_report_tlv->set_association_frame(client_info.assoc_req,
+                                                            client_info.asso_len);
+
     } else {
         client_capability_report_tlv->result_code() = wfa_map::tlvClientCapabilityReport::FAILURE;
 
         LOG(DEBUG) << "Result Code: FAILURE";
         LOG(DEBUG) << "STA specified in the Client Capability Query message is not associated with "
                       "any of the BSS operated by the Multi-AP Agent ";
-        //Add an Error Code TLV
+        // Add an Error Code TLV
         auto error_code_tlv = cmdu_tx.addClass<wfa_map::tlvErrorCode>();
         if (!error_code_tlv) {
             LOG(ERROR) << "addClass wfa_map::tlvErrorCode has failed";
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2530,7 +2530,7 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
                     // the timestamp of its last association.
                     for (const auto &associated_client : associated_clients) {
                         auto client_mac       = associated_client.first;
-                        auto association_time = associated_client.second;
+                        auto association_time = associated_client.second.time_stamp;
 
                         auto elapsed =
                             std::chrono::duration_cast<std::chrono::seconds>(now - association_time)
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -225,6 +225,14 @@ private:
 
     std::unique_ptr<beerocks::agent_ucc_listener> m_agent_ucc_listener;
 
+    struct sClientInfo {
+        sMacAddr client_mac;
+        sMacAddr bssid; // VAP mac
+        std::chrono::steady_clock::time_point time_stamp;
+        size_t asso_len;
+        uint8_t assoc_req[ASSOCIATION_FRAME_SIZE];
+    };
+
     /**
      * @brief Type definition for associated clients information.
      *
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -236,18 +236,21 @@ private:
     /**
      * @brief Type definition for associated clients information.
      *
-     * Associated client information consists of:
+     * Associated client information consists of sClientInfo strucrt, which has the 
+     * following fields:
      * - The MAC address of the 802.11 client that associates to a BSS.
      * - Timestamp of the 802.11 client's last association to this Multi-AP device.
+     * - The length of the association frame.
+     * - the association frame itself.
      *
      * Associated client information is gathered from
      * ACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION events received from slave threads.
      *
      * Associated client information is later used to fill in the Associated Clients TLV
-     * in the Topology Response message.
+     * in the Topology Response message and Client Capability Response message.
      */
-    typedef std::unordered_map<sMacAddr, std::chrono::steady_clock::time_point>
-        associated_clients_t;
+
+    typedef std::unordered_map<sMacAddr, sClientInfo> associated_clients_t;
 
     /**
      * @brief Information gathered about a radio (= slave).
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -102,6 +102,8 @@ private:
     void platform_notify_error(bpl::eErrorCode code, const std::string &error_data);
     bool send_slaves_enable();
 
+    void remove_client_from_all_radios(sMacAddr &client_mac);
+
     std::shared_ptr<bwl::sta_wlan_hal> get_wireless_hal(std::string iface = "");
 
 private:
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -2097,9 +2097,14 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             break;
         }
 
-        notification_out->iface_mac()  = hostap_params.iface_mac;
         notification_out->client_mac() = notification_in->mac();
         notification_out->bssid()      = notification_in->bssid();
+        if (!notification_in->association_frame_length()) {
+            LOG(DEBUG) << "no association frame";
+        } else {
+            notification_out->set_association_frame(notification_in->association_frame(),
+                                                    notification_in->association_frame_length());
+        }
 
         // Send the message
         LOG(DEBUG) << "send ACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION for client "
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -2081,7 +2081,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             return false;
         }
         LOG(TRACE) << "received ACTION_APMANAGER_CLIENT_ASSOCIATED_NOTIFICATION";
-        std::string client_mac = network_utils::mac_to_string(notification_in->params().mac);
+        std::string client_mac = network_utils::mac_to_string(notification_in->mac());
         LOG(INFO) << "client associated sta_mac=" << client_mac;
 
         if (!master_socket) {
@@ -2097,8 +2097,9 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             break;
         }
 
-        notification_out->client_mac() = notification_in->params().mac;
-        notification_out->bssid()      = notification_in->params().bssid;
+        notification_out->iface_mac()  = hostap_params.iface_mac;
+        notification_out->client_mac() = notification_in->mac();
+        notification_out->bssid()      = notification_in->bssid();
 
         // Send the message
         LOG(DEBUG) << "send ACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION for client "
@@ -2118,8 +2119,8 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             LOG(ERROR) << "addClass tlvClientAssociationEvent failed";
             return false;
         }
-        client_association_event_tlv->client_mac() = notification_in->params().mac;
-        client_association_event_tlv->bssid()      = notification_in->params().bssid;
+        client_association_event_tlv->client_mac() = notification_in->mac();
+        client_association_event_tlv->bssid()      = notification_in->bssid();
         client_association_event_tlv->association_event() =
             wfa_map::tlvClientAssociationEvent::CLIENT_HAS_JOINED_THE_BSS;
 
@@ -2135,10 +2136,10 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
                 return false;
             }
 
-            vs_tlv->mac()          = notification_in->params().mac;
-            vs_tlv->bssid()        = notification_in->params().bssid;
-            vs_tlv->vap_id()       = notification_in->params().vap_id;
-            vs_tlv->capabilities() = notification_in->params().capabilities;
+            vs_tlv->mac()          = notification_in->mac();
+            vs_tlv->bssid()        = notification_in->bssid();
+            vs_tlv->vap_id()       = notification_in->vap_id();
+            vs_tlv->capabilities() = notification_in->capabilities();
         }
 
         send_cmdu_to_controller(cmdu_tx);
--- a/common/beerocks/bcl/include/bcl/beerocks_defines.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_defines.h
@@ -40,13 +40,18 @@ static constexpr int DISCOVERY_NOTIFICATION_TIMEOUT_SEC = 60;
 namespace message {
 
 enum eStructsConsts {
-    VERSION_LENGTH                = 16,
-    NODE_NAME_LENGTH              = 32,
-    IFACE_NAME_LENGTH             = 32 + 4, //need extra 1 byte for null termination + alignment
-    SUPPORTED_CHANNELS_LENGTH     = 64,     //support upto # channels, every channel item is 32-bit
-    HOSTAP_ERR_MSG_LENGTH         = 64,
-    WIFI_DRIVER_VER_LENGTH        = 32 + 4,
-    WIFI_SSID_MAX_LENGTH          = 32 + 1 + 3, //need extra 1 byte for null termination + alignment
+    VERSION_LENGTH            = 16,
+    NODE_NAME_LENGTH          = 32,
+    IFACE_NAME_LENGTH         = 32 + 4, //need extra 1 byte for null termination + alignment
+    SUPPORTED_CHANNELS_LENGTH = 64,     //support upto # channels, every channel item is 32-bit
+    HOSTAP_ERR_MSG_LENGTH     = 64,
+    WIFI_DRIVER_VER_LENGTH    = 32 + 4,
+    WIFI_SSID_MAX_LENGTH      = 32 + 1 + 3, //need extra 1 byte for null termination + alignment
+    // The absolute maximum size of any frame according to the 802.11 specification
+    // is MMPDU size of 2304 bytes. The actual size of an (re)association frame
+    // should be in the range of a couple of hundreds of bytes, so be on the safe side
+    // and set the maximum size to 2KB
+    ASSOCIATION_MAX_LENGTH        = 2048,
     WIFI_PASS_MAX_LENGTH          = 64 + 1 + 3, //need extra 1 byte for null termination + alignment
     USER_PASS_LEN                 = 64 + 1 + 3, //need extra 1 byte for null termination + alignment
     DEV_INFO_STR_MAX_LEN          = 32,
--- a/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
@@ -380,9 +380,20 @@ bool ap_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
             LOG(ERROR) << "Failed reading mac parameter!";
             return false;
         }
-
         msg->params.mac = beerocks::net::network_utils::mac_from_string(tmp_str);
+        const char assoc_req[] =
+            "00003A01029A96FB591100504322565F029A96FB591110E431141400000E4D756C74692D41502D3234472D"
+            "31010802040B0C121618242102001430140100000FAC040100000FAC040100000FAC02000032043048606C"
+            "3B10515153547374757677787C7D7E7F80823B160C01020304050C161718191A1B1C1D1E1F202180818246"
+            "057000000000460571505000047F0A04000A82214000408000DD070050F2020001002D1A2D1103FFFF0000"
+            "000000000000000000000000000018E6E10900BF0CB079D133FAFF0C03FAFF0C03C70110DD07506F9A1603"
+            "0103";
 
+        //convert the hex string to binary
+        auto binary_str                      = get_binary_association_frame(assoc_req);
+        msg->params.association_frame_length = binary_str.length();
+
+        std::copy_n(&binary_str[0], binary_str.length(), msg->params.association_frame);
         bool caps_valid = true;
         SRadioCapabilitiesStrings caps_strings;
         if (!dummy_obj_read_str("SupportedRates", parsed_obj, &tmp_str)) {
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -1782,15 +1782,16 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         memset(msg_buff.get(), 0, sizeof(sACTION_APMANAGER_CLIENT_ASSOCIATED_NOTIFICATION));
         memset((char *)&msg->params.capabilities, 0, sizeof(msg->params.capabilities));
 
-        char VAP[SSID_MAX_SIZE]        = {0};
-        char MACAddress[MAC_ADDR_SIZE] = {0};
-        int supported_rates[16]        = {0};
-        int RRM_CAP[8]                 = {0};
-        int HT_MCS[16]                 = {0};
-        int16_t VHT_MCS[16]            = {0};
-        char ht_cap[8]                 = {0};
-        char vht_cap[16]               = {0};
-        size_t numOfValidArgs[20]      = {0};
+        char VAP[SSID_MAX_SIZE]                = {0};
+        char MACAddress[MAC_ADDR_SIZE]         = {0};
+        int supported_rates[16]                = {0};
+        int RRM_CAP[8]                         = {0};
+        int HT_MCS[16]                         = {0};
+        int16_t VHT_MCS[16]                    = {0};
+        char ht_cap[8]                         = {0};
+        char vht_cap[16]                       = {0};
+        size_t numOfValidArgs[20]              = {0};
+        char assoc_req[ASSOCIATION_FRAME_SIZE] = {0};
 
         FieldsToParse fieldsToParse[] = {
             {NULL /*opCode*/, &numOfValidArgs[0], DWPAL_STR_PARAM, NULL, 0},
@@ -1806,6 +1807,8 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
              "VHT_MCS=", sizeof(VHT_MCS)},
             {(void *)ht_cap, &numOfValidArgs[7], DWPAL_STR_PARAM, "HT_CAP=", sizeof(ht_cap)},
             {(void *)vht_cap, &numOfValidArgs[8], DWPAL_STR_PARAM, "VHT_CAP=", sizeof(vht_cap)},
+            {(void *)assoc_req, &numOfValidArgs[9], DWPAL_STR_PARAM,
+             "assoc_req=", sizeof(assoc_req)},
             {(void *)&msg->params.capabilities.btm_supported, &numOfValidArgs[9], DWPAL_CHAR_PARAM,
              "btm_supported=", 0},
             {(void *)&msg->params.capabilities.cell_capa, &numOfValidArgs[10], DWPAL_CHAR_PARAM,
@@ -1857,6 +1860,7 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         LOG(DEBUG) << "max_mcs          : " << (int)msg->params.capabilities.max_mcs;
         LOG(DEBUG) << "max_tx_power     : " << (int)msg->params.capabilities.max_tx_power;
         LOG(DEBUG) << "mumimo_supported : " << (int)msg->params.capabilities.mumimo_supported;
+        LOG(DEBUG) << "assoc_req: " << assoc_req;
 
         for (uint8_t i = 0; i < (sizeof(numOfValidArgs) / sizeof(size_t)); i++) {
             if (numOfValidArgs[i] == 0) {
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -1873,6 +1873,12 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         msg->params.mac          = beerocks::net::network_utils::mac_from_string(MACAddress);
         msg->params.capabilities = {};
 
+        //convert the hex string to binary
+        auto binary_str                      = get_binary_association_frame(assoc_req);
+        msg->params.association_frame_length = binary_str.length();
+
+        std::copy_n(&binary_str[0], binary_str.length(), msg->params.association_frame);
+
         std::string ht_cap_str(ht_cap);
         get_ht_mcs_capabilities(HT_MCS, ht_cap_str, msg->params.capabilities);
 
--- a/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
+++ b/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
@@ -196,6 +196,7 @@ typedef struct {
 
 typedef struct {
     sMacAddr mac;
+    sMacAddr bssid;
     beerocks::message::sRadioCapabilities capabilities;
     int8_t vap_id;
     uint8_t reserved1;
--- a/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
+++ b/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
@@ -201,6 +201,8 @@ typedef struct {
     uint8_t reserved1;
     uint8_t reserved2;
     uint8_t reserved3;
+    size_t association_frame_length;
+    uint8_t association_frame[beerocks::message::ASSOCIATION_MAX_LENGTH];
 } sClientAssociationParams;
 
 typedef struct {
--- a/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
+++ b/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
@@ -263,6 +263,9 @@ typedef struct {
 } sHOSTAP_ENABLED_NOTIFICATION;
 
 #define SSID_MAX_SIZE beerocks::message::WIFI_SSID_MAX_LENGTH
+// ASSOCIATION_MAX_LENGTH is defined to be the maximum of the binary frame
+// but since it is represented as hex here. we need two hex characters per byte + a terminating \0.
+#define ASSOCIATION_FRAME_SIZE (2 * beerocks::message::ASSOCIATION_MAX_LENGTH + 1)
 #define MAC_ADDR_SIZE 18
 #define MAX_SUPPORTED_20M_CHANNELS beerocks::message::SUPPORTED_CHANNELS_LENGTH
 #define MAX_SUPPORTED_CHANNELS                                                                     \
--- a/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
+++ b/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
@@ -130,6 +130,9 @@ cACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION:
   _type: class
   client_mac: sMacAddr
   bssid: sMacAddr
+  association_frame:
+    _type: uint8_t 
+    _length: [] 
 
 cACTION_BACKHAUL_CLIENT_DISCONNECTED_NOTIFICATION:
   _type: class
--- a/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
+++ b/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
@@ -140,12 +140,6 @@ sDfsChannelAvailable:
   bandwidth: uint8_t      #beerocks::eWiFiBandwidth
   vht_center_frequency: uint16_t  
 
-sClientAssociationParams:
-  _type: struct
-  mac: sMacAddr
-  bssid: sMacAddr
-  capabilities: beerocks::message::sRadioCapabilities
-  vap_id: int8_t
 
 sClientDisconnectionParams:
   _type: struct
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -1029,11 +1029,19 @@ bool master_thread::handle_cmdu_1905_client_capability_report_message(
         return false;
     }
 
-    //TODO: log the details so it can be checked in the test_flows
+    //log the details so it can be checked in the test_flows
     LOG(INFO) << "Received CLIENT_CAPABILITY_REPORT_MESSAGE, mid=" << std::hex << int(mid)
               << ", Result Code= " << result_code
               << ", client MAC= " << network_utils::mac_to_string(client_info_tlv->client_mac())
               << ", BSSID= " << network_utils::mac_to_string(client_info_tlv->bssid());
+
+    LOG_IF(client_capability_report_tlv->result_code() ==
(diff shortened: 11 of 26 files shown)


common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1812,15 +1812,16 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         memset(msg_buff.get(), 0, sizeof(sACTION_APMANAGER_CLIENT_ASSOCIATED_NOTIFICATION));
         memset((char *)&msg->params.capabilities, 0, sizeof(msg->params.capabilities));
 
-        char VAP[SSID_MAX_SIZE]        = {0};
```
tomereli, line 1785: Commit message is wrong - we need the (re)association frame, not the (re)authentication frame.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -253,6 +253,7 @@ typedef struct {
 } sHOSTAP_ENABLED_NOTIFICATION;
 
 #define SSID_MAX_SIZE beerocks::message::WIFI_SSID_MAX_LENGTH
+#define ASSOCIATION_FRAME_SIZE 1024
```
tomereli: This *has* to be verified in hostapd source code that this is indeed the maximum size of the association frame, otherwise, we are risking truncating the data here.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -190,6 +190,7 @@ typedef struct {
     uint8_t reserved1;
     uint8_t reserved2;
     uint8_t reserved3;
+    char *association_frame;
```
tomereli: This is not OK, why not use `std::string`? and if it is a `char *`, then also add `size_t association_frame_len`.
Also please change to `uint8_t` so it is aligned to the rest of the file.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1902,6 +1902,8 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         msg->params.vap_id       = beerocks::utils::get_ids_from_iface_string(VAP).vap_id;
         msg->params.mac          = beerocks::net::network_utils::mac_from_string(MACAddress);
         msg->params.capabilities = {};
+        std::copy_n(assoc_req, strnlen(assoc_req, ASSOCIATION_FRAME_SIZE) + 1,
```
tomereli: What's the `+ 1` for? is it for the null terminator? We need to verify in the MultiAP specification if the request should be null terminated or not. As a general rule, all strings are not null terminated in TLVs, take the manufacturer string in the M2 TLV as an example.

framework/tlvf/yaml/tlvf/wfa_map/tlvClientCapabilityReport.yaml
```diff
@@ -10,12 +10,9 @@ tlvClientCapabilityReport:
     _value_const: TLV_CLIENT_CAPABILITY_REPORT
   length: uint16_t
   result_code: eResultCode
-  association_frame_length:
-    _type: uint8_t
-    _length_var: True
   association_frame:
-    _type: uint8_t
-    _length: [ association_frame_length ] 
+    _type: char
```
tomereli: I now understand why you make it a `char` instead of `uint8_t` but you must explain that in the commit message - the reason is that currently TLVF has string APIs support for char types so we can easilly print the association frame using `association_frame_str()` which is autogenerated.
However, as I wrote in the previous commit, we must first understand if the association frame should include the null terminator. You can look at the wireshark capture of a passed test with a certified agent to see if it is null terminated.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1122,6 +1122,15 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
         notification->params().vap_id       = msg->params.vap_id;
         notification->params().bssid        = network_utils::mac_from_string(vap_node->second.mac);
         notification->params().capabilities = msg->params.capabilities;
+        if (!msg->params.association_frame) {
+            LOG(DEBUG) << "no association frame";
+        } else {
+            notification->params().association_frame =
+                new char[strnlen(msg->params.association_frame, ASSOCIATION_FRAME_SIZE) + 1];
```
tomereli: This really looks like a memory leak to me. You allocate a char array - where do you delete it?

tomereli: Also, why do you need to allocate it? you should not do any dynamic allocations when using TLVF, everything is static on the buffer. Please make sure to refactor these parts.

tomereli: Bottom line - you need to change the `sClientAssociationParams::association_frame` from `char *` to an array, and use the same max size used in hostapd.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2079,6 +2079,13 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         notification_out->iface_mac()  = hostap_params.iface_mac;
         notification_out->client_mac() = notification_in->params().mac;
         notification_out->bssid()      = notification_in->params().bssid;
+        if (!notification_in->params().association_frame) {
+            LOG(DEBUG) << "no association frame";
```
tomereli, line 2103: The log here is useless and is repeated throughout the entire packet flow, please remove it. It is enough to print once in the origin of the notification and once in the destination, so BWL and controller.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2079,6 +2079,13 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
...
+        } else {
+            notification_out->set_association_frame(
```
tomereli: This also makes it clear that the association frame should be a `uint8_t []` and not `char []` in the YAML files.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -245,7 +245,8 @@ class backhaul_manager : public btl::transport_socket_thread {
      * Associated client information is later used to fill in the Associated Clients TLV
      * in the Topology Response message.
      */
```
tomereli, line 253: Please update the comment describing the map.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1735,10 +1735,32 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+            auto associated_clients_map = slave.second.associated_clients_map;
+            for (const auto &vap : associated_clients_map) {
+                if (vap.second.find(msg->client_mac()) != vap.second.end()) {
+                    m_radio_info_map.erase(msg->client_mac());
+                }
+            }
+        }
         // Set client association information for associated client
+        std::string assoc_req;
+        if (!msg->association_frame()) {
```
tomereli: Why not simply leave it as empty string? no need for the debug print and if else.
```suggestion
        std::string assoc_req(msg->association_frame() ? msg->association_frame() : "");
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1735,10 +1735,32 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+        } else {
+            std::string assoc_frame(msg->association_frame());
+            assoc_req = assoc_frame;
+        }
         auto &associated_clients =
             m_radio_info_map[msg->iface_mac()].associated_clients_map[msg->bssid()];
-        associated_clients[msg->client_mac()] = std::chrono::steady_clock::now();
+
+        associated_clients[msg->client_mac()] =
+            make_tuple(std::chrono::steady_clock::now(), assoc_req);
```
tomereli: Why use tuple instead of adding a new structure?

```
typedef struct {
    sMacAddr client_mac;
    uint8_t assoc_req[MAX_ASSOC_REQ_SIZE];
    std::chrono::steady_clock::time_point timestamp;
} associated_client_t;

typedef std::unordered_map<sMacAddr, associated_client_t> associated_clients_t;
```

This will make the code much more readable IMO.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1856,11 +1856,15 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx
     //with any of the BSS operated by the Multi-AP Agent [ though the TLV does contain a BSSID, the specification
     // says that we should answer if the client is associated with any BSS on this agent.]
     bool associated_client_found = false;
+    sMacAddr client_ruid;
```
tomereli: Commit message typo: "For that reason the following changes ere made:" --> "For that reason the following changes are made:"

tomereli: So simply add this to the `associated_client_t` struct I suggested in the previous commit:
```
typedef struct {
    sMacAddr client_mac;
    sMacAddr radio_mac; // ruid
    sMacAddr bssid; // VAP mac
    ...
} associated_client_t;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1890,7 +1894,11 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx
     if (associated_client_found) {
         client_capability_report_tlv->result_code() = wfa_map::tlvClientCapabilityReport::SUCCESS;
         LOG(DEBUG) << "Result Code: SUCCESS";
-        //TODO: Add frame body of the most recently received (Re)Association Request frame from this client
+        //Add frame body of the most recently received (Re)Association Request frame from this client
```
tomereli: Add space after `//` (nitpick)

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -376,9 +376,12 @@ bool ap_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
             LOG(ERROR) << "Failed reading mac parameter!";
             return false;
         }
+        msg->params.association_frame = new char[30];
```
tomereli: Commit message should explain why we add a dummy association frame. If you do that as a preparation for a test (and you should) - then mention that.

tomereli (changes requested):
As usual, great job with the commit split and documentation. Your PR description is a sight to behold, I used it as an example for new contributors joining prplMesh :)

However, I have quite a lot of comments, some require big changes so requesting changes for now.
Please also verify this in the testbed so we know that with this we pass the full tests.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
arnout: I believe the absolute maximum (according to 802.11 specification, not hostapd) would be the MMPDU size of 2304 bytes. I think somebody with actual WiFi knowledge should be able to answer that easily.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
arnout: We shouldn't use std::string for a byte array. I would like us to start using `std::basic_string<uint8_t>`, but that would be the first time we use it in prplMesh so let's first get convergence on it.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
morantr: I think `strnlen` is one of the forbidden SDL functions, no?
@tomereli

framework/tlvf/AutoGenerated/include/tlvf/wfa_map/tlvClientCapabilityReport.h
```diff
@@ -22,6 +22,7 @@
 #include <tlvf/ClassList.h>
```
morantr: This commit (5ac1d342585845243a90ea271f7b31458260abce : tlvf: update cliaent caps report TLV auto-generated files) has a typo on the headline.
cliaent ->client

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -146,6 +146,9 @@ sClientAssociationParams:
   bssid: sMacAddr
```
morantr: This commit message headline (ad65f91c5bc80152c1fb29449e2795d4dfc2a58e tlvf: beerocks message common: add association frame to sClientAssoci…) is too long (>72 characteres).

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -146,6 +146,9 @@ sClientAssociationParams:
...
   capabilities: beerocks::message::sRadioCapabilities
   vap_id: int8_t
+  association_frame:
+    _type: char
```
morantr: Why char and not uint8_t?

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
```diff
@@ -125,6 +125,9 @@ cACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION:
   iface_mac: sMacAddr
```
morantr: This commit message (cdbaca92af7d91485d92f7b05bc8b2fc7e768e62: tlvf: beerocks message backhaul: add association frame to CLIENT_ASSO…) is too long.
You can remove "beerocks message backhaul" and just write "tlvf:".

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2093,7 +2093,7 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
                     // the timestamp of its last association.
```
morantr, line 2548: commit message (f72a6ab22aaaf49cbf4dddd9da7d1e0d24d232b7: agent: backhaul manager: update db to save association frame for each…) headline is too long.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -376,9 +376,12 @@ bool ap_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
...
+        msg->params.mac               = beerocks::net::network_utils::mac_from_string(tmp_str);
+        char assoc_req[]              = "dummy association frame";
```
morantr: Not sure why it is required to write `"dummy association frame"` on the tlv, but anyway it probably will better if you use std::string instead of c-string.

morantr (approved):
Besides Tomer's comments that I agree with, I added a few small comments.
Anyway, I'll approve it so you will not be blocked by me.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
tomereli: According to a WiFi expert (WAVE600 FW team leader) the size of the assoc_req is a couple of hundreds of bytes at the most. It is true that the MMPDU is the absolute maximum, so we can set it to a nice round number like 2048 bytes and add a comment saying why we chose this number.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Actually it is not, `strlen` is forbidden - see https://github.com/intel/safestringlib/wiki/SDL-List-of-Banned-Functions

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: I have tried to keep the code uniform since we use tuple most of the time, but I like your idea 👍

agent/src/beerocks/slave/son_slave_thread.cpp
CoralMalachi: I don't think it useless since if the association frame is empty it means something is wrong, and it will be printed only if something wrong..

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: I have added the +1 for the null terminator to get a result from the `strnlen` method, but when I'm filling the tlvClientCapabilityReport I copy it without the null terminator. @tomereli

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: So change to `LOG_IF(!notification_in->params().association_frame, ERROR) << "no association frame";`

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
CoralMalachi: Yes I know it was not ready yet :)

framework/tlvf/yaml/tlvf/wfa_map/tlvClientCapabilityReport.yaml
arnout: > so we can easilly print the association frame using `association_frame_str()`

The association frame is binary data. Printing it as a string will not lead to much joy.

I believe that what you parse out of the AP-STA-CONNECTED event from hostapd is the hex representation of the frame. So it should be converted back to binary (uint8_t) inside bwl and passed as binary in the Client Capability Report TLV.

As Tomer says, check with a certified device what the contents is, and make sure we generate the same thing. Actually, it would be nice if you could upload a pcap file with this TLV as generated by a certified device, so everybody can easily compare with it.

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -43,6 +43,7 @@ enum eStructsConsts {
     HOSTAP_ERR_MSG_LENGTH         = 64,
     WIFI_DRIVER_VER_LENGTH        = 32 + 4,
     WIFI_SSID_MAX_LENGTH          = 32 + 1 + 3, //need extra 1 byte for null termination + alignment
+    ASSOCIATION_MAX_LENGTH        = 2048,
```
arnout, line 54: I really would like some explanation (in the commit message or as a comment) about how you get this number.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1812,15 +1812,16 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
...
+        char VAP[SSID_MAX_SIZE]                = {0};
+        char MACAddress[MAC_ADDR_SIZE]         = {0};
+        int supported_rates[16]                = {0};
+        int RRM_CAP[8]                         = {0};
+        int HT_MCS[16]                         = {0};
+        int16_t VHT_MCS[16]                    = {0};
+        char ht_cap[8]                         = {0};
+        char vht_cap[16]                       = {0};
+        size_t numOfValidArgs[20]              = {0};
+        char assoc_req[ASSOCIATION_FRAME_SIZE] = {0};
```
arnout, line 1794: Note that *here*, char is the correct type.

However, this size should be `2*ASSOCIATION_FRAME_SIZE+1` to fix the hex representation + the terminating \0.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -252,6 +252,11 @@ typedef struct {
 } sHOSTAP_ENABLED_NOTIFICATION;
 
 #define SSID_MAX_SIZE beerocks::message::WIFI_SSID_MAX_LENGTH
+// The absolute maximum size of any frame according to the 802.11 specification
+// is MMPDU size of 2304 bytes. The actual size of an (re)association frame
+// should be in the range of a couple of hundreds of bytes, so be on the safe side
+// and set the maximum size to 2KB
```
arnout: Move this explanation to the place where the number 2048 is used.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1902,6 +1902,8 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         msg->params.vap_id       = beerocks::utils::get_ids_from_iface_string(VAP).vap_id;
         msg->params.mac          = beerocks::net::network_utils::mac_from_string(MACAddress);
         msg->params.capabilities = {};
+        std::copy_n(assoc_req, strnlen(assoc_req, ASSOCIATION_FRAME_SIZE) + 1,
+                    msg->params.association_frame);
```
arnout: As I note elsewhere: you shouldn't simply copy, but convert to binary.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -146,6 +146,9 @@ sClientAssociationParams:
...
+    _length: [ "beerocks::message::ASSOCIATION_MAX_LENGTH" ]
```
arnout: You really should use a dynamic length here, because there is no 0 terminator to indicate the length. But then it can't be a struct any more :-( But actually, this struct really serves no purpose at all, it can be inlined directly into cACTION_APMANAGER_CLIENT_ASSOCIATED_NOTIFICATION. So please do that (in a separate commit) and then add the association frame as a dynamic length field.

Note that in the 1905 message it's an unknown length field, but that's not possible here because we don't have a total length in the beerocks classes.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2105,6 +2105,14 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
...
         notification_out->client_mac() = notification_in->params().mac;
         notification_out->bssid()      = notification_in->params().bssid;
+        if (!notification_in->params().association_frame) {
+            LOG(DEBUG) << "no association frame";
+            notification_out->set_association_frame("");
+        } else {
+            notification_out->set_association_frame(
+                notification_in->params().association_frame,
+                strnlen(notification_in->params().association_frame, ASSOCIATION_FRAME_SIZE) + 1);
+        }
```
arnout, line 2107: Meh I hate all this useless copying around of notification messages...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1742,10 +1742,25 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
             return false;
         }
 
+        //remove this client from other radios
+        for (const auto &slave : m_radio_info_map) {
+            auto associated_clients_map = slave.second.associated_clients_map;
+            for (const auto &vap : associated_clients_map) {
```
arnout: vap? These are associated clients, right? So `client` would be a better name.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -379,9 +379,11 @@ bool ap_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
             LOG(ERROR) << "Failed reading mac parameter!";
             return false;
         }
+        msg->params.mac  = beerocks::net::network_utils::mac_from_string(tmp_str);
+        char assoc_req[] = "dummy association frame";
```
arnout: I think it would be better to include a real association frame here, e.g. something that you got from sniffing actual easymesh TLVs from a certified device.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1022,6 +1022,10 @@ bool master_thread::handle_cmdu_1905_client_capability_report_message(
               << ", Result Code= " << result_code
               << ", client MAC= " << network_utils::mac_to_string(client_info_tlv->client_mac())
               << ", BSSID= " << network_utils::mac_to_string(client_info_tlv->bssid());
+
+    LOG_IF(!result_code.compare("SUCCESS"), DEBUG)
```
arnout: (nitpick) `result_code == "SUCCESS"` is more natural to me.

Or actually, `client_capability_report_tlv->result_code() == wfa_map::tlvClientCapabilityReport::SUCCESS` is even better.

tests/test_flows.sh
```diff
@@ -197,6 +197,7 @@ test_client_capability_query() {
     dbg "Confirming client capability report message has been received on controller"
     check_log ${GATEWAY} controller "Received CLIENT_CAPABILITY_REPORT_MESSAGE"
     check_log ${GATEWAY} controller "Result Code= SUCCESS, client MAC= ${sta_mac2}, BSSID= ${mac_agent1_wlan0}"
+    check_log ${GATEWAY} controller "(Re)Association Request frame= dummy association frame"
```
arnout: Please update test_flows.py instead.

arnout (approved):
The confusion between bytes and hex is a major concern, but can be fixed easily.

common/beerocks/bcl/include/bcl/beerocks_defines.h
tomereli: I was sure I explained this in the commit which changed this originally.
I don't remember the maximum size - MPDU length, I think its ~2300 Bytes.
According to the FW guys, the association frame is normally a couple of hundreds of bytes so 2KB should be well enough.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
tomereli: Oh here it is!

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Shouldn't really be a nitpick - its in our CONTRIBUTING.md AFAIR.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
CoralMalachi: it used here - ASSOCIATION_MAX_LENGTH is defined to 2048 in beerocks_defines.h, do you want me to move this comment there?

common/beerocks/bcl/include/bcl/beerocks_defines.h
CoralMalachi: I thought it would be nicer to add a constant variable that is set to the desired size. I can move your explained to the place where ASSOCIATION_MAX_LENGTH  is defined (beerocks_defines.h). @tomereli :)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: why 2 times ASSOCIATION_FRAME_SIZE?

agent/src/beerocks/slave/son_slave_thread.cpp
CoralMalachi: Changed it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: Done 👍

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: I changed it but did not push it yet since I want to validate it works on testbed first (but I can't since the master branch using it now).

Currently, I'm facing an issue in which the test sometimes passes and sometimes not, without changing a thing (on a side branch).
I'll try to convert as you suggested in the ap_wlan_hal_dwpal.cpp (currently I'm converting it only before sending it as part of the client capability report message and maybe it's wrong..)

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: No, it's literally useless, because `association_frame` is an array, not a pointer:

```cpp
#include <iostream>

struct s {
    char foo[10];
};

int main()
{
    s s1;
    s1.foo[0] = '\0';
    if (!s1.foo)
        std::cout << "foo is empty" << std::endl;
    else
        std::cout << "foo is not empty" << std::endl;
}
```

The code above will print "foo is not empty".

It *could* be useful to say `if (notification_in->params().association_frame[0] != '\0')`.

However, as I wrote elsewhere, it should be a binary buffer, not a hex string. IMO.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: Because the 2048 was assumed to be the maximum of the *binary* frame, but it's represented as hex here. So you need two hex characters per byte + a terminating \0.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
arnout: Yes, that's what we mean.

The comment explains why the value 2048 is chosen, but that value is not defined here, it's defined in beerocks_defines.h.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
CoralMalachi: done.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
CoralMalachi: done

common/beerocks/bcl/include/bcl/beerocks_defines.h
tomereli: Please also add the explanation to the commit message, don't leave it empty.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -190,6 +190,7 @@ typedef struct {
     uint8_t reserved1;
     uint8_t reserved2;
     uint8_t reserved3;
+    char association_frame[beerocks::message::ASSOCIATION_MAX_LENGTH];
```
tomereli: Better to use `uint8_t` which makes it clear that this is binary data.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1902,6 +1902,10 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         msg->params.vap_id       = beerocks::utils::get_ids_from_iface_string(VAP).vap_id;
```
tomereli: Commit message "the assoc_req we get from the DWPAL the whole message" --> "the assoc_req we get from the DWPAL **contains** the whole message.
Please also add an explanation in the commit message and in the code - nice big comment - with reference to where in the specifications it says that the association request should be only the tagged parameters, that is if you know it. If not - perhaps @arnout does?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1902,6 +1902,10 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
...
         msg->params.mac          = beerocks::net::network_utils::mac_from_string(MACAddress);
         msg->params.capabilities = {};
+        auto assoc_frame_len     = strnlen(assoc_req, ASSOCIATION_FRAME_SIZE);
+        auto sub_str             = std::string(&assoc_req[56], assoc_frame_len - 56 - 18);
```
tomereli: Don't use magic numbers, please.  Why 56? why 18? These needs to be defined or at least documented in a comment with a TODO perhaps saying how this should be done properly.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1902,6 +1902,10 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
...
+
+        std::copy_n(&sub_str[0], sub_str.length() + 1, msg->params.association_frame);
```
tomereli: why `&sub_str[0]` and not just `sub_str`? Why the extra `1`? null termination is not required IMO.

framework/tlvf/yaml/tlvf/wfa_map/tlvClientCapabilityReport.yaml
```diff
@@ -10,12 +10,9 @@ tlvClientCapabilityReport:
     _value_const: TLV_CLIENT_CAPABILITY_REPORT
   length: uint16_t
   result_code: eResultCode
-  association_frame_length:
-    _type: uint8_t
-    _length_var: True
   association_frame:
     _type: uint8_t
-    _length: [ association_frame_length ] 
+    _length: []
```
tomereli, line 15: Commit message is wrong - the type is `uint8_t` (as it should be) and not `char`.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
tomereli: +1. we should use `uint8_t` everywhere now for the association frame, cause its binary data.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
```diff
@@ -125,6 +125,9 @@ cACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION:
...
   client_mac: sMacAddr
   bssid: sMacAddr
+  association_frame:
+    _type: char
```
tomereli: type should be `uint8_t`.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
```diff
@@ -125,6 +125,9 @@ cACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION:
   iface_mac: sMacAddr
   client_mac: sMacAddr
   bssid: sMacAddr
+  association_frame:
+    _type: char 
+    _length: []
```
tomereli, line 135: type should be `uint8_t`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -246,7 +246,8 @@ class backhaul_manager : public btl::transport_socket_thread {
...
-    typedef std::unordered_map<sMacAddr, std::chrono::steady_clock::time_point>
+    typedef std::unordered_map<sMacAddr,
+                               std::tuple<std::chrono::steady_clock::time_point, std::string>>
         associated_clients_t;
```
tomereli:
```suggestion

struct sClientInfo {
    sMacAddr client_mac;
    std::chrono::steady_clock::time_point;
    uint8_t assoc_req[MAX_ASSOC_REQ_LEN];
};

typedef std::unordered_map<sMacAddr, sClientInfo> associated_clients_t;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1742,10 +1742,25 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
             return false;
         }
 
+        //remove this client from other radios
```
tomereli, line 2036: Please move to a private helper function (like I did here https://github.com/prplfoundation/prplMesh/pull/1067/commits/f52ed42dee312e662cdfac61c114a3178ff1c018)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1742,10 +1742,25 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
+        for (const auto &slave : m_radio_info_map) {
+            auto associated_clients_map = slave.second.associated_clients_map;
+            for (const auto &vap : associated_clients_map) {
+                if (vap.second.find(msg->client_mac()) != vap.second.end()) {
+                    m_radio_info_map.erase(msg->client_mac());
+                }
+            }
+        }
         // Set client association information for associated client
+        std::string assoc_req(msg->association_frame() ? msg->association_frame() : "");
```
tomereli: Should be `uint8_t`

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: At least I see now that I'm consistent :)
So will you change it?

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -379,9 +379,19 @@ bool ap_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
             LOG(ERROR) << "Failed reading mac parameter!";
             return false;
         }
-
         msg->params.mac = beerocks::net::network_utils::mac_from_string(tmp_str);
-
+        char assoc_req[] =
```
tomereli: `const char`

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -379,9 +379,19 @@ bool ap_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
...
+            "00003A01029A96FB591100504322565F029A96FB591110E431141400000E4D756C74692D41502D3234472D"
+            "31010802040B0C121618242102001430140100000FAC040100000FAC040100000FAC02000032043048606C"
+            "3B10515153547374757677787C7D7E7F80823B160C01020304050C161718191A1B1C1D1E1F202180818246"
+            "057000000000460571505000047F0A04000A82214000408000DD070050F2020001002D1A2D1103FFFF0000"
+            "000000000000000000000000000018E6E10900BF0CB079D133FAFF0C03FAFF0C03C70110DD07506F9A1603"
+            "0103";
+
+        auto assoc_req_len = strnlen(assoc_req, ASSOCIATION_FRAME_SIZE);
+        auto sub_str       = std::string(&assoc_req[56], assoc_req_len - 56 - 18);
```
tomereli: Same thing - please no magic numbers.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1836,6 +1837,8 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
              "VHT_MCS=", sizeof(VHT_MCS)},
             {(void *)ht_cap, &numOfValidArgs[7], DWPAL_STR_PARAM, "HT_CAP=", sizeof(ht_cap)},
             {(void *)vht_cap, &numOfValidArgs[8], DWPAL_STR_PARAM, "VHT_CAP=", sizeof(vht_cap)},
+            {(void *)assoc_req, &numOfValidArgs[9], DWPAL_STR_PARAM,
```
tomereli, line 1810: Maybe it is better to do the conversion to the offset and size we need already here (nitpick)

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1022,6 +1022,13 @@ bool master_thread::handle_cmdu_1905_client_capability_report_message(
               << ", Result Code= " << result_code
               << ", client MAC= " << network_utils::mac_to_string(client_info_tlv->client_mac())
               << ", BSSID= " << network_utils::mac_to_string(client_info_tlv->bssid());
+
+    //TODO: convert the uint8_t array to ascii format so we can compare
+    LOG_IF(client_capability_report_tlv->result_code() ==
+               wfa_map::tlvClientCapabilityReport::SUCCESS,
+           DEBUG)
+        << "(Re)Association Request frame= ";
+
```
tomereli, line 1044: You can just use `utils::dump_buffer` - no need to convert to ascii.

tomereli (changes requested):
There is a big mess in the commits. The assoc_req needs to be uint8_t - so asking changes for now.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: I explained it in the commit message... I'll add a comment in the code as well.

controller/src/beerocks/master/son_master_thread.cpp
CoralMalachi: thanks

tests/test_flows.sh
CoralMalachi: done

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: fixed the commit message.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
CoralMalachi: done

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
CoralMalachi: done.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: If I pass just `sub_str` it does not compile...
about the extra byte, I removed it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: done 👍

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2105,11 +2119,18 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx
     //with any of the BSS operated by the Multi-AP Agent [ though the TLV does contain a BSSID, the specification
     // says that we should answer if the client is associated with any BSS on this agent.]
     bool associated_client_found = false;
+    sClientInfo s;
+    // sMacAddr client_ruid;
+    // sMacAddr client_vap;
     for (const auto &slave : m_radio_info_map) {
         auto associated_clients_map = slave.second.associated_clients_map;
```
```suggestion
        const auto &associated_clients_map = slave.second.associated_clients_map;
```

This was already wrong before this PR, but I notice it now...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2135,18 +2156,35 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx
...
+        LOG(DEBUG) << "******************************************************************";
+        LOG(DEBUG) << "          sAssociatedClients.asso_len / 2 = " << s.asso_len / 2;
+        LOG(DEBUG) << "          assoc_req.len = " << assoc_req.length() / 2;
+        LOG(DEBUG) << "          assoc_frame_body.len = " << assoc_frame_body.length();
+
+        LOG(DEBUG) << "******************************************************************";
+        client_capability_report_tlv->alloc_association_frame(s.asso_len / 2);
+        std::copy(&assoc_frame_body[0], &assoc_frame_body[0] + s.asso_len / 2,
+                  client_capability_report_tlv->association_frame());
+
```
arnout: This whole thing is wrong, no string conversion is needed here.

```suggestion
        client_capability_report_tlv->alloc_association_frame(s.asso_len);
        std::copy_n(s.assoc_req, s.asso_len, client_capability_report_tlv->association_frame());
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -242,21 +244,33 @@ class backhaul_manager : public btl::transport_socket_thread {
 
     std::unique_ptr<beerocks::agent_ucc_listener> m_agent_ucc_listener;
 
+    struct sClientInfo {
+        sMacAddr client_mac;
+        sMacAddr radio_mac; // ruid
+        sMacAddr bssid;     // VAP mac
+        std::chrono::steady_clock::time_point time_stamp;
+        size_t asso_len;
+        uint8_t assoc_req[ASSOCIATION_FRAME_SIZE];
```
arnout, line 235:
```suggestion
        std::array<uint8_t> assoc_req;
```

If you see a size and array combination in any structure that is not tlv-generated, it's probably wrong :-). Actually, if you see a C-style array it's probably wrong.

Hm, maybe I need to make a CONTRIBUTING.md update...

Note that with the std::array, the copy is simply `std::copy(s.assoc_req, client_capability_report_tlv->association_frame())`.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -380,9 +380,27 @@ bool ap_wlan_hal_dummy::process_dummy_event(parsed_obj_map_t &parsed_obj)
...
+        msg->params.association_frame_length = sub_str.length();
+
+        LOG(DEBUG)
+            << "*********************************************************************************";
+        LOG(DEBUG) << "msg->params.association_frame_length= "
+                   << msg->params.association_frame_length;
+        LOG(DEBUG)
+            << "*********************************************************************************";
+
+        std::copy_n(&sub_str[0], sub_str.length(), msg->params.association_frame);
```
arnout: The association_frame in the output is binary, not hex, so you have to convert the hex string to binary first.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: It should still be defined with static const variables.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1868,6 +1872,11 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         msg->params.vap_id       = beerocks::utils::get_ids_from_iface_string(VAP).vap_id;
         msg->params.mac          = beerocks::net::network_utils::mac_from_string(MACAddress);
         msg->params.capabilities = {};
+        auto assoc_frame_len     = strnlen(assoc_req, ASSOCIATION_FRAME_SIZE);
+        auto sub_str             = std::string(&assoc_req[56], assoc_frame_len - 56 - 18);
+        msg->params.association_frame_length = sub_str.length();
+
+        std::copy_n(&sub_str[0], sub_str.length() + 1, msg->params.association_frame);
```
arnout: Again, it should already have been converted to binary here.

I realize I said before that you could choose to pass it around as string internally, but really there is no reason to do that, so let's just do the right thing and convert to binary here.

tests/test_flows.py
```diff
@@ -466,6 +479,9 @@ def test_client_capability_query(self):
         self.check_log(env.controller,
                        r"Result Code= SUCCESS, client MAC= {}, BSSID= {}"
                        .format(sta2.mac, env.agents[0].radios[0].mac))
+               
+        for line in association_frame.splitlines():
+            self.check_log(env.controller, r"{}".format(line))
```
arnout, line 525: Can you use the start_line feature that was recently added to check_log to make sure that you're actually matching the right part of the log? The previous check_log call just before the loop will give you the starting line.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
CoralMalachi: sorry, the branch was not updated, that's why I named the commit message WIP :)
I've pushed now the side branch, and I'll go over all your comments @arnout, thanks.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: Ok 👍 @arnout

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
CoralMalachi: I agree it does not make sense to use size and array combination inside a structure.
The reason I did it is that I need to know the length of each association from so when I pass it in the client capability report it will have the correct length.

One option I thought about is to use a vector, which is dynamic but it isn't very comfortable to use (to push every time another `uint8_t`).
Another is to declare it as c array and to fill it with a null terminator (as we did in the ap_wlan_hal Event::STA_Connected) and then we can use `strnlen` after converting to string.

but `std::array` also requires a length argument so it's also not a dynamic array.
I want to know the length of each association frame and not take the fixed size, so how does `std::array` help me here? let me know if I did not understand your suggestion @arnout

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
CoralMalachi: done.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
CoralMalachi: done.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
arnout: OK, I was mixing things up a little, sorry about that...

`uint8_t assoc_req[ASSOCIATION_FRAME_SIZE];` is equivalent to `std::array<uint8_t, ASSOCIATION_FRAME_SIZE> assoc_req;` - but the latter would still require asso_len. So in this case, it should indeed be a `std::vector`.

However, as you observe, a vector is not very comfortable, because we later want to copy it out with set_association_frame(). Well, actually, it's not that bad, because we can use `std::copy_n` to fill in the vector, and `std::vector::data()` to access the data directly for set_association_frame().

However, as I indicated before, I think it's better to use a `basic_string<uint8_t>` for such a situation.

I'm working on a quick commit to show what I mean.

arnout: I pushed my proposal to dev/client-caps-report-assoc_req-with-bytes (additional fixup commits on top of this PR). If it looks good to @CoralMalachi and @tomereli just take those commits in this PR and rebase on master. I didn't create a PR for that branch.

Note that this is not terribly important because we already use this C-style array pattern in dozens of places in the code, so adding one more is not a huge concern. And we really should get this PR merged soonish, it's been dragging on for too long..

tomereli: I like it. I basic_string is a weird name for storing binary data so using a bytes class is welcome. I now understand why @mariomaz used std::basic_string in his PR - so this should probably be pushed as a separate PR so we can merge it quickly - @arnout can you do that so @mariomaz can rebase his PR and use the `bytes` class?

tests/test_flows.py
CoralMalachi: Do you mean instead of the loop or just before that? I don't think I understood the purpose of this method.

tomereli (approved):

tests/test_flows.py
arnout: Never mind, I'll just rework the startline functionality entirely at some point.

PR #891, opened 2020-02-26
auto-config received after the backhaul is operational

### LINKED ISSUE #886: [BUG] 4.8.1: slave not operational after slave_reset when broadcom device is the controller
The test plan specifies the Broadcom device should be used as the controller for the test`MAP-4.8.1_ETH_FH24G`.

When we run the test with the Broadcom controller, the slave for wlan0 does not become operational after the slave_reset, which causes the test to fail.

If we use the Marvell device as the controller for example, the problem does not occur.

Example of such a test using the Broadcom controller: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/445623302

rmelotte:
This is a quick attempt at fixing https://github.com/prplfoundation/prplMesh/issues/886

Although it fixes the particular issue we were facing in using the Broadcom controller as part of the test `MAP-4.8.1_ETH_FH24G`, I'd like to know more about the implication of the change in commit `3072fe02`, so please comment on it.

In particular:
- do you see any potential issue with this specific change
- why can't we call `finalize_slaves_connect_state(true, soc)` unconditionally when receiving an autoconfig-reply?

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -773,13 +773,9 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
     case EState::CONNECTED: {
 
         stop_on_failure_attempts = configuration_stop_on_failure_attempts;
-        LOG(DEBUG) << "Sending notifications...";
 
         LOG(DEBUG) << "clearing blacklist";
         ap_blacklist.clear();
-
-        finalize_slaves_connect_state(true);
-        LOG(DEBUG) << "Sent notifications, goto OPERATIONAL state";
 
         eth_link_poll_timer = std::chrono::steady_clock::now();
         m_eth_link_up       = network_utils::linux_iface_is_up_and_running(m_sConfig.wire_iface);
@@ -803,7 +799,6 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
             }
         } else if (pending_enable && m_sConfig.eType != SBackhaulConfig::EType::Invalid) {
             pending_enable = false;
-            finalize_slaves_connect_state(true);
         }
 
         break;
@@ -2225,12 +2220,7 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
                     LOG(DEBUG) << FSM_CURR_STATE_STR;
                     soc->controller_discovered = true;
                     controller_bridge_mac      = src_mac;
-                    // the backhaul was operational on slave registration.
-                    // This means that it was restarted and sending backhaul connected event
-                    // is required so that the slave will re-join the controller.
-                    if (soc->operational_on_registration) {
-                        finalize_slaves_connect_state(true, soc);
-                    }
+                    finalize_slaves_connect_state(true, soc);
                 }
             }
         }
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2197,54 +2197,44 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
     auto tlvSupportedFreqBand = cmdu_rx.getClass<ieee1905_1::tlvSupportedFreqBand>();
     if (tlvSupportedFreqBand) {
         LOG(DEBUG) << "tlvSupportedFreqBand->value()=" << int(tlvSupportedFreqBand->value());
-
-        if (tlvSupportedFreqBand->value() == ieee1905_1::tlvSupportedFreqBand::BAND_2_4G) {
-            LOG(DEBUG) << "received auto-config reply for 2.4GHz band";
-            for (auto soc : slaves_sockets) {
-                if (soc->freq_type == beerocks::eFreqType::FREQ_24G) {
-                    LOG(DEBUG) << "2.4GHz band socket found, iface=" << soc->hostap_iface;
-                    if (!soc->controller_discovered) {
-                        LOG(DEBUG) << "set controller_discovered to true for 2.4GHz band, iface="
-                                   << soc->hostap_iface;
-                        LOG(DEBUG) << FSM_CURR_STATE_STR;
-                        soc->controller_discovered = true;
-                        controller_bridge_mac      = src_mac;
-                        // the backhaul was operational on slave registration.
-                        // This means that it was restarted and sending backhaul connected event
-                        // is required so that the slave will re-join the controller.
-                        if (soc->operational_on_registration) {
-                            finalize_slaves_connect_state(true, soc);
-                        }
-                    }
-                }
-            }
-        } else if (tlvSupportedFreqBand->value() == ieee1905_1::tlvSupportedFreqBand::BAND_5G) {
-            LOG(DEBUG) << "received auto-config reply for 5GHz band";
-            for (auto soc : slaves_sockets) {
-                if (soc->freq_type == beerocks::eFreqType::FREQ_5G) {
-                    LOG(DEBUG) << "5GHz band socket found, iface=" << soc->hostap_iface;
-                    if (!soc->controller_discovered) {
-                        LOG(DEBUG) << "set controller_discovered to true for 5GHz band, iface="
-                                   << soc->hostap_iface;
-                        LOG(DEBUG) << FSM_CURR_STATE_STR;
-                        soc->controller_discovered = true;
-                        controller_bridge_mac      = src_mac;
-                        // the backhaul was operational on slave registration.
-                        // This means that it was restarted and sending backhaul connected event
-                        // is required so that the slave will re-join the controller.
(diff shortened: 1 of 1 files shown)


rmelotte:
There is a sample 4.8.1 test here (with the broadcom controller): https://gitlab.com/prpl-foundation/prplMesh/-/jobs/451251833

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -2276,22 +2276,25 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
...
+             beerocks::message::IFACE_NAME_LENGTH},
             {(void *)MACAddress, &numOfValidArgs[2], DWPAL_STR_PARAM, NULL, sizeof(MACAddress)},
             {(void *)&status_code, &numOfValidArgs[3], DWPAL_INT_PARAM, "status_code=", 0},
             /* Must be at the end */
             {NULL, NULL, DWPAL_NUM_OF_PARSING_TYPES, NULL, 0}};
 
-        if (dwpal_string_to_struct_parse(buffer, bufLen, fieldsToParse, sizeof(MACAddress)) ==
-            DWPAL_FAILURE) {
+        if (dwpal_string_to_struct_parse(buffer, bufLen, fieldsToParse,
+                                         sizeof(fieldsToParse)) == DWPAL_FAILURE) {
```
morantr: The last argument defined on `dwpal_string_to_struct_parse` from dwpal lib is defined with the following description:
`size_t userBufLen - The user total buffer size to hold parsed info from inside fieldsToParse`

This means it should have the sum of the sizes of vap_name, MACAddress, status_code.
I saw this example on dwpal code:
image

I have no idea why it is implemented like this, but as I can see we use this argument wrong on all dwpal code.
@kantera800 @tomereli

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -2305,6 +2308,10 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
 
         msg->params.mac         = beerocks::net::network_utils::mac_from_string(MACAddress);
         msg->params.status_code = status_code;
+        std::string mac;
+        beerocks::net::network_utils::linux_iface_get_mac(vap_name, mac);
+        LOG(DEBUG) << "BTM response source BSSID: " << mac;
+        msg->params.source_bssid = beerocks::net::network_utils::mac_from_string(mac);
```
morantr: You added to the struct also target_bssid, but I didn't see you fill it, also on other bwl hals.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2225,10 +2225,11 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
                     LOG(DEBUG) << FSM_CURR_STATE_STR;
                     soc->controller_discovered = true;
                     controller_bridge_mac      = src_mac;
-                    // the backhaul was operational on slave registration.
+                    // the backhaul was operational on slave registration or changed
+                    // to operational since the registration.
                     // This means that it was restarted and sending backhaul connected event
                     // is required so that the slave will re-join the controller.
-                    if (soc->operational_on_registration) {
+                    if (soc->operational_on_registration || FSM_IS_IN_STATE(OPERATIONAL)) {
```
morantr: This change reverts commit https://github.com/prplfoundation/prplMesh/commit/17ad041b5eb40b790d08d0e8ba225fe829d134e0.

The reason we shouldn't do it is that once we got the first discovery response message we would move to an operational state and send the "backhaul connected" message to all son_slave which are discovered.
This means that after we will get the second discovery response we would get inside the "finalize" function and send this message again. This means that the previous son_slave which we got the discovery response for, will receive a "backhaul connected" message twice.

morantr (dismissed):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
rmelotte: Ok, I didn't notice I was actually reverting a previous change.
Thanks for the explanation.
And what if we only sent the "backhaul connected" message only when we have the autoconfig response and not when moving to the operational state? Does it make any sense? 
I'll push the change to this branch so that you can see what I mean.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
arnout: If you look to the piece of INTERCAL code that is dwpal, you'll see that the `field` member of the fieldsToParse argument is assumed to point to a struct member of the first element of an array-of-structs, and an intricate way to avoid using the offsetof operator. The userBufLen argument is supposed to be th total size of that array-of-structs. So indeed, if you're not using an array-of-structs, you have to manually add up the size of the different fields.

arnout:
It took me a long time to understand what is going on, but I definitely like the direction in which this is going. Actually, I'd make a few additional modifications.

- Further refactor the finalize_slaves_connect_state, split it into two functions for the connect and disconnect branches. They're completely disjoint. Actually, since each function is called only once, they could be inlined, but they're rather big so let's keep them as functions.
- After the last commit, pSocket will always be non-nullptr, so the looping over the different sockets becomes redundant. I would then move the loop over socs that is currently in handle_1905_autoconfiguration_response down into finalize_slaves_connect_state, and just pass the freq_type as argument instead of pSocket.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: I saw the same today with Denis and we also didn't get why we pass a partial size.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: >This means that the previous son_slave which we got the discovery response for, will receive a "backhaul connected" message twice.

So? The slave can simply ignore the second message. This is merely an optimization, and this PR proves it was wrong since it didn't cover all cases.

I think that instead of letting Raphael struggle with a new solution we should revert the optimization and be done with it.

@arnout @vitalybu what do you think?

tomereli:
Definitely looks good to me, you can change it from RFC to a normal PR and IMO you can just drop the WIP commit so we can merge it tomorrow and pass the test officially before the arch call.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: This 

> > This means that the previous son_slave which we got the discovery response for, will receive a "backhaul connected" message twice.
> 
> So? The slave can simply ignore the second message. This is merely an optimization, and this PR proves it was wrong since it didn't cover all cases.
> 
> I think that instead of letting Raphael struggle with a new solution we should revert the optimization and be done with it.
> 
> @arnout @vitalybu what do you think?

This is actually bad because when the son_slave will receive this message it will change its state to a different state and send messages it already sent.
Don't forget, we both did this because we noticed it causing a mess on the son_slave and the logs too.
BTW, this is one of the reasons we need unified Agent and not two independent "son_slaves".

tomereli: So we should add a check for that in the son slave - if it is already operational it can ignore the second message.
I know that this is probably not the best solution so if you have another one please speak up, we need this resolved **today**

rmelotte: @tomereli @morantr could you comment on commit ce621129ae9adf2ff0acc822bd544e0adf291206 ? 
Why not call the finalize function when receiving the autoconfig reply only?
It avoids the duplicate M1 and shouldn't put the slave in a wrong state as far as I understand.

morantr: I don't remember why we did it like this, but your suggestion might be a good idea.
I think that the logic behinds it is not to hold back auto-config flow if the other radio hasn't been answered, but basically I don't see any reason why would it happen.

rmelotte:
> * Further refactor the finalize_slaves_connect_state, split it into two functions for the connect and disconnect branches. They're completely disjoint. Actually, since each function is called only once, they could be inlined, but they're rather big so let's keep them as functions.
> 
> * After the last commit, pSock

Since we need to fix this issue ASAP, I opened https://github.com/prplfoundation/prplMesh/issues/911 with your additional suggestions.

rmelotte:
Rebased on master, squashed the commits and changed the commit messages.
A new `MAP-4.8.1_ETH_FH24G` test (broadcom controller) has been run here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/459082965 (passed)

arnout (approved):

tomereli (approved):

morantr (approved):

PR #897, opened 2020-02-27
Dev/client steering optimization

kantera800:
This PR fixes client steering bugs and performance issues.

Found:
while performing client steering basic validation on RDKB platform.

- [x] Disassociate timer is too long– change default from 15sec to 200ms (use in all cases in optimal path task)

- [x] client_steering_task wrong “success” criteria – add check if we steered to the requested bssid.

- [x] [bug] Disallow message not sent  – controller should send VAP mac instead of AP mac in bssid field . disallow message should be sent per VAP.

### DIFF
--- a/controller/src/beerocks/master/tasks/client_steering_task.cpp
+++ b/controller/src/beerocks/master/tasks/client_steering_task.cpp
@@ -23,11 +23,11 @@ using namespace son;
 
 client_steering_task::client_steering_task(db &database_, ieee1905_1::CmduMessageTx &cmdu_tx_,
                                            task_pool &tasks_, std::string sta_mac_,
-                                           std::string hostap_mac_, bool disassoc_imminent_,
+                                           std::string target_bssid_, bool disassoc_imminent_,
                                            int disassoc_timer_ms_, bool steer_restricted_,
                                            std::string task_name_)
     : task(task_name_), database(database_), cmdu_tx(cmdu_tx_), tasks(tasks_), sta_mac(sta_mac_),
-      hostap_mac(hostap_mac_), //Chosen VAP BSSID to steer the client to
+      target_bssid(target_bssid_), //Chosen VAP BSSID to steer the client to
       disassoc_imminent(disassoc_imminent_), disassoc_timer_ms(disassoc_timer_ms_),
       steer_restricted(steer_restricted_)
 {
@@ -40,6 +40,8 @@ void client_steering_task::work()
         int prev_task_id = database.get_steering_task_id(sta_mac);
         tasks.kill_task(prev_task_id);
         database.assign_steering_task_id(sta_mac, id);
+
+        original_bssid = database.get_node_parent(sta_mac);
 
         steer_sta();
 
@@ -56,19 +58,19 @@ void client_steering_task::work()
     case FINALIZE: {
 
         if (!steering_success && disassoc_imminent) {
-            TASK_LOG(DEBUG) << "steering failed for " << sta_mac << " from " << original_hostap_mac
-                            << " to " << hostap_mac;
+            TASK_LOG(DEBUG) << "steering failed for " << sta_mac << " from " << original_bssid
+                            << " to " << target_bssid;
 
             /*
                  * might need to split this logic to high and low bands of 5GHz
                  * since some clients can support one but not the other
                  */
-            if (database.is_node_24ghz(original_hostap_mac) && database.is_node_5ghz(hostap_mac)) {
+            if (database.is_node_24ghz(original_bssid) && database.is_node_5ghz(target_bssid)) {
                 TASK_LOG(DEBUG) << "steering from 2.4GHz to 5GHz failed --> updating failed 5ghz "
                                    "steering attempt";
                 database.update_node_failed_5ghz_steer_attempt(sta_mac);
-            } else if (database.is_node_5ghz(original_hostap_mac) &&
-                       database.is_node_24ghz(hostap_mac)) {
+            } else if (database.is_node_5ghz(original_bssid) &&
+                       database.is_node_24ghz(target_bssid)) {
                 TASK_LOG(DEBUG) << "steering from 5GHz to 2.4GHz failed, updating failed 2.4ghz "
                                    "steering attempt";
                 database.update_node_failed_24ghz_steer_attempt(sta_mac);
@@ -92,8 +94,8 @@ void client_steering_task::steer_sta()
         }
     }
 
-    std::string radio_mac = database.get_node_parent_radio(hostap_mac);
-    original_hostap_mac   = database.get_node_parent(sta_mac);
+    std::string radio_mac = database.get_node_parent_radio(target_bssid);
+    original_bssid        = database.get_node_parent(sta_mac);
 
     // Send 17.1.27	Client Association Control Request
     if (!cmdu_tx.create(0, ieee1905_1::eMessageType::CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE)) {
@@ -110,7 +112,7 @@ void client_steering_task::steer_sta()
     }
 
     association_control_request_tlv->bssid_to_block_client() =
-        network_utils::mac_from_string(hostap_mac);
+        network_utils::mac_from_string(target_bssid);
     association_control_request_tlv->association_control() =
         wfa_map::tlvClientAssociationControlRequest::UNBLOCK;
     association_control_request_tlv->validity_period_sec() = 0;
@@ -119,14 +121,14 @@ void client_steering_task::steer_sta()
     std::get<1>(sta_list_unblock) = network_utils::mac_from_string(sta_mac);
 
     auto agent_mac = database.get_node_parent_ire(radio_mac);
-    TASK_LOG(DEBUG) << "sending allow request for " << sta_mac << " to " << radio_mac
+    TASK_LOG(DEBUG) << "sending allow request for " << sta_mac << " to bssid " << target_bssid
                     << " id=" << int(id);
     son_actions::send_cmdu_to_agent(agent_mac, cmdu_tx, database, radio_mac);
 
     // update bml listeners
     bml_task::client_allow_req_available_event client_allow_event;
     client_allow_event.sta_mac    = sta_mac;
-    client_allow_event.hostap_mac = hostap_mac;
+    client_allow_event.hostap_mac = target_bssid;
     client_allow_event.ip         = database.get_node_ipv4(sta_mac);
     tasks.push_event(database.get_bml_task_id(), bml_task::CLIENT_ALLOW_REQ_EVENT_AVAILABLE,
                      &client_allow_event);
@@ -141,14 +143,14 @@ void client_steering_task::steer_sta()
             LOG(ERROR) << "Failed building message!";
             return;
         }
-        roam_request->params().bssid   = network_utils::mac_from_string(hostap_mac);
-        roam_request->params().channel = database.get_node_channel(hostap_mac);
+        roam_request->params().bssid   = network_utils::mac_from_string(target_bssid);
+        roam_request->params().channel = database.get_node_channel(target_bssid);
         son_actions::send_cmdu_to_agent(agent_mac, cmdu_tx, database, radio_mac);
 
         // update bml listeners
         bml_task::bh_roam_req_available_event bh_roam_event;
-        bh_roam_event.bssid   = hostap_mac;
-        bh_roam_event.channel = database.get_node_channel(hostap_mac);
+        bh_roam_event.bssid   = target_bssid;
+        bh_roam_event.channel = database.get_node_channel(target_bssid);
         tasks.push_event(database.get_bml_task_id(), bml_task::BH_ROAM_REQ_EVENT_AVAILABLE,
                          &bh_roam_event);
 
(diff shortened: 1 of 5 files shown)


tomereli (approved):

tomereli (approved):
LGTM although I have some comments, so adding the don't merge tag so you can fix them before merge.

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -1191,10 +1191,8 @@ void optimal_path_task::work()
                             << sta_mac << " steer from AP " << current_hostap_vap << " to AP "
                             << chosen_hostap;
             bool disassoc_imminent = true;
-            int disassoc_timer_ms  = 100;
-            steering_task_id =
-                son_actions::steer_sta(database, cmdu_tx, tasks, sta_mac, chosen_hostap,
-                                       disassoc_imminent, disassoc_timer_ms);
+            steering_task_id       = son_actions::steer_sta(database, cmdu_tx, tasks, sta_mac,
```
vitalybu, line 1194: I don't understand the commit message.
Also, did the signature of `steer_sta()` changed?
You replaced an `int` variable with a `bool` one...

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -249,11 +249,18 @@ void client_steering_task::steer_sta()
...
     if (event_type == STA_CONNECTED) {
-        TASK_LOG(DEBUG) << "steering successful for sta " << sta_mac;
-        steering_success = true;
+        auto current_bssid = database.get_node_parent(sta_mac);
+        if (target_bssid == current_bssid) {
+            TASK_LOG(DEBUG) << "steering successful for sta " << sta_mac << " to bssid "
+                            << current_bssid;
+            steering_success = true;
+        } else {
+            TASK_LOG(DEBUG) << "sta " << sta_mac << " steered to bssid " << current_bssid
```
vitalybu: What happens in this case?
You don't mark the steering as successful, the task will retry to steer this station again?
As far as I remember, the task won't finish as long as `steering_success != true`

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -163,40 +163,48 @@ void client_steering_task::steer_sta()
...
-            return;
-        }
-
-        auto association_control_block_request_tlv =
-            cmdu_tx.addClass<wfa_map::tlvClientAssociationControlRequest>();
-        if (!association_control_block_request_tlv) {
-            LOG(ERROR) << "addClass wfa_map::tlvClientAssociationControlRequest failed";
-            return;
+        auto hostap_vap_lits = database.get_hostap_vap_list(hostap);
+        auto ssid            = database.get_hostap_ssid(target_bssid);
```
vitalybu: Use const references...
```suggestion
        const auto& hostap_vaps = database.get_hostap_vap_list(hostap);
        const auto& ssid            = database.get_hostap_ssid(target_bssid);
```

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -95,7 +95,6 @@ void client_steering_task::steer_sta()
     }
 
     std::string radio_mac = database.get_node_parent_radio(hostap_mac);
-    auto current_ap_mac   = database.get_node_parent(sta_mac);
```
arnout, line 98: Instead of removing this, I'd move the setting of original_hostap_mac to here instead of in the work() function. That way, the work() function just has the boilerplate of managing state, and the steer_sta function has all the logic. (nitpick)

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -163,40 +163,48 @@ void client_steering_task::steer_sta()
...
-
-        auto association_control_block_request_tlv =
-            cmdu_tx.addClass<wfa_map::tlvClientAssociationControlRequest>();
-        if (!association_control_block_request_tlv) {
-            LOG(ERROR) << "addClass wfa_map::tlvClientAssociationControlRequest failed";
-            return;
+        auto hostap_vap_lits = database.get_hostap_vap_list(hostap);
+        auto ssid            = database.get_hostap_ssid(target_bssid);
+        for (auto &hostap_vap : hostap_vap_lits) {
+            if (hostap_vap.second.ssid == ssid) {
```
vitalybu: Remove extra indentation:
```suggestion
            if (hostap_vap.second.ssid != ssid) {
                continue;
            }
            .
            .
            .
```

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -163,40 +163,48 @@ void client_steering_task::steer_sta()
...
-        }
-
-        auto association_control_block_request_tlv =
-            cmdu_tx.addClass<wfa_map::tlvClientAssociationControlRequest>();
-        if (!association_control_block_request_tlv) {
-            LOG(ERROR) << "addClass wfa_map::tlvClientAssociationControlRequest failed";
-            return;
+        auto hostap_vap_lits = database.get_hostap_vap_list(hostap);
+        auto ssid            = database.get_hostap_ssid(target_bssid);
+        for (auto &hostap_vap : hostap_vap_lits) {
```
vitalybu:
```suggestion
        for (const auto &hostap_vap : hostap_vaps) {
```

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -23,11 +23,11 @@ using namespace son;
 
 client_steering_task::client_steering_task(db &database_, ieee1905_1::CmduMessageTx &cmdu_tx_,
                                            task_pool &tasks_, std::string sta_mac_,
-                                           std::string hostap_mac_, bool disassoc_imminent_,
+                                           std::string target_bssid_, bool disassoc_imminent_,
```
arnout, line 26: *Excellent*!

Now also use sMacAddr instead of std::string and I'm going to be really happy :-) (Just to be clear: no need to do that in this PR)

Unfortunately, though, it is not correct. In the current optimal path code, the hostap_mac is actually a radio mac, not a BSSID. At least, I think so... So to make sure this code is entirely correct, I think it's better to *first* fix up the naming in optimal_path_task, and only afterwards fix it in client_steering_task.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -218,7 +217,7 @@ void client_steering_task::steer_sta()
     steering_request_tlv->request_flags().btm_disassociation_imminent_bit = disassoc_imminent;
 
     steering_request_tlv->btm_disassociation_timer_ms() = disassoc_timer_ms;
-    steering_request_tlv->bssid() = network_utils::mac_from_string(current_ap_mac);
+    steering_request_tlv->bssid() = network_utils::mac_from_string(original_hostap_mac);
```
morantr: Since `original_hostap_mac` is actually "BSSID mac" and not the "radio mac", it is probably will be a good idea to change its name on the class to `m_current_bssid` or something like this.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -249,11 +249,18 @@ void client_steering_task::steer_sta()
...
+                            << current_bssid;
+            steering_success = true;
+        } else {
+            TASK_LOG(DEBUG) << "sta " << sta_mac << " steered to bssid " << current_bssid
+                            << " ,target bssid was " << target_bssid;
+        }
     } else if (event_type == STA_DISCONNECTED) {
         TASK_LOG(DEBUG) << "sta " << sta_mac
-                        << " disconnected after successful steering, proceeding to unblock";
+                        << " disconnected after steering, proceeding to unblock";
```
arnout: Actually, the "proceeding to unblock" part is also wrong, we don't do any unblocking. Instead, we just have the steering_wait_time_ms which is the amount of time it will be blocked.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -163,40 +163,48 @@ void client_steering_task::steer_sta()
...
-                << "cmdu creation of type CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE, has failed";
-            return;
-        }
-
-        auto association_control_block_request_tlv =
-            cmdu_tx.addClass<wfa_map::tlvClientAssociationControlRequest>();
-        if (!association_control_block_request_tlv) {
-            LOG(ERROR) << "addClass wfa_map::tlvClientAssociationControlRequest failed";
-            return;
+        auto hostap_vap_lits = database.get_hostap_vap_list(hostap);
```
arnout: lits -> list ?

Thinking of the renaming you did in a previous commit, it would make sense to call it `radio_list` instead.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -163,40 +163,48 @@ void client_steering_task::steer_sta()
...
+
+                association_control_block_request_tlv->association_control() =
+                    wfa_map::tlvClientAssociationControlRequest::BLOCK;
+                association_control_block_request_tlv->validity_period_sec() =
+                    steering_wait_time_ms / 1000;
+                association_control_block_request_tlv->alloc_sta_list();
+                auto sta_list_block         = association_control_block_request_tlv->sta_list(0);
+                std::get<1>(sta_list_block) = network_utils::mac_from_string(sta_mac);
+                association_control_block_request_tlv->bssid_to_block_client() =
+                    network_utils::mac_from_string(hostap_vap.second.mac);
```
arnout: Please keep the order of setting TLV fields to the same order as defined in the structure, because in some cases the tlvf is sensitive to the order. It was correct in the original.

tests/test_flows.sh
```diff
@@ -509,7 +509,7 @@ test_client_steering_dummy() {
     send_bwl_event ${REPEATER1} wlan2 "EVENT AP-STA-CONNECTED ${sta_mac}"
     dbg "Confirm steering success by client connected"
     check_log ${GATEWAY} controller "steering successful for sta ${sta_mac}"
-    check_log ${GATEWAY} controller "sta ${sta_mac} disconnected after successful steering"
+    check_log ${GATEWAY} controller "sta ${sta_mac} disconnected after steering, proceeding to unblock"
```
arnout: Please squash this with the commit that changes the message. And perhaps do that in a separate commit, not together with the change of what "success" means.

arnout (approved):
Very good PR with good commit messages, thanks!

Although there is one comment that is essential (hostap_mac actually is radio_uid at the moment and not bssid), it is already broken in the current code (or I'm wrong of course) so OK to merge as is.

controller/src/beerocks/master/tasks/client_steering_task.cpp
arnout: The task finishes as soon as *a* DISCONNECT and *a* CONNECT event is received. In the FINALIZE state, the STA node is updated with update_node_failed_5ghz_steer_attempt if success is not true.

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -41,7 +41,7 @@ void client_steering_task::work()
         tasks.kill_task(prev_task_id);
         database.assign_steering_task_id(sta_mac, id);
 
-        original_hostap_mac = database.get_node_parent(sta_mac);
+        original_bssid = database.get_node_parent(sta_mac);
```
morantr, line 44: If you renaming it, then I think "current_bssid" is more suitable, and since it is a class member I would also add "m_" prefix -> "m_current_bssid". Same for "target_bssid" -> "m_target_bssid".

controller/src/beerocks/master/tasks/client_steering_task.cpp
```diff
@@ -249,11 +249,18 @@ void client_steering_task::steer_sta()
 void client_steering_task::handle_event(int event_type, void *obj)
 {
     if (event_type == STA_CONNECTED) {
-        TASK_LOG(DEBUG) << "steering successful for sta " << sta_mac;
-        steering_success = true;
+        auto current_bssid = database.get_node_parent(sta_mac);
```
morantr: This is confusing. I would like it more if it was "newly_connected_bssid'. (nitpick)

controller/src/beerocks/master/tasks/client_steering_task.cpp
morantr: @arnout it is not "radio_list" it is actually "bssid_list", meaning the vaps under the radio.

morantr (approved):

controller/src/beerocks/master/tasks/client_steering_task.cpp
kantera800: I checked, all steer_sta() calls are with bssid.
The naming are wrong in optimal path that's for sure but since we had a lot of bugs on the steering task I chose to fix them first. Next thing I'm going to do is to refractor optimal path task :)

controller/src/beerocks/master/tasks/optimal_path_task.cpp
kantera800: now that the "disassoc_timer" default is a valid number (200ms) no need to set it when calling steer_clint() since its set to default in client_steering_task() class constructor.

controller/src/beerocks/master/tasks/client_steering_task.cpp
kantera800: "original" is correct since the task still exists also after steering to the new bssid.
also , its changed to "original_bssid" in next commits.

regarding the m_ prefix ,I'll do it in different PR since  it needs to be done to all params in the class.

controller/src/beerocks/master/tasks/client_steering_task.cpp
kantera800: I'll do it in different PR since it needs to be done to all params in the class.

PR #898, opened 2020-02-27
Bugfix/fail to read cross rssi measurement change

kantera800:
When rssi measurement notification arrives it's supposed to be saved
to an unordered map with a radio mac as key.
Currently set_node_cross_rx_rssi sent with vap mac (bssid) instead of radio mac.
So, we can't find it when try and look for it using radio mac as map key.
As a result optimal path task fails to reach a correct decision.

Use radio mac as map key to save cross rssi  measurement.


tomereli (approved):
Nicely done! Great split to commits, good commit messages.

morantr (approved):

PR #899, opened 2020-02-27
Bugfix/channel switch wrong bandwidth

### LINKED ISSUE #344: [TASK] Pass flow 4.2.1 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.2.1 in the certification testbed.

# Exit criteria

- Certification test passes

### LINKED ISSUE #345: 4.2.1 Multi-AP Ethernet Onboarding and Initialization
The prplMesh agent on Turris Omnia with OpenWrt needs to pass the certification test case 4.2.1 as defined by the Wi-Fi CERTIFIED EasyMesh (tm) Test Plan.

- [x] #336 Support wfa_ca in agent
- [x] #346 AP configuration on Turris Omnia / RAX40 OpenWrt
- [x] #343 Test flow for 4.2.1 Multi-AP Ethernet Onboarding and Initialization
- [x] #344 Pass flow 4.2.1 test in testbed

tomereli:
Fix several bugs in the channel selection implementation:
- Fix wrong bandwidth use in DWPAL which caused chan_switch command to fail
- Fix bug in agent's channel selection which selected a restricted channel
- Fix bug in agent's channel selection which caused the agent not to send operating channel report when channel switch was not needed.
- Fix bug in agent's channel selection which did not follow the Multi-AP specification by not marking channel preference with empty channel list as non-operable

WiFi Alliance support case # 00129476 helped identify the final bug. With this, we pass 4.2.1 without any workarounds.

Fixes #344

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -2256,7 +2256,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             // on the first channel on the list and sFlags itself.
             // See: https://github.com/prplfoundation/prplMesh/issues/8
 
-            op_class_channels->operating_class() = preference.preference.oper_class;
+            op_class_channels->operating_class() = preference.oper_class;
             if (!op_class_channels->alloc_channel_list(preference.channels.size())) {
                 LOG(ERROR) << "alloc_channel_list() has failed!";
                 return false;
@@ -2269,9 +2269,9 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             }
 
             // Update channel list flags
-            op_class_channels->flags().preference = preference.preference.preference;
+            op_class_channels->flags().preference = preference.preference;
             op_class_channels->flags().reason_code =
-                (wfa_map::cPreferenceOperatingClasses::eReasonCode)preference.preference.reason;
+                (wfa_map::cPreferenceOperatingClasses::eReasonCode)preference.reason;
 
             // Push operating class object to the list of operating class objects
             if (!channel_preference_tlv->add_operating_classes_list(op_class_channels)) {
@@ -4387,14 +4387,14 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
 beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
 {
     for (const auto &preference : channel_preferences) {
-        LOG(DEBUG) << "Preference operating class: " << int(preference.preference.oper_class);
+        LOG(DEBUG) << "Preference operating class: " << int(preference.oper_class);
         for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
             auto channel         = hostap_params.supported_channels[i];
             auto operating_class = wireless_utils::get_operating_class_by_channel(
                 channel.channel, static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth));
 
             // Skip channels from other operating classes
-            if (operating_class != preference.preference.oper_class) {
+            if (operating_class != preference.oper_class) {
                 continue;
             }
             // Skip DFS channels
@@ -4431,14 +4431,14 @@ bool slave_thread::channel_selection_current_channel_restricted()
     for (const auto &preference : channel_preferences) {
         // for now we handle only non-operable preference
         // TODO - handle as part of https://github.com/prplfoundation/prplMesh/issues/725
-        if (static_cast<beerocks::eChannelPreference>(preference.preference.preference) !=
-            beerocks::eChannelPreference::NON_OPERABLE) {
+        if (preference.preference !=
+            wfa_map::cPreferenceOperatingClasses::ePreference::NON_OPERABLE) {
             LOG(WARNING) << "Ignoring operable channels preference";
             continue;
         }
         // According to Table 23 in the MultiAP Specification, an empty channel list field
         // indicates that the indicated preference applies to all channels in the operating class.
-        if (preference.preference.oper_class == operating_class && preference.channels.empty()) {
+        if (preference.oper_class == operating_class && preference.channels.empty()) {
             LOG(INFO) << "Current operating class " << int(operating_class) << " restricted,"
                       << "channel switch required";
             return true;
@@ -4515,9 +4515,12 @@ bool slave_thread::channel_selection_get_channel_preference(ieee1905_1::CmduMess
                 channels_list.push_back(wifi_channel);
             }
             LOG(DEBUG) << ss.str();
-            auto pref =
-                beerocks::message::sPreference({operating_class, preference, uint8_t(reason_code)});
-            channel_preferences.push_back({pref, channels_list});
+            wireless_utils::sChannelPreference pref;
+            pref.oper_class = operating_class;
+            pref.preference = preference;
+            pref.reason     = uint8_t(reason_code);
+            pref.channels   = channels_list;
+            channel_preferences.push_back(pref);
         }
     }
 
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4557,9 +4557,21 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
         return false;
     }
 
-    if (!switch_required && !power_limit_received) {
+    // Currently, setting and reading tx power limit is not implemented
+    // in any or our BWL backends.
+    // Therefore, updating only the tx power limit does not result with an event
+    // which should trigger the agent to send the operating channel report
+    // with the new tx power, thus failing certification onboarding test 4.2.1.
+    // As a temporary workaround, always send the operating channel report regardless
+    // of the tx power setting.
+    // TODO - Fix as part of https://github.com/prplfoundation/prplMesh/issues/725
+    if (!switch_required) {
+        LOG(INFO) << "Channel switch not required, sending operating channel report";
         send_operating_channel_report();
-        return true;
+        if (!power_limit_received) {
+            LOG(INFO) << "No power limit received - nothing else to do";
+            return true;
+        }
     }
 
     auto request_out = message_com::create_vs_message<
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4477,6 +4477,8 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
                          });
 
         if (is_restricted != preference.channels.end()) {
+            LOG(INFO) << "Current channel " << int(hostap_cs_params.channel)
+                      << " is restricted, switch required";
             switch_required = 1;
             break;
         }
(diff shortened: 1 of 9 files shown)


YuriyMasechko:
According to design channel selection request should contain only restricted channels. But it still could contain empty list and we should count this. So pls skip operating classes with empty list and don't add  them to preferences.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4555,9 +4555,13 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
         return false;
     }
 
-    if (!switch_required && !power_limit_received) {
+    if (!switch_required) {
+        LOG(INFO) << "Channel switch not required, sending operating channel report";
         send_operating_channel_report();
```
YuriyMasechko, line 4616: Is this temporary solution? If power limit was received, channel report should contain new power limit.

tomereli: Good point, will fix it.

tomereli:
> According to design channel selection request should contain only restricted channels. But it still could contain empty list and we should count this. So pls skip operating classes with empty list and don't add them to preferences.

See https://github.com/prplfoundation/prplMesh/pull/899/commits/566198a812c7d30a359cca5f894a08eb4c3d8e76

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4475,37 +4475,50 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
                          });
 
         if (is_restricted != preference.channels.end()) {
+            LOG(INFO) << "Current channel " << int(hostap_cs_params.channel)
+                      << " is restricted, switch required";
             switch_required = 1;
```
morantr: Why `switch_required` is defined as `int`? Clearly it should be `bool` and be set to `true` or `false` instead of `1` or `0`.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4475,37 +4475,50 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
...
-                                    (beerocks::eWiFiBandwidth)hostap_params.supported_channels[i]
-                                        .channel_bandwidth) &&
-                            hostap_params.supported_channels[i].channel == wifi_channel.channel);
-                    });
-
-                if (is_restricted == preference.channels.end() &&
-                    !hostap_params.supported_channels[i].is_dfs_channel) {
-                    channel_to_switch = hostap_params.supported_channels[i];
-                    break;
+        for (auto preference : channel_preferences) {
```
morantr: It needs to be const reference.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Yes this is a temporary solution so we can pass certification. I added a comment and a TODO message linking this to https://github.com/prplfoundation/prplMesh/issues/725

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4475,37 +4475,50 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
                          });
```
morantr: I don't understand why we set `is_restricted` if the channel from the list is equal to the current channel.
AFAIR it is only restricted if the channel preference is 0.
So IMO it should be looking like this:
```cpp
    for (const auto &preference : channel_preferences) {
        // Check if channel list is with preference of restricted channel
        if (preference.preference.preference == 0) {
            continue;
        }

        auto current_channel_is_restricted = preference.channels.end() == 
                std::find_if(preference.channels.begin(), preference.channels.end(),
                [&hostap_cs_params.channel](const beerocks::message::sWifiChannel &wifi_channel) {
	              return hostap_cs_params.channel == wifi_channel.channel;
	        );

        if (current_channel_is_restricted) {
            LOG(INFO) << "Current channel " << int(hostap_cs_params.channel)
                      << " is restricted, switch required";
            switch_required = true;
            break;
        }
    }
```

Please notice that on my code suggestion, I changed a few things:
1. Use const reference on the main for-loop.
2. Add only `hostap_cs_params.channe` as reference to the lambda function. Adding everything as a reference is useless and it is translated to passing everything possible as a reference to the function, so it hurts performance.
3. Added a `break` when `switch_required` is set to `true`. There is no need to continue with furthermore iterations. 

I'm not sure, but it seems to me that it is possible to merge this loop and the one down below.

tomereli:
> Don't you always need to send the operating channel report, regardless of whether the switch is required?

Yes I do, and I updated the PR to do that. With this we pass the test.

morantr (commented):
I didn't finish, this is what I got so far.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Correct, but it was int to begin with so I didn't want to change that.
If you think it is critical (you didn't mark this as (nitpick), I can add another commit which does that.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Correct but again this was the same in previous implementation. Will change as part of a separate commit.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: You are again correct. The current code assumes that `preference` is always `0`, meaning always non-operable. The correct solution is to take the actual preference into account but then we will need to add much more code which is not related to this PR and should be handled in #725 - this I prefer not to do as part of a bug fix PR.

Regarding the change it self - yours is wrong, since if channel list is empty, you don't mark *any* channel as non-operable (restricted). 

However thank you for this comment - I fixed everything you requested.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: NVM apparently I already changed this as part of c7b5669

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: fixed

tomereli:
[4.2.1_PASS.zip](https://github.com/prplfoundation/prplMesh/files/4281488/4.2.1_PASS.zip)

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: Just to clarify, this comment was on the upper for-loop (line 4470)and not the one you have changed.

rmelotte: Just a quick note because I've seen it before, this is not correct:
> Adding everything as a reference is useless and it is translated to passing everything possible as a reference to the function"

Only the variables you use in the lambda body will be captured by reference so you don't need to explicitly list them (see [cppreference](https://en.cppreference.com/w/cpp/language/lambda)). 
Of course you can still list them if you prefer but if it's for "performance reasons", then you don't have to. The 2 are equivalent.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4422,29 +4422,37 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
 {
-    for (auto preference : channel_preferences) {
+    auto channel         = hostap_cs_params.channel;
+    auto bw              = static_cast<beerocks::eWiFiBandwidth>(hostap_cs_params.bandwidth);
+    auto operating_class = wireless_utils::get_operating_class_by_channel(channel, bw);
+
+    for (const auto &preference : channel_preferences) {
+        // for now we handle only non-operable preference
+        // TODO - handle as part of https://github.com/prplfoundation/prplMesh/issues/725
+        if (preference.preference.preference != 0) {
```
mariomaz: Define an enum instead of using a magic number (i.e.: NON_OPERABLE: 0, MOST_PREFERRED: 15)
(nitpick)

mariomaz: Surely there must be a better name for this: `preference.preference.preference`
(nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4382,16 +4382,66 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
     return message_com::send_cmdu(ap_manager_socket, cmdu_tx);
 }
 
-bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
 {
-    const auto mid = cmdu_rx.getMessageId();
-    LOG(DEBUG) << "Received CHANNEL_SELECTION_REQUEST_MESSAGE, mid=" << std::dec << int(mid);
+    for (auto preference : channel_preferences) {
```
morantr: const reference

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4382,16 +4382,66 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
-    int switch_required       = 0;
-    int power_limit           = 0;
-    beerocks::message::sWifiChannel channel_to_switch;
+    LOG(ERROR) << "Could not find a suitable channel";
+    return beerocks::message::sWifiChannel();
+}
+
+bool slave_thread::channel_selection_current_channel_restricted()
+{
+    for (auto preference : channel_preferences) {
```
morantr: const reference

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4448,6 +4498,12 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
         }
     }
 
+    return true;
+}
+
+bool slave_thread::channel_selection_get_transmit_power_limit(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                              int &power_limit)
+{
     for (auto tx_power_limit_tlv : cmdu_rx.getClassList<wfa_map::tlvTransmitPowerLimit>()) {
```
mariomaz: You should use 'const auto&'
I think `cmdu_rx` parameter should also be a const reference

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4382,16 +4382,66 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
+    LOG(INFO) << "Current channel " << int(hostap_cs_params.channel)
+              << " not restricted, channel switch not required";
+    return false;
+}
+
+bool slave_thread::channel_selection_get_channel_preference(ieee1905_1::CmduMessageRx &cmdu_rx)
+{
     channel_preferences.clear();
 
     for (auto channel_preference_tlv : cmdu_rx.getClassList<wfa_map::tlvChannelPreference>()) {
```
mariomaz, line 4467: You should use 'const auto&'
I think `cmdu_rx` parameter should also be a const reference

mariomaz (changes requested):

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: It bothers me too.

morantr (changes requested):

morantr (approved):

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Done

tomereli: Not part of this PR, but I can fix it in a separate commit. Will do.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: Done

agent/src/beerocks/slave/son_slave_thread.cpp
mariomaz: This one has been marked as resolved but it is not.
This file deserves a full review regarding the 'const auto&' usage.

mariomaz (approved):
son_slave_thread.cpp deserves a full review regarding the 'const auto&' usage.

PR #900, opened 2020-02-27
Feature/add support for nl80211 standard commands to BWL

mariomaz:
Create C++ wrappers around libnl and libnl-genl to facilitate sending
NL80211 commands through a NL80211 netlink socket.

Classes have been created in bwl/shared so NL80211 standard commands
will be available to all flavours of the BWL library except the Dummy one.

Add NL80211_CMD_GET_STATION command implementation as an example of
how standard commands are to be implemented.

Once this design had been validated, the rest of standard commands will
be added. Also we will do a code refactoring to use the new classes and
eliminate duplicated code.

### DIFF
--- a/common/beerocks/bwl/CMakeLists.txt
+++ b/common/beerocks/bwl/CMakeLists.txt
@@ -49,6 +49,7 @@ if(BWL_TYPE STREQUAL "DWPAL")
 
     file(GLOB bwl_platform_sources
         ${bwl_platform_sources}
+        ${MODULE_PATH}/shared/*.c*
         ${MODULE_PATH}/dwpal/*.c*
     )
 
@@ -75,6 +76,7 @@ if(BWL_TYPE STREQUAL "DWPAL")
 elseif(BWL_TYPE STREQUAL "NL80211")
 
     file(GLOB bwl_platform_sources
+        ${MODULE_PATH}/shared/*.c*
         ${MODULE_PATH}/nl80211/*.c*
 
         # WPA Control Interface
@@ -103,6 +105,18 @@ elseif(BWL_TYPE STREQUAL "NL80211")
 
     list(APPEND BWL_LIBS nl-tiny)
 
+elseif(BWL_TYPE STREQUAL "DUMMY")
+
+    set(bwl_platform_sources
+        ${bwl_platform_sources}
+        ${MODULE_PATH}/dummy/nl80211_client_dummy.cpp
+        ${MODULE_PATH}/dummy/nl80211_client_factory_dummy.cpp
+        ${MODULE_PATH}/dummy/mon_wlan_hal_dummy.cpp
+        ${MODULE_PATH}/dummy/ap_wlan_hal_dummy.cpp
+        ${MODULE_PATH}/dummy/base_wlan_hal_dummy.cpp
+        ${MODULE_PATH}/dummy/sta_wlan_hal_dummy.cpp
+    )
+    
 else()
 
     string(TOLOWER ${BWL_TYPE} BWL_PATH)
--- a/common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
+++ b/common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
@@ -1,0 +1,51 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "nl80211_client_dummy.h"
+
+namespace bwl {
+
+nl80211_client_dummy::nl80211_client_dummy() {}
+
+nl80211_client_dummy::~nl80211_client_dummy() {}
+
+bool nl80211_client_dummy::get_sta_info(const std::string &local_interface_name,
+                                        const sMacAddr &sta_mac_address, sStaInfo &sta_info)
+{
+    // Suppress "unused parameter" warning
+    (void)local_interface_name;
+    (void)sta_mac_address;
+
+    static int inactive_time_ms        = 0;
+    static int rx_bytes                = 0;
+    static int rx_packets              = 0;
+    static uint32_t tx_bytes           = 0;
+    static uint32_t tx_packets         = 0;
+    static uint32_t tx_retries         = 0;
+    static uint32_t tx_failed          = 0;
+    static uint8_t signal_dBm          = 0;
+    static uint8_t signal_avg_dBm      = 0;
+    static uint16_t tx_bitrate_100kbps = 0;
+    static uint16_t rx_bitrate_100kbps = 0;
+
+    sta_info.inactive_time_ms   = inactive_time_ms++;
+    sta_info.rx_bytes           = rx_bytes++;
+    sta_info.rx_packets         = rx_packets++;
+    sta_info.tx_bytes           = tx_bytes++;
+    sta_info.tx_packets         = tx_packets++;
+    sta_info.tx_retries         = tx_retries++;
+    sta_info.tx_failed          = tx_failed++;
+    sta_info.signal_dBm         = signal_dBm++;
+    sta_info.signal_avg_dBm     = signal_avg_dBm++;
+    sta_info.tx_bitrate_100kbps = tx_bitrate_100kbps++;
+    sta_info.rx_bitrate_100kbps = rx_bitrate_100kbps++;
+
+    return true;
+}
+
+} // namespace bwl
--- a/common/beerocks/bwl/dummy/nl80211_client_factory_dummy.cpp
+++ b/common/beerocks/bwl/dummy/nl80211_client_factory_dummy.cpp
@@ -1,0 +1,22 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include <bwl/nl80211_client_factory.h>
+
+#include "nl80211_client_dummy.h"
+
+#include <bcl/beerocks_backport.h>
+
+namespace bwl {
+
+std::unique_ptr<nl80211_client> nl80211_client_factory::create_instance()
+{
+    return std::make_unique<nl80211_client_dummy>();
+}
+
+} // namespace bwl
--- a/common/beerocks/bwl/include/bwl/nl80211_client.h
+++ b/common/beerocks/bwl/include/bwl/nl80211_client.h
@@ -1,0 +1,81 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+#ifndef __BWL_NL80211_CLIENT_H__
+#define __BWL_NL80211_CLIENT_H__
+
+#include <tlvf/common/sMacAddr.h>
+
+#include <string>
+
+namespace bwl {
+
+/**
+ * @brief NL80211 client interface.
+ *
+ * This interface lists the methods to read and write wireless hardware status and configuration
+ * through a NL80211 socket.
+ *
+ * Programming to an interface allows dependent classes to remain unaware of the different
+ * implementations of the interface as well as facilitates the creation of mock implementations for
+ * unit testing.
+ *
+ * This is a C++ interface: an abstract class that is designed to be specifically used as a base
+ * class and which derived classes (implementations) will override each pure virtual function.
+ *
+ * Known implementations: nl80211_client_impl (uses NL80211 socket), nl80211_client_dummy (fake
+ * implementation that returns dummy data).
+ */
+class nl80211_client {
+
+public:
+    /**
+     * @brief Station information
+     *
+     * Information obtained with NL80211_CMD_GET_STATION command through a NL80211 socket.
+     * See NL80211_STA_INFO_* in <linux/nl80211.h> for a description of each field.
+     */
+    struct sStaInfo {
+        uint32_t inactive_time_ms   = 0;
+        uint32_t rx_bytes           = 0;
+        uint32_t rx_packets         = 0;
+        uint32_t tx_bytes           = 0;
+        uint32_t tx_packets         = 0;
+        uint32_t tx_retries         = 0;
+        uint32_t tx_failed          = 0;
+        uint8_t signal_dBm          = 0;
+        uint8_t signal_avg_dBm      = 0;
+        uint16_t tx_bitrate_100kbps = 0;
+        uint16_t rx_bitrate_100kbps = 0;
+    };
+
+    /**
+     * @brief Class destructor.
+     */
+    virtual ~nl80211_client() = default;
+
+    /**
+     * @brief Gets station information.
+     *
+     * Station information contains basically metrics associated to the link between given local
+     * interface and a the interface of a station whose MAC address is 'sta_mac_address'.
+     *
+     * @param[in] local_interface_name Virtual AP (VAP) interface name.
+     * @param[in] sta_mac_address MAC address of a station connected to the local interface.
+     * @param[out] sta_info Station information.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_sta_info(const std::string &local_interface_name,
+                              const sMacAddr &sta_mac_address, sStaInfo &sta_info) = 0;
+
+    // TODO: add remaining methods
+};
+
+} // namespace bwl
+
+#endif
--- a/common/beerocks/bwl/shared/README.md
+++ b/common/beerocks/bwl/shared/README.md
@@ -1,0 +1,6 @@
+# Shared files directory
+
+The `shared` directory contains classes shared among all the implementations that use actual WiFi radios (i.e.: the DWPAL and NL80211 flavors but not the Dummy one).
+
+Currently, these classes provide the following functionality:
+- Support for NL80211 standard commands through a NL80211 socket. 
--- a/common/beerocks/bwl/shared/netlink_socket.cpp
+++ b/common/beerocks/bwl/shared/netlink_socket.cpp
@@ -1,0 +1,154 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "netlink_socket.h"
+
+#include <netlink/msg.h>
+#include <netlink/netlink.h>
+
+#include <easylogging++.h>
+
+namespace bwl {
+
+netlink_socket::netlink_socket(int protocol)
+    : m_nl_socket(nl_socket_alloc(), nl_socket_free), m_protocol(protocol)
+{
+    LOG_IF(!m_nl_socket, ERROR) << "Failed to allocate netlink socket!";
+}
+
+bool netlink_socket::connect()
+{
+    if (!m_nl_socket) {
+        LOG(ERROR) << "Cannot use unallocated socket!";
+        return false;
+    }
+
+    // Connect the socket
+    if (nl_connect(m_nl_socket.get(), m_protocol) != 0) {
+        LOG(ERROR) << "Failed to connect netlink socket!";
+        return false;
+    }
+
+    return true;
+}
+
+void netlink_socket::close()
+{
+    if (!m_nl_socket) {
+        LOG(ERROR) << "Cannot use unallocated socket!";
+    }
+
+    // Terminate connection and close socket
+    nl_close(m_nl_socket.get());
+}
+
+bool netlink_socket::send_receive_msg(std::function<bool(struct nl_msg *msg)> msg_create,
+                                      std::function<bool(struct nl_msg *msg)> msg_handle)
+{
+    if (!m_nl_socket) {
+        LOG(ERROR) << "Cannot use unallocated socket!";
+        return false;
+    }
+
+    // The Netlink message to send
+    std::unique_ptr<nl_msg, void (*)(nl_msg *)> nl_message(nlmsg_alloc(), nlmsg_free);
+    if (!nl_message) {
+        LOG(ERROR) << "Failed creating netlink message!";
+        return false;
+    }
+
+    // The Netlink callback set
+    std::unique_ptr<nl_cb, void (*)(nl_cb *)> nl_callback(nl_cb_alloc(NL_CB_DEFAULT), nl_cb_put);
+    if (!nl_callback) {
+        LOG(ERROR) << "Failed creating netlink callback!";
+        return false;
+    }
+
+    // Termination flag for the loop that receives the response messages. Possible values are:
+    // error == 1: initial value, request message has not be sent yet
+    // error == 0: response has been successfully received
+    // error < 0: some error occurred while receiving response
+    // Final value is used to compute the result code of this method.
+    int error = 1;
+
+    // Create standard callbacks
+    static auto nl_err_cb = [](struct sockaddr_nl *nla, struct nlmsgerr *err, void *arg) -> int {
+        int *error = (int *)arg;
+        *error     = err->error;
+        LOG(ERROR) << "Failed to process netlink message! Error: " << *error;
+        return NL_STOP;
+    };
+
+    static auto nl_finish_cb = [](struct nl_msg *msg, void *arg) -> int {
+        int *error = (int *)arg;
+        *error     = 0;
+        return NL_SKIP;
+    };
+
+    static auto nl_ack_cb = [](struct nl_msg *msg, void *arg) -> int {
+        int *error = (int *)arg;
+        *error     = 0;
+        return NL_STOP;
+    };
+
+    // Passing a lambda with capture is not supported for standard C function
+    // pointers. As a workaround, we create a static (but thread local) wrapper
+    // function that calls the capturing lambda function.
+    static thread_local std::function<int(struct nl_msg * msg, void *arg)> nl_handler_cb_wrapper;
+    nl_handler_cb_wrapper = [&](struct nl_msg *msg, void *arg) -> int {
+        if (!msg_handle(msg)) {
+            LOG(ERROR) << "User's netlink handler function failed!";
+        }
+        return NL_SKIP;
+    };
+    auto nl_handler_cb = [](struct nl_msg *msg, void *arg) -> int {
+        return nl_handler_cb_wrapper(msg, arg);
+    };
+
+    // Call the user's message create function
+    if (!msg_create(nl_message.get())) {
+        LOG(ERROR) << "User's netlink create function failed!";
+        return false;
+    }
+
+    // Set the callbacks
+    nl_cb_err(nl_callback.get(), NL_CB_CUSTOM, nl_err_cb, &error); // error
+    nl_cb_set(nl_callback.get(), NL_CB_FINISH, NL_CB_CUSTOM, nl_finish_cb,
+              &error);                                                        // finish
+    nl_cb_set(nl_callback.get(), NL_CB_ACK, NL_CB_CUSTOM, nl_ack_cb, &error); // ack
+    nl_cb_set(nl_callback.get(), NL_CB_VALID, NL_CB_CUSTOM, nl_handler_cb,
+              nullptr); // response handler
+
+    // Send the netlink message
+    int rc = nl_send_auto_complete(m_nl_socket.get(), nl_message.get());
+    if (rc < 0) {
+        LOG(ERROR) << "Failed to send netlink message! Error: " << rc;
+        return false;
+    }
+
+    // Receive the response messages
+    // Note that call to nl_recvmsgs() is blocking and loop terminates when one of these
+    // conditions is met:
+    // - nl_recvmsgs() fails (because internal call to nl_recv() in turn fails)
+    // - One of the callback functions sets error to 0 (ok)
+    // - One of the callback functions sets error to a value lower than 0 (error)
+    // Loop is required just in case more than one message is received. Handling callback must
+    // process them all.
+    while (error > 0) {
+        int rc = nl_recvmsgs(m_nl_socket.get(), nl_callback.get());
+        if (rc < 0) {
+            LOG(ERROR) << "Failed to receive netlink messages! Error: " << rc;
+            return false;
+        }
+    }
+
+    // Return true on success and false otherwise
+    return (0 == error);
+}
+
+} // namespace bwl
--- a/common/beerocks/bwl/shared/netlink_socket.h
+++ b/common/beerocks/bwl/shared/netlink_socket.h
@@ -1,0 +1,112 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+#ifndef __BWL_NETLINK_SOCKET_H__
+#define __BWL_NETLINK_SOCKET_H__
+
+#include <functional>
+#include <memory>
+
+/**
+ * Forward declarations
+ */
+struct nl_sock;
+struct nl_msg;
+
+namespace bwl {
+
+/**
+ * @brief C++ wrapper class around the core library of the Netlink Protocol Library Suite (libnl).
+ *
+ * The Netlink socket is Linux kernel interface used for inter-process communication (IPC) between
+ * both the kernel and userspace processes.
+ *
+ * The Netlink core library contains C functions implementing the fundamentals required to use the
+ * Netlink protocol such as socket handling, message construction and parsing, and sending and
+ * receiving of data.
+ *
+ * This is a very simple C++ wrapper class around the Netlink core library. Its aim is to
+ * facilitate the use of the library from C++ applications. The class can be used as is or extended
+ * by derived classes to support extended versions of the netlink protocol like, for example, the
+ * generic netlink protocol.
+ *
+ * Known derived classes: nl_genl_socket.
+ */
+class netlink_socket {
+
+public:
+    /**
+     * @brief Class constructor.
+     *
+     * Allocates the Netlink socket structure.
+     *
+     * @param protocol Netlink protocol to use. Possible values are listed in <linux/netlink.h>
+     */
+    explicit netlink_socket(int protocol);
+
+    /**
+     * @brief Class destructor.
+     *
+     * Deallocates the Netlink socket structure.
+     */
+    virtual ~netlink_socket() = default;
+
+    /**
+     * @brief Connects the Netlink socket.
+     *
+     * Creates a new Netlink socket and binds it to the protocol specified in class constructor.
+     *
+     * @return True on success and false otherwise (e.g.: socket is already connected).
+     */
+    virtual bool connect();
+
+    /**
+     * @brief Closes the Netlink socket.
+     *
+     * Call this method to terminate connection and free resources.
+     */
+    virtual void close();
+
+    /**
+     * @brief Performs a request/response transaction.
+     *
+     * Requires Netlink socket to be connected.
+     *
+     * Creates a new Netlink message and sends it over the previously connected Netlink socket.
+     * Then waits until response message is received and handles it.
+     *
+     * Message to send is created using provided creation callback function and received message
+     * is parsed using provided handling callback function.
+     *
+     * @param[in] msg_create Message creation callback function (called on message to be sent).
+     * @param[in] msg_handle Message handling callback function (called on received message).
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool send_receive_msg(std::function<bool(struct nl_msg *msg)> msg_create,
+                                  std::function<bool(struct nl_msg *msg)> msg_handle);
+
+protected:
+    /**
+     * Netlink socket structure containing the netlink socket and all related attributes,
+     * including the actual file descriptor.
+     *
+     * This member is protected so derived classes can use it to extend or refine the
+     * functionality provided by this class.
+     */
+    std::unique_ptr<struct nl_sock, void (*)(struct nl_sock *)> m_nl_socket;
+
+private:
+    /**
+     * Netlink protocol to use. Possible values are listed in <linux/netlink.h>
+     */
+    int m_protocol;
+};
+
+} // namespace bwl
+
+#endif
--- a/common/beerocks/bwl/shared/nl80211_client_factory.cpp
+++ b/common/beerocks/bwl/shared/nl80211_client_factory.cpp
@@ -1,0 +1,34 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include <bwl/nl80211_client_factory.h>
+
+#include "nl80211_client_impl.h"
+
+#include <bcl/beerocks_backport.h>
+
+#include <easylogging++.h>
+
+namespace bwl {
+
+std::unique_ptr<nl80211_client> nl80211_client_factory::create_instance()
+{
+    // Create NL80211 socket to communicate with WiFi driver
+    auto socket = std::make_unique<bwl::nl80211_socket>();
+
+    // Connect NL80211 socket (it will be automatically closed when it is not needed any more)
+    if (!socket->connect()) {
+        LOG(ERROR) << "Failed to connect nl80211 socket!";
+        return nullptr;
+    }
+
+    // Create NL80211 client to send standard NL80211 commands to WiFi driver using NL80211 socket
+    return std::make_unique<nl80211_client_impl>(std::move(socket));
+}
+
+} // namespace bwl
--- a/common/beerocks/bwl/shared/nl80211_client_impl.cpp
+++ b/common/beerocks/bwl/shared/nl80211_client_impl.cpp
@@ -1,0 +1,185 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2020 MaxLinear
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "nl80211_client_impl.h"
+
+#include <easylogging++.h>
+
+#include <linux/if_ether.h>
+#include <linux/nl80211.h>
+#include <net/if.h>
+#include <netlink/genl/genl.h>
+
+#include <math.h>
+
+namespace bwl {
+
+nl80211_client_impl::nl80211_client_impl(std::unique_ptr<nl80211_socket> socket)
+    : m_socket(std::move(socket))
+{
+}
+
+bool nl80211_client_impl::get_sta_info(const std::string &local_interface_name,
+                                       const sMacAddr &sta_mac_address, sStaInfo &sta_info)
+{
+    static struct nla_policy stats_policy[NL80211_STA_INFO_MAX + 1];
+    stats_policy[NL80211_STA_INFO_INACTIVE_TIME] = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_RX_BYTES]      = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_RX_PACKETS]    = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_BYTES]      = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_PACKETS]    = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_RETRIES]    = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_FAILED]     = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_SIGNAL]        = {NLA_U8, 0, 0};
+    stats_policy[NL80211_STA_INFO_SIGNAL_AVG]    = {NLA_U8, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_BITRATE]    = {NLA_NESTED, 0, 0};
+    stats_policy[NL80211_STA_INFO_RX_BITRATE]    = {NLA_NESTED, 0, 0};
+
+    static struct nla_policy rate_policy[NL80211_RATE_INFO_MAX + 1];
+    rate_policy[NL80211_RATE_INFO_BITRATE]   = {NLA_U16, 0, 0};
+    rate_policy[NL80211_RATE_INFO_BITRATE32] = {NLA_U32, 0, 0};
+
+    sta_info = {};
+
+    if (!m_socket) {
+        LOG(ERROR) << "Socket is NULL!";
+        return false;
+    }
+
+    // Get the interface index for given interface name
+    int iface_index = if_nametoindex(local_interface_name.c_str());
+    if (0 == iface_index) {
+        LOG(ERROR) << "Failed to read the index of interface " << local_interface_name << ": "
+                   << strerror(errno);
+
+        return false;
+    }
+
+    return m_socket.get()->send_receive_msg(
+        NL80211_CMD_GET_STATION, NLM_F_DUMP,
+        [&](struct nl_msg *msg) -> bool {
+            nla_put_u32(msg, NL80211_ATTR_IFINDEX, iface_index);
+            nla_put(msg, NL80211_ATTR_MAC, ETH_ALEN, sta_mac_address.oct);
+
+            return true;
+        },
+        [&](struct nl_msg *msg) -> bool {
+            struct nlattr *tb[NL80211_ATTR_MAX + 1];
+            struct genlmsghdr *gnlh = (struct genlmsghdr *)nlmsg_data(nlmsg_hdr(msg));
+            struct nlattr *sinfo[NL80211_STA_INFO_MAX + 1];
+
+            // Parse the netlink message
+            if (nla_parse(tb, NL80211_ATTR_MAX, genlmsg_attrdata(gnlh, 0), genlmsg_attrlen(gnlh, 0),
+                          NULL)) {
+                LOG(ERROR) << "Failed to parse netlink message!";
+                return false;
+            }
+
+            if (!tb[NL80211_ATTR_STA_INFO]) {
+                LOG(ERROR) << "STA stats missing!";
+                return false;
+            }
+
+            // Parse nested station stats
+            if (nla_parse_nested(sinfo, NL80211_STA_INFO_MAX, tb[NL80211_ATTR_STA_INFO],
+                                 stats_policy)) {
+                LOG(ERROR) << "Failed to parse nested STA attributes!";
+                return false;
+            }
+
+            if (sinfo[NL80211_STA_INFO_INACTIVE_TIME]) {
+                sta_info.inactive_time_ms = nla_get_u32(sinfo[NL80211_STA_INFO_INACTIVE_TIME]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_INACTIVE_TIME attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_RX_BYTES]) {
+                sta_info.rx_bytes = nla_get_u32(sinfo[NL80211_STA_INFO_RX_BYTES]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_RX_BYTES attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_RX_PACKETS]) {
+                sta_info.rx_packets = nla_get_u32(sinfo[NL80211_STA_INFO_RX_PACKETS]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_RX_PACKETS attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_TX_BYTES]) {
+                sta_info.tx_bytes = nla_get_u32(sinfo[NL80211_STA_INFO_TX_BYTES]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_TX_BYTES attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_TX_PACKETS]) {
+                sta_info.tx_packets = nla_get_u32(sinfo[NL80211_STA_INFO_TX_PACKETS]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_TX_PACKETS attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_TX_RETRIES]) {
+                sta_info.tx_retries = nla_get_u32(sinfo[NL80211_STA_INFO_TX_RETRIES]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_TX_RETRIES attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_TX_FAILED]) {
+                sta_info.tx_failed = nla_get_u32(sinfo[NL80211_STA_INFO_TX_FAILED]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_TX_FAILED attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_SIGNAL]) {
+                sta_info.signal_dBm = nla_get_u8(sinfo[NL80211_STA_INFO_SIGNAL]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_SIGNAL attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_SIGNAL_AVG]) {
+                sta_info.signal_avg_dBm = nla_get_u8(sinfo[NL80211_STA_INFO_SIGNAL_AVG]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_SIGNAL_AVG attribute is missing";
+            }
+
+            // Bit rate parsing helper function
+            auto parse_bitrate_func = [&](struct nlattr *bitrate_attr) -> int {
+                struct nlattr *rinfo[NL80211_RATE_INFO_MAX + 1];
+
+                int rate = 0;
+                if (nla_parse_nested(rinfo, NL80211_RATE_INFO_MAX, bitrate_attr, rate_policy)) {
+                    LOG(ERROR) << "Failed to parse nested rate attributes!";
+                } else if (rinfo[NL80211_RATE_INFO_BITRATE32]) {
+                    rate = nla_get_u32(rinfo[NL80211_RATE_INFO_BITRATE32]);
+                } else if (rinfo[NL80211_RATE_INFO_BITRATE]) {
+                    rate = nla_get_u16(rinfo[NL80211_RATE_INFO_BITRATE]);
+                } else {
+                    LOG(DEBUG) << "NL80211_RATE_INFO_BITRATE attribute is missing";
+                }
+
+                return rate;
+            };
+
+            if (sinfo[NL80211_STA_INFO_TX_BITRATE]) {
+                sta_info.tx_bitrate_100kbps =
+                    parse_bitrate_func(sinfo[NL80211_STA_INFO_TX_BITRATE]);
+            } else {
+                LOG(DEBUG) << "NL80211_STA_INFO_TX_BITRATE attribute is missing";
+            }
+
+            if (sinfo[NL80211_STA_INFO_RX_BITRATE]) {
+                sta_info.rx_bitrate_100kbps =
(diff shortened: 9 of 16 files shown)


common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,147 @@
...
+    static auto nl_ack_cb = [](struct nl_msg *msg, void *arg) -> int {
+        int *error = (int *)arg;
+        *error     = 0;
+        return NL_STOP;
+    };
+
+    // Passing a lambda with capture is not supported for standard C function
+    // pointers. As a workaround, we create a static (but thread local) wrapper
+    // function that calls the capturing lambda function.
+    static __thread std::function<int(struct nl_msg * msg, void *arg)> nl_handler_cb_wrapper;
```
ghost: Why not use the standard keyword `thread_local`? https://en.cppreference.com/w/cpp/keyword/thread_local

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include <netlink/netlink.h>
+
+#include <easylogging++.h>
+
+namespace bwl {
+
+netlink_socket::netlink_socket(int protocol) : m_protocol(protocol)
+{
+    // Allocate the netlink socket structure
+    m_nl_socket = nl_socket_alloc();
```
rmelotte: Why couldn't m_nl_socket also be using a smart pointer instead of a raw pointer?

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,147 @@
...
+netlink_socket::~netlink_socket()
+{
+    // Deallocate the netlink socket structure
+    if (m_nl_socket) {
+        nl_socket_free(m_nl_socket);
+    }
+}
+
+bool netlink_socket::connect()
+{
```
rmelotte: Maybe it would make sense to call connect from the constructor, and close from the destructor? This way you know that if you hold a netlink_socket, it's connected and you can use it, and similarly you don't need to worry about calling close, it will be done when destructing the object.

However for that to be possible, we have to allow the usage of exceptions so that the constructor correctly fails if it cannot connect to the socket.

I would be interested to have other's opinion on the matter.

Alternatively if we really can't use exceptions, we could probably at least close the socket from the destructor if it's not already closed.

rmelotte: EDIT: in fact after reading [the documentation of libnl](http://www.infradead.org/~tgr/libnl/doc/api/group__send__recv.html#gac86a4e3b131a4e8ae21b6e28a4713748) it looks like you don't need to worry about calling close() at all if you call nl_socket_free().

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,147 @@
...
+    // Terminate connection and close socket
+    nl_close(m_nl_socket);
+}
+
+bool netlink_socket::send_receive_msg(std::function<bool(struct nl_msg *msg)> msg_create,
+                                      std::function<bool(struct nl_msg *msg)> msg_handle)
+{
+    // The Netlink message to send
+    std::shared_ptr<nl_msg> nl_message =
+        std::shared_ptr<struct nl_msg>(nlmsg_alloc(), [](struct nl_msg *obj) {
```
rmelotte: Is there a reason to use a smart_ptr instead of a unique_ptr?
It doesn't look like you share the ownership of nl_message with anything else?

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,147 @@
...
+            if (obj) {
+                nlmsg_free(obj);
+            }
+        });
+    LOG_IF(!nl_message, FATAL) << "Failed creating netlink message!";
+
+    // The Netlink callback set
+    std::shared_ptr<nl_cb> nl_callback =
```
rmelotte: Same question than for nl_message: why a shared_ptr and not a unique_ptr?

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: Changed.

common/beerocks/bwl/shared/nl80211_socket.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include "nl80211_socket.h"
+
+#include <netlink/genl/ctrl.h>
+#include <netlink/genl/family.h>
+#include <netlink/genl/genl.h>
+
+#include <easylogging++.h>
+
+#define NETLINK_BUFFER_SIZE (8192)
```
rmelotte: Why a macro and not a constexpr?

common/beerocks/bwl/shared/netlink_socket.cpp
rmelotte: (we could even avoid the need for the destructor, copy constructor, and assignment operator if the free was handled by the unique_ptr deleter)

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,168 @@
...
+#include <linux/nl80211.h>
+#include <net/if.h>
+#include <netlink/genl/genl.h>
+
+#include <math.h>
+
+namespace bwl {
+
+nl80211_client_impl::nl80211_client_impl(const std::shared_ptr<nl80211_socket> &socket)
+    : m_socket(socket)
```
rmelotte: Maybe I'm missing something here, but I think nl_80211_client_impl would be the only one holding the socket right?
In that case and if you don't want to use exceptions and create the socket from this constructor, you could still make it take a unique_ptr, to make it clear that nl_80211_client_impl takes the ownership of the socket (you would need to explicitly std::move it).

common/beerocks/bwl/shared/nl80211_client_factory.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+    // Connect NL80211 socket (it will be automatically closed when it is not needed any more)
+    if (!socket->connect()) {
+        LOG(ERROR) << "Failed to connect nl80211 socket!";
+        return nullptr;
+    }
+
+    // Create NL80211 client to send standard NL80211 commands to WiFi driver using NL80211 socket
+    return std::make_unique<nl80211_client_impl>(socket);
+}
```
rmelotte, line 32: Ok now I understand the need for the factory: you don't want to have to handle the call to `connect` everytime.
Note that the error will be printed twice: once inside the connect function, and once here.
Again it would be easier to handle using exceptions  (and would also avoid the need for the factory), but of course it's not related to this PR since exceptions have been avoided up to know.

Maybe it's time for another discussion/poll? : 
Should we keep on excluding exceptions from the code?
If yes why?
If not why?

@tomereli 
@fvbogaert 
@arnout 
@morantr

common/beerocks/bwl/dummy/nl80211_client_factory_dummy.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include "nl80211_client_dummy.h"
+
+#include <bcl/beerocks_backport.h>
+
+namespace bwl {
+
+std::unique_ptr<nl80211_client> nl80211_client_factory::create_instance()
+{
+    return std::make_unique<nl80211_client_dummy>();
+}
```
rmelotte, line 20: This factory doesn't look very useful... I guess it's there for consistency with other implementations?

rmelotte (approved):
Everything is documented :heavy_check_mark: 
Dynamic allocation uses smart pointers :heavy_check_mark: 
It looks excellent! 

I have a few comments you might want to read though so I'm approving, but also adding the "don't merge" tag to prevent Mergify from merging it automatically in cases someone else approves it in the meantime.

Overall I'm also not a big fan of having everything in a single commit, but I don't think it's a blocker.

edjaguilar:
@rmelotte asked me to review this change as well.
I just have one comment to add, I changed the CMakeLists to search dynamically for directories with the structure of the DUMMY bwl, based on strings in the path. If I merge this change to the econet bwl later I might remove the separate case for DUMMY bwl.

common/beerocks/bwl/shared/nl80211_client_factory.cpp
ghost: @rmelotte your wish is my command ;-)

https://www.strawpoll.me/19474302

tomereli: I think google can provide plenty of reasons why using exceptions in embedded is bad, and plenty of why exceptions are great.
If using exceptions makes our binaries larger - then its off the table.
But even if its not, the fact remains that prplmesh is built with a strict "no exceptions" - so adding them now will just make the code a mess.
So - my vote is not to add exceptions but I'm happy to hear other suggestions.

3r5l41

common/beerocks/bwl/shared/netlink_socket.cpp
tomereli: I think that the reason for that is since `nl_socket_alloc` is a libnl library function.

tomereli (changes requested):
Very good job documenting everything.
I think this should have been split up to more commits, but I don't think its worth it at this point.
I don't really like the fact we need to treat BWL DUMMY mode differently in the CMAKE. The fact that we implement this in `shared` leads to code duplication (that you probably mean to remove in future PR), but I think it should be handled as part of this PR.

The way I would do it, is by introducing another level of inheritance to the DWPAL implementation like @vitalybu suggested when we discussed this - have BWL DWPAL derive from BWL nl80211, and override everything *except* for the nl80211 standard commands.
The BWL Dummy should just implement stubs like you did in this PR.
I mark this as needs changes so @arnout and @vitalybu can comment on the direction in which this is going.

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -0,0 +1,81 @@
...
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_sta_info(const std::string &local_interface_name,
+                              const sMacAddr &sta_mac_address, sNl80211StaInfo &sta_info) = 0;
+
+    // TODO: add remaining methods
+};
+
+} // namespace bwl
+
```
vitalii-komisarenko, line 80: I guess a dummy and a real clients should have similar/same interface. 
I suggest having the header file shared between the two implementations or making both implementations child classes of the same base class.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,147 @@
...
+    int rc = nl_send_auto_complete(m_nl_socket, nl_message.get());
+    if (rc < 0) {
+        LOG(ERROR) << "Failed to send netlink message! Error: " << rc;
+        return false;
+    }
+
+    // Receive the response messages
+    // Note that call to nl_recvmsgs() is blocking and loop terminates when one of the callback
+    // functions sets error to 0 (ok) or to a value lower than 0 (error)
+    while (error > 0) {
```
vitalii-komisarenko, line 142: I guess `error` should be of atomic_t type.

common/beerocks/bwl/shared/nl80211_client_factory.cpp
vitalii-komisarenko: AFAIK exception support in C++ causes overhead even if they are not used. 
So we should either use them everywhere or not use them anywhere (if performance is important to us).

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -103,6 +105,18 @@ elseif(BWL_TYPE STREQUAL "NL80211")
...
+elseif(BWL_TYPE STREQUAL "DUMMY")
+
+    set(bwl_platform_sources
+        ${bwl_platform_sources}
+        ${MODULE_PATH}/dummy/nl80211_client_dummy.cpp
+        ${MODULE_PATH}/dummy/nl80211_client_factory_dummy.cpp
+        ${MODULE_PATH}/dummy/mon_wlan_hal_dummy.cpp
+        ${MODULE_PATH}/dummy/ap_wlan_hal_dummy.cpp
+        ${MODULE_PATH}/dummy/base_wlan_hal_dummy.cpp
+        ${MODULE_PATH}/dummy/sta_wlan_hal_dummy.cpp
```
vitalybu, line 117: Why not use `file(GLOB...)` to  grab all `*.cpp` files inside the `dummy` folder?

common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
```diff
@@ -0,0 +1,50 @@
...
+
+nl80211_client_dummy::nl80211_client_dummy() {}
+
+nl80211_client_dummy::~nl80211_client_dummy() {}
+
+bool nl80211_client_dummy::get_sta_info(const std::string &local_interface_name,
+                                        const sMacAddr &sta_mac_address, sNl80211StaInfo &sta_info)
+{
+    (void)local_interface_name;
+    (void)sta_mac_address;
```
vitalybu, line 22: Adding a comment explaining that it's done to suppress the unused variables warning is probably a good idea

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -0,0 +1,81 @@
...
+namespace bwl {
+
+/**
+ * @brief Station information
+ *
+ * Information obtained with NL80211_CMD_GET_STATION command through a NL80211 socket.
+ * See NL80211_STA_INFO_* in <linux/nl80211.h> for a description of each field.
+ */
+struct sNl80211StaInfo {
```
vitalybu: If it's used only by the "client" type classes, maybe consider putting inside the class and renaming to "sStaInfo".
This way it'll be accessed by `nl80211_client::sStaInfo`

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -0,0 +1,81 @@
...
+     * Station information contains basically metrics associated to the link between given local
+     * interface and a the interface of a station whose MAC address is 'sta_mac_address'.
+     *
+     * @param[in] local_interface_name Name of the local interface.
+     * @param[in] sta_mac_address MAC address of a station connected to the local interface.
+     * @param[out] sta_info Station information.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_sta_info(const std::string &local_interface_name,
```
vitalybu, line 73: Instead of passing the interface name to every function, perhaps a better implementation would be to pass it once in the constructor (similarly to the AP implementation of BWL).

common/beerocks/bwl/shared/netlink_socket.cpp
vitalybu: Completely agree with @rmelotte here. We do this a lot in other places.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+namespace bwl {
+
+netlink_socket::netlink_socket(int protocol) : m_protocol(protocol)
+{
+    // Allocate the netlink socket structure
+    m_nl_socket = nl_socket_alloc();
+    if (!m_nl_socket) {
+        LOG(ERROR) << "Failed to allocate netlink socket!";
+    }
```
vitalybu:
```suggestion
    LOG_IF(!m_nl_socket, ERROR) << "Failed to allocate netlink socket!";
```

common/beerocks/bwl/shared/netlink_socket.cpp
vitalybu: Nope, it is set and read from the same context

vitalybu (commented):
Overall looks good, but I do have some comments, so not approving just yet

common/beerocks/bwl/CMakeLists.txt
tomereli: I think @arnout said something against globbing sometime in the past.

common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
tomereli: Using `__attribute__((unused))` is probably an even better idea (nitpick)

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: Very good point. Done!

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: I'd rather leave it as it is. Although it might never be the case, you could create an instance of `netlink_socket` once and later call `connect` and `close` as many times you wanted. But most important of all is that derived classes can configure the socket in their constructor **before** the call to `connect` can ever be made. That's just the case with derived class `nl80211_socket`, which increases the socket's internal buffer size in its constructor.

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: You are right. I've changed shared_ptr to unique_ptr and simplified code.
I've inspected the source code of the `nlmsg_free` and checked that it does nothing if parameter is NULL.

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: You are right. I've changed shared_ptr to unique_ptr and simplified code.
I've inspected the source code of the `nl_cb_put` and checked that it does nothing if parameter is NULL.

common/beerocks/bwl/shared/nl80211_socket.cpp
mariomaz: Done

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_factory.cpp
mariomaz: I don't want to start a war flame here about the use exceptions but, just for the record, I like them and therefore voted "yes".
As an advantage and just because @rmelotte has mentioned it right here, we shouldn't bother about if an error message has been printed to the log or not: log it when you catch it, no need to clutter the code with log calls everywhere. 
I also understand @tomereli's position: use of exceptions is something to be decided at the beginning of the project and if we added them now, we'd end up having prplMess instead of prplMesh.

mariomaz: Regarding the use of a factory: its just a static method to create a client with which to send commands to WiFi driver. There are two different factories: one for the Dummy flavor of the BWL which creates dummy clients and one for the rest of flavors which creates real clients (those that use a socket to communicate with the driver).
Socket is connected at the beginning and stays open all the time for performance reasons. Documentation of `nl80211_client_impl` warns that socket must be connected before calling its methods, but it could have been implemented in such a way that socket was connected and closed on each call (or even created/connected/used/closed/destroyed on each call).

common/beerocks/bwl/dummy/nl80211_client_factory_dummy.cpp
mariomaz: As said, this is the implementation of the factory in the Dummy flavor of the BWL library. The factory is defined to return an instance of a class that implements the interface `nl80211_client`. This particular implementation returns an instance of `nl80211_client_dummy`. Methods of this class do nothing but return dummy data.

mariomaz:
> Typo - DWPAL

Fixed


> Missing line break

Fixed



> Can `m_socket` be `nullptr` here?

Fixed

mariomaz:
> Very good job documenting everything.
> I think this should have been split up to more commits, but I don't think its worth it at this point.
> I don't really like the fact we need to treat BWL DUMMY mode differently in the CMAKE. The fact that we implement this in `shared` leads to code duplication (that you probably mean to remove in future PR), but I think it should be handled as part of this PR.
> 
> The way I would do it, is by introducing another level of inheritance to the DWPAL implementation like @vitalybu suggested when we discussed this - have BWL DWPAL derive from BWL nl80211, and override everything _except_ for the nl80211 standard commands.
> The BWL Dummy should just implement stubs like you did in this PR.
> I mark this as needs changes so @arnout and @vitalybu can comment on the direction in which this is going.

I evaluated @vitalybu's suggestion but decided to go with this solution because it's much simpler. Also, if BWL DWPAL derives from BWL nl80211, then it will take longer to compile and the binary will be bigger. 

BWL DWPAL, BWL nl80211 and whatever other part of the application can all be users of the `nl80211_client_factory`, create an instance of a class implementing the `nl80211_client` interface and use it to send NL80211 standard commands. Although I prefer to inject the dependencies to facilitate unit-testing (if we ever had to), so most probably BWL DWPAL and BWL nl80211 classes that send NL80211 standard commands will have a `nl80211_client` parameter in their constructor.

Another goal of this design is that it can be easily extended to send NETLINK_ROUTE commands (probably with a `netlink_route_client`). Maybe we should move all these classes from BWL to BPL because they are not intended for WiFi only.

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: That's exactly what we have: `nl80211_client_dummy` and `nl80211_client_impl` both implement the `nl80211_client` interface. `nl80211_client` interface is defined in nl80211_client.h which is located in bwl/include so it's shared (and public).

common/beerocks/bwl/CMakeLists.txt
mariomaz: I saw that files for the Dummy flavor of the BWL where picked differently than in the other flavors of the library and stuck to it without even questioning why. Shall I change it then?

common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
mariomaz: Although we will never use another compiler, `__attribute__((unused))` is a proprietary GCC extension, so I'll stick to `(void)` and add a comment

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Some methods will not require the interface name, so it cannot be a parameter to the constructor.

common/beerocks/bwl/include/bwl/nl80211_client.h

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: Done

common/beerocks/bwl/CMakeLists.txt
arnout: See the NOTE in the glob section of the [CMake documentation](https://cmake.org/cmake/help/v3.1/command/file.html):

> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

I have been hit by this a couple of times already in prplMesh. So indeed, I'd prefer to remove all GLOBs.

arnout:
> > The way I would do it, is by introducing another level of inheritance to the DWPAL implementation like @vitalybu suggested when we discussed this - have BWL DWPAL derive from BWL nl80211, and override everything _except_ for the nl80211 standard commands.

Note that this looks like implementation inheritance, and 25 years ago I was thought that [this is evil](http://whats-in-a-game.com/implementation-inheritance-is-evil/).


> > The BWL Dummy should just implement stubs like you did in this PR.
> > I mark this as needs changes so @arnout and @vitalybu can comment on the direction in which this is going.
> 
> I evaluated @vitalybu's suggestion but decided to go with this solution because it's much simpler. Also, if BWL DWPAL derives from BWL nl80211, then it will take longer to compile and the binary will be bigger.
> 
> BWL DWPAL, BWL nl80211 and whatever other part of the application can all be users of the `nl80211_client_factory`, create an instance of a class implementing the `nl80211_client` interface and use it to send NL80211 standard commands. Although I prefer to inject the dependencies to facilitate unit-testing (if we ever had to), so most probably BWL DWPAL and BWL nl80211 classes that send NL80211 standard commands will have a `nl80211_client` parameter in their constructor.
> 
> Another goal of this design is that it can be easily extended to send NETLINK_ROUTE commands (probably with a `netlink_route_client`). Maybe we should move all these classes from BWL to BPL because they are not intended for WiFi only.

Please include this reasoning (more or less copy&paste of the text above) in one of the commit messages.

common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
arnout: We'll only ever use GCC *or clang*. But both support `__attribute__((unused))`.

The typical way to handle this is something like

```
#if defined(__GCC__) || defined(__CLANG__)
#define UNUSED __attribute__((unused))
#else
#define UNUSED
#endif
```

However, for this particular case, it's about unused parameters, and we pass `-Wno-unused-parameter` in CMAKE_C_FLAGS, so why do we need this at all? Note BTW that `-Wunused-parameter` is a `-Wextra` warning because it is very common that parameters are unused.

tomereli: The reason we should still support it is that when prplMesh is built in-tree (OpenWRT, prplWrt, UGW, RDKB), these flags may be passed from the build system (unless I'm mistaken) - in which case I think we should be on the safe side and be strict with our CMAKE flags default settings.
I think the typical way of handling this @arnout suggested is a good compromise.

common/beerocks/bwl/shared/README.md
```diff
@@ -0,0 +1,6 @@
+# Shared files directory
+
+The `shared` directory contains classes shared among all the implementations that use actual WiFi radios (i.e.: the BWPAL and NL80211 flavors but not the Dummy one).
```
tomereli: Might be worth adding documentation to the README.md for the nl80211 support which is added in this PR.
It might be useful to read through the README.md which will serve as the DR for the implementation before browsing the code.
Anyway, this is one hell of a nitpick, so ignore it if you like :)

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -18,9 +18,7 @@ namespace bwl {
 netlink_socket::netlink_socket(int protocol)
     : m_nl_socket(nl_socket_alloc(), nl_socket_free), m_protocol(protocol)
 {
-    if (!m_nl_socket) {
-        LOG(ERROR) << "Failed to allocate netlink socket!";
-    }
+    LOG_IF(!m_nl_socket, ERROR) << "Failed to allocate netlink socket!";
```
tomereli, line 21:
```suggestion
    LOG_IF(!m_nl_socket, FATAL) << "Failed to allocate netlink socket!";
```

This will allow aborting the application if `el::Loggers::addFlag(el::LoggingFlag::DisableApplicationAbortOnFatalLog);` is removed from our `beerocks_logging::apply_settings()` implementation.

@vitalybu why don't we want to abort on FATAL? Letting constructors fail on abort in such cases is much better than just having an error log and fail later.

tomereli (approved):
A couple of nitpicks but strong LGTM!

common/beerocks/bwl/include/bwl/nl80211_client.h
vitalybu: Please correct me if I'm wrong, but as far as I understand, all methods will read some information or perform an operation of a client interface. What methods do you think will be interface agnostic?

common/beerocks/bwl/shared/netlink_socket.cpp
vitalybu: We don't always want to abort the application.
If for some reason prplMesh will run on a platform without access to netlink sockets, it could still function (to some extent).

tomereli: OK, valid point. Lets keep it as `ERROR` - but I think we should remove `el::Loggers::addFlag(el::LoggingFlag::DisableApplicationAbortOnFatalLog)` and start using `LOG(FATAL)` appropriately.

common/beerocks/bwl/dummy/nl80211_client_factory_dummy.cpp
arnout: (nitpick) the rest of the bwl uses a factory function instead of a factory class, e.g. `ap_wlan_hal_create()`. It would be more consistent to use a factory function `nl80211_client_create()` here as well.

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Programming to an interface allows dependent classes to remain unaware of the different
+ * implementations of the interface as well as facilitates the creation of mock implementations for
+ * unit testing.
+ *
+ * This is a C++ interface: an abstract class that is designed to be specifically used as a base
+ * class and which derived classes (implementations) will override each pure virtual function.
+ *
+ * Known implementations: nl80211_client_impl (uses NL80211 socket), nl80211_client_dummy (fake
+ * implementation that returns dummy data).
```
arnout, line 31: Very good documentation, thank you!

common/beerocks/bwl/include/bwl/nl80211_client.h
arnout: I think there will be only one instance of this class per radio, but the local_interface_name is the VAP inteface name (@mariomaz please correct me if I'm wrong). So it can't be part of the constructor.

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -0,0 +1,81 @@
...
+     */
+    virtual ~nl80211_client() = default;
+
+    /**
+     * @brief Gets station information.
+     *
+     * Station information contains basically metrics associated to the link between given local
+     * interface and a the interface of a station whose MAC address is 'sta_mac_address'.
+     *
+     * @param[in] local_interface_name Name of the local interface.
```
arnout: Please specify if this is the VAP or the radio interface.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    LOG_IF(!m_nl_socket, ERROR) << "Failed to allocate netlink socket!";
+}
+
+netlink_socket::~netlink_socket() {}
+
+bool netlink_socket::connect()
+{
+    // Connect the socket
+    if (nl_connect(m_nl_socket.get(), m_protocol) != 0) {
```
arnout, line 32: (For my understanding, not a change request:) Why is the explicit get() needed here? Is it because nl_connect is an extern C that there's no automatic conversion?

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // Set the callbacks
+    nl_cb_err(nl_callback.get(), NL_CB_CUSTOM, nl_err_cb, &error); // error
+    nl_cb_set(nl_callback.get(), NL_CB_FINISH, NL_CB_CUSTOM, nl_finish_cb,
+              &error);                                                        // finish
+    nl_cb_set(nl_callback.get(), NL_CB_ACK, NL_CB_CUSTOM, nl_ack_cb, &error); // ack
+    nl_cb_set(nl_callback.get(), NL_CB_VALID, NL_CB_CUSTOM, nl_handler_cb,
+              nullptr); // response handler
+
+    // Send the netlink message
+    int rc = nl_send_auto_complete(m_nl_socket.get(), nl_message.get());
```
arnout, line 128: Maybe it would be nice if we did an automatic connect here, i.e. if there was an additional flag to keep track of the connected state of the socket (or maybe there's a libnl function for that), and call connect() if not already connected.

Not something we need to do now but a nice idea to improve if it is needed/useful for some user.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+{
+    // Terminate connection and close socket
+    nl_close(m_nl_socket.get());
+}
+
+bool netlink_socket::send_receive_msg(std::function<bool(struct nl_msg *msg)> msg_create,
+                                      std::function<bool(struct nl_msg *msg)> msg_handle)
+{
+    // The Netlink message to send
+    std::unique_ptr<nl_msg, void (*)(nl_msg *)> nl_message(nlmsg_alloc(), nlmsg_free);
```
arnout, line 59: Would it be useful (and even possible) to add a local

```
template<class T> using unique_ptr = std::uniq_ptr<T, void (*)(T *)>;

...

    unique_ptr<nl_msg> nl_message(nlmsg_alloc(), nlmsg_free);
```

(if that even works...)

(nitpick, of course)

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+    LOG_IF(!nl_message, FATAL) << "Failed creating netlink message!";
```
arnout: So, just to be clear: if allocation fails, we abort here instead of returning false. Except if we remove the AbortOnFatalLog flag, in which case we will log an error and continue as if the allocation succeeded.

The second option is obviously not acceptable. So, either we say that we keep the abort flag and start relying on it to simply crash in out-of-memory conditions (which IMO is very reasonable), or we need to return false here in case allocation failed (but then it should probably not be fatal).

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // error == 0: response has been successfully received
+    // error < 0: some error occurred while receiving response
+    // Final value is used to compute the result code of this method.
+    int error = 1;
+
+    // Create standard callbacks
+    static auto nl_err_cb = [](struct sockaddr_nl *nla, struct nlmsgerr *err, void *arg) -> int {
+        int *error = (int *)arg;
+        *error     = err->error;
+        LOG(ERROR) << "Failed to process netlink message! " << strerror(*error) << " (" << *error
```
arnout: There's something wrong here. The comment above says that `error < 0` indicates error. However, the error numbers expected by `strerror` are positive number; e.g. EPERM is 1, not -1.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    static auto nl_ack_cb = [](struct nl_msg *msg, void *arg) -> int {
+        int *error = (int *)arg;
+        *error     = 0;
+        return NL_STOP;
+    };
+
+    // Passing a lambda with capture is not supported for standard C function
+    // pointers. As a workaround, we create a static (but thread local) wrapper
+    // function that calls the capturing lambda function.
```
arnout, line 101: AFAIU, this stuff is needed for dwpal because dwpal is not reentrant.

However, the libnl functions take a void* context argument. So it is perfectly possible to wrap all this into an internal class that converts the void* back to the proper class and calls the callback member on the class. This class could even be exported, so the msg_handle can derive from it instead of being a function.

Anyway, nitpick.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+    };
+
+    // Call the user's message create function
+    if (!msg_create(nl_message.get())) {
+        LOG(ERROR) << "User's netlink create function failed!";
+        return false;
+    }
+
+    // Set the callbacks
+    nl_cb_err(nl_callback.get(), NL_CB_CUSTOM, nl_err_cb, &error); // error
```
arnout, line 120: Those comments are pretty redundant... `nl_cb_err` obviously sets the error callback.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+    // Send the netlink message
+    int rc = nl_send_auto_complete(m_nl_socket.get(), nl_message.get());
+    if (rc < 0) {
+        LOG(ERROR) << "Failed to send netlink message! Error: " << rc;
+        return false;
+    }
+
+    // Receive the response messages
+    // Note that call to nl_recvmsgs() is blocking and loop terminates when one of the callback
+    // functions sets error to 0 (ok) or to a value lower than 0 (error)
```
arnout: I think in the future we might want to refactor this to be non-blocking, by making a separate send and receive function. But that's not needed now.

More importantly, though: as this comment notes, nl_recvmsgs will continue reading until the callback returns STOP or nl_recv() returns 0 or error. nl_recv() can only return 0 if the socket is non-blocking, but we create a blocking socket. So I don't see why the while loop is needed. Can you add a comment that explains that?

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -0,0 +1,128 @@
...
+        *error     = err->error;
+        LOG(ERROR) << "Failed to process netlink message! " << strerror(*error) << " (" << *error
+                   << ")";
+        return NL_STOP;
+    };
+
+    static auto nl_finish_cb = [](struct nl_msg *msg, void *arg) -> int {
+        int *error = (int *)arg;
+        *error     = 0;
+        return NL_SKIP;
```
arnout, line 90: Shouldn't this be NL_STOP to stop the nl_recvmsgs() function?

common/beerocks/bwl/shared/netlink_socket.h
```diff
@@ -0,0 +1,112 @@
...
+
+public:
+    /**
+     * @brief Class constructor.
+     *
+     * Allocates the Netlink socket structure.
+     *
+     * @param protocol Netlink protocol to use. Possible values are listed in <linux/netlink.h>
+     */
+    explicit netlink_socket(int protocol);
```
arnout, line 49: In addition to the explicit, I think we also need

```
netlink_socket() = delete;
netlink_socket(const netlink_socket&) = delete;
```

common/beerocks/bwl/shared/netlink_socket.h
```diff
@@ -0,0 +1,112 @@
...
+    virtual ~netlink_socket();
+
+    /**
+     * @brief Connects the Netlink socket.
+     *
+     * Creates a new Netlink socket and binds it to the protocol specified in class constructor.
+     *
+     * @return True on success and false otherwise (e.g.: socket is already connected).
+     */
+    virtual bool connect();
```
arnout, line 65: I don't think it's very useful to make these or any of the other functions virtual.

Specifically for this one, I think it's more useful to add a `virtual bool post_connect() {}` that is overridden in nl80211_socket.

common/beerocks/bwl/shared/netlink_socket.h
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * @brief Closes the Netlink socket.
+     *
+     * Call this method to terminate connection and free resources.
+     */
+    virtual void close();
```
arnout, line 72: So here a `pre_close()` would be more useful than making close itself virtual. However, don't add it now since it's not used.

common/beerocks/bwl/shared/netlink_socket.h
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * Message to send is created using provided creation callback function and received message
+     * is parsed using provided handling callback function.
+     *
+     * @param[in] msg_create Message creation callback function (called on message to be sent).
+     * @param[in] msg_handle Message handling callback function (called on received message).
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool send_receive_msg(std::function<bool(struct nl_msg *msg)> msg_create,
```
arnout, line 90: Again, I don't think it is useful to make this virtual. Overridden implementations will behave differently than what is specified in the documentation above, e.g. the nl80211 override already creates the first attribute in the message, and it requires additional arguments.

common/beerocks/bwl/shared/netlink_socket.h
```diff
@@ -0,0 +1,112 @@
...
+#define __BWL_NETLINK_SOCKET_H__
+
+#include <functional>
+#include <memory>
+
+/**
+ * Forward declarations
+ */
+struct nl_sock;
+struct nl_msg;
```
arnout, line 18: (nitpick) since this is only used for *real* netlink implementations, wouldn't it make sense to include the netlink headers that define those structures here?

common/beerocks/bwl/shared/nl80211_client_factory.cpp
arnout: Just to summarize this discussion:

- We need something like `nl80211_client_factory::create_instance()` to create an instance of the abstract nl80211_client class. (It could have been a factory function instead of a factory class, but that's just a detail.)
- We want connect() to be done already in its return value. (That bit could be avoided by doing the connect() automatically in the send_receive_message function if it is not connected already.)
- We can't do connect() in the nl80211_client_impl constructor because it may fail and you can't fail in a constructor (you can with an exception, but even that is a bit iffy and we anyway don't want exceptions).

For these reasons, we have a factory and it has the two std::move calls.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -103,6 +105,18 @@ elseif(BWL_TYPE STREQUAL "NL80211")
 
     list(APPEND BWL_LIBS nl-tiny)
 
+elseif(BWL_TYPE STREQUAL "DUMMY")
+
+    set(bwl_platform_sources
+        ${bwl_platform_sources}
+        ${MODULE_PATH}/dummy/nl80211_client_dummy.cpp
+        ${MODULE_PATH}/dummy/nl80211_client_factory_dummy.cpp
```
arnout, line 113: I wonder why we even need a dummy instance of the nl80211_client. nl80211_client is only going to be used internally inside a bwl implementation, right?

Without the dummy implementation, we also don't need the factories and the abstract base classes, since there's only one implementation.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,173 @@
...
+{
+}
+
+nl80211_client_impl::~nl80211_client_impl() {}
+
+bool nl80211_client_impl::get_sta_info(const std::string &local_interface_name,
+                                       const sMacAddr &sta_mac_address, sStaInfo &sta_info)
+{
+    static struct nla_policy stats_policy[NL80211_STA_INFO_MAX + 1];
+    stats_policy[NL80211_STA_INFO_INACTIVE_TIME] = {NLA_U32, 0, 0};
```
arnout, line 31: Can't we use an array constructor here or is that not standard C++?

```
    static cont struct nla_policy stats_policy[NL80211_STA_INFO_MAX + 1] = {
        [NL80211_STA_INFO_INACTIVE_TIME] = {NLA_U32, 0, 0},
...
```

The problem with this implementation is that we define a static object that is overwritten with each call to this function, so it's not really reentrant.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    // Get the interface index for given interface name
+    int iface_index = if_nametoindex(local_interface_name.c_str());
+    if (0 == iface_index) {
+        LOG(ERROR) << "Failed to read the index of interface " << local_interface_name << ": "
+                   << strerror(errno);
+
+        return false;
+    }
```
arnout, line 61: Maybe not for this PR, but I would really rather have that we use ifindex rather than ifname in all our code. ifindex is monotonically incrementing when interfaces are removed and re-added, but an ifname can "change" under the hood due to removals or renames. So I feel that using an interface name is unreliable.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            // Parse nested station stats
+            if (nla_parse_nested(sinfo, NL80211_STA_INFO_MAX, tb[NL80211_ATTR_STA_INFO],
+                                 stats_policy)) {
+                LOG(ERROR) << "Failed to parse nested attributes!";
+                return false;
+            }
+
+            // Inactive time
```
arnout: This comment just repeats what is pretty explicit in the code below.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,173 @@
...
+            if (sinfo[NL80211_STA_INFO_INACTIVE_TIME]) {
+                sta_info.inactive_time_ms = nla_get_u32(sinfo[NL80211_STA_INFO_INACTIVE_TIME]);
+            }
```
arnout: I think we should either explicitly clear sta_info in the beginning, or have an else branch here to set a default value.

Probably it would be useful to have a helper function `uint32 get_u32(struct nlattr *attr, uint32 default_value);` that handles the condition.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,173 @@
...
+                sta_info.signal_dBm = nla_get_u8(sinfo[NL80211_STA_INFO_SIGNAL]);
+            }
+
+            // Signal avg
+            if (sinfo[NL80211_STA_INFO_SIGNAL_AVG]) {
+                sta_info.signal_avg_dBm = nla_get_u8(sinfo[NL80211_STA_INFO_SIGNAL_AVG]);
+            }
+
+            // Bit rate parsing helper function
+            auto parse_bitrate_func = [&](struct nlattr *bitrate_attr) -> int {
```
arnout, line 150: I have a negative prejudice against using lambdas for local function definitions, but this is actually quite elegant!

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,173 @@
...
+                if (nla_parse_nested(rinfo, NL80211_RATE_INFO_MAX, bitrate_attr, rate_policy)) {
```
arnout: AFAICS rinfo is used only locally within this function and shouldn't carry state from one call to the next, so I'd make it a local variable of this function.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -0,0 +1,173 @@
...
+                    LOG(ERROR) << "Failed to parse nested rate attributes!";
```
arnout: We don't report anything when any of the other attributes is missing, so why an error here? Actually, I think there should be an INFO message in all cases when it's missing.

common/beerocks/bwl/shared/nl80211_client_impl.h
```diff
@@ -0,0 +1,69 @@
...
+public:
+    /**
+     * @brief Class constructor.
+     *
+     * Note: provided socket must be connected before calling any of the methods in this class.
+     *
+     * @param[in] socket NL80211 socket to send messages and receive responses to/from the WiFi
+     * driver.
+     */
+    explicit nl80211_client_impl(std::unique_ptr<nl80211_socket> socket);
```
arnout, line 38: I think we should also delete the two default constructors.

common/beerocks/bwl/shared/nl80211_client_impl.h
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * @brief Class destructor.
+     */
+    virtual ~nl80211_client_impl();
+
+    /**
+     * @brief Gets station information.
+     *
+     * Station information contains basically metrics associated to the link between given local
+     * interface and a the interface of a station whose MAC address is 'sta_mac_address'.
```
arnout:
```suggestion
     * Station information contains basic metrics associated to the link between given local
     * interface and the interface of a station with MAC address 'sta_mac_address'.
```

common/beerocks/bwl/shared/nl80211_socket.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include "nl80211_socket.h"
+
+#include <netlink/genl/ctrl.h>
+#include <netlink/genl/family.h>
+#include <netlink/genl/genl.h>
+
+#include <easylogging++.h>
+
+/**
+ * Receive and transmit socket buffer size in bytes.
```
arnout, line 18: Could be useful to mention why this is changed from the default.

common/beerocks/bwl/shared/nl80211_socket.cpp
```diff
@@ -0,0 +1,85 @@
...
+    bool result = netlink_socket::connect();
+
+    // Resolve the generic nl80211 family id
+    if (result) {
+        const char *family_name = "nl80211";
+
+        m_family_id = genl_ctrl_resolve(m_nl_socket.get(), family_name);
+        if (0 > m_family_id) {
+            LOG(ERROR) << "'" << family_name << "' family not found!";
+            result = false;
```
arnout, line 57: So in my proposal, this bit would go into `post_connect()` and the error handling (i.e. calling close()) would go into the non-virtual `connect()` function.

common/beerocks/bwl/shared/nl80211_socket.h
```diff
@@ -0,0 +1,80 @@
...
+#define __BWL_NL80211_SOCKET_H__
+
+#include "nl_genl_socket.h"
+
+namespace bwl {
+
+/**
+ * @brief Extension to the generic Netlink protocol class for NL80211.
+ *
+ * The NL80211 is the 802.11 Netlink-based userspace interface for the new cfg80211 configuration
```
arnout:
```suggestion
 * The NL80211 is the 802.11 Netlink-based userspace interface for the cfg80211 configuration
```

It was introduced in v2.6.24 so it's hardly "new" any more.

common/beerocks/bwl/shared/nl_genl_socket.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+#include "nl_genl_socket.h"
+
+#include <linux/netlink.h>
+
+namespace bwl {
+
+nl_genl_socket::nl_genl_socket() : netlink_socket(NETLINK_GENERIC) {}
+
+nl_genl_socket::~nl_genl_socket() {}
```
arnout: I personally prefer to put such empty or very simple function defintions inline, so the cpp file can be removed entirely.

Note that putting stuff inline like that not only saves code size and performance, but sometimes even compile time because the compiler can elide the call right away and simplify its intermediate representation. (In a previous life, I did research on compiler optimisations.)

common/beerocks/bwl/shared/nl_genl_socket.h
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @brief C++ wrapper class around the generic Netlink protocol library (libnl-genl).
+ *
+ * This is a very simple C++ wrapper class around the generic Netlink protocol library. Its aim is
+ * to facilitate the use of the library from C++ applications. This class extends the
+ * netlink_socket base class for the generic Netlink protocol (NETLINK_GENERIC).
+ *
+ * Known derived classes: nl80211_socket.
+ */
```
arnout, line 23: With no functionality other than setting the protocol, I'm not sure if it's worthwhile to introduce this additional class. IMO it makes the code harder to understand with no benefit at all.

This class is only useful if later on 
- there is a second, non-nl80211 netlink socket (and there probably should be one, for nlrt, but that one will not be in bwl so code needs to be refactored anyway);
- *and* there is some common code between them that should be in this class and not in netlink_socket itself.

But in that case, we can just introduce the class at that time.

arnout (approved):
*Excellent* documentation in this PR. I wish all the code was like that.

As often, I have a truckload of comments, but nothing critical. Except for one thing: the negative `error` number. It would have been great if there were a unit test that would validate such issues :-)

Given that there is nothing critical, I'm approving, so please do the following:
- apply the fixups that you agree with and are easy to do;
- for bigger changes that you think really should be done, make an issue;
- for the hints of the type "code could be refactored like this", just leave them as unresolved - we can refer back to this PR when development on the netlink stuff continues;
- squash and rebase everything, and mergify!

common/beerocks/bwl/shared/netlink_socket.cpp
tomereli: #936

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: Honestly, I didn't know it was even possible to perform such implicit conversion from smart pointer to raw pointer type. But even if it was possible, in this case I'd rather use the explicit access.

If I remove the  call to `get()`, then I get this error when compiling for the RAX40:
```
../common/beerocks/bwl/shared/netlink_socket.cpp: In member function 'virtual bool bwl::netlink_socket::connect()':
../common/beerocks/bwl/shared/netlink_socket.cpp:29:43: error: cannot convert 'std::unique_ptr<nl_sock, void (*)(nl_sock*)>' to 'nl_sock*' for argument '1' to 'int nl_connect(nl_sock*, int)'
     if (nl_connect(m_nl_socket, m_protocol) != 0) {
                                           ^
```
I found these quite interesting articles about the topic:
- [7.7 Implicit Conversion to Raw Pointer Types](https://www.informit.com/articles/article.aspx?p=31529&seqNum=7)
- [Reader Q&A: Why don’t modern smart pointers implicitly convert to *?](https://herbsutter.com/2012/06/21/reader-qa-why-dont-modern-smart-pointers-implicitly-convert-to/)

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: I see your point but I prefer if a explicit call to connect() is required. Let's see if lazy connection it's needed/useful in the future.

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: Changed to log error and return false in that cases

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: You must think of nl80211_client as a programmatic way of doing what the `iw` command line tool does. Surely we will not implement all the `iw` options, but at least we will use the same nl80211_client instance to query information for different VAP interfaces in the same radio, or to query information for a PHY device, or simply to list available interfaces. Thus `local_interface_name` cannot be a parameter to the constructor.

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Done (it's the VAP interface name)

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: Removed call to strerror()

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: Probably it has the same effect returning NL_STOP than NL_SKIP. 

According to documentation,  NL_CB_FINISH gets called at the end of a multipart response message. According to source code, function nl_recvmsgs() continues parsing messages if NL_SKIP is received as a result of invoking callback and there are more messages to parse. So, if NL_CB_FINISH callback returns NL_SKIP and it's the last message in response, it has the same effect as if NL_STOP would have been returned.

However, extra messages after the multipart message would be ignored if NL_STOP were returned but not if NL_SKIP is returned. I don't know if this can be even possible, but just in case we should leave it as it is. All the examples I've checked in the Internet and original code in `base_wlan_hal_nl80211::send_nl80211_msg` return NL_SKIP too.

common/beerocks/bwl/shared/netlink_socket.h
mariomaz: If you try to create a `netlink_socket` using default constructor, then you get a compilation error because it's not defined:
```
error: no matching function for call to 'bwl::netlink_socket::netlink_socket()'
```
If you try to create a `netlink_socket`using copy constructor, then you also get a compilation error because such constructor is implicitly deleted due to the unique_ptr member.
```
error: use of deleted function 'bwl::netlink_socket::netlink_socket(const bwl::netlink_socket&)'
```
I can however add those constructors if there is any coding style rule about this.

common/beerocks/bwl/shared/netlink_socket.h
mariomaz: If you don't mind, I prefer to stick to current implementation unless you see a real issue with it.

Virtual methods can be overridden in derived classes as needed, in a case by case basis: adding code before or after calling base class method, or even removing/replacing the call to base class method. And the beauty of declaring the method as virtual in the base class is that we must not know all those use cases in advance. If we add a `virtual bool post_connect()` now, tomorrow we might need a `pre_connect()` method. And base class should have to be modified. Current implementation allows all such use cases in a very clean way.

common/beerocks/bwl/shared/netlink_socket.h
mariomaz: Again it's virtual just in case a derived class would ever need to override it. 

`nl80211_socket::send_receive_msg` is not overriding the method in base class with the same name because parameter type lists are not the same. 

What I could do here is to override the method in base class so it does nothing but log a message and return an error to avoid ill-formed request messages, thus breaking the contract provided by base class.

common/beerocks/bwl/shared/netlink_socket.h
mariomaz: Forward declarations reduce build times.
In this case, it's hard to say when it's worthwhile to #include the headers or the forward declarations of what is actually needed.

common/beerocks/bwl/CMakeLists.txt
mariomaz: `nl80211_client` will be used outside the BWL implementation, at least to get link metrics for links between wireless interfaces and stations. So yes, we need the dummy implementation.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: I tried your proposal, but I got a compile error:
```
../common/beerocks/bwl/shared/nl80211_client_impl.cpp: In member function 'virtual bool bwl::nl80211_client_impl::get_sta_info(const string&, const sMacAddr&, bwl::nl80211_client::sStaInfo&)':
../common/beerocks/bwl/shared/nl80211_client_impl.cpp:52:5: sorry, unimplemented: non-trivial designated initializers not supported
     };
     ^
../common/beerocks/bwl/shared/nl80211_client_impl.cpp:52:5: sorry, unimplemented: non-trivial designated initializers not supported
```

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: I've initialized the struct at the beginning

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: `nla_parse_nested()` is called after checking that corresponding attribute does exist. In such case, it should succeed but, if it doesn't, then logging the error is the appropriate action IMO.

I agree with you that an INFO message should be added to log when an expected attribute is missing. Done

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Removed

common/beerocks/bwl/shared/nl80211_client_impl.h
mariomaz: https://github.com/prplfoundation/prplMesh/pull/900#discussion_r390433136

common/beerocks/bwl/shared/nl80211_client_impl.h
mariomaz: Done

common/beerocks/bwl/shared/nl80211_socket.cpp
mariomaz: Honestly, I don't know. I've been investigating and found that the call to `nl_socket_set_buffer_size` ends up calling `setsockopt(SO_SNDBUF)` and `setsockopt(SO_RCVBUF)`. Some people report that they were losing events with default values and had to increase the receive buffer size to fix it. Some tutorials I've found also change default sizes but don't explain why.

I borrowed this code from `base_wlan_hal_nl80211::send_nl80211_msg` and thought that who initially wrote that code had considered it was necessary to call `nl_socket_set_buffer_size` so I kept it. I didn't question myself why it was actually needed nor if values used were the most appropriate.

While investigating, I've realized that the call to `nl_socket_set_buffer_size` requires the socket to be connected or otherwise it fails with NLE_BAD_SOCK. Thus I've removed the call from the constructor and put it after the socket connection has been made.

` base_wlan_hal_nl80211` is also calling `nl_socket_set_buffer_size` but before connecting the socket so I guess it must be failing too (no error code is being checked after the call) and buffer size is not being changed. Thus it looks like, for the commands being executed up to now in BWL, it's not necessary to increase buffer size.

common/beerocks/bwl/shared/nl80211_socket.h
mariomaz: Done

common/beerocks/bwl/shared/nl_genl_socket.cpp
mariomaz: Done

common/beerocks/bwl/shared/netlink_socket.cpp
mariomaz: I've borrowed this code from existing implementation where the loop was already there. I've seen some examples on the Internet and there's a loop too. 

I think (but I cannot prove it) that it's possible to receive more than one message after a request. Callbacks must handle all of them and notify termination when done.

I've realized also that return code from nl_recvmsgs wasn't being checked. An error in this call is also a condition to exit the loop.

PR #902, opened 2020-02-28
OpenWrt: build the full firmwares

rmelotte:
Instead of building only the OpenWrt SDK and making a docker image out of it, compile the full firmware.

The full OpenWrt tree is now used to compile the prplMesh ipks.

Also remove the deletion of the docker images on the Gitlab runners.
Runners will now have to take care of deleting the images after some time themselves.

The firmwares are not deployed at the moment, it will be done in an other PR.


tomereli (approved):
The only part I'm not sure about is not deleting images.

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -46,68 +62,40 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
 
 WORKDIR /home/openwrt/openwrt
 
-COPY --chown=openwrt:openwrt configs/sdk_${SDK_CONFIG}_config /home/openwrt/sdk_config
+COPY --chown=openwrt:openwrt configs/base_${BASE_CONFIG}_config /home/openwrt/base_config
```
MuteSpirit, line 65: Maybe use `$HOME` instead of explicit `/home/openwrt` to avoid possible issue if in future name of user will be changed

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -46,68 +62,40 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
...
-        && scripts/feeds install intel_mips ; \
+        && scripts/feeds install intel_mips \
+        && scripts/feeds install -a -p intel ; \
     fi ;\
-    cp /home/openwrt/sdk_config .config \
-    && echo "CONFIG_SDK=y" >> .config \
-    && echo "CONFIG_TARGET_${TARGET}=y" >> .config \
+    scripts/feeds update -a \
+    && scripts/feeds install -a \
+    && cp /home/openwrt/base_config .config \
```
MuteSpirit, line 79: Maybe use `$HOME` instead of explicit `/home/openwrt`?

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -46,68 +62,40 @@ RUN git clone "$OPENWRT_REPOSITORY" openwrt \
...
+    && echo "CONFIG_TARGET_${TARGET_SYSTEM}=y" >> .config \
+    && echo "CONFIG_TARGET_${TARGET_SYSTEM}_${SUBTARGET}=y" >> .config \
+    && echo "CONFIG_TARGET_${TARGET_SYSTEM}_${SUBTARGET}_${TARGET_PROFILE}=y" >> .config \
+    && echo "CONFIG_PACKAGE_prplmesh${PRPLMESH_VARIANT}=y" >> .config \
```
MuteSpirit: How do you think will it be more flexible to move this logic into some build script and break the dependency on usage only this docker and not another build environment?

tools/docker/builder/openwrt/Dockerfile
rmelotte: HOME is not set as an argument to the Dockerfile, so not accessible here.
We could add it, but since the actual username or path used inside the container don't really matter, I'm not sure it's worth it.

tools/docker/builder/openwrt/Dockerfile
rmelotte: HOME is not set as an argument to the Dockerfile, so not accessible here.
We could add it, but since the actual username or path used inside the container don't really matter, I'm not sure it's worth it.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Are you refering to this PR? https://github.com/prplfoundation/prplMesh/pull/917
Yes I was going to suggest not to add another Dockerfile.
If the goal is to be able to compile a development version of prplWrt (by that I mean with changes not yet part of the repository), we could mount a directory instead of cloning the repository.
It's not strictly necessary to move this logic to a build script to do that, but it may be more convenient to debug build failures.

tools/docker/builder/openwrt/Dockerfile
MuteSpirit: Shell commands executing via `RUN` may use not only env vars defined via `ENV` or `ARG`, some basic env vars like `HOME` are also present.
You may try to run next to see available env vars:
```
FROM ubuntu:18.04
RUN useradd -ms /bin/bash openwrt
USER openwrt
WORKDIR $HOME
RUN env
```

rmelotte: It's true for shell commands executed via RUN (so true for my other comment), but neither for COPY nor for WORKDIR (or it has changed in recent docker versions - does what you pasted work for you? With which docker version?).

So yes we can use it for the `cp` command below, but we still have to hard-code it here and when using WORKDIR, so there is not much value in switching to $HOME in some places if it doesn't work in others.

What we could do if we really wanted to use another username, is to create a home directory that doesn't have the user name in it's path, but again I can't come up with a use-case where you would need to replicate the user name and home path inside the container.

MuteSpirit: I withdraw my proposal, i didn't know that shell env vars doesn't work for COPY, WORKDIR (and maybe to some another Dockerfile specific commands too). So if it's impossible to avoid code duplication then my suggestion doesn 't make sense.

tools/docker/builder/openwrt/Dockerfile
MuteSpirit: Yes, #917 is good example.
Could you please say what is purpose to build separate docker image for each target/subtarget/... for each commit?

How do you think will be better to have a single docker image with build environment (gcc, make, etc.) required to build OpenWrt and keep workspaces with prplWrt/OpenWrt sources on host?

tests/openwrt/test_status.sh
```diff
@@ -34,6 +34,8 @@ EOF
 # The first start currently never succeeds, we need to restart it first.
 echo "Attempting to start/restart prplMesh ..."
 ssh "$TARGET" <<"EOF"
+# Some devices still use the old path (outside of "scripts")
+ln -s /opt/prplmesh/scripts/prplmesh_utils.sh /opt/prplmesh/prplmesh_utils.sh || true
```
arnout, line 38: You'd normally use `ln -sf` to overwrite an already existing one (nitpick).

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -33,10 +33,22 @@ ARG INTEL_FEED
 ARG OPENWRT_REPOSITORY
 # Which OpenWrt version (commit hash) to use:
 ARG OPENWRT_VERSION
-# Target to build for (can be any existing value after CONFIG_TARGET_):
-ARG TARGET
-# SDK .config to use as baseline
-ARG SDK_CONFIG=default
+
+# Target to build for (CONFIG_TARGET will be prepended).
```
arnout:
```suggestion
# Target to build for (CONFIG_TARGET_ will be prepended).
```

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -33,10 +33,22 @@ ARG INTEL_FEED
...
+# Example: TARGET_SYSTEM=mvebu
+ARG TARGET_SYSTEM
+
+# Subtarget (CONFIG_TARGET will be prepended).
```
arnout:
```suggestion
# Subtarget (CONFIG_TARGET_${TARGET_SYSTEM}_ will be prepended).
```

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -33,10 +33,22 @@ ARG INTEL_FEED
...
+# Example: SUBTARGET=cortexa9
+ARG SUBTARGET
+
+# Target profile (CONFIG_TARGET will be prepended).
+# Example: TARGET_PROFILE=DEVICE_cznic_turris-omnia
```
arnout, line 50: It might be useful to add the DEVICE part automatically (nitpick).

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -79,39 +91,11 @@ RUN if [ -n "$INTEL_FEED" ] ; then \
...
-    && cat .config \
-    && make package/feeds/prpl/prplmesh/compile -j$(nproc) V=sc \
-    && make package/feeds/prpl/prplmesh/compile -j$(nproc) \
-    && make package/feeds/prpl/prplmesh/clean
-    # For some reason, make has to be run twice because
-    # it still rebuilds some packages on the second run.
-    # On the third run, it's able to compile prplmesh only.
-    # V=sc turns on verbosity for the first command to
-    # make debugging easier.
+RUN make package/feeds/prpl/prplmesh/clean
```
arnout: If that is the only thing that we do here, it doesn't make a lot of sense to make a separate image for it. IOW, this can be merged into openwrt-prebuilt (which should then be renamed to prplmesh-builder, of course).

(nitpick)

arnout (approved):
Nitpicks, nitpicks... Let's get this merged quickly. Feel free to remove the don't merge and merge as is if you have something better to do.

tools/docker/builder/openwrt/Dockerfile
rmelotte: Each docker image contains a pre-built OpenWrt tree for a specific target device that we use to produce prplMesh ipks.
The ipks are then deployed to the different devices that we have before the CI tests are run.

Note that the docker images are only built once on each CI runner we have, then the build result is cached and re-used for later runs, so we don't rebuild the full docker image every time.
The images are only rebuilt from a previous step if something changes, like the OpenWrt version hash, the base configuration, and so on.

The single docker image you refer to already exist, it's just not kept in CI.
If you look at the `openwrt-prerequisites` stage it's exactly that: it contains all you need to build with OpenWrt (you can build it using the `--target` option of `docker build`).

If you need to work on OpenWrt and want to compile a development version locally, you could adapt the existing scripts and the Dockerfile to use a local OpenWrt tree instead of a version hash.
In that case, yes, you might want to move the lines we are commenting on here to a separate script for convenience, in order to build prplWrt when running the container and not when building it.

However: if you only want to build prplMesh and/or a full image for an OpenWrt target based on a public version of OpenWrt, there is no need to do all that.
Just running `tools/docker/builder/openwrt/build.sh -d <target>` will produce both the full image (when this is merged into master) and the prplMesh ipk.
It will take some time to complete on the first run because it has to build everything from scratch, then it will be much quicker on the next runs because it will only rebuild prplMesh.

tests/openwrt/test_status.sh
rmelotte: If the link name exists, it means the target doesn't, so we don't want to overwrite it (the link would be broken).

tools/docker/builder/openwrt/Dockerfile
rmelotte: Indeed, it doesn't make a lot of sense to keep it.
Thanks for the suggestion, done.

rmelotte:
run_tests failed on `test_client_steering_dummy`: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/464609295

Since it's unrelated to this PR, I just re-ran the job again so this can be merged.

tools/docker/builder/openwrt/build.sh
```diff
@@ -46,15 +48,20 @@ build_prplmesh() {
...
-           -v "$scriptdir/scripts:/home/openwrt/openwrt_sdk/build_scripts/:ro" \
+           -v "$scriptdir/scripts:/home/openwrt/openwrt/build_scripts/:ro" \
            -v "${rootdir}:/home/openwrt/prplMesh_source:ro" \
            "$image_tag" \
            ./build_scripts/build.sh
     mkdir -p "$build_dir"
-    docker cp "${container_name}:/home/openwrt/openwrt_sdk/prplmesh-${TARGET}-${OPENWRT_VERSION}-${PRPLMESH_VERSION}.ipk" "$build_dir"
+    # Note: docker cp does not support globbing, so we need to copy the folder
+    docker cp "${container_name}:/home/openwrt/openwrt/artifacts/" "$build_dir"
+    mv "$build_dir/artifacts/"* "$build_dir"
```
MuteSpirit, line 63: Maybe on `docker run` mount `$build_dir/artifacts` into `/home/openwrt/openwrt/artifacts` inside container to avoid operation of copying artifacts from container to host?

tools/docker/builder/openwrt/Dockerfile
MuteSpirit: I understood described use case.

So if presence of big images with built OpenWrt in docker cache is suitable then i'd like to withdraw my offer.

PR #907, opened 2020-03-03
build.sh uses wrong path to mount rootdir

abelog:
### Description

Currently,  `build.sh all` script produces next output:
```
ASSERT: Command '['cmake', '--build', '/local/workspace/prpl/sources/prplMesh/build/nng', '--target', 'install', '--', '-j', '5']' returned non-zero exit status 2
```
As result - it's impossible to build project using provided Dockerfiles and scripts.
This is caused by `rootdir` variable which mounts only `prplMesh` folder (but not `3rdparty`, `hostap` dirs which are required to` build.sh all`) in `prplmesh-builder` Docker image.

### Changes

- Currently, rootdir points to `prplMesh` folder. Substituting one more additional level makes Dockerimage to mount all repo tool projects required to `build.sh all`. 
- In order to keep path to `maptools.py`, make image `--workdir `point to `prplMesh` dir

### DIFF
--- a/tools/docker/build.sh
+++ b/tools/docker/build.sh
@@ -7,10 +7,9 @@
 ###############################################################
 
 scriptdir="$(cd "${0%/*}"; pwd)"
-rootdir="${scriptdir%/*/*}"
+rootdir=${scriptdir%/*/*/*}
 
-. ${rootdir}/tools/docker/functions.sh
-
+. ${scriptdir}/functions.sh
 main() {
     docker image inspect prplmesh-builder >/dev/null 2>&1 || {
         echo "Image prplmesh-build does not exist, creating..."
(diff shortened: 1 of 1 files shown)


tools/docker/build.sh
```diff
@@ -7,10 +7,9 @@
 ###############################################################
 
 scriptdir="$(cd "${0%/*}"; pwd)"
-rootdir="${scriptdir%/*/*}"
-
-. ${rootdir}/tools/docker/functions.sh
```
tomereli, line 12: I'm missing a description for the PR, so can't really review cause I don't know what is fixed

abelog: This is draft yet. Will be ready by EOD

tomereli (changes requested):
This PR is not needed. There is a confusion, building should not use `./build.sh all`, you should only build using `./build.sh map`. Building prplmesh with the hostapd and dwpal we have in github is deprecated and will not work.

ghost:
@abelog Can you compile prplmesh with `./build.sh map`? 
If you thought you needed to run `./build.sh all` because of some (outdated) documentation, please update that documentation (or mention it on `#prplmesh-dev` so someone else can update it)

abelog:
@fvbogaert `build.sh map` works flawlessly.

I am closing PR as not required.
So other newcomers won't be confused too, after conversation with @tomereli on `#prplmesh-dev`  it was decided to:
 - Remove mentions of outdated args from documentation (`dep` and `all`)
 - Make `map` as default arg for `maptools.py`

PR #908, opened 2020-03-04
tests: send_CAPI_command: add tlv class

arnout:
The dev_send_1905 function needs a list of TLVs as arguments. Up to now,
they were given as tuples, which is a bit hard to understand for the
developer.

Create a tlv class and use that as the arguments to dev_send_1905. It
allows us to move the format_tlv local function to the tlv class.

Update test_flows.py accordingly. It's just a matter of calling the tlv
constructor instead of a simple tuple.

This was requested in https://github.com/prplfoundation/prplMesh/pull/859#discussion_r386341137

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -15,6 +15,7 @@ import time
 import json
 
 import send_CAPI_command
+from send_CAPI_command import tlv
 
 '''Regular expression to match a MAC address in a bytes string.'''
 RE_MAC = rb"(?P<mac>([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})"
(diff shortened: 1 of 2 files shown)


arnout:
I've created this as a separate PR so it can be reviewed and merged immediately. However, it will also be part of the second test-flows-to-python-PR, so if I created that one before this one is merged, I'll just close this one.

tests/test_flows.py
```diff
@@ -15,6 +15,7 @@
 import json
 
 import send_CAPI_command
+from send_CAPI_command import tlv
```
tomereli, line 18: Perhaps we should rename send_CAPI_command to UCC or CAPI? (nitpick)

tomereli (approved):

arnout:
Included in #922.

PR #909, opened 2020-03-04
bpl: cfg_get_prplmesh_param_int: add default value

vitalii-komisarenko:
Currently BPL subsystem returns error code if a configuration parameter is missing in the DB.
This error is usually propagated by its clients and it ends up shutting down the whole system.
This effect is undesired for non-critical parameters.

One approach is to keep configuration updated for every supported platform.
Forcing developers to explicitly set all the parameters can potentially lead to better tuned firmware.
The downside is higher maintanance time/cost.

Another approach is to use default values for missing configuration parametes.
Falling back to a default value will keep FW running thus making it more robust.
It also allows to configure smaller fraction of available parameters in order to decrese time to market for new platforms.

This commit provides API for integer parameters for the second approach.
It adds a default parameter to a bunch of cfg_get_* functions that is used if the parameter is missing in the DB.
Error is still returned in case of syntax error (non-integer value).

The affected code in Linux and UCI implementation is reorganized in a similar manner in order to facilitate their code merge in the future.

Signed-off-by: Vitalii Komisarenko <vitalii.komisarenko@essensium.com>

### DIFF
--- a/framework/platform/bpl/include/bpl/bpl_cfg.h
+++ b/framework/platform/bpl/include/bpl/bpl_cfg.h
@@ -84,6 +84,13 @@ namespace bpl {
 #define BPL_GW_DB_MANAGE_MODE_LEN (127 + 1) /* Maximal length of MANAGEMENT MODE string */
 #define BPL_GW_DB_OPER_MODE_LEN (127 + 1)   /* Maximal length of OPERATING MODE string */
 
+/* Default values */
+#define DEFAULT_STOP_ON_FAILURE_ATTEMPTS 1
+#define DEFAULT_RDKB_EXTENSIONS 0
+#define DEFAULT_BAND_STEERING 1
+#define DEFAULT_DFS_REENTRY 1
+#define DEFAULT_CLIENT_ROAMING 1
+
 /****************************************************************************/
 /******************************* Structures *********************************/
 /****************************************************************************/
--- a/framework/platform/bpl/uci/cfg/bpl_cfg.cpp
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg.cpp
@@ -113,9 +113,11 @@ int cfg_get_certification_mode()
 int cfg_get_stop_on_failure_attempts()
 {
     int retVal = -1;
-    if (cfg_get_prplmesh_param_int("stop_on_failure_attempts", &retVal) == RETURN_ERR) {
-        MAPF_ERR("cfg_get_stop_on_failure_attempts: Failed to read stop_on_failure_attempts "
-                 "parameter\n");
+    if (cfg_get_prplmesh_param_int_default("stop_on_failure_attempts", &retVal,
+                                           DEFAULT_STOP_ON_FAILURE_ATTEMPTS) == RETURN_ERR) {
+
+        MAPF_INFO("cfg_get_stop_on_failure_attempts: Failed to read stop_on_failure_attempts "
+                  "parameter\n");
         return RETURN_ERR;
     }
     return retVal;
@@ -134,8 +136,9 @@ int cfg_is_onboarding()
 int cfg_get_rdkb_extensions()
 {
     int retVal = -1;
-    if (cfg_get_prplmesh_param_int("rdkb_extensions", &retVal) == RETURN_ERR) {
-        printf("cfg_get_rdkb_extensions: Failed to read RDKB Extensions parameter\n");
+    if (cfg_get_prplmesh_param_int_default("rdkb_extensions", &retVal, DEFAULT_RDKB_EXTENSIONS) ==
+        RETURN_ERR) {
+        MAPF_INFO("cfg_get_rdkb_extensions: Failed to read RDKB Extensions parameter\n");
         return RETURN_ERR;
     }
     return retVal;
@@ -144,8 +147,9 @@ int cfg_get_rdkb_extensions()
 int cfg_get_band_steering()
 {
     int retVal = -1;
-    if (cfg_get_prplmesh_param_int("band_steering", &retVal) == RETURN_ERR) {
-        MAPF_ERR("cfg_get_band_steering: Failed to read BandSteering parameter\n");
+    if (cfg_get_prplmesh_param_int_default("band_steering", &retVal, DEFAULT_BAND_STEERING) ==
+        RETURN_ERR) {
+        MAPF_INFO("cfg_get_band_steering: Failed to read BandSteering parameter\n");
         return RETURN_ERR;
     }
     return retVal;
@@ -154,8 +158,9 @@ int cfg_get_band_steering()
 int cfg_get_dfs_reentry()
 {
     int retVal = -1;
-    if (cfg_get_prplmesh_param_int("dfs_reentry", &retVal) == RETURN_ERR) {
-        printf("cfg_get_dfs_reentry: Failed to read DfsReentry parameter\n");
+    if (cfg_get_prplmesh_param_int_default("dfs_reentry", &retVal, DEFAULT_DFS_REENTRY) ==
+        RETURN_ERR) {
+        MAPF_INFO("cfg_get_dfs_reentry: Failed to read DfsReentry parameter\n");
         return RETURN_ERR;
     }
     return retVal;
@@ -164,8 +169,10 @@ int cfg_get_dfs_reentry()
 int cfg_get_client_roaming()
 {
     int retVal = -1;
-    if (cfg_get_prplmesh_param_int("client_roaming", &retVal) == RETURN_ERR) {
-        MAPF_ERR("cfg_get_client_roaming: Failed to read ClientRoaming parameter\n");
+    if (cfg_get_prplmesh_param_int_default("client_roaming", &retVal, DEFAULT_CLIENT_ROAMING) ==
+        RETURN_ERR) {
+
+        MAPF_INFO("cfg_get_client_roaming: Failed to read ClientRoaming parameter\n");
         return RETURN_ERR;
     }
     return retVal;
--- a/framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
+++ b/framework/platform/bpl/uci/cfg/bpl_cfg_helper.cpp
@@ -11,6 +11,8 @@
 #include "bpl_cfg_helper.h"
 #include "bpl_cfg_uci.h"
 
+#include "mapf/common/logger.h"
+
 namespace beerocks {
 namespace bpl {
 
@@ -66,20 +68,46 @@ int cfg_get_prplmesh_radio_param(int radio_id, const std::string &radio_param, c
     return cfg_uci_get(path, buf, buf_len);
 }
 
-int cfg_get_prplmesh_param_int(const std::string &param, int *buf)
+static int cfg_get_prplmesh_param_int_raw(const std::string &param, int *buf)
 {
     int status;
     char val[MAX_UCI_BUF_LEN] = "";
 
     status = cfg_get_prplmesh_param(param, val, MAX_UCI_BUF_LEN);
     if (status == RETURN_ERR)
-        return RETURN_ERR;
+        return RETURN_ERR_NOT_FOUND;
 
     status = sscanf_s(val, "%d", buf);
     if (status != 1)
-        return RETURN_ERR;
+        return RETURN_ERR_PARSE;
 
     return RETURN_OK;
+}
+
+int cfg_get_prplmesh_param_int(const std::string &param, int *buf)
+{
+    int status = cfg_get_prplmesh_param_int_raw(param, buf);
+    return status == RETURN_OK ? RETURN_OK : RETURN_ERR;
+}
+
+int cfg_get_prplmesh_param_int_default(const std::string &param, int *buf, int default_value)
+{
+    int status = cfg_get_prplmesh_param_int_raw(param, buf);
+    switch (status) {
+    case RETURN_OK:
+        return RETURN_OK;
+    case RETURN_ERR_NOT_FOUND:
+        *buf = default_value;
+        MAPF_INFO(__FUNCTION__ << ": missing parameter '" << param
+                               << "', using default: " << default_value);
+        return RETURN_OK;
+    case RETURN_ERR_PARSE:
+        MAPF_ERR(__FUNCTION__ << ": can't parse value for '" << param << "'");
+        return RETURN_ERR;
+    default:
+        MAPF_ERR(__FUNCTION__ << ": cfg_get_prplmesh_param_int_raw returned unexpected value");
+        return RETURN_ERR;
+    }
 }
 
 int cfg_get_channel(const std::string &interface_name, int *channel)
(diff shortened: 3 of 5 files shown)


framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -213,9 +213,10 @@ struct BPL_WLAN_IFACE {
  *
  * @return 1 if enabled.
  * @return 0 if disabled.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_is_enabled();
+int cfg_is_enabled(int default_value = -1);
```
tomereli: I don't think default value should be a parameter - for each case we should handle on its own and decide if a default value should be used or if that specific parameter is mandatory.
In the `cfg_is_enabled` case - this should be mandatory.

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -256,62 +257,69 @@ int cfg_get_management_mode();
  *
  * @return BPL_CERTIFICATION_MODE_ON Certification mode is ON.
  * @return BPL_CERTIFICATION_MODE_OFF Certification mode is OFF.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_get_certification_mode();
+int cfg_get_certification_mode(int default_value = -1);
```
tomereli: Again,  no need to change the function prototype. The default value should be false.

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -256,62 +257,69 @@ int cfg_get_management_mode();
...
 
 /**
  * Returns the maximum number of failures allowed on agent before stopping its execution.
  *
  * @return Maximum number of failures allowed or 0 to retry indefinitely.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_get_stop_on_failure_attempts();
+int cfg_get_stop_on_failure_attempts(int default_value = -1);
```
tomereli: default value for stop_on_failure should be 1

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -256,62 +257,69 @@ int cfg_get_management_mode();
...
 /**
  * Returns whether the platform is in onboarding state.
  *
  * @return 1 Platform is in onboarding state.
  * @return 0 Platform is NOT in onboarding state.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_is_onboarding();
+int cfg_is_onboarding(int default_value = -1);
```
tomereli: No default value should be used.

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -256,62 +257,69 @@ int cfg_get_management_mode();
...
 /**
  * Checks the state of the RDKB Extensions feature.
  *
  * @return 1 Enabled.
  * @return 0 Disabled.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_get_rdkb_extensions();
+int cfg_get_rdkb_extensions(int default_value = -1);
```
tomereli: the default value should be 0 (false)

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -256,62 +257,69 @@ int cfg_get_management_mode();
...
 /**
  * Checks the state of the Band Steering feature.
  *
  * @return 1 Enabled.
  * @return 0 Disabled.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_get_band_steering();
+int cfg_get_band_steering(int default_value = -1);
```
tomereli: Default should be false

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -256,62 +257,69 @@ int cfg_get_management_mode();
...
 /**
  * Checks the state of the DFS Reentry feature.
  *
  * @return 1 Enabled.
  * @return 0 Disabled.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_get_dfs_reentry();
+int cfg_get_dfs_reentry(int default_value = -1);
```
tomereli: Default should be false

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -256,62 +257,69 @@ int cfg_get_management_mode();
...
 /**
  * Checks the state of the Client Roaming feature.
  *
  * @return 1 Enabled.
  * @return 0 Disabled.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_get_client_roaming();
+int cfg_get_client_roaming(int default_value = -1);
```
tomereli: Default should be false

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -381,9 +389,10 @@ int cfg_get_beerocks_credentials(const int radio_dir, char ssid[BPL_SSID_LEN],
  * the wpa-supplicant will get from there in runtime, without supply it on `connect` API, or not.
  *
  * @return mem_only_psk flag on success.
+ * @return default_value if value not set
  * @return -1 Error.
  */
-int cfg_get_security_policy();
+int cfg_get_security_policy(int default_value = -1);
```
tomereli: Not sure about this one - @kantera800 ?

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -65,21 +65,33 @@ int cfg_get_param(const std::string &param, std::string &value)
     return RETURN_ERR;
 }
 
-int cfg_get_param_int(const std::string &param, int &value)
+static int cfg_get_prplmesh_param_int(const std::string &param, int default_value)
```
tomereli: This should not change at all - each API on its own should take care of the default value which is used only if reading from the DB fails.

framework/platform/bpl/linux/bpl_cfg.cpp
```diff
@@ -65,21 +65,33 @@ int cfg_get_param(const std::string &param, std::string &value)
...
         MAPF_ERR("Failed reading param " << param);
-        return RETURN_ERR;
+        return default_value;
     }
 
-    value = utils::stoi(str_value);
+    int value = utils::stoi(str_value);
 
-    return RETURN_OK;
+    return value;
```
tomereli: Again, not OK - we do not return value means function succeeded or not.

framework/platform/bpl/linux/prplmesh_platform_db
```diff
@@ -2,3 +2,10 @@ management_mode=Multi-AP-Controller-and-Agent
 operating_mode=Gateway
 stop_on_failure_attempts=1
 certification_mode=1
+enable=1
```
tomereli: Here you should add only the mandatory arguments and mention that any argument which is not specified is using default value

tomereli (changes requested):
Not the right direction IMO. We should not change the APIs for supporting default values.
We should go over each API and decide if we want to add default value or not, and add the default in the API itself.
To reduce code duplication, we can define the default values in the header file for each API:

```
#define DEFAULT_IS_ENABLED false
#define DEFAULT_IS_RDKB_EXTENSIONS false
//etc.
```

framework/platform/bpl/include/bpl/bpl_cfg.h
vitalii-komisarenko: It is 1 in Linux implementation. Should it be put into DB since it is non-default?

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -114,9 +114,9 @@ int cfg_get_stop_on_failure_attempts()
 {
     int retVal = -1;
     if (cfg_get_prplmesh_param_int("stop_on_failure_attempts", &retVal) == RETURN_ERR) {
-        MAPF_ERR("cfg_get_stop_on_failure_attempts: Failed to read stop_on_failure_attempts "
+        MAPF_INFO("cfg_get_stop_on_failure_attempts: Failed to read stop_on_failure_attempts "
                  "parameter\n");
-        return RETURN_ERR;
+        return DEFAULT_STOP_ON_FAILURES_ATTEMPTS;
```
arnout: This is still not the way I would like it.

I would prefer to add a second version of `cfg_get_prplmesh_param_int` like this:

```
int cfg_get_prplmesh_param_int(const std::string &param, int *buf, int default)
{
    int status;
    char val[MAX_UCI_BUF_LEN] = "";

    status = cfg_get_prplmesh_param(param, val, MAX_UCI_BUF_LEN);
    if (status == RETURN_ERR) {
        // Not found, use default
        *buf = default;
        return RETURN_OK;
    }
    status = sscanf_s(val, "%d", buf);
    if (status != 1)
        return RETURN_ERR;

    return RETURN_OK;
}
```

and use that version here.

framework/platform/bpl/include/bpl/bpl_cfg.h
arnout: If we're not sure, leave it as no default for the time being.

arnout (approved):
Although I think it can be done better, what is proposed now is already good enough, so approving.

arnout:
BTW, [the build fails](https://prpl-foundation.gitlab.io/-/prplMesh/-/jobs/459183989/artifacts/build/netgear-rax40/openwrt-build.log):

```
[01m[K/home/openwrt/openwrt_sdk/build_dir/target-mips_24kc+nomips16_musl/prplmesh-dwpal/prplmesh-1.4.0-d78175d4/framework/platform/bpl/uci/cfg/bpl_cfg.cpp:[m[K In function '[01m[Kint beerocks::bpl::cfg_get_stop_on_failure_attempts()[m[K':
[01m[K/home/openwrt/openwrt_sdk/build_dir/target-mips_24kc+nomips16_musl/prplmesh-dwpal/prplmesh-1.4.0-d78175d4/framework/platform/bpl/uci/cfg/bpl_cfg.cpp:119:16:[m[K [01;31m[Kerror: [m[K'[01m[KDEFAULT_STOP_ON_FAILURES_ATTEMPTS[m[K' was not declared in this scope
         return [01;31m[KDEFAULT_STOP_ON_FAILURES_ATTEMPTS[m[K;
                [01;31m[K^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[m[K
[01m[K/home/openwrt/openwrt_sdk/build_dir/target-mips_24kc+nomips16_musl/prplmesh-dwpal/prplmesh-1.4.0-d78175d4/framework/platform/bpl/uci/cfg/bpl_cfg.cpp:119:16:[m[K [01;36m[Knote: [m[Ksuggested alternative: '[01m[KDEFAULT_STOP_ON_FAILURE_ATTEMPTS[m[K'
         return [01;36m[KDEFAULT_STOP_ON_FAILURES_ATTEMPTS[m[K;
                [01;36m[K^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~[m[K
                [32m[KDEFAULT_STOP_ON_FAILURE_ATTEMPTS[m[K
```

framework/platform/bpl/include/bpl/bpl_cfg.h
tomereli: defaults can be different for different implementations, good point.
In Linux, most should be enabled by default - I would keep everything as it is today.

framework/platform/bpl/include/bpl/bpl_cfg.h
```diff
@@ -84,6 +84,13 @@ namespace bpl {
 #define BPL_GW_DB_MANAGE_MODE_LEN (127 + 1) /* Maximal length of MANAGEMENT MODE string */
 #define BPL_GW_DB_OPER_MODE_LEN (127 + 1)   /* Maximal length of OPERATING MODE string */
 
+/* Default values */
+#define DEFAULT_STOP_ON_FAILURE_ATTEMPTS 1
+#define DEFAULT_RDKB_EXTENSIONS 0
+#define DEFAULT_BAND_STEERING 0
```
tomereli: Thinking about it again - lets use the following defaults:
band steering - 1
dfs_reentry - 1
client roaming - 1
rdkb extensions - 0

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -135,8 +135,8 @@ int cfg_get_rdkb_extensions()
 {
     int retVal = -1;
     if (cfg_get_prplmesh_param_int("rdkb_extensions", &retVal) == RETURN_ERR) {
-        printf("cfg_get_rdkb_extensions: Failed to read RDKB Extensions parameter\n");
-        return RETURN_ERR;
+        MAPF_INFO("cfg_get_rdkb_extensions: Failed to read RDKB Extensions parameter\n");
```
tomereli, line 141:
```suggestion
        MAPF_INFO("cfg_get_rdkb_extensions: Failed to read RDKB Extensions parameter, using default " << int(DEFAULT_RDKB_EXTENSIONS));
```

tomereli: Same for all others - it is useful to see in the logs what is the default in case it will change in the future and we are debugging an old prplmesh.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
tomereli: OK this makes sense I guess cause it will reduce code so I'm not against it as I previousely was.
I prefer the following signature:
```
int cfg_get_prplmesh_param_int(const std::string &param, int *value, int default)
```

Then it is used as follows:
```
int cfg_get_rdkb_extensions()
{
    int value;
    if (cfg_get_prplmesh_param_int("rdkb_extensions", &value, DEFAULT_RDKB_EXTENSIONS) == RETURN_OK) {
        return value;
    }
    return RETURN_ERR;
}
```

And we should do the same for other parameter types - it will save a lot of duplicated code.
I guess this was the purpose for this change to begin with, so my appologies @vitalii-komisarenko  for changing my mind again :)

framework/platform/bpl/include/bpl/bpl_cfg.h
vitalii-komisarenko: Since the defaults for linux and uci can be different I feel that the common header is the wrong place to put them.

tomereli: I disagree - we should decide on one default set of values and if some implementation wants to override the default - well, this is why we have the DB for - simply put it in the UCI DB.

vitalii-komisarenko: If using a signle set of default value is an option I'd like to use it.

tomereli (approved):

PR #912, opened 2020-03-04
gitlab-ci: clang-format: properly exit on failure

arnout:
When clang-format fails (that is, when it leads to differences), the job
doesn't fail. That's because `git diff` doesn't exit failure when there
are differences. In fact, we saved the exit code in $ret specifically
for this purpose, but the `exit` command uses $? instead of $ret.

Fix that now.

Failing job on master that doesn't report failure: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/458965322

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -33,7 +33,7 @@ clang-format:
         ret=$?
         echo "Inconsistent formatting, please apply patch from artifacts"
         git diff > correct-formatting.patch
-        exit $?
+        exit $ret
       }
   artifacts:
     paths:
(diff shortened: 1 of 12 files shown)


tomereli (approved):

ghost (approved):

ghost (approved):

arnout:
The job now reports

```
Uploading artifacts...
WARNING: correct-formatting.patch: no matching files 
ERROR: No files to upload 
```

It still passes though, so I don't know if we need to fix this.

arnout:
Thanks to @rmelotte, the artefacts error is fixed: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/459139709

rmelotte (approved):

PR #913, opened 2020-03-04
tools: make "map" default build module

abelog:
### Description
Originally raised in https://github.com/prplfoundation/prplMesh/pull/907

When no arguments specified, build.sh reports error.
As result, newcommer may try to build unsupported 'all' or 'dep' targets.
Root cause is located in 'build.py', where modules is positional argument with no default value.
This leads to a confusion.

3 options to resolve:
1. Add check for no args in build.sh, add "map" in case no args given. Which adds more logic to Docker launcher script.
2. Append "map" inside maptools.py (if no args except build are  given), more complicated due to mismatch of object types.
3. Set map as default value in build.py and allow empty arg case which is most minimal change with no additional logic added.

### Changes
Set map as default value to modules argument of build.py.
Change modules nargs to allow no argument case.

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -45,14 +45,9 @@ subcommand help:
 ./maptools.py <subcommand> --help
 ```
 ### Examples
-build all (default):
+build map (default):
 ```
-./maptools.py build all
-```
-
-build only map modules:
-```
-./maptools.py build map
+./maptools.py build
 ```
 
 build only dep modules(nng, safeclib, dwpal):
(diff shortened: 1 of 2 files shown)


tomereli (approved):
Very good commit message! the change itself is welcome, but make sure to add a commit updating the README.md.

abelog:
@tomereli just added commit to fix Documentation to this PR, as they are united by topic

tomereli (approved):
Very good for a first PR. Just note that for next time, once the PR is ready for review, you should mark it as such so its not kept as draft.

MuteSpirit:
There is misspell in PR description:
> Root cause is located in 'build.**b**y',
correct is
> Root cause is located in 'build.**p**y',

abelog:
Fixed! @MuteSpirit , thank you for paying attention :)

abelog:
Branch was force-pushed to rebase on latest master.

ghost (approved):

tools/README.md
```diff
@@ -45,39 +45,34 @@ subcommand help:
...
 ```
-./maptools.py build all -f MSGLIB=nng
+./maptools.py build map -f MSGLIB=nng
 ```
 
 Clean and rebuild controller only:
 ```
-./maptools.py build controller -c clean make
+./maptools.py build map -c clean make
 ```
```
arnout, line 66: Actually, building controller only is no longer supported so this paragraph should just be removed.

arnout (commented):
Truly *excellent* commit message: giving the different options that were considered and explaining why this one was chosen is exactly what we want.

PR #917, opened 2020-03-04
tools/docker: Dockerfile for prplwrt RAX40 build

VladyslavTupikin:
Currently, we do not have dockerfile for building prplwrt for RAX40.
Without dockerfile developer must find and install all depednencies
in own environment. Eeach developer has own non-standard environment.
Which can cause hard-to-debug hard-to-reproduce build runtime errors.

Create new dockerfile and entrypoint script to prepare clear and unified
build environment for all developers.

### DIFF
--- a/tools/docker/builder/openwrt/toolchains/Dockerfile
+++ b/tools/docker/builder/openwrt/toolchains/Dockerfile
@@ -1,0 +1,20 @@
+FROM ubuntu:16.04
+
+RUN dpkg --add-architecture i386
+
+RUN apt-get -y update && apt-get install --no-install-recommends -y sudo git-core \
+gawk wget unzip make uml-utilities bridge-utils debootstrap gcc-multilib \
+perl automake autoconf fakeroot build-essential crash makedumpfile \
+kernel-wedge git-core libncurses5 libncurses5-dev libelf-dev asciidoc binutils-dev \
+bison imagemagick openssl cmake git-core expect xutils-dev genext2fs m4 qemu \
+libqt4-dev qt4-dev-tools python-setuptools \
+python-pip python3.5 python3.5-dev apt-file gcc gcc-multilib \
+libstdc++6:i386 python-zope.interface python-bjsonrpc \
+python-crypto python-openssl python-twisted curl libcurl3 libcurl4-openssl-dev libjson0-dev \
+fuse libfuse2 libfuse-dev libpcre3 libpcre3-dev zlib1g zlib1g-dev \
+libidn2-0 libxml2-dev unzip subversion nmap python-qt4 cflow lua5.1 libssl1.0.0 libssl-dev
+
+RUN pip install --upgrade pip && pip install pyyaml
+
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+ENTRYPOINT ["/bin/bash","/usr/local/bin/entrypoint.sh"]
--- a/tools/docker/builder/openwrt/toolchains/README.txt
+++ b/tools/docker/builder/openwrt/toolchains/README.txt
@@ -1,0 +1,42 @@
+Building prplwrt for Nighthawk RAX40
+
+# First execution:
+
+1. Download prplmesh
+
+    git clone https://github.com/prplfoundation/prplMesh.git
+
+2. Enter in openwrt/toolchains directory
+
+    cd prplMesh/tools/docker/builder/openwrt/toolchains
+
+3. Build docker image
+
+    docker build -t prplwrt-build .
+
+4. Put prplwrt-env in /usr/bin/
+
+    mv prplwrt-env /usr/bin/
+
+5. Retrieve prplwrt in new empty folder
+
+    git clone https://git.prpl.dev/prplmesh/prplwrt.git|
+
+6. From new folder run prplwrt-env script
+
+    prplwrt-env bash
+
+7. After entering in docker container shell start building
+
+    make world -j4 V=99
+
+# Next executions:
+
+1. Enter in directory with prplwrt folder
+2. Run prplwrt-env script
+
+     prplwrt-env bash
+
+3. After entering in docker container shell start building
+
+     make world -j4 V=99
--- a/tools/docker/builder/openwrt/toolchains/entrypoint.sh
+++ b/tools/docker/builder/openwrt/toolchains/entrypoint.sh
@@ -1,0 +1,53 @@
+#!/bin/sh
+
+cmd="$@"
+
+# Check if the command argument is a service command
+if [ "${cmd}" = "show-launch-script" ] ; then
+    cat /usr/local/bin/openwrt-env
+    exit
+fi
+
+FILE_SYSTEM_OUTPUT=/fs-output
+ROOT=/workspace
+WORK_DIR=${ROOT}
+
+# Create a user and his primary group that match a host's user inside a Docker
+# container.
+# Note: We don't create a user's home directory because it is already created by
+# mounting a host's user .ssh directory in 'docker run' command.
+groupadd -g ${GROUP_ID} ${GROUP_NAME}
+useradd -g ${GROUP_ID} -s "/bin/bash" -u ${USER_ID} ${USER_NAME}
+
+# Prepare user's home directory
+cp -pr /etc/skel/. /home/${USER_NAME}
+
+cat <<EOF >> /home/${USER_NAME}/.profile
+PS1='OPENWRT-ENV:\w\$ '
+export FILE_SYSTEM_OUTPUT=${FILE_SYSTEM_OUTPUT}
+export ROOT=${ROOT}
+cd ${WORK_DIR}
+
+./scripts/feeds setup "src-git,packages,https://git.openwrt.org/feed/packages.git;openwrt-19.07" "src-git,intel_feed,https://git.prpl.dev/prplmesh/feed-intel.git;intel-19.07" "src-git,iwlwav,https://git.prpl.dev/prplmesh/iwlwav.git;iwlwav-19.07" "src-git,routing,https://github.com/openwrt-routing/packages.git" "src-git,luci,https://git.openwrt.org/project/luci.git^c0e73d3f9567f227cbe36ba12af53efbfdd4343d" "src-git,prpl_feed,https://git.prpl.dev/prplwrt/feed-prpl.git"
+./scripts/feeds update
+./scripts/feeds install -a -p intel_feed
+./scripts/feeds install -a -p prpl_feed
+./scripts/feeds install -a -p iwlwav
+./scripts/feeds install -a -p luci
+./scripts/feeds install intel_mips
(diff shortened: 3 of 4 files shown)


tools/docker/builder/openwrt/toolchains/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM ubuntu:16.04
```
tomereli, line 1: Why specifically 16.04?

tools/docker/builder/openwrt/toolchains/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN dpkg --add-architecture i386
```
tomereli, line 3: Can you add a comment for this command? What does it do and what is it needed for?

tools/docker/builder/openwrt/toolchains/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+RUN apt-get -y update && apt-get install --no-install-recommends -y sudo git-core \
```
tomereli, line 5: 1. We prefer to split it to single package per line.
2. No need for sudo - you are root by default in docker
3. Are you sure we need all the tools below?

tools/docker/builder/openwrt/toolchains/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+gawk wget unzip make uml-utilities bridge-utils debootstrap gcc-multilib \
+perl automake autoconf fakeroot build-essential crash makedumpfile \
+kernel-wedge git-core libncurses5 libncurses5-dev libelf-dev asciidoc binutils-dev \
+bison imagemagick openssl cmake git-core expect xutils-dev genext2fs m4 qemu \
+libqt4-dev qt4-dev-tools python-setuptools \
```
tomereli, line 10: I don't see why we need QT - care to explain?

tools/docker/builder/openwrt/toolchains/entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+FILE_SYSTEM_OUTPUT=/fs-output
+ROOT=/workspace
+WORK_DIR=${ROOT}
+
+# Create a user and his primary group that match a host's user inside a Docker
+# container.
+# Note: We don't create a user's home directory because it is already created by
+# mounting a host's user .ssh directory in 'docker run' command.
+groupadd -g ${GROUP_ID} ${GROUP_NAME}
+useradd -g ${GROUP_ID} -s "/bin/bash" -u ${USER_ID} ${USER_NAME}
```
tomereli, line 20: Please refer to `toolchains/rdkb/Dockerfile` - I did this differently there also without creating home directory since I map the current user home directory into the container.

tomereli (commented):
I'm not sure how this should be used - I expect to see some script, which will be used to build the toolchain and call `docker container run` with the $HOME, $USER, $USER_ID, $GROUP_ID and $OPENWRT_ROOT as environment variables like we do in `tools/docker/builder/rdkb/toolchain/bitbake.sh`.
In general I prefer that this will be as close as possible to how the rdkb builder is implemented.

tomereli:
Looking really good so far - only missing an update to the readme, but I think you can go ahead and mark this as ready for review.

rmelotte (changes requested):
I didn't make specific review comments because not sure we want to achieve here.
Is it about being able to compile a development version of prplWrt (meaning with changes not part of any public repo) or only to be able to compile prplWrt locally?

In the former case, we could adapt the existing Dockerfile to:
- bind-mount the prplwrt repository instead of cloning it
- maybe add a possibility to do the build at runtime instead of when building the docker image (makes it easier to debug build failures)

In the latter case, it's already covered by https://github.com/prplfoundation/prplMesh/pull/902

In any case, I don't think we should add another Dockerfile, so marking as changes requested.

tools/docker/builder/openwrt/toolchains/Dockerfile
MuteSpirit: The most usable commands for package or kernel options selection are `make defconfig`, `make menuconfig` and so on which can work in console, but maybe some developers use [xconfig](https://en.wikipedia.org/wiki/Xconfig) for the same purposes. That tool requires Qt

VladyslavTupikin:
@rmelotte 

Goal of this PR is to commit a Dockerfile which may be used by developers to locally build prplWRT for RAX40. Using one common Docker image to build prplWRT images for RAX40 should simplify build procedure for everyone (no need to look for dependencies and install them to local workstation) and also unifies build environment across all developers. Latter one help to avoid some per-station-specific build issues (wrong version of dependencies installed, etc.)

tools/docker/builder/openwrt/toolchains/entrypoint.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/sh
+
+cmd="$@"
+
+# Check if the command argument is a service command
+if [ "${cmd}" = "show-launch-script" ] ; then
+    cat /usr/local/bin/openwrt-env
```
MuteSpirit, line 7: I don't see any mention of `openwrt-env` script in your Dockerfile.
Did you mean `prplwrt-env`?
If yes, then please make it a part of docker image

tools/docker/builder/openwrt/toolchains/README.txt
```diff
@@ -0,0 +1,42 @@
...
+
+    cd prplMesh/tools/docker/builder/openwrt/toolchains
+
+3. Build docker image
+
+    docker build -t prplwrt-build .
+
+4. Put prplwrt-env in /usr/bin/
+
+    mv prplwrt-env /usr/bin/
```
mariomaz, line 19: You need root access to move files to /usr/bin. I cannot do that in my box

tools/docker/builder/openwrt/toolchains/entrypoint.sh
```diff
@@ -0,0 +1,53 @@
...
+
+./scripts/feeds setup "src-git,packages,https://git.openwrt.org/feed/packages.git;openwrt-19.07" "src-git,intel_feed,https://git.prpl.dev/prplmesh/feed-intel.git;intel-19.07" "src-git,iwlwav,https://git.prpl.dev/prplmesh/iwlwav.git;iwlwav-19.07" "src-git,routing,https://github.com/openwrt-routing/packages.git" "src-git,luci,https://git.openwrt.org/project/luci.git^c0e73d3f9567f227cbe36ba12af53efbfdd4343d" "src-git,prpl_feed,https://git.prpl.dev/prplwrt/feed-prpl.git"
+./scripts/feeds update
+./scripts/feeds install -a -p intel_feed
+./scripts/feeds install -a -p prpl_feed
+./scripts/feeds install -a -p iwlwav
+./scripts/feeds install -a -p luci
+./scripts/feeds install intel_mips
+./scripts/feeds install zmq
+./scripts/gen_config.sh rax40 luci
```
mariomaz, line 39: I think this instructions are required to configure the build environment and thus should be executed in the first execution only.

tools/docker/builder/openwrt/toolchains/README.txt
```diff
@@ -0,0 +1,42 @@
...
+# Next executions:
+
+1. Enter in directory with prplwrt folder
+2. Run prplwrt-env script
+
+     prplwrt-env bash
+
+3. After entering in docker container shell start building
+
+     make world -j4 V=99
```
mariomaz, line 42: Depends on the number of cores you have

mariomaz (changes requested):
Other build scripts do not use docker interactive mode. IMO both possibilities should exist.
This docker image could also be adapted to build prplMesh for RAX40 (once you have already built prplWrt)

tools/docker/builder/openwrt/toolchains/Dockerfile
tomereli: IMO the whole interactive option should be removed from the `entrypoint.sh` - developers should prepare everything in the build server natively, and just use a docker container to run `make` - in which case we should limit the tools we install only to what is needed for building prplWrt.
See `tools/docker/builder/rdkb/toolchain/bitbake.sh` for how we do this for building RDKB - I don't think that prplWrt should be any different.

Actually, the only reason we did that for RDKB since it requires Ubuntu 16.04 for some reason and our build server is Ubuntu18.04 LTS which is too new. There is no such requirement for prplWrt for all I know - which raises the question, why do we need this PR in the first place?

Provided we do, I would expect this Dockerfile to be part of prplWrt and not prplMesh, but since for convenience we have this for RDKB then OK, I guess.

If we do need this, then I expect that the following to be done **outside** the container:
```
git clone https://git.prpl.dev/prplmesh/prplwrt.git
cd prplwrt
./scripts/feeds setup "src-git,packages,https://git.openwrt.org/feed/packages.git;openwrt-19.07" "src-git,intel_feed,https://git.prpl.dev/prplmesh/feed-intel.git;intel-19.07" "src-git,iwlwav,https://git.prpl.dev/prplmesh/iwlwav.git;iwlwav-19.07" "src-git,routing,https://github.com/openwrt-routing/packages.git"
./scripts/feeds update 
./scripts/feeds install -a -p intel_feed
./scripts/feeds install -a -p iwlwav
./scripts/feeds install intel_mips
./scripts/feeds install zmq
./scripts/gen_config.sh rax40
```
Not sure about the `gen_config.sh` part though - need to check if it depends on host tools.
Then, like `bitbake.sh`, add a build `tools/docker/builder/prplwrt/make.sh` which will run make from within a container while mapping the prplWrt clone folder, and the home folder if necessary, to the container.

tools/docker/builder/openwrt/toolchains/entrypoint.sh
tomereli: I think these should be done natively in the build server if possible

tools/docker/builder/openwrt/toolchains/README.txt
tomereli: Better to change to `make world -j$(nproc) V=s`

tomereli (changes requested):
I don't see the benefit of the interactive option, can you explain how you use it and why it is better than the approach taken in RDKB builder?

tools/docker/builder/openwrt/toolchains/Dockerfile
VladyslavTupikin: This command adds ability to install packages for i386 architecture.

tools/docker/builder/openwrt/toolchains/Dockerfile
VladyslavTupikin: 1. I will fix it according to commentary
2. I will fix it according to commentary
3. Yes, I am sure, this Dockerfile we use for building OpenWRT.

ghost:
Closing this PR due to replication of existing functionality

PR #918, opened 2020-03-04
tests: 4.2.1 Multi-AP Ethernet Onboarding and Initialization. #343

VladyslavTupikin:
This test verifies that a MAUT can be onboarded via Ethernet
to a Multi-AP network and initialized.

### DIFF
--- a/tests/test_flows.sh
+++ b/tests/test_flows.sh
@@ -13,10 +13,12 @@
 # since it uses all radios. A better solution would of course be to properly configure all APs,
 # but just doing the optimal_path_dummy test first is simpler for the time being.
 # FIXME optimal_path_dummy temporarily disabled since it's broken
-ALL_TESTS="topology ap_config_renew ap_config_bss_tear_down
-           client_capability_query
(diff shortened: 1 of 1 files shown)


tomereli:
I'm not sure this is going in the right direction. We already have the autoconfig and autoconfig renew tests which basically checks onboarding and autoconfiguration.
The only part which is missing to cover all the parts in 4.2.1 is the channel selection, for which we also have a test - so we just need to make sure the channel selection test covers everything defined in the 4.2.1_ETH test.

tomereli (changes requested):
I'm not sure this is going in the right direction. We already have the autoconfig and autoconfig renew tests which basically checks onboarding and autoconfiguration.
The only part which is missing to cover all the parts in 4.2.1 is the channel selection, for which we also have a test - so we just need to make sure the channel selection test covers everything defined in the 4.2.1_ETH test.

VladyslavTupikin:
Pull request closed because function for checking 4.2.1 ETH test flow exist in the python script with name channel_selection.

tomereli:
@VladyslavTupikin please also delete the branch if not needed.

PR #922, opened 2020-03-05
Convert test_flows.sh to python (part 2)

arnout:
More tests converted to python, with additional helpers.

### DIFF
--- a/tests/send_CAPI_command.py
+++ b/tests/send_CAPI_command.py
@@ -3,7 +3,7 @@ import argparse
 import logging
 import socket
 from enum import Enum
-from typing import Dict, Tuple
+from typing import Dict
 
 logger = logging.getLogger(__name__)
 
@@ -14,6 +14,55 @@ class CAPIReply(Enum):
     COMPLETE = "status,COMPLETE"  # parameters may follow
     INVALID = "status,INVALID"
     ERROR = "status,ERROR"
+
+
+class tlv:
+    """Representation of an 1905.1 TLV."""
+
+    def __init__(self, type_: int, length: int, value: str):
+        """A TLV has a type, a length and a value.
+
+        Parameters
+        ----------
+        type_ : int
+            The TLV type.
+
+        length : int
+            The length of the TLV. This must correspond to the length of `value`.
+
+        value : str
+            The TLV value, as a string. It must be formatted according to the UCC rules, i.e. with
+            curly braces and hexadecimal numbers.
+        """
+        self.type = type_
+        self.length = length
+        self.value = value
+
+    def format(self, tlv_num: int = 0) -> str:
+        """Format the TLV for the dev_send_1905 CAPI command.
+
+        Parameters
+        ----------
+        tlv_num : int
+            The TLV counter. In the CAPI command, if there are multiple TLVs, they must be numbered
+            starting from 1. `tlv_num` is that number. If there is only a single TLV, set it to
+            zero.
+
+        Returns
+        -------
+        str
+            The string representation of the TLV according to CAPI definition.
+        """
+        if tlv_num:
+            tlv_num_str = str(tlv_num)
+        else:
+            tlv_num_str = ''
+        return "tlv_type{tlv_num_str},0x{tlv_type:02x}," \
+               "tlv_length{tlv_num_str},0x{tlv_length:04x}," \
+               "tlv_value{tlv_num_str},{tlv_value}".format(tlv_num_str=tlv_num_str,
+                                                           tlv_type=self.type,
+                                                           tlv_length=self.length,
+                                                           tlv_value=self.value)
 
 
 class UCCSocket:
@@ -115,7 +164,7 @@ class UCCSocket:
         reply = self.cmd_reply("dev_get_parameter,program,map,parameter,{}".format(parameter))
         return reply[parameter]
 
-    def dev_send_1905(self, dest: str, message_type: int, *tlvs: Tuple[int, int, str]) -> int:
+    def dev_send_1905(self, dest: str, message_type: int, *tlvs: tlv) -> int:
         """Call dev_send_1905 to `dest` with CMDU type `message_type` and additional `tlvs`.
 
         Parameters
@@ -126,31 +175,20 @@ class UCCSocket:
         message_type : int
             The message type of the 1905.1 message to be sent, as an integer.
 
-        tlvs : Tuple[int, int, str]
-            A list of TLV descriptions. Each TLV is a 3-tuple (type, length, value). The type and
-            length are given as integers. The value is given as a string, formatted according to the
-            UCC rules, i.e. with curly braces and hexadecimal numbers.
+        tlvs : tlv
+            Additional arguments are the TLVs in the 1905.1 message, as `tlv` objects.
 
         Returns
         -------
         The MID of the message, as an integer.
         """
-        def format_tlv(tlv_num, tlv_type, tlv_length, tlv_value):
-            if tlv_num:
-                tlv_num = str(tlv_num)
-            else:
-                tlv_num = ''
-            return "tlv_type{tlv_num},0x{tlv_type:02x}," \
-                   "tlv_length{tlv_num},0x{tlv_length:04x}," \
-                   "tlv_value{tlv_num},{tlv_value}".format(**locals())
-
         cmd = "DEV_SEND_1905,DestALid,{dest:s},MessageTypeValue,0x{message_type:04x}"\
             .format(**locals())
         if len(tlvs) > 1:
-            formatted_tlvs = [format_tlv(tlv_num + 1, *tlv) for (tlv_num, tlv) in enumerate(tlvs)]
+            formatted_tlvs = [tlv.format(tlv_num + 1) for (tlv_num, tlv) in enumerate(tlvs)]
             cmd += ',' + ','.join(formatted_tlvs)
         elif tlvs:
-            cmd += ',' + format_tlv('', *tlvs[0])
+            cmd += ',' + tlvs[0].format()
         return int(self.cmd_reply(cmd)["mid"], base=0)
 
 
--- a/tests/send_CAPI_command.py
+++ b/tests/send_CAPI_command.py
@@ -159,10 +159,28 @@ class UCCSocket:
             self.send_cmd(command)
             return self.get_reply(verbose)
 
-    def dev_get_parameter(self, parameter: str) -> str:
-        """Call dev_get_parameter and return the parameter, or raise KeyError if it is missing."""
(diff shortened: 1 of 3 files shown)


tests/send_CAPI_command.py
```diff
@@ -15,6 +15,54 @@ class CAPIReply(Enum):
...
+
+        value : str
+            The TLV value, as a string. It must be formatted according to the UCC rules, i.e. with
+            curly braces and hexadecimal numbers.
+        """
+        self.type = type_
+        self.length = length
+        self.value = value
+
+    def format(self, tlv_num: int = 0) -> str:
```
ghost, line 41: for ease of debugging, it could be useful to rename this function `__str__` (or `__repr__`), so you can just to `print(tlv)`.

tests/test_flows.py
```diff
@@ -225,6 +295,82 @@ def test_initial_ap_config(self):
...
+            tlv(0x0F, 0x0001, "{0x00}"),
+            tlv(0x10, 0x0001, "{0x00}"))
+
+        time.sleep(3)
+        self.check_log(self.repeater1, "agent_wlan0", r".* tear down radio")
+        conn_map = self.get_conn_map()
+        repeater1 = conn_map[self.mac_repeater1]
+        repeater1_wlan0 = repeater1.radios[self.mac_repeater1_wlan0]
+        for vap in repeater1_wlan0.vaps.values():
+            if vap.ssid != b'N/A':
```
ghost, line 370: Nitpick: `SSID`s should probably be strings, not `bytes`, since they can contain unicode

ghost (approved):
Definitively an improvement :)

Some more nitpicks:
* Consider making the tlv class a named tuple (or a subclass of one). This kind-of conflicts with my earlier suggestion (about changing `tlv.format` to `tlv.__str__`) though, so do whichever one you like
* I don't like `send_CAPI_command.py` as a filename, I'd rather have something like `base_classes.py` (or simply `base.py`). Or maybe call it `capi.py` and put the tlv class in a seperate file

tests/send_CAPI_command.py
arnout: This format function does something different than simply make a string representation, so NACK. I could add a __str__ that prints the three members though.

tests/test_flows.py
arnout: SSIDs can contain any bytes, not limited to UTF-8. So it should most definitely be `bytes`.

arnout:
>     * Consider making the tlv class a named tuple (or a subclass of one).

I don't see how a namedtuple is useful here... IMO it is way more awkward than a class.

I would love to use [attrs](https://github.com/hynek/attrs/) but that's unfortunately not a standard Python module so I wouldn't want to rely on it in the top-level test infra.

>     * I don't like `send_CAPI_command.py` as a filename, I'd rather have something like `base_classes.py` (or simply `base.py`). Or maybe call it `capi.py` and put the tlv class in a seperate file

I fully agree, but since send_CAPI_command.py is called from the existing test_flows.sh, I'd prefer to do that rename when test_flows.sh is removed.

arnout:
I'll continue converting more tests, but it can be merged as is as well so ready for review.

tests/send_CAPI_command.py
```diff
@@ -16,14 +16,63 @@ class CAPIReply(Enum):
...
+            starting from 1. `tlv_num` is that number. If there is only a single TLV, set it to
+            zero.
+
+        Returns
+        -------
+        str
+            The string representation of the TLV according to CAPI definition.
+        """
+        if tlv_num:
+            tlv_num = str(tlv_num)
```
LiorAmram: I think it's bad practice to change the type of a variable, especially when you explicitly mark it as int in the method signature. 
I would prefer defining a tlv_num_str variable.

tests/send_CAPI_command.py
```diff
@@ -126,29 +193,20 @@ def dev_send_1905(self, dest: str, message_type: int, *tlvs: Tuple[int, int, str
...
-                tlv_num = str(tlv_num)
-            else:
-                tlv_num = ''
-            return "tlv_type{tlv_num},0x{tlv_type:02x}," \
-                   "tlv_length{tlv_num},0x{tlv_length:04x}," \
-                   "tlv_value{tlv_num},{tlv_value}".format(**locals())
-
-        cmd = "DEV_SEND_1905,DestALid,{dest:s},MessageTypeValue,0x{message_type:04x}".format(**locals())
+        cmd = "DEV_SEND_1905,DestALid,{dest:s},MessageTypeValue,0x{message_type:04x}"\
+            .format(locals())
```
LiorAmram: Why did you remove the unpacking? Now the format method receives a dict, and this code should fail...

tests/test_flows.py
```diff
@@ -234,6 +234,33 @@ def test_initial_ap_config(self):
         self.check_log(self.repeater1, "agent_wlan0", r".* Controller configuration \(WSC M2 Encrypted Settings\)")
         self.check_log(self.repeater1, "agent_wlan2", r".* Controller configuration \(WSC M2 Encrypted Settings\)")
 
+    def test_ap_config_renew(self):
+        # Regression test: MAC address should be case insensitive
+        mac_repeater1_upper = self.mac_repeater1.upper()
+        # Configure the controller and send renew
+        self.gateway_ucc.cmd_reply("DEV_RESET_DEFAULT")
```
tomereli, line 305: Since autoconfiguration renew should be reused by multiple tests (4.4.3, 4.2.1, 4.8.1, etc.) - we should implement a helper function for ap_config_renew, preferable in a followup commit.
Something like:
```
def trigger_ap_config_renew(self, al_mac, credentials-2.4g, credentials-5g)
```

(nitpick) since this can be done in a separate PR, but I rather keep this comment open till its done.

tests/test_flows.py
```diff
@@ -261,6 +261,37 @@ def test_ap_config_renew(self):
...
+            "DEV_SET_CONFIG,bss_info1,{self.mac_repeater1} 8x".format(self = self))
+        # Send renew message
+        self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x000A,
+            tlv(0x01, 0x0006, "{" + self.mac_gateway + "}"),
+            tlv(0x0F, 0x0001, "{0x00}"),
+            tlv(0x10, 0x0001, "{0x00}"))
+
+        time.sleep(3)
+        self.check_log(self.repeater1, "agent_wlan0", r".* tear down radio")
+        # TODO check that conn map was updated on controller
```
tomereli: Why do we care about the connection map? The TODO should say that we should trigger topology query and check the AP operational BSS TLV which should be empty for all radios.

tests/test_flows.py
```diff
@@ -261,6 +261,37 @@ def test_ap_config_renew(self):
         if not bssid1:
             self.fail("repeater1 didn't configure Multi-AP-24G-1")
 
+    def test_ap_config_bss_tear_down(self):
+        # Configure the controller and send renew
```
tomereli, line 329: So this should already use a helper function.

LiorAmram (dismissed):

tests/test_flows.py
```diff
@@ -133,6 +162,31 @@ def docker_command(self, device: str, *command: str) -> bytes:
         '''Execute `command` in docker container `device` and return its output.'''
         return subprocess.check_output(("docker", "exec", device) + command)
 
+    def beerocks_command(self, command: str) -> bytes:
```
tomereli: `beerocks_cli_command` is more appropriate IMO (nitpick)

tests/test_flows.py
```diff
@@ -277,7 +331,16 @@ def test_ap_config_bss_tear_down(self):
 
         self.check_log(self.repeater1, "agent_wlan0", r"Received credentials for ssid: Multi-AP-24G-3 .* bss_type: 2")
         self.check_log(self.repeater1, "agent_wlan2", r".* tear down radio")
-        # TODO check that conn map was updated on controller
+        conn_map = self.get_conn_map()
```
tomereli, line 344: So now I understand the conn map TODO comment, but still, we should imitate the exact steps which are tested in the testbed, so once #924 is available we should convert this to check the AP Operational BSS TLV instead of relying on the connection map.

tests/test_flows.py
```diff
@@ -245,7 +245,7 @@ def check_log(self, device: str, program: str, regex: str) -> bool:
         '''Verify that on "device" the logfile for "program" matches "regex", fail if not.'''
         logfilename = os.path.join(self.rootdir, 'logs', device, 'beerocks_{}.log'.format(program))
         try:
-            # HACK check_log is often used immediately after triggering a message on the other side.
+            # HACK self.check_log(is, "often", rused immediately after triggering a message on the other side.)
```
tomereli: What's up with this comment?

tests/test_flows.py
```diff
@@ -453,6 +453,46 @@ def test_combined_infra_metrics(self):
         self.check_log(self.repeater1, "agent", "Received TLV_TRANSMITTER_LINK_METRIC")
         self.check_log(self.repeater1, "agent", "Received TLV_RECEIVER_LINK_METRIC")
 
+    def test_client_capability_query(self):
+        sta_mac1 = '00:00:00:11:00:22'
+        sta_mac2 = '00:00:00:11:00:33'
+
+        self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8009,
```
tomereli, line 464: A comment explaining the 1905 command being sent will help understand the flow better.

tomereli (approved):
mainly nitpick comments, but overall I can't really fault this. LGTM

tests/send_CAPI_command.py
arnout: Good point. Done.

tests/send_CAPI_command.py
arnout: That was an accident, fixed.

tests/test_flows.py
arnout: There should be no need to trigger topology query, since the controller should do that automatically upon receiving a topology notification -> #925.

I added the TODO because I removed the corresponding code that existed in test_flows.sh, and then reimplemented it in a different way in the next commit. Since it wasn't a one-to-one mapping I thought it was better to do it in a separate commit, and I didn't want to bother with implementing it the original way first.

tests/test_flows.py
arnout: I disagree that we should do the exact steps of the testbed.

We should *at least* do the same as the testbed, but I think it's definitely OK if we do more.

tests/test_flows.py
arnout: Hahaha, search-replace effect!

PR #923, opened 2020-03-05
test_flows.py: correct bridge interface when using docker

arnout:
test_flows.py calls `docker network inspect` to get the interface on
which to sniff with tcpdump. However, this was only tested with podman,
not upstream docker, and podman has different inspect output.

Distinguish podman and docker with the `plugins` key which docker
doesn't have.

Docker doesn't actually report the interface name in the inspect output,
so construct it based on the Id, like was done in test_flows.sh.

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -99,7 +99,15 @@ class test_flows:
             self.debug("Starting tcpdump, output file {}".format(outputfile))
             inspect = json.loads(subprocess.check_output(('docker', 'network', 'inspect',
                                                           'prplMesh-net-{}'.format(self.opts.unique_id))))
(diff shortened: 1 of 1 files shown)


tomereli (approved):
Where do we actually use `podman`? In CI? Is it aliased to `docker`?

arnout:
> Where do we actually use `podman`?

On my laptop :-) So it did work on my laptop, but not on docker. In CI, the tcpdump option isn't used that's why I didn't notice.

I use podman instead of docker because docker is still using the deprecated cgroupsv1 API which I don't have enabled in my kernel.

rmelotte (approved):

PR #924, opened 2020-03-05
Tests: add a list of ieee 1905 messages captured by tcpdump for easier testing

Galvien:
Add a list of tracked tcpdump messages with ieee1905 data for easier testing of sent and received messages in test_flows.py.

This hasn't been tested in the `test_flows.py` environment so note that this is very much a work in progress!

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -20,6 +20,7 @@ import send_CAPI_command
 RE_MAC = rb"(?P<mac>([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})"
 
 class test_flows:
+    ieee1905_capture = []
     def __init__(self):
         self.tests = [attr[len('test_'):] for attr in dir(self) if attr.startswith('test_')]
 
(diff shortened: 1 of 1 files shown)


tests/test_flows.py
```diff
@@ -112,6 +113,14 @@ def tcpdump_start(self):
             else:
                 self.err("tcpdump terminated")
                 self.tcpdump_proc = None
+            # after tcpdump finishes, add a list of all ieee1905 packets for easier analysis
+            # first, generate a json file of the pcap dump
+            self.json_from_pcap = subprocess.run(['tshark', '-r', outputfile, '-T', 'json' '>tmp.json'])
```
arnout, line 194: Ugh... How about just parsing the pcap file with [pypcap](https://pypi.org/project/pcap-ct/) instead of relying on an external tool?

tests/test_flows.py
```diff
@@ -112,6 +113,14 @@ def tcpdump_start(self):
...
+            if self.json_from_pcap.returncode == 0:
+                with open('tmp.json', 'r') as f:
+                    #filter out only relevant packets from the file
+                   ieee1905_packets =  filter(lambda x: "ieee1905" in x["_source"]["layers"],json.load(f))
```
arnout, line 198: We can also filter at the tcpdump level, that's probably more efficient...

While we're at it, we should also add the `-U` option to tcpdump to make sure the pcap file is always up to date.

tests/test_flows.py
Galvien: I've played around with a bunch of python pcap libraries and neither one allowed for convenient work with ieee1905 messages. on the other hand the objects generated using this method are very readable and convenient.
I don't believe I've checked pypcap, I've tried 2 other libraries, dpkg and scapy.
I'll check it out right now.

tests/test_flows.py
Galvien: I've tried adding a `-j` filter for ieee1905 messages but it was way too strict and resulted in a huge loss of information.

arnout: `-j` filter? I mean a simple pcap-filter expression when doing the tcpdump capture, something like `ether proto 0xwhateveritis`.

Speaking of which: even simpler would be to replace tcpdump with tshark directly, using a combination of `-w` and `-P` to both write to a pcap file and print to stdout. We can then just read from the `self.tcpdump_proc.stdout` and process in the middle of the test.

BTW, this PR is definitely going in the right direction :-)

tomereli: +1 for using tshark directly

Issue #927, opened 2020-03-06
[BUG] run-tests gets the same IP for 2 different devices


rmelotte:
It's already the second time run-tests (test_flows) fails on master in a similar way:

```
 7 / 9 tests failed
```

See the following jobs:
https://gitlab.com/prpl-foundation/prplMesh/-/jobs/461578546
https://gitlab.com/prpl-foundation/prplMesh/-/jobs/458747170

I don't have more information on the possible causes at the moment.

arnout:
Both failing test have the following in their output:
```
test_flows.sh: mac_agent1 = aa:bb:cc:11:00:00
test_flows.sh: mac_agent2 = aa:bb:cc:11:00:00
```

And indeed, I see in the log files that repeater1 doesn't receive any UCC message, while repeater2 receives a lot of duplicates.

So, somehow it seems that repeater1 and repeater2 receive the same IP address...

I think I'll create a quick debugging PR.

rmelotte:
It's frustratingly hard to reproduce "manually" in CI, but happened once again in yesterday's schedules: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/463790642

It also seems that one was caught in `run-tests-py` rather than `run-tests` here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/463991767

arnout:
> It also seems that one was caught in run-tests-py rather than run-tests here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/463991767

And this time, it's the gateway that has the IP address of repeater1...

rmelotte:
It looks like we're still "randomly" generating the IP using this:
```
    rand=`shuf -i 2-254 -n 1`
```
Although I'm not surprised we encounter the issue, I wonder why we suddenly encounter it *so* frequently.

One of the possible solutions was proposed in [461](https://github.com/prplfoundation/prplMesh/issues/461): let docker generate the IP instead of trying to generate it ourself.

PR #929, opened 2020-03-06
Debugging for #927

arnout:
Just to trigger CI so we can debug #927


arnout:
Oops this should have been draft.

arnout:
Meh, I started 6 parallel run-tests jobs and all succeed...

arnout:
Since there's a very good idea now about what the issue is, this is no longer needed.

Issue #445, opened 2019-10-22
[TASK] Using of the validity period in BWL


CoralMalachi:
## Description

The current BWL implementations don't use the validity period option, and in the FINALIZE state of the client steering task, the controller sends an unblock message to all agents.

This task is to make use of the validity period in the BWL implementations.
Then, the unblocking in the FINALIZE state can be removed since it is not really needed due to the validity period.

## Exit criteria

* Using the validity period in BWL 
* The unblocking in the FINALIZE state is removed

PR #930, opened 2020-03-08
Stations remain in blacklist after unblock

kantera800:
I'll start with an intro for client steering flow:

1. all active hostaps moved to blacklist - block messages sent to each hostap's bssid.
2. target hostap allowed - unblock message sent to target bssid
3. client steer request sent to target bssid.
4. client respond "steer success/failure".
5. all hostaps removed from blacklist - unblock messages sent to all hostap's bssid.

The bug is that in step 5 same bssid (current hostap's) is used for all unblock messages.
In result all hostaps rather than the current stay blocked.

Solution is to find the correct bssid for each hostap and use it in the unplock message.
Go over all active hostaps, search for VAPs with same
SSID as the current hostap and send each an unblock message with
the VAP's mac (bssid).


tomereli (changes requested):
The commit message is bad. I honestly don't even understand what this commit aims to solve.

kantera800:
> The commit message is bad. I honestly don't even understand what this commit aims to solve.

I'll start with an intro for client steering flow:
1. all active hostaps moved to blacklist - block messages sent to each hostap's bssid.
2. target hostap allowed - unblock message sent to target bssid
3. client steer request sent to target bssid.
4. client respond "steer success/failure".
5. all hostaps removed from blacklist - unblock messages sent to all hostap's bssid.

The bug is that in step 5 same bssid (current hostap's) is used for **all** unblock messages.
In result all hostaps rather than the current stay blocked.

Solution is to find the correct bssid for each hostap and use it in the unplock message.
Go over all active hostaps, search for VAPs with same
SSID as the current hostap and send each an unblock message with
the VAP's mac (bssid).

If this is more clear I'll add it instead.

arnout (commented):
Commenting only, because I think there's a better approach, but I'm willing to approve (after reviewing in detail) if you disagree.

AFAIU, this issue pops up because commit 5c9d0e627871 updated the client steering task to actually block the other VAPs.

My first question is: why is unblocking needed at all? The blocking is restricted to steering_wait_time_ms (25s) so unblocking should happen automatically. Although it is probably a good idea to unblock after successful steering, it's not really needed.

*If* unblocking is needed, then I think the code should be refactored first. Unblocking is only done by the steering task, so it should be done by the steering task. Also, the blocking is done directly in the steering task so it makes sense to unblocking there as well.

With unblocking moved to steering task, I think there's a lot of refactoring opportunity. From a quick scan of the code, it looks more or less the same as the blocking part.

But actually, I think it makes more sense to store the list of blocked VAPs in the steering task itself, and jut iterate over that list instead of re-requesting the VAPs. Indeed, if a VAP was added in the mean time, it will not have received a block request so there's no need to unblock.

So, my proposal would be to add a list of blocked VAPs to the steering task, and send the unblock request only to them.

Also maybe an additional state should be added for the unblocking - this is relevant if we handle the ACKs. But since we don't do that at the moment, it can stay as is.

kantera800:
> Commenting only, because I think there's a better approach, but I'm willing to approve (after reviewing in detail) if you disagree.
> 
> AFAIU, this issue pops up because commit [5c9d0e6](https://github.com/prplfoundation/prplMesh/commit/5c9d0e627871bd794a7b13d7e03d91e000330fcc) updated the client steering task to actually block the other VAPs.
> 
yes you are correct.

> My first question is: why is unblocking needed at all? The blocking is restricted to steering_wait_time_ms (25s) so unblocking should happen automatically. Although it is probably a good idea to unblock after successful steering, it's not really needed.
> 
I agree that the  steering_wait_time_ms can be less than 25sec.
not sure what you mean by unblocking automatically but we need to make sure that "stupid" STAs don't try to reconnect to the same radio right after it was disassociated with imminent.
Only when STA rejects we send the unblock right away.
Regarding successful steering, not sure  why we also wait 25sec before unblocking. maybe @vitalybu 
could answer that . 
> _If_ unblocking is needed, then I think the code should be refactored first. Unblocking is only done by the steering task, so it should be done by the steering task. Also, the blocking is done directly in the steering task so it makes sense to unblocking there as well.
> 
> With unblocking moved to steering task, I think there's a lot of refactoring opportunity. From a quick scan of the code, it looks more or less the same as the blocking part.
> 
> But actually, I think it makes more sense to store the list of blocked VAPs in the steering task itself, and jut iterate over that list instead of re-requesting the VAPs. Indeed, if a VAP was added in the mean time, it will not have received a block request so there's no need to unblock.
> 
good point.
> So, my proposal would be to add a list of blocked VAPs to the steering task, and send the unblock request only to them.
> 
> Also maybe an additional state should be added for the unblocking - this is relevant if we handle the ACKs. But since we don't do that at the moment, it can stay as is.

Generally i agree with the refractory proposal but not part of this PR.
I propose to merge this fix regardless and open new task for steering task refactoring.

arnout:
> > My first question is: why is unblocking needed at all? The blocking is restricted to steering_wait_time_ms (25s) so unblocking should happen automatically. Although it is probably a good idea to unblock after successful steering, it's not really needed.
> 
> I agree that the steering_wait_time_ms can be less than 25sec.

That was not my point. The wait time anyway only applies when the steering is not succesful.

> not sure what you mean by unblocking automatically

The block message (Client Association Control Request CMDU) has a Validity Period field. The spec has this to say about it:

---
If a Multi-AP Agent receives a Client Association Control Request message with Association Control field set to 0x00 (indicating Client Blocking) and all the following conditions are true, it shall reject a first attempt by a STA specified in the request to associate to the specified BSS operated by the Multi-AP Agent and should not respond to any Probe Request frames sent by the specified STA(s) to the specified BSS:

- The time delta since the message was received is less than the value of Validity Period field in the Client Association Control Request message.
- The Multi-AP Agent has not subsequently received a Client Association Control Request message with Association Control field set to 0x01 (indicating Client Unblocking) specifying the STA.

Otherwise, the Multi-AP Agent shall not perform blocking of the STA.

---

In other words, the blocking should be lifted automatically after 25s. If this doesn't happen, it's a bug in the agent.

> but we need to make sure that "stupid" STAs don't try to reconnect to the same radio right after it was disassociated with imminent.

Yes, that's why we need the blocking. I'm not disputing that.

> Only when STA rejects we send the unblock right away.

Ah, that would be a reason to do an explicit unblock, indeed. However, *do* we really need to unblock right away?

I'm a bit concerned about the load created by this. If we have six agents with two radios each, we'll send a total of 25 messages for one steering attempt...

arnout:
Even if we do want to do the unblocking sooner than after 25s, then I don't think it can ever be an urgent fix, so I *do* want the refactoring to be included in this PR.

If the automatic unblocking doesn't happen, then it's really a bug in the agent so it should be an urgent fix in the agent.

tomereli:
> > The commit message is bad. I honestly don't even understand what this commit aims to solve.
> 
> I'll start with an intro for client steering flow:
> 
> 1. all active hostaps moved to blacklist - block messages sent to each hostap's bssid.
> 2. target hostap allowed - unblock message sent to target bssid
> 3. client steer request sent to target bssid.
> 4. client respond "steer success/failure".
> 5. all hostaps removed from blacklist - unblock messages sent to all hostap's bssid.
> 
> The bug is that in step 5 same bssid (current hostap's) is used for **all** unblock messages.
> In result all hostaps rather than the current stay blocked.
> 
> Solution is to find the correct bssid for each hostap and use it in the unplock message.
> Go over all active hostaps, search for VAPs with same
> SSID as the current hostap and send each an unblock message with
> the VAP's mac (bssid).
> 
> If this is more clear I'll add it instead.

Much better.

tomereli (approved):
We can merge it if its critical, however lets add a task to do this right for this sprint. @kantera800 please link this PR to the new task you add **before** you merge - I'll add the don't merge tag for now.

arnout (approved):
Turns out validity period is indeed ignored in the agent - cfr. #445. So that is the real issue.

However, this change is also going in the right direction - though I'd prefer it to be refactored more, that can be done later.

Therefore, approving.

I've assigned #445 to Alex to follow up on this.

vitalybu:
> I agree that the steering_wait_time_ms can be less than 25sec.
> not sure what you mean by unblocking automatically but we need to make sure that "stupid" STAs don't try to reconnect to the same radio right after it was disassociated with imminent.
> Only when STA rejects we send the unblock right away.
> Regarding successful steering, not sure why we also wait 25sec before unblocking. maybe @vitalybu
> could answer that .

If I remember correctly, in the past we observed stations that decided to disconnect after a successful steer (maybe due to some connectivity issues). The 25 seconds are probably to prevent these station from re-connecting to the "wrong" AP. Not sure it's relevant anymore.

kantera800:
> We can merge it if its critical, however lets add a task to do this right for this sprint. @kantera800 please link this PR to the new task you add **before** you merge - I'll add the don't merge tag for now.

I updated the commit message.
opened task #943 for the steering refactoring.

removing the don't merge flag

Issue #925, opened 2020-03-05
[TASK] Detect topology changes on the network


morantr:
based on #906 summary
## Description

Implement a mechanism which detects on the controller side that an Agents Radio/BSSID went down.

## Problems
In the past, we used to have a TCP/IP socket between the Controller and the Agent.
When a radio went down, the son_slave perform reset and closed the socket to the controller.
When that happened, the socket on the controller side was closed too, and the controller removed the node of the son_slave from the database.

Nowadays, we have a BUS infrastructure that replaces the TCP/IP socket.  This means that if the son_slave perform reset,
No socket is closed and the controller doesn’t aware that and Agent radio is down.

## IEEE 1905.1 Topology Discovery flow brief
The topology discovery protocol uses the multicast discovery procedure, the unicast topology
query/response procedure, and the relayed multicast topology notification procedure to enable a
1905.1 management entity to discover the network topology. These procedures make use of the following
messages:
- Topology discovery message (neighbor multicast)
- IEEE 802.1 bridge discovery message (neighbor multicast)
- Topology query message (unicast)
- Topology response message (unicast)
- Topology notification message (relayed multicast)

Once all interfaces are ready and the Agent establishes a backhaul link then within 1 s, a 1905.1 management entity shall transmit a topology discovery message every 60 seconds.
The 1905.1 management entity shall transmit a topology discovery message on neighbor multicast.

Each 1905.1 management entity (Agent), shall subscribe and listen to the Discovery Topology messages and add all the neighbors it knows to neighbor device TLV on Topology Response message.

If Agent2 disconnects from a different Agent1, it will not get Discovery Topology message for more than 60 seconds and report it on Associated Clients TLV and 1905.1 neighbor device TLV that is in the Topology response message.

### Phase 1 - Update Agent node on Controller database
This phase is required for detecting radio on the Agent that went down.
 - [x] Send Topology Query message on the Controller upon receiving auto-config **M1 or Search** message (?). 
 - [x] Send Topology Query message on the Controller upon receiving Topology Notification Message with zero Client Association Event TLV. 
 - [x] Add missing radio interfaces as dummy information to device information TLV which is sent on Topology Response message on the Agent (backhaul manager).
 - [x] If a son_slave socket of a specific radio interface is not connected, the Agent (backhaul manager) shall not add it to the device information TLV which is sent on Topology Response message it sends.
- [x] When son_slave socket disconnects or connects, the Agent should send a topology notification with no TLVs.
[...]

Issue #524, opened 2019-11-29
[REFACTOR] master database should have different classes for different nodes


arnout:
Currently, the master database only has a `node` class, with a type field to indicate if it is a gateway, repeater, bridge, radio or BSS. However, this doesn't fit very well with how 1905.1/EasyMesh works. In addition, each node contains a lot of redundant data that is not relevant for that particular type. Finally, each node requires a unique MAC address, while in reality it is possible that repeater and bridge and radio and BSS all have the same MAC address.

Instead, it would be nicer to have separate classes for different types, i.e. device (=repeater or gateway), bridge, radio, BSS, with containment relationships between the different classes.

morantr:
I have an idea which I planned in my mind for a long time how to do it.
The most important thing IMO is to get rid of the hierarchy in the database which complexes things up.
As for the class separation, I think that the criteria for separation is if the entity can stand on its own with no dependency, meaning that the only nodes in the database will be "platform" and "station".
as for the fronthaul radio, vaps and backhaul, they will be internally inside the platform node.

I want to take this task.

PR #931, opened 2020-03-08
Update the controller on son_slave connect/disconnect event

morantr:
In the past, we used to have a TCP/IP socket between the Controller and the Agent.
When a radio went down, the son_slave perform reset and closed the socket to the controller.
When that happened, the socket on the controller side was closed too, and the controller removed the node of the son_slave from the database.

Nowadays, we have a BUS infrastructure that replaces the TCP/IP socket. This means that if the son_slave perform reset,
No socket is closed and the controller doesn’t aware that and Agent radio is down.

This PP adds a Topology Query message send after the controller receives M1
and add missing fronthaul radio information to Topology Response message, so the Controller
could remove from its database inactive radio (son_salve) nodes.

#925

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -25,6 +25,7 @@
  * over time this list is going to get very long
  */
 #include <tlvf/ieee_1905_1/eMessageType.h>
+#include <tlvf/ieee_1905_1/s802_11SpecificInformation.h>
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvDeviceInformation.h>
@@ -1989,6 +1990,92 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
     // This is something that cannot be done at this moment because the MediaType field
     // must be computed with information obtained through NL80211_CMD_GET_WIPHY command of DWPAL,
     // which is currently not supported. See "Send standard NL80211 commands using DWPAL #782"
+    // For now, fill from radio data with dummy values based on information we have.
+    for (const auto &soc : slaves_sockets) {
+        // Iterate on front radio iface and then switch to back radio iface
+        auto fill_radio_iface_info = [&](bool front_iface) {
+            LOG(DEBUG) << "filling interface information on radio="
+                       << (front_iface ? soc->radio_mac : soc->radio_mac + " backhaul");
+
+            // Skip Backhaul iteration iface when STA BWL is not allocated (Eth connection or GW).
+            if (!front_iface && !soc->sta_wlan_hal) {
+                LOG(TRACE) << "Skip radio interface with no active STA BWL, front_radio="
+                           << soc->radio_mac;
+                return false;
+            }
+
+            auto localInterfaceInfo = tlvDeviceInformation->create_local_interface_list();
+
+            localInterfaceInfo->mac() =
+                front_iface ? network_utils::mac_from_string(soc->radio_mac)
+                            : network_utils::mac_from_string(soc->sta_wlan_hal->get_radio_mac());
+
+            LOG(DEBUG) << "Added radio interface to tlvDeviceInformation: "
+                       << localInterfaceInfo->mac();
+
+            ieee1905_1::eMediaType media_type = ieee1905_1::eMediaType::UNKNONWN_MEDIA;
+            if (soc->freq_type == beerocks::eFreqType::FREQ_24G) {
+                media_type = ieee1905_1::eMediaType::IEEE_802_11N_2_4_GHZ;
+            } else if (soc->freq_type == beerocks::eFreqType::FREQ_5G) {
+                media_type = ieee1905_1::eMediaType::IEEE_802_11AC_5_GHZ;
+            } else {
+                LOG(ERROR) << "Unsupported freq_type=" << int(soc->freq_type)
+                           << ", iface=" << soc->hostap_iface;
+                return false;
+            }
+            localInterfaceInfo->media_type() = media_type;
+
+            ieee1905_1::s802_11SpecificInformation media_info = {};
+            localInterfaceInfo->alloc_media_info(sizeof(media_info));
+
+            // BSSID field is not defined well for interface. The common definition is in simple
+            // words "the AP/ETH mac that we are connected to".
+            // For fronthaul radio interface or unused backhaul interface put zero mac.
+            if (local_gw || m_sConfig.eType == SBackhaulConfig::EType::Wired || front_iface ||
+                (m_sConfig.eType == SBackhaulConfig::EType::Wireless &&
+                 soc->sta_iface != m_sConfig.wireless_iface)) {
+                media_info.network_membership = network_utils::ZERO_MAC;
+            } else {
+                media_info.network_membership =
+                    network_utils::mac_from_string(soc->sta_wlan_hal->get_bssid());
+            }
+
+            media_info.role =
+                front_iface ? ieee1905_1::eRole::AP : ieee1905_1::eRole::NON_AP_NON_PCP_STA;
+
+            // TODO: The Backhaul manager does not hold the information on the front radios.
+            // For now, put zeros and when the Agent management will be move to unified Agent thread
+            // this field will be filled. #435
+            media_info.ap_channel_bandwidth               = 0;
+            media_info.ap_channel_center_frequency_index1 = 0;
+            media_info.ap_channel_center_frequency_index2 = 0;
+
+            auto *media_info_ptr = localInterfaceInfo->media_info(0);
+            if (media_info_ptr == nullptr) {
+                LOG(ERROR) << "media_info is nullptr";
+                return false;
+            }
+
+            std::copy_n(reinterpret_cast<uint8_t *>(&media_info), sizeof(media_info),
+                        media_info_ptr);
+
+            tlvDeviceInformation->add_local_interface_list(localInterfaceInfo);
+
+            return true;
+        };
+
+        if (!fill_radio_iface_info(true)) {
+            LOG(DEBUG) << "filling interface information on radio=" << soc->radio_mac
+                       << " has failed!";
+            return false;
+        }
+
+        if (!fill_radio_iface_info(false)) {
+            LOG(DEBUG) << "filling interface information on radio=" << soc->radio_mac
+                       << " backhaul has failed!";
+            return false;
+        }
+    }
 
     auto tlvSupportedService = cmdu_tx.addClass<wfa_map::tlvSupportedService>();
     if (!tlvSupportedService) {
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -276,6 +276,18 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
                 finalize_slaves_connect_state(true, slaves_sockets.front());
             }
             LOG(INFO) << "disconnected slave sockets has been deleted";
+
+            if (m_eFSMState >= EState::CONNECT_TO_MASTER) {
+                LOG(INFO) << "Sending topology notification on son_slave disconnect";
+                auto cmdu_header =
+                    cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE);
+                if (!cmdu_header) {
+                    LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
+                    return false;
+                }
+                cmdu_header->flags().relay_indicator = true;
+                send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
+            }
             return false;
         } else {
             ++it;
@@ -1487,6 +1499,18 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
                 LOG(ERROR) << "failed start agent_ucc_listener";
                 return false;
             }
+        }
+
+        if (m_eFSMState >= EState::CONNECT_TO_MASTER) {
+            LOG(INFO) << "Sending topology notification on reconnected son_slave";
+            auto cmdu_header =
+                cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE);
+            if (!cmdu_header) {
+                LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
+                return false;
+            }
+            cmdu_header->flags().relay_indicator = true;
+            return send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
         }
 
         LOG(DEBUG) << "ACTION_BACKHAUL_REGISTER_REQUEST sta_iface=" << soc->sta_iface
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -1510,11 +1510,19 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
     auto mid = cmdu_rx.getMessageId();
     LOG(DEBUG) << "Received TOPOLOGY_NOTIFICATION_MESSAGE, mid=" << std::hex << int(mid);
 
+    // IEEE 1905.1 defines that TOPOLOGY_NOTIFICATION_MESSAGE must containt one 1905.1 AL MAC
+    // address type TLV, and MultiAp standard extend it with zero or one Client Association Event
+    // TLV. So if we didn't receive Client Association Event TLV, we need to send
+    // TOPOLOGY_QUERY_MESSAGE to figure out what has changed on the topology.
     auto client_association_event_tlv = cmdu_rx.getClass<wfa_map::tlvClientAssociationEvent>();
     if (!client_association_event_tlv) {
-        // The standard allows having zero TLV_CLIENT_ASSOCIATION_EVENT, in that case, there is no
-        // point to continue the handling of this message.
-        LOG(INFO) << "wfa_map::tlvClientAssociationEvent not found";
+        LOG(INFO) << "wfa_map::tlvClientAssociationEvent not found, sending TOPOLOGY_QUERY_MESSAGE";
+
+        if (!cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE)) {
+            LOG(ERROR) << "Failed building message!";
+            return false;
+        }
+        son_actions::send_cmdu_to_agent(src_mac, cmdu_tx, database);
         return true;
     }
 
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -33,10 +33,13 @@
 #include <beerocks/tlvf/beerocks_message_1905_vs.h>
 #include <beerocks/tlvf/beerocks_message_control.h>
 
+#include <tlvf/ieee_1905_1/eMediaType.h>
 #include <tlvf/ieee_1905_1/eMessageType.h>
 #include <tlvf/ieee_1905_1/eTlvType.h>
+#include <tlvf/ieee_1905_1/s802_11SpecificInformation.h>
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
+#include <tlvf/ieee_1905_1/tlvDeviceInformation.h>
 #include <tlvf/ieee_1905_1/tlvEndOfMessage.h>
 #include <tlvf/ieee_1905_1/tlvMacAddress.h>
 #include <tlvf/ieee_1905_1/tlvReceiverLinkMetric.h>
@@ -100,6 +103,7 @@ bool master_thread::init()
             ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE,
             ieee1905_1::eMessageType::STEERING_COMPLETED_MESSAGE,
             ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE,
+            ieee1905_1::eMessageType::TOPOLOGY_RESPONSE_MESSAGE,
             ieee1905_1::eMessageType::LINK_METRIC_RESPONSE_MESSAGE,
             ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE,
             ieee1905_1::eMessageType::AP_CAPABILITY_REPORT_MESSAGE,
@@ -297,6 +301,8 @@ bool master_thread::handle_cmdu_1905_1_message(const std::string &src_mac,
         return handle_cmdu_1905_higher_layer_data_message(src_mac, cmdu_rx);
     case ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE:
         return handle_cmdu_1905_topology_notification(src_mac, cmdu_rx);
+    case ieee1905_1::eMessageType::TOPOLOGY_RESPONSE_MESSAGE:
+        return handle_cmdu_1905_topology_response(src_mac, cmdu_rx);
     case ieee1905_1::eMessageType::LINK_METRIC_RESPONSE_MESSAGE:
         return handle_cmdu_1905_link_metric_response(src_mac, cmdu_rx);
     case ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE:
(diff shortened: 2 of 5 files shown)


controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -818,6 +818,13 @@ bool master_thread::handle_cmdu_1905_autoconfiguration_WSC(const std::string &sr
     }
 
     if (!database.setting_certification_mode()) {
+        // trigger Topology query
```
tomereli, line 827: Commit message typo - "upn" --> "upon"

Also, the commit message doesn't explain at all why we send the topology query at this time.
Finally - why do we send it only in certification mode? This will mean it will not work in RDKB and UGW since by default certification mode is not set in these platforms.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,50 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
     // This is something that cannot be done at this moment because the MediaType field
     // must be computed with information obtained through NL80211_CMD_GET_WIPHY command of DWPAL,
     // which is currently not supported. See "Send standard NL80211 commands using DWPAL #782"
+    // For now, fill from radio data with dummy values based on information we have.
```
tomereli, line 2017: It looks like #900 is about to be merged, so I guess it makes more sense to rebase on that instead of adding dummy code. The TODO should mention #900 and not the closed issue in case you insist on not rebasing.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,50 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+    for (const auto soc : slaves_sockets) {
```
tomereli:
```suggestion
    for (const auto &soc : slaves_sockets) {
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,50 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+        std::shared_ptr<ieee1905_1::cLocalInterfaceInfo> localInterfaceInfo =
```
tomereli:
```suggestion
        auto localInterfaceInfo =
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -276,6 +276,16 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
                 finalize_slaves_connect_state(true, slaves_sockets.front());
             }
             LOG(INFO) << "disconnected slave sockets has been deleted";
+
```
tomereli, line 279: Commit message wrong grammar - "he will be notified" -> "it will be notified"

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2011,47 +2011,61 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
     // which is currently not supported. See "Send standard NL80211 commands using DWPAL #782"
```
tomereli, line 2016: Commit 9fe1245e3381926fb5846b87f4ef0f1951c2d59b should be squashed to commit dedefbf. Also, do not put the commit numbers in the commit message unless you're sure you won't change it again, cause if you force push this commit will be deleted.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1611,6 +1616,83 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
+        LOG(ERROR) << "ieee1905_1::tlvDeviceInformation not found";
+        return false;
+    }
+
+    const auto &al_mac = tlvDeviceInformation->mac();
+
+    // Get all Agent known fronthaul radios if exist, and compare it to what the Agent reports.
+    // If a radio exist on the database but not the tlvDeviceInformation, this radio node needs
+    // to be removed.
+    // This shall work in the opposite way as well: If it is not exist on the database, but reported
```
tomereli:
```suggestion
    // This shall work in the opposite way as well: If it does not exist on the database, but reported
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1611,6 +1616,83 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
+    // in the TLV, then needs to add the node to the database. Of course this will remain as
```
tomereli:
```suggestion
    // in the TLV, then we need to add the node to the database. Of course, this will remain as
```

tomereli (commented):
Mainly nitpick comments, but not approving yet cause I think this should be rebased on #900 instead of using dummy data.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Basically it is a pointer (shared), but you are right, a shared pointer size is 16 and is greater than 4.
fixed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: done

controller/src/beerocks/master/son_master_thread.cpp
morantr: done

controller/src/beerocks/master/son_master_thread.cpp
morantr: done

controller/src/beerocks/master/son_master_thread.cpp
morantr: done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: squashed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: From what I can see, #900 does not contain relevant code for getting the info we need here.

mariomaz: #900 implements NL80211_CMD_GET_STATION only but NL80211_CMD_GET_WIPHY is required here. Now that the design for the nl80211 standard commands support has been validated, adding the NL80211_CMD_GET_WIPHY command is the next task I'll be working on.

tomereli: #900 provides the infrastructure for sending NL standard commands - if this does not provide you what you need to get the information you need, then what is? Is there an open issue for that?

tomereli: Got it - thanks @mariomaz , is there an open issue for the NL80211_CMD_GET_WIPHY support? If not, can you create one?
@morantr please add a comment and a task to replace the dummy values once @mariomaz adds the support.

tomereli (approved):
Since support for WIPHY commands via NL80211 is not ready yet - approved, but please add a task and a TODO comment in the code and commit message which mention that.

framework/tlvf/yaml/tlvf/ieee_1905_1/s802_11SpecificInformation.yaml
```diff
@@ -15,7 +15,7 @@ s802_11SpecificInformation:
   _type: struct
   network_membership: sMacAddr
   role: eRole
-  ap_channel_band:
+  ap_channel_bandwidth:
```
arnout, line 18: Yes, I already got confused by that one twice!

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,81 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
     // This is something that cannot be done at this moment because the MediaType field
     // must be computed with information obtained through NL80211_CMD_GET_WIPHY command of DWPAL,
     // which is currently not supported. See "Send standard NL80211 commands using DWPAL #782"
+    // For now, fill from radio data with dummy values based on information we have.
+    for (const auto &soc : slaves_sockets) {
+        // Iterate on front radio iface and then switch to back radio iface
+        for (uint8_t i = 0; i < 2; i++) {
+            bool front_iface = bool(i);
```
arnout: (nitpick) I find this super ugly. I think it's much better to factor the common part in a function or lambda and call it twice.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,81 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+                return false;
+            }
+            localInterfaceInfo->media_type() = media_type;
+
+            ieee1905_1::s802_11SpecificInformation media_info = {};
+            localInterfaceInfo->alloc_media_info(sizeof(media_info));
+
+            // BSSID field is not defined well for interface. The common definition is in simple
+            // words "the AP/ETH mac that we are connected to".
+            // For fronthaul radio interface or unused backhaul interface put zero mac.
```
arnout, line 2057: Given that we consider only the radio as an interface, not each VAP, the logical thing for me would be:
- for backhaul STA: the BSSID the STA is associated to (or 0 if not associated)
- for AP: the BSSID of the first VAP, if it is configured. So generally equal to the radio MAC.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,81 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+                (m_sConfig.eType == SBackhaulConfig::EType::Wireless &&
+                 soc->sta_iface != m_sConfig.wireless_iface)) {
+                media_info.network_membership = network_utils::ZERO_MAC;
+            } else {
+                media_info.network_membership =
+                    network_utils::mac_from_string(soc->sta_wlan_hal->get_bssid());
+            }
+
+            // Note: Not sure that "NON_AP_NON_PCP_STA" is the correct definition to our backhaul
+            // interface.
```
arnout: I am sure :-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,81 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+            media_info.role =
+                front_iface ? ieee1905_1::eRole::AP : ieee1905_1::eRole::NON_AP_NON_PCP_STA;
+
+            // TODO: The Backhaul manager does not hold the information on the front radios.
+            // For now, put zeros and when the Agent management will be move to unified Agent thread
+            // this field will be filled. #435
```
arnout, line 2072: This is not true. The unified agent is about unifying the ap manager and monitor threads. There would still be a separate process for each radio. The idea is that the per-radio process needs to link to bwl and is more prone to crashing, plus it may need more capabilities or different SELinux configuration.

The topology response will always be constructed by the main agent process, not by the per-radio process, so #435 will not solve this.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1989,6 +1990,81 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+            media_info.ap_channel_center_frequency_index2 = 0;
+
+            auto *media_info_ptr = localInterfaceInfo->media_info(0);
+            if (media_info_ptr == nullptr) {
+                LOG(ERROR) << "media_info is nullptr";
+                return false;
+            }
+
+            std::copy_n(reinterpret_cast<uint8_t *>(&media_info), sizeof(media_info),
+                        media_info_ptr);
```
arnout, line 2084: This is so annoying... There should be a tlvf-generated function that does exactly this (similar to the std::string functions when type is char).

(not for this PR)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -276,6 +276,16 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
                 finalize_slaves_connect_state(true, slaves_sockets.front());
             }
             LOG(INFO) << "disconnected slave sockets has been deleted";
+
+            if (m_eFSMState >= EState::CONNECT_TO_MASTER) {
+                LOG(INFO)
+                    << "Sending topology notification to the controller on son_slave disconnect";
```
arnout: Topology notification is multicast, so no "to the controller".

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -276,6 +276,16 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
...
+                if (!cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE)) {
+                    LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
+                    return false;
+                }
+                send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
```
```suggestion
                send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -276,6 +276,16 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
                 finalize_slaves_connect_state(true, slaves_sockets.front());
             }
             LOG(INFO) << "disconnected slave sockets has been deleted";
+
+            if (m_eFSMState >= EState::CONNECT_TO_MASTER) {
```
arnout, line 280: Since it's not to controller (see below) it should be any time that the slaves are connected.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1489,6 +1499,15 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
...
         }
 
+        if (m_eFSMState >= EState::CONNECT_TO_MASTER) {
+            LOG(INFO) << "Sending topology notification to the controller on reconnected son_slave";
+            if (!cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE)) {
+                LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
+                return false;
+            }
+            return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+        }
```
arnout: Same remarks as above. Maybe move the entire thing into a function (including the state condition).

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1611,6 +1616,88 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
+
+    for (uint8_t i = 0; i < tlvDeviceInformation->local_interface_list_length(); i++) {
+        const auto iface_info_tuple = tlvDeviceInformation->local_interface_list(i);
+        if (!std::get<0>(iface_info_tuple)) {
+            LOG(ERROR) << "Failed to get " << int(i) << " element of local iface info "
+                       << "on Device Information TLV";
+            return false;
+        }
+
+        auto &iface_info = std::get<1>(iface_info_tuple);
```
arnout, line 1662:
```suggestion
        const auto &iface_info = std::get<1>(iface_info_tuple);
```

Please please pretty please: do not use strings for MAC addresses.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1611,6 +1616,88 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
+    // If a radio exist on the database but not the tlvDeviceInformation, this radio node needs
+    // to be removed.
+    // This shall work in the opposite way as well: If it doesn't exist on the database, but
+    // reported in the TLV, then we need to add the node to the database. Of course this will remain
+    // as "TODO" for future task. Meanwhile parse properly the TLV, leaving the unused parts in
+    // comment for future implementation.
+    auto fronthaul_radios_on_db = database.get_node_children(
+        network_utils::mac_to_string(al_mac), beerocks::TYPE_SLAVE, beerocks::STATE_CONNECTED);
+
+    std::unordered_set<std::string> reported_fronthaul_radios;
```
arnout, line 1652:
```suggestion
    std::unordered_set<sMacAddress> reported_fronthaul_radios;
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1611,6 +1616,88 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
+        if (iface_info.media_info_length() == 10) {
+            const auto media_info = reinterpret_cast<ieee1905_1::s802_11SpecificInformation *>(
+                iface_info.media_info(0));
+            const auto iface_role = media_info->role;
+
+            // For future implementation
+            // const auto &iface_bssid = media_info->network_membership;
+            // const auto iface_bw     = media_info->ap_channel_bandwidth;
+            // const auto iface_cf1    = media_info->ap_channel_center_frequency_index1;
+            // const auto iface_cf2    = media_info->ap_channel_center_frequency_index2;
```
arnout, line 1684:
```suggestion
            // In MultiAP context, bssid, bw, cf1 and cf2 are not actually useful since they are more accurately
            // represented in other TLVs (operating channel report, AP operational BSS). 
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1611,6 +1616,88 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
+        const auto &iface_mac = iface_info.mac();
+        auto iface_mac_str    = network_utils::mac_to_string(iface_mac);
+
+        // For future implementation
+        // const auto media_type        = iface_info.media_type();
+
+        // For wireless interface it is defined on IEEE 1905.1 that the size of the media info
+        // is n=10 octets, which the size of s802_11SpecificInformation struct.
+        // For wired interface n=0.
+        if (iface_info.media_info_length() == 10) {
```
arnout: I think you should also check the media type, not just the length. It's possible that with some extension, a different media type with a different media info format happens to have the same length.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1611,6 +1616,88 @@ bool master_thread::handle_cmdu_1905_topology_notification(const std::string &sr
...
+    const auto &al_mac = tlvDeviceInformation->mac();
+
+    // Get all Agent known fronthaul radios if exist, and compare it to what the Agent reports.
+    // If a radio exist on the database but not the tlvDeviceInformation, this radio node needs
+    // to be removed.
+    // This shall work in the opposite way as well: If it doesn't exist on the database, but
+    // reported in the TLV, then we need to add the node to the database. Of course this will remain
+    // as "TODO" for future task. Meanwhile parse properly the TLV, leaving the unused parts in
+    // comment for future implementation.
+    auto fronthaul_radios_on_db = database.get_node_children(
```
arnout, line 1649: That's not just fronthaul radios; I think it's really AP radios that you mean (VAPs on that radio can be fronthaul and/or backhaul VAPs; the point it's that it's not backhaul STA).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Created "[TASK] Add NL80211_CMD_GET_WIPHY command #967"

arnout (approved):
Looks great! I have a few important comments, but everything can be fixed easily, so approve with don't merge.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Unify Agent is the entity that will connect to two fronthaul managers (per radio) that each one will have a monitor and ap manager. Meaning, it will be the main Agent process.
The backhaul manager will be responsible only to link establishment and remove from him all the Agent stuff.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: ` ¯\_(ツ)_/¯`

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: done.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: For the backhaul STA, I agree and that what I did.
Regarding the fronthaul interface, I'm not sure this is right. Also, the MAC is already mentioned, so it seems weird to me to write it again on the TLV.
I will keep it as is for now.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Ok, I'll remove the comment.
The standard defines the following options:

> 
“0100” – non-AP/non-PCP STA
“1000” – Wi-Fi P2P Client (see [B04])
“1001” – Wi-Fi P2P Group Owner (see [B04])
“1010” – 802.11adPCP

Do you know what are the others?
@arnout

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Changed to relayed multicast.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Changed to relayed multicast, but I prefer to not turning it to a function since it is not a lot of code and I would like to see the print with the right location.

controller/src/beerocks/master/son_master_thread.cpp
morantr: It is impossible to add the const.

controller/src/beerocks/master/son_master_thread.cpp
morantr: I did it because of the database return `set` of `std::string`.
I don't want to change the database in this PR scope.
Let's give #524 more priority.

controller/src/beerocks/master/son_master_thread.cpp
morantr: AFAIR, the current channel selection task does use it. I'm not sure that we will not use it. Keep it for now.

controller/src/beerocks/master/son_master_thread.cpp
morantr: According to the standard, only wireless interface has a length of 10. But Ok, I added also an enum check.

controller/src/beerocks/master/son_master_thread.cpp
morantr: No, this tlv refers only to the radio interface and not vaps.
The vaps are handled in AP operational BSS TLV.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: removed "to the controller".

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: > Created "[TASK] Add NL80211_CMD_GET_WIPHY command #967"

Add "Enabler" and "Certification" labels

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: @LiorAmram Please open a task for this and work on this, you can take a break from the UTAF tasks. Keeping comment open till @LiorAmram sees this.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Still, the decision is to use `sMacAddress` instead of `std::string` even if we don't change everything at once, so please change to `sMacAddress` as part of this PR.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: Our mindset is to delete what we don't use **now** - so I don't see why we should keep it. In addition, the channel selection task will be stripped down to nothing very soon so we don't really care if it uses it since its obsolete.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: P2P is 11s mesh (AFAIU)

11ad PCP is the mesh-like thing introduced for 60GHz.

Neither of them are covered by MultiAP (AFAIU).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: #973

controller/src/beerocks/master/son_master_thread.cpp
arnout: Indeed. It's OK for `fronthaul_radios_on_db` because that is indeed returned by the database. But `reported_fronthaul_radios` gets filled with the MAC addresses that come from the device information so they're not strings. And where you do `reported_fronthaul_radios.find(fronthaul_radio_on_db)` you can easily do `reported_fronthaul_radios.find(network_utils::string_to_mac(fronthaul_radio_on_db))` instead.

Anyway, I'm not going to block the merge for this. But please take it into account next time.

controller/src/beerocks/master/son_master_thread.cpp
arnout: Channel selection task will use this information - but coming from the AP Radio Basic Capabilities and the Channel Preference Report, which is a lot more accurate than the limited information given in this TLV.

I hoped that my suggested edit expressed this feeling accurately...

controller/src/beerocks/master/son_master_thread.cpp
arnout: Yes, that's what I mean: there is no such thing as a "fronthaul radio". There are fronthaul and/or backhaul VAPs on a radio, but every radio can be used as fronthaul or backhaul AP.

The only limitation you can have on a radio is that it is not going to be used as an AP.

Since this term "fronthaul radio" is introduced by this PR, I think it's rather important to *not* introduce it if it is not correct (or potentially confusing).

arnout: Actually, the term "fronthaul radio" *is* used all over the place in the test plan, though it is not defined anywhere. But therefore it's OK to use it here as well.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Yes, the 1905 standard does not say it is for the controller but the reason for that is that IEEE 1905.1 does not define entities such as the "Controller" and "Agent". But the whole point of the message is that the controller will receive it.
So there is no point to send it if the controller will not receive it.

arnout: Not necessarily. An agent can also do topology queries. That can be used for keeping track of the backhaul link (#951) or for agent-initiated steering.

controller/src/beerocks/master/son_master_thread.cpp
arnout: After further discusion - it turns out that p_channel_center_frequency_index2 is missing from the Operating Channel Report TLV, so we *will* need to use this for 80+80MHz operation... Which is super-weird...

PR #932, opened 2020-03-08
Feature/autoconfig search handshake

### LINKED ISSUE #423: [TASK] Pass flow 4.12.1 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.12.1 in the certification testbed.

# Exit criteria

### LINKED ISSUE #424: [TASK] Pass flow 4.12.2 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.12.2 in the certification testbed.

# Exit criteria

### LINKED ISSUE #425: 4.12.1,2 Higher layer data payload over 1905
The prplMesh agent on Turris Omnia with OpenWrt needs to pass the certification test cases 4.12.1 and 4.12.2 as defined by the Wi-Fi CERTIFIED EasyMesh (tm) Test Plan.

- [ ] #423 Pass flow 4.12.1 test in testbed
- [ ] #424 Pass flow 4.12.2 test in testbed

### LINKED ISSUE #675: [TASK] prplMesh autoconfig handshake
## Description

Currently, prplMesh agent and controller always send VS TLVs to some messages even though it is possible the other side will not process these TLVs. Moreover, some testbed certified units don't cope well with fragmented CMDUs (R1 certification has no test for fragmentation/reassembly), so when adding a VS TLV big enough, the CMDU gets fragmented in the transport and then ignored by these units, causing autoconfiguration to fail.

It is possible to avoid this altogether by adding a handshake during the autoconfig search-reply phase - the prplMesh agent will add a VS TLV to the autoconfig search, and the controller will mark this agent as prplMesh if it contains this VS TLV and will answer it with an autoconfig response with a similar VS TLV. The agent will then understand that the controller is also prplMesh, so VS TLVs and VS CMDUs can be used.

In addition, we can use the handshake VS TLVs to add a list of extra supported features that are not part of the specification which can be used in the future by different vendors to differentiate their products.

## Exit criteria

- prplMesh agent always sends autoconfig search with an extra VS TLV with a currently empty list of extra supported features.
- prplMesh controller marks the agent as prplMesh if it sent the autoconfig search with that TLV, and can send VS TLVs and CMDUs to it from now on. Otherwise, it ceases to send VS TLVs to this agent.
-prplMesh agent does the same - if the autoconfig search response contains the VS TLV, it can continue sending VS TLVs and VS CMDUs. Otherwise, it ceases to send those to the controller.

### LINKED ISSUE #921: [TASK] Test onboarding with all controllers
## Description

It seems there was an issue with a test when using a Qualcomm controller, but not with Marvell.
To see if this is a problem with the hardware or with the test, we should really do an onboarding test with all possible configurations, to weed out the various issues.

tomereli:
Currently, prplMesh agent and controller always send VS TLVs to some messages even though it is possible the other side will not process these TLVs. Moreover, some testbed certified units don't cope well with fragmented CMDUs (R1 certification has no test for fragmentation/reassembly), so when adding a VS TLV big enough, the CMDU gets fragmented in the transport and then ignored by these units, causing autoconfiguration to fail.

It is possible to avoid this altogether by adding a handshake during the autoconfig search-reply phase - the prplMesh agent will add a VS TLV to the autoconfig search, and the controller will mark this agent as prplMesh if it contains this VS TLV and will answer it with an autoconfig response with a similar VS TLV. The agent will then understand that the controller is also prplMesh, so VS TLVs and VS CMDUs can be used.

This PR adds the basic handshake using an empty handshake request and reply messages attached to the autoconfig search and response messages.
By doing this, we mark the prplmesh controller as prplmesh or not prplmesh, and can use it to decide if to send VS TLVs and VS messages later on in the agent.

In addition, update the agent's node in the controller DB so that in a future PR we can use it to refrain from sending VS messages to non-prplmesh agents.

Fixes #675 
Fixes #425 
Fixes #423 
Fixes #424 
Fixes #921 

Passing 4.12.1 - https://gitlab.com/prpl-foundation/prplMesh/pipelines/124523357
Passing 4.12.2 - https://gitlab.com/prpl-foundation/prplMesh/-/jobs/464474890


common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -30,18 +30,20 @@ eActionOp_CONTROL:
...
-  ACTION_CONTROL_CONTROLLER_PING_REQUEST: 4
-  ACTION_CONTROL_CONTROLLER_PING_RESPONSE: 5
-  ACTION_CONTROL_AGENT_PING_REQUEST: 6
-  ACTION_CONTROL_AGENT_PING_RESPONSE: 7
-  ACTION_CONTROL_SLAVE_JOINED_4ADDR_MODE_NOTIFICATION: 8
-  ACTION_CONTROL_CHANGE_MODULE_LOGGING_LEVEL: 9
-  ACTION_CONTROL_ARP_QUERY_REQUEST: 10
-  ACTION_CONTROL_ARP_QUERY_RESPONSE: 11
-  ACTION_CONTROL_PLATFORM_OPERATIONAL_NOTIFICATION: 12
+  ACTION_CONTROL_SLAVE_HANDSHAKE_REQUEST: 1
```
vitalii-komisarenko, line 33: It breaks backward compartibility.
Are there on the market any older devices?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -3599,3 +3599,16 @@ void db::clear_vap_list()
 }
 
 const std::shared_ptr<db::vaps_list_t> db::get_vap_list() { return m_vap_list; }
+
+bool db::is_prplmesh(const std::string &mac)
+{
+    auto node = get_node(mac);
+    return node ? node.get()->is_prplmesh : false;
```
vitalii-komisarenko: Maybe I should have added a debug output here to signal that the node is not in the DB yet and we read `is_prplmesh` too early.

vitalii-komisarenko (approved):

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
tomereli: prplMesh is not deployed it, it is in development, so no.

rmelotte:
Also fixes the issue seen in https://github.com/prplfoundation/prplMesh/issues/921, so I linked it.

rmelotte (approved):
:rocket: 
I added the don't merge tag just in case you want to squash the commits updating the generated files.

tomereli:
> 🚀
> I added the don't merge tag just in case you want to squash the commits updating the generated files.

Good thing you did - I'd like @arnout to also review this before we merge.

controller/src/beerocks/master/db/db.cpp
tomereli: Fixed.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -3600,3 +3600,20 @@ void db::clear_vap_list()
 }
 
 const std::shared_ptr<db::vaps_list_t> db::get_vap_list() { return m_vap_list; }
+
+bool db::is_prplmesh(const std::string &mac)
```
arnout: I *don't* want new interfaces that use strings for MAC addresses.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -3600,3 +3600,20 @@ void db::clear_vap_list()
...
+    if (!node) {
+        LOG(ERROR) << "can't find node with mac " << mac << ", consider as not prplmesh";
+        return false;
+    }
+    return node->is_prplmesh;
+}
+void db::set_prplmesh(const std::string &mac)
+{
+    if (!get_node(mac)) {
+        add_node(network_utils::mac_from_string(mac));
```
arnout: I don't think a new node should be added in this case. The type of the node is going to be wrong. When we implement #524 we will need to know which type of node to create, so this isn't even possible.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2197,6 +2198,16 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
 
+    auto beerocks_header = message_com::parse_intel_vs_message(cmdu_rx);
+    if (beerocks_header &&
+        beerocks_header->action_op() == beerocks_message::ACTION_CONTROL_SLAVE_HANDSHAKE_RESPONSE) {
+        // mark controller as prplMesh
+        LOG(DEBUG) << "prplMesh controller: received ACTION_CONTROL_SLAVE_HANDSHAKE_RESPONSE from "
+                   << src_mac;
+        prplmesh_controller = true;
+    } else {
+        LOG(DEBUG) << "Not prplMesh controller " << src_mac;
```
arnout, line 2211: Since we get multiple responses in the backhaul manager, perhaps we should log an error message if we already had prplmesh_controller == true. And we should probably reset it to false as well.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -380,6 +380,7 @@ bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
         network_utils::iface_info iface_info;
         bool backhaul_manager_exist = false;
 
+        notification->params().prplmesh_controller = pSocket->prplmesh_controller;
```
arnout: I don't think there's a need to save this flag per socket. We can just use `this->prplmesh_controller` here.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -108,6 +108,7 @@ class backhaul_manager : public btl::transport_socket_thread {
         bool sta_iface_filter_low        = false;
         bool slave_is_backhaul_manager   = false;
         bool controller_discovered       = false;
+        bool prplmesh_controller         = false;
```
arnout: So I think this member is unnecessary.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -228,6 +229,7 @@ class backhaul_manager : public btl::transport_socket_thread {
     bool pending_enable = false;
 
     std::string controller_bridge_mac;
+    bool prplmesh_controller = false;
```
arnout: Nitpick: I prefer names of booleans to be clearly boolean, so `is_prplmesh_controller` or `controller_is_prplmesh`.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -579,7 +579,7 @@ sBackhaulParams:
   backhaul_is_wireless: uint8_t 
   backhaul_iface_type: uint8_t 
   is_backhaul_manager: uint8_t 
-  gw_manage_type: uint8_t
```
arnout, line 582: Removal of this unused field should be made explicit in the commit message.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1942,7 +1942,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         client_association_event_tlv->association_event() =
             wfa_map::tlvClientAssociationEvent::CLIENT_HAS_LEFT_THE_BSS;
 
-        if (config.no_vendor_specific) {
+        if (!backhaul_params.prplmesh_controller) {
             LOG(DEBUG) << "non-Intel, not adding ClientAssociationEvent VS TLV";
```
arnout: Maybe also change this to "non-prplmesh" then?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3737,6 +3737,9 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
     }
 
     if (cmdu_tx.getMessageType() == ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE) {
+        if (!backhaul_params.prplmesh_controller) {
```
arnout: Ideally, we should never even get here... So maybe throw a warning if we do?

tests/test_gw_repeater.sh
```diff
@@ -77,12 +77,10 @@ main() {
...
     [ "$START_REPEATER" = "true" ] && {
         index=0
-        no_vendor="-n"
         for repeater in $REPEATER_NAMES; do
             status "Start Repeater (Remote Agent): $repeater"
-            ${rootdir}/tools/docker/run.sh -u ${UNIQUE_ID} ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${repeater} -m aa:bb:cc:$index$index -- $no_vendor "$@"
+            ${rootdir}/tools/docker/run.sh -u ${UNIQUE_ID} ${VERBOSE_OPT} ${FORCE_OPT} start-agent -d -n ${repeater} -m aa:bb:cc:$index$index -- "$@"
             index=$((index+1))
-            no_vendor=
         done
```
arnout, line 84: So, this means we no longer have a test of an agent that acts as a non-prplmesh agent. Is that acceptable?

Maybe it's better to keep all this and just force the controller_is_prplmesh option to false if the no_vendor_specific option is set.

arnout (approved):
Nothing major so approving. Well, I consider the string MAC address a major issue but I assume you can fix it up :-)

There's a typo in one of the commit messages: resonse instead of response.

controller/src/beerocks/master/db/db.cpp
tomereli: I think we create the node only in during the autoconfig M1, so the node will always be created first when we get the autoconfig search request arrives, and if we don't create it here it we won't mark it as prplmesh when we mean to.
@vitalii-komisarenko please confirm.

tests/test_gw_repeater.sh
tomereli: I think it is acceptable since we are testing agent certification on a nightly basis with non-prplmesh controllers and controller certification with non-prplmesh agents.

controller/src/beerocks/master/db/db.cpp
tomereli: Ok.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: NACK - I did this here instead of not sending the messages and checking if the controller is prplMesh everywhere else where we send the VS message.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
tomereli: Oops I forgot to update the commit message, mergify beat me to it

PR #934, opened 2020-03-09
easylogging: resolveFilename: use same file for all debug levels

### LINKED ISSUE #832: [BUG] framework: logging may open two files
**Describe the bug** -

This was discovered in CI job https://gitlab.com/prpl-foundation/prplMesh/-/jobs/437812485

Apparently, for each debug level the log file name is calculated separately. Normally, it's always the same name, so the log file will be opened only once. However, if the second happens to tick by in the middle of the initialisation of the logging, we will have two different log files with 1 second difference, and half the log ends up in one log file, the other half in the other log file.

In the case of that CI job, it lead to a failure because we grep the log file to check the status of the agent, and we were grepping the wrong log file.

## Steps to reproduce

This is nearly impossible to reproduce...

For testing purposes it can probably be reproduced by adding a 1 second sleep at some point in the logging initialisation.

vitalii-komisarenko:
For each debug level the log file name is calculated separately. Normally, it's always the same name, so the log file will be opened only once. However, if the second happens to tick by in the middle of the initialisation of the logging, we will have two different log files with 1 second difference, and half the log ends up in one log file, the other half in the other log file.

In the case of that CI job, it lead to a failure because we grep the log file to check the status of the agent, and we were grepping the wrong log file.

This commit ensures that all file name calculations use the same time stamp.

### DIFF
--- a/framework/external/easylogging/easylogging++.cc
+++ b/framework/external/easylogging/easylogging++.cc
@@ -1758,6 +1758,19 @@ unsigned long TypedConfigurations::getULong(std::string confVal) {
   return atol(confVal.c_str());
 }
 
+std::string TypedConfigurations::getDateTimeForFilename(const std::string &fmt) {
+  m_dateTimeMutex.lock();
+
+  if (m_dateTimeNow.empty()) {
+    base::SubsecondPrecision ssPrec(3);
+    m_dateTimeNow = base::utils::DateTime::getDateTime(fmt.c_str(), &ssPrec);
+  }
+
+  m_dateTimeMutex.unlock();
+
+  return m_dateTimeNow;
+}
+
 std::string TypedConfigurations::resolveFilename(const std::string& filename) {
   std::string resultingFilename = filename;
   std::size_t dateIndex = std::string::npos;
@@ -1788,8 +1801,7 @@ std::string TypedConfigurations::resolveFilename(const std::string& filename) {
       } else {
         fmt = std::string(base::consts::kDefaultDateTimeFormatInFilename);
       }
-      base::SubsecondPrecision ssPrec(3);
-      std::string now = base::utils::DateTime::getDateTime(fmt.c_str(), &ssPrec);
+      std::string now = getDateTimeForFilename(fmt);
       base::utils::Str::replaceAll(now, '/', '-'); // Replace path element since we are dealing with filename
       base::utils::Str::replaceAll(resultingFilename, dateTimeFormatSpecifierStr, now);
     }
@@ -1847,6 +1859,7 @@ bool TypedConfigurations::unsafeValidateFileRolling(Level level, const PreRollOu
     ELPP_INTERNAL_INFO(1, "Truncating log file [" << fname << "] as a result of configurations for level ["
                        << LevelHelper::convertToString(level) << "]");
     fs->close();
+    m_dateTimeNow.clear();
     preRollOutCallback(fname.c_str(), currFileSize);
     fs->open(fname, std::fstream::out | std::fstream::trunc);
     return true;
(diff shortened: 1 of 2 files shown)


framework/external/easylogging/easylogging++.cc
```diff
@@ -1786,7 +1786,7 @@ std::string TypedConfigurations::resolveFilename(const std::string& filename) {
         fmt = std::string(base::consts::kDefaultDateTimeFormatInFilename);
       }
       base::SubsecondPrecision ssPrec(3);
-      std::string now = base::utils::DateTime::getDateTime(fmt.c_str(), &ssPrec);
+      static std::string now = base::utils::DateTime::getDateTime(fmt.c_str(), &ssPrec);
       base::utils::Str::replaceAll(now, '/', '-'); // Replace path element since we are dealing with filename
```
rmelotte, line 1805: I'm not sure about this change. I only took a quick look, but it seems like the time would now always be the same for all loggers, which is probably not what we want.

Isn't it enough to keep the time as a member variable?

Unrelated: please also wrap the commit message so that it stays under 72 chars long (see [CONTRIBUTING.md](https://github.com/prplfoundation/prplMesh/blob/master/CONTRIBUTING.md)).

vitalii-komisarenko: Indeed I wanted to use the same time stamp for all the loggers. (It would be clearly visible that a bunch of *.log files belong to a single run; it may help to examine the logs with tools like `find` and `grep`).
But if it is not what we want - yes, using a member variable would be preferable.

ghost (approved):

framework/external/easylogging/easylogging++.cc
```diff
@@ -1755,6 +1755,20 @@ unsigned long TypedConfigurations::getULong(std::string confVal) {
...
 }
 
+std::string TypedConfigurations::getDateTimeForFilename(const std::string &fmt) {
+  m_dateTimeMutex.lock();
+
+  if (!m_dateTimeCalculated) {
+    base::SubsecondPrecision ssPrec(3);
+    m_dateTimeNow = base::utils::DateTime::getDateTime(fmt.c_str(), &ssPrec);
+    m_dateTimeCalculated = true;
+  }
```
vitalybu: You don't need 2 member variables here.
Checking if `m_dateTimeNow` is empty should be enough:
```suggestion
  if (m_dateTimeNow.empty()) {
    base::SubsecondPrecision ssPrec(3);
    m_dateTimeNow = base::utils::DateTime::getDateTime(fmt.c_str(), &ssPrec);
  }
```

vitalybu (commented):
Not approving as I'm not sure if this change is required (especially since it's modifying a third-party library).
Perhaps the CI issue can be handled differently:
1. Allow specifying the log files DateTime format in the configuration files and drop the seconds.
2. Do not timestamp the log files at all when running the tests, since we'll probably have only one set files anyway.

arnout:
Unfortunately, I don't think this is acceptable...

There's another feature that we use, log cycling. I don't remember exactly how it's configured, but after some time the log files will be closed and reopened under a new name. However, the "new name" bit is just that it has a new timestamp. With this change, the timestamp is calculated only once, so log cycling will just reopen the same file again.

I could be wrong though - maybe the TypedConfigurations object is re-created when the logs are cycled, in that case all is well.

IIRC the log cycling is done within easylogging++ itself, so it should be possible to simply reset m_dateTimeNow when log cycling is done.

arnout:
> Not approving as I'm not sure if this change is required (especially since it's modifying a third-party library).

Well, it really is a bug in the upstream library. In easylogging++, if you want to put DEBUG and INFO logs in the same file, the only way to do that is to configure DEBUG and INFO to use the same filename (pattern). The problem is that if that pattern contains a timestamp, they *may* expand to different filenames.

> Perhaps the CI issue can be handled differently:

It's not just a CI issue, it's also a production issue. You really don't want the logs to be in two separate files.

PR #938, opened 2020-03-09
agent: channel sel: skip non-operable op classes

### LINKED ISSUE #937: [TASK] Pass test MAP-4.8.1_ETH_FH5GH in testbed
This issue is to keep track of the 5GH variant of test 4.8.1. The original issue for 4.8.1 was here: https://github.com/prplfoundation/prplMesh/issues/400

A job where the test is failing is here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/464453896

It currently fails with:
```
No association request from CTT STA to MAUT
```

It has never passed yet, and I suspect prplMesh is not operating on the correct channel.

If it's the case, it's probably not specific to 4.8.1 and will be fixed when fixing this issue: https://github.com/prplfoundation/prplMesh/issues/368

- [ ] when the issue is fixed, the test is added to nightly CI

tomereli:
Commit 82fe3c5785216af08e2b90737cb5228c436fffeb fixed the channel selection logic to ignore non-operable channels, but missed an important case where the channel preference channel list is empty, which means that all channels in that operating class are non-operable.

We missed it because the UCC is using op class 115 channel 36 in 4.2.1 test case, which is the first operating class we iterate on in channel_selection_select_channel loop when we try to find out if a specific channel from that operating class is non-operable.
Since channel 36 is the only channel which is not present in the preference channel list, it was selected as it should have.

In case of channel 149 however, the UCC marks the entire 115 operating class as non-operable, by having an empty channel list (according to Table 23 in the MultiAP Specification, an empty channel list field indicates that the indicated preference applies to all channels in the operating class).

The current channel selection logic iterates through all supported channels, and for each channel it checks if it is marked as non-operable by checking if it is part of the channel preference channel list.
Since the channel list for operating class 115 is empty, the first channel in that operating class is selected.

The solution is to simply skip the entire preference if the channel list is empty and is introduced in the first commit.

Another bug we have is that operating classes 128,129,130 defined in the `operating_classes_list` are wrong, these operating classes use center channels instead of starting channels and are using bandwidths 80 and 160.
However, hostapd reports the starting channel, for example, 36 instead of 42. This collides with the UCC command which is also following the 80211 specification and marks center channels as non-operable, which results with the agent concluding that the current channel (36 for that manner) is not restricted.

The solution for this is added in the third commit, which updates the channels for operating classes 128,129 and 130 to use center channels, and use the center channel when checking if the current channel is restricted if the bandwidth is 80MHz or 160MHz.

Fixes #937

MAP-4.8.1_ETH_ETH5GH status: https://gitlab.com/prpl-foundation/prplMesh/pipelines/124908827

### DIFF
--- a/common/beerocks/bcl/source/son/son_wireless_utils.cpp
+++ b/common/beerocks/bcl/source/son/son_wireless_utils.cpp
@@ -43,9 +43,11 @@ static const std::map<uint8_t, sOperatingClass> operating_classes_list = {
     {125,       {{149, 153, 157, 161, 165, 169},                               beerocks::BANDWIDTH_20}},
     {126,       {{149, 157},                                                   beerocks::BANDWIDTH_40}},
     {127,       {{153, 161},                                                   beerocks::BANDWIDTH_40}},
-    {128,       {{36,  40,  44,  48,  52,  56,  60,  64,  100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144, 149, 153, 157, 161}, beerocks::BANDWIDTH_80}},
-    {129,       {{50, 66, 82, 98, 114},                                        beerocks::BANDWIDTH_160}},
-    {130,       {{36,  40,  44,  48,  52,  56,  60,  64,  100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144, 149, 153, 157, 161}, beerocks::BANDWIDTH_160}}};
+//  {OP Class   {Channel center Frequency index},                              Bandwidth              }}
+    {128,       {{42, 58, 106, 122, 138, 155},                                 beerocks::BANDWIDTH_80}},
+    {129,       {{50, 114},                                                    beerocks::BANDWIDTH_160}},
+    {130,       {{42, 58, 106, 122, 138, 155},                                 beerocks::BANDWIDTH_80}}
+};
 // clang-format on
 
 constexpr beerocks::eWiFiAntNum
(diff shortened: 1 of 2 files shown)


rmelotte (approved):

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -645,7 +647,7 @@ uint8_t wireless_utils::get_operating_class_max_tx_power(
 /**
  * @brief get operating class number by channel and channel bandwidth 
  *
- * @param channel current channel
+ * @param channel current channel (center channel for bandwidth 80 and 160)
```
dob71: It's very easy to make a mistake if just the comment there explains that the caller must pass the center channel.
It might be better to rename the function to "get_operating_class_by_center_channel"

dob71 (approved):

common/beerocks/bcl/source/son/son_wireless_utils.cpp
tomereli: But it doesn't always accept center channel. If the BW is 20/40 - it uses the starting channel. If the BW is 80/160 - it uses the center channel. How would you do it?

tomereli: OK - see https://github.com/prplfoundation/prplMesh/pull/938/commits/cf6c38031fb9c9dca108927f0a199d033d7e9a28 if its better

PR #940, opened 2020-03-10
Make subset of python files PEP8-compatible

vitalii-komisarenko:
Making non-test-related python file pep8-compatible.


tools/commands/deploy.py
```diff
@@ -37,15 +35,17 @@ def __connect__(self):
...
+                            username=self.proxy['user'], password=self.proxy['pass'])
 
     def run(self, commands=[]):
         ''' Run commands on target via proxy '''
         for c in commands:
             logger.debug("@{} --> {}".format(self.proxy['ip'], c))
             stdin, stdout, stderr = self.client.exec_command(c)
             if stdout.channel.recv_exit_status():
-                logger.error("Command failed:\n{}\nexit code - {}, stderr - {}".format(c, stdout.channel.recv_exit_status(), stderr.read()))
+                logger.error("Command failed:\n{}\nexit code - {}, stderr - {}".format(
```
arnout, line 47: Maybe we should decide on a coding style for this kind of situation...

```
        "blahblahblah".format(
            x, y, z)
```

or

```
        "blahblahblah".\
            format(x, y, z)
```

Both are a bit ugly, but I have a slight preference for the second option.

Well, actually, I'd really prefer to use f-strings which would remove the need for ever using `"...".format()`, but apparently not everyone has python 3.6 :-(

tools/commands/deploy.py
```diff
@@ -107,33 +124,39 @@ def mkdir_p(path):
...
 
+
 class mapdeploy(object):
     def __init__(self, args):
-        if args.verbose: logger.setLevel(logging.DEBUG)
-        modules = deploy_modules if 'all' in args.modules else [m for m in deploy_modules if m in args.modules]
+        if args.verbose:
+            logger.setLevel(logging.DEBUG)
+        modules = deploy_modules if 'all' in args.modules else [
+            m for m in deploy_modules if m in args.modules]
```
arnout, line 139: Meh, it's a bit silly to fix this since it actually doesn't work any more... There are no longer any modules...

Oh well, let's just fix it for pep8.

tools/commands/config.py
```diff
@@ -113,15 +123,16 @@ def __guess_setup_id(self):
         try:
             from chdlab.commands.config import CHDLAB_config
             return CHDLAB_config().setup_number
-        except:
+        except Exception:
```
arnout, line 126: The previous commit still leaves a catch-all except, so why not do the same here?

tools/commands/build.py
```diff
@@ -42,8 +50,11 @@ def make(self):
         raise NotImplementedError('make() function must be overrided')
 
 # cmake-based builder class
+
+
```
arnout: The double newline should be *before* the command, the comment is tied to the cmakebuilder class.

tools/commands/build.py
```diff
@@ -85,8 +96,11 @@ def make(self):
         subprocess.check_call(cmd, env=self.env)
 
 # automake-based builder class
+
+
```
arnout: Same here, the comment is tied to the acbuilder class.

tools/commands/build.py
```diff
@@ -129,6 +143,8 @@ def make(self):
         subprocess.check_call(cmd, env=self.env)
 
 # hostap builder class
+
+
```
arnout: And same here again.

tools/commands/build.py
```diff
@@ -159,12 +175,15 @@ def make(self):
...
 class mapbuild(object):
     def __init__(self, args):
-        if args.verbose: logger.setLevel(logging.DEBUG)
+        if args.verbose:
+            logger.setLevel(logging.DEBUG)
 
         commands = args.commands
-        _dep_modules = dep_modules if 'all' in args.modules or 'dep' in args.modules else [m for m in dep_modules if m in args.modules]
+        _dep_modules = dep_modules if 'all' in args.modules or 'dep' in args.modules else [
+            m for m in dep_modules if m in args.modules]
```
arnout: I don't think this is a nice way to split. Better split on the else:

```suggestion
        _dep_modules = dep_modules if 'all' in args.modules or 'dep' in args.modules 
            else [m for m in dep_modules if m in args.modules]
```

arnout (approved):
I think it would be good to merge this soon. There's no need to have everything PEP8 compliant before we merge this.

Also, it may be a good idea to already add the pep8 check to CI, but just allow it to fail.

adam1985d:
> Making deploy.py PEP8-compartible as a part of
> preparation for automatic style checks of python files on CI.
> 
> Signed-off-by: Vitalii Komisarenko [vitalii.komisarenko@gmail.com](mailto:vitalii.komisarenko@gmail.com)

I think you meant `compatible` and not `compartible` - both in PR description and in PR name

this comment is also relevant to all commit headlines and messages

tools/commands/config.py
vitalii-komisarenko: flake8 compains with `E722 do not use bare except`, so I add `Exception`

vitalii-komisarenko:
beerocks_message_endian_converter.py is a python2 file.
Shall it be converted to python3?

tomereli:
> beerocks_message_endian_converter.py is a python2 file.
> Shall it be converted to python3?

I think we can delete it. Right @vitalybu ?

tools/beerocks_analyzer/beerocks_analyzer.py
```diff
@@ -13,46 +13,54 @@
...
+    logger.info("    -gw_ip=<IP of GW>                                     "
+                "- GW ip to get updates from. If empty, the GW is assumed to be running locally")
+    logger.info("    -bin_path=<path_to_beerocks_cli>                      "
+                "- Path to the beerocks_cli binary. If empty, use build from <top_level>/../")
+    logger.info("    -my_ip=[IP of PC]                                     "
+                "- PC ip to send updates, if not set will automaticly set")
+    logger.info("    -ssh_port=[ssh port of GW]                            "
+                "- SSH port used to connect to GW")
+    logger.info("    -docker_container=[name of container]                 "
+                "- Name of the docker container on which to run beerocks_cli")
```
rmelotte: What a mess... This shouldn't even be here, it should be handled by a library like argparse.

For now if you have the opportunity, could you just remove all the extra spaces before the second dash? So the first one would be like this for example:
```
"    -f=<input log file> - input log file"
```
It should make it look a tiny bit less messy.

tools/beerocks_analyzer/beerocks_analyzer_widget.py
```diff
@@ -184,19 +195,25 @@ def getCommandLineArgs(self, argv):
...
     def printUsage(self):
         self.logger.info("usage: BeeRocksAnalyzer [options]\n")
-        self.logger.info("    -f=<log file>                                         - input log file\n")
-        self.logger.info("    -plot<fig num><sub plot>=<Label 1>,<Label 1>,..       - enable and configure plots\n")
-        self.logger.info("    -realtime=<update interval sec>,<samples>             - enable real time plot\n")
-        self.logger.info("    -print_vals                                           - print log values to console\n")
-        self.logger.info("    -print_labels                                         - print log labels to console\n")
-        self.logger.info("    -output_csv                                           - output params to csv file\n")
+        self.logger.info(
+            "    -f=<log file>                                     - input log file\n")
```
rmelotte: Same here, if you have the time/opportunity, please remove all the extra spaces after the single dash.

rmelotte (approved):
Pretty hard to review in details of course, but overall LGTM.

Can you also add the ones that passes flake8 to `FLAKE8_FILES` in `.gitlab-ci.yml`?

vitalybu:
> > beerocks_message_endian_converter.py is a python2 file.
> > Shall it be converted to python3?
> 
> I think we can delete it. Right @vitalybu ?

Yup

tools/commands/deploy.py
vitalii-komisarenko: Fixed for PEP8

.gitlab-ci.yml
```diff
@@ -28,7 +28,7 @@ flake8:
   stage: build
   image: pipelinecomponents/flake8:f087c4c
   variables:
-    FLAKE8_FILES: "tests/*.py"
+    FLAKE8_FILES: "tests/*.py tools/beerocks_analyzer/conf/conf1.py tools/beerocks_analyzer/*.py tools/commends/*.py"
```
arnout, line 31:
```suggestion
    FLAKE8_FILES: "tests/*.py tools/beerocks_analyzer/conf/conf1.py tools/commands/*.py"
```

vitalii-komisarenko: beerocks_analyzer already passes the code check.
typo commEnds->commAnds is addressed in https://github.com/prplfoundation/prplMesh/pull/1118

Issue #686, opened 2020-01-16
[TASK][Channel scan] use AP Radio Basic Capabilities TLV  to validate channel pool


kantera800:
currently channel pool validated by static table supported_channels_array[] of channels in db::set_dcs_channel_pool()

This is  incorrect  since radio setup could vary with different country configuration.

We need permanent and generic prplMesh solution:
- [x] refactor read_supported_channels to use get_radio_info #1024  (PR #1043)
- [x] update Radio Basic Capabilities TLV using the new supported_channels (PR  #1065)
- [x] [bug] missing channels when using nl80211_client::get_radio_info #1046 (PR #1053)
- [x]  use supported_channels from AP Radio Basic Capabilities data for channel scan pool validation in db instead of static table (PR #1093)

poc PR - do not merge #941

arnout:
> create new structure in node.h std::vector<beerocks_message::sWifiChannel> phy_supported_channels
> save GET_HW_FEATURES report during in slave_join to DB.

This is exactly what ap_wlan_hal->read_supported_channels() does, no? So what exactly needs to be done here?

kantera800:
> > create new structure in node.h std::vector<beerocks_message::sWifiChannel> phy_supported_channels
> > save GET_HW_FEATURES report during in slave_join to DB.
> 
> This is exactly what ap_wlan_hal->read_supported_channels() does, no? So what exactly needs to be done here?

in case we are currently on a static channel and we want to scan , read_supported_channels() returns just this channel.
also ,  i heard that read_supported_channels() is not going to be used soon.
we need to store in the db latest channel list that fw supports for the dcs feature.

PR #941, opened 2020-03-10
HW Supported Channels for channel pool validation

### LINKED ISSUE #961: [BUG] agent: MAP-4.5.1_ETH fails with "No Channel Preference Report message found"
Job: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/468547961

logs: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/468547961/artifacts/browse/logs/FAIL/MAP-4.5.1_ETH/

itayx:
currently channel pool validated by static table supported_channels_array[] of channels in db::set_dcs_channel_pool()

This is  incorrect  since radio setup could vary with different country configuration.

We need permanent prplmesh solution:

create  new structure  in node.h     std::vector<beerocks_message::sWifiChannel> phy_supported_channels
save GET_HW_FEATURES report during in slave_join to DB.
see ap_manager_thread::handle_hostapd_attached

ap_wlan_hal->read_supported_channels();

use it to Validate new dcs channel pool in db::set_dcs_channel_pool()
remove static table supported_channels_array[]  from db::set_dcs_channel_pool()

Task #686 
Fixes #961

### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1469,9 +1469,13 @@ void ap_manager_thread::handle_hostapd_attached()
 {
     LOG(DEBUG) << "handling enabled hostapd";
 
+    // Read HW supported channels by default.
+    ap_wlan_hal->read_supported_channels();
+    
     if (acs_enabled) {
         LOG(DEBUG) << "retrieving ACS report";
         int read_acs_attempt = 0;
+        // During each attempt the supported_channels are wiped and reset.
         while (!ap_wlan_hal->read_acs_report()) {
             read_acs_attempt++;
             if (read_acs_attempt >= READ_ACS_ATTEMPT_MAX) {
@@ -1483,8 +1487,6 @@ void ap_manager_thread::handle_hostapd_attached()
 
             usleep(ACS_READ_SLEEP_USC);
         }
-    } else {
-        ap_wlan_hal->read_supported_channels();
     }
 
     auto notification =
@@ -1516,6 +1518,8 @@ void ap_manager_thread::handle_hostapd_attached()
 
     // Copy the channels supported by the AP
     copy_radio_supported_channels(ap_wlan_hal, notification->params().supported_channels);
+    // Copy the channels supported by the AP
+    copy_radio_supported_channels(ap_wlan_hal, notification->params().hw_supported_channels, true);
 
     LOG(INFO) << "send ACTION_APMANAGER_JOINED_NOTIFICATION";
     LOG(INFO) << " mac = " << ap_wlan_hal->get_radio_mac();
@@ -1526,6 +1530,8 @@ void ap_manager_thread::handle_hostapd_attached()
     LOG(INFO) << " current bw = " << ap_wlan_hal->get_radio_info().bandwidth;
     LOG(INFO) << " supported_channels = " << std::endl
               << get_radio_supported_channels_string(ap_wlan_hal);
+    LOG(INFO) << " hw_supported_channels = " << std::endl
+              << get_radio_supported_channels_string(ap_wlan_hal, true);
 
     // Send CMDU
     message_com::send_cmdu(slave_socket, cmdu_tx);
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -34,9 +34,12 @@ using namespace beerocks::net;
 //////////////////////////////////////////////////////////////////////////////
 
 static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
-                                          beerocks::message::sWifiChannel supported_channels[])
+                                          beerocks::message::sWifiChannel supported_channels[],
+                                          bool use_hw_supported_channels = false)
 {
-    auto radio_channels = ap_wlan_hal->get_radio_info().supported_channels;
+    auto radio_channels = use_hw_supported_channels
+                              ? ap_wlan_hal->get_hw_supported_channels()
+                              : ap_wlan_hal->get_radio_info().supported_channels;
 
     // Copy the channels
     for (uint i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH && i < radio_channels.size();
@@ -67,10 +70,15 @@ static int8_t get_tx_power(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
 }
 
 static std::string
-get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
+get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
+                                    bool use_hw_supported_channels = false )
 {
+    auto supported_channels = use_hw_supported_channels
+                                  ? ap_wlan_hal->get_hw_supported_channels()
+                                  : ap_wlan_hal->get_radio_info().supported_channels;
     std::ostringstream os;
-    for (auto val : ap_wlan_hal->get_radio_info().supported_channels) {
+    os << " Count: " << supported_channels.size();
+    for (auto val : supported_channels) {
         if (val.channel > 0) {
             os << " ch = " << int(val.channel) << " | dfs = " << int(val.tx_pow)
                << " | tx_pow = " << int(val.is_dfs) << " | noise = " << int(val.noise)
--- a/common/beerocks/bwl/common/base_wlan_hal.cpp
+++ b/common/beerocks/bwl/common/base_wlan_hal.cpp
@@ -32,6 +32,7 @@ base_wlan_hal::base_wlan_hal(HALType type, std::string iface_name, IfaceType ifa
 
     // Initialize complex containers of the radio_info structure
     m_radio_info.supported_channels.resize(128 /* TODO: Get real value */);
+    m_hw_supported_channels.resize(128 /* TODO: Get real value */);
 }
 
 base_wlan_hal::~base_wlan_hal()
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -1550,7 +1550,7 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
     FieldsToParse fieldsToParse[] = {
         {(void *)&chan, &numOfValidArgs[0], DWPAL_INT_PARAM, "chan=", 0},
         {(void *)dfs_state, &numOfValidArgs[1], DWPAL_STR_PARAM,
-         "(DFS state = ", sizeof(dfs_state)},
+        "(DFS state = ", sizeof(dfs_state)},
         /* Must be at the end */
         {NULL, NULL, DWPAL_NUM_OF_PARSING_TYPES, NULL, 0}};
 
@@ -1569,9 +1569,14 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
             return false;
         }
 
-        // LOG(DEBUG) << "numOfValidArgs[0]= " << numOfValidArgs[0] << " chan= " << chan;
-        // LOG(DEBUG) << "numOfValidArgs[1]= " << numOfValidArgs[1] << " dfs_state= " << dfs_state;
-
+        LOG(DEBUG) << "numOfValidArgs[0]= " << numOfValidArgs[0] << " chan= " << chan;
+        LOG(DEBUG) << "numOfValidArgs[1]= " << numOfValidArgs[1] << " dfs_state= " << dfs_state;
+        
+        m_hw_supported_channels[i].bandwidth   = 20;
+        m_hw_supported_channels[i].channel     = chan;
+        m_hw_supported_channels[i].bss_overlap = 0;
+        m_hw_supported_channels[i].is_dfs      = numOfValidArgs[1];
+        
         m_radio_info.supported_channels[i].bandwidth   = 20;
         m_radio_info.supported_channels[i].channel     = chan;
         m_radio_info.supported_channels[i].bss_overlap = 0;
--- a/common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_common.h
+++ b/common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_common.h
@@ -371,16 +371,23 @@ typedef struct sNodeHostap {
     uint8_t conducted_power;
     char driver_version[beerocks::message::WIFI_DRIVER_VER_LENGTH];
     beerocks::message::sWifiChannel supported_channels[beerocks::message::SUPPORTED_CHANNELS_LENGTH];
+    beerocks::message::sWifiChannel hw_supported_channels[beerocks::message::SUPPORTED_CHANNELS_LENGTH];
     void struct_swap(){
         iface_mac.struct_swap();
         for (size_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++){
             (supported_channels[i]).struct_swap();
+        }
+        for (size_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++){
+            (hw_supported_channels[i]).struct_swap();
         }
     }
     void struct_init(){
         iface_mac.struct_init();
             for (size_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
                 (supported_channels[i]).struct_init();
+            }
+            for (size_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+                (hw_supported_channels[i]).struct_init();
             }
     }
 } __attribute__((packed)) sNodeHostap;
--- a/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
+++ b/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
@@ -271,6 +271,9 @@ sNodeHostap:
   supported_channels:
     _type: beerocks::message::sWifiChannel 
     _length: [ "beerocks::message::SUPPORTED_CHANNELS_LENGTH" ]          
+  hw_supported_channels:
+    _type: beerocks::message::sWifiChannel 
+    _length: [ "beerocks::message::SUPPORTED_CHANNELS_LENGTH" ]          
 
 sVapsList:
   _type: struct
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -1165,7 +1165,7 @@ int db::get_hostap_conducted_power(std::string mac)
 }
 
 bool db::set_hostap_supported_channels(std::string mac, beerocks::message::sWifiChannel *channels,
-                                       int length)
+                                       int length, bool use_hw_supported_channels)
 {
     auto n = get_node(mac);
     if (!n) {
@@ -1176,13 +1176,16 @@ bool db::set_hostap_supported_channels(std::string mac, beerocks::message::sWifi
         return false;
     }
     std::vector<beerocks::message::sWifiChannel> supported_channels_(channels, channels + length);
-    n->hostap->supported_channels = supported_channels_;
+
+    (use_hw_supported_channels ? n->hostap->hw_supported_channels : n->hostap->supported_channels)
+        .assign(supported_channels_.begin(), supported_channels_.end());
 
     if (n->hostap->supported_channels.size() == 0) {
         LOG(ERROR) << "No supported channels";
         return false;
     }
 
+    if (!use_hw_supported_channels) {
     if (wireless_utils::which_freq(n->hostap->supported_channels[0].channel) ==
         eFreqType::FREQ_5G) {
         n->supports_5ghz = true;
@@ -1194,12 +1197,13 @@ bool db::set_hostap_supported_channels(std::string mac, beerocks::message::sWifi
                    << int(n->hostap->supported_channels[0].channel);
         return false;
     }
+    }
 
     return true;
 }
 
 const std::vector<beerocks::message::sWifiChannel>
-db::get_hostap_supported_channels(std::string mac)
+db::get_hostap_supported_channels(std::string mac, bool use_hw_supported_channels)
 {
     auto n = get_node(mac);
     if (!n) {
@@ -1209,13 +1213,15 @@ db::get_hostap_supported_channels(std::string mac)
         LOG(WARNING) << __FUNCTION__ << "node " << mac << " is not a valid hostap!";
         return std::vector<beerocks::message::sWifiChannel>();
     }
-    return n->hostap->supported_channels;
+    return use_hw_supported_channels ? n->hostap->hw_supported_channels
+                                     : n->hostap->supported_channels;
 }
 
-std::string db::get_hostap_supported_channels_string(const std::string &radio_mac)
+std::string db::get_hostap_supported_channels_string(const std::string &radio_mac,
+                                                     bool use_hw_supported_channels)
 {
     std::ostringstream os;
-    auto supported_channels = get_hostap_supported_channels(radio_mac);
+    auto supported_channels = get_hostap_supported_channels(radio_mac, use_hw_supported_channels);
     for (const auto &val : supported_channels) {
         if (val.channel > 0) {
             os << " ch = " << int(val.channel) << " | dfs = " << int(val.is_dfs_channel)
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -1184,7 +1184,7 @@ bool db::set_hostap_supported_channels(std::string mac, beerocks::message::sWifi
         LOG(ERROR) << "No supported channels";
         return false;
     }
-
+    
     if (!use_hw_supported_channels) {
     if (wireless_utils::which_freq(n->hostap->supported_channels[0].channel) ==
         eFreqType::FREQ_5G) {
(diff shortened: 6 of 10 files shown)


common/beerocks/bwl/common/base_wlan_hal.cpp
```diff
@@ -32,6 +32,7 @@ base_wlan_hal::base_wlan_hal(HALType type, std::string iface_name, IfaceType ifa
 
     // Initialize complex containers of the radio_info structure
     m_radio_info.supported_channels.resize(128 /* TODO: Get real value */);
+    m_hw_supported_channels.resize(128 /* TODO: Get real value */);
```
kantera800, line 35: why copy paste,   
find the max available channels , use the 5G (it has more channels) 

edited:
why not use also SUPPORTED_CHANNELS_LENGTH

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1569,9 +1569,14 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
             return false;
         }
 
-        // LOG(DEBUG) << "numOfValidArgs[0]= " << numOfValidArgs[0] << " chan= " << chan;
-        // LOG(DEBUG) << "numOfValidArgs[1]= " << numOfValidArgs[1] << " dfs_state= " << dfs_state;
-
+        LOG(DEBUG) << "numOfValidArgs[0]= " << numOfValidArgs[0] << " chan= " << chan;
```
kantera800, line 1572: please do cleanups in separate commit

**edited:**
 leave there prints , its for debugging.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1569,9 +1569,14 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
...
+        LOG(DEBUG) << "numOfValidArgs[1]= " << numOfValidArgs[1] << " dfs_state= " << dfs_state;
+        
+        m_hw_supported_channels[i].bandwidth   = 20;
+        m_hw_supported_channels[i].channel     = chan;
+        m_hw_supported_channels[i].bss_overlap = 0;
```
kantera800, line 1577: I'm pretty sure we don't need or use "bandwidth" and "bss_overlap" in the m_hw_supported_channels structure

  GET_HW_FEATURES reply from hostapd gives us:
Allowed channel: mode=2 chan=144 freq=5720 MHz max_tx_power=24 dBm (DFS state = available)

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1508,6 +1518,8 @@ void ap_manager_thread::handle_hostapd_attached()
 
     // Copy the channels supported by the AP
     copy_radio_supported_channels(ap_wlan_hal, notification->params().supported_channels);
+    // Copy the channels supported by the AP
+    copy_radio_supported_channels(ap_wlan_hal, notification->params().hw_supported_channels, true);
```
kantera800, line 1522: this is not correct , you shouldn't **over write** supported channels with hw_supported channels.
For now hw_supported channels should be copied  along with the supported_channels in the  SLAVE_JOIN message 

in the future supported_channels will be removed.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -34,9 +34,12 @@ using namespace beerocks::net;
 //////////////////////////////////////////////////////////////////////////////
 
 static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
-                                          beerocks::message::sWifiChannel supported_channels[])
+                                          beerocks::message::sWifiChannel supported_channels[],
+                                          bool use_hw_supported_channels = false)
```
kantera800, line 38: you shouldn't copy hw_supported_channels to supported_channels.
please remove this commit.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1165,7 +1165,7 @@ int db::get_hostap_conducted_power(std::string mac)
 }
 
 bool db::set_hostap_supported_channels(std::string mac, beerocks::message::sWifiChannel *channels,
-                                       int length)
+                                       int length, bool use_hw_supported_channels)
```
kantera800, line 1168: again something isn't right with your solution, 
this flag is not needed (same for the other functions)

edit:
please use dedicated  method for it 
`set_hw_supported_channels()`

kantera800 (changes requested):
the solution seems a bit off track.

**hw_supported_channels** holds a list of  channels that FW supports.
**supported_channels** holds a list of current radio supported channels.
so in case of static channel,  supported_channels will hold on the only channel it supports.

hw_supported_channels **shouldn't overwrite** AP's current supported_channels list.

please make the appropriate changes.

edit:
I can see now that there is no overwrite, but PR still needs work.

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: I don't overwrite the supported_channels
If you look at line 1522, it functions like so:
copy the supported channel from `ap_wlan_hal` into `hw_supported_channels` and use the HW supported channels in the `bwl`

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: like I said before, I don't copy the `hw_supported_channels` into `supported_channels`
since when called I use the notification params' **hw_supported_channels**

kantera800: In that case don't change this method.
just create a dedicated method for copying hw supported channels
`copy_hw_supported_channels()`

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -67,10 +70,15 @@ static int8_t get_tx_power(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
 }
 
 static std::string
-get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
+get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
```
kantera800, line 73: also here , please create a dedicated method
`get_hw_supported_channels_string()`

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1526,6 +1530,8 @@ void ap_manager_thread::handle_hostapd_attached()
     LOG(INFO) << " current bw = " << ap_wlan_hal->get_radio_info().bandwidth;
     LOG(INFO) << " supported_channels = " << std::endl
               << get_radio_supported_channels_string(ap_wlan_hal);
+    LOG(INFO) << " hw_supported_channels = " << std::endl
+              << get_radio_supported_channels_string(ap_wlan_hal, true);
```
kantera800, line 1534:
```suggestion
              << get_hw_supported_channels_string(ap_wlan_hal);
```

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: then its very confusing.
If you want to copy hw_supported_channels to reply message then do it right after you read it from hostapd.
line 1473
`ap_wlan_hal->read_supported_channels();`

also, don't reuse `copy_radio_supported_channels()` by adding a flag, use dedicated function.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -271,6 +271,9 @@ sNodeHostap:
   supported_channels:
     _type: beerocks::message::sWifiChannel 
     _length: [ "beerocks::message::SUPPORTED_CHANNELS_LENGTH" ]          
+  hw_supported_channels:
```
tomereli, line 274: Wait - what is the difference between `supported_channels` and `hw_supported_channels`?
The commit message should explain what are the AP supported channels and what are the HW supported channels.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_common.h
```diff
@@ -371,17 +371,24 @@ typedef struct sNodeHostap {
     uint8_t conducted_power;
     char driver_version[beerocks::message::WIFI_DRIVER_VER_LENGTH];
     beerocks::message::sWifiChannel supported_channels[beerocks::message::SUPPORTED_CHANNELS_LENGTH];
+    beerocks::message::sWifiChannel hw_supported_channels[beerocks::message::SUPPORTED_CHANNELS_LENGTH];
```
tomereli, line 374: Can be squashed to the previous commit.

tomereli (commented):
I can't really review this before I understand what are the AP supported channels and what are the HW supported channels. This should be explained in the PR and in the commits which add the HW supported channels.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1483,8 +1487,6 @@ void ap_manager_thread::handle_hostapd_attached()
 
             usleep(ACS_READ_SLEEP_USC);
         }
-    } else {
-        ap_wlan_hal->read_supported_channels();
```
kantera800, line 1479: commit message should mention such changes.
something like:

> 
> We want always to save GET_HW_FEATURES when hostapd attached.
read_supported_channels() reads from GET_HW_FEATURES channel report and save supported channel list .
> But in case ACS report exists then its used instead of reading GET_HW_FEATURES .
> Since supported channels container is cleaned before reading the acs report read_supported_channels() can be read out side the **acs_enabled** condition.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1550,7 +1550,7 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
     FieldsToParse fieldsToParse[] = {
         {(void *)&chan, &numOfValidArgs[0], DWPAL_INT_PARAM, "chan=", 0},
         {(void *)dfs_state, &numOfValidArgs[1], DWPAL_STR_PARAM,
-         "(DFS state = ", sizeof(dfs_state)},
+        "(DFS state = ", sizeof(dfs_state)},
```
arnout, line 1553: This change makes clang-format fail.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1184,7 +1184,7 @@ bool db::set_hostap_supported_channels(std::string mac, beerocks::message::sWifi
         LOG(ERROR) << "No supported channels";
         return false;
     }
-
+
```
arnout, line 1187: Another clang-format error is introduced here.

arnout (changes requested):
I can't say I'm happy with this PR...

First of all I think the naming is off. AFAIU, what is called "hw_supported_channels" here, really is "supported_channels", and what is called "supported_channels" is actually "channel_preferences". I think a few of @kantera800's comments are triggered by this confusion.

Second, I don't think it's right to use the sWifiChannel structure for something which is just a channel number, not an actual channel (with bandwidth).

Most importantly, though, it is yet again going in the wrong direction, using a VS message instead of the EasyMesh message. In this case, the EasyMesh way is already implemented - with AP Radio Basic Capabilities. So the controller should just parse that and save it in the database, rather than adding an extra set of channels to the VS message. Note that this will also cover my second general remark.

I can be convinced to relax on these general worries, but for now it's changes requested.

kantera800:
> I can't say I'm happy with this PR...
> 
> First of all I think the naming is off. AFAIU, what is called "hw_supported_channels" here, really is "supported_channels", and what is called "supported_channels" is actually "channel_preferences". I think a few of @kantera800's comments are triggered by this confusion.
> 
> Second, I don't think it's right to use the sWifiChannel structure for something which is just a channel number, not an actual channel (with bandwidth).
> 
> Most importantly, though, it is yet again going in the wrong direction, using a VS message instead of the EasyMesh message. In this case, the EasyMesh way is already implemented - with AP Radio Basic Capabilities. So the controller should just parse that and save it in the database, rather than adding an extra set of channels to the VS message. Note that this will also cover my second general remark.
> 
> I can be convinced to relax on these general worries, but for now it's changes requested.

@arnout there is no confusion , I just wasn't sure if easymesh protocol already covers it so we went on the VS implementation to get the job done.
Now that you mention it easymesh way is better solution.

I checked the spec and current master implementation this is what i found:
Currently we build tlvApRadioBasicCapabilities and send it at slave_join.
but, the operation class list is generated still from the **supported_channels** list .
AFAIU in case of static channel we will have only 1 operating class in the tlvApRadioBasicCapabilities as a result  we could scan only 1 channel from the DCS channel pool.
So i see the following changes that has to be done:
1.  rename hw_supported_channels to supported_channels
2.  rename supported_channels to channel_preferences
3.  in add_ap_radio_basic_capabilities,  build  operation class list from **supported_channels**
4.  in controller, validate DCS channel pool against supported_channel list from previously saved tlvApRadioBasicCapabilities  data.

is this what you had in mind ?

edit:
@tomereli Just saw that in handle_cmdu_1905_autoconfiguration_WSC we are currently validating bss_info with radio_basic_caps so, AFAIU this change is needed unless we wouldn't pass certification when the radio operate with static channel.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1469,9 +1469,13 @@ void ap_manager_thread::handle_hostapd_attached()
...
     LOG(DEBUG) << "handling enabled hostapd";
 
+    // Read HW supported channels by default.
+    ap_wlan_hal->read_supported_channels();
+    
     if (acs_enabled) {
         LOG(DEBUG) << "retrieving ACS report";
         int read_acs_attempt = 0;
+        // During each attempt the supported_channels are wiped and reset.
         while (!ap_wlan_hal->read_acs_report()) {
```
tomereli, line 1479: I agree with @arnout - The ACS report provides the channel preference. The supported channels are the hw supported channels. This name change is critical.

arnout:
> So i see the following changes that has to be done:
> 
>     1. rename hw_supported_channels to supported_channels
>     2. rename supported_channels to channel_preferences

I would swap those two :-)

>     3. in add_ap_radio_basic_capabilities,  build  operation class list from **supported_channels**
>     4. in controller, validate DCS channel pool against supported_channel list from previously saved tlvApRadioBasicCapabilities  data.
> 
> is this what you had in mind ?

It's exactly what I had in mind.

There's one more thing to take into account: the supported channels list should contain all supported bandwidths, so e.g. channel 52 is there twice, once for 20MHz and once for 40MHz. Also remember to take into account that the 80 and 160MHz bandwidths use center frequency, not base frequency. I *think* both are OK, but think about it as well.

kantera800:
> > So i see the following changes that has to be done:
> > ```
> > 1. rename hw_supported_channels to supported_channels
> > 2. rename supported_channels to channel_preferences
> > ```
> 
> I would swap those two :-)
👍 
> 
> There's one more thing to take into account: the supported channels list should contain all supported bandwidths, so e.g. channel 52 is there twice, once for 20MHz and once for 40MHz. Also remember to take into account that the 80 and 160MHz bandwidths use center frequency, not base frequency. I _think_ both are OK, but think about it as well.
👍 

@arnout one thing i still don't understand.
 after this change,  how the controller will be notified of radio's static channel , is there a diffrent TLV for that?

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli:
```suggestion
    copy_supported_channels(ap_wlan_hal, notification->params().supported_channels);
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1516,6 +1518,8 @@ void ap_manager_thread::handle_hostapd_attached()
 
     // Copy the channels supported by the AP
     copy_radio_supported_channels(ap_wlan_hal, notification->params().supported_channels);
```
tomereli, line 1520:
```suggestion
    copy_channel_preference(ap_wlan_hal, notification->params().channel_preference);
```

controller/src/beerocks/master/db/db.cpp
tomereli: Please align to the new terminology - `set_hostap_supported_channels` and `set_hostap_channel_preference` - split this to 2 functions.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2276,16 +2276,13 @@ bool db::set_channel_scan_pool(const sMacAddr &mac, const std::unordered_set<uin
...
-    static size_t supported_channels_array_size =
-        sizeof(supported_channels_array) / sizeof(uint8_t);
-    std::set<uint8_t> supported_channels(supported_channels_array,
-                                         supported_channels_array + supported_channels_array_size);
+    auto supported_channels =
+        get_hostap_supported_channels(network_utils::mac_to_string(mac), true);
     for (auto channel : channel_pool) {
-        if (supported_channels.find(channel) == supported_channels.end()) {
+        if (std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&channel](const beerocks::message::sWifiChannel &wifi_channel) {
```
tomereli, line 2283:
```suggestion
                         [&](const beerocks::message::sWifiChannel &wifi_channel) {
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2276,16 +2276,13 @@ bool db::set_channel_scan_pool(const sMacAddr &mac, const std::unordered_set<uin
...
-        64, 100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144, 149, 153, 157, 161, 165};
-    static size_t supported_channels_array_size =
-        sizeof(supported_channels_array) / sizeof(uint8_t);
-    std::set<uint8_t> supported_channels(supported_channels_array,
-                                         supported_channels_array + supported_channels_array_size);
+    auto supported_channels =
+        get_hostap_supported_channels(network_utils::mac_to_string(mac), true);
     for (auto channel : channel_pool) {
-        if (supported_channels.find(channel) == supported_channels.end()) {
+        if (std::find_if(supported_channels.begin(), supported_channels.end(),
```
tomereli, line 2282:
```suggestion
        auto channel = std::find_if(supported_channels.begin(), supported_channels.end(),...);
        if (channel == supported_channels.end()) {...}
```

tomereli (changes requested):
Please refactor this PR to use supported channels and channel preference as it should, then ask for a re-review.

tomereli:
First - I fully agree with @arnout about the direction.

Regarding
> edit:
> @tomereli Just saw that in handle_cmdu_1905_autoconfiguration_WSC we are currently validating bss_info with radio_basic_caps so, AFAIU this change is needed unless we wouldn't pass certification when the radio operate with static channel.

Can you explain which test will fail due to this? Not sure I understand.

kantera800:
> First - I fully agree with @arnout about the direction.
> 
> Regarding
> 
> > edit:
> > @tomereli Just saw that in handle_cmdu_1905_autoconfiguration_WSC we are currently validating bss_info with radio_basic_caps so, AFAIU this change is needed unless we wouldn't pass certification when the radio operate with static channel.
> 
> Can you explain which test will fail due to this? Not sure I understand.
>
AFAIU in case of static channel, ACS report overwrites **supported channels** with just this channel.
So, I assume that all tests that during handle_cmdu_1905_autoconfiguration_WSC rely on radio_basic_caps to include all radio **supported channels**  will go to teardown.

tomereli:
> > First - I fully agree with @arnout about the direction.
> > Regarding
> > > edit:
> > > @tomereli Just saw that in handle_cmdu_1905_autoconfiguration_WSC we are currently validating bss_info with radio_basic_caps so, AFAIU this change is needed unless we wouldn't pass certification when the radio operate with static channel.
> > 
> > 
> > Can you explain which test will fail due to this? Not sure I understand.
> https://assets.grammarly.com/emoji/v1/1f91d.svg
> AFAIU in case of static channel, ACS report overwrites **supported channels** with just this channel.
> So, I assume that all tests that during handle_cmdu_1905_autoconfiguration_WSC rely on radio_basic_caps to include all radio **supported channels** will go to teardown.

You are correct. It is inherently wrong to use the ACS report for getting the supported channels, since I expect that if static channel is configured, the ACS report to be **empty** - I just verified this in the RAX40. @rmelotte this probably explains #961 -
In the CI, the DUT is rarely rebooted. The UCI default configuration is ACS, so `hostapd_cli -iwlan0 acs_report` will get some channel preference (not supported, but preferred).
Each time there is a scan, the ACS report gets updated. If there is no scan, the previous ACS report is fetched, but after 24 hours (according to one of my colleagues) it will be deleted, and the report will come out empty. This still has to be confirmed by the WLAN team in Intel, but a simple check on RAX40 in the CI will prove this check if you get acs_report, if its empty, reboot and check again. Wait 24 hours, then check again - should be empty again.

Bottom line is that I think this PR will solve #961 and possibly many others.

arnout:
> after this change, how the controller will be notified of radio's static channel , is there a diffrent TLV for that?

I'm not sure what you mean with "static channel".

If it is the channel that it's currently operating at: after a channel selection request, the agent is supposed to send an Operating Channel Report CMDU to the controller. It should also do that when it changes channels on its own initiative.

It would actually make sense to send an unsollicited Operating Channel Report CMDU immediately after autoconfig search completes (i.e. just before sending M1). Otherwise, indeed, there's no way for the controller to know the current operating channel.

kantera800:
> > after this change, how the controller will be notified of radio's static channel , is there a diffrent TLV for that?
> 
> I'm not sure what you mean with "static channel".

What i meant is that the radio **not** configured  with auto channel, so it can't switch channel on its own.
In this case the ACS report will have only one channel  (that's the "channel_preferences" we decided to rename earlier).
I want to make sure where exactly do we need to use channel_preferences (from acs) and where to use supported_channels (from radio capabilities /hw features) since this PR is going to change exactly that.
So actually my additional question is which easymesh message is responsible to notify the channel_preferences (from acs report)  to  the controller?

> 
> If it is the channel that it's currently operating at: after a channel selection request, the agent is supposed to send an Operating Channel Report CMDU to the controller. It should also do that when it changes channels on its own initiative.

that is when the radio configured with **auto channel** 
> 
> It would actually make sense to send an unsollicited Operating Channel Report CMDU immediately after autoconfig search completes (i.e. just before sending M1). Otherwise, indeed, there's no way for the controller to know the current operating channel.
>
 good point we need to verify that happens .

arnout:
> 
> In this case the ACS report will have only one channel (that's the "channel_preferences" we decided to rename earlier).
> 

Oh, that will have to be checked... The Channel Preference TLV mentions only channels that are not preferred, with a preference value ranging from 0 (= this channel will not be used) to 14 (= just below maximum preference). Channels that were reported to be supported in the AP Radio Basic Capabilities but that are not mentioned in the Channel Preference TLV are considered to have maximum preference.

So, to be verified how the construction of the Channel Preference TLV treats an ACS report that has only a single channel. I suspect that it will create a Channel Preference TLV with all channels except the current one set to 0 (blocked). But that is not what we want - we probably should set the preference of the current channel to maximum, and the preference of all other channels to just below that (14, or maybe something like 12). Or at the very least, we should report no preference at all (so everything at maximum preference).

Ideally there should be a test of this :-)

> 
> I want to make sure where exactly do we need to use channel_preferences (from acs) and where to use supported_channels (from radio capabilities /hw features) since this PR is going to change exactly that.
> 

That's easy: where we use read_acs_report, it should be channel_preferences, and where we use read_supported_channels, it should be supported_channels.

> 
> So actually my additional question is which easymesh message is responsible to notify the > channel_preferences (from acs report) to the controller?
> 

Very difficult to find this: Channel Preference Report CMDU containing the Channel Preference TLV. And the supported_channels are sent in the AP Radio Basic Capabilities.

It would actually make sense to send an unsollicited Operating Channel Report CMDU immediately after autoconfig search completes (i.e. just before sending M1). Otherwise, indeed, there's no way for the controller to know the current operating channel.

> 
> good point we need to verify that happens .
> 

I can tell you: it doesn't happen at the moment.

As I mentioned earlier: the controller has no way to query the current channel, but now I think about it more, there is actually a trick that it can use. It can send a Channel Selection Request with no TLVs. The agent is supposed to send an Operating Channel Report message after receiving it, even if it doesn't switch channels. But of course, this is not tested in certification, so who knows how other agents will behave...

kantera800:
> 
> But that is not what we want - we probably should set the preference of the current channel to maximum, and the preference of all other channels to just below that (14, or maybe something like 12). Or at the very least, we should report no preference at all (so everything at maximum preference).
> 

When I disable auto channel in my router I want him stay on the channel I selected ( bridging for example). 
AFAIU what you propose contradicts radio's "static channel" definition since now on the next BG scan DCS task will be able to switch to other channels. 

> 
> That's easy: where we use read_acs_report, it should be channel_preferences, and where we use read_supported_channels, it should be supported_channels.
> 

:) 
I was refering controller side. 
Currently supported channel (channel preference) used everywhere. 

p.s
i accidentally edited and reverted your last reply

arnout:
> > But that is not what we want - we probably should set the preference of the current channel to maximum, and the preference of all other channels to just below that (14, or maybe something like 12). Or at the very least, we should report no preference at all (so everything at maximum preference).
> 
> When I disable auto channel in my router I want him stay on the channel I selected ( bridging for example).
> AFAIU what you propose contradicts radio's "static channel" definition since now on the next BG scan DCS task will be able to switch to other channels.

Right, so we really need 3 cases:
1. ACS -> channel_preference is set according to ACS scan.
2. No channel preference -> channel preference is 15 for current channel and 14 for everything else (to avoid switching channels if there's no reason to do so).
3. Static channel set by user -> channel preference is 15 for current channel and 0 or maybe 1 for all other channels.


> > That's easy: where we use read_acs_report, it should be channel_preferences, and where we use read_supported_channels, it should be supported_channels.
> 
> :)
> I was refering controller side.
> Currently supported channel (channel preference) used everywhere.

Controller side I don't think channel preferences are really used anywhere, except in the channel_selection task which has to be revamped anyway. But I haven't checked.

tomereli:
Please note that the supported channels are available via NL80211 as implemented in https://github.com/prplfoundation/prplMesh/pull/977

kantera800:
> Please note that the supported channels are available via NL80211as implemented in #977

we already use get_hw_features from hostapd.

But you reminded me of something else,
 since we are opening multiple sockets to nl (DWPAL, NL80211, arp monitor etc...) we are going to have a problem if we don't protect nl reply buffer from collision.
I already see it when scanning simultaneously on both radio's .   
not related to this PR.

PR #947, opened 2020-03-11
Bugfix/fix channel selection

### LINKED ISSUE #937: [TASK] Pass test MAP-4.8.1_ETH_FH5GH in testbed

tomereli:
agent: channel_sel: restrict correct channels for op class 128,129,130 …
03ba771

Commit 23de848 tried to fix a bug in the channel selection where
operating classes 128,129,130 which use a center channel instead of
start channel, but it broke in the last code review iteration since the
center channel was used correctly to get the operating class, but the
start channel was wrongly used to check if the channel is restricted
later on.

Fix that by explicitely adding center channel to the
get_operating_class_by_channel API, then explicitely use the center
channel for operating classes 128,129 and 130.

Tests run: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/467558322

Fixes #937 

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
             continue;
         }
         for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
-            auto channel         = hostap_params.supported_channels[i];
-            auto operating_class = wireless_utils::get_operating_class_by_channel(
-                channel.channel, static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth));
+            auto channel        = hostap_params.supported_channels[i];
+            auto bw             = static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth);
+            auto center_channel = wireless_utils::get_5g_center_channel(channel.channel, bw, true);
+            auto operating_class =
+                wireless_utils::get_operating_class_by_channel(channel.channel, center_channel, bw);
 
             // Skip channels from other operating classes
             if (operating_class != preference.oper_class) {
                 continue;
             }
             // Skip DFS channels
-            if (channel.is_dfs_channel) {
+            if (hostap_params.supported_channels[i].is_dfs_channel) {
                 LOG(DEBUG) << "Skip DFS channel " << int(channel.channel);
                 continue;
             }
             auto is_restricted =
                 std::find_if(preference.channels.begin(), preference.channels.end(),
                              [&](const beerocks::message::sWifiChannel &wifi_channel) {
-                                 return (channel.channel == wifi_channel.channel);
+                                 // operating classes 128,129,130 use center channel **unlike the other classes** (See Table E-4 in 802.11 spec)
+                                 if (operating_class == 128 || operating_class == 129 ||
+                                     operating_class == 130) {
+                                     return center_channel == wifi_channel.channel;
+                                 } else {
+                                     return (channel.channel == wifi_channel.channel);
+                                 }
                              });
             // Skip restricted channels
             if (is_restricted != preference.channels.end()) {
-                LOG(DEBUG) << "Skip restricted channel " << int(channel.channel);
+                LOG(DEBUG) << "Skip restricted channel " << int(channel.channel)
+                           << " (center channel " << int(center_channel) << ")";
                 continue;
             }
             // If we got this far, we found a candidate channel, so switch to it
             LOG(DEBUG) << "Selected channel " << int(channel.channel);
-            return channel;
+            return hostap_params.supported_channels[i];
         }
     }
 
@@ -4475,12 +4484,15 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
 
 bool slave_thread::channel_selection_current_channel_restricted()
 {
-    auto channel         = hostap_cs_params.channel;
-    auto bw              = static_cast<beerocks::eWiFiBandwidth>(hostap_cs_params.bandwidth);
-    auto operating_class = wireless_utils::get_operating_class_by_channel(channel, bw);
+    auto bw      = static_cast<beerocks::eWiFiBandwidth>(hostap_cs_params.bandwidth);
+    auto channel = hostap_cs_params.channel;
+    auto center_channel =
+        beerocks::utils::wifi_freq_to_channel(hostap_cs_params.vht_center_frequency);
+    auto operating_class =
+        wireless_utils::get_operating_class_by_channel(channel, center_channel, bw);
 
-    LOG(DEBUG) << "Current channel " << int(channel) << " bw " << int(bw) << " oper_class "
-               << int(operating_class);
+    LOG(DEBUG) << "Current channel " << int(channel) << " center channel " << int(center_channel)
+               << " bw " << int(bw) << " oper_class " << int(operating_class);
     for (const auto &preference : channel_preferences) {
         // for now we handle only non-operable preference
         // TODO - handle as part of https://github.com/prplfoundation/prplMesh/issues/725
@@ -4496,19 +4508,24 @@ bool slave_thread::channel_selection_current_channel_restricted()
                       << "channel switch required";
             return true;
         }
-        auto is_restricted =
-            std::find_if(preference.channels.begin(), preference.channels.end(),
-                         [&channel](const beerocks::message::sWifiChannel &wifi_channel) {
-                             return channel == wifi_channel.channel;
-                         });
+        auto is_restricted = std::find_if(
+            preference.channels.begin(), preference.channels.end(),
+            [&](const beerocks::message::sWifiChannel &wifi_channel) {
+                // operating classes 128,129,130 use center channel **unlike the other classes** (See Table E-4 in 802.11 spec)
+                if (operating_class == 128 || operating_class == 129 || operating_class == 130) {
+                    return center_channel == wifi_channel.channel;
+                } else {
+                    return channel == wifi_channel.channel;
+                }
+            });
 
         if (is_restricted != preference.channels.end()) {
-            LOG(INFO) << "Current channel " << int(channel)
-                      << " is restricted, channel switch required";
+            LOG(INFO) << "Current channel " << int(channel) << " center channel"
+                      << int(center_channel) << " is restricted, channel switch required";
             return true;
         }
     }
-    LOG(INFO) << "Current channel " << int(channel)
+    LOG(INFO) << "Current channel " << int(channel) << " center channel" << int(center_channel)
               << " not restricted, channel switch not required";
     return false;
 }
@@ -4734,11 +4751,20 @@ bool slave_thread::send_operating_channel_report()
         return false;
     }
 
-    auto &operating_class_entry           = std::get<1>(operating_class_entry_tuple);
-    operating_class_entry.operating_class = wireless_utils::get_operating_class_by_channel(
-        hostap_cs_params.channel, (beerocks::eWiFiBandwidth)hostap_cs_params.bandwidth);
-    operating_class_entry.channel_number = hostap_cs_params.channel;
+    auto &operating_class_entry = std::get<1>(operating_class_entry_tuple);
+    auto bw                     = static_cast<beerocks::eWiFiBandwidth>(hostap_cs_params.bandwidth);
+    auto channel                = hostap_cs_params.channel;
+    auto center_channel =
+        beerocks::utils::wifi_freq_to_channel(hostap_cs_params.vht_center_frequency);
+    auto operating_class =
+        wireless_utils::get_operating_class_by_channel(channel, center_channel, bw);
 
+    operating_class_entry.operating_class = operating_class;
+    // operating classes 128,129,130 use center channel **unlike the other classes** (See Table E-4 in 802.11 spec)
+    operating_class_entry.channel_number =
+        (operating_class == 128 || operating_class == 129 || operating_class == 130)
+            ? center_channel
+            : channel;
     operating_channel_report_tlv->current_transmit_power() = hostap_cs_params.tx_power;
 
     return send_cmdu_to_controller(cmdu_tx);
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4432,6 +4432,50 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
     return message_com::send_cmdu(ap_manager_socket, cmdu_tx);
 }
 
+/**
+ * @brief Get the channel preference 
+ * 
+ * @param channel channel to check
+ * @param preference preference 
+ * @return NON_OPERABLE if channel is restricted, channel preference otherwise
+ */
+static uint8_t get_channel_preference(const beerocks::message::sWifiChannel channel,
+                                      const son::wireless_utils::sChannelPreference &preference)
+{
+    uint8_t center_channel = 0;
+    auto bw                = static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth);
+    auto operating_class   = wireless_utils::get_operating_class_by_channel(channel);
+
+    // consider channels in different operating classes as restricted
+    if (operating_class != preference.oper_class) {
+        return preference.preference;
+    }
+    // operating classes 128,129,130 use center channel **unlike the other classes**,
+    // so convert channel and bandwidth to center channel.
+    // For more info, refer to Table E-4 in the 802.11 specification.
+    if (operating_class == 128 || operating_class == 129 || operating_class == 130) {
+        center_channel = wireless_utils::get_5g_center_channel(channel.channel, bw, true);
+    }
+
+    // According to Table 23 in the MultiAP Specification, an empty channel list field
+    // indicates that the indicated preference applies to all channels in the operating class.
+    if (preference.channels.empty()) {
+        return preference.preference;
+    }
+    // explicitely restrict non-operable channels
+    auto channel_to_check =
+        (operating_class == 128 || operating_class == 129 || operating_class == 130)
+            ? center_channel
+            : channel.channel;
+    for (const auto &ch : preference.channels) {
+        if (channel_to_check == ch.channel) {
+            return preference.preference;
+        }
+    }
+    // default to the highest preference
+    return wfa_map::cPreferenceOperatingClasses::ePreference::PREFERRED14;
+}
+
 beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
 {
     for (const auto &preference : channel_preferences) {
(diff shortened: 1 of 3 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
             continue;
         }
         for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
-            auto channel         = hostap_params.supported_channels[i];
-            auto operating_class = wireless_utils::get_operating_class_by_channel(
-                channel.channel, static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth));
+            auto channel        = hostap_params.supported_channels[i];
+            auto bw             = static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth);
+            auto center_channel = wireless_utils::get_5g_center_channel(channel.channel, bw, true);
```
vitalybu: Doesn't `hostap_params.supported_channels` also contain 2.4G channels?
On line 4445 you're calling `wireless_utils::get_5g_center_channel()` which will work fine with 2.4G channels as well, but if it's used not only for 5G channels, maybe the name should be changed to `get_center_channel`.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
             continue;
         }
         for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
-            auto channel         = hostap_params.supported_channels[i];
```
vitalybu, line 4443: Why not store a reference to the channel if it's used in more than one place?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
                 continue;
             }
             auto is_restricted =
                 std::find_if(preference.channels.begin(), preference.channels.end(),
                              [&](const beerocks::message::sWifiChannel &wifi_channel) {
-                                 return (channel.channel == wifi_channel.channel);
+                                 // operating classes 128,129,130 use center channel **unlike the other classes** (See Table E-4 in 802.11 spec)
+                                 if (operating_class == 128 || operating_class == 129 ||
+                                     operating_class == 130) {
+                                     return center_channel == wifi_channel.channel;
```
vitalybu:
```suggestion
                                     return (center_channel == wifi_channel.channel);
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4496,19 +4508,24 @@ bool slave_thread::channel_selection_current_channel_restricted()
...
-                             return channel == wifi_channel.channel;
-                         });
+        auto is_restricted = std::find_if(
+            preference.channels.begin(), preference.channels.end(),
+            [&](const beerocks::message::sWifiChannel &wifi_channel) {
+                // operating classes 128,129,130 use center channel **unlike the other classes** (See Table E-4 in 802.11 spec)
+                if (operating_class == 128 || operating_class == 129 || operating_class == 130) {
+                    return center_channel == wifi_channel.channel;
+                } else {
+                    return channel == wifi_channel.channel;
```
vitalybu:
```suggestion
                    return (center_channel == wifi_channel.channel);
                } else {
                    return (channel == wifi_channel.channel);
```

vitalybu (approved):
Have some comments, but nothing critical, so approving.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: The reason I called it `get_5g_center_channel` is because there is no use for center channel for 2.4G band. I can change the name to `get_center_channel` but then it might be confusing and cause people to use it when they shouldn't..

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
             auto is_restricted =
                 std::find_if(preference.channels.begin(), preference.channels.end(),
                              [&](const beerocks::message::sWifiChannel &wifi_channel) {
-                                 return (channel.channel == wifi_channel.channel);
+                                 // operating classes 128,129,130 use center channel **unlike the other classes** (See Table E-4 in 802.11 spec)
+                                 if (operating_class == 128 || operating_class == 129 ||
+                                     operating_class == 130) {
+                                     return center_channel == wifi_channel.channel;
+                                 } else {
+                                     return (channel.channel == wifi_channel.channel);
```
kantera800: a bit confusing , consider changing to something like wifi_chan_params.channel (nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: actually now that you mention it, `const auto&` is indeed more appropriate. Just a reference without const is risky.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
-            if (channel.is_dfs_channel) {
+            if (hostap_params.supported_channels[i].is_dfs_channel) {
                 LOG(DEBUG) << "Skip DFS channel " << int(channel.channel);
                 continue;
             }
             auto is_restricted =
                 std::find_if(preference.channels.begin(), preference.channels.end(),
                              [&](const beerocks::message::sWifiChannel &wifi_channel) {
-                                 return (channel.channel == wifi_channel.channel);
+                                 // operating classes 128,129,130 use center channel **unlike the other classes** (See Table E-4 in 802.11 spec)
```
kantera800: 802.11 spec say its VS ,
 so if this is our decision why not use first channel as well as the other classes?

edit: I checked the spec from 2012 (not sure what is the latest one)  , and it say **reserved** not VS.
but still same question why center channel?

tomereli: Don't know what you mean by "spec say its VS". When we get channel preference TLV from the controller, it includes all the non-operable channels for each operating class.
For operating classes 128,129,130 the 802.11 specification says that the center channel should be used instead of the start channel, which is also reflected in the channel set that the UCC sends in the channel selection request:
```
DEBUG 22:54:00:028 <1988852512> son_slave_thread.cpp[4566] --> operating class=128, preference=0, reason=0, channel_list={42,58,106,122,138}
DEBUG 22:54:00:028 <1988852512> son_slave_thread.cpp[4566] --> operating class=129, preference=0, reason=0, channel_list={}
DEBUG 22:54:00:029 <1988852512> son_slave_thread.cpp[4566] --> operating class=130, preference=0, reason=0, channel_list={}
```

Since operating class 128 bandwidth is 80MHz, channel 42 is the center channel while the start channel (which is what is expected/reported by hostapd) is 36.
So in this code snippet, when we compare the channel to the non-operable channels, we should compare the center channel instead of the start channel when the operating class is 128-130.

I hope this is more clear.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: `channel.channel` is fine IMO and it was that before this PR, so NACK.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4496,19 +4508,24 @@ bool slave_thread::channel_selection_current_channel_restricted()
...
             return true;
         }
-        auto is_restricted =
-            std::find_if(preference.channels.begin(), preference.channels.end(),
-                         [&channel](const beerocks::message::sWifiChannel &wifi_channel) {
-                             return channel == wifi_channel.channel;
-                         });
+        auto is_restricted = std::find_if(
+            preference.channels.begin(), preference.channels.end(),
+            [&](const beerocks::message::sWifiChannel &wifi_channel) {
```
rmelotte: This lambda looks a lot like the other one. Are-they the same? If yes I would take the opportunity to make it a function

rmelotte (approved):
Approving already because it fixes the issue we have with any 5GH test.
Added the don't merge tag so you can decide if you want it merged or not.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -260,7 +260,8 @@ bool slave_thread::work()
 
 void slave_thread::process_keep_alive()
 {
-    if (!config.enable_keep_alive || !son_config.slave_keep_alive_retries) {
+    if (!config.enable_keep_alive || !son_config.slave_keep_alive_retries ||
+        !backhaul_params.is_prplmesh_controller) {
```
arnout: I think @morantr planned to remove this entirely, but until then, let's do it like this indeed.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Since we have the full channel structure here, why not pass that to the function instead of the separate bandwidth?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
         }
         for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
-            auto channel         = hostap_params.supported_channels[i];
-            auto operating_class = wireless_utils::get_operating_class_by_channel(
-                channel.channel, static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth));
+            auto channel        = hostap_params.supported_channels[i];
+            auto bw             = static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth);
+            auto center_channel = wireless_utils::get_5g_center_channel(channel.channel, bw, true);
+            auto operating_class =
+                wireless_utils::get_operating_class_by_channel(channel.channel, center_channel, bw);
```
arnout: This makes little sense to me... The conversion of channel+bw to operating class should be one-to-one IMO. So *either* you pass just channel.channel and do the get_center_channel inside the function, *or* the function assumes that the center_channel is used for the 80MHz channels.

arnout (commented):
Some quick comments already.

agent/src/beerocks/slave/son_slave_thread.cpp
kantera800: didn't see it in  802.11 2012 spec

image

so in that case we can decide what we send in the preference TLV for operative classes 128-130


edit: found it in 802.11 2016 
now it makes sense :)

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: @vitalybu can you update CODING_STYLE.md with this pattern? We should record the patterns that we often use.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Forget this.

What we really should have is something like a wireless_operating_channel class which has constructors based on (operatingclass, channel) and on (channel, bandwidth). This class would be used everywhere internally, and it would have member functions to convert internally to whatever format is needed.

That was actually my plan all along when operating classes were introduced.

However, changing the code like that is *a lot* of work, with limited benefit. So let's just forget about it.

So for this PR, I'm OK with the proposal here. get_operating_class_by_channel *always* needs to be combined with the get_center_frequency because that's what is needed for the "operating class" form.

Or maybe one improvement would be to introduce just a single additional function in wireless_utils: `get_operating_class_channel(channel, bandwidth)`. So when converting to the "operating class" format, you derive both the operating class itself and the "operating class channel" i.e. the channel as used in the "operating class" format. Does that make sense?

tomereli: Why forget this? `get_operating_class_by_channel` should accpet a single parameter, so swifiChannel makes perfect sense, this until we have the wireless_operating_channel class which I like even more but definitely not for this PR.

Anyway, see if 6b4dd98 is better.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
 
             // Skip channels from other operating classes
             if (operating_class != preference.oper_class) {
                 continue;
             }
             // Skip DFS channels
-            if (channel.is_dfs_channel) {
+            if (hostap_params.supported_channels[i].is_dfs_channel) {
```
arnout: I must have missed something... Why this change?

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: @tomereli You marked this as resolved but didn't actually make the change.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,32 +4440,41 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
             if (is_restricted != preference.channels.end()) {
-                LOG(DEBUG) << "Skip restricted channel " << int(channel.channel);
+                LOG(DEBUG) << "Skip restricted channel " << int(channel.channel)
+                           << " (center channel " << int(center_channel) << ")";
                 continue;
             }
             // If we got this far, we found a candidate channel, so switch to it
             LOG(DEBUG) << "Selected channel " << int(channel.channel);
-            return channel;
+            return hostap_params.supported_channels[i];
```
arnout: Again, why this change?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4432,6 +4432,52 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
+    // so convert channel and bandwidth to center channel.
+    // For more info, refer to Table E-4 in the 802.11 specification.
+    if (operating_class == 128 || operating_class == 129 || operating_class == 130) {
+        center_channel = wireless_utils::get_5g_center_channel(channel.channel, bw, true);
+    }
+
+    // restrict channels from other operating classes
+    if (operating_class != preference.oper_class) {
+        return true;
+    }
```
arnout: This comment makes little sense to me. I think you mean "consider channels in different operating classes as restricted".

Also I'd do this check more in the beginning.

nitpick both

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4432,6 +4432,52 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
+    // According to Table 23 in the MultiAP Specification, an empty channel list field
+    // indicates that the indicated preference applies to all channels in the operating class.
+    if (preference.channels.empty()) {
+        return true;
+    }
+    // explicitely restrict non-operable channels
+    for (const auto &ch : preference.channels) {
+        if (operating_class == 128 || operating_class == 129 || operating_class == 130) {
```
arnout: (nitpick) I'd move this condition outside of the loop, i.e. calculate the channel to check only once.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4432,6 +4432,52 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
 
+/**
+ * @brief check if channel is restricted in a given preference
+ * 
+ * @param channel channel to check
+ * @param preference channel preference
+ * @return true if channel is restricted
+ * @return false if channel is not restricted
+ */
+static bool is_channel_restricted(const beerocks::message::sWifiChannel channel,
```
arnout: (nitpick) it would make more sense to return the preference value instead, or NON_OPERABLE.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4440,36 +4486,15 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
...
-                                     return center_channel == wifi_channel.channel;
-                                 } else {
-                                     return (channel.channel == wifi_channel.channel);
-                                 }
-                             });
             // Skip restricted channels
-            if (is_restricted != preference.channels.end()) {
-                LOG(DEBUG) << "Skip restricted channel " << int(channel.channel)
-                           << " (center channel " << int(center_channel) << ")";
+            if (is_channel_restricted(channel, preference)) {
```
arnout: So when returning the preference value, you'd check here if it was equal to NON_OPERABLE. And it's trivial to extend it later to multiple preference values.

agent/src/beerocks/slave/son_slave_thread.cpp
vitalybu: It'll work just fine for 2.4G channels (simply doing nothing), but since we're using it for all channels (not just 2.4G), I think that `get_center_channel` is a more appropriate name.

arnout (approved):
Nitpick only. I'd just like to undo the two unnecessary changes.

arnout:
@kantera800 @adam1985d Since DCS (and maybe other parts of channel selection) is using a mix of EasyMesh messages and VS messages, and since the VS messages *don't* use the center freq for the 80MHz operating classes, I expect there will be additional changes needed either in the controller or in the DCS messages to make sure things don't break again.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4433,49 +4433,48 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
 
-    // restrict channels from other operating classes
-    if (operating_class != preference.oper_class) {
-        return true;
-    }
     // According to Table 23 in the MultiAP Specification, an empty channel list field
     // indicates that the indicated preference applies to all channels in the operating class.
     if (preference.channels.empty()) {
-        return true;
+        return wfa_map::cPreferenceOperatingClasses::ePreference::NON_OPERABLE;
```
arnout: Err, no, the idea was to return preference.preference here.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4433,49 +4433,48 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
-        if (operating_class == 128 || operating_class == 129 || operating_class == 130) {
-            if (center_channel == ch.channel) {
-                return true;
-            }
-        } else {
-            if (channel.channel == ch.channel) {
-                return true;
-            }
+        if (channel_to_check == ch.channel) {
+            wfa_map::cPreferenceOperatingClasses::ePreference::NON_OPERABLE;
```
arnout: The `return` seems to have gone missing here :-)

But again, I meant `return preference.preference;`

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4433,49 +4433,48 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
...
         }
     }
-    return false;
+    // default to the highest preference
+    return wfa_map::cPreferenceOperatingClasses::ePreference::PREFERRED14;
```
arnout, line 4476: This should actually be PREFERRED15 - but that is apparently not defined...

arnout (approved):
Not entirely what I meant :-) but close!

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: This is according to spec:
```
enum ePreference {
            NON_OPERABLE = 0x0,
            PREFERRED1 = 0x1,
            PREFERRED2 = 0x2,
            PREFERRED3 = 0x3,
            PREFERRED4 = 0x4,
            PREFERRED5 = 0x5,
            PREFERRED6 = 0x6,
            PREFERRED7 = 0x7,
            PREFERRED8 = 0x8,
            PREFERRED9 = 0x9,
            PREFERRED10 = 0xa,
            PREFERRED11 = 0xb,
            PREFERRED12 = 0xc,
            PREFERRED13 = 0xd,
            PREFERRED14 = 0xe,
            RESERVED = 0xf,
        };
```

tomereli: image

kantera800:
> @kantera800 @adam1985d Since DCS (and maybe other parts of channel selection) is using a mix of EasyMesh messages and VS messages, and since the VS messages _don't_ use the center freq for the 80MHz operating classes, I expect there will be additional changes needed either in the controller or in the DCS messages to make sure things don't break again.

@arnout @tomereli 
DCS currently  use VS messages only.
This change needs to be done in channel selection task. 
Just a thought, do we need to keep channel selection task ? 
1. currently its running but it does nothing 
2.  it needs serious refactoring unless we won't pass certification.
3. most of its functionality can be done in DCS task.

I'm bringing this up since we are currently deciding on the next DCS phases..

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Exactly: if it is not explicitly specified, it should be a larger preference value than 14. So it should default to 15 (or any larger number). The spec literally says "the most preferred score 15 is inferred" (i.e. not part of the message). Internally, we still have to use that value.

arnout:
> > @kantera800 @adam1985d Since DCS (and maybe other parts of channel selection) is using a mix of EasyMesh messages and VS messages, and since the VS messages _don't_ use the center freq for the 80MHz operating classes, I expect there will be additional changes needed either in the controller or in the DCS messages to make sure things don't break again.
> 
> @arnout @tomereli
> DCS currently use VS messages only.
> This change needs to be done in channel selection task.
> Just a thought, do we need to keep channel selection task ?

Discussed orally so just for the record: let's keep channel_selection_task.cpp but remove all logic from it and restart from scratch with the DCS approach.

PR #949, opened 2020-03-11
framework: check the destination mac when looking for duplicates

### LINKED ISSUE #945: [BUG] test_flows.sh: test_client_steering_dummy occasionnaly fail with missing ACK
Example of a job where it fails here: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/465223282

The only thing I noticed up to now is that:
- the target agent indeed doesn't receive any ACK message
- the controller tries to send the ACK message according to the logs

I created the branch `dev/test_client_steering_dummy` to debug the issue.

rmelotte:
When an ACK message is sent to different destination MACs
successively, the packet might be marked as duplicate and not sent,
resulting in some targets not receiving the packet.

This was discovered as part of issue #945:
https://github.com/prplfoundation/prplMesh/issues/945

Add the destination to the keys that are compared when looking for
duplicates, so that the same packets with a different destination are
not marked as duplicates.

While we're at it, replace 3 C-style casts with static casts.

Fixes #945.

### DIFF
--- a/framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
+++ b/framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
@@ -152,6 +152,7 @@ private:
 
     struct DeDuplicationKey {
         uint8_t src[ETH_ALEN];
+        uint8_t dst[ETH_ALEN];
         uint16_t
             messageType; // required to distinguish the case of Request / Reply (when the reply carries an out of sync MID)
         uint16_t messageId;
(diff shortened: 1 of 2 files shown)


tomereli (approved):

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
```diff
@@ -161,10 +162,13 @@ class Ieee1905Transport {
...
         {
-            // compare based on src address first then messageId, then fragmentId
+            // compare based on src and dst address first then messageId, then fragmentId
             int srccmp = memcmp(lhs.src, rhs.src, ETH_ALEN);
             if (srccmp)
                 return srccmp < 0;
+            int dstcmp = memcmp(lhs.dst, rhs.dst, ETH_ALEN);
+            if (dstcmp)
+                return dstcmp < 0;
             else if (lhs.messageType != rhs.messageType)
```
adam1985d, line 172: (nitpick)
no need for the `else if` since `return` is performed if the previous condition is met

adam1985d (approved):

PR #953, opened 2020-03-12
flow 4.7.7 -- intermediate results

vitalii-komisarenko:


arnout:
Based on the test plan and my own common sense, I don't think we should create link metrics per connected STA. Indeed, those types of metrics are already covered by the Associated STA Traffic Stats. What is important to report, and indeed the only thing that is verified by the certification test plan 4.7.3, are the link metrics for the backhaul link. 

This PR, on the other hand, is supposed to be for 4.7.7, which is about Associated STA Link Metrics. Those are completely different from the 1905.1 Link Metrics. Associated STA Link Metrics just have:
- Estimated MAC Data Rate in downlink
- Estimated MAC Data Rate in uplink
- Uplink RCPI

Note that the MAC Data Rate is *not* a measure of how much throughput is generated, but rather it's a measure of the link speed assuming that the STA would be able to use 100% of the airtime. I *think* that's just the rx/tx_phy_rate as reported in struct SStaStats, but I'm not sure - maybe overhead needs to be subtracted. Any wifi expert who can confirm?

Make sure you carefully read section 10.3.1, it has a lot to say about how the measurements should be done. We may need to adapt mon_wlan_hal to cater to it.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1632,6 +1646,40 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
...
+        if (!response->alloc_per_bss_associated_sta_link_metrics(response->bss_num())) {
+            LOG(ERROR) << "Failed alloc_per_bss_associated_sta_link_metrics";
+            return false;
+        }
+
+        for (size_t i=0; i<response->bss_num(); ++i) {
+            auto &resp = std::get<1>(response->per_bss_associated_sta_link_metrics(i));
+            auto &metr = metrics_result->per_bss_metrics[i];
+
+            memcpy(resp.bssid.oct, metr.bssid.oct, sizeof(resp.bssid.oct));
```
vitalii-komisarenko, line 1671: TODO: convert mac addresses properly (including extra fields)

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1632,6 +1646,42 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
...
+        auto response = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE>(cmdu_tx);
+
+        if (!response) {
+            LOG(ERROR) <<
+                "Failed building cACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE msg";
+            return false;
+        }
+
+        memcpy(response->sta_mac().oct, metrics_result->sta_mac.oct,
```
arnout, line 1659: Don't use memcpy. IIRC we have an operator= for sMacAddr so you can just assign it; otherwise, use std::copy.

vitalii-komisarenko: BWL has a separate struct for the MAC address. `=` is not working because types are technically different.

vitalii-komisarenko:
Replaced by #1149

PR #959, opened 2020-03-12
Feature/cppcheck framework master

rmelotte:
Fixes the cppcheck issues of severity warning and error for the framework folder, then enforces cppcheck on the framework folder.


framework/discovery_agent/include/mapf/topology_discovery_agent/topology_discovery_agent_common.h
```diff
@@ -19,7 +19,7 @@ enum topology_discovery_agent_common_macros {
     MAX_NAME_LEN        = 64,
     MAX_IPv4_ADDR_LEN   = 4,
     MAX_IPv6_ADDR_LEN   = 16,
-    MAX_SSID_LEN        = 6,
```
arnout, line 22: Or maybe better: `rm -rf framework/discovery_agent`...

framework/transport/ieee1905_transport/test/ieee1905_transport_test.cpp
```diff
@@ -336,8 +336,8 @@ int main(int argc, char *argv[])
             std::copy_n(ieee1905_header, sizeof(ieee1905_header), cmdu_tx_msg.data());
 
             MAPF_INFO("sending CmduTxMessage: " << std::endl << cmdu_tx_msg);
-            rc = bus.publisher().Send(cmdu_tx_msg);
-            if (rc != true) {
+            bool ret = bus.publisher().Send(cmdu_tx_msg);
+            if (!ret) {
```
arnout, line 340: I personally have a slight preference for

```suggestion
            if (!bus.publisher().Send(cmdu_tx_msg)) {
```

but we use both patterns in the code at the moment, and this is what it was originally, so OK. Maybe we should choose and track it in CODING_STYLE.md though.

framework/common/nng/broker.cpp
```diff
@@ -18,7 +18,8 @@ namespace mapf {
 
 Broker::Broker() : frontend_(new msglib_socket), backend_(new msglib_socket)
 {
-    int rc = nng_pub0_open_raw(&backend_->sd_);
+    context_ = nullptr; // TODO: implement
```
arnout: IMO it's more appropriate to do this in the initialiser list than in the function body.

The TODO is wrong - there simply is no context needed in the nng case. This stuff is implemented in a bit an awkward way: there are two implementations, but they use the same header file. For that reason the context_ is a void pointer instead of something useful.

framework/platform/bpl/uci/arp/monitor/arp_monitor.h
```diff
@@ -88,7 +88,7 @@ class arp_monitor {
     std::list<std::shared_ptr<SProbeEntry>> m_lstProbe;
 
     // Buffer for reading ARP packets
-    char m_arrArpPacket[ETH_FRAME_LEN];
+    char m_arrArpPacket[ETH_FRAME_LEN]{0};
```
arnout, line 91: I didn't know this syntax, I thought it had to be 

```suggestion
    char m_arrArpPacket[ETH_FRAME_LEN] = {0};
```

#til

arnout (approved):
Nice!

framework/discovery_agent/include/mapf/topology_discovery_agent/topology_discovery_agent_common.h
tomereli: Yes please!

framework/transport/ieee1905_transport/ieee1905_transport_network.cpp
```diff
@@ -32,6 +32,10 @@ class Ieee1905SocketFilter {
     // The two addresses are here to specify the AL MAC address and the interface's hardware address
     Ieee1905SocketFilter(const uint8_t *addr0 = NULL, const uint8_t *addr1 = NULL)
     {
+        if (!addr0 && !addr1) {
```
tomereli: Can we define `MAPF_LOG_IF` or just use `LOG_IF` here?

tomereli (approved):

tomereli:
Added the don't merge tag in case you want to fix the remaining comments.

morantr (approved):

PR #964, opened 2020-03-12
agent: channel_sel: ignore channels from other operating classes

### LINKED ISSUE #960: [BUG] agent: MAP-4.2.1 fails with "No Operating Channel Report message found"
Job: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/468547961

logs: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/468547961/artifacts/browse/logs/FAIL/MAP-4.2.1/

tomereli:
There is a bug in the current channel selection implementation which was
introduced in the recent changes to channel selection.
The issue is that if the agent receives a channel selection request
with the same preference as the channel it is currently operating on,
the agent mistakingly identifies that the channel is non-operable and
therefore does not send operating channel report as it expects channel
selection to complete with CSA_Finished event.
The CSA_Finished event never arrives since no channel switch was
actually needed.

The root cause is in the get_channel_preference function when it is
called from channel_selection_current_channel_restricted() - it is
called for each channel preference before concluding that channel switch
is not required, so while the channel in question may belong to the
first preference operating class, we still continue and check the second
preference for which the channel will definitely not belong and
therefore hit be marked as non-operable.
Reproduced 100%.

The solution is to always skip channels which do not belong to the
preference operating class in channel_selection_select_channel() and in
channel_selection_current_channel_restricted(), and abort if we get non
matching channel in get_channel_preference().

Fixes #960 

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4434,7 +4434,8 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
 
 /**
  * @brief Get the channel preference 
- * 
+ *
+ * @pre The channel operating class and the preference operating class have to match.
  * @param channel channel to check
  * @param preference preference 
  * @return NON_OPERABLE if channel is restricted, channel preference otherwise
@@ -4446,10 +4447,10 @@ static uint8_t get_channel_preference(const beerocks::message::sWifiChannel chan
     auto bw                = static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth);
     auto operating_class   = wireless_utils::get_operating_class_by_channel(channel);
 
-    // consider channels in different operating classes as restricted
-    if (operating_class != preference.oper_class) {
-        return preference.preference;
-    }
+    LOG_IF(operating_class != preference.oper_class, FATAL)
+        << "Invalid channel operating class " << int(operating_class)
+        << ", preference operating class is " << int(preference.oper_class);
+
     // operating classes 128,129,130 use center channel **unlike the other classes**,
     // so convert channel and bandwidth to center channel.
     // For more info, refer to Table E-4 in the 802.11 specification.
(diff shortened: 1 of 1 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4446,10 +4448,10 @@ static uint8_t get_channel_preference(const beerocks::message::sWifiChannel chan
     auto bw                = static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth);
     auto operating_class   = wireless_utils::get_operating_class_by_channel(channel);
 
-    // consider channels in different operating classes as restricted
-    if (operating_class != preference.oper_class) {
-        return preference.preference;
```
arnout, line 4451: This should have stayed at INOPERABLE instead of preference.preference. But asserting is fine as well.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4434,7 +4434,9 @@ bool slave_thread::handle_channel_preference_query(Socket *sd, ieee1905_1::CmduM
 
 /**
  * @brief Get the channel preference 
- * 
+ * Note that the channel operating class and the preference operating class has to match,
```
arnout: (nitpick) leave an empty line before, otherwise it will be part of the 'brief'.

Maybe even better, mark is as a precondition:
```
@pre The channel operating class and the preference operating class have to match.
```

arnout (approved):
I can't say I fully understand the issue, but the code changes definitely seem correct.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4446,10 +4448,10 @@ static uint8_t get_channel_preference(const beerocks::message::sWifiChannel chan
     auto bw                = static_cast<beerocks::eWiFiBandwidth>(channel.channel_bandwidth);
     auto operating_class   = wireless_utils::get_operating_class_by_channel(channel);
 
-    // consider channels in different operating classes as restricted
-    if (operating_class != preference.oper_class) {
-        return preference.preference;
-    }
+    LOG_IF(operating_class != preference.oper_class, FATAL)
```
mariomaz, line 4450: Shall this really abort execution if it does ever happen? Or should it be treated as an error and ignored?

mariomaz (commented):
I don't either fully understand the issue here but changes look good so I approve it.
This is a method for which we could have easily created a unit-test

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: This will result with `abort()`.

mariomaz (approved):
Approving

PR #965, opened 2020-03-12
test_flows.py: Enhance channel selection test

### LINKED ISSUE #926: tests:test_flow.py:test_channel_selection cover 4.2.1 Certification test
### Description
`test_flow.py:test_channel_selection` currently sends empty channel selection request. As result - `test_channel_selection` test doesn't verify agent reaction to `CHANNEL_SELECTION_REQUEST_MESSAGE` due to empty payload.
As agent already able to pass 4.2.1 on testbed, we can complete `test_channel_selection`. Thus, `test_flows.py` will cover whole MAP-4.2.1 Certification test scenario.

### Acceptance criteria
* `test_channel_selection` send full payload for `0x8006` command
* `test_channel_selection` verifies that proper channel was selected ( `Operating Channel 
Report` was sent to controller)

VladyslavTupikin:
test_flow.py:test_channel_selection currently sends empty channel
selection request. As result - test_channel_selection test doesn't
verify agent reaction to CHANNEL_SELECTION_REQUEST_MESSAGE due to
empty payload.

We add payload for channel selection message for wlan0 and wlan2
and add checks in log files for CHANNEL_PREFERENCE_QUERY_MESSAGE,
CHANNEL_SELECTION_REQUEST_MESSAGE,tlvTransmitPowerLimit,
ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -379,18 +379,88 @@ class TestFlows:
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
 
+        payload_empty = "0x14"
+
+        # payload_wlan0 - request for change channel on 6
+        payload_wlan0 = (
+            "0x14 "
+            "{0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} "
+            "{0x52 {0x00 0x00}} "
+            "{0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}} "
+            "{0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} "
+            "{0x73 {0x00 0x00}} "
+            "{0x74 {0x00 0x00}} "
+            "{0x75 {0x00 0x00}} "
+            "{0x76 {0x00 0x00}} "
+            "{0x77 {0x00 0x00}} "
+            "{0x78 {0x00 0x00}} "
+            "{0x79 {0x00 0x00}} "
+            "{0x7A {0x00 0x00}} "
+            "{0x7B {0x00 0x00}} "
+            "{0x7C {0x00 0x00}} "
+            "{0x7D {0x00 0x00}} "
+            "{0x7E {0x00 0x00}} "
+            "{0x7F {0x00 0x00}} "
+            "{0x80 {0x00 0x00}} "
+            "{0x81 {0x00 0x00}} "
+            "{0x82 {0x00 0x00}} "
+            )
+
+        # payload_wlan2  - request for change channel on 36
+        payload_wlan2 = (
+            "0x14 "
+            "{0x51 {0x00 0x00}} "
+            "{0x52 {0x00 0x00}} "
+            "{0x53 {0x00 0x00}} "
+            "{0x54 {0x00 0x00}} "
+            "{0x73 0x03 {0x28 0x2C 0x30} 0x00} "
+            "{0x74 0x01 {0x2C} 0x00} "
+            "{0x75 {0x00 0x00}} "
+            "{0x76 {0x00 0x00}} "
+            "{0x77 {0x00 0x00}} "
+            "{0x78 {0x00 0x00}} "
+            "{0x79 {0x00 0x00}} "
+            "{0x7A {0x00 0x00}} "
+            "{0x7B {0x00 0x00}} "
+            "{0x7C {0x00 0x00}} "
+            "{0x7D {0x00 0x00}} "
+            "{0x7E {0x00 0x00}} "
+            "{0x7F {0x00 0x00}} "
+            "{0x80 0x05 {0x3A 0x6A 0x7A 0x8A 0x9B} 0x00} "
+            "{0x81 {0x00 0x00}} "
+            "{0x82 {0x00 0x00}}"
+            )
+
         debug("Send channel selection request")
-        self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8006)
+        mid = self.gateway_ucc.dev_send_1905(
+                self.mac_repeater1,
+                0x8006,
+                tlv(0x8B, 0x005F, "{} {}".format(self.mac_repeater1_wlan0, payload_wlan0)),
+                tlv(0x8D, 0x0007, "{} {}".format(self.mac_repeater1_wlan0, payload_empty)),
+                tlv(0x8B, 0x004C, "{} {}".format(self.mac_repeater1_wlan2, payload_wlan2)),
+                tlv(0x8D, 0x0007, "{} {}".format(self.mac_repeater1_wlan2, payload_empty))
+            )
         time.sleep(1)
+
+        debug("Confirming channel selection request has been received on controller")
+        self.check_log(self.gateway, "controller", r"CHANNEL_SELECTION_RESPONSE_MESSAGE, mid={}".format(mid))
+
         debug("Confirming channel selection request has been received on agent")
+
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
 
-        # debug("Confirming 1905.1 Ack Message request was received on agent")
-        # TODO: When creating handler for the ACK message on the agent, replace lookup of this string
-        # TODO: currently controller sends empty channel selection request, so no switch is performed
-        # self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
-        # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
+        self.check_log(
+            self.repeater1,
+            "agent_wlan0",
+            "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+
+        self.check_log(
+            self.repeater1,
+            "agent_wlan2",
+            "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+
+        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
 
     def test_ap_capability_query(self):
         self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8001)
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -384,7 +384,7 @@ class TestFlows:
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
 
-        payload_empty = "0x14"
+        tp20dBm = 0x14
 
         # payload_wlan0 - request for change channel on 6
         payload_wlan0 = (
@@ -436,36 +436,83 @@ class TestFlows:
             "{0x82 {0x00 0x00}}"
             )
 
-        debug("Send channel selection request")
-        mid = self.gateway_ucc.dev_send_1905(
-                self.mac_repeater1,
-                0x8006,
-                tlv(0x8B, 0x005F, "{} {}".format(self.mac_repeater1_wlan0, payload_wlan0)),
-                tlv(0x8D, 0x0007, "{} {}".format(self.mac_repeater1_wlan0, payload_empty)),
-                tlv(0x8B, 0x004C, "{} {}".format(self.mac_repeater1_wlan2, payload_wlan2)),
-                tlv(0x8D, 0x0007, "{} {}".format(self.mac_repeater1_wlan2, payload_empty))
+        """
+        Step 1: Trigger channel selection to channel 6 and 36. Check that
+                operating channel report was sent.
+
+        Step 2: Trigger channel selection to channel 6 and 36 again - check that
+                operating channel report is sent again. This is to catch bugs
+                when we don't send channel report when there is no need to
+                switch channel
+        """
+        for i in range(1, 3):
+            debug("Send channel selection request, step {}".format(i))
+            mid = self.gateway_ucc.dev_send_1905(
+                    self.mac_repeater1,
+                    0x8006,
+                    tlv(0x8B, 0x005F, '{} {}'.format(self.mac_repeater1_wlan0, payload_wlan0)),
+                    tlv(0x8D, 0x0007, '{} 0x{:2x}'.format(self.mac_repeater1_wlan0, tp20dBm)),
+                    tlv(0x8B, 0x004C, '{} {}'.format(self.mac_repeater1_wlan0, payload_wlan2)),
+                    tlv(0x8D, 0x0007, '{} 0x{:2x}'.format(self.mac_repeater1_wlan2, tp20dBm))
             )
-        time.sleep(1)
+            time.sleep(1)
 
-        debug("Confirming channel selection request has been received on controller")
-        self.check_log(self.gateway, "controller", r"CHANNEL_SELECTION_RESPONSE_MESSAGE, mid={}".format(mid))
+            debug(
+                "Confirming channel selection request has been received on controller,\
+                    step {}".format(i))
+            self.check_log(
+                self.gateway,
+                "controller",
+                r"CHANNEL_SELECTION_RESPONSE_MESSAGE, mid={}".format(mid))
 
-        debug("Confirming channel selection request has been received on agent")
+            debug(
+                "Confirming channel selection request has been received on agent,step {}".format(i))
 
-        self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
-        self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+            self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+            self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
 
-        self.check_log(
-            self.repeater1,
-            "agent_wlan0",
-            "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+            debug(
+                "Confirming tlvTransmitPowerLimit has been received with correct value on \
+                    agent, step {}".format(i))
 
-        self.check_log(
-            self.repeater1,
-            "agent_wlan2",
-            "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+            self.check_log(
+                self.repeater1, "agent_wlan0", "tlvTransmitPowerLimit {}".format(tp20dBm))
 
-        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
+            self.check_log(self.repeater1, "agent_wlan0",
+                           "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+
+            self.check_log(
+                self.repeater1, "agent_wlan2", "tlvTransmitPowerLimit {}".format(tp20dBm))
+
+            self.check_log(self.repeater1, "agent_wlan2",
+                           "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+
+            debug("Confirming CHANNEL_SELECTION_RESPONSE_MESSAGE has been received, \
+                step {}".format(i))
+
+            self.check_log(self.gateway, "controller", "CHANNEL_SELECTION_RESPONSE_MESSAGE")
+
+            debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE has been received on \
+                controller step {}".format(i))
+
+            result, ocrm_line, mid_match = self.check_log(
+                                        self.gateway, "controller",
+                                        r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)')
+
+            if (mid_match):
+                self.check_log(
+                    self.repeater1,
+                    "agent_wlan0",
+                    "ACK_MESSAGE, mid={}".format(mid_match[0]))
+
+                self.check_log(
+                    self.repeater1,
+                    "agent_wlan2",
+                    "ACK_MESSAGE, mid={}".format(mid_match[0]))
+
+            debug("Confirming tx_power has been received with correct value on controller, \
+                step {}".format(i))
+            self.check_log(self.gateway, "controller", "tx_power={}".format(tp20dBm))
 
     def test_ap_capability_query(self):
         self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8001)
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -384,7 +384,92 @@ class TestFlows:
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
 
+        debug("Send empty channel selection request")
+        cs_req_mid = self.gateway_ucc.dev_send_1905(self.mac_repeater1,
+                                                    0x8006, tlv(0x00, 0x0000, "{}"))
+        time.sleep(1)
+
+        debug("Confirming channel selection request has been received on controller")
+        self.check_log(
+            self.gateway,
+            "controller",
+            r"CHANNEL_SELECTION_RESPONSE_MESSAGE, mid={}".format(cs_req_mid)
+        )
+
+        debug("Confirming empty channel selection request has been received on agent")
+        self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+        self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+
+        debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message has been received on \
+            controller with mid")
+
+        result, ocrm_line, mid_match = self.check_log(
+                                            self.gateway, "controller",
+                                            r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)'
+                                        )
+        if (mid_match):
+            debug("Confirming ACK_MESSAGE from the controller \
+                with same mid as OPERATING_CHANNEL_REPORT_MESSAGE")
+            self.check_log(
+                self.repeater1,
+                "agent_wlan0",
+                "ACK_MESSAGE, mid={}".format(mid_match[0])
+            )
+
+            self.check_log(
+                self.repeater1,
+                "agent_wlan2",
+                "ACK_MESSAGE, mid={}".format(mid_match[0])
+            )
+
         tp20dBm = 0x14
+        tp21dBm = 0x15
+
+        for payload_transmit_power in (tp20dBm, tp21dBm):
+            debug("Send empty channel selection request with changing tx_power_limit")
+            cs_req_mid = self.gateway_ucc.dev_send_1905(
+                            self.mac_repeater1,
+                            0x8006,
+                            tlv(0x8D, 0x0007, '{} 0x{:02x}'.format(self.mac_repeater1_wlan0,
+                                payload_transmit_power)),
+                            tlv(0x8D, 0x0007, '{} 0x{:02x}'.format(self.mac_repeater1_wlan2,
+                                payload_transmit_power))
+                        )
+            time.sleep(1)
+
+            self.check_log(
+                self.gateway,
+                "controller",
+                r"CHANNEL_SELECTION_RESPONSE_MESSAGE, mid={}".format(cs_req_mid)
+            )
+
+            self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+            self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+
+            self.check_log(self.repeater1, "agent_wlan0",
+                           "tlvTransmitPowerLimit {}".format(payload_transmit_power))
+            self.check_log(self.repeater1, "agent_wlan2",
+                           "tlvTransmitPowerLimit {}".format(payload_transmit_power))
+
+            self.check_log(self.gateway, "controller", "tx_power={}".format(payload_transmit_power))
+
+            result, ocrm_line, mid_match = self.check_log(
+                                                self.gateway, "controller",
+                                                r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)',
+                                                ocrm_line
+                                            )
+            if (result):
+                self.check_log(
+                    self.repeater1,
+                    "agent_wlan0",
+                    "ACK_MESSAGE, mid={}".format(mid_match[0])
+                )
+
+                self.check_log(
+                    self.repeater1,
+                    "agent_wlan2",
+                    "ACK_MESSAGE, mid={}".format(mid_match[0])
+                )
 
         # payload_wlan0 - request for change channel on 6
         payload_wlan0 = (
@@ -463,7 +548,8 @@ class TestFlows:
             self.check_log(
                 self.gateway,
                 "controller",
-                r"CHANNEL_SELECTION_RESPONSE_MESSAGE, mid={}".format(mid))
+                r"CHANNEL_SELECTION_RESPONSE_MESSAGE, mid={}".format(mid)
+            )
 
             debug(
                 "Confirming channel selection request has been received on agent,step {}".format(i))
@@ -497,7 +583,8 @@ class TestFlows:
 
             result, ocrm_line, mid_match = self.check_log(
                                         self.gateway, "controller",
-                                        r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)')
+                                        r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)',
+                                        ocrm_line)
 
             if (mid_match):
                 self.check_log(
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -220,17 +220,20 @@ class TestFlows:
         self.mac_repeater2_wlan2 = re.search(rb"link/ether " + RE_MAC, mac_repeater2_wlan2_output).group('mac').decode()
         debug("Repeater2 wl2: {}".format(self.mac_repeater2_wlan2))
 
-    def _check_log_internal(self, device: str, program: str, regex: str) -> bool:
+    def _check_log_internal(self, device: str, program: str, regex: str, start_line: int):
         '''Search for regex in logfile for program on device.'''
         logfilename = os.path.join(self.rootdir, 'logs', device, 'beerocks_{}.log'.format(program))
         with open(logfilename) as logfile:
-            for line in logfile.readlines():
-                if re.search(regex, line):
+            for (i, v) in enumerate(logfile.readlines()):
+                if i <= start_line:
+                    continue
+                search = re.search(regex, v)
+                if search:
                     debug("Found '{}'\n\tin {}".format(regex, logfilename))
-                    return True
-        return False
+                    return (True, i, search.groups())
+        return (False, start_line, None)
 
-    def check_log(self, device: str, program: str, regex: str) -> bool:
+    def check_log(self, device: str, program: str, regex: str, start_line=0):
         '''Verify that on "device" the logfile for "program" matches "regex", fail if not.'''
         try:
             # HACK check_log is often used immediately after triggering a message on the other side.
@@ -238,12 +241,14 @@ class TestFlows:
             # we tend to check it too quickly. As a simple workaround, add a small sleep here.
             # The good solution is to retry with a small timeout.
             time.sleep(.1)
-            if self._check_log_internal(device, program, regex):
-                return True
+            result, line, value = self._check_log_internal(device, program, regex, start_line)
+            if result:
+                return True, line, value
             else:
-                return self.fail("'{}'\n\tin log of {} on {}".format(regex, program, device))
+                return \
+                  self.fail("'{}'\n\tin log of {} on {}".format(regex, program, device)), line, None
         except OSError:
-            return self.fail("Can't read log of {} on {}".format(program, device))
+            return self.fail("Can't read log of {} on {}".format(program, device)), line, None
 
     def wait_for_log(self, device: str, program: str, regex: str, timeout: float) -> bool:
         deadline = time.monotonic() + timeout


tests/test_flows.py
```diff
@@ -383,17 +383,33 @@ def test_channel_selection(self):
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
 
         self.debug("Send channel selection request")
-        self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8006)
+
+        payload_wlan0="0x14 {0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x52 {0x00 0x00}} {0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}} {0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x73 {0x00 0x00}} {0x74 {0x00 0x00}} {0x75 {0x00 0x00}} {0x76 {0x00 0x00}} {0x77 {0x00 0x00}} {0x78 {0x00 0x00}} {0x79 {0x00 0x00}} {0x7A {0x00 0x00}} {0x7B {0x00 0x00}} {0x7C {0x00 0x00}} {0x7D {0x00 0x00}} {0x7E {0x00 0x00}} {0x7F {0x00 0x00}} {0x80 {0x00 0x00}} {0x81 {0x00 0x00}} {0x82 {0x00 0x00}}"
```
tomereli: Please add comment to the test explaining what are the steps taken.
Also for such payloads, please add a comment explaining it.

I guess both payloads are channel selection requests, one to change to channel 6 and one to channel 36?

tests/test_flows.py
```diff
@@ -383,17 +383,33 @@ def test_channel_selection(self):
...
-        # TODO: currently controller sends empty channel selection request, so no switch is performed
-        # self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
-        # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
+        self.check_log(self.repeater1, "agent_wlan0", "tlvTransmitPowerLimit")
+        self.check_log(self.repeater1, "agent_wlan0", "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+
+        self.check_log(self.repeater1, "agent_wlan2", "tlvTransmitPowerLimit")
+        self.check_log(self.repeater1, "agent_wlan2", "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+
+        self.check_log(self.gateway, "controller", "PERATING_CHANNEL_REPORT_MESSAGE")
```
tomereli:
```suggestion
        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
```

tests/test_flows.py
```diff
@@ -383,17 +383,33 @@ def test_channel_selection(self):
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
 
         self.debug("Send channel selection request")
-        self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8006)
```
tomereli, line 383: Commit headline of the first and second commits are identical - please use a different headline (Part 1 and Part 2) can work.

Also there is a typo in the commit message - "teststs" --> "tests"

This can work: `test_flows.py: Enhance channel selection test - part 1`

tests/test_flows.py
```diff
@@ -384,22 +384,67 @@ def test_channel_selection(self):
...
+            '{0x7C {0x00 0x00}}',
+            '{0x7D {0x00 0x00}}',
+            '{0x7E {0x00 0x00}}',
+            '{0x7F {0x00 0x00}}',
+            '{0x80 {0x00 0x00}}',
+            '{0x81 {0x00 0x00}}',
+            '{0x82 {0x00 0x00}}',
+        ])
+
+        payload_wlan2=' '.join([
```
tomereli: I don't understand this change - I get the ACK change, but not the payload. Please explain it better in the commit message and in a comment. Also, please keep in 1-3 lines, it doesn't help much to spread this over a full screen, especially since the values are just hex.

tomereli (changes requested):
Some comments inline, but there are more important things to test in channel selection, based on my recent experience.
These are the tests we need, which if we had them, we would save a lot of debugging time:
1. Trigger channel selection to channel 6. Check that operating channel report was sent - dummy mode should simulate channel 1 at init, so switch to channel 6 may be needed to be implemented in dummy.
2. Trigger channel selection to channel 6 **again** - check that operating channel report is sent again. This is to catch bugs when we don't send channel report when there is no need to switch channel

tests/test_flows.py
VladyslavTupikin: Yes. payload for wlan0 - request for change on channel 6, payload for wlan2 - request for change on channel 36

tests/test_flows.py
```diff
@@ -401,18 +401,92 @@ def test_channel_selection(self):
...
-        self.debug("Confirming channel selection request has been received on agent")
-        self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
-        self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
-
-        #self.debug("Confirming 1905.1 Ack Message request was received on agent")
-        # TODO: When creating handler for the ACK message on the agent, replace lookup of this string
-        # TODO: currently controller sends empty channel selection request, so no switch is performed
-        # self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
-        # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
+        payload_empty="0x14"
```
abelog: Suggest to rename it to something like: `payload_transmit_power` to avoid confusion

tomereli: actually if we keep to the payload_XXX structure, I think payload_empty is fine here. We should add a comment though explaining what an empty channel selection payload means.

tests/test_flows.py
```diff
@@ -401,18 +401,92 @@ def test_channel_selection(self):
...
+
+        # payload_wlan0 - request for change channel on 6
+        payload_wlan0=' '.join([
```
tomereli: There is nothing to gain by splitting this to multiple lines - can you keep it in 2-3 lines?
Or even better - add it to another yaml file which you can import from this file which can parse into a dictionary of messages:
```yaml
tlv_data:
    channel_selection:
         channel6: [...],
         channel36: [...],
         ...
    autoconfig:
        ...
};
```

tomereli (approved):
Small changes requested but are easy to make, so approving.

tests/test_flows.py
VladyslavTupikin: I have chosen 1 option - 3 strings with payload.

tomereli (changes requested):
Please also rebase on #982. I add support in dummy mode for setting tx_power.
Please add the following tests:
1. Change only tx power and verify that channel preference report is sent
1.2 Make sure the channel preference tx power matches the tx power setting sent in the preference report (add a print if its missing in the controller)
2. Send empty channel selection request and verify that channel preference report is sent

VladyslavTupikin:
Updated branch with force push key because branch rebased on latest master branch which contains tx_power feature for dummy mode.

tests/test_flows.py
```diff
@@ -401,18 +401,88 @@ def test_channel_selection(self):
...
-        # TODO: When creating handler for the ACK message on the agent, replace lookup of this string
-        # TODO: currently controller sends empty channel selection request, so no switch is performed
-        # self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
-        # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
+        self.debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message has been received on controller")
+        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
+
+        # Send channel selection request with chnging tx_power_limit 
+        # Payload for transmit power - 20 for both radio
+        payload_transmit_power ="0x14"
```
MuteSpirit: Please use assignment code style as in another file: with spaces around `=`:
```
payload_transmit_power = "0x14"
```

tests/test_flows.py
```diff
@@ -401,18 +401,88 @@ def test_channel_selection(self):
...
+            self.gateway_ucc.dev_send_1905(
+                self.mac_repeater1,
+                0x8006,
+                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan0} {payload_transmit_power}'),
+                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan2} {payload_transmit_power}')
+            )
+
+            self.check_log(self.repeater1, "agent_wlan0", "tlvTransmitPowerLimit %d" % int(payload_transmit_power,16))
+            self.check_log(self.repeater1, "agent_wlan2", "tlvTransmitPowerLimit %d" % int(payload_transmit_power,16))
+            self.check_log(self.gateway, "controller", "tx_power=%d" % int(payload_transmit_power,16))
```
MuteSpirit: Maybe keep result of `int(payload_transmit_power,16)` in local variable and avoid calculation duplication?

tests/test_flows.py
```diff
@@ -401,18 +401,88 @@ def test_channel_selection(self):
...
-        # self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
-        # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
+        self.debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message has been received on controller")
+        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
+
+        # Send channel selection request with chnging tx_power_limit 
+        # Payload for transmit power - 20 for both radio
+        payload_transmit_power ="0x14"
+
+        for i in range(1, 3):
```
MuteSpirit: If goal of this loop is set 2 different values of transmit power then maybe do that more explicitly:
```
tp20dBm = "0x14"
tp21dBm = "0x15"
for payload_transmit_power in (tp20dBm, tp21dBm):
```

MuteSpirit (changes requested):

tests/test_flows.py
arnout: :-1: part 1 part 2 are not useful commit messages. The summary line should say what is done, the body why and more details of what.

So something like:

test_flows.py: channel_selection: add test for empty channel selection request

test_flows.py: channel_selection: add test for power limit

test_flows.py: channel_selection: add test for channel switch

test_flows.py: channel_selection: add test for no channel switch

tests/test_flows.py
```diff
@@ -401,18 +401,86 @@ def test_channel_selection(self):
...
+                when we don't send channel report when there is no need to
+                switch channel
+        """
+        for i in range(1, 3):
+            transmit_power = int(tp20dBm,16)
+            self.debug("Send channel selection request, step %i" % i)
+            self.gateway_ucc.dev_send_1905(
+                self.mac_repeater1,
+                0x8006,
+                tlv(0x8B,0x005F, f'{self.mac_repeater1_wlan0} {payload_wlan0}'),
```
rmelotte: I believe some people only have access to Python 3.5, which doesn't have f-strings unfortunately.

Maybe we should check if it's still the case.

tests/test_flows.py
```diff
@@ -401,18 +401,86 @@ def test_channel_selection(self):
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
 
-        self.debug("Send channel selection request")
-        self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8006)
-        time.sleep(1)
-        self.debug("Confirming channel selection request has been received on agent")
+        # Send empty channel selection request
+        self.debug("Send empty channel selection request")
```
arnout: A comment that repeats the log message is redundant.

However, this is not what happens (or should happen). When dev_send_1905 without TLVs is done, the controller should add the TLVs itself. Cfr. step 5 of 5.5.1: "MCUT to send a Channel Selection Request message to CTT Agent1 for each Channel Preference TLV received in the Channel Preference Report message. (MCUT is expected to
generate the Channel Preference TLV)"

So the problem is actually in the controller, in that it doesn't send a channel preference TLV while it should do so. Probably it's because the agent sends an empty channel preference TLV.

So the problem is in handle_cmdu_1905_channel_preference_report() which doesn't correctly fill the certification buffer.

That said, we should obviously still have this test with a completely empty channel selection request, because it does expose a corner case with a bug. A trick to achieve that is:

```
self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8006, tlv(0x00, 0x0000, "")
```

That adds an EOM TLV which would otherwise anyway be added automatically, but to the CAPI parser it looks like a TLV was provided.

tests/test_flows.py
```diff
@@ -401,18 +401,86 @@ def test_channel_selection(self):
...
+        self.gateway_ucc.dev_send_1905(self.mac_repeater1,0x8006)
```
arnout: Please don't introduce PEP8 issues (i.e., space after comma).

rmelotte:
It looks like there is still an overflow.

I guess this needs to be rebased to integrate this change? https://github.com/prplfoundation/prplMesh/pull/1014

tests/test_flows.py
```diff
@@ -401,18 +401,86 @@ def test_channel_selection(self):
...
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
 
-        #self.debug("Confirming 1905.1 Ack Message request was received on agent")
-        # TODO: When creating handler for the ACK message on the agent, replace lookup of this string
-        # TODO: currently controller sends empty channel selection request, so no switch is performed
-        # self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
-        # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
+        self.debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message has been received on controller")
+        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
```
arnout: Please also check the CHANNEL_SELECTION_RESPONSE.

tests/test_flows.py
```diff
@@ -401,18 +401,86 @@ def test_channel_selection(self):
...
+
+        # Send channel selection request with chnging tx_power_limit 
+        # Payload for transmit power (tp)
+        tp20dBm = "0x14"
+        tp21dBm = "0x15"
+
+        for payload_transmit_power in (tp20dBm,tp21dBm):
+            transmit_power = int(payload_transmit_power,16)
```
arnout:
```suggestion
        tp20dBm = 0x14
        tp21dBm = 0x15

        for payload_transmit_power in (tp20dBm,tp21dBm):
```

tests/test_flows.py
```diff
@@ -401,18 +401,86 @@ def test_channel_selection(self):
...
+            self.debug("Send empty channel selection request with changing tx_power_limit")
+            self.gateway_ucc.dev_send_1905(
+                self.mac_repeater1,
+                0x8006,
+                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan0} {payload_transmit_power}'),
+                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan2} {payload_transmit_power}')
```
arnout: Unfortunately, @tomereli and friends are stuck with Python 3.5 or worse, so no f-strings...

```suggestion
                tlv(0x8D, 0x0007, '{} {:02x}'.format(self.mac_repeater1_wlan0, payload_transmit_power)),
                tlv(0x8D, 0x0007, '{} {:02x}'.format(self.mac_repeater2_wlan0, payload_transmit_power)),
```

(and with f-string it could have been `tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan0} {payload_transmit_power:02x}

tests/test_flows.py
```diff
@@ -401,18 +401,86 @@ def test_channel_selection(self):
...
+
+            self.debug("Confirming tlvTransmitPowerLimit has been received with correct value on agent, step %i" % i)
+            self.check_log(self.repeater1, "agent_wlan0", "tlvTransmitPowerLimit %d" % transmit_power)
+            self.check_log(self.repeater1, "agent_wlan0", "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+            self.check_log(self.repeater1, "agent_wlan2", "tlvTransmitPowerLimit %d" % transmit_power)
+            self.check_log(self.repeater1, "agent_wlan2", "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+
+            self.debug("Confirming ACK_MESSAGE has been received on agent, step %i" % i)
+            self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
+            self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
```
arnout: Which ACK_MESSAGE?

For an ACK, we should definitely always check the MID because there are so many acks that it will always hit one.

arnout (approved):
Although I have some important comments, I'm already going to approve. Please make the necessary fixes and do an interactive rebase to update each commit separately as appropriate. Then you can push and set the ReadyForMerge label.

The remaining issue that the controller doesn't add a TLV to the channel selection request while it should doesn't need to be fixed in this PR. Actually, it doesn't really need to get fixed, because it does work fine in the certification setup, since there the channel preference TLV will have some contents.

tests/test_flows.py
VladyslavTupikin: This solution does not work. It should be like:

`tlv(0x8D, 0x0007, '{} {}'.format(self.mac_repeater1_wlan0, hex(payload_transmit_power))),`

arnout: What does not work about it?

`hex` is not good, because it doesn't include a leading zero for values < 16.

Oh, now I see, I forgot the 0x bit, is that it?
```suggestion
                tlv(0x8D, 0x0007, '{} 0x{:02x}'.format(self.mac_repeater1_wlan0, payload_transmit_power)),
                tlv(0x8D, 0x0007, '{} 0x{:02x}'.format(self.mac_repeater1_wlan0, payload_transmit_power))
```

VladyslavTupikin: Sure, but we should pay attention to `#define TX_POWER_MIN 19`.
As I understood we can't send tx_power less than 19

tomereli (approved):

abelog (approved):

arnout:
https://github.com/prplfoundation/prplMesh/pull/965#discussion_r398702745: 
> part 1 part 2 are not useful commit messages. The summary line should say what is done, the body why and more details of what.

You marked this as resolved but the commit message summary still says "part 1, part 2, ...". You have every right to disagree with me, but then please explain why.

VladyslavTupikin:
> [#965 (comment)](https://github.com/prplfoundation/prplMesh/pull/965#discussion_r398702745):
> 
> > part 1 part 2 are not useful commit messages. The summary line should say what is done, the body why and more details of what.
> 
> You marked this as resolved but the commit message summary still says "part 1, part 2, ...". You have every right to disagree with me, but then please explain why.

VladyslavTupikin:
> [#965 (comment)](https://github.com/prplfoundation/prplMesh/pull/965#discussion_r398702745):
> 
> > part 1 part 2 are not useful commit messages. The summary line should say what is done, the body why and more details of what.
> 
> You marked this as resolved but the commit message summary still says "part 1, part 2, ...". You have every right to disagree with me, but then please explain why.

I added this header according to Tomer's commentary and I did not see your answer in that thread. 
I agree with your explanations from the thread - summary should shortly describe work which done in the commit. I can fix it if need.

MuteSpirit (approved):

rmelotte (approved):
The splitting into commits still looks really weird:
- the first commit introduces the payloads on one line.
- the second one switches to braces and newlines.
- the fourth one changes them back to one line

There is also at least one mention of "part 3", but we don't know anymore what it refers to. 

Overall the changes look ok though, so approving.

tests/test_flows.py
```diff
@@ -381,17 +381,33 @@ def test_channel_selection(self):
...
+          (0x8B,0x005F,"{%s %s}" % (self.mac_repeater1_wlan0,payload_wlan0)), \
+          (0x8D,0x0007,"{%s %s}" % (self.mac_repeater1_wlan0,payload_empty)), \
+          (0x8B,0x004C,"{%s %s}" % (self.mac_repeater1_wlan2,payload_wlan2)), \
+          (0x8D,0x0007,"{%s %s}" % (self.mac_repeater1_wlan2,payload_empty)))
         time.sleep(1)
+
         self.debug("Confirming channel selection request has been received on agent")
+
+        self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
+        self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
```
arnout: These are already checked above (after the channel preference query was sent), so not needed here.

tests/test_flows.py
arnout: You marked this as resolved but the issue is still there... However, it gets removed again later so OK.

tests/test_flows.py
```diff
@@ -381,17 +381,33 @@ def test_channel_selection(self):
...
+
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
 
-        #self.debug("Confirming 1905.1 Ack Message request was received on agent")
-        # TODO: When creating handler for the ACK message on the agent, replace lookup of this string
-        # TODO: currently controller sends empty channel selection request, so no switch is performed
-        # self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
-        # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
+        self.check_log(self.repeater1, "agent_wlan0", "tlvTransmitPowerLimit")
```
arnout: It would be better to first check that the controller has received the ACK for the channel selection request, by storing the mid returned by dev_send_1905 and including that in the regex:
```
        self.check_log(self.gateway, "controller", r"ACK_MESSAGE, mid=0x{:04x}".format(mid))
```

tests/test_flows.py
```diff
@@ -380,24 +380,71 @@ def test_channel_selection(self):
...
+            '{0x7A {0x00 0x00}}',
+            '{0x7B {0x00 0x00}}',
+            '{0x7C {0x00 0x00}}',
+            '{0x7D {0x00 0x00}}',
+            '{0x7E {0x00 0x00}}',
+            '{0x7F {0x00 0x00}}',
+            '{0x80 {0x00 0x00}}',
+            '{0x81 {0x00 0x00}}',
+            '{0x82 {0x00 0x00}}',
+        ])
```
arnout: IMO it's better to use string constant concatenation instead of `join`:
```suggestion
        payload_wlan0 = (
            "0x14"
            " {0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}}"
            " {0x52 {0x00 0x00}}"
            " {0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}}"
            " {0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}}"
            " {0x73 {0x00 0x00}}"
            " {0x74 {0x00 0x00}}"
            " {0x75 {0x00 0x00}}"
            " {0x76 {0x00 0x00}}"
            " {0x77 {0x00 0x00}}"
            " {0x78 {0x00 0x00}}"
            " {0x79 {0x00 0x00}}"
            " {0x7A {0x00 0x00}}"
            " {0x7B {0x00 0x00}}"
            " {0x7C {0x00 0x00}}"
            " {0x7D {0x00 0x00}}"
            " {0x7E {0x00 0x00}}"
            " {0x7F {0x00 0x00}}"
            " {0x80 {0x00 0x00}}"
            " {0x81 {0x00 0x00}}"
            " {0x82 {0x00 0x00}}")
```

Also note: correct spacing around ` = `, and we mostly use double quotes for strings.

Note: this comment is more a way for both of us to learn new things and to explore possibilities than an actual suggestion of "you should change it like this".

tests/test_flows.py
```diff
@@ -380,24 +380,71 @@ def test_channel_selection(self):
...
+            tlv(0x8B,0x004C, f'{self.mac_repeater1_wlan2} {payload_wlan2}'),
+            tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan2} {payload_empty}')
+        )
         time.sleep(1)
 
         self.debug("Confirming channel selection request has been received on agent")
 
-        self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
-        self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_PREFERENCE_QUERY_MESSAGE")
-
```
arnout: So here you remove that redundant stuff again... Please try to make your commits clean, i.e. not introduce something that you fix later.

In this case, I think the first two commits can just be squashed.

tests/test_flows.py
```diff
@@ -406,8 +453,10 @@ def test_channel_selection(self):
 
         self.check_log(self.repeater1, "agent_wlan2", "tlvTransmitPowerLimit")
         self.check_log(self.repeater1, "agent_wlan2", "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
+        self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE")
```
arnout: Checking for ACK should always include the MID, because the logs are full of ACKs. However, in this case it's not that easy, because this is the ACK of the operating channel report message, which is sent autonomously by the agent. So you'd have to parse the MID out of the controller log...

tests/test_flows.py
```diff
@@ -431,32 +431,41 @@ def test_channel_selection(self):
...
-        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
+        """
+        Step 1: Trigger channel selection to channel 6 and 36. Check that
+                operating channel report was sent.
+
+        Step 2: Trigger channel selection to channel 6 and 36 again - check that
+                operating channel report is sent again. This is to catch bugs
+                when we don't send channel report when there is no need to
+                switch channel
+        """
```
arnout, line 532: I think I mentioned this before, but: excellent explanation!

tests/test_flows.py
```diff
@@ -431,32 +431,41 @@ def test_channel_selection(self):
...
+        for i in range(1, 3):
```
arnout, line 533: Didn't someone comment that this is easier to understand as
```suggestion
        for i in (1, 2):
```

tests/test_flows.py
```diff
@@ -380,57 +380,44 @@ def test_channel_selection(self):
...
-            '{0x7B {0x00 0x00}}',
-            '{0x7C {0x00 0x00}}',
-            '{0x7D {0x00 0x00}}',
-            '{0x7E {0x00 0x00}}',
-            '{0x7F {0x00 0x00}}',
-            '{0x80 {0x00 0x00}}',
-            '{0x81 {0x00 0x00}}',
-            '{0x82 {0x00 0x00}}',
-        ])
+        payload_wlan0="0x14 {0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x52 {0x00 0x00}} {0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}} {0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x73 {0x00 0x00}} {0x74 {0x00 0x00}} {0x75 {0x00 0x00}} {0x76 {0x00 0x00}} {0x77 {0x00 0x00}} {0x78 {0x00 0x00}} {0x79 {0x00 0x00}} {0x7A {0x00 0x00}} {0x7B {0x00 0x00}} {0x7C {0x00 0x00}} {0x7D {0x00 0x00}} {0x7E {0x00 0x00}} {0x7F {0x00 0x00}} {0x80 {0x00 0x00}} {0x81 {0x00 0x00}} {0x82 {0x00 0x00}}"
```
arnout: OK, so you prefer it like this after all? Then make up your mind and do it correctly right away.

Note that you'll need to learn to work with `git rebase -i` to fix this. You really should make sure that when you make changes after a review, the changes end up in the right commit. It makes for a cleaner history.

The commit message says "Reject payload variables view, decide keep it in 2 long strings." but that is not something that needs to be mentioned in a commit message. Instead, write it as a reaction to the comment that asked to split the line and be done with it.

Finally, this line really is too long and won't pass PEP8 checks. So some kind of splitting will need to be done.

tests/test_flows.py
```diff
@@ -380,57 +380,44 @@ def test_channel_selection(self):
...
+        self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+        self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+
+        self.debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message has been received on controller")
+        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
+
+        # Send channel selection request with chnging tx_power_limit 
+        # Payload for transmit power (tp)
+        tp20dBm = "0x14"
+        tp21dBm = "0x15"
```
arnout: I suggested before to write this as
```suggestion
        # Send channel selection request with changing tx_power_limit 
        # Payload for transmit power (tp)
        tp20dBm = 0x14
        tp21dBm = 0x15
```
You marked that as resolved with no feedback, so that sounds as if you accepted the proposal. But it's still here...

tests/test_flows.py
```diff
@@ -441,32 +428,39 @@ def test_channel_selection(self):
...
             self.debug("Send channel selection request, step %i" % i)
             self.gateway_ucc.dev_send_1905(
                 self.mac_repeater1,
                 0x8006,
                 tlv(0x8B,0x005F, f'{self.mac_repeater1_wlan0} {payload_wlan0}'),
-                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan0} {payload_empty}'),
+                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan0} {tp20dBm}'),
                 tlv(0x8B,0x004C, f'{self.mac_repeater1_wlan2} {payload_wlan2}'),
-                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan2} {payload_empty}')
+                tlv(0x8D,0x0007, f'{self.mac_repeater1_wlan2} {tp20dBm}')
```
arnout: f-strings are unfortunately not allowed since this script needs to run with Python 3.5.

tests/test_flows.py
```diff
@@ -441,32 +428,39 @@ def test_channel_selection(self):
...
             )
             time.sleep(1)
 
-            self.debug("Confirming channel selection request has been received on agent step %i" % i)
+            self.debug("Confirming channel selection request has been received on agent, step %i" % i)
```
arnout: Please do this right the first time.

Actually, I also don't see much reason to keep this as a separate commit. For review and for development it's nice to do it step by step, but now you can squash them.

tests/test_flows.py
```diff
@@ -223,6 +223,20 @@ def init(self):
...
+            for (i, v) in enumerate(logfile.readlines()):
+                if i <= start_line:
+                    continue
+                match = re.match(regex, v)
+                if match:
+                    self.debug("Found '{}'\n\tin {}\tline {} \t mid {}".format(regex, logfilename, i,  # noqa: E501
+                                match.group(1)))
+                    return (True, i, match.group(1))
+        return self.fail("'{}'\n\tin log of {} on {}".format(regex, program, device))
+
```
arnout: Instead of rewriting check_log, I think it's better to modify check_log (or actually, _check_log_internal) directly. It can return (line, match) instead of just True or False. If it is not found, it would return (last line, None).

For the start_line, it can also easily be an optional argument of check_log (default 0).

tests/test_flows.py
```diff
@@ -380,43 +394,69 @@ def test_channel_selection(self):
...
         self.debug("Confirming empty channel selection request has been received on agent")
         self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
         self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
 
-        self.debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message has been received on controller")
-        self.check_log(self.gateway, "controller", "OPERATING_CHANNEL_REPORT_MESSAGE")
+        self.check_log(self.gateway, "controller", "CHANNEL_SELECTION_RESPONSE_MESSAGE")
+
+        self.debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message \
+            has been received on controller with mid")
```
arnout: Prefer the following type of string splitting:
```suggestion
        self.debug("Confirming OPERATING_CHANNEL_REPORT_MESSAGE message"
            "has been received on controller with mid")
```

(backslash isn't needed in this case since it's between `(...)`)

tests/test_flows.py
```diff
@@ -380,43 +394,69 @@ def test_channel_selection(self):
...
+
+        ocrm_line = 0
+        result, ocrm_line, mid = self.parse_log(self.gateway, "controller",
+                                                r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)',
+                                                ocrm_line)
+        if (result):
```
arnout: So with my suggestion this would become something like:
```suggestion
        ocrm_line, mid_match = self.check_log(self.gateway, "controller",
                                                r'OPERATING_CHANNEL_REPORT_MESSAGE, mid=(\d+)')
        if mid_match:
            mid = mid_match.group(1)
```

tests/test_flows.py
```diff
@@ -380,43 +394,69 @@ def test_channel_selection(self):
...
+            self.debug("Confirming ACK_MESSAGE from the controller \
+                with same mid as OPERATING_CHANNEL_REPORT_MESSAGE")
+            self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE, mid=%s" % mid)
+            self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE, mid=%s" % mid)
```
arnout: Please prefer format instead of %:
```suggestion
            self.check_log(self.repeater1, "agent_wlan0", "ACK_MESSAGE, mid={}".format(mid))
            self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE, mid={}".format(mid))
```

tests/test_flows.py
```diff
@@ -380,43 +394,69 @@ def test_channel_selection(self):
...
 
-        # Send channel selection request with chnging tx_power_limit 
         # Payload for transmit power (tp)
-        tp20dBm = "0x14"
-        tp21dBm = "0x15"
+        tp20dBm = 0x14
+        tp21dBm = 0x15
```
arnout, line 426: Ah, again, the change was not applied to the correct commit...

tests/test_flows.py
```diff
@@ -380,43 +394,69 @@ def test_channel_selection(self):
...
+                           payload_transmit_power)
+            self.check_log(self.repeater1, "agent_wlan2", "tlvTransmitPowerLimit %d" %
+                           payload_transmit_power)
+
+            self.check_log(self.gateway, "controller", "tx_power=%d" % payload_transmit_power)
+            self.check_log(self.gateway, "controller", "CHANNEL_SELECTION_RESPONSE_MESSAGE")
+
+            result, ocrm_line, mid = self.parse_log(self.gateway, "controller",
+                                                    r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)',  # noqa: E501
+                                                    ocrm_line)
```
arnout: In this case, a better way (IMO) to work around the line length problem is to not use visual indent:
```suggestion
            result, ocrm_line, mid = self.parse_log(self.gateway, "controller",
                r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)',
                ocrm_line)
```
I *think* flake8 accepts that (it doesn't complain about visual indent if visual indent would make it exceed the line length limit).

tests/test_flows.py
```diff
@@ -380,43 +394,69 @@ def test_channel_selection(self):
...
         # payload_wlan0 - request for change channel on 6
-        payload_wlan0="0x14 {0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x52 {0x00 0x00}} {0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}} {0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x73 {0x00 0x00}} {0x74 {0x00 0x00}} {0x75 {0x00 0x00}} {0x76 {0x00 0x00}} {0x77 {0x00 0x00}} {0x78 {0x00 0x00}} {0x79 {0x00 0x00}} {0x7A {0x00 0x00}} {0x7B {0x00 0x00}} {0x7C {0x00 0x00}} {0x7D {0x00 0x00}} {0x7E {0x00 0x00}} {0x7F {0x00 0x00}} {0x80 {0x00 0x00}} {0x81 {0x00 0x00}} {0x82 {0x00 0x00}}"
+        payload_wlan0 = "0x14 {0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x52 {0x00 0x00}} {0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}} {0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} {0x73 {0x00 0x00}} {0x74 {0x00 0x00}} {0x75 {0x00 0x00}} {0x76 {0x00 0x00}} {0x77 {0x00 0x00}} {0x78 {0x00 0x00}} {0x79 {0x00 0x00}} {0x7A {0x00 0x00}} {0x7B {0x00 0x00}} {0x7C {0x00 0x00}} {0x7D {0x00 0x00}} {0x7E {0x00 0x00}} {0x7F {0x00 0x00}} {0x80 {0x00 0x00}} {0x81 {0x00 0x00}} {0x82 {0x00 0x00}}"  # noqa: E501
```
arnout: Here I think it's better to split the line instead of adding a noqa.

You could also use a helper function to generate the repetitive non-operable lists:
```
def gen_disabled_opclass(opclass: int):
    return '{0x%02x {0x00 0x00}}' % opclass

payload_wlan0 = "0x14" \
    " {0x51 {0x0C {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}} " + \
    gen_disabled_opclass(0x52) + \
    " {0x53 {0x08 {0x01 0x02 0x03 0x04 0x05 0x07 0x08 0x09} 0x00}}" \
    " {0x54 {0x08 {0x05 0x07 0x08 0x09 0x0A 0x0B 0x0C 0x0D} 0x00}}" + \
    ' '.join([gen_disabled_opclass(opclass) for opclass in range(0x73, 0x82)])
```
(note: just a suggestion, I'm not that convinced it looks better. But it is future safe if we want to test other channel selections later.)

arnout (changes requested):
Looks like there was a problem with the commit splitting... I'm going to mark this as changes requested this time, since it's good to use this PR to learn how to do commit splitting and rebasing properly.

There are two changes that I really want you to make to the code itself:
- check the ACK of the channel selection request on the controller side including its MID;
- extend check_log instead of introducing parse_log.

The rest of the comments are more informational.

The review comments I'm making will look pretty confusing because I've reviewed commit per commit, so I saw a lot of things that were not OK but that were fixed in later commits. That is one indication why proper commit splitting is important. The most important reason, however, is for future reference. We want to be able to use the git log or git blame to find out why the code looks like it is (e.g. why some bug was introduced, or why something was done in that particular way). If there are a lot of commits in between that change things in irrelevant ways, it becomes much more difficult to effectively use the git log.

Normally, I'd suggest to just squash everything in a single commit and be done with it - the overall changes are sufficiently related to warrant that. However, I think this is a good learning opportunity to work with rebase and proper commit splitting. So let's try to rework this into four commits:
1. Use full channel selection request instead of empty one (essentially the first two commits squashed together, but with proper PEP8 and no f-strings, and with the check of the ACK on the controller).
2. Extend check_log (but not using the new feature yet).
3. Extend the "full" channel selection request test with power limit and doing it two times.
4. Additional channel selection tests (empty request, different power limits).

You will also need to rebase it on master because it is conflicting. And it's also going to conflict heavily with #1031 so we need to synchronise on that.

tests/test_flows.py
arnout: Turns out that I was wrong about that. Visual indent is always required, but you can work around it like this:
```suggestion
            result, ocrm_line, mid = self.parse_log(
                self.gateway, "controller",
                r'.+OPERATING_CHANNEL_REPORT_MESSAGE,\smid=(\d+)',
                ocrm_line)
```
However, that does *not* improve readability, so let's stick with the noqa after all.

tests/test_flows.py
VladyslavTupikin: In the updated implementation for check_log we should exactly write esult, ocrm_line, mid = self.check_log, because it returns bool int string, if we skip result we will get bool int, skipping string.

tests/test_flows.py
```diff
@@ -220,30 +220,35 @@ def init(self):
...
+            for (i, v) in enumerate(logfile.readlines()):
+                if i <= start_line:
+                    continue
+                search = re.search(regex, v)
+                if search:
                     debug("Found '{}'\n\tin {}".format(regex, logfilename))
-                    return True
-        return False
+                    return (True, i, search.groups())
+        return (False, start_line, None)
```
arnout, line 234: In general, you don't know if there will be groups in the match object. For full flexibility, just return the match object. That also removes the need to have a True, False as first return value, because the None already functions as a truth value.

Also, I think it makes much more sense to return the last line in the failure case, because that allows us to continue searching later.

In other words:

```suggestion
                    return i, search
        return i, None
```

That said, since currently nothing is actually using the return value, it's fine as it is for now.

tests/test_flows.py
arnout: This comment is still not taken into account.

tests/test_flows.py
```diff
@@ -436,36 +436,83 @@ def test_channel_selection(self):
...
+                when we don't send channel report when there is no need to
+                switch channel
+        """
+        for i in range(1, 3):
+            debug("Send channel selection request, step {}".format(i))
+            mid = self.gateway_ucc.dev_send_1905(
+                    self.mac_repeater1,
+                    0x8006,
+                    tlv(0x8B, 0x005F, '{} {}'.format(self.mac_repeater1_wlan0, payload_wlan0)),
+                    tlv(0x8D, 0x0007, '{} 0x{:2x}'.format(self.mac_repeater1_wlan0, tp20dBm)),
```
arnout, line 539: This must be 
```suggestion
                    tlv(0x8D, 0x0007, '{} 0x{:02x}'.format(self.mac_repeater1_wlan0, tp20dBm)),
```

It happens to work now because we use values > 16 so it already is two digits, but if it weren't, the leading 0 is really essential.

tests/test_flows.py
```diff
@@ -436,36 +436,83 @@ def test_channel_selection(self):
...
+                    tlv(0x8B, 0x004C, '{} {}'.format(self.mac_repeater1_wlan0, payload_wlan2)),
+                    tlv(0x8D, 0x0007, '{} 0x{:2x}'.format(self.mac_repeater1_wlan2, tp20dBm))
             )
-        time.sleep(1)
+            time.sleep(1)
+
+            debug(
+                "Confirming channel selection request has been received on controller,\
+                    step {}".format(i))
```
arnout, line 547: This line splitting is really ugly... Better keep it on one line, if needed with a `  # noqa E501`

(nitpick)

tests/test_flows.py
```diff
@@ -436,36 +436,83 @@ def test_channel_selection(self):
...
+
+            self.check_log(self.repeater1, "agent_wlan0", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+            self.check_log(self.repeater1, "agent_wlan2", "CHANNEL_SELECTION_REQUEST_MESSAGE")
+
+            debug(
+                "Confirming tlvTransmitPowerLimit has been received with correct value on \
+                    agent, step {}".format(i))
+
+            self.check_log(
+                self.repeater1, "agent_wlan0", "tlvTransmitPowerLimit {}".format(tp20dBm))
```
arnout, line 565: Here as well: we prefer line splitting like this.
```suggestion
            self.check_log(self.repeater1, 
                           "agent_wlan0", "tlvTransmitPowerLimit {}".format(tp20dBm))
```

arnout (approved):
The summary line of "test_flow.py: check_log: Modify check log function" is still not OK: "modify" doesn't say what you're doing. Should be something like: "test_flows.py: check_log: use start_line and return match".

I still have a number of comments, but nothing that can't be fixed later. Since this will be conflicting with other PRs, let's merge it!

tests/test_flows.py
VladyslavTupikin: In some places, we still can get lines greater than 100 symbols. And the solution - transfer all parameters to the new string including first.

PR #968, opened 2020-03-13
Convert test_flows to python - final

arnout:
This converts the rest of the tests. It's a one-to-one conversion, not even adding newlines for proper line width.


tests/test_flows.py
```diff
@@ -244,23 +244,42 @@ def init(self):
...
+                    self.debug("Found '{}'\n\tin {}".format(regex, logfilename))
+                    return True
+        return False
+
     def check_log(self, device: str, program: str, regex: str) -> bool:
         '''Verify that on "device" the logfile for "program" matches "regex", fail if not.'''
-        logfilename = os.path.join(self.rootdir, 'logs', device, 'beerocks_{}.log'.format(program))
         try:
-            # HACK check_log is often used immediately after triggering a message on the other side.
+            # HACK self.check_log(is, "often", rused immediately after triggering a message on the other side.)
```
ghost: This seems to be a find/replace job gone wrong `s/check_log (\S+) (\S+) (.*)/self.check_log(\1, "\2", r\3)/g`

tests/test_flows.py
```diff
@@ -244,23 +244,42 @@ def init(self):
...
-                        self.debug("Found '{}'\n\tin {}".format(regex, logfilename))
-                        return True
+            if self._check_log_internal(device, program, regex):
+                return True
+            else:
+                return self.fail("'{}'\n\tin log of {} on {}".format(regex, program, device))
+        except OSError:
+            return self.fail("Can't read log of {} on {}".format(program, device))
+
+    def wait_for_log(self, device: str, program: str, regex: str, timeout: float) -> bool:
```
ghost, line 272: there's a couple of things I dislike here:
- polling a file every second
- doing so inefficiently (by constantly closing and re-opening the file, instead of just reading what's new since you've last read it).

I know that's the easiest approach, but I'd be tempted to rewrite it using asynchronous IO (O_ASYNC and check for EAGAIN, or whatever that translates to in python)
Or just use `poll` (or `select`, but `poll` is better), where you can also pass a timeout.
Then again, it's probably not going to be a real issue, right?

ghost: Actually I don't think that would work, since it's a different process writing the file. There would most likely not be an update on your file descriptor if a different process writes to the file (I think?)
You could still use something like `inotify`, but that's probably overkill for this

ghost (approved):
No other issues found, so approving.

ghost:
https://gitlab.com/prpl-foundation/prplMesh/-/jobs/470873622

client_steering_dummy test fails on gitlab

tomereli (approved):

tests/test_flows.py
arnout: As noted in the commit message, it's a pretty literal translation to Python.

And anyway I hope that most of the check_log (and wait_for_log) calls can be replaced with a status event. With the shell script, we have to use files because sockets and pipes are tricky in shell. But with python we have all the power we need.

PR #970, opened 2020-03-13
beerocks_ucc_listener.cpp: Controller drops first CHANNEL_SELECTION...

### LINKED ISSUE #969: [BUG] Controller drops first CHANNEL_SELECTION_REQUEST_MESSAGE
## Steps to reproduce

1. Build
2. Run controller & agent
3. Send `CHANNEL_SELECTION_REQUEST_MESSAGE` with payload
4. Check agent logs. No reaction to message sent
5. Send `CHANNEL_SELECTION_REQUEST_MESSAGE` with payload
6. Check agent logs. Channel is set according to request, reaction according to MAP-4.2.1 Cert. test

## Expected behaviour

Controller and agent reacts to the first `CHANNEL_SELECTION_REQUEST_MESSAGE` as defined in MAP-4.2.1 Cert. test

## Target

- Host: [Docker Ubuntu 18.04]
- Platform: [Linux DUMMY mode with ZMQ]
- How you built prplMesh: prplMesh$ ./docker-builder.sh
- How you ran prplMesh: prplMesh$ ./tests/test_flows.py

## Additional context
This issue was originally discussed on Slack (#prplmesh-dev) and found during #926 .
Brief about current state from @tomereli :
> current implementation tailored for passing test 5.5.1 which requires the MCUT to send channel selection request message in step 5 which is based on the information received from the agent.
> This is why currently, in  master_thread::handle_cmdu_1905_channel_preference_report we prepare the cert_cmdu_tx with CHANNEL_SELECTION_REQUEST_MESSAGE .
> Later on, the UCC (step 5 in test 5.5.1) sends a channel selection request message without TLV data, and the prplMesh controller sends the prefilled cmdu.
> The test Nikolai Merinov and his team are writing sends channel selection request with TLV data which is ignored.
> The following solution will solve this:
> if cmdu_cert_tx for a specific message type is prefilled, and UCC send_1905_cmdu with TLV data is received, ignore the prefilled CMDU and send only the one with TLV data.
> Then the only question is if we should drop the prefilled cmdu.
Root of issue is located in `master_thread::handle_cmdu_1905_channel_preference_report ` symbol. Exact condition which is omitted when command send for the first time located at 699 line.

## Proposed resolution
From @tomereli :

> But before that I would create an issue to solve it as first step like we discussed yesterday Anton Bilohai - UCC Listener simply sends the CMDU if no tlv data, and builds a cmdu and ignores the certification cmdu if it gets tlv data.
> The original issue is that the UCC Listener ignores tlv data if exists in the capi command if it already has a saved certification cmdu

## Acceptance criteria

Controller properly reacts to first `CHANNEL_SELECTION_REQUEST_MESSAGE`

VladyslavTupikin:
The channel selection test flow uses a fully specified Channel
Selection Request message. However, current implemenation has a
predefined channel selection buffer with empty tlv paylod and
sends it instead preparing new CMDU message with full tlv payload.

For solving issue added one more check in CMDU mesage type condition from
case eWfaCaCommand::DEV_SEND_1905 - checking tlv_hex_list variable it should
be empty for sending predefined message. If tlv_hex_list variable has a data
clear CMDU buffer for erasing predefined message, create and send new CMDU
message.

Signed-off-by: Vladyslav Tupikin <v.tupikin@inango-systems.com>

### DIFF
--- a/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
+++ b/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
@@ -696,7 +696,7 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
         }
 
         // Check if CMDU message type has preprepared CMDU which can be loaded
-        if (m_cmdu_tx.is_finalized() &&
+        if (tlv_hex_list.empty() && m_cmdu_tx.is_finalized() &&
             m_cmdu_tx.getMessageType() == ieee1905_1::eMessageType(message_type)) {
             m_cmdu_tx.setMessageId(g_mid); // force mid
             LOG(DEBUG) << "Send preset cmdu with mid " << std::hex << g_mid;
@@ -710,7 +710,7 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
         } else {
             // CMDU was not loaded from preprepared buffer, and need to be created manually, and
             // use TLV data from the command (if exists)
-
+            m_cmdu_tx.reset();
             if (!cmdu_tx.create(g_mid, ieee1905_1::eMessageType(message_type))) {
                 LOG(ERROR) << "cmdu creation of type 0x" << message_type_str << ", has failed";
                 reply_ucc(eWfaCaStatus::ERROR, err_internal);


tomereli (approved):

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -710,7 +711,7 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
         } else {
             // CMDU was not loaded from preprepared buffer, and need to be created manually, and
             // use TLV data from the command (if exists)
-
+            m_cmdu_tx.reset();
```
arnout, line 713: The commit message doesn't mention why this is needed.

arnout (approved):
The commit message could be a bit clearer about what the issue is. Something like "The channel selection test flow uses a fully specified Channel Selection Request message. However, if the certification buffer was already prepared with a Channel Selection Request message, the prepared one is ignored and the one specified by UCC is ignored."

The second paragraph of the commit message is very clear, no change needed there.

arnout:
BTW the clang-format job failed, so please run tools/docker/clang-format.sh to make sure it's formatted correctly.

abelog (approved):

Issue #973, opened 2020-03-16
[TASK] TLVF: Add copy methods to uint8 lists


LiorAmram:
## Description

Add auto-generated copy methods to uint8 lists in tlvf classes, just like copy methods for strings that exist today.
This is needed for uniformity of the code as well as convenience.

## Exit criteria

Passing build and current tests

tomereli:
Fixed by #974

PR #974, opened 2020-03-16
Feature/tlvf add copy methods to uint8 lists

LiorAmram:
Add copy methods to uint8_t lists on tlvf, applying this feature: https://github.com/prplfoundation/prplMesh/issues/973


tomereli (approved):

morantr (approved):
Approved, but you need to remove the merge commits and rebase to the master instead.
Added "don't merge" label so you could fix it before auto-merge will take action.

PR #977, opened 2020-03-16
Feature/add method to obtain radio information

### LINKED ISSUE #967: [TASK] Add NL80211_CMD_GET_WIPHY command
## Description

Add a new method to nl80211_client to obtain the information required to compute the MediaType value for IEEE802.11 interfaces. This field is used in several TLVs and described in Table 6-12—Media type (intfType) of IEEE 1905.1 document.

MediaType value must be computed from HT/VHT capabilities. The nl80211 socket command to obtain such information is NL80211_CMD_GET_WIPHY.

mariomaz:
Add method to obtain radio information by issuing command
NL80211_CMD_GET_WIPHY.

This method will later be used to compute MediaType field required by
several TLVs and to fill in the HT/VHT capabilities TLVs.

### DIFF
--- a/common/beerocks/bwl/dummy/nl80211_client_dummy.h
+++ b/common/beerocks/bwl/dummy/nl80211_client_dummy.h
@@ -32,30 +32,41 @@ public:
     virtual ~nl80211_client_dummy();
 
     /**
+     * @brief Gets radio information.
+     *
+     * Radio information contains HT/VHT capabilities and the list of supported channels.
+     *
+     * @param[in] interface_name Interface name, either radio or Virtual AP (VAP).
+     * @param[out] radio_info Radio information.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_radio_info(const std::string &interface_name, radio_info &radio_info) override;
+
+    /**
      * @brief Gets station information.
      *
      * Fills station information with dummy data.
      *
-     * @param[in] local_interface_name Virtual AP (VAP) interface name.
+     * @param[in] interface_name Virtual AP (VAP) interface name.
      * @param[in] sta_mac_address MAC address of a station connected to the local interface.
      * @param[out] sta_info Station information.
      *
      * @return Dummy implementation returns always true.
      */
-    virtual bool get_sta_info(const std::string &local_interface_name,
-                              const sMacAddr &sta_mac_address, sStaInfo &sta_info) override;
+    bool get_sta_info(const std::string &interface_name, const sMacAddr &sta_mac_address,
+                      sta_info &sta_info) override;
 
     /**
      * @brief Set the tx power limit
      *
      * Set tx power limit for a radio
      *
-     * @param[in] local_interface_name radio interface name.
-     * @param[in] limit tx power limit in dBM to set 
+     * @param[in] interface_name radio interface name.
+     * @param[in] limit tx power limit in dBm to set
      * @return true success and false otherwise
      */
-    virtual bool set_tx_power_limit(const std::string &local_interface_name,
-                                    uint32_t limit) override;
+    bool set_tx_power_limit(const std::string &interface_name, uint32_t limit) override;
 };
 
 } // namespace bwl
--- a/common/beerocks/bwl/include/bwl/nl80211_client.h
+++ b/common/beerocks/bwl/include/bwl/nl80211_client.h
@@ -8,9 +8,31 @@
 #ifndef __BWL_NL80211_CLIENT_H__
 #define __BWL_NL80211_CLIENT_H__
 
+#include <bcl/beerocks_defines.h>
+#include <bcl/beerocks_message_structs.h>
+#include <bcl/son/son_wireless_utils.h>
 #include <tlvf/common/sMacAddr.h>
 
+#include <algorithm>
 #include <string>
+#include <unordered_map>
+#include <vector>
+
+/**
+ * @brief Length of HT MCS set.
+ *
+ * According to <linux/nl80211.h>, NL80211_BAND_ATTR_HT_MCS_SET is a 16-byte attribute containing
+ * the MCS set as defined in 802.11n
+ */
+static constexpr size_t ht_mcs_set_size = 16;
+
+/**
+ * @brief Length of VHT MCS set.
+ *
+ * According to <linux/nl80211.h>, NL80211_BAND_ATTR_VHT_MCS_SET is a 32-byte attribute containing
+ * the MCS set as defined in 802.11ac
+ */
+static constexpr size_t vht_mcs_set_size = 32;
 
 namespace bwl {
 
@@ -34,12 +56,144 @@ class nl80211_client {
 
 public:
     /**
+     * @brief Channel information
+     *
+     * Used in band_info structure as each band contains a map of supported channels.
+     *
+     * Information obtained with NL80211_CMD_GET_WIPHY command through a NL80211 socket.
+     * See NL80211_FREQUENCY_ATTR_* in <linux/nl80211.h> for a description of each field.
+     */
+    struct channel_info {
+
+        /**
+         * Channel number.
+         * Obtained from NL80211_FREQUENCY_ATTR_FREQ (see 802.11-2007 17.3.8.3.2 and Annex J)
+         */
+        uint8_t number = 0;
+
+        /**
+         * Supported channel bandwidths.
+         */
+        std::vector<beerocks::eWiFiBandwidth> supported_bandwidths;
+
+        /**
+         * Maximum transmission power in dBm.
+         * Obtained as 0.01 * NL80211_FREQUENCY_ATTR_MAX_TX_POWER
+         */
+        uint8_t tx_power = 0;
+
+        /**
+         * Radar detection is mandatory on this channel in current regulatory domain.
+         * Set to true if attribute NL80211_FREQUENCY_ATTR_RADAR is present.
+         */
+        bool is_dfs = false;
+
+        /**
+         * True if current state for DFS is NL80211_DFS_UNAVAILABLE.
+         * Obtained from NL80211_FREQUENCY_ATTR_DFS_STATE
+         */
+        bool radar_affected = false;
+
+        /**
+         * @brief Gets the maximum supported bandwidth by the channel.
+         *
+         * @return Maximum supported bandwidth
+         */
+        beerocks::eWiFiBandwidth get_max_bandwidth() const
+        {
+            if (supported_bandwidths.empty()) {
+                return beerocks::eWiFiBandwidth::BANDWIDTH_UNKNOWN;
+            }
+
+            return *std::max_element(supported_bandwidths.begin(), supported_bandwidths.end());
+        }
+    };
+
+    /**
+     * @brief Band information
+     *
+     * Used in radio_info structure as each radio contains a list of bands.
+     *
+     * Information obtained with NL80211_CMD_GET_WIPHY command through a NL80211 socket.
+     * See NL80211_BAND_ATTR_* in <linux/nl80211.h> for a description of each field.
+     */
+    struct band_info {
+
+        /**
+         * Value of NL80211_BAND_ATTR_HT_CAPA, see iw/util.c print_ht_capability() as a bit
+         * interpretation example.
+         */
+        uint16_t ht_capability = 0;
+
+        /**
+         * Value of NL80211_BAND_ATTR_HT_MCS_SET, see iw/util.c print_ht_mcs() as a byte
+         * interpretation example.
+         */
+        uint8_t ht_mcs_set[ht_mcs_set_size];
+
+        /**
+         * Value of NL80211_BAND_ATTR_VHT_CAPA, see iw/util.c print_vht_info() as a bit
+         * interpretation example.
+         */
+        uint32_t vht_capability = 0;
+
+        /**
+         * Value of NL80211_BAND_ATTR_VHT_MCS_SET, see iw/util.c print_vht_info() as a byte
+         * interpretation example.
+         */
+        uint8_t vht_mcs_set[vht_mcs_set_size];
+
+        /**
+         * Channels supported in this band (obtained from NL80211_BAND_ATTR_FREQS).
+         * Map key is the channel number and map value is the channel information.
+         */
+        std::unordered_map<uint8_t, channel_info> supported_channels;
+
+        /**
+         * @brief Gets the frequency band of this band.
+         *
+         * Return value can be obtained from whatever of the supported channels.
+         *
+         * @return Frequency band
+         */
+        beerocks::eFreqType get_frequency_band() const
+        {
+            if (supported_channels.empty()) {
+                return beerocks::eFreqType::FREQ_UNKNOWN;
+            }
+
+            return son::wireless_utils::which_freq(supported_channels.begin()->first);
+        }
+
+        /**
+         * @brief Checks if this band is the 5GHz band.
+         *
+         * @return True if this band is the 5GHz band and false otherwise.
+         */
+        bool is_5ghz_band() const { return beerocks::eFreqType::FREQ_5G == get_frequency_band(); }
+    };
+
+    /**
+     * @brief Radio information
+     *
+     * Information obtained with NL80211_CMD_GET_WIPHY command through a NL80211 socket.
+     * See <linux/nl80211.h> for a description of each field.
+     */
+    struct radio_info {
+
+        /**
+         * Bands included in this radio (obtained from NL80211_ATTR_WIPHY_BANDS)
+         */
+        std::vector<band_info> bands;
+    };
+
+    /**
      * @brief Station information
      *
      * Information obtained with NL80211_CMD_GET_STATION command through a NL80211 socket.
      * See NL80211_STA_INFO_* in <linux/nl80211.h> for a description of each field.
      */
-    struct sStaInfo {
+    struct sta_info {
         uint32_t inactive_time_ms   = 0;
         uint32_t rx_bytes           = 0;
         uint32_t rx_packets         = 0;
@@ -47,8 +201,8 @@ public:
         uint32_t tx_packets         = 0;
         uint32_t tx_retries         = 0;
         uint32_t tx_failed          = 0;
-        uint8_t signal_dBm          = 0;
-        uint8_t signal_avg_dBm      = 0;
+        uint8_t signal_dbm          = 0;
+        uint8_t signal_avg_dbm      = 0;
         uint16_t tx_bitrate_100kbps = 0;
         uint16_t rx_bitrate_100kbps = 0;
     };
@@ -59,31 +213,42 @@ public:
     virtual ~nl80211_client() = default;
 
     /**
+     * @brief Gets radio information.
+     *
+     * Radio information contains HT/VHT capabilities and the list of supported channels.
+     *
+     * @param[in] interface_name Interface name, either radio or Virtual AP (VAP).
+     * @param[out] radio_info Radio information.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_radio_info(const std::string &interface_name, radio_info &radio_info) = 0;
+
+    /**
      * @brief Gets station information.
      *
      * Station information contains basically metrics associated to the link between given local
      * interface and a the interface of a station whose MAC address is 'sta_mac_address'.
      *
-     * @param[in] local_interface_name Virtual AP (VAP) interface name.
+     * @param[in] interface_name Virtual AP (VAP) interface name.
      * @param[in] sta_mac_address MAC address of a station connected to the local interface.
      * @param[out] sta_info Station information.
      *
      * @return True on success and false otherwise.
      */
-    virtual bool get_sta_info(const std::string &local_interface_name,
-                              const sMacAddr &sta_mac_address, sStaInfo &sta_info) = 0;
+    virtual bool get_sta_info(const std::string &interface_name, const sMacAddr &sta_mac_address,
+                              sta_info &sta_info) = 0;
+
     /**
      * @brief Set the tx power limit
      *
      * Set tx power limit for a radio
      *
-     * @param[in] local_interface_name radio interface name.
-     * @param[in] limit tx power limit in dBM to set 
+     * @param[in] interface_name radio interface name.
+     * @param[in] limit tx power limit in dBm to set
      * @return true success and false otherwise
      */
-    virtual bool set_tx_power_limit(const std::string &local_interface_name, uint32_t limit) = 0;
-
-    // TODO: add remaining methods
+    virtual bool set_tx_power_limit(const std::string &interface_name, uint32_t limit) = 0;
 };
 
 } // namespace bwl
--- a/common/beerocks/bwl/shared/nl80211_client_impl.cpp
+++ b/common/beerocks/bwl/shared/nl80211_client_impl.cpp
@@ -6,7 +6,32 @@
  * See LICENSE file for more details.
  */
 
+/**
+ * This file uses code copied from `iw` (http://git.sipsolutions.net/iw.git/)
+ *
+ * Copyright (c) 2007, 2008 Johannes Berg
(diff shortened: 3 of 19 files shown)


common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -33,6 +51,116 @@ namespace bwl {
...
 public:
+    /**
+     * @brief Channel information
+     *
+     * Used in sBandInfo structure as each band contains a list of supported channels.
+     *
+     * Information obtained with NL80211_CMD_GET_WIPHY command through a NL80211 socket.
+     * See NL80211_FREQUENCY_ATTR_* in <linux/nl80211.h> for a description of each field.
+     */
+    struct sChannelInfo {
```
arnout: Is it really necessary/useful to copy the sWifiChannel structure?

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -33,6 +51,116 @@ namespace bwl {
...
+
+        /**
+         * Channel number.
+         * Obtained from NL80211_FREQUENCY_ATTR_FREQ (see 802.11-2007 17.3.8.3.2 and Annex J)
+         */
+        uint16_t channel = 0;
+
+        /**
+         * Maximum channel bandwidth in MHz.
+         * Possible values are 0 (unknown), 20, 40, 80 and 160 MHz.
```
arnout: Do you know how nl80211 represents 80+80?

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
 
+/**
+ * @brief Obtains the channel number that corresponds to given frequency value.
+ *
+ * This function has been borrowed from iw/util.c
+ *
+ * @param freq frequency value in MHz.
+ * @return channel number.
+ */
+static uint16_t ieee80211_frequency_to_channel(uint32_t freq)
```
arnout: I think it's better to move beerocks::utils::wifi_freq_to_channel() to wireless_utils and use that.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
 
 #include <math.h>
 
+/**
+ * @brief Obtains the channel number that corresponds to given frequency value.
+ *
+ * This function has been borrowed from iw/util.c
```
arnout: When copying from somewhere else, please also copy the license blurb and copyright.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
 namespace bwl {
 
 nl80211_client_impl::nl80211_client_impl(std::unique_ptr<nl80211_socket> socket)
     : m_socket(std::move(socket))
 {
 }
 
+bool nl80211_client_impl::get_radio_info(const std::string &interface_name, sRadioInfo &radio_info)
+{
+    static struct nla_policy stats_policy[NL80211_STA_INFO_MAX + 1];
```
arnout: This isn't used. Cut&paste from update_stations_stats? :-)

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+    stats_policy[NL80211_STA_INFO_TX_BYTES]      = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_PACKETS]    = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_RETRIES]    = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_FAILED]     = {NLA_U32, 0, 0};
+    stats_policy[NL80211_STA_INFO_SIGNAL]        = {NLA_U8, 0, 0};
+    stats_policy[NL80211_STA_INFO_SIGNAL_AVG]    = {NLA_U8, 0, 0};
+    stats_policy[NL80211_STA_INFO_TX_BITRATE]    = {NLA_NESTED, 0, 0};
+    stats_policy[NL80211_STA_INFO_RX_BITRATE]    = {NLA_NESTED, 0, 0};
+
+    static struct nla_policy rate_policy[NL80211_RATE_INFO_MAX + 1];
```
arnout: Also unused

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+    rate_policy[NL80211_RATE_INFO_BITRATE]   = {NLA_U16, 0, 0};
+    rate_policy[NL80211_RATE_INFO_BITRATE32] = {NLA_U32, 0, 0};
+
+    struct channels_ctx {
+        int last_band;
+        bool width_40;
+        bool width_80;
+        bool width_160;
+    } ctx{};
```
arnout: (nitpick) why make this a struct? We can just as well use plain local variables in the closure.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+                        band.ht_capability = nla_get_u16(tb_band[NL80211_BAND_ATTR_HT_CAPA]);
+
+                        if (band.ht_capability & BIT(1)) {
+                            ctx.width_40 = true;
+                        }
+                    } else {
+                        LOG(DEBUG) << "NL80211_BAND_ATTR_HT_CAPA attribute is missing";
+                    }
+
+                    if (tb_band[NL80211_BAND_ATTR_HT_MCS_SET]) {
```
arnout, line 129: (nitpick) this is only useful/valid if HT_CAPA was set, and it's an ERROR if HT_CAPA was set and HT_MCS is not set.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+                        } else {
+                            LOG(DEBUG) << "Invalid length for NL80211_BAND_ATTR_VHT_MCS_SET "
+                                          "attribute. Expected length: "
+                                       << expected_length << ", actual length: " << actual_length;
+                        }
+                    } else {
+                        LOG(DEBUG) << "NL80211_BAND_ATTR_VHT_MCS_SET attribute is missing";
+                    }
+
+                    if (tb_band[NL80211_BAND_ATTR_FREQS]) {
```
arnout: I think we can treat it as an error (and return) if this is missing. That saves us some indentation depth below.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+
+                            if (!tb_freq[NL80211_FREQUENCY_ATTR_FREQ] ||
+                                tb_freq[NL80211_FREQUENCY_ATTR_DISABLED] ||
+                                (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR] &&
+                                 tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) ||
+                                tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
+                                continue;
+                            }
+
+                            if (tb_band[NL80211_FREQUENCY_ATTR_FREQ]) {
```
arnout:
```suggestion
                            if (tb_freq[NL80211_FREQUENCY_ATTR_FREQ]) {
```

However, the previous condition just checked that, so this is redundant.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+                                LOG(ERROR)
+                                    << "Failed to parse frequency " << nl_freq->nla_type + 1 << "!";
+                                return false;
+                            }
+
+                            if (!tb_freq[NL80211_FREQUENCY_ATTR_FREQ] ||
+                                tb_freq[NL80211_FREQUENCY_ATTR_DISABLED] ||
+                                (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR] &&
+                                 tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) ||
+                                tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
```
arnout: `(a && b) || b` is equivalent to just `b`. In other words:
```suggestion
                                tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
```

But maybe I missed something.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+                                uint32_t freq   = nla_get_u32(tb_freq[NL80211_FREQUENCY_ATTR_FREQ]);
+                                channel.channel = ieee80211_frequency_to_channel(freq);
+                            } else {
+                                LOG(DEBUG) << "NL80211_FREQUENCY_ATTR_FREQ attribute is missing";
+                            }
+
+                            if (tb_freq[NL80211_FREQUENCY_ATTR_MAX_TX_POWER]) {
+                                channel.tx_pow_dBm =
+                                    0.01 *
+                                    nla_get_u32(tb_freq[NL80211_FREQUENCY_ATTR_MAX_TX_POWER]);
```
arnout: (nitpick) This way you convert to float and then back to int, so in the end you could just as well do
```suggestion
                                    nla_get_u32(tb_freq[NL80211_FREQUENCY_ATTR_MAX_TX_POWER]) / 100;
```

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+                                    << "NL80211_FREQUENCY_ATTR_MAX_TX_POWER attribute is missing";
+                            }
+
+                            if (tb_freq[NL80211_FREQUENCY_ATTR_RADAR]) {
+                                channel.is_dfs = true;
+                            }
+
+                            if (!tb_freq[NL80211_FREQUENCY_ATTR_NO_20MHZ]) {
+                                channel.bandwidth_MHz = 20;
+                            }
```
arnout: This will only be set for channels that are limited to 5 or 10MHz, which is probably not something we will ever want to use. So I think we can safely skip a channel that has NO_20MHZ set.

If not, we'll have to initialize bandwidth_MHz with a non-zero value, or it will *sometimes* be ignored...

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -17,13 +21,276 @@
...
+                            if (ctx.width_40 && (!tb_freq[NL80211_FREQUENCY_ATTR_NO_HT40_MINUS] ||
+                                                 !tb_freq[NL80211_FREQUENCY_ATTR_NO_HT40_PLUS])) {
+                                channel.bandwidth_MHz = 40;
+                            }
+                            if (ctx.width_80 && !tb_freq[NL80211_FREQUENCY_ATTR_NO_80MHZ]) {
+                                channel.bandwidth_MHz = 80;
+                            }
+                            if (ctx.width_160 && !tb_freq[NL80211_FREQUENCY_ATTR_NO_160MHZ]) {
+                                channel.bandwidth_MHz = 160;
+                            }
```
arnout: I'm not sure how this will be used, but I believe we in the end will need a sWifiChannel for each separate bandwidth at which we can operate. At least, that's the assumption when constructing the EasyMesh messages (AP Radio Basic Capabilities). @kantera800 @YuriyMasechko @vitalybu please comment.

In practice, of course, the maximum bandwidth is sufficient since we can (probably) always operate at a lower bandwidth.

arnout (approved):
As often, some major comments but I don't think there's anything that can't be easily fixed.

Also as usual for @mariomaz: great documentation!

common/beerocks/bwl/include/bwl/nl80211_client.h
tomereli: I don't see why, I think we should use the same structure here.
Same for the sBandInfo I guess.

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -33,6 +51,116 @@ namespace bwl {
...
+         */
+        uint16_t bandwidth_MHz = 0;
```
tomereli: Please don't mix camel case and lower snake case. The comment is enough actually so its fine to just call it `bandwidth`
```suggestion
        uint16_t bandwidth = 0;
```

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -33,6 +51,116 @@ namespace bwl {
...
+
+        /**
+         * Maximum transmission power in dBm.
+         * Obtained as 0.01 * NL80211_FREQUENCY_ATTR_MAX_TX_POWER
+         */
+        uint32_t tx_pow_dBm = 0;
```
tomereli:
```suggestion
        uint32_t tx_power = 0;
```

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -33,6 +51,116 @@ namespace bwl {
...
+        beerocks::eWiFiBandwidth max_bandwidth = beerocks::eWiFiBandwidth::BANDWIDTH_20;
+    };
+
+    /**
+     * @brief Radio information
+     *
+     * Information obtained with NL80211_CMD_GET_WIPHY command through a NL80211 socket.
+     * See <linux/nl80211.h> for a description of each field.
+     */
+    struct sRadioInfo {
```
tomereli: There is already a radio info struct in `base_wlan_hal_types.h` - can we extend it and use it here instead?

tomereli (approved):
small nitpicks, looks good

tomereli:
It has just occured to me that we desperately also need `NL80211_CMD_SET_WIPHY` support to set tx_power limit`. Please add a task for it:)

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Replaced sChannelInfo with sWifiChannel

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: No idea about WiFi. 
From code borrowed from `iw`:
```c
                        switch ((band.vht_capability >> 2) & 3) {
                        case 2:
                            /* width_80p80 = true; */
                            /* fall through */
                        case 1:
                            ctx.width_160 = true;
                            break;
                        }
```

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: There is a FIXME in `beerocks::utils::wifi_freq_to_channel()` that I think has been fixed in `ieee80211_frequency_to_channel` borrowed from `iw`. @arnout / @tomereli  Can you please confirm which of two versions should I move to wireless_utils?

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Removed

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Removed

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Long story. Code borrowed from `iw` creates a variable of type channels_ctx and passes it around because response can be split into several messages

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Changed to use NL80211_BAND_ATTR_HT_MCS_SET only if NL80211_BAND_ATTR_HT_CAPA was set and ignored otherwise

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Fixed

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Fixed

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Supported channels in this new implementation are obtained the same way as in previous implementation (which I guess will soon be replaced by this one).
I've also seen that nl80211 command returns a nested array of operating bands in `NL80211_ATTR_WIPHY_BANDS`. However, previous implementation considers just one band, the first one, and ignores the rest

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Ok

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Ok

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: There is a radio info struct in `base_wlan_hal_types.h` and another one in `beerocks_message_structs.h`. @arnout / @tomereli  Which one should I use?
The same happens with the WiFi channel struct.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
arnout: The version you borrowed from `iw` is obviously better, but I still think there should be only one. So two commits are needed: one to move the function to wireless_utils, and a second one to fix it to correspond to how iw does it.

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: According to what has been agreed on meeting, I've recovered and used sChannelInfo. 
Now there are 3 different structures containing channel information: in nl80211_client.h, in beerocks_message_structs.h and in base_wlan_hal_types.h. This is something to be fixed in a future PR

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Changed to support 80+80 MHz channels

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Modified as it was agreed during meeting: each channel includes a vector of supported_bandwidths.
I've also added a method to get maximum bandwidth supported

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Done as agreed during the meeting: use new sRadioInfo structure

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -33,6 +55,138 @@ namespace bwl {
...
 public:
+    /**
+     * @brief Channel information
+     *
+     * Used in sBandInfo structure as each band contains a map of supported channels.
+     *
+     * Information obtained with NL80211_CMD_GET_WIPHY command through a NL80211 socket.
+     * See NL80211_FREQUENCY_ATTR_* in <linux/nl80211.h> for a description of each field.
+     */
+    struct sChannelInfo {
```
arnout: I realize I'm royally late with giving this feedback, but according to our CODING_STYLE.md, we use lower snake case everywhere (except for cases of historical accident, like sMacAddress). So this should be `struct channel_info`.

(nitpick)

common/beerocks/bwl/dummy/nl80211_client_dummy.h
```diff
@@ -31,6 +31,18 @@ class nl80211_client_dummy : public nl80211_client {
...
+     * @brief Gets radio information.
+     *
+     * Radio information contains HT/VHT capabilities and the list of supported channels.
+     *
+     * @param[in] interface_name Interface name, either radio or Virtual AP (VAP).
+     * @param[out] radio_info Radio information.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_radio_info(const std::string &interface_name, sRadioInfo &radio_info) override;
```
rmelotte: No need for virtual when there is already override (nitpick)

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -16,14 +41,236 @@
...
+    return m_socket.get()->send_receive_msg(
+        NL80211_CMD_GET_WIPHY, 0,
+        [&](struct nl_msg *msg) -> bool {
+            nla_put_u32(msg, NL80211_ATTR_IFINDEX, iface_index);
+
+            return true;
+        },
+        [&](struct nl_msg *msg) -> bool {
+            struct nlattr *tb[NL80211_ATTR_MAX + 1];
+            struct genlmsghdr *gnlh = (struct genlmsghdr *)nlmsg_data(nlmsg_hdr(msg));
```
rmelotte: No C-style cast, please use a static_cast if you still find the time/motivation to change it

common/beerocks/bwl/shared/nl80211_client_impl.h
```diff
@@ -42,6 +42,18 @@ class nl80211_client_impl : public nl80211_client {
...
+     * @brief Gets radio information.
+     *
+     * Radio information contains HT/VHT capabilities and the list of supported channels.
+     *
+     * @param[in] interface_name Interface name, either radio or Virtual AP (VAP).
+     * @param[out] radio_info Radio information.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_radio_info(const std::string &interface_name, sRadioInfo &radio_info) override;
```
rmelotte: Same comment, no need for virtual here (there are a lot of places where it's done like this, but also others where we drop the override)

rmelotte (approved):

common/beerocks/bwl/shared/nl80211_client_impl.h
arnout: I think we didn't come to a conclusion about it in the cppcheck PR. Vote in #democracy?

common/beerocks/bwl/include/bwl/nl80211_client.h

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/dummy/nl80211_client_dummy.h
mariomaz: Using `virtual` is an old habit from the times before C++11, when there was no `override`. It wasn't necessary either, but it was considered good practice to advertise programmers that a method was `virtual` without having to check the base class definition. I've also heard that some compilers might complain if `virtual` is not used in derived class, but that's not our case. 

Since `virtual` is being used together with `override` all over the code and it doesn't hurt, I'm going to leave it here as it is now. If we ever vote it in #democracy, I'll say do not use `virtual` if there is `override`. But I'll also say the use of `override` should be mandatory to avoid pitfalls.

PR #978, opened 2020-03-16
tools: add download_ipk.sh

tomereli:
Add a helper script which can be used to download latest ipks from a
specific branch in gitlab.

This will be used by a cron job in each testbed setup not connected to gitlab CI which will allow for a simple script to deploy latest prplmesh master IPK to the RAX40.

This can also be used for Broadfarm, or any other testing framework which uses real devices.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -124,7 +124,7 @@ run-tests:
     # kill prplmesh on each target to make sure they don't interfere with the test
     - for i in $ALL_TARGETS ; do ssh "$i" 'pgrep -f beerocks | xargs kill -9 2>/dev/null' || true ; done
   script:
-    - tools/deploy_ipk.sh $TARGET_DEVICE_NAME "build/$TARGET_DEVICE/"*.ipk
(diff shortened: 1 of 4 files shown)


ghost (approved):
Might be interesting to download only the ipk file rather than the whole archive.
For this, we would need a symlink to be made during the build (discussed during call).
However, I don't think this is a big issue, so approving now

arnout (approved):
Please also make sure a README.md or wiki page that explains how to use this script.

tomereli:
> Please also make sure a README.md or wiki page that explains how to use this script.

I don't think we need a wiki / README.md for every little script, especially when the script is well documented and provides -h option.

tomereli:
> Might be interesting to download only the ipk file rather than the whole archive.
> For this, we would need a symlink to be made during the build (discussed during call).
> However, I don't think this is a big issue, so approving now

Using symlink didn't work since the symlink was downloaded as text. I changed the ipk filename to `prplmesh.ipk` and added `prplmesh.buildinfo` which stores the opernwrt version, prplmesh version and target profile instead of embedding them in the ipk file name.

Issue #273, opened 2019-09-09
[TASK] Architecture for handling ACKs and retransmissions


arnout:
## Description

The Multi-AP specification defines the concept of acknowledgements and replies. Although no retransmission scheme is explicitly specified, the underlying assumption is that requests are retransmitted if there is no response within one second.

Currently, prplMesh does no retransmissions at all. This task is to define a scheme for retransmissions and implement it. Note that this is not required for certification.

## Exit criteria

- A document exists that describes the architecture for retransmissions.
- Acknowledgements and responses are handled somewhere.
- If no acknowledgement is received within one second, the request is retransmitted (with a new MID).
- If an acknowledgement is received after the retransmission, the acknowledgement is ignored.
- After a number of retransmissions, attempts are stopped.
- Some feedback mechanism is provided if we give up.
- A test exists for all the above functionality.

## Detailed description

Two basic approaches are possible.

1. Handle retransmissions centrally, i.e. in ieee1905_transport. This requires a way to indicate in the original transmission what the expected acknowledgement is. Probably we also need a way to abort retransmission.
2. Handle retransmissions in the end-points, i.e. in agent, controller, .... This makes it easier to handle failures in the place where we know how to handle them. It also avoids the need to send messages about failures. However, it requires repeating the one-second-timeout logic everywhere, so probably some abstraction in a common library is needed.

PR #979, opened 2020-03-17
[BUG] agent: autoconfig renew with broadcom controller

tomereli:
**Describe the bug** -
Currently, there is an issue in the testbed where all tests involving Broadcom controller sometimes fail on "MSG: No AP-Autoconfiguration WSC message with M2 found" after the autoconfig renew step in the 4.4.3 tests.

Reproduces 8/10 tries.

## Steps to reproduce

Run any of the 4.4.3 tests in the testbed which involve the Broadcom controller. For example, 4.4.3_ETH_FH5GL.

## Expected behavior

The test should pass 100% of the times.

## Target

PTK1 testbed

## Logs

https://ftp.essensium.com/owncloud/index.php/s/ketl3eFhIgwcweZ?path=%2Fprpl%20agent%20tests%2FFAIL%2FMAP-4.4.3_ETH_FH24G%202020-03-16%2FMAP-4.4.3_ETH_FH24G_Mar-16-2020__03-09-48

https://ftp.essensium.com/owncloud/index.php/s/ketl3eFhIgwcweZ?path=%2Fprpl%20agent%20tests%2FFAIL%2FMAP-4.4.3_ETH_FH5GL%202020-03-16%2FMAP-4.4.3_ETH_FH5GL_Mar-16-2020__03-15-06

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.h
+++ b/agent/src/beerocks/slave/son_slave_thread.h
@@ -156,7 +156,7 @@ private:
 private:
     const int SELECT_TIMEOUT_MSEC                                     = 200;
     const int SLAVE_INIT_DELAY_SEC                                    = 4;
-    const int WAIT_FOR_JOINED_RESPONSE_TIMEOUT_SEC                    = 30;
+    const int WAIT_FOR_JOINED_RESPONSE_TIMEOUT_SEC                    = 5;
     const int WAIT_BEFORE_SEND_SLAVE_JOINED_NOTIFICATION_SEC          = 1;
     const int WAIT_BEFORE_SEND_BH_ENABLE_NOTIFICATION_SEC             = 1;
     const int WAIT_FOR_PLATFORM_MANAGER_REGISTER_RESPONSE_TIMEOUT_SEC = 600;


tomereli:
## Analysis

After autoconfig renew is sent by the controller, prplMesh agent immediately sends M1 which goes without an answer. Then there is a 30 second timeout for the next try, for which there is an answer this time and the prplmesh agent completes the autoconfig successfully.
The problem is that by this time the test ends and fails since there is no M2 found from the controller.

## Solution

The current state timer for waiting for the autconfig reply is far too long - 30 seconds, so decrease this to 5 seconds. This means that if the prplmesh agent doesn't get an autoconfig response in 5 seconds, it will try again, instead of waiting 30 seconds.
With this, the test pass 100%.

arnout (approved):
Grmbl, again a certified controller that doesn't actually follow the spec. "If a Multi-AP Controller receives a WSC TLV containing an M1, then it shall respond within one second with.."

Given the one second deadline, we could actually set the timeout to 1 instead of 5. However, I think it is better to handle this in the transport layer (#273). When the transport does e.g. 3 retransmissions, then a 5 second higher-level timeout is perfect.

ghost (approved):
Looks good to me. Approved!

PR #980, opened 2020-03-17
documentation: test_flows.py test description

abelog:
There is no description of test which are done in test_flows.py.
So no one have confirmed info about test_flows.py -> Cert test mapping
and general overview of tests are missing.

General intention of this PR is to form up such document in form of `README.md`
in `tests`section. As i am collecting info on it for F2C presentation.

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,177 @@
+This directory contains the end-to-end tests of prplMesh
+
+## Description of tests
+
+### ap_config_renew
+
+- send 0x0006 + 0x0001 + 2x 0x0001 (1905 Topology Notification message) to Gateway
+- check log wlan0 received credentials SSID 1 bss_type 2
+- check log wlan0 received credentials SSID 2 bss_type 1
+- check log wlan2 tear down radio
+
+### ap_config_bss_tear_down
+- send 0x0006 (1905 Topology Notification message) to repeater 1 as Gateway
+- sleep 3
+- check log repeater 1 wlan0 received credentials SSID 3 bss_type 2
+- check log repeater 1 wlan2 tear down radio
+- send 0x0006 (1905 Topology Notification message) to repeater 1 as Gateway
+- check log repeater 1 wlan2 tear down radio
+
+### channel_selection
+- send 0x8004 (Channel Preference Query message) to repeater 1 as Gateway
+- check log repeater 1 wlan0 message CHANNEL_PREFERENCE_QUERY_MESSAGE received
+- check log repeater 1 wlan2 message CHANNEL_PREFERENCE_QUERY_MESSAGE received
+- send 0x8006 (Channel Selection Request message) to repeater 1 as Gateway
+- check log repeater 1 wlan0 message CHANNEL_SELECTION_REQUEST_MESSAGE received
+- check log repeater 1 wlan2 message CHANNEL_SELECTION_REQUEST_MESSAGE received
+
+### ap_capability_query
+- send 0x8001 (AP Capability Query message) to repeater 1 as Gateway
+- check log repeater 1 agent message AP_CAPABILITY_QUERY_MESSAGE received
+- check log controller AP_CAPABILITY_REPORT_MESSAGE received
+
+### combined_infra_metrics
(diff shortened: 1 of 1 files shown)


tests/README.md
```diff
@@ -0,0 +1,179 @@
...
+- send 0x8016 (Client Association Control Request message) to repeater1 wlan0 as Gateway
+- check logs repeater 1 wlan0 CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE
+- check logs repeater 1 wlan2 CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE
+- check logs controller ACK_MESSAGE
+
+### topology
+- send 0x0002 (Topology Query message) to repeater 1 as Gateway 
+- check log repeater 1 agent TOPOLOGY_QUERY_MESSAGE
+
+### ...
```
arnout: What is this?

tests/README.md
```diff
@@ -0,0 +1,179 @@
...
+- send 0x0002 (1905 Topology Query message) to repeater 1 as Gateway
+- check logs repeater 1 agent TOPOLOGY_QUERY_MESSAGE
+- send 0x8016 (Client Association Control Request message) to repeater1 wlan0 as Gateway
+- check logs repeater 1 wlan0 CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE
+- check logs repeater 1 wlan2 CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE
+- check logs controller ACK_MESSAGE
+
+### topology
+- send 0x0002 (Topology Query message) to repeater 1 as Gateway 
+- check log repeater 1 agent TOPOLOGY_QUERY_MESSAGE
```
arnout: I didn't review all of the above -- too boring. I am a bit unsure what value it adds, it seems pretty easy to get the same information by just looking at the source code...

tests/README.md
```diff
@@ -0,0 +1,179 @@
...
+### ...
+
+## Mapping of certification tests
+
+Flow tests between parenthesis don't fully cover the certification test.
+In addition, many certification tests include a configuration phase which is not covered in the flow test.
+
+| Certification test | test flow |
+| ------------------ | --------- |
+| 4.2.1 MAUT §5 Multi-AP Ethernet Onboarding and Initialization test | (channel_selection)
```
arnout: Maybe we should explicitly indicate the step(s) that are covered by a particular flow.

Maybe we should abbreviate the certification test description a little (can be done with a simple search-replace: `M[AC]UT §[^ ]* Multi-AP ` to nothing; `test |` -> `|`).

```suggestion
| 4.2.1 Ethernet Onboarding and Initialization | step 1-9: -<br>step 10: channel_selection<br>step 11-12: (topology) |
```

Note that (after #965 is merged) I think the channel selection test covers step 10 completely, so no parenthesis any more.

That said, I think we can merge it as is now and fill in details later. Just one or two examples that are worked out in detail to show the direction we want to go to.

arnout (changes requested):
I guess the idea is to squash everything in a single commit before applying? If not, I think you should. The individual commits have little merit and just pollute the history. If you plan on squashing commits, it's a good idea to make sure at least one commit has no Signed-off-by tag - that way, you're sure the merging is blocked until you actually squashed it, because the DCO check will fail. Also, to allow reviewers to validate the commit message as well as the change itself, it's convenient to:
- use fixup commits whenever you don't change the commit message;
- use squash commits with the full new commit message when you do change the commit message;
- (optionally) point out you did this in the PR comments.

Like for #965, I'm going to mark this as changes requested and re-review when the commit messages are updated.

tests/README.md
MuteSpirit: Removed

tests/README.md
MuteSpirit: Agree
But i suggest to not delete it right now and not broke document structure

tests/README.md
MuteSpirit: Apply your code suggestion
And add similar test steps cover description for 4.3.2

arnout (approved):
You'll also need to change the commit message after squashing the two commits. Everything after the first paragraph can be removed IMO. Oh, and keep two signoffs, both Anton's and Ivan's.

VladyslavTupikin (approved):

PR #981, opened 2020-03-17
[RFC] .mergify.yml: require "ready for merge" label

arnout:
Currently, a PR will be merged automatically as soon as it has two
reviews and passes CI. However, there have been a few cases where this
was a bit premature, i.e. something was still supposed to have changed.

Require an explicit "ready for merge" label to trigger automatic
merging.

Update CONTRIBUTING.md with this workflow.

For the time being, the "don't merge" label is kept as well.
If it turns out we're not using it any more, we can still remove it.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -3,6 +3,7 @@ pull_request_rules:
     conditions:
       - "#approved-reviews-by>=2" # Require at least two approving reviewers
       - "#changes-requested-reviews-by=0" # Changes requested blocks the merge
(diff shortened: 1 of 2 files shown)


tomereli (approved):

vitalybu (approved):

rmelotte (approved):

tomereli:
Why isn't this automatically merged?

arnout:
mergify doesn't merge PRs that edit .mergify.yml to avoid shooting yourself in the foot.

PR #982, opened 2020-03-17
bwl: nl80211_client: add set_tx_power_limit API

### LINKED ISSUE #983: [TASK] set_tx_power support
## Description

Add support for setting tx power limit in the agent per radio. This is not available in hostapd, so we need to use nl80211 for that.

### Steps

- [x] Add a new method to nl80211_client to set tx power limit using NL80211_CMD_SET_WIPHY.
- [x] Add a nl80211_client to the base_wlan_hal which will be used by all BWL backends to get and set WLAN nl80211 generic commands.
- [x] Move `set_tx_power_limit` API to ap_wlan_hal
- [x] Implement the stub in `set_tx_power_limit` to use the `nl80211_client` for setting tx_power limit
- [ ] Implement set_tx_power limit in dummy BWL implementation which will store the limit
- [ ] Enhance channel selection dummy test to set new tx power limit as done in the certification

Edit - the last two are a bit tricky with the current implementation since the `nl80211_client` doesn't have access to the `m_radio_info` which holds the tx power reading - this is read from hostapd using `radio_info` command which reports the tx power via the TxPower parameter, and is saved in the `m_radio_info` structure which belongs to the `base_wlan_hal` class.
So, unless there are better ideas I suggest not to implement this in dummy mode at all. We will have coverage for this in real devices test flows.

tomereli:
Some EasyMesh flows require setting tx power limit, which is not
supported via hostapd (tx_power is readable via radio_info command),
Add set_tx_power_limit() API to the nl80211 client implementation which
uses NL80211_CMD_SET_WIPHY to set the limit.

Currently doesn't work since set tx power is not supported by the driver, but this can be merged since we get the same status as when doing this with `iw` - so marking as ready for review.
**Tested on AXEPOINT3000 (UGW) and everything works as expected.**

Since this PR changes some channel selection handling, triggered gitlan CI pipeline to verify nothing broke, which tests the following tests and verifies autoconfig, onboarding and channel selection:
```
MAP-4.2.1
MAP-4.5.2_ETH_FH5GL
MAP-4.5.3_ETH_FH5GL
MAP-4.5.3_ETH_FH5GH
```
https://gitlab.com/prpl-foundation/prplMesh/pipelines/128634565

Fixes #983
 
### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -52,20 +52,6 @@ static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_
     }
 }
 
-static int8_t get_tx_power(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
-{
-    auto radio_channels = ap_wlan_hal->get_radio_info().supported_channels;
-    uint8_t channel     = ap_wlan_hal->get_radio_info().channel;
-
-    for (uint8_t i = 0;
-         i < beerocks::message::SUPPORTED_CHANNELS_LENGTH && i < radio_channels.size(); i++) {
-        if (radio_channels[i].channel == channel)
-            return radio_channels[i].tx_pow;
-    }
-
-    return 0;
-}
-
 static std::string
 get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
 {
@@ -997,7 +983,7 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
 void ap_manager_thread::fill_cs_params(beerocks_message::sApChannelSwitch &params)
 {
-    params.tx_power  = get_tx_power(ap_wlan_hal);
+    params.tx_power  = ap_wlan_hal->get_radio_info().conducted_power;
     params.channel   = ap_wlan_hal->get_radio_info().channel;
     params.bandwidth = uint8_t(
         beerocks::utils::convert_bandwidth_to_enum(ap_wlan_hal->get_radio_info().bandwidth));
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -582,8 +582,24 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
                    << beerocks::utils::convert_bandwidth_to_int(
                           (beerocks::eWiFiBandwidth)request->cs_params().bandwidth);
 
-        if (request->cs_params().channel == 0)
-            LOG(DEBUG) << "Start ACS";
+        LOG_IF(request->cs_params().channel == 0, DEBUG) << "Start ACS";
+
+        // Set transmit power
+        if (request->tx_limit_valid()) {
+            ap_wlan_hal->set_tx_power_limit(request->tx_limit());
+            if (ap_wlan_hal->get_radio_info().channel == request->cs_params().channel) {
+                LOG(DEBUG) << "Setting tx power without channel switch, send CSA notification";
+                auto notification = message_com::create_vs_message<
+                    beerocks_message::cACTION_APMANAGER_HOSTAP_CSA_NOTIFICATION>(cmdu_tx);
+                if (!notification) {
+                    LOG(ERROR) << "Failed building message!";
+                    return false;
+                }
+                ap_wlan_hal->refresh_radio_info();
+                fill_cs_params(notification->cs_params());
+                message_com::send_cmdu(slave_socket, cmdu_tx);
+            }
+        }
 
         // Set AP channel
         if (ap_wlan_hal->get_radio_info().channel != request->cs_params().channel &&
@@ -607,10 +623,6 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             message_com::send_cmdu(slave_socket, cmdu_tx);
             return false;
         }
-        //TODO: in case only tx power received need to send notification to slave as there
-        // is no event for this
-        if (request->tx_limit_valid())
-            ap_wlan_hal->set_tx_power_limit(request->tx_limit());
         break;
     }
     case beerocks_message::ACTION_APMANAGER_HOSTAP_SET_NEIGHBOR_11K_REQUEST: {
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1046,6 +1046,7 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
     case Event::CSA_Finished: {
 
         ap_wlan_hal->read_acs_report();
+        ap_wlan_hal->refresh_radio_info();
 
         LOG(INFO) << ((event == Event::ACS_Completed) ? "ACS_Completed" : "CSA_Finished:")
                   << " channel = " << int(ap_wlan_hal->get_radio_info().channel)
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -603,17 +603,7 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
                 LOG(ERROR) << "Failed building message!";
                 return false;
             }
-
-            notification->cs_params().channel  = ap_wlan_hal->get_radio_info().channel;
-            notification->cs_params().tx_power = get_tx_power(ap_wlan_hal);
-            notification->cs_params().bandwidth =
-                uint8_t(beerocks::utils::convert_bandwidth_to_enum(
-                    ap_wlan_hal->get_radio_info().bandwidth));
-            notification->cs_params().channel_ext_above_primary =
-                ap_wlan_hal->get_radio_info().channel_ext_above;
-            notification->cs_params().vht_center_frequency =
-                ap_wlan_hal->get_radio_info().vht_center_freq;
-            notification->cs_params().switch_reason = beerocks::CH_SWITCH_REASON_UNKNOWN;
+            fill_cs_params(notification->cs_params());
             message_com::send_cmdu(slave_socket, cmdu_tx);
             return false;
         }
@@ -993,6 +983,18 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
     return true;
 }
 
+void ap_manager_thread::fill_cs_params(beerocks_message::sApChannelSwitch &params)
+{
+    params.tx_power  = get_tx_power(ap_wlan_hal);
+    params.channel   = ap_wlan_hal->get_radio_info().channel;
+    params.bandwidth = uint8_t(
+        beerocks::utils::convert_bandwidth_to_enum(ap_wlan_hal->get_radio_info().bandwidth));
+    params.channel_ext_above_primary = ap_wlan_hal->get_radio_info().channel_ext_above;
+    params.vht_center_frequency      = ap_wlan_hal->get_radio_info().vht_center_freq;
+    params.switch_reason             = uint8_t(ap_wlan_hal->get_radio_info().last_csa_sw_reason);
+    params.is_dfs_channel            = ap_wlan_hal->get_radio_info().is_dfs_channel;
+}
+
 bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event_ptr)
 {
     if (!event_ptr) {
@@ -1043,9 +1045,6 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
                   << " last_csa_switch_reason enum = "
                   << int(ap_wlan_hal->get_radio_info().last_csa_sw_reason);
 
-        // CMDU Message
-        beerocks_message::sApChannelSwitch *msg_params = nullptr;
-
         if (event == Event::ACS_Completed) {
             auto notification = message_com::create_vs_message<
                 beerocks_message::cACTION_APMANAGER_HOSTAP_ACS_NOTIFICATION>(cmdu_tx);
@@ -1055,7 +1054,7 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
             }
             auto tuple_supported_channels = notification->supported_channels_list(0);
             copy_radio_supported_channels(ap_wlan_hal, &std::get<1>(tuple_supported_channels));
-            msg_params               = &notification->cs_params();
+            fill_cs_params(notification->cs_params());
             acs_completed_vap_update = true;
         } else {
             auto notification = message_com::create_vs_message<
@@ -1064,17 +1063,8 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
                 LOG(ERROR) << "Failed building message!";
                 return false;
             }
-            msg_params = &notification->cs_params();
+            fill_cs_params(notification->cs_params());
         }
-
-        msg_params->tx_power  = get_tx_power(ap_wlan_hal);
-        msg_params->channel   = ap_wlan_hal->get_radio_info().channel;
-        msg_params->bandwidth = uint8_t(
-            beerocks::utils::convert_bandwidth_to_enum(ap_wlan_hal->get_radio_info().bandwidth));
-        msg_params->channel_ext_above_primary = ap_wlan_hal->get_radio_info().channel_ext_above;
-        msg_params->vht_center_frequency      = ap_wlan_hal->get_radio_info().vht_center_freq;
-        msg_params->switch_reason  = uint8_t(ap_wlan_hal->get_radio_info().last_csa_sw_reason);
-        msg_params->is_dfs_channel = ap_wlan_hal->get_radio_info().is_dfs_channel;
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
 
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -983,7 +983,7 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
 void ap_manager_thread::fill_cs_params(beerocks_message::sApChannelSwitch &params)
 {
-    params.tx_power  = ap_wlan_hal->get_radio_info().conducted_power;
+    params.tx_power  = ap_wlan_hal->get_radio_info().tx_power;
     params.channel   = ap_wlan_hal->get_radio_info().channel;
     params.bandwidth = uint8_t(
         beerocks::utils::convert_bandwidth_to_enum(ap_wlan_hal->get_radio_info().bandwidth));
@@ -1485,10 +1485,10 @@ void ap_manager_thread::handle_hostapd_attached()
 
     notification->params().iface_type = uint8_t(ap_wlan_hal->get_iface_type());
     notification->params().iface_mac = network_utils::mac_from_string(ap_wlan_hal->get_radio_mac());
-    notification->params().iface_is_5ghz   = ap_wlan_hal->get_radio_info().is_5ghz;
-    notification->params().ant_num         = ap_wlan_hal->get_radio_info().ant_num;
-    notification->params().conducted_power = ap_wlan_hal->get_radio_info().conducted_power;
-    notification->cs_params().channel      = ap_wlan_hal->get_radio_info().channel;
+    notification->params().iface_is_5ghz = ap_wlan_hal->get_radio_info().is_5ghz;
+    notification->params().ant_num       = ap_wlan_hal->get_radio_info().ant_num;
+    notification->params().tx_power      = ap_wlan_hal->get_radio_info().tx_power;
+    notification->cs_params().channel    = ap_wlan_hal->get_radio_info().channel;
     notification->cs_params().channel_ext_above_primary =
         ap_wlan_hal->get_radio_info().channel_ext_above;
     notification->cs_params().vht_center_frequency = ap_wlan_hal->get_radio_info().vht_center_freq;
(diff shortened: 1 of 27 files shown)


common/beerocks/bwl/dummy/nl80211_client_dummy.h
```diff
@@ -44,6 +44,18 @@ class nl80211_client_dummy : public nl80211_client {
      */
     virtual bool get_sta_info(const std::string &local_interface_name,
                               const sMacAddr &sta_mac_address, sStaInfo &sta_info) override;
+
+    /**
+     * @brief Set the tx power limit object
+     *
+     * Set tx power limit for a radio
+     *
+     * @param[in] local_interface_name Virtual AP (VAP) interface name.
```
mariomaz: I think it can be a physical radio interface as well as a virtual interface

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
```diff
@@ -325,7 +325,10 @@ class ap_wlan_hal : public virtual base_wlan_hal {
      *
      * @return true on success or false on error.
      */
-    virtual bool set_tx_power_limit(int tx_pow_limit) = 0;
+    virtual bool set_tx_power_limit(int tx_pow_limit)
```
mariomaz: I think parameter should be uint32_t

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -72,6 +72,15 @@ class nl80211_client {
      */
     virtual bool get_sta_info(const std::string &local_interface_name,
                               const sMacAddr &sta_mac_address, sStaInfo &sta_info) = 0;
+    /**
+     * @brief Set the tx power limit object
+     *
+     * Set tx power limit for a radio
+     * @param[in] local_interface_name Virtual AP (VAP) interface name.
```
mariomaz: I think it works both with radio or virtual interface name

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -182,4 +182,34 @@ bool nl80211_client_impl::get_sta_info(const std::string &local_interface_name,
...
+    return m_socket.get()->send_receive_msg(
+        NL80211_CMD_SET_WIPHY, 0,
+        [&](struct nl_msg *msg) -> bool {
+            nla_put_u32(msg, NL80211_ATTR_IFINDEX, iface_index);
+            nla_put_u32(msg, NL80211_ATTR_WIPHY_TX_POWER_SETTING, NL80211_TX_POWER_LIMITED);
+            nla_put_u32(msg, NL80211_ATTR_WIPHY_TX_POWER_LEVEL, limit * 100); //mBm
+            return true;
+        },
+        [&](struct nl_msg *msg) -> bool {
+            return true;
```
mariomaz: There must be some kind of return code to check if operation succeeded or not

common/beerocks/bwl/shared/nl80211_client_impl.h
```diff
@@ -57,6 +57,17 @@ class nl80211_client_impl : public nl80211_client {
     virtual bool get_sta_info(const std::string &local_interface_name,
                               const sMacAddr &sta_mac_address, sStaInfo &sta_info) override;
 
+    /**
+     * @brief Set the tx power limit object
+     *
+     * Set tx power limit for a radio
+     * @param[in] local_interface_name Virtual AP (VAP) interface name.
```
mariomaz: Radio interface

common/beerocks/bwl/shared/nl80211_client_impl.h
```diff
@@ -57,6 +57,17 @@ class nl80211_client_impl : public nl80211_client {
     virtual bool get_sta_info(const std::string &local_interface_name,
                               const sMacAddr &sta_mac_address, sStaInfo &sta_info) override;
 
+    /**
+     * @brief Set the tx power limit object
```
mariomaz: What do you mean by 'object'?

common/beerocks/bwl/common/base_wlan_hal.cpp
```diff
@@ -32,6 +33,9 @@ base_wlan_hal::base_wlan_hal(HALType type, std::string iface_name, IfaceType ifa
 
     // Initialize complex containers of the radio_info structure
     m_radio_info.supported_channels.resize(128 /* TODO: Get real value */);
+
+    m_nl80211_client = nl80211_client_factory::create_instance();
```
mariomaz: nl80211_client is a dependency and thus should be injected, not hard-coded.
If we ever create a unit test for this class, nl80211_client should be provided as a parameter.

mariomaz (approved):
Only nitpick comments so approving.

common/beerocks/bwl/shared/nl80211_client_impl.h
tomereli: typo

common/beerocks/bwl/common/base_wlan_hal.cpp
tomereli: I'm not sure I follow - can you explain what is hard coded here and what do you mean by injecting it?

mariomaz: `nl80211_client` is a dependency for `base_wlan_hal`. `base_wlan_hal` should not create it. `nl80211_client` should be injected as a parameter to the constructor of `base_wlan_hal`.

But this is only actually required if we want to create a unit test for `base_wlan_hal`. In such unit test, we would pass a mock `nl80211_client` instead of the real implementation.

mariomaz:
BTW, I've realized just now that you've moved `set_tx_power_limit` from `ap_wlan_hal_*` to `base_wlan_hal`. This is a base class for `ap_wlan_hal_*`, `base_wlan_hal_*`, `mon_wlan_hal_*` and `sta_wlan_hal_*`classes so `set_tx_power_limit` will be available to all of them. Was that the intention or a side effect? I think there will be more instances of `nl80211_client` that what they are really needed.

Moreover, since `ap_wlan_hal_*` classes extend both `base_wlan_hal` and `base_wlan_hal_*` there will be two instances of `nl80211_client`. Again, more than needed. 

I think `nl80211_client` should be a member of `ap_wlan_hal_*` and not of `base_wlan_hal`

tomereli:
> BTW, I've realized just now that you've moved `set_tx_power_limit` from `ap_wlan_hal_*` to `base_wlan_hal`. This is a base class for `ap_wlan_hal_*`, `base_wlan_hal_*`, `mon_wlan_hal_*` and `sta_wlan_hal_*`classes so `set_tx_power_limit` will be available to all of them. Was that the intention or a side effect? I think there will be more instances of `nl80211_client` that what they are really needed.
> 
> Moreover, since `ap_wlan_hal_*` classes extend both `base_wlan_hal` and `base_wlan_hal_*` there will be two instances of `nl80211_client`. Again, more than needed.
> 
> I think `nl80211_client` should be a member of `ap_wlan_hal_*` and not of `base_wlan_hal`

True, didn't think about this. Thanks. I added the don't merge tag for now.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
tomereli: I couldn't find this - from what I could understand, there is a default callback for error so I don't need to handle errors here.

common/beerocks/bwl/dummy/nl80211_client_dummy.h
tomereli: I think it will cause confusion, so lets keep it for radio only.

common/beerocks/bwl/include/bwl/nl80211_client.h
tomereli: I checked on UGW where I have only radio and it works, I didn't check VAP so I'll change the comment to say only the radio name is supported.

common/beerocks/bwl/common/base_wlan_hal.cpp
tomereli: Done

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
tomereli: This was already like that apparently and I reverted this original change so I prefer not to change this as part of this PR.

tomereli:
@mariomaz I refactored this PR based on your last comment, so I think this requires a new review.

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -72,6 +72,16 @@ class nl80211_client {
      */
     virtual bool get_sta_info(const std::string &local_interface_name,
                               const sMacAddr &sta_mac_address, sStaInfo &sta_info) = 0;
+    /**
+     * @brief Set the tx power limit object
```
arnout: Didn't you say that "object" was a typo and that you fixed it?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -713,9 +715,11 @@ update_vap_credentials_configure_wpa(const std::string &vap_if,
 // NOTE: Since *base_wlan_hal_dwpal* inherits *base_wlan_hal* virtually, we
 //       need to explicitly call it's from any deriving class
 ap_wlan_hal_dwpal::ap_wlan_hal_dwpal(std::string iface_name, hal_event_cb_t callback,
-                                     hal_conf_t hal_conf)
+                                     hal_conf_t hal_conf,
+                                     std::unique_ptr<nl80211_client> nl80211_client)
     : base_wlan_hal(bwl::HALType::AccessPoint, iface_name, IfaceType::Intel, callback, hal_conf),
-      base_wlan_hal_dwpal(bwl::HALType::AccessPoint, iface_name, callback, hal_conf)
+      base_wlan_hal_dwpal(bwl::HALType::AccessPoint, iface_name, callback, hal_conf),
+      m_nl80211_client(std::move(nl80211_client))
```
arnout: I believe the std::move is redundant here because unique_ptr only has a move constructor, no copy constructor. (nitpick)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -2521,7 +2524,10 @@ bool ap_wlan_hal_dwpal::process_dwpal_nl_event(struct nl_msg *msg)
 std::shared_ptr<ap_wlan_hal> ap_wlan_hal_create(std::string iface_name, hal_conf_t hal_conf,
                                                 base_wlan_hal::hal_event_cb_t callback)
 {
-    return std::make_shared<dwpal::ap_wlan_hal_dwpal>(iface_name, callback, hal_conf);
+    auto nl80211_client = nl80211_client_factory::create_instance();
+    LOG_IF(!nl80211_client, FATAL) << "Failed to create nl80211_client instance";
+    return std::make_shared<dwpal::ap_wlan_hal_dwpal>(iface_name, callback, hal_conf,
+                                                      std::move(nl80211_client));
 }
```
arnout: @mariomaz I still don't really understand how the mocking would work. Now the nl80211_client_factory is hardcoded here, in ap_wlan_hal_create. If we would want to create a unit test for ap_wlan_hal_dwpal, we'd have to create a "mocking" factory function to replace this one, correct?

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -327,7 +327,8 @@ bool ap_wlan_hal_dummy::read_supported_channels() { return true; }
 
 bool ap_wlan_hal_dummy::set_tx_power_limit(int tx_pow_limit)
 {
-    LOG(TRACE) << __func__ << " power limit: " << tx_pow_limit;
+    LOG(TRACE) << __func__ << " setting power limit: " << tx_pow_limit * 100 << " mBm";
+    m_radio_info.conducted_power = tx_pow_limit * 100;
```
arnout: IMO this could have been squashed into the first commit.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -582,8 +582,24 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+            ap_wlan_hal->set_tx_power_limit(request->tx_limit());
+            if (ap_wlan_hal->get_radio_info().channel == request->cs_params().channel) {
+                LOG(DEBUG) << "Setting tx power without channel switch, send CSA notification";
+                auto notification = message_com::create_vs_message<
+                    beerocks_message::cACTION_APMANAGER_HOSTAP_CSA_NOTIFICATION>(cmdu_tx);
+                if (!notification) {
+                    LOG(ERROR) << "Failed building message!";
+                    return false;
+                }
+                ap_wlan_hal->refresh_radio_info();
```
arnout, line 584: Since nothing has changed, refreshing radio info shouldn't be needed.

Looking at it in more detail, it's even worse. Setting the TX power will not be reflected in the result of get_tx_power(), because that is using the tx power limit from the supported channels, and that, in turn, is purely based on the regulatory limits and the hardware limit, *not* on the power limit set by NL80211_ATTR_WIPHY_TX_POWER_SETTING. So I think we need to add an actual power limit setting in bwl::RadioInfo and make sure it gets filled in by refresh_radio_info(). But that's for another PR...

arnout (approved):
Just one remaining typo in a comment.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: It looks like @mariomaz already included the max tx power using `NL80211_FREQUENCY_ATTR_MAX_TX_POWER` in #977 - I think we should decide how we want to proceed with @mariomaz 's PR before we merge this one.
I have UGW so I can test it correctly all the way, no need to rush and merge this half way done.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Apparently it is not :/

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: So apparently my previous comment is wrong, we don't need to rebase on #977 in order to read the tx power since it is reflected in `TxPower` when running `hostapd_cli -iwlan0 radio_info` hence available in `m_radio_info.conducted_power`. Added a commit which fixes this.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -182,4 +182,32 @@ bool nl80211_client_impl::get_sta_info(const std::string &local_interface_name,
...
+{
+    if (!m_socket) {
+        LOG(ERROR) << "Socket is NULL!";
+        return false;
+    }
+
+    // Get the interface index for given interface name
+    int iface_index = if_nametoindex(local_interface_name.c_str());
+    if (0 == iface_index) {
+        LOG(ERROR) << "Failed to read the index of interface " << local_interface_name << ": "
```
vitalybu: I think that you should either validate that `local_interface_name` is not empty, or at least add `''` to the log (so that it'll be easier to spot an empty value):
```suggestion
        LOG(ERROR) << "Failed to read the index of interface '" << local_interface_name << "': "
```

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -182,4 +182,32 @@ bool nl80211_client_impl::get_sta_info(const std::string &local_interface_name,
...
+                   << strerror(errno);
+
+        return false;
+    }
+
+    return m_socket.get()->send_receive_msg(
+        NL80211_CMD_SET_WIPHY, 0,
+        [&](struct nl_msg *msg) -> bool {
+            nla_put_u32(msg, NL80211_ATTR_IFINDEX, iface_index);
+            nla_put_u32(msg, NL80211_ATTR_WIPHY_TX_POWER_SETTING, NL80211_TX_POWER_LIMITED);
```
vitalybu, line 206: Shouldn't we allow returning the power setting to `NL80211_TX_POWER_AUTOMATIC`?
Or perhaps even support all TX_POWER modes:
- NL80211_TX_POWER_AUTOMATIC
- NL80211_TX_POWER_FIXED
- NL80211_TX_POWER_LIMITED

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1590,8 +1594,7 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
 
 bool ap_wlan_hal_dwpal::set_tx_power_limit(int tx_pow_limit)
 {
-    LOG(DEBUG) << "set_tx_power_limit(): missing function implementation";
-    return true;
+    return m_nl80211_client->set_tx_power_limit(m_radio_info.iface_name, tx_pow_limit);
```
vitalybu, line 1593: Since now `nl80211_client` is passed to the constructor (and has a default value of `nullptr`), you can end up with a null reference exception here.
You should either remove the default value and make sure it's not `nullptr`, or check if it's not `nullptr` before every call.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -572,8 +576,7 @@ bool ap_wlan_hal_nl80211::read_supported_channels()
 
 bool ap_wlan_hal_nl80211::set_tx_power_limit(int tx_pow_limit)
 {
-    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED!";
-    return true;
+    return m_nl80211_client->set_tx_power_limit(m_radio_info.iface_name, tx_pow_limit);
```
vitalybu, line 575: dwpal (previous commit) comments apply here as well

common/beerocks/bwl/shared/nl80211_client_impl.cpp
tomereli: The only thing which is needed by EasyMesh is setting power limit.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: But in dwpal we always pass it, and override the default value - and if it fails then we assert with LOG_IF(FATAL) - do you still think it is necessary to check every call?

vitalybu: No, I think you should remove the default value from the constructor and force passing a valid instance, while also validating that it's not nullptr.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
vitalybu: ¯\\_(ツ)_/¯

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: OK, done. I removed the default value and just moved the initialization to `base_wlan_hal_*`. IMO it makes much more sense there instead of manually specifying it for every derived class.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: I moved the `create_instance` to `base_wlan_hal_XXX` constructor since the nl80211_client can be and will be used by the base class to update radio info which is implemented in the base class.

As @vitalybu said:
>Dependency injection is useful when you want to "inject" different implementations of the same interface to some class. I'm just not sure if it's the case here. Why would we ever use a non-DWPAL nl80211 client with DWPAL BWL?

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
tomereli: IMHO it makes more sense as a separate commit.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
```diff
@@ -88,6 +89,8 @@ class base_wlan_hal_dwpal : public virtual base_wlan_hal,
     bool dwpal_nl_cmd_scan_dump();
     void *get_dwpal_nl_ctx() const { return (m_dwpal_nl_ctx); }
 
+    std::unique_ptr<nl80211_client> m_nl80211_client;
```
arnout, line 92: What is the point of moving it to base_wlan_hal_dwpal? I though we said that it wouldn't be used by the monitor or station HALs...

More importantly though: if you move it here, it should also be taken as an argument in the base_wlan_hal_dwpal constructor and set there, not in the derived class's constructor.

arnout (approved):
Still looks good to me, with one minor comment.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
tomereli: The point is that the `refresh_radio_info` is implemented in the base, so I thought that in order to use @mariomaz's #977 to read the TxPower directly from nl80211 I will need to do that from `refresh_radio_info`.
It turns out that the `TxPower` we get from hostapd is fine, but nevertheless we will probably update `refresh_radio_info` to complement the radio info parameters we get from hostapd with the values we read from nl80211, so I decided to keep it there.

tomereli: >More importantly though: if you move it here, it should also be taken as an argument in the base_wlan_hal_dwpal constructor and set there, not in the derived class's constructor.

What's wrong with instantiating it in `base_wlan_hal_dwpal` like I did here? It is not set in the derived class constructor:
```
base_wlan_hal_dwpal::base_wlan_hal_dwpal(HALType type, std::string iface_name,
                                         hal_event_cb_t callback, hal_conf_t hal_conf)
    : base_wlan_hal(type, iface_name, IfaceType::Intel, callback, hal_conf),
      beerocks::beerocks_fsm<dwpal_fsm_state, dwpal_fsm_event>(dwpal_fsm_state::Delay),
      m_nl80211_client(nl80211_client_factory::create_instance())
{
    LOG_IF(!m_nl80211_client, FATAL) << "Failed to create nl80211_client instance";
    // Initialize the FSM
    fsm_setup();
}
```

tomereli:
```
base_wlan_hal_dwpal::base_wlan_hal_dwpal(HALType type, std::string iface_name,
                                         hal_event_cb_t callback, hal_conf_t hal_conf)
    : base_wlan_hal(type, iface_name, IfaceType::Intel, callback, hal_conf),
      beerocks::beerocks_fsm<dwpal_fsm_state, dwpal_fsm_event>(dwpal_fsm_state::Delay),
      m_nl80211_client(nl80211_client_factory::create_instance())
{
    LOG_IF(!m_nl80211_client, FATAL) << "Failed to create nl80211_client instance";
    // Initialize the FSM
    fsm_setup();
}
```

arnout: I'm sorry, I misread the code.

PR #991, opened 2020-03-22
controller: son_management: check single scan in progress

### LINKED ISSUE #990: [BUG] controller: on BML single scan request, there is no check if single scan is in progress
**Describe the bug** -
The son_management is responsible to send a `dynamic_channel_selection_task::eEvent::TRIGGER_SINGLE_SCAN` request to the Dynamic channel selection task, when a `beerocks_message::cACTION_BML_CHANNEL_SCAN_START_SCAN_REQUEST` event is received.
Before triggering the request, the son_management must verify that a scan is not in progress since the task supports one single-scan request at a time.
The above-mentioned check is not performed and the scan parameters of the already triggered single-scan can be overridden.

## Steps to reproduce

Run 2  consecutive (very fast) single scan request one after the other on the same radio interface.

## Expected behavior

The first single-scan will complete successfully, while the second request will return an error "scan-in-progress".

## Target

- Tested on UGW but is reproducible on any platform

adam1985d:
The son_management is responsible to send a
`dynamic_channel_selection_task::eEvent::TRIGGER_SINGLE_SCAN` request to
the Dynamic Channel Selection task, when a
`beerocks_message::cACTION_BML_CHANNEL_SCAN_START_SCAN_REQUEST` event
is received.
Before triggering the request, the son_management must verify that a
a scan is not in progress since the task supports one single-scan
request at a time.
The above-mentioned check is not performed and the scan parameters of
the already triggered single-scan can be overridden.

Added a check of single-scan in progress. If a scan is in progress a
`CHANNEL_SCAN_OP_SCAN_IN_PROGRESS` error code is returned and the scan request is not passed to the task.

Fixes #990 

### DIFF
--- a/controller/src/beerocks/master/son_management.cpp
+++ b/controller/src/beerocks/master/son_management.cpp
@@ -2071,6 +2071,18 @@ void son_management::handle_bml_message(Socket *sd,
             LOG(ERROR) << "Failed building message cACTION_BML_CHANNEL_SCAN_START_SCAN_RESPONSE !";
             return;
         }
+
+        // Get single scan status
+        auto single_scan_in_progress =
+            database.get_channel_scan_in_progress(request->scan_params().radio_mac, true);
+        if (single_scan_in_progress) {
+            LOG(ERROR) << "Single scan is still running!";
+            response->op_error_code() =
+                uint8_t(eChannelScanOpErrCode::CHANNEL_SCAN_OP_SCAN_IN_PROGRESS);
+            message_com::send_cmdu(sd, cmdu_tx);
+            break;
+        }
+
         auto radio_mac         = request->scan_params().radio_mac;
         auto dwell_time_ms     = request->scan_params().dwell_time_ms;
         auto channel_pool_size = request->scan_params().channel_pool_size;


controller/src/beerocks/master/son_management.cpp
```diff
@@ -2071,6 +2071,18 @@ void son_management::handle_bml_message(Socket *sd,
             LOG(ERROR) << "Failed building message cACTION_BML_CHANNEL_SCAN_START_SCAN_RESPONSE !";
             return;
         }
+
+        // Get scan status
+        auto scan_in_progress =
+            database.get_channel_scan_in_progress(request->scan_params().radio_mac, true);
+        if (scan_in_progress) {
+            LOG(ERROR) << "Single scan is still running!";
```
tomereli, line 2079: Who said its a single scan which is running? Can't it be in continues scan?

tomereli (approved):
small nitpick

controller/src/beerocks/master/son_management.cpp
adam1985d: 1. this flow is for single-scan only
2. the single-scan and continuous-scan have separate configuration params in DB - thus a continuous scan in progress should not block a single scan configuration. Moreover - the received single-scan request will now be pending and triggered only after the continuous scan is complete - but this is just elaboration on the flow and not related to the issue described :-)

kantera800: then, 
 either use define "SINGLE_CHANNEL_SCAN" instead of **true** in database.get_channel_scan_in_progres(...)
or/and  add comments 
or/and rename scan_in_progress to  `single_scan_in_progress `

kantera800 (changes requested):

controller/src/beerocks/master/son_management.cpp
adam1985d: edited the comment to `Get single scan status`
and
edited the `scan_in_progress` to `single_scan_in_progress`

kantera800 (approved):

PR #993, opened 2020-03-22
Hotfix/dynamic channel selection task: increase timeout

### LINKED ISSUE #992: [BUG] Dynamic Channel Selection Task : Observing time out while performing a scan with a large channel pool
The Dynamic Channel Selection uses a channel scan to get the neighboring AP results.
The channel pool is set in the DB and is then used when a scan is triggered (both for the single and continuous scan modes).

The issue rises if a large channel pool is set, or if a large quantity of DFS channels are present in the channel pool. If that happens the scan has a chance of reaching "CHANNEL_SCAN_RESULTS_READY_EVENT_TIMEOUT"
Even though the scan does not actually fail

## Steps to reproduce
1. Build
2. Run beerocks_cli
3. Run the single scan with a large (and valid) channel pool
bml_start_dcs_single_scan ac:9a:96:fb:2d:f6 45 100,36,108,60,64,128,161,153,157,136

## Expected behavior

The scan will reach a "CHANNEL_SCAN_RESULTS_READY_EVENT_TIMEOUT" event and set the status accordingly.

To fix the issue the simplest solution would be to increase the wait time to a larger value,
The suggested value is 3 and a half minutes (210000 milliseconds).
Since the value is used to determine the "last resort" catch scenario, increasing the value is the most optimal solution

itayx:
In the Dynamic Channel Selection (DCS) task, the task reads the channel
pool avaliable for the current radio (and scan type) then sents that
pool the the monitor to trigger a channel scan.
The task then waits for events sent from the monitor and acts
accordingly.

If the sent pool is very large and/or contains DFS channels the task
might read the abnormaly losg wait time as too long thus resultig in
a "CHANNEL_SCAN_RESULTS_READY_EVENT_TIMEOUT" or a similer timeout
event\status, even if the scan eventually finishes.

Issue: #992

### DIFF
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -12,8 +12,8 @@
 #include <bcl/network/network_utils.h>
 #include <easylogging++.h>
 
-#define SCAN_TRIGGERED_WAIT_TIME_MSEC 20000
-#define SCAN_RESULTS_DUMP_WAIT_TIME_MSEC 40000
+#define SCAN_TRIGGERED_WAIT_TIME_MSEC 20000     //20 Sec
+#define SCAN_RESULTS_DUMP_WAIT_TIME_MSEC 210000 //3.5 Min
 
 std::string s_ar_states[] = {FOREACH_DCS_STATE(GENERATE_STRING)};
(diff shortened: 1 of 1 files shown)


kantera800 (approved):
there is no need to open an bug issue when you are already solved it and opened a hotfix PR (nitpick)
approved

tomereli (approved):
There are a bunch of typos in the commit message, please fix before merging.

PR #994, opened 2020-03-22
Hotfix/fix reconnect of son slave on topology flow

morantr:
When a son_slave reconnects to the backhaul manager after it disconnection, the Backhaul Manager
sends topology notification that results in getting Topology Query message.
Filling the device information on the Topology Response message might fail because of
a race condition. The radio_mac and the freq_type are being updated on the BACKHAUL_ENABLE message,
but we trigger the Topology notification on BACKHAUL REGISTER message, so if the Topology Query
message would get before getting the BACKHAUL ENABLE, the filling of the device information
will fail.

Moved the sending of the Topology Notification from BACKHAUL REGISTER to BACKHAUL ENABLE.

Created a pipeline that checks the following: 
```
MAP-4.2.1
MAP-4.3.1_ETH
MAP-4.3.2_ETH_FH24G
MAP-4.3.2_ETH_FH5GL
MAP-4.3.2_ETH_FH5GH
```

https://gitlab.com/prpl-foundation/prplMesh/pipelines/128752410

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1725,18 +1725,6 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
             }
         }
 
-        if (m_eFSMState >= EState::CONNECT_TO_MASTER) {
-            LOG(INFO) << "Sending topology notification on reconnected son_slave";
-            auto cmdu_header =
-                cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE);
-            if (!cmdu_header) {
-                LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
-                return false;
-            }
-            cmdu_header->flags().relay_indicator = true;
-            return send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
-        }
-
         LOG(DEBUG) << "ACTION_BACKHAUL_REGISTER_REQUEST sta_iface=" << soc->sta_iface
                    << " local_master=" << int(local_master) << " local_gw=" << int(local_gw)
                    << " hostap_iface=" << soc->hostap_iface;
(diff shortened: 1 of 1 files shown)


tomereli (approved):

kantera800 (changes requested):
@morantr 
The original problem was spotted on UGW so we need to test this fix there also.
In addition please perform the following test:
->kill the monitor process manually 
->check if agent restarts it

adding the don't merge flag until then

morantr:
> @morantr
> The original problem was spotted on UGW so we need to test this fix there also.
> In addition please perform the following test:
> ->kill the monitor process manually
> ->check if agent restarts it
> 
> adding the don't merge flag until then

passed

arnout (approved):
The second commit message doesn't convey anything useful. The change is not related at all to commit 10f137c - that commit changed the return of fill_radio_iface_info, and only in one specific case; here you're changing the return of handle_1905_topology_query.

A better commit message:

```
agent: handle_topology_query(): return true also on failure

If a handle_... function returns false, it means that the message will be forwarded
to the slaves. That's not what we want here, so always return true.
```

However, with this reasoning, we should return true *everywhere*. Indeed, except for maybe one or two cases, if the message has a handle function, it really shouldn't be forwarded.

So I think a better solution would be to add an addition argument to handle_1905_1_message() and use that to indicate if it needs to be forwarded or not. If the function returns false, it's an error so you don't forward. If it returns true, you check the forward_to_slaves value and if true you forward. What do you think?

morantr:
> The second commit message doesn't convey anything useful. The change is not related at all to commit [10f137c](https://github.com/prplfoundation/prplMesh/commit/10f137c235a0e9a26142617b63e33609ca46d0c7) - that commit changed the return of fill_radio_iface_info, and only in one specific case; here you're changing the return of handle_1905_topology_query.
> 
> A better commit message:
> 
> ```
> agent: handle_topology_query(): return true also on failure
> 
> If a handle_... function returns false, it means that the message will be forwarded
> to the slaves. That's not what we want here, so always return true.
> ```
> 
> However, with this reasoning, we should return true _everywhere_. Indeed, except for maybe one or two cases, if the message has a handle function, it really shouldn't be forwarded.
> 
> So I think a better solution would be to add an addition argument to handle_1905_1_message() and use that to indicate if it needs to be forwarded or not. If the function returns false, it's an error so you don't forward. If it returns true, you check the forward_to_slaves value and if true you forward. What do you think?

I don't think we should put effort into that since when we will create the unified Agent, all of this will be moved to there, and no forwarding will be done.

kantera800 (approved):

arnout:
@morantr @kantera800 Remember it is now not sufficient any more to remove the don't merge tag, you also have to set the Ready for merge tag if you want mergify to proceed.

I merged manually now.

PR #1002, opened 2020-03-23
Feature/cppcheck controller

rmelotte:
Fix issues of severity warning and error for the 'controller' folder, and enforce cppcheck on it.

Also enables inline suppression for cppcheck.

Next is the agent folder, then we will be able to run cppcheck on the whole tree (warnings and errors only, until the others are also fixed).


controller/src/beerocks/master/db/node.h
```diff
@@ -239,7 +239,7 @@ class node {
         ap_metrics_data(){};
         ~ap_metrics_data(){};
 
-        sMacAddr bssid;
+        sMacAddr bssid                               = {0};
```
tomereli:
```suggestion
        sMacAddr bssid                               = network_utils::ZERO_MAC;
```

controller/src/beerocks/cli/beerocks_cli_bml.h
```diff
@@ -67,11 +67,11 @@ class cli_bml : public cli {
     cli_bml(std::string beerocks_conf_path_);
     virtual ~cli_bml();
 
-    bool connect();
-    void disconnect();
-    bool is_connected();
-    bool start() { return true; };
-    void stop(){};
+    bool connect() override;
```
tomereli, line 70: shouldn't we also add the `virtual` keyword, while at it?

controller/src/beerocks/cli/beerocks_cli_bml.h
```diff
@@ -80,7 +80,7 @@ class cli_bml : public cli {
 
 private:
     // Help functions
-    void setFunctionsMapAndArray();
+    void setFunctionsMapAndArray() override;
```
tomereli, line 83: Again, shouldn't we use `virtual`?
```suggestion
    virtual void setFunctionsMapAndArray() override;
```

controller/src/beerocks/cli/beerocks_cli_socket.h
```diff
@@ -35,7 +35,7 @@ class cli_socket : public socket_thread, public cli {
     virtual bool socket_disconnected(Socket *sd) override;
 
 private:
-    void setFunctionsMapAndArray();
+    void setFunctionsMapAndArray() override;
```
tomereli, line 38:
```suggestion
    virtual void setFunctionsMapAndArray() override;
```

controller/src/beerocks/cli/beerocks_cli_socket.h
```diff
@@ -21,11 +21,11 @@ class cli_socket : public socket_thread, public cli {
     cli_socket(std::string temp_path_, std::string proxy_ip_ = std::string());
     ~cli_socket();
 
-    bool connect();
-    void disconnect();
-    bool is_connected() { return (master_socket != nullptr); }
-    bool start() { return socket_thread::start(); }
-    void stop() { socket_thread::stop(); }
+    bool connect() override;
```
tomereli, line 24:
```suggestion
    virtual bool connect() override;
```
Its not mandatory and is exactly the same functionality-wise, but is clearer.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -2164,6 +2164,7 @@ int bml_internal::get_wifi_credentials(int vap_id, char *ssid, char *pass, int *
     //clear the memory with password in it.
     memset(&sWifiCredentials.pass, 0, sizeof(sWifiCredentials.pass));
     //memset might be optimized and compiler might not set it 0 if its not used after memset
+    // cppcheck-suppress selfAssignment
     *(volatile char *)sWifiCredentials.pass = *(volatile char *)sWifiCredentials.pass;
```
tomereli: I think this is a terrible hack. A short google search says that `memset_s` will not be optimized out - https://www.cryptologie.net/article/419/zeroing-memory-compiler-optimizations-and-memset_s/
I would really prefer to fix the code to use memset_s with a comment instead of this hack.
In general, I prefer not to add the `// cppcheck-suppress ` comments in our production code.

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -12,7 +12,6 @@ INCLUDES="\
     -I$rootdir/framework/common/include \
     -I$rootdir/framework/tlvf/src/include \
     -I$rootdir/framework/tlvf/AutoGenerated/include \
-    -I$rootdir/framework/discovery_agent/include \
```
tomereli: Can we please please delete the discovery agent while we're at it? Its not used and will not be used.

tomereli (approved):
One major comment which can be fixed rather easily - I prefer to avoid the need to add comments in the code which are used by external tools, I dislike the //clang-format-off and the cppcheck comment is the same.. In this particular case there is a better solution, so I suggest to use it.

controller/src/beerocks/master/db/node.h
arnout: Alternatively, we could give sMacAddr a default constructor, or initialize its fields.

controller/src/beerocks/cli/beerocks_cli_bml.h
arnout: Hm, coding style question...

The `virtual` only needs to be added in the base class declaration, it is implicit in subclasses. And for the programmer, the `override` clearly indicates that it's a virtual function. So my opinion would be: save some horizontal space and drop the `virtual`.

Currently, we have roughly 400 cases of `override`, and about 300 of those also have `virtual`. That is not counting the generated code, which has about 600 more instances of `override` without `virtual` (but changing 3 lines of Python converts all of them, of course). So, if we take the current situation as a guideline, the `virtual` should still be there.

Poll, @fvbogaert ?

controller/src/beerocks/cli/beerocks_cli_socket.cpp
```diff
@@ -31,7 +31,7 @@ cli_socket::cli_socket(std::string temp_path_, std::string proxy_ip_)
     temp_path           = temp_path_;
 }
 
-cli_socket::~cli_socket() { disconnect(); }
+cli_socket::~cli_socket() { cli_socket::disconnect(); }
```
arnout, line 34: This is exactly the same case as for cli_bml. I think it would be clearer if the two commits were squashed (nitpick).

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -31,6 +31,7 @@ run_cppcheck() {
              --error-exitcode=1 \
              --enable=warning,style,information \
              -j"$(nproc)" \
+             --inline-suppr \
```
arnout: If we follow @tomereli's advice on working around the problem rather than suppressing it, we also should remove this commit.

controller/src/beerocks/bml/internal/bml_internal.cpp
arnout: memset_s is C11, not C++. cppreference [advises to use std::fill with volatile instead](https://en.cppreference.com/w/cpp/string/byte/memset#Notes). I like that solution a lot better.

arnout (approved):
I think both my and @tomereli's comments are not worth fixing, except for the alternative for memset.

controller/src/beerocks/cli/beerocks_cli_bml.h
rmelotte: Since override only applies to virtual function, I'd vote to drop it as well, it doesn't add any useful information.
If/when we introduce clang-tidy, there is even an option to do that automatically: https://clang.llvm.org/extra/clang-tidy/checks/modernize-use-override.html

controller/src/beerocks/bml/internal/bml_internal.cpp
rmelotte: I find the cppreference advice confusing. Does the pointer itself really need to be volatile?

Said differently: AFAIK calling std:fill with a `volatile char *p` should be fine. But does it really guarantee that it won't be optimized by any C++11-compliant compiler?
Or should we call it with both the pointer and the pointee volatile? (`volatile char *p volatile`) ?

arnout: Indeed, they mean with `volatile char* p`. "with volatile pointers" means "with pointers to volatile data". An actual `char * volatile p` is pretty useless, and casting to it is not even possible.

PR #1003, opened 2020-03-23
ucc listener: some fixes for being able to run all tests (even ones we don't support at the moment)

tomereli:
This PR fixes two issues which currently prevent us from running all agent tests in a single run:

1. Reply `INVALID` if non-eth backhaul onboarding command is received
2. Always reply with `RUNNING` first as it appears that the UCC hangs since it expects RUNNING regardless of the command success/fail criteria

### DIFF
--- a/agent/src/beerocks/slave/agent_ucc_listener.cpp
+++ b/agent/src/beerocks/slave/agent_ucc_listener.cpp
@@ -192,6 +192,13 @@ bool agent_ucc_listener::handle_dev_set_config(std::unordered_map<std::string, s
 
     auto &backhaul_param = params["backhaul"];
     std::transform(backhaul_param.begin(), backhaul_param.end(), backhaul_param.begin(), ::tolower);
+    // TODO - add wireless backhaul support.
+    // For now, this causes slave reset and the test to hang,
+    // so return not supported.
+    if (backhaul_param != "eth") {
+        err_string = "wireless backhaul not supported";
+        return false;
+    }
     m_selected_backhaul = backhaul_param;
 
     auto timeout =
--- a/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
+++ b/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
@@ -506,6 +506,12 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
 {
     LOG(DEBUG) << "Received WFA-CA, command=\"" << command << "\"";
 
+    // send back first reply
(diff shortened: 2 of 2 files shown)


agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -192,6 +192,13 @@ bool agent_ucc_listener::handle_dev_set_config(std::unordered_map<std::string, s
 
     auto &backhaul_param = params["backhaul"];
     std::transform(backhaul_param.begin(), backhaul_param.end(), backhaul_param.begin(), ::tolower);
+    // TODO - add wireless backhaul support.
+    // For now, this causes slave reset and the test to hang,
+    // so return not supported.
+    if (backhaul_param != "eth") {
```
vitalybu, line 198: LGTM, but wondering if `backhaul_param` can also be empty..

vitalybu (approved):

agent/src/beerocks/slave/agent_ucc_listener.cpp
tomereli: It cannot, its invalid if it is

rmelotte (approved):
There are some typos in the commit messages: 
immidiate -> immediate
immidietely -> immediately

Otherwise LGTM, so I'm approving to not hold the merge, but removing the merge tag so you can decide when you want it merged.

arnout (approved):
> Apparently the UCC expects to always receive status,RUNNING before any
> other status, so even if there is a failure with which we reply with
> status,INVALID,* - the UCC ignores this and hangs.

That is so insane... Figure 1 of the Control API Specification document clearly shows that INVALID comes on its own, and ERROR comes after RUNNING.

Given this situation, I think it would be better to never return INVALID at all, and return ERROR instead - that way we at least stay a little closer to the specification, and we work around this UCC issue.

But that really is nitpicking, and I don't think we should spent too much time on this since it's only for certification, so approving.

BTW there are still two INVALID branches that precede the RUNNING: forbidden_chars and cmd_tokens_vec.empty().

tomereli:
> BTW there are still two INVALID branches that precede the RUNNING: forbidden_chars and cmd_tokens_vec.empty().

Fixed.

arnout:
Note: CI hit #832 (cfr. https://gitlab.com/prpl-foundation/prplMesh/-/jobs/483262491/artifacts/browse/logs/gateway-30394/ - there are two log files for beerocks_agent_wlan0). I manually retriggered the failing test and now it should pass.

PR #1005, opened 2020-03-24
ci: add run_easymesh_cert.sh (PATCH 1/2)

tomereli:
Add a shell script which can be used to run certification tests.
This is a simple wrapper script that includes downloading of the latest
prplmesh IPK from gitlab, deploying it to the target, and running the
certification tests.
This script will be used by a cron job in each testbed for running full
nightly certifications.

For now, uploading the logs to owncloud is left as TODO.

While we're at it, also:
1. Move functions.sh to tools directory so it can be used by non-docker scripts
2. Also download prplmesh.buildinfo in `download_ipk.sh`


ci/run_easymesh_cert.sh
```diff
@@ -0,0 +1,115 @@
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2020 Tomer Eliyahu (Intel)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+#!/bin/bash
```
arnout: This has to be the first line of the script. Only the first 2 characters are read by the kernel, and based on that it decides that it's a shebang line.

ci/run_easymesh_cert.sh
```diff
@@ -0,0 +1,115 @@
...
+    printf '\033[1;31m%s\033[0m\n' "$*"
+}
+
+info() {
+    printf '\033[1;35m%s: %s\033[0m\n' "$(basename "$0")" "$*"
+}
+
+success() {
+    printf '\033[1;32m%s\033[0m\n' "$*"
+}
```
arnout: Why not use functions.sh? Well, it'd have to move to a better directory than tools/docker of course.

ci/run_easymesh_cert.sh
```diff
@@ -0,0 +1,115 @@
...
+    [ -n "$TESTS" ] || TESTS="$EASYMESH_CERT_PATH/tests/all_agent_tests.txt"
+    [ -n "$LOG_FOLDER" ] || LOG_FOLDER="$EASYMESH_CERT_PATH/logs/$(date +%F_%H-%M-%S)"
+    info "Logs location: $LOG_FOLDER"
+    info "Tests to run: $TESTS"
+
+    info "download latest ipk"
+    run "$PRPLMESH_PATH"/tools/download_ipk.sh --branch "$BRANCH"
+
+    info "deploy latest ipk"
+    run "$PRPLMESH_PATH"/tools/deploy_ipk.sh "$USER"@"$IPADDR" "$IPK"
```
arnout: Given that those two scripts are there, I think it's more appropriate to put this run_easymesh_cert script in the same directory. (nitpick)

Also I think it's worth to make a variable for the tools directory.

ci/run_easymesh_cert.sh
```diff
@@ -0,0 +1,115 @@
...
+    fi
+    info "Passed $passed/$total"
+    info "Uploading results to $UPLOAD_PATH not supported yet"
+    # TODO upload results
+    info "done"
+}
+
+USER=root
+BRANCH=master
+IPADDR=192.168.250.170
```
arnout: I really don't like that the IP address is hardcoded in the script, but there's not really an alternative I guess. Except, maybe it would make sense to default it to `"${TARGET_DEVICE}-1"`.

arnout (approved):
I don't really like commit messages that say `patch 1/2` - in the overall history, it's not usually relevant to consider something patch 1/2, and if it is, then probably it should have been a single patch.

So, I would rephrase it like this:

```
 ci: add run_easymesh_cert.sh

Add a shell script which can be used to run certification tests.
This is a simple wrapper script that includes downloading of the latest
prplmesh IPK from gitlab, deploying it to the target, and running the
certification tests.

This script will be used by a cron job in each testbed for running full
nightly certifications.

For now, uploading the logs to owncloud is left as a TODO.
```

ci/run_easymesh_cert.sh
tomereli: So we count on ~/.ssh/config. OK, make sense I guess.

ci/run_easymesh_cert.sh
```diff
@@ -0,0 +1,115 @@
...
+     OPTS=$(getopt -o 'hb:o:e:u:v' --long help,branch:,log-folder:,easymesh-cert:,upload:,verbose -n 'parse-options' -- "$@")
+
+    if [ $? != 0 ] ; then echo "Failed parsing options." >&2 ; usage; exit 1 ; fi
+
+    eval set -- "$OPTS"
+
+    while true; do
+        case "$1" in
+            -h | --help)            usage; exit 0;;
+            -b | --branch)          BRANCH="$2"; shift 2;;
```
rmelotte, line 49: Since there will be no way to know what the prplMesh version was once the logs are on Owncloud, please make sure to include the full hash of the commit it's running on somewhere.

rmelotte (approved):
Please run shellcheck on it and fix the issues (I know we're not running shellcheck in CI at the moment, but let's not introduce more issues).

ci/run_easymesh_cert.sh
tomereli: Added prplmesh.buildinfo

ci/run_easymesh_cert.sh
tomereli: Done.

ci/run_easymesh_cert.sh
tomereli: @rmelotte preferred to have a new folder `ci` since we will probably have more scripts related to CI in the future, so I left it as is. I did use a variable for the tools directory, however.

rmelotte: We could arguably move `ci` to `tools/ci`, but yes I think it makes sense to group the scripts related to CI in the same folder (we definitely have more to come).

PR #1006, opened 2020-03-24
CI: run flake8

arnout:
We want to validate PEP8 compliance in CI.
Since we currently have one file (tests/capi.py) that actually passes flake8, add a CI job for that one file.
More can be added once they start passing flake8 as well.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,6 +19,14 @@ cppcheck:
   tags:
     - docker
 
+flake8:
+  stage: build
+  image: pipelinecomponents/flake8:f087c4c
+  variables:
+      FLAKE8_FILES: "tests/capi.py"
+  script:
+    - flake8 --verbose $FLAKE8_FILES
+
 clang-format:
   extends: .in-prplmesh-builder
   stage: build
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -27,6 +27,13 @@ flake8:
   script:
     - flake8 --verbose $FLAKE8_FILES
 
+flake8-diff:
+  stage: build
+  image: pipelinecomponents/flake8:f087c4c
+  script:
+    - apk add --update --no-cache git
+    - git diff -U0 origin/master | flake8 --verbose --diff
+
 clang-format:
   extends: .in-prplmesh-builder
   stage: build
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,2 @@
+[flake8]
+max_line_length=100


.gitlab-ci.yml
```diff
@@ -19,6 +19,21 @@ cppcheck:
   tags:
     - docker
 
+flake8:
+  stage: build
+  image: pipelinecomponents/flake8:latest
```
rmelotte: I'm personally not a big fan of using "rolling" images like "latest", because it makes it harder to reproduce tests if the image was silently updated from one job to another.
On the other hand, it avoids the need to manually check and update images and it's frequently used, so I guess it's a split.

.gitlab-ci.yml
```diff
@@ -19,6 +19,21 @@ cppcheck:
...
+  variables:
+      FLAKE8_FILES: "tests/capi.py"
+  script:
+    - flake8 --verbose $FLAKE8_FILES
+
+flake8-diff:
+  stage: build
+  image: pipelinecomponents/flake8:latest
+  script:
+    - apk add git
```
rmelotte: I'm surprised this works without `--update`, unless the image already contains a cache.
I'm not too familiar with apk though.

Another option you might want to use regardless is `--no-cache`.

rmelotte (approved):
Awesome! :+1:

.gitlab-ci.yml
```diff
@@ -19,6 +19,21 @@ cppcheck:
...
+    - git diff -U0 origin/master | flake8 --verbose --diff
```
rmelotte, line 35: Can it always safely run on partial code taken out of their context?

tomereli (approved):

.gitlab-ci.yml
```diff
@@ -19,6 +19,21 @@ cppcheck:
   tags:
     - docker
 
+flake8:
+  stage: build
+  image: pipelinecomponents/flake8:latest
+  variables:
+      FLAKE8_FILES: "tests/capi.py"
+  script:
+    - flake8 --verbose $FLAKE8_FILES
```
MuteSpirit, line 28: Hi,

Just a note: there is a trick with "git ls-files" to get list list of files under source control and has needed extension
   See more here: https://github.com/koalaman/shellcheck/wiki/GitLab-CI

That will allow not hard-code Python files paths in CI related file

.gitlab-ci.yml
arnout: I guess they didn't clear the cache.

But indeed, going forward --update --no-cache is needed, because the cached reference may no longer be available.

.gitlab-ci.yml
arnout: It actually runs on the full file, it just only reports errors in the hunks mentioned by the diff.

.gitlab-ci.yml
arnout: Did you read the commit message? The reason to pass them explicitly is that most python files currently don't pass flake8, and we want to check only the ones that pass already.

.gitlab-ci.yml
rmelotte: That's awesome! :-)

.gitlab-ci.yml
MuteSpirit: Yes, i've read.
I've no objections for this PR.
My comment was not an issue for current PR but just a note which possibly you know to what future task to add.

PR #1009, opened 2020-03-24
use validity period in client disallow

kantera800:
check if validity period parameter is set in client_disallow message
then add it to the "disallowed client timeouts" list.
This list will be polled and when validity period is timed-out
CLIENT_ALLOW message will be sent to the bssid.

The unblocking in the client steering task FINALIZE state can
be removed since it is not really needed due to the validity period.
Unblocking will be triggered on in case client rejects the steering
request , since then there is no point of waiting till the end of the
period and we need to ALLOW all the client in all APs ASAP.

this commit will be split later on .

issue #445

Following test was performed:
platform: RDKB 
1. attach both 2 clients to 5G radio
2. steer client_1 to 2.4G radio
3. disallow client_1 from 5G message sent with validity period = 25 [sec] 
4. wait 5 sec
5. steer client_2  to 2.4G radio
6. disallow client_2 from 5G message sent with validity period = 25 [sec]

result:
->   t0: client_1 steered  successfully to 2.4G radio
->   t5: client_2 steered  successfully to 2.4G radio
-> t25: ALLOW client_1 on 5G message auto sent by the radio 
-> t30: ALLOW client_2 on 5G message auto sent by the radio

### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -693,6 +693,17 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
         std::string sta_mac = network_utils::mac_to_string(request->mac());
         std::string bssid   = network_utils::mac_to_string(request->bssid());
+
+        const auto &vap_unordered_map = ap_wlan_hal->get_radio_info().available_vaps;
+        auto it = std::find_if(vap_unordered_map.begin(), vap_unordered_map.end(),
+                               [&](const std::pair<int, bwl::VAPElement> &element) {
+                                   return element.second.mac == bssid;
+                               });
+
+        if (it == vap_unordered_map.end()) {
+            //AP does not have the requested vap, probably will be handled on the other AP
+            return true;
+        }
         LOG(DEBUG) << "CLIENT_DISALLOW: mac = " << sta_mac << ", bssid = " << bssid;
 
         ap_wlan_hal->sta_deny(sta_mac, bssid);
@@ -708,6 +719,18 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
         std::string sta_mac = network_utils::mac_to_string(request->mac());
         std::string bssid   = network_utils::mac_to_string(request->bssid());
+
+        const auto &vap_unordered_map = ap_wlan_hal->get_radio_info().available_vaps;
+        auto it = std::find_if(vap_unordered_map.begin(), vap_unordered_map.end(),
+                               [&](const std::pair<int, bwl::VAPElement> &element) {
+                                   return element.second.mac == bssid;
+                               });
+
+        if (it == vap_unordered_map.end()) {
+            //AP does not have the requested vap, probably will be handled on the other AP
+            return true;
+        }
+
         LOG(DEBUG) << "CLIENT_ALLOW: mac = " << sta_mac << ", bssid = " << bssid;
 
         ap_wlan_hal->sta_allow(sta_mac, bssid);
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1667,3 +1667,17 @@ void ap_manager_thread::send_steering_return_status(beerocks_message::eActionOp_
     }
     return;
 }
+
+void ap_manager_thread::remove_client_from_disallowed_list(const sMacAddr &mac,
+                                                           const sMacAddr &bssid)
+{
+    auto it = std::find_if(m_disallowed_clients.begin(), m_disallowed_clients.end(),
+                           [&](const son::ap_manager_thread::disallowed_client_t &element) {
+                               return ((element.mac == mac) && (element.bssid == bssid));
+                           });
+
+    if (it != m_disallowed_clients.end()) {
+        // remove client from the disallow list
+        m_disallowed_clients.erase(it);
+    }
+}
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -387,6 +387,9 @@ void ap_manager_thread::after_select(bool timeout)
                 now + std::chrono::seconds(HEARTBEAT_NOTIFICATION_DELAY_SEC);
         }
     }
+
+    // allow clients with expired blocking period timer
+    allow_expired_clients();
 }
 
 bool ap_manager_thread::socket_disconnected(Socket *sd)
@@ -704,9 +707,41 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             //AP does not have the requested vap, probably will be handled on the other AP
             return true;
         }
+
         LOG(DEBUG) << "CLIENT_DISALLOW: mac = " << sta_mac << ", bssid = " << bssid;
 
         ap_wlan_hal->sta_deny(sta_mac, bssid);
+
+        // Check if validity period is set then add it to the "disallowed client timeouts" list
+        // This list will be polled in after_select()
+        // When validity period is timed out sta_allow will be called.
+        if (request->validity_period_sec()) {
+
+            disallowed_client_t disallowed_client;
+
+            // calculate new disallow timeout from client validity period parameter [sec]
+            disallowed_client.timeout = std::chrono::steady_clock::now() +
+                                        std::chrono::seconds(request->validity_period_sec());
+            disallowed_client.mac   = request->mac();
+            disallowed_client.bssid = request->bssid();
+
+            // Remove old disallow period timeout from the list before inserting new
+            remove_client_from_disallowed_list(request->mac(), request->bssid());
+
+            // insert new disallow timeout to the list
+            m_disallowed_clients.push_back(disallowed_client);
+
+            LOG(DEBUG) << "client " << disallowed_client.mac
+                       << " will be allowed to accosiate with bssid " << disallowed_client.bssid
+                       << " in "
+                       << std::chrono::duration_cast<std::chrono::seconds>(
+                              disallowed_client.timeout - std::chrono::steady_clock::now())
+                              .count()
+                       << "sec";
+        } else {
+            LOG(WARNING) << "CLIENT_DISALLOW validity period set to 0, STA mac " << request->mac()
+                         << " will remain blocked from bssid " << request->bssid();
+        }
         break;
     }
     case beerocks_message::ACTION_APMANAGER_CLIENT_ALLOW_REQUEST: {
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -766,9 +766,11 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             return true;
         }
 
-        LOG(DEBUG) << "CLIENT_ALLOW: mac = " << sta_mac << ", bssid = " << bssid;
+        remove_client_from_disallowed_list(request->mac(), request->bssid());
 
+        LOG(DEBUG) << "CLIENT_ALLOW: mac = " << sta_mac << ", bssid = " << bssid;
         ap_wlan_hal->sta_allow(sta_mac, bssid);
+
         break;
     }
     case beerocks_message::ACTION_APMANAGER_READ_ACS_REPORT_REQUEST: {
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1681,3 +1681,18 @@ void ap_manager_thread::remove_client_from_disallowed_list(const sMacAddr &mac,
         m_disallowed_clients.erase(it);
     }
 }
+
+void ap_manager_thread::allow_expired_clients()
+{
+    // check if any client disallow period has expired and allow it.
+    for (auto it = m_disallowed_clients.begin(); it != m_disallowed_clients.end();) {
+        if (std::chrono::steady_clock::now() > it->timeout) {
+            LOG(DEBUG) << "CLIENT_ALLOW: mac = " << it->mac << ", bssid = " << it->bssid;
+            ap_wlan_hal->sta_allow(network_utils::mac_to_string(it->mac),
+                                   network_utils::mac_to_string(it->bssid));
+            m_disallowed_clients.erase(it);
+        } else {
+            it++;
+        }
+    }
+}
--- a/agent/src/beerocks/slave/ap_manager_thread.h
+++ b/agent/src/beerocks/slave/ap_manager_thread.h
@@ -47,6 +47,18 @@ public:
         eBackhaulVapType type;
     };
 
+    /**
+     * disallowed client parameters
+     * Used to save clients mac, bssid that the client is disallowed from and 
+     * validity period of time (blocking) 
+     * so we could unblock it when the period expires
+     */
+    struct disallowed_client_t {
+        sMacAddr mac;
+        sMacAddr bssid;
+        std::chrono::steady_clock::time_point timeout;
+    };
+
     enum eThreadErrors : uint32_t {
         APMANAGER_THREAD_ERROR_NO_ERROR            = 0,
         APMANAGER_THREAD_ERROR_HOSTAP_DISABLED     = 1,
@@ -82,6 +94,7 @@ private:
     bool low_filter;
     int bss_steer_valid_int;
     int bss_steer_imminent_valid_int;
+    std::vector<disallowed_client_t> m_disallowed_clients;
 
     std::list<backhaul_vap_list_element_t> backhaul_vaps_list;
 
--- a/agent/src/beerocks/slave/ap_manager_thread.h
+++ b/agent/src/beerocks/slave/ap_manager_thread.h
@@ -88,6 +88,7 @@ private:
     void send_steering_return_status(beerocks_message::eActionOp_APMANAGER ActionOp,
                                      int32_t status);
     void remove_client_from_disallowed_list(const sMacAddr &mac, const sMacAddr &bssid);
+    void allow_expired_clients();
 
     std::string slave_uds;
     uint8_t wifi_channel;
--- a/agent/src/beerocks/slave/ap_manager_thread.h
+++ b/agent/src/beerocks/slave/ap_manager_thread.h
@@ -87,6 +87,7 @@ private:
     void send_heartbeat();
     void send_steering_return_status(beerocks_message::eActionOp_APMANAGER ActionOp,
                                      int32_t status);
+    void remove_client_from_disallowed_list(const sMacAddr &mac, const sMacAddr &bssid);
 
     std::string slave_uds;
     uint8_t wifi_channel;
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4305,8 +4305,9 @@ bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::Cmd
             return false;
         }
 
-        request_out->mac()   = sta_mac;
-        request_out->bssid() = bssid;
+        request_out->mac()                 = sta_mac;
+        request_out->bssid()               = bssid;
+        request_out->validity_period_sec() = association_control_request_tlv->validity_period_sec();
     }
 
     message_com::send_cmdu(ap_manager_socket, cmdu_tx);
--- a/common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_apmanager.h
+++ b/common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_apmanager.h
@@ -625,6 +625,7 @@ class cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST : public BaseClass
         }
         sMacAddr& mac();
         sMacAddr& bssid();
+        uint16_t& validity_period_sec();
         void class_swap() override;
         bool finalize() override;
         static size_t get_initial_size();
@@ -634,6 +635,7 @@ class cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST : public BaseClass
         eActionOp_APMANAGER* m_action_op = nullptr;
         sMacAddr* m_mac = nullptr;
         sMacAddr* m_bssid = nullptr;
+        uint16_t* m_validity_period_sec = nullptr;
 };
 
 class cACTION_APMANAGER_CLIENT_ALLOW_REQUEST : public BaseClass
--- a/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
+++ b/common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
@@ -140,6 +140,7 @@ cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST:
   _type: class
   mac: sMacAddr
   bssid: sMacAddr
+  validity_period_sec: uint16_t
 
 cACTION_APMANAGER_CLIENT_ALLOW_REQUEST:
   _type: class
--- a/controller/src/beerocks/master/son_management.cpp
+++ b/controller/src/beerocks/master/son_management.cpp
@@ -309,8 +309,6 @@ void son_management::handle_cli_message(Socket *sd,
             network_utils::mac_from_string(hostap_mac);
         association_control_request_tlv->association_control() =
             wfa_map::tlvClientAssociationControlRequest::UNBLOCK;
-        //TODO: Get real validity_period_sec
-        association_control_request_tlv->validity_period_sec() = 1;
         association_control_request_tlv->alloc_sta_list();
         auto sta_list         = association_control_request_tlv->sta_list(0);
         std::get<1>(sta_list) = network_utils::mac_from_string(client_mac);
--- a/common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_apmanager.cpp
+++ b/common/beerocks/tlvf/AutoGenerated/src/beerocks/tlvf/beerocks_message_apmanager.cpp
@@ -2008,11 +2008,16 @@ sMacAddr& cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST::bssid() {
     return (sMacAddr&)(*m_bssid);
 }
 
+uint16_t& cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST::validity_period_sec() {
+    return (uint16_t&)(*m_validity_period_sec);
+}
+
 void cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST::class_swap()
 {
     tlvf_swap(8*sizeof(eActionOp_APMANAGER), reinterpret_cast<uint8_t*>(m_action_op));
     m_mac->struct_swap();
     m_bssid->struct_swap();
+    tlvf_swap(16, reinterpret_cast<uint8_t*>(m_validity_period_sec));
 }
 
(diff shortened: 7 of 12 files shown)


agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -708,6 +737,51 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        if (request->validity_period()) {
+
+            disallowed_client_params_t disallowed_client_params;
+            disallowed_client_params.mac   = request->mac();
+            disallowed_client_params.bssid = request->bssid();
+
+            // check if STA is already disallowed from this bssid
+            auto disallowed_client_timeout = std::find_if(m_disallowed_client_timeouts.begin(), m_disallowed_client_timeouts.end(),
+                        [&](const std::pair<std::chrono::steady_clock::time_point, disallowed_client_params_t> &element) {
+                            return ((element.second.mac == disallowed_client_params.mac) && (element.second.bssid == disallowed_client_params.bssid));
```
arnout: Unless I'm mistaken, the default operator== of disallowed_client_params_t should do exactly this. In other words:
```suggestion
                            return ((element.second == disallowed_client_params));
```

However, my other comment about adding timeout to the struct will make this no longer true, unless you add an explicit operator== that ignores the timeout.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -708,6 +737,51 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        // Check if validity period is set then add it to the "disallowed client timeouts" list
+        // This list will be polled in after_select()
+        // When validity period is timed out CLIENT_ALLOW message will be sent.
+        if (request->validity_period()) {
+
+            disallowed_client_params_t disallowed_client_params;
+            disallowed_client_params.mac   = request->mac();
+            disallowed_client_params.bssid = request->bssid();
+
+            // check if STA is already disallowed from this bssid
```
arnout: You'll need to do the same in the handling of the ALLOW case, so better factor this into a function.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -708,6 +737,51 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+                        });
+            if (disallowed_client_timeout != m_disallowed_client_timeouts.end()) {
+                // remove old disallow timeout from the list before inserting the new one
+                m_disallowed_client_timeouts.erase(disallowed_client_timeout);
+                LOG(DEBUG) << "remove old disallow timeout from the list before inserting the new one";
+            }
+
+            // calculate new disallow timeout from client validity period parameter [sec]
+            // when timeout expires allow message will be sent to the client.
+            auto disallow_timeout_in_sec =
```
arnout: The `_in_sec` bit is not correct, since this will be a timepoint, not a value in seconds.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -708,6 +737,51 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+                std::chrono::steady_clock::now() + std::chrono::seconds(request->validity_period());
+            
+            LOG(DEBUG) << "insert new disallow timeout to the list";
+            // insert new disallow timeout to the list
+            m_disallowed_client_timeouts.insert(
+                std::pair<std::chrono::steady_clock::time_point, disallowed_client_params_t>(
+                    disallow_timeout_in_sec, disallowed_client_params));
```
arnout: Wouldn't it be simpler to add the timeout to disallowed_client_params_t instead of using pairs?

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -708,6 +737,51 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+            // for (auto it = m_disallowed_client_timeouts.begin(); it != m_disallowed_client_timeouts.end(); it++)
+            // {
+            //     LOG(DEBUG) << " STA mac " << it->second.mac << " bssid " << it->second.bssid
+            //                << " disallow timeout "
+            //                << std::chrono::duration_cast<std::chrono::seconds>(
+            //                       it->first - std::chrono::steady_clock::now())
+            //                       .count();
+            // }
+        } else {
+            LOG(WARNING) << "CLIENT_DISALLOW validity period set to 0, STA mac " << request->mac() << " will remain blocked from bssid " << request->bssid();
```
arnout: Nice!

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -401,6 +401,35 @@ void ap_manager_thread::after_select(bool timeout)
                 now + std::chrono::seconds(HEARTBEAT_NOTIFICATION_DELAY_SEC);
         }
     }
+
+    // check if any client disallow period has expired
+    // in case it did we need to removed the client from blacklist
+    if (!m_disallowed_client_timeouts.empty()) {
+        // map key holds the disallow timout
+        // since its an ordered map check always the first element (lowest value)
+        auto it                          = m_disallowed_client_timeouts.begin();
```
arnout: If there are multiple timeouts running, it is possible and even likely (*) that more than one expires at the same time. So you really have to iterate here.

Because of that, I'd think it's easier to use a std::vector and just iterate over everything.

(*) Likely because: either we're steering just a single client, and then there will be only one element in this list. Or we're steering all clients away, e.g. because we're going to switch channels, and in that case all the timeouts will expire at the same time. So the only situation where we have multiple elements in the list that don't expire at the same time is when a second client is steered a few seconds after the first one. It happens, but not very likely.

arnout (commented):
I know you asked for "no nitpick comments", but my comments really aren't fundamental so they can be considered nitpicks :-) I really like the direction in which this is going!

I would also like an update to test_flows where this functionality is tested, by checking after the validity period that the client is unblocked on its original BSS.

arnout (commented):
Since it is possible that at some point, the validity period will be handled by hostapd itself, maybe it would be better to add it as an argument to sta_deny() and let the return value of sta_deny() indicate if it will handle validity period itself or not.

But actually, that's something we can do when there's a bwl that does handle validity period internally, so never mind.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -140,6 +140,7 @@ cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST:
   _type: class
   mac: sMacAddr
   bssid: sMacAddr
+  validity_period: uint16_t
```
arnout: (nitpick) can we add an explicit unit here?
```suggestion
  validity_period_sec: uint16_t
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4306,8 +4307,9 @@ bool slave_thread::handle_client_association_request(Socket *sd, ieee1905_1::Cmd
             return false;
         }
 
-        request_out->mac()   = sta_mac;
-        request_out->bssid() = bssid;
+        request_out->mac()             = sta_mac;
+        request_out->bssid()           = bssid;
+        request_out->validity_period() = validity_period;
```
arnout: Why do we need an additional variable here?
```suggestion
        request_out->validity_period() = association_control_request_tlv->validity_period_sec();
```

agent/src/beerocks/slave/ap_manager_thread.h
```diff
@@ -47,6 +47,12 @@ class ap_manager_thread : public beerocks::socket_thread {
         eBackhaulVapType type;
     };
 
+    struct disallowed_client_t {
```
arnout, line 56: Note: although https://www.strawpoll.me/19651625/r indicates that we will use `sDisallowedClient`, I think consistency is more important so it should indeed be like this.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -705,6 +705,17 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
         std::string bssid   = network_utils::mac_to_string(request->bssid());
+
+        const auto &vap_unordered_map = ap_wlan_hal->get_radio_info().available_vaps;
+        auto it = std::find_if(vap_unordered_map.begin(), vap_unordered_map.end(),
+                               [&](const std::pair<int, bwl::VAPElement> &element) {
+                                   return element.second.mac == bssid;
+                               });
+
+        if (it == vap_unordered_map.end()) {
+            //AP does not have the requested vap, probably will be handled on the other AP
```
arnout, line 707: (nitpick) IMO "AP" is not a very accurate term, I prefer to use "radio" which (IMO again) is less likely to be confused with VAP.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -156,6 +156,21 @@ static bool enable_backhaul_vap(
     return false;
 }
 
+static void remove_client_from_disallowed_list(
+    std::vector<son::ap_manager_thread::disallowed_client_t> &disallowed_clients,
```
arnout: This can only be m_disallowed_clients, right? I think it would be more appropriate to add it as a private member function of ap_manager_thread. (nitpick, and maybe I'm wrong)

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -416,6 +416,23 @@ void ap_manager_thread::after_select(bool timeout)
...
+                LOG(DEBUG) << "CLIENT_ALLOW: mac = " << m_disallowed_clients[i].mac
+                           << ", bssid = " << m_disallowed_clients[i].bssid;
+                // send allow client msg to bssid
+                ap_wlan_hal->sta_allow(network_utils::mac_to_string(m_disallowed_clients[i].mac),
+                                       network_utils::mac_to_string(m_disallowed_clients[i].bssid));
+                // erase client from disallowed list
+                m_disallowed_clients.erase(m_disallowed_clients.begin() + i);
+            }
+        }
+    }
```
arnout: I don't understand why this has to be so complicated...
```suggestion
    for (auto it = m_disallowed_clients.begin; it != m_disallowed_clients.end(); ) {
        if (std::chrono::steady_clock::now() > it->timeout) {
            LOG(DEBUG) << "CLIENT_ALLOW: mac = " << m_disallowed_clients[i].mac
                           << ", bssid = " << m_disallowed_clients[i].bssid;
            // allow client on bssid
            ap_wlan_hal->sta_allow(network_utils::mac_to_string(it->mac),
                                   network_utils::mac_to_string(it->bssid));
            // erase client from disallowed list
            m_disallowed_clients.erase(m_disallowed_clients.begin() + i);
        } else {
            it++;
        }
    }
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -731,9 +731,42 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
             return true;
         }
+
         LOG(DEBUG) << "CLIENT_DISALLOW: mac = " << sta_mac << ", bssid = " << bssid;
 
         ap_wlan_hal->sta_deny(sta_mac, bssid);
+
+        // Check if validity period is set then add it to the "disallowed client timeouts" list
+        // This list will be polled in after_select()
+        // When validity period is timed out CLIENT_ALLOW message will be sent.
```
arnout: There's no message that is sent, we just call the sta_allow bwl function (which indeed sends a CLIENT_ALLOW command to hostapd, but that's hidden in the bwl layer). (nitpick)

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -808,9 +808,12 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             return true;
         }
 
-        LOG(DEBUG) << "CLIENT_ALLOW: mac = " << sta_mac << ", bssid = " << bssid;
+        // remove client from disallowed list
```
arnout: This comment is literally saying what is right below it, so I really don't think it adds value. @RanRegev disagrees with me though :-)

arnout (approved):
Excellent PR with good commit messages. The only thing that is missing is an update of test_flows :-)

agent/src/beerocks/slave/ap_manager_thread.cpp
RanRegev: The point is not to disagree... The idea is to tell a story with the comments. IMO one should be able to understand the intent of the code without reading the code itself, just by reading the comments. And yes, this approach sometimes results in stating the obvious. But sometimes few lines of code are doing something that is summed up in a single comment. When functionality is not trivial this 'story' expressed in comments is very useful for later maintenance.
For example, it is not uncommon for me to write a code like
`// do we have enough?`
`if ( enough ){`
`// it seems that we are good, continue...`
`}`
`else {`
`// we don't have enough, lets try another approach`
`}`
In this specific case that we are chatting on: the comment is completely redundant and should be removed as there are no other comments inside this case clause that supports this paradigm.

ghost: No, I think that's wrong. We're programmers, we know how to read code (so long as it's written sanely, of course).
Comments should describe what happens at a higher layer than can be seen in the code directly. As such, they should only describe overall function behavior and how that function is used, rather than what it does. (i.e. instead of having a bunch of "remove item from array"-type comments within the function, the function should just have a description like "disassociates agent from the controller and removes its credentials so that it has to pair again").
Comments within functions  should be avoided, and only used to explain especially tricky, non-intuitive code (which should also be avoided if possible, of course).

"Telling a story with comments" just means the code is now telling two stories, and woe betide you if you let them get out of sync

kantera800:
Slipped my mind, will add a commit for that.

agent/src/beerocks/slave/ap_manager_thread.cpp
RanRegev: I agree, I don't think we have an argue here. I don't comment trivial code that can be read as is. As you said, there is  no point in doing so. However, sometimes the code is not trivial even if it doesn't do tricky things. It just has its logic that might be berried in few lines of technical details that have nothing to do with the logic itself. In these cases I find it extremely maintainable to comment the different sections in the function and their logic.
A small example can be seen in the place this discussion was started: https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/son_slave_thread.cpp#L4373-L4405
This else clause starts with a story about what is done in it. In this case not what would you expect. It also states why it is fine and suggest future changes.
Later in the code the promise that was taken in the initial comment is implemented.
I find it useful to read the first comment, then look forward in the code for _more_ comments that support the story on top, and only then dive into the code to ensure it does what it says it does.

All in all commenting code is a very delicate process that its benefits are revealed only *over time*. Comments are not compiled therefore their target is not correct code. They serve as another mean for maintaining a code base.
One indication for poor commenting process is that developers don't read comments. It means that they get used to comments that are irrelevant, hard to read or just outdated. This situation is in many cases worse then non commented code at all, because developers, seeking for assistance are reading misleading comments.

tomereli: Bottom line - comments in **prplMesh** are saying what **isn't obvious** from the code, and contributors are expected to align to this approach.

arnout: I think @RanRegev's long remark is very valuable and it should be put in a new "Comments" section in CODINGSTYLE.md. @RanRegev can you do that?

RanRegev: #1062

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -140,6 +140,7 @@ cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST:
   _type: class
   mac: sMacAddr
   bssid: sMacAddr
+  validity_period_sec: uint16_t
```
tomereli, line 143: Please explain a bit more in the commit message - at least what is the validity period and where it will be used.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_apmanager.h
```diff
@@ -625,6 +625,7 @@ class cACTION_APMANAGER_CLIENT_DISALLOW_REQUEST : public BaseClass
         }
         sMacAddr& mac();
         sMacAddr& bssid();
+        uint16_t& validity_period_sec();
```
tomereli, line 628: This is autogenerated code - commit message should mention that. In this case it should however be squashed to the previous commit, since its a small change so easy to review.

controller/src/beerocks/master/son_management.cpp
```diff
@@ -309,8 +309,12 @@ void son_management::handle_cli_message(Socket *sd,
...
         association_control_request_tlv->association_control() =
             wfa_map::tlvClientAssociationControlRequest::UNBLOCK;
-        //TODO: Get real validity_period_sec
-        association_control_request_tlv->validity_period_sec() = 1;
+        /*
+        According to section 11.6 in the Multi-AP Specification
+        The Validity Period field in a Client Association Control Request message 
+        with Association Control field set to 0x01 (Client Unblocking) is ignored
+        */
+        association_control_request_tlv->validity_period_sec() = 0;
```
tomereli: So this needs to be defined in the YAML file - you can add `_value = 0` which will be the default value.

agent/src/beerocks/slave/ap_manager_thread.h
tomereli, line 56: For new structs - please add a Doxygen comment.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -693,6 +693,17 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
         std::string sta_mac = network_utils::mac_to_string(request->mac());
         std::string bssid   = network_utils::mac_to_string(request->bssid());
+
+        const auto &vap_unordered_map = ap_wlan_hal->get_radio_info().available_vaps;
```
tomereli, line 700: The param name shouldn't include the param type - it doesn't help make the code more readable.
Use `const auto &vaps` instead.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -693,6 +693,17 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        auto it = std::find_if(vap_unordered_map.begin(), vap_unordered_map.end(),
+                               [&](const std::pair<int, bwl::VAPElement> &element) {
+                                   return element.second.mac == bssid;
+                               });
+
+        if (it == vap_unordered_map.end()) {
```
tomereli, line 706: This is wrong - please move the handling to the backhaul manager where we have the list of VAPs and associated clients, then send the message only to the AP to which the client is connected. Since its a bigger change, you can do this in a separate PR - but make sure to add a task for it and TODO in the code. We should stop sending messages which may be ignored directly to the son slaves.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1667,3 +1667,18 @@ void ap_manager_thread::send_steering_return_status(beerocks_message::eActionOp_
     }
     return;
 }
+
+void ap_manager_thread::remove_client_from_disallowed_list(
```
tomereli: Big like for adding helper functions such as this one - but if you do, do it for all operations:
`allow_client`
`dissallow_client`
`is_client_allowed`

These names are better IMO.

agent/src/beerocks/slave/ap_manager_thread.h
```diff
@@ -81,6 +81,9 @@ class ap_manager_thread : public beerocks::socket_thread {
     void send_heartbeat();
     void send_steering_return_status(beerocks_message::eActionOp_APMANAGER ActionOp,
                                      int32_t status);
+    void remove_client_from_disallowed_list(
+        std::vector<son::ap_manager_thread::disallowed_client_t> &disallowed_clients,
```
tomereli: No - why do you pass the vector to the function? Isn't it class parameter? the function should get only the client MAC and optionally the bssid.
So, in this case, two helper functions are needed:
```
bool allow_client(const sMacAddr &client_mac); //allows client globally for the radio
bool allow_client(const sMacAddr &client_mac, const sMacAddr &bssid); //allows client for a specific BSSID
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -387,6 +387,19 @@ void ap_manager_thread::after_select(bool timeout)
                 now + std::chrono::seconds(HEARTBEAT_NOTIFICATION_DELAY_SEC);
         }
     }
+
+    // check if any client disallow period has expired and allow it.
+    for (auto it = m_disallowed_clients.begin(); it != m_disallowed_clients.end();) {
```
tomereli: perhaps you can move this to a helper function - `allow_expired_clients()`

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -704,9 +717,42 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+            disallowed_client_t disallowed_client;
+
+            // calculate new disallow timeout from client validity period parameter [sec]
+            disallowed_client.timeout = std::chrono::steady_clock::now() +
+                                        std::chrono::seconds(request->validity_period_sec());
+            disallowed_client.mac   = request->mac();
+            disallowed_client.bssid = request->bssid();
+
+            // Remove old disallow period timeout from the list before inserting new
+            remove_client_from_disallowed_list(m_disallowed_clients, request->mac(),
```
tomereli: No need to pass a class member to a class function.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -777,9 +777,11 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             return true;
         }
 
-        LOG(DEBUG) << "CLIENT_ALLOW: mac = " << sta_mac << ", bssid = " << bssid;
+        remove_client_from_disallowed_list(m_disallowed_clients, request->mac(), request->bssid());
 
+        LOG(DEBUG) << "CLIENT_ALLOW: mac = " << sta_mac << ", bssid = " << bssid;
```
tomereli, line 771: This log should be moved to `remove_client_from_dissallowed_list()`

tomereli (changes requested):
This is a really good PR @kantera800! I still had some major comments so asking for changes.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_apmanager.h
kantera800: "rendered" means exactly that

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: you are probably right , but its used like this several times in ap_manager_thread and i didn't want to break the consistency.

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: I disagree, this should be a part of  to the "unified agent" task  that @morantr is working on right now.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_apmanager.h
tomereli: We use "autogenerated" in all commits, don't change that please.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: By all means - break it!

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: If its a task to the unified agent - then it means that until the unified agent is ready this should be in the backhaul manager, so I'm not sure why you disagree.
If you need to handle error cases this is a bug - for example in the sta measurements flow, we need to reply with an error TLV if the STA is not connected to any of the BSSes in the agent, so this is why it has to be in the backhaul manager.
Since this is part of EasyMesh steering - its the same. Anyway, it can wait to the unified agent.

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: no,
1. since there is chance that remove_client_from_dissallowed_list() doesn't find the client on the list.
2. ALLOW message not actual sent in remove_client_from_dissallowed_list() so this log could be misleading.

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: unnecessary since the "remove" helper function is the only one that is used more than once. 

About the naming ,
 `allow_client` is misleading since removing a client from this list **doesn't** always mean the client will be automatically allowed it could be added again with different parameters (timeout, target bssid...)

agent/src/beerocks/slave/ap_manager_thread.h
kantera800: you are right , i removed the vector.
Although, bssid parameter is always needed since there could be multiple VAPs on this radio that the client could be blocked from.

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: I'm not sure its needed since this is checked only once, also its pretty clear from code

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
kantera800: Sure I'll add this message  after you approve the fixups to not ruin the history.

"Time period in seconds (from reception of the Client Association ControlRequest message) for which a blocking request is valid "

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: Even if it is only used once - it still make sense to put in a helper function simply because its more readable.
Naming - you choose then.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: As I said before I think it makes the code more readable, its not about optimization.

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: I prefer not to since Its not related to this PR. 
also, it should be removed by #1082

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_apmanager.h
kantera800: sure I'll fix the commit title after you approve fixups

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: missed it, sure i'll add it after you approve the fixups

tomereli (approved):

PR #1014, opened 2020-03-25
Agent: Variable overflow in fill_cs_params

### LINKED ISSUE #1013: [BUG] Agent: Variable overflow in fill_cs_params function
**Describe the bug** -
Problem in function [`fill_cs_params`](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/ap_manager_thread.cpp#L984) in string `params.tx_power  = ap_wlan_hal->get_radio_info().tx_power`.
Variable `param.tx_power` has type` int8_t` (1 byte), variable` ap_wlan_hal->get_radio_info().tx_power `has type` int` (4 bytes), and when value in variable `p_wlan_hal->get_radio_info().tx_power` greater than 127 it cause a problem - type overflow and variable `params.tx_power` contains garbage instead correct `tx_power` value. 

Example where `tx_power` contains value greater than 127 found  function [` set_tx_power`](https://github.com/prplfoundation/prplMesh/blob/master/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp#L328), string `m_radio_info.tx_power = tx_pow_limit * 100;` Here both variables `m_radio_info.tx_power `and` tx_pow_limit` have type `int`.

The issue was found in dummy mode in handling the message `ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START`, where the first call - function [`set_tx_power`](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/ap_manager_thread.cpp#L575) then, creating a notification message with data, which fills up (with garbage in `tx_power`) by [`fill_cs_params`](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/ap_manager_thread.cpp#L585)` function.

As a result, the agent gets  [`ACTION_APMANAGER_HOSTAP_CSA_NOTIFICATION`](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/son_slave_thread.cpp#L1800) message and after handling sends [operating channel report](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/son_slave_thread.cpp#L1822) with garbage in `tx_power`.

## Steps to reproduce

1. Build
2. Run controller & agent
3. Send channel selection message with payload( channel list and tx power limit)
4. Check the value of tx_power after `CHANNEL_SELECTION `message in the log for the agent and after `OPERATIN_CHANNEL_REPORT `message in the controller - values different.

## Expected behavior

The values of tx_power should be the same in both log files for controller and agent.

## Target

- Host: Ubuntu 18.04
- Platform: Linux DUMMY mode with ZMQ
- How you built prplMesh - ./docker-builder.sh
- How you ran prplMesh - tests/test_flow.py

## Additional context

According to the documentation transmit power limit TLV consists of:  tlvType - 1 byte, tlvLength - 2 bytes, tlvValue - 6 bytes for MAC, 1 byte for transmit power limit.

VladyslavTupikin:
Function set_tx_power_limit break the logic for
tx_power variable when multiple tx_pow_limit on 100
and assign it to the m_radio_info.tx_power, because in
other cases as example in the function fill_cs_params
tx_power has type int8_t, and it triggers overflow when
value in tx_power greater than 127.

Solution - remove multiplication from set_tx_power and
add type conversion from int to int8_t in fill_cs_params
for tx_power.

### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -983,7 +983,7 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
 void ap_manager_thread::fill_cs_params(beerocks_message::sApChannelSwitch &params)
 {
-    params.tx_power  = (int8_t)ap_wlan_hal->get_radio_info().tx_power;
+    params.tx_power  = static_cast<int8_t>(ap_wlan_hal->get_radio_info().tx_power);
     params.channel   = ap_wlan_hal->get_radio_info().channel;
     params.bandwidth = uint8_t(
         beerocks::utils::convert_bandwidth_to_enum(ap_wlan_hal->get_radio_info().bandwidth));
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -983,7 +983,7 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
 void ap_manager_thread::fill_cs_params(beerocks_message::sApChannelSwitch &params)
 {
-    params.tx_power  = ap_wlan_hal->get_radio_info().tx_power;
+    params.tx_power  = (int8_t)ap_wlan_hal->get_radio_info().tx_power;
     params.channel   = ap_wlan_hal->get_radio_info().channel;
     params.bandwidth = uint8_t(
         beerocks::utils::convert_bandwidth_to_enum(ap_wlan_hal->get_radio_info().bandwidth));
--- a/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
@@ -327,8 +327,8 @@ bool ap_wlan_hal_dummy::read_supported_channels() { return true; }
 
 bool ap_wlan_hal_dummy::set_tx_power_limit(int tx_pow_limit)
 {
-    LOG(TRACE) << __func__ << " setting power limit: " << tx_pow_limit * 100 << " mBm";
-    m_radio_info.tx_power = tx_pow_limit * 100;
+    LOG(TRACE) << __func__ << " setting power limit: " << tx_pow_limit;
+    m_radio_info.tx_power = tx_pow_limit;
     return true;
 }
 
--- a/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
@@ -327,7 +327,8 @@ bool ap_wlan_hal_dummy::read_supported_channels() { return true; }
 
 bool ap_wlan_hal_dummy::set_tx_power_limit(int tx_pow_limit)
 {
-    LOG(TRACE) << __func__ << " setting power limit: " << tx_pow_limit;
+    LOG(TRACE) << __func__ << " setting power limit: " << static_cast<int8_t>(tx_pow_limit)
+               << " dBm";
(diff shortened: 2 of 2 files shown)


agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -983,7 +983,7 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
 
 void ap_manager_thread::fill_cs_params(beerocks_message::sApChannelSwitch &params)
 {
-    params.tx_power  = ap_wlan_hal->get_radio_info().tx_power;
+    params.tx_power  = (int8_t)ap_wlan_hal->get_radio_info().tx_power;
```
tomereli:
```suggestion
    params.tx_power  = static_cast<int8_t>(ap_wlan_hal->get_radio_info().tx_power);
```

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -327,8 +327,8 @@ bool ap_wlan_hal_dummy::read_supported_channels() { return true; }
 
 bool ap_wlan_hal_dummy::set_tx_power_limit(int tx_pow_limit)
 {
-    LOG(TRACE) << __func__ << " setting power limit: " << tx_pow_limit * 100 << " mBm";
-    m_radio_info.tx_power = tx_pow_limit * 100;
+    LOG(TRACE) << __func__ << " setting power limit: " << tx_pow_limit;
```
tomereli:
```suggestion
    LOG(TRACE) << __func__ << " setting power limit: " << tx_pow_limit << " dBM";
```

tomereli (approved):
Please fix the C style cast to a static cast.
In addition, the commit message only says there is an overflow due to the fact we multiply the tx power value by 100, which is true, but doesn't explain why this is done - the reason is that the actual NL command expects the tx power in mBm which is why we multiply this by 100 in the actual implementation, but when we read the value from hostapd, it is in dBM so we don't need to divide by 100.
In the dummy implementation though, there is a single API which already sets the dBM value, and it was wrong to multiply by 100 in the first place.

Other than this, LGTM!

agent/src/beerocks/slave/ap_manager_thread.cpp
arnout: Actually, for conversions between scalar types, it [should use `T{e}` construction notation](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-construct).
```suggestion
    params.tx_power  = int8_t{ap_wlan_hal->get_radio_info().tx_power};
```

However, this particular one is redundant, because `params.tx_power` is already an int8_t. For the other two it's needed though.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
arnout:
```suggestion
    LOG(TRACE) << __func__ << " setting power limit: " << int8_t{tx_pow_limit} << " dBm";
```

arnout (approved):

arnout:
Forgot to mention: nice catch!

tomereli:
@VladyslavTupikin please fix the comments, then add the `ready for merge` tag so that `mergify` will merge this PR automatically.

arnout:
@VladyslavTupikin *Not* with fixup commits if it's intended to be merged. I removed the don't merge tag again since I'm not sure if mergify takes into account the DCO check.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
arnout: This apparently doesn't work...

abelog (approved):

arnout:
If you have the time, it would be nice if you could also add something to test_flows.py that captures this error. Or is that covered by #965?

PR #1018, opened 2020-03-25
ci: update run_easymesh_cert.sh to support upload to owncloud (Patch 2)

tomereli:
Add a script for uploading file[s] / folders to owncloud via curl.
(based on https://doc.owncloud.com/server/user_manual/files/access_webdav.html#accessing-files-using-curl) and update `run_easymesh_cert.sh` to support uploading to owncloud.

Example logs uploaded by the script:
https://ftp.essensium.com/owncloud/index.php/apps/files/?dir=/Nightly/agent_certification/2020-03-25_19-19-40&fileid=712397

### DIFF
--- a/ci/run_easymesh_cert.sh
+++ b/ci/run_easymesh_cert.sh
@@ -22,9 +22,10 @@ usage() {
     echo "      -b|--branch - prplmesh branch to use (default master latest)"
     echo "      -o|--log-folder - path to put the logs to (make sure it does not contain previous logs)."
     echo "      -e|--easymesh-cert - path to easymesh_cert repository (default ../easymesh_cert)"
-    echo "      -u|--upload - upload results"
+    echo "      --owncloud-upload - whether or not to upload results to owncloud. (default: false)"
+    echo "      --owncloud-path - relative path in the owncloud server to upload results to. (default: $OWNCLOUD_PATH)"
     echo "      -v|--verbose - set verbosity (ucc logs also redirected to stdout)"
-    echo "      -s|--ssh - target device ssh name (defined in ~/.ssh/config). Default is $TARGET_DEVICE_SSH"
+    echo "      -s|--ssh - target device ssh name (defined in ~/.ssh/config). (default: $TARGET_DEVICE_SSH)"
     echo "'test' can either be a test name, or the path to a file containing"
     echo " test names (newline-separated)."
     echo
@@ -34,10 +35,15 @@ usage() {
     echo "$(basename "$0") MAP-5.3.1 tests/all_controller_test.txt"
     echo ""
     echo "The default is to run all agent certification tests."
+    echo ""
+    echo "Credentials for uploading to owncloud are taken from $HOME/.netrc"
+    echo "Make sure it has a line like this:"
+    echo "  machine ftp.essensium.com login <user> password <password>"
+    echo ""
 }
 
 main() {
-     OPTS=$(getopt -o 'hb:o:e:u:s:v' --long help,branch:,log-folder:,easymesh-cert:,upload:,ssh:verbose -n 'parse-options' -- "$@")
+    OPTS=$(getopt -o 'hb:o:e:s:v' --long help,branch:,log-folder:,easymesh-cert:,owncloud-upload,owncloud-path:,ssh:,verbose -n 'parse-options' -- "$@")
 
     if [ $? != 0 ] ; then echo "Failed parsing options." >&2 ; usage; exit 1 ; fi
 
@@ -50,7 +56,8 @@ main() {
             -o | --log-folder)      LOG_FOLDER="$2"; shift 2;;
             -v | --verbose)         VERBOSE=true; VERBOSE_OPT="-v"; shift;;
             -e | --easymesh-cert)   EASYMESH_CERT_PATH="$2"; shift 2;;
-            -u | --upload)          UPLOAD_PATH="$2"; shift 2;;
+            --owncloud-upload)      OWNCLOUD_UPLOAD=true; shift;;
+            --owncloud-path)        OWNCLOUD_PATH="$2"; shift 2;;
             -s | --ssh)             TARGET_DEVICE_SSH="$2"; shift 2;;
             -- ) shift; break ;;
             * ) err "unsupported argument $1"; usage; exit 1;;
@@ -75,19 +82,12 @@ main() {
 
     info "Start running tests"
     run "$EASYMESH_CERT_PATH"/run_test_file.sh -o "$LOG_FOLDER" -d "$TARGET_DEVICE" "$TESTS" "$VERBOSE_OPT"
+    mv prplmesh.{ipk,buildinfo} "$LOG_FOLDER"
 
-    passed=$(find "$LOG_FOLDER"/PASS/ -maxdepth 1 -type d -printf x | wc -c)
-    failed=$(find "$LOG_FOLDER"/FAIL/ -maxdepth 1 -type d -printf x | wc -c)
-    total=$((failed+passed))
-    if [ "$VERBOSE" = "true" ]; then
-        info "Passed tests:"
-        ls "$LOG_FOLDER/PASS"
-        info "Failed tests:"
-        ls "$LOG_FOLDER/FAIL"
+    if [ -n "$OWNCLOUD_UPLOAD" ]; then
+        info "Uploading $LOG_FOLDER to $OWNCLOUD_PATH"
+        "$scriptdir"/upload_to_owncloud.sh "$OWNCLOUD_PATH" "$LOG_FOLDER"
     fi
-    info "Passed $passed/$total"
-    info "Uploading results to $UPLOAD_PATH not supported yet"
-    # TODO upload results
     info "done"
 }
 
@@ -98,5 +98,6 @@ TOOLS_PATH="$rootdir/tools"
 EASYMESH_CERT_PATH=$(realpath "$rootdir/../easymesh_cert")
 TARGET_DEVICE="netgear-rax40"
 TARGET_DEVICE_SSH="$TARGET_DEVICE-1"
+OWNCLOUD_PATH=Nightly/agent_certification
 
 main "$@"
--- a/ci/run_easymesh_cert.sh
+++ b/ci/run_easymesh_cert.sh
@@ -78,11 +78,12 @@ main() {
     cat "$PRPLMESH_BUILDINFO"
 
     info "deploy latest ipk to $TARGET_DEVICE_SSH"
-    run "$TOOLS_PATH"/deploy_ipk.sh "$TARGET_DEVICE_SSH" "$IPK"
+    run "$TOOLS_PATH"/deploy_ipk.sh "$TARGET_DEVICE_SSH" "$PRPLMESH_IPK"
 
     info "Start running tests"
     run "$EASYMESH_CERT_PATH"/run_test_file.sh -o "$LOG_FOLDER" -d "$TARGET_DEVICE" "$TESTS" "$VERBOSE_OPT"
-    mv prplmesh.{ipk,buildinfo} "$LOG_FOLDER"
+    mv "$PRPLMESH_IPK" "$LOG_FOLDER"
+    mv "$PRPLMESH_BUILDINFO" "$LOG_FOLDER"
 
     if [ -n "$OWNCLOUD_UPLOAD" ]; then
         info "Uploading $LOG_FOLDER to $OWNCLOUD_PATH"
@@ -92,8 +93,8 @@ main() {
 }
 
 BRANCH=master
-IPK=prplmesh.ipk
-PRPLMESH_BUILDINFO=prplmesh.buildinfo
+PRPLMESH_IPK=prplmesh.ipk
+PRPLMESH_BUILDINFO=prplmesh_buildinfo.txt
 TOOLS_PATH="$rootdir/tools"
 EASYMESH_CERT_PATH=$(realpath "$rootdir/../easymesh_cert")
 TARGET_DEVICE="netgear-rax40"
--- a/ci/run_easymesh_cert.sh
+++ b/ci/run_easymesh_cert.sh
@@ -72,13 +72,19 @@ main() {
     info "Tests to run: $TESTS"
 
     info "download latest ipk from branch $BRANCH"
-    run "$TOOLS_PATH"/download_ipk.sh --branch "$BRANCH"
+    "$TOOLS_PATH"/download_ipk.sh --branch "$BRANCH" || {
+        err "Failed to download prplmesh.ipk, abort"
+        exit 1
+    }
     
     info "prplmesh build info:"
     cat "$PRPLMESH_BUILDINFO"
 
     info "deploy latest ipk to $TARGET_DEVICE_SSH"
-    run "$TOOLS_PATH"/deploy_ipk.sh "$TARGET_DEVICE_SSH" "$PRPLMESH_IPK"
+    "$TOOLS_PATH"/deploy_ipk.sh "$TARGET_DEVICE_SSH" "$PRPLMESH_IPK" || {
+        err "Failed to deploy prplmesh.ipk, abort"
+        exit 1
+    }
 
     info "Start running tests"
     "$EASYMESH_CERT_PATH"/run_test_file.sh -o "$LOG_FOLDER" -d "$TARGET_DEVICE" "$TESTS" "$VERBOSE_OPT"
@@ -87,7 +93,10 @@ main() {
 
     if [ -n "$OWNCLOUD_UPLOAD" ]; then
         info "Uploading $LOG_FOLDER to $OWNCLOUD_PATH"
-        "$scriptdir"/upload_to_owncloud.sh "$OWNCLOUD_PATH" "$LOG_FOLDER"
+        "$scriptdir"/upload_to_owncloud.sh "$OWNCLOUD_PATH" "$LOG_FOLDER" || {
+            err "Failed to upload $LOG_FOLDER"
+            exit 1
+        }
     fi
     info "done"
 }
--- a/ci/run_easymesh_cert.sh
+++ b/ci/run_easymesh_cert.sh
@@ -72,7 +72,7 @@ main() {
     info "Tests to run: $TESTS"
 
     info "download latest ipk from branch $BRANCH"
-    "$TOOLS_PATH"/download_ipk.sh --branch "$BRANCH" || {
+    "$TOOLS_PATH"/download_ipk.sh --branch "$BRANCH" ${VERBOSE:+ -v} || {
         err "Failed to download prplmesh.ipk, abort"
         exit 1
     }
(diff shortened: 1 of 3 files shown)


ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# Copyright (c) 2020 Tomer Eliyahu (Intel)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+scriptdir="$(cd "${0%/*}" && pwd)"
+rootdir=$(realpath "$scriptdir/../")
+
+# shellcheck source=functions.sh
```
arnout: Does this work? I'd expect it should be
```suggestion
# shellcheck source=../tools/functions.sh
```

tools/download_ipk.sh
```diff
@@ -22,12 +23,12 @@ download() {
...
     # curl options - fail on error (-f), follow redirect (-L)
-    curl -f -L "$IPK_URL" --output "$IPK_PATH"/prplmesh.ipk
-    curl -f -L "$BUILDINFO_URL" --output "$IPK_PATH"/prplmesh_buildinfo.txt
+    curl -f -L "$IPK_URL" --output "$IPK_PATH"/prplmesh.ipk "$REDIRECT"
+    curl -f -L "$BUILDINFO_URL" --output "$IPK_PATH"/prplmesh_buildinfo.txt "$REDIRECT"
 }
 
 main() {
-    if ! OPTS=$(getopt -o 'h' --long help,branch:,device:,path: -n 'parse-options' -- "$@"); then
+    if ! OPTS=$(getopt -o 'hq' --long help,quite,branch:,device:,path: -n 'parse-options' -- "$@"); then
```
rmelotte: From the parsing and the help, 'q' should be 'v' and 'quite' should be 'verbose'.

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+. "$rootdir/tools/functions.sh"
+
+usage() {
+    echo "usage: $(basename "$0") [-hv] <user> [user] <password> [password] <remote-path> [path] <local-path> [path]"
```
arnout: I read this as: you can call it as `upload_to_owncloud.sh arnout arnoutpassword path/on/server ./file` or as `upload_to_owncloud.sh arnout user arnoutpassword password path/on/server path ./file path` - which makes no sense.

In other words:
```suggestion
    echo "usage: $(basename "$0") [-hv] <user> <password> <remote-path> <local-path>"
```

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+    if ! OPTS=$(getopt -o 'hvu:' --long help,verbose,url: -n 'parse-options' -- "$@"); then
+        echo echo "Failed parsing options." >&2; usage; exit 1
+    fi
+
+    eval set -- "$OPTS"
+
+    while true; do
+        case "$1" in
+            -h | --help)            usage; exit 0;;
+            -v | --verbose)         export VERBOSE=true; REDIRECT=">/dev/null 2>&1"; shift;;
```
arnout: Shouldn't this be
```suggestion
            -v | --verbose)         export VERBOSE=true; REDIRECT=""; shift;;
```

and set REDIRECT as a default?

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+    [ -n "$1" ] || { usage; err "Missing user name"; exit 1; }
+    user="$1"; shift
+    [ -n "$1" ] || { usage; err "Missing password"; exit 1; }
+    password="$1"; shift
+    [ -n "$1" ] || { usage; err "Missing remote-path"; exit 1; }
+    remote_path="$1"; shift
+    [ -n "$1" ] || { usage; err "Missing local-path"; exit 1; }
+    local_path="$1"; shift
+
+    info "upload $local_path to $OWNCLOUD_URL/$remote_path user $user password $password"
```
arnout: I wouldn't echo the password - ever.

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+    echo "      -h|--help - display this help."
+    echo "      -v|--verbose - enable verbosity"
+    echo "      -u|--url - webDAV URL (default https://ftp.essensium.com/owncloud/remote.php/dav/files/USERNAME)"
+    echo
+    echo "  positional arguments:"
+    echo "      user - owncloud user name to use for uploading file[s]"
+    echo "      password - owncloud password"
+    echo "      remote-path - path in the cloud, relative to the user home to upload"
+    echo "      local-path - local file or folder to upload"
+    echo ""
```
arnout:
```suggestion
    echo "usage: $(basename "$0") [-hv] <remote-path> <local-path>"
    echo "  options:"
    echo "      -h|--help - display this help."
    echo "      -v|--verbose - enable verbosity"
    echo "      -u|--url - webDAV URL (default https://ftp.essensium.com/owncloud/remote.php/dav/files/USERNAME)"
    echo
    echo "  positional arguments:"
    echo "      remote-path - path in the cloud, relative to the user home to upload"
    echo "      local-path - local file or folder to upload"
    echo ""
    echo "  Credentials are taken from $HOME/.netrc"
    echo "  Make sure it has a line like this:"
    echo "    machine ftp.essensium.com login <user> password <password>"
    echo ""
```

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+    [ -n "$1" ] || { usage; err "Missing local-path"; exit 1; }
+    local_path="$1"; shift
+
+    info "upload $local_path to $OWNCLOUD_URL/$remote_path user $user password $password"
+    find "$local_path" -type d -exec \
+        realpath {} --relative-to="$(dirname "$local_path")" \; | \
+            while read -r -s dir; do
+                printf . # show progress
+                dbg "Create directory: $dir" 
+                curl -u "$user":"$password" -X MKCOL "$OWNCLOUD_URL/$user/$remote_path/$dir" "$REDIRECT"
```
arnout: Passwords on the command line are a security nightmare. True, it's going to use an account with very limited permissions, but still we don't like exposing passwords like that.

I would propose to leave out username and password entirely and instead use --netrc. The user that runs the script should make sure the netrc file is there. And it should be documented in the help text, as in my suggestion above.

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+                printf . # show progress
+                # get the list of files to upload in the format <file>,<file>,...,<file>
+                files=$(find "$(dirname "$local_path")/$dir/" -type f -maxdepth 1 -print0 | tr '\0' ',' | sed 's/,$//')
+                dbg "$files"
+                [ -n "$files" ] && {
+                     curl -u "$user":"$password" -T "{$files}" "$OWNCLOUD_URL/$user/$remote_path/$dir/" "$REDIRECT"
```
arnout: If you quote $REDIRECT, it will not work. I know shellcheck requests it, but shellcheck is wrong here :-).

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,76 @@
...
+                }
+            done
+            echo # show progress
```
arnout: Indentation is wrong here - this should be at the level of the `find`.

arnout (changes requested):
I only reviewed the first commit but since I have so much feedback and you're actively working on it, I'm already requesting changes.

I'll change it to approve later.

ci/run_easymesh_cert.sh
```diff
@@ -22,7 +22,10 @@ usage() {
     echo "      -b|--branch - prplmesh branch to use (default master latest)"
     echo "      -o|--log-folder - path to put the logs to (make sure it does not contain previous logs)."
     echo "      -e|--easymesh-cert - path to easymesh_cert repository (default ../easymesh_cert)"
-    echo "      -u|--upload - upload results"
+    echo "      --owncloud-upload - weather or not to upload results to owncloud"
```
arnout:
```suggestion
    echo "      --owncloud-upload - wether or not to upload results to owncloud"
```

ci/run_easymesh_cert.sh
```diff
@@ -22,7 +22,10 @@ usage() {
...
+    echo "      --owncloud-user - owncloud username"
+    echo "      --owncloud-password - owncloud password"
```
arnout: Again, I'd rather rely on .netrc.

ci/run_easymesh_cert.sh
```diff
@@ -22,7 +22,10 @@ usage() {
...
+    echo "      --owncloud-path - relative path in the owncloud server to upload results to"
```
arnout: Please mention the default. Note that you can use 
```suggestion
    echo "      --owncloud-path - relative path in the owncloud server to upload results to"
    echo "                                   (default: $OWNCLOUD_PATH)"
```

ci/run_easymesh_cert.sh
```diff
@@ -80,7 +80,7 @@ main() {
     run "$TOOLS_PATH"/deploy_ipk.sh "$TARGET_DEVICE_SSH" "$PRPLMESH_IPK"
 
     info "Start running tests"
-    run "$EASYMESH_CERT_PATH"/run_test_file.sh -o "$LOG_FOLDER" -d "$TARGET_DEVICE" "$TESTS" "$VERBOSE_OPT"
+    "$EASYMESH_CERT_PATH"/run_test_file.sh -o "$LOG_FOLDER" -d "$TARGET_DEVICE" "$TESTS" "$VERBOSE_OPT"
```
rmelotte: I'm not sure quoting "$TESTS" will work if the intention is to pass multiple tests. Did you try it?
I didn't see the rest of the file, but if run_easymesg_cert is just a wrapper to run_test_file.sh then you should be able to use "$@" instead.

ci/run_easymesh_cert.sh
```diff
@@ -75,19 +81,12 @@ main() {
...
+    mv prplmesh.{ipk,buildinfo} "$LOG_FOLDER"
 
-    passed=$(find "$LOG_FOLDER"/PASS/ -maxdepth 1 -type d -printf x | wc -c)
-    failed=$(find "$LOG_FOLDER"/FAIL/ -maxdepth 1 -type d -printf x | wc -c)
-    total=$((failed+passed))
-    if [ "$VERBOSE" = "true" ]; then
-        info "Passed tests:"
-        ls "$LOG_FOLDER/PASS"
-        info "Failed tests:"
-        ls "$LOG_FOLDER/FAIL"
```
arnout, line 86: This bit is *not* done in run_test_file.sh, and I think it's useful to do it.

tools/download_ipk.sh
```diff
@@ -23,7 +23,7 @@ download() {
     BUILDINFO_URL="https://gitlab.com/prpl-foundation/prplmesh/-/jobs/artifacts/$BRANCH/raw/build/$DEVICE/prplmesh.buildinfo?job=build-for-$DEVICE"
     # curl options - fail on error (-f), follow redirect (-L)
     curl -f -L "$IPK_URL" --output "$IPK_PATH"/prplmesh.ipk
-    curl -f -L "$BUILDINFO_URL" --output "$IPK_PATH"/prplmesh.buildinfo
+    curl -f -L "$BUILDINFO_URL" --output "$IPK_PATH"/prplmesh_buildinfo.txt
```
arnout: Good idea!

tools/download_ipk.sh
```diff
@@ -22,12 +23,12 @@ download() {
     IPK_URL="https://gitlab.com/prpl-foundation/prplmesh/-/jobs/artifacts/$BRANCH/raw/build/$DEVICE/prplmesh.ipk?job=build-for-$DEVICE"
     BUILDINFO_URL="https://gitlab.com/prpl-foundation/prplmesh/-/jobs/artifacts/$BRANCH/raw/build/$DEVICE/prplmesh.buildinfo?job=build-for-$DEVICE"
     # curl options - fail on error (-f), follow redirect (-L)
-    curl -f -L "$IPK_URL" --output "$IPK_PATH"/prplmesh.ipk
-    curl -f -L "$BUILDINFO_URL" --output "$IPK_PATH"/prplmesh_buildinfo.txt
+    curl -f -L "$IPK_URL" --output "$IPK_PATH"/prplmesh.ipk "$REDIRECT"
```
arnout: As mentioned elsewhere, the non-verbose redirect will not work due to the quotes.

However, specfically for curl, you should use `-s` instead of redirecting output. Also, you probably *do* want to keep errors, so it's better to use `-s -S`.
```suggestion
    curl ${VERBOSE:+ -s -S} -f -L "$IPK_URL" --output "$IPK_PATH"/prplmesh.ipk
```

Note the optional argument handling suggestion from [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086). Note that this means VERBOSE needs to be defined as empty instead of false by default, so if you want to apply this trick in any script that uses functions.sh, it's a bigger change.

tools/download_ipk.sh
arnout:
```suggestion
    if ! OPTS=$(getopt -o 'hv' --long help,verbose,branch:,device:,path: -n 'parse-options' -- "$@"); then
```

tools/download_ipk.sh
```diff
@@ -41,6 +42,9 @@ main() {
                 usage
                 exit 0
                 ;;
+            -v|--verbose)
+                REDIRECT=""
```
arnout:
```suggestion
                REDIRECT=""
                shift
```

Something tells me you didn't really test this feature :-)

ci/run_easymesh_cert.sh
```diff
@@ -71,7 +71,7 @@ main() {
     info "Tests to run: $TESTS"
 
     info "download latest ipk from branch $BRANCH"
-    run "$TOOLS_PATH"/download_ipk.sh --branch "$BRANCH"
+    run "$TOOLS_PATH"/download_ipk.sh --quite --branch "$BRANCH" "$VERBOSE_OPT"
```
arnout: quite -> quiet - but that option is not actually implemented in download_ipk.sh.

VERBOSE_OPT shouldn't be quoted, otherwise you're passing an extra empty argument when verbose is not set. So with shellcheck's suggestion:
```suggestion
    run "$TOOLS_PATH"/download_ipk.sh --branch "$BRANCH" ${VERBOSE:+ -v}
```

arnout (approved):
I wouldn't call them nitpicks, but nothing I need to re-review.

arnout:
Oh, one more thing: you have commit summary lines like: `ci: run_easymesh_cert.sh add upload support`. For me, that doesn't read very comfortably because it sounds like `run_easymesh_cert.sh` is the subject of the sentence, which means that `add` should be `adds`. So please add an additional colon: `ci: run_easymesh_cert.sh: add upload support`

Also, there's a typo in one of the commit messages: exists -> exits.

ci/run_easymesh_cert.sh
rmelotte: Actually I've been wanting to at least print the list of failing tests at the end of the job for some times now, I think it would be useful to do it from run_test_file.sh.

rmelotte (approved):

ci/upload_to_owncloud.sh
tomereli: It doesn't work either way, but I'll change to your suggestion

ci/run_easymesh_cert.sh
tomereli: Its actually `whether`, fixed.

tools/download_ipk.sh
tomereli: Unfortunately, we need the opposite - if `VERBOSE` **is** defined not to add the `-s -S` options. So it has to be `--quite` instead, no?

ci/run_easymesh_cert.sh
tomereli: Yes it works:
```
Done
run_easymesh_cert.sh: Start running tests
The following test(s) will be run:

MAP-4.2.1 MAP-4.3.2_ETH
Applying radio info configuration
Using the following MasterTestInfo.xml: /home/tester/work/easymesh_cert/config/mastertestinfo/default.xml

Running test MAP-4.2.1 (1/2)

Time: 2020-03-27_15-30-29

Applying configuration for an agent tes
```

This was a suggestion by `shellcheck`. Originally I used `"$@"`, and this was the error:
```
Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.shellcheck(SC2124)
```

arnout: Wow... This is an artefact from how run_test_file.sh is written.

As Raphaél observed, the tests will be passed a single argument to run_test_file.sh.

The read_test_names parses that into one line per argument. So all the explicitly specified tests are on a single line. You can see that in the output above: normally, the "The following test(s) will be run" output should print one test per line.

But then, the _run_tests function uses `wc -w` to get the number of tests, and `for t in $test_names` (unquoted!) to iterate over them. So for both cases, word splitting is done again, and it magically works.

It still is a horrible hack though.

The proper solution would be to switch to bash and user arrays.

ci/upload_to_owncloud.sh
tomereli: Good idea, changed to use .netrc

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,80 @@
...
+    [ -n "$1" ] || { usage; err "Missing local-path"; exit 1; }
+    local_path="$1"; shift
+
+    info "upload $local_path to $OWNCLOUD_BROWSE_URL/$remote_path/$(basename $local_path) (using user $user)"
+    find "$local_path" -type d -exec \
+        realpath {} --relative-to="$(dirname "$local_path")" \; | \
+            while read -r -s dir; do
+                printf . # show progress
+                dbg "Create directory: $dir" 
+                curl ${QUIET:+ -s -S} -f -n -X MKCOL "$OWNCLOUD_URL/$user/$remote_path/$dir"
```
rmelotte: Can you exit the script with an exit code different than 0 if any of these fail? (not necessarily at the first error, but it would be great to know that something went wrong)
It would allow the job to fail in CI when there is an issue with the upload: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/488726605

ci/upload_to_owncloud.sh
```diff
@@ -0,0 +1,80 @@
...
+                        err "Failed to upload files to $remote_path/$dir/"
+                    }
+                }
+            done
+    printf '\n'
+    info "done"
+}
+
+OWNCLOUD_URL="https://ftp.essensium.com/owncloud/remote.php/dav/files"
+OWNCLOUD_BROWSE_URL="https://ftp.essensium.com/owncloud/index.php/apps/files/?dir=/prplmesh/certification"
```
rmelotte, line 83: Is this only informational for the user?
It won't always upload to `prplmesh/certification` right?

ci/upload_to_owncloud.sh
tomereli: Done

ci/upload_to_owncloud.sh
tomereli: Of course not, this is the base URL for user to be able to click on the link, since the URL used by `curl` can't be used for that.

See here how it is used:
```
info "upload $local_path to $OWNCLOUD_BROWSE_URL/$remote_path/$(basename "$local_path") (using user $user)"
```

ci/run_easymesh_cert.sh
tomereli: Lets do this in a separate PR then and enjoy this moment of bliss that something magically works :)

ci/upload_to_owncloud.sh
arnout: Ah, but that URL is not usable for most people. Instead, you should use
```suggestion
OWNCLOUD_BROWSE_URL="https://ftp.essensium.com/owncloud/index.php/s/ketl3eFhIgwcweZ?path=" (assuming `remote_path` starts with `/Nightly`)
```

PR #1019, opened 2020-03-25
Feature/detect agent down with topology flow

### LINKED ISSUE #925: [TASK] Detect topology changes on the network

morantr:
This PR removed the Topology Discovery handler from the controller and move it to the Agent.
The Agent now uses it to store newly discovered 1905 Devices on a list, and notify the Controller about any change on that list.
If the Controller detects that the Agent did not report a neighbor device that appears on the Controller database, it removed this node.
This whole mechanism purpose it to detect Agent that went down like power off for example.

Currently, there is a problem with this PR. It is observed that running `tests/test_gw_repeater.sh -f -v` is failing since the repeater so_slaves get a timeout on waiting for "JOINED_RESPONSE" multiple times until they get it (takes 60 seconds).
After some debugging I did, it seems that if the controller sends the Topology Notification with the relay flag set, it causes the problem.
Also, I checked on Wireshark and I saw all the messages from the repeater that the controller did not receive.

HELP!

Edit:
For some reason test flows fully passed now. 😱 

Created a pipeline that checks the following: 
```
MAP-4.2.1
MAP-4.3.1_ETH
MAP-4.3.2_ETH_FH24G
MAP-4.3.2_ETH_FH5GL
MAP-4.3.2_ETH_FH5GH
```
https://gitlab.com/prpl-foundation/prplMesh/pipelines/129730277

#925

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -511,7 +511,6 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
                     LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
                     return false;
                 }
-                cmdu_header->flags().relay_indicator = true;
                 send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
             }
             return false;
@@ -624,7 +623,6 @@ void backhaul_manager::after_select(bool timeout)
             LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
             return;
         }
-        cmdu_header->flags().relay_indicator = true;
         send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
     }
 }
@@ -1813,7 +1811,6 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
                 LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
                 return false;
             }
-            cmdu_header->flags().relay_indicator = true;
             send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
         }
 
@@ -2771,7 +2768,6 @@ bool backhaul_manager::handle_1905_topology_discovery(const std::string &src_mac
             LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
             return false;
         }
-        cmdu_header->flags().relay_indicator = true;
         send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
     }
     return true;
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -340,6 +340,7 @@ bool backhaul_manager::init()
     }
 
     if (!bus_subscribe(std::vector<ieee1905_1::eMessageType>{
+            ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE,
             ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE,
             ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_RESPONSE_MESSAGE,
             ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_RENEW_MESSAGE,
@@ -593,6 +594,37 @@ void backhaul_manager::after_select(bool timeout)
         (m_eFSMState == EState::OPERATIONAL && elapsed.count() == 60)) {
         discovery_timestamp = now;
         send_1905_topology_discovery_message();
+    }
+
+    // Each platform must send Topology Discovery message every 60 seconds to its first circle
+    // 1905.1 neighbors.
+    // Iterate on each of known 1905.1 neighbors and check if we have received in the last
+    // 60 seconds a Topology Discovery message from it. If not, remove this neighbor from our list
+    // and send a Topology Notification message.
+    bool neighbors_list_changed = false;
+    for (auto it = m_1905_neighbor_devices.begin(); it != m_1905_neighbor_devices.end();) {
+        const auto &last_topology_discovery = it->second;
+        if (last_topology_discovery + std::chrono::seconds(DISCOVERY_NEIGHBOUR_REMOVAL_TIMEOUT) <
+            std::chrono::steady_clock::now()) {
+            const auto &device_al_mac = it->first;
+            LOG(INFO) << "Removed 1905.1 device " << device_al_mac << " from neighbors list";
+            it                     = m_1905_neighbor_devices.erase(it);
+            neighbors_list_changed = true;
+            continue;
+        }
+        it++;
+    }
+
+    if (neighbors_list_changed) {
+        LOG(INFO) << "Sending topology notification on removeing of 1905.1 neighbors";
+        auto cmdu_header =
+            cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE);
+        if (!cmdu_header) {
+            LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
+            return;
+        }
+        cmdu_header->flags().relay_indicator = true;
+        send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
     }
 }
 
@@ -2037,6 +2069,8 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
      * false if the message needs to be forwarded by the calling function
      */
     switch (cmdu_rx.getMessageType()) {
+    case ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE:
+        return handle_1905_topology_discovery(src_mac, cmdu_rx);
     case ieee1905_1::eMessageType::AP_AUTOCONFIGURATION_RESPONSE_MESSAGE: {
         return handle_1905_autoconfiguration_response(cmdu_rx, src_mac);
     }
@@ -2664,6 +2698,46 @@ bool backhaul_manager::handle_1905_combined_infrastructure_metrics(
     }
     LOG(DEBUG) << "sending ACK message to the originator, mid=" << std::hex << int(mid);
     return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
+}
+
+bool backhaul_manager::handle_1905_topology_discovery(const std::string &src_mac,
+                                                      ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto tlvAlMac = cmdu_rx.getClass<ieee1905_1::tlvAlMacAddressType>();
+    if (!tlvAlMac) {
+        LOG(ERROR) << "getClass tlvAlMacAddressType failed";
+        return false;
+    }
+
+    // Filter out the messages we have sent.
+    if (tlvAlMac->mac() == network_utils::mac_from_string(bridge_info.mac)) {
+        return true;
+    }
+
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received TOPOLOGY_DISCOVERY_MESSAGE from AL MAC=" << tlvAlMac->mac()
+              << ", mid=" << std::hex << int(mid);
+
+    auto new_device =
+        m_1905_neighbor_devices.find(tlvAlMac->mac()) == m_1905_neighbor_devices.end();
+
+    // Add/Update the device on our list.
+    m_1905_neighbor_devices[tlvAlMac->mac()] = std::chrono::steady_clock::now();
+
+    // If it is a new device, then our 1905.1 neighbors list has changed and we are required to send
+    // Topology Notification Message.
+    if (new_device) {
+        LOG(INFO) << "Sending Topology Notification on newly discovered 1905.1 device";
+        auto cmdu_header =
+            cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE);
+        if (!cmdu_header) {
+            LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
+            return false;
+        }
+        cmdu_header->flags().relay_indicator = true;
+        send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
+    }
+    return true;
 }
 
 bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMessageRx &cmdu_rx,
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -27,6 +27,7 @@
  */
 #include <tlvf/ieee_1905_1/eMessageType.h>
 #include <tlvf/ieee_1905_1/s802_11SpecificInformation.h>
+#include <tlvf/ieee_1905_1/tlv1905NeighborDevice.h>
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvDeviceInformation.h>
@@ -2365,6 +2366,42 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
                        << " backhaul has failed!";
             return true;
         }
+    }
+
+    // The tlv1905NeighborDevice we construct here is not entirely correct, but close enough.
+    // We should create a tlv1905NeighborDevice for every local interface on which a 1905 neighbor
+    // is visible, and then add a list of neighbors visible on that interface. However, since we
+    // don't keep track of the incoming interface of the received topology discovery messages, we
+    // can't do that. Instead, we model it as a single interface, the bridge interface, behind which
+    // all neighbors are visible, with a bridge between them. This is accurate enough to allow us to
+    // work in practice.
+    auto tlv1905NeighborDevice = cmdu_tx.addClass<ieee1905_1::tlv1905NeighborDevice>();
+    if (!tlv1905NeighborDevice) {
+        LOG(ERROR) << "addClass ieee1905_1::tlv1905NeighborDevice failed, mid=" << std::hex
+                   << (int)mid;
+        return false;
+    }
+    tlv1905NeighborDevice->mac_local_iface() = network_utils::mac_from_string(bridge_info.mac);
+
+    if (!tlv1905NeighborDevice->alloc_mac_al_1905_device(m_1905_neighbor_devices.size())) {
+        LOG(ERROR) << "alloc_mac_al_1905_device() has failed";
+        return true;
+    }
+
+    size_t index = 0;
+    for (const auto &neighbor_device : m_1905_neighbor_devices) {
+        const auto &neighbor_al_mac = neighbor_device.first;
+
+        auto mac_al_1905_device_tuple = tlv1905NeighborDevice->mac_al_1905_device(index);
+        if (!std::get<0>(mac_al_1905_device_tuple)) {
+            LOG(ERROR) << "getting mac_al_1905_device element has failed";
+            return true;
+        }
+        auto &mac_al_1905_device = std::get<1>(mac_al_1905_device_tuple);
+        mac_al_1905_device.mac   = neighbor_al_mac;
+        mac_al_1905_device.bridges_exist =
+            ieee1905_1::tlv1905NeighborDevice::eBridgesExist::AT_LEAST_ONE_BRIDGES_EXIST;
+        index++;
     }
 
     auto tlvSupportedService = cmdu_tx.addClass<wfa_map::tlvSupportedService>();
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -13,6 +13,7 @@
 
 #include <bcl/beerocks_backport.h>
 #include <bcl/beerocks_config_file.h>
+#include <bcl/beerocks_defines.h>
 #include <bcl/beerocks_socket_thread.h>
 #include <bcl/network/network_utils.h>
 #include <btl/btl.h>
@@ -210,9 +211,8 @@ private:
     const int DEAUTH_REASON_PASSPHRASE_MISMACH        = 2;
     const int AUTOCONFIG_DISCOVERY_TIMEOUT_SECONDS    = 1;
     const int MAX_FAILED_AUTOCONFIG_SEARCH_ATTEMPTS   = 20;
-    const int DISCOVERY_NOTIFICATION_TIMEOUT_SEC      = 60; // According to specification
     const int DISCOVERY_NEIGHBOUR_REMOVAL_TIMEOUT =
-        DISCOVERY_NOTIFICATION_TIMEOUT_SEC + 3; // 3 seconds grace period
+        ieee1905_1_consts::DISCOVERY_NOTIFICATION_TIMEOUT_SEC + 3; // 3 seconds grace period
 
     std::chrono::steady_clock::time_point state_time_stamp_timeout;
     int state_attempts;
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -286,6 +286,16 @@ private:
      */
     bool get_neighbor_links(const sMacAddr &neighbor_mac_filter,
                             std::map<std::string, std::vector<sMacAddr>> &neighbor_links_map);
+    /*
+     * @brief List of known 1905 neighbor devices
+     * 
+     * key:     1905.1 device AL-MAC
+     * value:   Last timestam receiving discovery message from AL-MAC
+     * Devices are being added to the list when receiving a 1905.1 Topology Discovery message from
+     * an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and
+     * we update the time stamp in which the message is received.
+     */
+    std::unordered_map<sMacAddr, std::chrono::steady_clock::time_point> m_1905_neighbor_devices;
 
     /*
  * State Machines
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -75,6 +75,8 @@ private:
     bool handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx, const std::string &src_mac);
 
     // 1905 messages handlers
+    bool handle_1905_topology_discovery(const std::string &src_mac,
+                                        ieee1905_1::CmduMessageRx &cmdu_rx);
     bool handle_1905_autoconfiguration_response(ieee1905_1::CmduMessageRx &cmdu_rx,
                                                 const std::string &src_mac);
     bool handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmdu_rx, const std::string &src_mac);
@@ -208,6 +210,9 @@ private:
     const int DEAUTH_REASON_PASSPHRASE_MISMACH        = 2;
     const int AUTOCONFIG_DISCOVERY_TIMEOUT_SECONDS    = 1;
     const int MAX_FAILED_AUTOCONFIG_SEARCH_ATTEMPTS   = 20;
+    const int DISCOVERY_NOTIFICATION_TIMEOUT_SEC      = 60; // According to specification
+    const int DISCOVERY_NEIGHBOUR_REMOVAL_TIMEOUT =
+        DISCOVERY_NOTIFICATION_TIMEOUT_SEC + 3; // 3 seconds grace period
 
     std::chrono::steady_clock::time_point state_time_stamp_timeout;
     int state_attempts;
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -1004,6 +1004,35 @@ std::set<std::string> db::get_node_children(std::string mac, int type, int state
     return children_macs;
 }
 
+std::list<sMacAddr> db::get_1905_1_neighbors(const sMacAddr &al_mac)
+{
+    auto al_mac_str = network_utils::mac_to_string(al_mac);
+    std::list<sMacAddr> neighbors_al_macs;
+    auto all_al_macs = get_nodes(beerocks::TYPE_IRE);
+
+    // According to IEEE 1905.1 a neighbor is defined as a first circle only, so we need to filter
+    // out the childrens from second circle and above.
+    for (const auto &al_mac_iter : all_al_macs) {
+        if (get_node_parent_ire(al_mac_iter) == al_mac_str) {
+            neighbors_al_macs.push_back(network_utils::mac_from_string(al_mac_iter));
+        }
+    }
+
+    // Add the parent bridge as well to the neighbors list
+    auto parent_bridge = get_node_parent_ire(network_utils::mac_to_string(al_mac));
+    if (!parent_bridge.empty()) {
+        neighbors_al_macs.push_back(network_utils::mac_from_string(parent_bridge));
+    }
+
+    // Add siblings Nodes
+    auto siblings = get_node_siblings(al_mac_str, beerocks::TYPE_IRE);
+    for (const auto &sibling : siblings) {
+        neighbors_al_macs.push_back(network_utils::mac_from_string(sibling));
+    }
+
+    return neighbors_al_macs;
+}
+
 //
 // Capabilities
 //
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -37,6 +37,7 @@
 #include <tlvf/ieee_1905_1/eMessageType.h>
 #include <tlvf/ieee_1905_1/eTlvType.h>
 #include <tlvf/ieee_1905_1/s802_11SpecificInformation.h>
+#include <tlvf/ieee_1905_1/tlv1905NeighborDevice.h>
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvAutoconfigFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvDeviceInformation.h>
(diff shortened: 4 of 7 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -287,6 +287,22 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+     * 
+     * key:     1905.1 device AL-MAC
+     * value:   s1905NeighborDevice
+     * Devices are being added to the list when receiving a 1905.1 Topology Discovery message from
+     * an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and
+     * we update the time stamp in which the message is received.
+     */
+    struct s1905NeighborDevice {
+        std::chrono::steady_clock::time_point last_seen;
+        sMacAddr local_iface_mac;
```
arnout: In general, it's possible that multiple routes exist between us and the neighbour device. In fact, 1905.1 is also intended to be able to manage such a situation. So, in general, this needs to be a vector (which in most cases has just a single element).

Also, you need to store both the local iface and the peer iface. You get the peer iface from the MAC address type TLV in the discovery message.

Finally (but probably not for this PR), I would like to introduce a new struct to represent local interfaces (wireless, bridge, radio, VAP, ...) with ifindex as its main member and mac and ifname as getter functions. The ifindex is the only thing that is really unique. Multiple ifaces may have the same MAC address (e.g. the bridge typically takes the MAC address of the first port added to it), and an interface can be renamed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -287,6 +287,22 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+        bool bridge_exist; // Not sure from where to take this information
```
arnout: This can have two meanings (and I think you mean the first one):
- A bridge exists between this particular interface on this device and the peer interface on the neighbour. This is detected using the LLDP protocol. See section 8.1 of the IEEE 1905.1 spec. Note that in this case, the bridge boolean has to be per interface pair, not per neighbour.
- The neighbour is a bridge. This is indicated by the device bridging capability TLV in the topology response.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2666,6 +2669,54 @@ bool backhaul_manager::handle_1905_combined_infrastructure_metrics(
...
+        return true;
+    }
+
+    auto mid = cmdu_rx.getMessageId();
+    LOG(INFO) << "Received TOPOLOGY_DISCOVERY_MESSAGE from AL MAC=" << tlvAlMac->mac()
+              << ", mid=" << std::hex << int(mid);
+
+    auto new_device = m_1905_neighbor_devices.find(tlvAlMac->mac()) == m_1905_neighbor_devices.end()
+                          ? true
+                          : false;
```
arnout: This is redundant, just use
```suggestion
auto new_device = m_1905_neighbor_devices.find(tlvAlMac->mac()) == m_1905_neighbor_devices.end();
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -287,6 +287,22 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+     * @brief List of known 1905 neighbor devices
+     * 
+     * key:     1905.1 device AL-MAC
+     * value:   s1905NeighborDevice
+     * Devices are being added to the list when receiving a 1905.1 Topology Discovery message from
+     * an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and
+     * we update the time stamp in which the message is received.
+     */
+    struct s1905NeighborDevice {
+        std::chrono::steady_clock::time_point last_seen;
```
arnout: (nitpick) since this is only updated when we receive a topology discovery, not when we actually see it from any other message, it would be more appropriate to call it `last_topology_discovery`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -287,6 +287,22 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+        sMacAddr local_iface_mac;
+        bool bridge_exist; // Not sure from where to take this information
+    };
+    std::unordered_map<sMacAddr, s1905NeighborDevice> m_1905_neighbor_devices;
```
arnout: (nitpick) Since backhaul_manager_thread is already way too big, I think it's worth moving this to a new class and file `topology`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -287,6 +287,22 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+    /*
+     * @brief List of known 1905 neighbor devices
+     * 
+     * key:     1905.1 device AL-MAC
+     * value:   s1905NeighborDevice
+     * Devices are being added to the list when receiving a 1905.1 Topology Discovery message from
+     * an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and
+     * we update the time stamp in which the message is received.
+     */
+    struct s1905NeighborDevice {
```
arnout: According to CODING_STYLE.md, our naming convention is lower snake case without prefix (except for `m_` for members), so this should be `ieee1905_neighbor_device`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2666,6 +2669,54 @@ bool backhaul_manager::handle_1905_combined_infrastructure_metrics(
...
+
+    // Add/Update the device on our list.
+    m_1905_neighbor_devices[tlvAlMac->mac()].last_seen = std::chrono::steady_clock::now();
+
+    auto tlvMac = cmdu_rx.getClass<ieee1905_1::tlvMacAddress>();
+    if (!tlvMac) {
+        LOG(ERROR) << "Failed to getClass tlvMacAddress";
+        return false;
+    }
+    m_1905_neighbor_devices[tlvAlMac->mac()].local_iface_mac = tlvMac->mac();
```
arnout: tlvMac->mac() is the peer iface mac, not the local iface mac.

Unfortunately, the local iface information is discarded by transport_socket_thread::handle_cmdu_message_bus(). It is available in `cmdu_rx_msg->metadata()->if_index` but that doesn't get forwarded into handle_cmdu(). So some more refactoring will be needed to be able to properly populate the neighbour list in the topology response.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -210,6 +210,7 @@ class backhaul_manager : public btl::transport_socket_thread {
     const int DEAUTH_REASON_PASSPHRASE_MISMACH        = 2;
     const int AUTOCONFIG_DISCOVERY_TIMEOUT_SECONDS    = 1;
     const int MAX_FAILED_AUTOCONFIG_SEARCH_ATTEMPTS   = 20;
+    const int DISCOVERY_NOTIFICATION_TIMEOUT_SECONDS  = 63; // 60 sec (standard) + 3 sec grace
```
arnout: (nitpick) I'd call it something like `DISCOVERY_REMOVAL_TIMEOUT_SECONDS` because it's more about when you consider the node as down than about sending the notification.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2367,6 +2368,33 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
         }
     }
 
+    for (const auto &neighbor_device : m_1905_neighbor_devices) {
```
arnout: This is not entirely correct: there should be one tlv1905NeighborDevice per local interface, not per neighbour. And for each of those, *all* the neighbours visible from that interface should be added.

So in practice, this means inverting the m_1905_neighbor_devices map into a map-of-vector-of-neighbor-device.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2367,6 +2368,33 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+
+        auto mac_al_1905_device_tuple = tlv1905NeighborDevice->mac_al_1905_device(0);
+        if (!std::get<0>(mac_al_1905_device_tuple)) {
+            LOG(ERROR) << "getting mac_al_1905_device element has failed";
+            return true;
+        }
+        auto &mac_al_1905_device = std::get<1>(mac_al_1905_device_tuple);
+        mac_al_1905_device.mac   = neighbor_al_mac;
+        mac_al_1905_device.bridges_exist =
+            ieee1905_1::tlv1905NeighborDevice::eBridgesExist(neighbor_device.second.bridge_exist);
```
arnout: Does this work correctly? Isn't `true` going to be converted to `0x01` instead of `0x80`?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1004,6 +1004,36 @@ std::set<std::string> db::get_node_children(std::string mac, int type, int state
     return children_macs;
 }
 
+std::list<sMacAddr> db::get_1905_1_neighbors(const sMacAddr &al_mac)
```
arnout, line 1007: :tada: sMacAddr :1st_place_medal:

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1004,6 +1004,36 @@ std::set<std::string> db::get_node_children(std::string mac, int type, int state
...
+    std::list<sMacAddr> neighbors_al_macs;
+    auto all_al_macs = get_nodes(beerocks::TYPE_IRE);
+
+    // According to IEEE 1905.1 a neighbor is defined as a first circle only, so we need to filter
+    // out the childrens from second circle and above.
+    for (const auto &al_mac_iter : all_al_macs) {
+        if (get_node_parent_ire(al_mac_iter) != al_mac_str) {
+            continue;
+        }
+        neighbors_al_macs.push_back(network_utils::mac_from_string(al_mac_iter));
```
arnout: It would feel more natural to me to invert the condition:
```suggestion
        if (get_node_parent_ire(al_mac_iter) == al_mac_str) {
            neighbors_al_macs.push_back(network_utils::mac_from_string(al_mac_iter));
        }
```

arnout (changes requested):
I didn't make it to the end, only up to "add 1905 neighbor device TLV to topology response". But I've made enough comments already to warrant a re-review, so changes request.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: The "peer iface" is the al_mac and it is the key of the unordered_map, and I take it from `tlvAlMacAddressType` on the discovery message and not the `tlvMacAddress`. The "local iface" mac I take from `tlvMacAddress`.
I think it is more convenient to store it in a map than a vector, but if there is more than one route I could change it to `unordered_multimap`  if it is ok with you?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: I mean the field "Existence of IEEE 802.1 bridges" which is on "1905.1 neighbor device TLV" on IEEE 1905.1-2013 Table 6-15.
We need to send this TLV in the Topology Response message.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: done.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: changed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: maybe, but after we will implement the unified Agent. No point doing it now.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: How it can be the peer iface mac? (by peer you mean neighbor al_mac right?)
Because the other tlv is al_mac_tlv.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: I don't know about it...  "Notification timeout" seems right to me.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: No. For some reason, I thought it is using the first bit.
Fixing.
good catch.

controller/src/beerocks/master/db/db.cpp
morantr: ok

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
tomereli: For structs I agree, looks better as lower snake case

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
tomereli: I disagree - I'm not sure when we will implement the unified agent. So there is definitely a point in doing it now - you can and should do it as the last commit which moves all the implementation to a separate class in a separate file.
But since this can be done in step 3 (refactoring), just make sure it is listed as one of the sub tasks there.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: Looks like this was missed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
tomereli: I tend to agree with @arnout here - "notification timeout" sounds like a timeout on this agent notification, not about other agents. How about defining 2:
```
const int DISCOVERY_NOTIFICATION_TIMEOUT_SEC = 60; // according to specification
const int DISCOVERY_NEIGHBOUR_REMOVAL_TIMEOUT = DISCOVERY_NOTIFICATION_TIMEOUT_SEC + 3 ; // 3 seconds grace period
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Ok, let's make this clear. Because clearly it is not clear to me.
I'm filling the container from Discovery message which has:
1905.1 AL MAC address type TLV and MAC address type TLV.
1905.1 AL MAC address type is the neighbor (peer?) mac?
and MAC address type TLV is the local iface_mac (or not?)?

Anyway, you say that I should store the information like this?
```cpp
    /*
     * key:     Local interface mac
     * value:   s1905NeighborDevice
     * Devices are being added to the list when receiving a 1905.1 Topology Discovery message from
     * an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and
     * we update the time stamp in which the message is received.
     */
    struct s1905NeighborDevice {
        std::chrono::steady_clock::time_point last_topology_discovery;
        std::list<sMacAddr> al_mac;
        bool bridge_exist; // Not sure from where to take this information
    };
    std::unordered_map<sMacAddr, s1905NeighborDevice> m_1905_neighbor_devices;

```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
arnout: So that's the first case. It needs to be detected using the LLDP protocol. And as mentioned, this has to be per local interface (I wrote per interface pair but that was wrong).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: peer iface mac is the MAC address of the interface on which the neighbour sends out its topology discovery messages. I thought it was part of the 1905.1 neighbour TLV, but apparently I was mistaken - it's only included in the link metrics.

If there is a single link between two neighbours, you have four MAC addresses involved: the AL-MAC address of this device, the AL-MAC address of the neighbour device, the MAC address of the interface on which the topology discovery is received by this device, and the MAC address of the interface on which the topology discovery message is sent by the neighbour device. The AL-MAC address and the sending interface MAC address of the neighbour device come from the topology discover message - AL MAC address type resp. MAC address type TLV. The local interface on which the topology discovery message was received has to be recorded by the receiving AL Entity, but, as mentioned, currently we don't.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
arnout: And then use DISCOVERY_NOTIFICATION_TIMEOUT_SEC instead of the hardcoded 60 in the backhaul manager to send out the discovery.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: it is on a fixup
https://github.com/prplfoundation/prplMesh/pull/1019/commits/e55c5493dbb21b8330b272192c8d0d6e21ec6363

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: Ok, I will change the comment about it on a fixup.
Not implementing it as part of this PR. Out of scope.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: The last phase if mostly for code removal. I can add this as a new subtask. But I really think that "not sure when we will implement the unified agent" is not a good idea. We should allocate time for it and put efforts into it ASAP. It is just a matter of making a decision to do it in my POV.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: will be decided by the poll on the democracy channel.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Currently, we fill the interface mac on the discovery message with our bridge mac.
Second, I don't understand how could we know from each interface the message will be sent with. I mean, we send something to the bus. We don't get information about the routes it used to send the message. Also, the Discovery message uses multicast. It could, in theory, use all available interfaces, but we could only fill one interface.
It is not so clear to me.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: I don't agree, but ok.... 😑

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1679,6 +1680,53 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
         }
     }
 
+    // The reported neighbors list might not be correct since the reporting al_mac hasn't received 
+    // a Topology Discovery from its neighbors yet. Therefore, remove a neighbor node only if more
+    // than 65 seconds have passed since we added this node. This promise that the reported al_mac
+    // will get the Topology Discovery messages from its neighbors and add them to the report.
+    auto last_state_change_timestamp = database.get_last_state_change(src_mac);
+    if (last_state_change_timestamp + std::chrono::seconds(65) < std::chrono::steady_clock::now()) {
```
tomereli: Don't use hard coded values - I suggest to move `DISCOVERY_NOTIFICATION_TIMEOUT_SECONDS` to a common place and use this here instead. In addition, you need to say in the comment why specifically 65 seconds.
Its 60 seconds timeout + 5 seconds grace period?

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1679,6 +1680,53 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
...
+        LOG(TRACE) << "Checking if one of " << src_mac << " is no longer connected";
```
tomereli: I don't understand this comment - you never use src_mac after this line...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -511,7 +511,6 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
                     LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
                     return false;
                 }
-                cmdu_header->flags().relay_indicator = true;
```
tomereli, line 512: I think this should be fixed before the PR can be merged, no?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -286,6 +286,16 @@ class backhaul_manager : public btl::transport_socket_thread {
...
     bool get_neighbor_links(const sMacAddr &neighbor_mac_filter,
                             std::map<std::string, std::vector<sMacAddr>> &neighbor_links_map);
+    /*
+     * @brief List of known 1905 neighbor devices
+     * 
+     * key:     1905.1 device AL-MAC
+     * value:   Last timestam receiving discovery message from AL-MAC
+     * Devices are being added to the list when receiving a 1905.1 Topology Discovery message from
+     * an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and
+     * we update the time stamp in which the message is received.
```
arnout, line 301:
```suggestion
     * \li key:     1905.1 device AL-MAC
     * \li value:   Last timestamp receiving discovery message from AL-MAC
     *
     * Devices are being added to the list when receiving a 1905.1 Topology Discovery message from
     * an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and
     * we update the time stamp in which the message is received.
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -594,6 +595,37 @@ void backhaul_manager::after_select(bool timeout)
...
+        const auto &last_topology_discovery = it->second;
+        if (last_topology_discovery + std::chrono::seconds(DISCOVERY_NEIGHBOUR_REMOVAL_TIMEOUT) <
+            std::chrono::steady_clock::now()) {
+            const auto &device_al_mac = it->first;
+            LOG(INFO) << "Removed 1905.1 device " << device_al_mac << " from neighbors list";
+            it                     = m_1905_neighbor_devices.erase(it);
+            neighbors_list_changed = true;
+            continue;
+        }
+        it++;
```
arnout, line 615: I find this easier to read:
```suggestion
        } else {
            it++;
        }
```

(nitpick and personal preference only anyway)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -594,6 +595,37 @@ void backhaul_manager::after_select(bool timeout)
...
+    // 1905.1 neighbors.
+    // Iterate on each of known 1905.1 neighbors and check if we have received in the last
+    // 60 seconds a Topology Discovery message from it. If not, remove this neighbor from our list
+    // and send a Topology Notification message.
+    bool neighbors_list_changed = false;
+    for (auto it = m_1905_neighbor_devices.begin(); it != m_1905_neighbor_devices.end();) {
+        const auto &last_topology_discovery = it->second;
+        if (last_topology_discovery + std::chrono::seconds(DISCOVERY_NEIGHBOUR_REMOVAL_TIMEOUT) <
+            std::chrono::steady_clock::now()) {
+            const auto &device_al_mac = it->first;
```
arnout, line 609: I don't like to create a variable for something that is used only once (nitpick and personal preference only).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2666,6 +2700,46 @@ bool backhaul_manager::handle_1905_combined_infrastructure_metrics(
     return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::handle_1905_topology_discovery(const std::string &src_mac,
+                                                      ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto tlvAlMac = cmdu_rx.getClass<ieee1905_1::tlvAlMacAddressType>();
+    if (!tlvAlMac) {
+        LOG(ERROR) << "addClass tlvAlMacAddressType failed";
```
```suggestion
        LOG(ERROR) << "getClass tlvAlMacAddressType failed";
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: I want a big comment here that explains that our 1905 neighbour device TLV is not entirely correct (but close enough).
```
        // The tlv1905NeighborDevice we construct here is not entirely correct, but close enough.
        // We should create a tlv1905NeighborDevice for every local interface on which a 1905 neighbor
        // is visible, and then add a list of neighbors visible on that interface. However, since we don't
        // keep track of the incoming interface of the received topology discovery messages, we can't do that.
        // Instead, we model it as a single interface, the bridge interface, behind which all neighbors are
        // visible, with a bridge between them. This is accurate enough to allow us to work in practice.
```

Also, I think we should create an issue (for M3 or M5) to correct this.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2367,6 +2368,33 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
         }
     }
 
+    for (const auto &neighbor_device : m_1905_neighbor_devices) {
+        auto tlv1905NeighborDevice = cmdu_tx.addClass<ieee1905_1::tlv1905NeighborDevice>();
```
arnout: This is not correct: there should be a tlv1905NeighborDevice *per local interface*. Since we only model 1 local interface (the bridge), the should be only one tlv1905NeighborDevice. So something like:
```
        auto tlv1905NeighborDevice = cmdu_tx.addClass<ieee1905_1::tlv1905NeighborDevice>();
        tlv1905NeighborDevice->mac_local_iface() = network_utils::mac_from_string(bridge_info.mac);
        if (!tlv1905NeighborDevice->alloc_mac_al_1905_device(m_1905_neighbor_devices.size()))
            ...
        size_t index = 0;
        for (const auto &neighbor_device : m_1905_neighbor_devices) {
            auto mac_al_1905_device_tuple = tlv1905NeighborDevice->mac_al_1905_device(index);
            ...
            index++;
        }
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: no :(

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1004,6 +1004,35 @@ std::set<std::string> db::get_node_children(std::string mac, int type, int state
...
+    auto parent_bridge = get_node_parent_ire(network_utils::mac_to_string(al_mac));
+    if (!parent_bridge.empty()) {
+        neighbors_al_macs.push_back(network_utils::mac_from_string(parent_bridge));
+    }
+
+    // Add siblings Nodes
+    auto siblings = get_node_siblings(al_mac_str, beerocks::TYPE_IRE);
+    for (const auto &sibling : siblings) {
+        neighbors_al_macs.push_back(network_utils::mac_from_string(sibling));
+    }
```
arnout, line 1031: Just to be clear: I don't understand how the topology is represented in the db, so I haven't really reviewed this code.

@mariomaz if you're looking for something that is easy to unit test, I think the master db is definitely a good example. It's a leaf class so no mocking is needed even.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1679,6 +1680,53 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
...
+        for (const auto &tlv1905NeighborDevice : tlv1905NeighborDeviceList) {
+            if (!tlv1905NeighborDevice) {
+                LOG(ERROR) << "ieee1905_1::tlv1905NeighborDevice has invalid pointer";
+                return false;
+            }
+
+            const auto neighbor_al_mac_tuple = tlv1905NeighborDevice->mac_al_1905_device(0);
+            if (!std::get<0>(neighbor_al_mac_tuple)) {
+                LOG(ERROR) << "Getting al_mac element has failed";
+            }
```
arnout: Here we need to iterate over all neighbours, not just the first one.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1679,6 +1680,53 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
...
+        for (const auto &neighbor_al_mac_on_db : neighbor_al_macs_on_db) {
+            // If reported al_mac is on the db skip it, otherwise remove the node.
+            LOG(DEBUG) << "Checks if al_mac " << al_mac << " neighbor " << neighbor_al_mac_on_db
+                       << " is reported in this message";
+            if (reported_neighbor_al_macs.find(neighbor_al_mac_on_db) !=
+                reported_neighbor_al_macs.end()) {
+                continue;
+            }
+            LOG(DEBUG) << "known neighbor al_mac  " << neighbor_al_mac_on_db
+                       << " is not reported on 1905 Neighbor Device TLV, removing the al_mac node";
```
arnout: Actually, that's not enough. It is possible that re-routing took place, and the node is now a neighbour of some other node. So we should now check the neighbours of `neighbor_al_mac_on_db` and remove it if it doesn't have any (other than `src_mac`).

controller/src/beerocks/master/son_master_thread.cpp
morantr: yes, it is grace. I did what you asked.

controller/src/beerocks/master/son_master_thread.cpp
morantr: Yes I was missing a word there. It should have been:
"Checking if one of " << src_mac << " **neighbors** is no longer connected"
fixed.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1679,6 +1680,53 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
...
+
+            std::string neighbor_al_mac_on_db_str =
+                network_utils::mac_to_string(neighbor_al_mac_on_db);
+            auto backhhaul_mac = database.get_node_parent(neighbor_al_mac_on_db_str);
+            son_actions::handle_dead_node(backhhaul_mac, database.get_node_parent(backhhaul_mac),
+                                          database, cmdu_tx, tasks);
+        }
+    }
```
arnout, line 1746: I think we should also send a topology query to all the reported neighbours that are not currently in the database, and add them (actually I'd add them on reception of the topology response, not right away).

Also, I'm missing actually adding the neighbour information in the database. I think now this is only done with the intel slave join, and in the non-intel slave join we assume it's connected directly to the gateway. Now would be a good time to fix that (or add an issue to fix it for M3).

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1679,6 +1680,53 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
         }
     }
 
+    // The reported neighbors list might not be correct since the reporting al_mac hasn't received 
+    // a Topology Discovery from its neighbors yet. Therefore, remove a neighbor node only if more
+    // than 65 seconds have passed since we added this node. This promise that the reported al_mac
+    // will get the Topology Discovery messages from its neighbors and add them to the report.
+    auto last_state_change_timestamp = database.get_last_state_change(src_mac);
```
arnout, line 1690: I think this should be `al_mac` instead of `src_mac`. Normally they should be the same, but if they are not, we should actually somewhere make src_mac an alias of al_mac. That of course doesn't need to be done now, it's a much bigger change, but just use al_mac to make things consistent.

arnout (approved):
Again some important comments but can be fixed without re-review.

The last commit really shouldn't be there... but if we can't fix the issue quickly, I think we should indeed unblock ourselves and apply it, and create an issue to revert it later.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
tomereli: When you have time from APM tasks, go ahead :)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: In this case I think it is good to create the variable since it tells the reader that `it->first` is the device al mac, and will be optimized anyway by the compiler.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
morantr: I don't think we should get to this resolution in the Doxygen comments. 
"/li" meaning is not clear from first sight and I had to check what it is. I prefer not to use it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Yes, this is what I've tried to do.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: added

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Yes, you are right.
fixed it.

controller/src/beerocks/master/son_master_thread.cpp
morantr: done

controller/src/beerocks/master/son_master_thread.cpp
morantr: Good catch. Added one more check to filter such cases.

controller/src/beerocks/master/son_master_thread.cpp
morantr: Currently, we send a query to Agents when receiving M1 from them.
It might be a good idea to send a query on new neighbor that has been discovered, and send the M1 after that instead. Need to think about it...
Anyway, what you are talking about is not in the task scope.

controller/src/beerocks/master/son_master_thread.cpp
morantr: No, the src_mac is always the al_mac of the sender.

tomereli (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
arnout: Without the `\li`, the rendered text is:

key:     1905.1 device AL-MAC value:   Last timestam receiving discovery message from AL-MAC Devices are being added to the list when receiving a 1905.1 Topology Discovery message from an unknown 1905.1 device. Every 1905.1 device shall send this message every 60 seconds, and we update the time stamp in which the message is received. 

which is not at all readable.

PR #1020, opened 2020-03-25
Feature - Packet capture using TShark in test_flows.py

### LINKED ISSUE #942: [TASK] Capture packets in test_flows.py using tshark
## Description

Currently, both `test_flows` implementations search the logs in order to confirm that messages were received properly.
However, this causes the logs to be flooded with unnecessary prints and doesn't provide good debugging information for the script, as sometimes it can find irrelevant prints, and the script's string expressions being searched for in the logs need to be identical to the cpp implementation, causing even more confusion.

A proposed improvement to this would be to analyze the raw packet data using the python script.
An attempt was made in PR #924, utilizing tshark to generate a json file from the tcpdump and later opening it in python, however, as suggested by @arnout and @tomereli, a more elegant solution would be to capture the packets directly using tshark, and handling them later as we desire, e.g. filter them or dump them into a file.

## Challenge

At the moment I'm attempting to implement this using [pyshark](https://github.com/KimiNewt/pyshark) however, after attempting to capture the packets in the background on a different thread, I ran into a bug with pyshark's implementation causing an issue when attempting to [run the sniffer in a different thread.](https://github.com/KimiNewt/pyshark/issues/353)
 
So we need to think of a different approach.

Galvien:
I'll go into more detail on the description later, but for now I've created a draft PR with sorted commits.
Consistency needs some work, and I assume that for slower computers, the `sleep()` timers might need to be adjusted.
Will close #942 

### To Do:

- [ ] Check if it's possible (and if it matters, performance-wise) to call tshark without starting it in a different thread, this is a workaround from older methodologies I've tried and might be unnecessary.
 Only requires editing 84286ef50838a619d907af832d080ccd549cff38, 8c982bfbcdf7b4bd61862d974d5ed3e323580c56 and calls to `self.tshark_instances`.

- [ ] Consistency - the capture fails sometimes, find out why, and if it depends on whether or not 
`--skip-init` argument is given.

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -17,6 +17,7 @@ import json
 
 import send_CAPI_command
 from send_CAPI_command import tlv
+from datetime import datetime
 
 '''Regular expression to match a MAC address in a bytes string.'''
 RE_MAC = rb"(?P<mac>([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})"
@@ -96,6 +97,15 @@ class test_flows:
             bridge_id = prplmesh_net['Id']
             bridge = 'br-' + bridge_id[:12]
         return bridge
+
+    def check_time_in_bounds(self, timestamp, end=time.time(), start=None):
+        if start:
+            return start < timestamp and timestamp < end
+        return timestamp < end
+
+    def get_time_for_packet(self, packet):
+        return float(packet['_source']['layers']['frame']['frame.time_epoch'])
+        # return datetime.fromtimestamp(float(packet_time))
 
     def message(self, message: str, color: int = 0):
         '''Print a message, optionally in a color, preceded by the currently running test.'''
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -525,9 +525,12 @@ class test_flows:
         # self.check_log(self.repeater1, "agent_wlan2", "ACK_MESSAGE")
 
     def test_ap_capability_query(self):
+        start_time = time.time()
         self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8001)
-        time.sleep(1)
-
+        time.sleep(2)
+        captured = self.check_tshark()
+        found = self.check_tlv_name('ap capability', captured)
+        self.status(f"AP Capability captured? {found}")
         self.debug("Confirming ap capability query has been received on agent")
         self.check_log(self.repeater1, "agent", "AP_CAPABILITY_QUERY_MESSAGE")
 
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -14,7 +14,7 @@ import sys
 import time
 from typing import Dict
 import json
-
+from threading import Thread
 import send_CAPI_command
 from send_CAPI_command import tlv
 from datetime import datetime
@@ -97,6 +97,22 @@ class test_flows:
             bridge_id = prplmesh_net['Id']
             bridge = 'br-' + bridge_id[:12]
         return bridge
+
+    def start_tshark(self, bridge, duration=None, reset=False):
+        self.tshark_instances = {}
+        self.tshark_capture = {}
+        if reset:
+            for t in self.tshark_instances:
+                try:
+                    t.captured_packets = []
+                finally:
+                    pass
+        for test_index, test_name in enumerate(self.opts.tests):
+            self.tshark_instances[test_name] = Thread(
+                target=self.__capture, daemon=True, args=(test_index, test_name, bridge, duration,))
+
+        for test_name, test_thread in self.tshark_instances.items():
+            test_thread.start()
 
     def __capture(self, test_index, test_name, bridge, duration=None):
         
@@ -251,6 +267,9 @@ class test_flows:
         self.gateway = 'gateway-' + self.opts.unique_id
         self.repeater1 = 'repeater1-' + self.opts.unique_id
         self.repeater2 = 'repeater2-' + self.opts.unique_id
+        self.start_tshark(self.get_bridge_mac(), reset=True)
+        self.status('Starting tshark processes, wait 10s...')
+        time.sleep(10)
         if not self.opts.skip_init:
             self.tcpdump_start()
             try:
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -17,6 +17,7 @@ import json
 from threading import Thread
 import send_CAPI_command
 from send_CAPI_command import tlv
+import signal
 from datetime import datetime
 
 '''Regular expression to match a MAC address in a bytes string.'''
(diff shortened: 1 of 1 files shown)


tests/test_flows.py
```diff
@@ -98,6 +98,24 @@ def get_bridge_mac(self):
...
+    def __capture(self, test_index, test_name, bridge, duration=None):
+        
+        # `test_index` currently unused, but might be useful for file output
+        
+        # for optional file output
+        # self.outputfile = os.path.join(
+        #     self.rootdir, 'logs', f'tshark_log_{test_name}.json')
+        
+        try:
+            tshark_args = ['tshark', '-Q', '-i', bridge, '-T', 'json']
```
arnout, line 128: If this is going to be used for actual hardware, not just for docker, there may be quite a lot of traffic that we're not interested in. Therefore, it would be good to add a pcap filter:
```suggestion
            tshark_args = ['tshark', '-Q', '-i', bridge, '-f', 'ether proto 0x88CC or ether proto 0x893A', '-T', 'json']
```

tests/test_flows.py
```diff
@@ -98,6 +98,24 @@ def get_bridge_mac(self):
...
+            if duration:
+                self.tshark_instances[test_name].proc = subprocess.Popen(
+                    tshark_args+['-a', f'duration:{duration}'], shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
arnout, line 131: `shell=False` is the default, no need to add it.

Putting both stdout and stderr in separate pipes makes it more difficult to make sure we don't block. And I anyway don't think we need to capture stderr - it's better to send it directly to the terminal (i.e., leave it at its default).

tests/test_flows.py
```diff
@@ -98,6 +98,24 @@ def get_bridge_mac(self):
...
+            else:
+                self.tshark_instances[test_name].proc = subprocess.Popen(
+                    tshark_args+['-a', 'duration:900'], shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
arnout, line 134: You can just default duration to 900 in the function and remove the condition.

tests/test_flows.py
```diff
@@ -132,6 +133,47 @@ def __capture(self, test_index, test_name, bridge, duration=None):
...
+        `end=time.time()` (datetime) upper bound for captured packets timestamp
+
+        `start=None` (datetime) lower bound for captured packets timestamp
+
+        `file_output=False` flag to determine if the captured data would be 
+        written to a file
+
+        ### Returns
+
+        a collection of captured `ieee1905` packets
```
arnout, line 163: *Very* good documentation, we can use this as an example (for CODING_STYLE.md) of how to write docstrings.

Just one thing though: the first line should be a summary:

```
    '''Get tshark captured data between `start` and `end`.

    ### Description
    ...
```

tests/test_flows.py
```diff
@@ -132,6 +133,47 @@ def __capture(self, test_index, test_name, bridge, duration=None):
...
+        '''
+        # if file_output:
+        #     with open(self.outputfile, "w+") as f:
+        #         f.write(f'\n{time.time()} - test {self.running}\n')
+        if not target_thread_name:
```
arnout, line 168: Instead of simple `not`, it's better to explicitly compare with None:
`if target_thread_name is not None`.

Also, instead of conditions all over, simply set the default value:
```
    if target_thread_name is None:
        target_thread_name = self.running
```

Note that you can't set the default value in the function declaration here, because it refers to one of the arguments.

tests/test_flows.py
```diff
@@ -525,9 +525,12 @@ def test_channel_selection(self):
...
 
     def test_ap_capability_query(self):
+        start_time = time.time()
         self.gateway_ucc.dev_send_1905(self.mac_repeater1, 0x8001)
-        time.sleep(1)
-
+        time.sleep(2)
+        captured = self.check_tshark()
+        found = self.check_tlv_name('ap capability', captured)
+        self.status(f"AP Capability captured? {found}")
```
arnout, line 533: The `status` bit should move into check_tlv_name.

tests/test_flows.py
```diff
@@ -134,6 +134,13 @@ def __capture(self, test_index, test_name, bridge, duration=None):
                     tshark_args+['-a', 'duration:900'], shell=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
         except subprocess.CalledProcessError as e:
             self.err(f'test {self.run_tests} failed, output: {e.output}')
+
+    def check_tlv_name(self, st: str, packets):
```
arnout, line 138: I think we want to generalize this to a function that checks for the presence of a specific CMDU type (optionally with a specific MID), and returns all packets matching it, or False if there is no match.

Then there is a second function that checks for specific TLVs in those packets, and returns those TLVs.

And finally we can add a function that analyzes other info in those TLVs.

arnout (commented):
I really like how this PR is built up: adding infrastructure bit by bit, and then finally changing just a single test as an example of how it will be used.

What I don't like at all, however, is the way the tshark calls are made, terminating it every time we need to get its input.

As an alternative, I would propose to use pexpect. We will have to use pexpect anyway at some point, e.g. when we start to use boardfarm. And pexpect has a read_nonblocking(). We can use that to read in the JSON objects as they come. The annoying thing is that they are output as a list, so you can't parse it with a normal JSON parser until it ends. The 'ek' output format is a bit better since it dumps objects, but then the TLV contents is dumped in an inconvenient way. Alternatively, we could use the fields output of tshark. Then we get one packet per line, which makes it really easy to match things directly with pexpect. However, it has the same inconvenient way of dumping TLVs as the ek output, and you need to explicitly specify the fields that should be dumped. We could use tshark -G to get all 492 fields for ieee1905, but that's not exactly convenient...

Now that we have this draft PR, I think it would be good to set up a meeting with @abelog and myself to discuss what the best approach is.

tests/test_flows.py
rmelotte: Can I suggest we stick to a single style of docstring?
What we've been using in the recent python scripts we added ([capi.py](https://github.com/prplfoundation/prplMesh/blob/master/tests/capi.py) for example) is based on Numpy's docstring: https://numpydoc.readthedocs.io/en/latest/format.html

I don't want to impose this one over another, but I think we should definitely choose one and try to stick to it.

rmelotte: Here is more or less what it would look like using Numpy's style:
```suggestion
        """Get tshark captured data between `start` and `end`.
        
        Looks into the tshark captured data, updating the global capture data
        and returning the captured data from `start` to `end`.

        Parameters
        ----------
        target_thread_name: str, optional
            Dictionary key (in self.tshark_instances) of the thread running tshark (default is None, which means the current running test will be used).
        end: datetime
            Upper bound for captured packets timestamp.
        start: datetime, optional
            Lower bound for captured packets timestamp (default is None).
        file_output: bool, optional
            Flag to determine if the captured data would be written to a file (default is False).

        Returns
        -------
        List[replace_with_the_classname_of_temp_capture]
            A collection of captured `ieee1905` packets.
        """
```

tests/test_flows.py
```diff
@@ -83,6 +86,110 @@ def __init__(self):
...
+            # ID.
+            bridge_id = prplmesh_net['Id']
+            bridge = 'br-' + bridge_id[:12]
+        return bridge
+
+    def start_tshark(self, bridge, duration=None, reset=False):
+        self.tshark_instances = {}
+        self.tshark_capture = {}
+        if reset:
+            for t in self.tshark_instances:
```
LiorAmram, line 107: I might be missing something here, but `self.tshark_instances` should be an empty dict, so no point in iterating over its items

tests/test_flows.py
```diff
@@ -83,6 +86,110 @@ def __init__(self):
...
+                try:
+                    t.captured_packets = []
+                finally:
```
LiorAmram, line 110: What's the point in this clause?
if there's an exception in the try clause it will still crush the program. 
If you want to ignore exceptions you should use `except` and not `finally`.

tests/test_flows.py
LiorAmram: Actually we can avoid using if altogether:
`tshark_proc = self.tshark_instances[target_thread_name or self.running].proc`
it will take `target_thread_name` if it's not None, and `self.running` otherwise.

tests/test_flows.py
LiorAmram: small addition: in this case, if there is no match it's better to return None and not False.

tests/test_flows.py
Galvien: Regarding stderr, while calling `communicate()` we have access to both stderr and stdout, so I thought that might add a level of convenience.
But If you think this is unnecessary I'll just remove it 😅

tests/test_flows.py
Galvien: Python is cool

tests/test_flows.py
Galvien: This is a relic from an older revision, where the initialization of those dictionaries was in `init()`, will be fixed.

arnout:
Replaced by #1145

PR #1029, opened 2020-03-26
Bugfix/add relay indicator to capi commads

### LINKED ISSUE #1011: [BUG] AP-Autoconfiguration Renew message should set relay indicator to 1
**Describe the bug** -
The AP-Autoconfiguration Renew message CMDU header relay indicator is 0.
The 1905.1 SPEC said:

6.3.10 AP-autoconfiguration renew message format
The AP-autoconfiguration renew message is a relayed multicast message. The following TLVs shall be included in this message:
— One 1905.1 AL MAC address type TLV (see Table 6-8)

## Steps to reproduce

1. use send_CAPI_command.py send AP-Autoconfiguration Renew message.

## Expected behavior

AP-Autoconfiguration Renew message should set relay indicator to 1

## Screenshots
2EDE4CC3-D08A-4ACA-8A94-28A27106D1D6

## Target

- Host: [e.g. Ubuntu 18.04]
- Platform: [e.g. Linux DUMMY mode with ZMQ]
- How you built prplMesh - full build command (for example `sudo ./build-image.sh && sudo ./build.sh -f MSGLIB=zmq CMAKE_BUILD_TYPE=Debug BUILD_TESTS=ON`
- How you ran prplMesh - for example `sudo build/install/scripts/prplmesh_utils.sh start`

morantr:
Add relay indicator to CAPI commands with message types that require it according to Table 6-4 on IEEE 1905.1-2013, to solve #1011.

As part of this PR, changed also other messages that should have been sent as relayed multicast, to relayed multicast.  

**Help Needed!**
For some reason sending messages with relay indicator casing to some other message to not be received. For example, when the repeater sends an auto-config search message (relay=1), the controller response to it (relay=0) but the response is not being received on the repeater side.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -3751,8 +3751,12 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
         beerocks_header->actionhdr()->radio_mac() = hostap_params.iface_mac;
         beerocks_header->actionhdr()->direction() = beerocks::BEEROCKS_DIRECTION_CONTROLLER;
     }
-    return message_com::send_cmdu(master_socket, cmdu_tx, backhaul_params.controller_bridge_mac,
-                                  backhaul_params.bridge_mac);
+
+    auto dst_addr =
+        cmdu_tx.getMessageType() == ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE
+            ? network_utils::MULTICAST_1905_MAC_ADDR
+            : backhaul_params.controller_bridge_mac;
+    return message_com::send_cmdu(master_socket, cmdu_tx, dst_addr, backhaul_params.bridge_mac);
 }
 
 /**
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1667,7 +1667,7 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
         } else { // Forward the data (cmdu) to bus
             // LOG(DEBUG) << "forwarding slave->master message, controller_bridge_mac=" << controller_bridge_mac;
             cmdu_rx.swap(); //swap back before forwarding
-            send_cmdu_to_bus(cmdu_rx, controller_bridge_mac, bridge_info.mac, length);
+            send_cmdu_to_bus(cmdu_rx, dst_mac, bridge_info.mac, length);
         }
     }
 
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -484,7 +484,8 @@ bool backhaul_manager::socket_disconnected(Socket *sd)
                     LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
                     return false;
                 }
-                send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
+
+                send_cmdu_to_bus(cmdu_tx, network_utils::MULTICAST_1905_MAC_ADDR, bridge_info.mac);
             }
             return false;
         } else {
@@ -596,7 +597,7 @@ void backhaul_manager::after_select(bool timeout)
             LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
             return;
         }
-        send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
+        send_cmdu_to_bus(cmdu_tx, network_utils::MULTICAST_1905_MAC_ADDR, bridge_info.mac);
     }
 }
 
@@ -1136,7 +1137,7 @@ bool backhaul_manager::send_1905_topology_discovery_message()
     tlvMac->mac() = network_utils::mac_from_string(bridge_info.mac);
 
     LOG(DEBUG) << "send_1905_topology_discovery_message, bridge_mac=" << bridge_info.mac;
-    return send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
+    return send_cmdu_to_bus(cmdu_tx, network_utils::MULTICAST_1905_MAC_ADDR, bridge_info.mac);
 }
 
 bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<SSlaveSockets> soc)
@@ -1226,7 +1227,7 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<SSlaveSock
     auto beerocks_header                      = message_com::get_beerocks_header(cmdu_tx);
     beerocks_header->actionhdr()->direction() = beerocks::BEEROCKS_DIRECTION_CONTROLLER;
     LOG(DEBUG) << "sending autoconfig search message, bridge_mac=" << bridge_info.mac;
-    return send_cmdu_to_bus(cmdu_tx, MULTICAST_MAC_ADDR, bridge_info.mac);
+    return send_cmdu_to_bus(cmdu_tx, network_utils::MULTICAST_1905_MAC_ADDR, bridge_info.mac);
 }
 
 bool backhaul_manager::backhaul_fsm_wireless(bool &skip_select)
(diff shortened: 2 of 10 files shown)


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3748,8 +3748,10 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
         beerocks_header->actionhdr()->radio_mac() = hostap_params.iface_mac;
         beerocks_header->actionhdr()->direction() = beerocks::BEEROCKS_DIRECTION_CONTROLLER;
     }
-    return message_com::send_cmdu(master_socket, cmdu_tx, backhaul_params.controller_bridge_mac,
-                                  backhaul_params.bridge_mac);
+
+    static const std::string MULTICAST_MAC_ADDR = "01:80:c2:00:00:13";
```
arnout: Can't we use transport_socket_thread::MULTICAST_MAC_ADDR here?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3748,8 +3748,10 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
...
+    auto dst_addr = multicast ? MULTICAST_MAC_ADDR : backhaul_params.controller_bridge_mac;
```
arnout: Since multicast is never used for VS messages, there is nothing in common between the multicast and non-multicast case. So I think it's better to create a separate function send_cmdu_multicast. (nitpick)

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1926,10 +1926,13 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
...
 
         // build 1905.1 message CMDU to send to the controller
-        if (!cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE)) {
+        auto cmdu_header =
+            cmdu_tx.create(0, ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE);
+        if (!cmdu_header) {
             LOG(ERROR) << "cmdu creation of type TOPOLOGY_NOTIFICATION_MESSAGE, has failed";
             return false;
         }
+        cmdu_header->flags().relay_indicator = true;
```
arnout: (Not for this PR) Maybe it would be better to move this code entirely to backhaul_manager. Indeed, if we just forward the notification as-is to backhaul manager, we can handle everything there. backhaul_manager anyway needs to handle the notification since it maintains a cache of the state of all radios together.

arnout (approved):
Looks very good - except that CI fails. From the log files, I don't immediately see what the issue is. Is this the problem we talked about on zoom a couple of days ago? If you want, I'm willing to debug it.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -350,6 +350,23 @@ bool beerocks_ucc_listener::get_send_1905_1_tlv_hex_list(
     return false;
 }
 
+static void set_relay_indicator(std::shared_ptr<ieee1905_1::cCmduHeader> &cmdu_header)
```
tomereli: This should be done directly in `CmduMessageTx::create`, not here in the listener, so we are sure we don't ever forget to set the relay indicator correctly, not only in the UCC listener. This way you will also take care of the destination (multicast) mac address transparently instead of having to manually set it whenever you create such a CMDU.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3728,7 +3728,7 @@ bool slave_thread::ap_manager_heartbeat_check()
     return true;
 }
 
-bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
+bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx, bool multicast)
```
tomereli: This should be done automatically - why do you need a parameter? You can check the message type and use multicast if the message type is one of the messages defined in Table 6-4.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: I guess not since it is defined in the BTL, and the son slave does not link with it.
@morantr - please add this to `network_utils` instead - you should name the variable `MULTICAST_1905_MAC_ADDR` since its the 1905 multicast mac according to the 1905.1 specification.

tomereli (changes requested):

arnout:
I couldn't reproduce the CI failure locally so I rebased on master and pushed again...

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
morantr: Good idea.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: done, although it wouldn't be necessary if we had a unified Agent.
I'm saying it a lot, do I?

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: not critical IMO if it is a default argument or a different function.
Anyway, when we will move to a unified Agent, we send it directly to the bus. So this function will look differently.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: Or... move the code from the backhaul manager to the unified Agent...

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: Good idea.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -687,7 +687,8 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
             // CMDU was not loaded from preprepared buffer, and need to be created manually, and
             // use TLV data from the command (if exists)
             m_cmdu_tx.reset();
-            if (!cmdu_tx.create(g_mid, ieee1905_1::eMessageType(message_type))) {
+            auto cmdu_header = cmdu_tx.create(g_mid, ieee1905_1::eMessageType(message_type));
```
tomereli: This change is no longer needed.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3748,8 +3748,12 @@ bool slave_thread::send_cmdu_to_controller(ieee1905_1::CmduMessageTx &cmdu_tx)
         beerocks_header->actionhdr()->radio_mac() = hostap_params.iface_mac;
         beerocks_header->actionhdr()->direction() = beerocks::BEEROCKS_DIRECTION_CONTROLLER;
     }
-    return message_com::send_cmdu(master_socket, cmdu_tx, backhaul_params.controller_bridge_mac,
-                                  backhaul_params.bridge_mac);
+
+    auto dst_addr =
+        cmdu_tx.getMessageType() == ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE
```
tomereli, line 3756: Shouldn't we send with multicast to more message types?

tomereli (approved):
some small comments but looks great!

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
morantr: removed

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: not from son_slave.

arnout:
I suspect we are seeing the combination of a number of issues in the failing test...

- Since we are using pcap to capture packets (#30, #1090), we receive outgoing packets as well as incoming ones. For normal packets this doesn't happen because they are sent on the bridge and we don't sniff the bridge. For relayed multicast packets, however, we send it to each individual interface. Thus, whenever a relayed multicast packet is sent, it is also received 3 times (wlan0, wlan2, eth0).
- Since we don't have an ebtable rule to disable bridge handling of ieee1905.1 packets, the br-lan bridge will relay the packets even if the relayed multicast bit is not set. Thus, every relayed multicast packet will be received twice on each interface.
- Because of the combination of the above two, it is possible that the first time the transport receives a packet, it came from one of the dummy interfaces (wlan0, wlan2) instead of the interface it really came from (eth0). Therefore, it will be relayed back again to eth0 and return to the sender.
- We send to autoconfig search messages back-to-back, with a different MID.
- The transport de-duplication only seems to keep track of the last received MID. Thus, when the second autoconfig search has been received, and then a duplicate of the first one is received, it will be considered as a new one so it will be relayed again.

The combination of all of the above leads to a rush of duplicate packets coming in into the transport, which overflows the pcap buffer and thus the autoconfig response message is lost. The autoconfig response is unicast and is not retransmitted, so it really only arrives once in the repeater, so it is easy to loose it between all the rest.

I think fixing any one of the above issues will fix the problem. E.g. not setting the relay bit will fix it because only the bridge will do relaying, not the transport.

I think the easiest fix will be to set the ebtables rule in the containers so the bridge itself doesn't do any relaying: `ebtables -A FORWARD  -d 01:80:c2:00:00:13 -j DROP`.

morantr:
Thanks @arnout I did it on our server and it worked. Added one more commit that fixes it.
I had to update the docker image on our machine to make work. So I guess it still fail on gitlab is since we need to update on the gitlab server too. @rmelotte can you help with that?

PR #1031, opened 2020-03-26
Refactor test_flows.py (patch 6)

### LINKED ISSUE #896: [TASK] Refactor python test framework
## Description

Subtask of #738 

The current python test framework (made in issue #894 , PR #859 ) is a simple transition from shell to python, which is a good first step.

The next step to improving the test framework is refactoring it into something sensible, based on python unittest, with each test in a separate file.

Another thing that needs to be done in this task is to implement the new architecture specified by arnout (TO DO: link to UML diagram)

This is needed for #893

arnout:
This is hopefully the last of the test_flows refactoring.
Everything is moved down into separate classes, with a model of the test devices.
At this point, it is not yet committed to unittest or boardfarm (or something else) - it can still go either way.

test_flows.py is no longer specific for docker - except for the main function.
Since it's not clear yet in which direction we'll go further, the way this will be refactored will depend on that choice.
But the tests themselves are completely agnostic of whether they run in docker or on a device.
All the tests are still kept in a single test_flows.py file, since the way they should be split up is probably different for unittest and boardfarm.

Fixes: #896


arnout:
Note that this is still a draft PR. The description is the dream towards which I'm working, the commits don't fully reflect that yet.

Since this PR will again conflict heavily with all other PRs affecting test flows, I will not consider it ready until #965 and #1020 are merged and this is rebased on top of that.

Galvien:
I would also suggest running a PEP8 formatting on top of everything.
In addition to making the file neater and more readable, any time I clean up using auto-format in vs-code a few hundred lines change and I assume I'm not the only one this happens to.
vs-code's default formatter is either Black or PEP8 so I assume just running one of these should be fine.
Some formatting tools also make sure all of the strings are kept in the same format, i.e. either double quotation or single quotation marks.

tests/environment.py
```diff
@@ -0,0 +1,175 @@
...
+import os
+import re
+import subprocess
+
+from capi import UCCSocket
+from opts import debug
+import sniffer
+
+
+class AlEntity:
```
ghost: nitpick: maybe make the class name `ALEntity`, to avoid confusion (I originally read it as "A1 entity")

tests/environment.py
```diff
@@ -0,0 +1,175 @@
...
+    '''Abstract representation of a MultiAP device (1905.1 AL Entity).
+
+    Derived classes provide concrete implementations for a specific device (e.g. docker).
+
+    This provides basic information about the entity, e.g. its AL MAC address. How this information
+    is retrieved is implementation-specific.
+
+    It also provides an abstract interface to interact with the entity, e.g. for sending CAPI
```
ghost, line 28: e.g.? What other means of communication would be supported through this abstract interface?

tests/environment.py
```diff
@@ -0,0 +1,175 @@
...
+        self.ucc_socket = ucc_socket
+        self.is_controller = is_controller
+        self.radios = []
+
+        # Convenience functions that propagate to ucc_socket
+        self.cmd_reply = self.ucc_socket.cmd_reply
+        self.dev_get_parameter = self.ucc_socket.dev_get_parameter
+        self.dev_send_1905 = self.ucc_socket.dev_send_1905
+
+    def command(self, *command: str) -> bytes:
```
ghost, line 49: Please specify the type of command that should be sent with this method.
Also, examples in the docstring would be helpful
e.g.
```
    def sendCAPIcommand(self, *command: str) -> bytes:
    '''Send the specified CAPI command to the device.
    The command is a variable length argument list comprising the different words of the command, e.g.
    device.sendCAPIcommand("reset_device")
    device.sendCAPIcommand("configure_device", "00:11:22:33:44:55")
```

tests/environment.py
```diff
@@ -0,0 +1,175 @@
...
+agents = []
+
+
+# Concrete implementation with docker
+
+rootdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+installdir = os.path.join(rootdir, 'build', 'install')
+
+
+class AlEntityDocker(AlEntity):
```
ghost: Same comment as before: I'd capitalize ALEntity, even if just to avoid confusion (nitpick)

tests/environment.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+# The following variables are initialized as None, and have to be set when a concrete test
+# environment is started.
+wired_sniffer = None
+controller = None
+agents = []
+
+
+# Concrete implementation with docker
```
ghost, line 140: Perhaps move the different implementations of the interface to different files?

tests/environment.py
```diff
@@ -0,0 +1,175 @@
...
+        if is_controller:
+            config_file_name = 'beerocks_controller.conf'
+        else:
+            config_file_name = 'beerocks_agent.conf'
+        with open(os.path.join(installdir, 'config', config_file_name)) as config_file:
+            ucc_port = \
+                re.search(r'ucc_listener_port=(?P<port>[0-9]+)', config_file.read()).group('port')
+
+        device_ip_output = self.command('ip', '-f', 'inet', 'addr', 'show', self.bridge_name)
+        device_ip = re.search(r'inet (?P<ip>[0-9.]+)', device_ip_output.decode('utf-8')).group('ip')
```
ghost: Some sanity checks might be useful here (e.g. check for command failure)
Unless this is dealt with using exceptions, of course ...
(nitpick)

tests/environment.py
ghost: Actually, I notice now that this interface is for CLI commands. Still, the same principle applies :)

tests/environment.py
```diff
@@ -0,0 +1,175 @@
...
+        # This is normally done by test_gw_repeater.sh, but we need it earlier to be able to
+        # start tcpdump
+        # Raise an exception if it fails (check=True).
+        subprocess.run(('docker', 'network', 'create', docker_network), check=True,
+                       stdout=subprocess.DEVNULL)
+        # Inspect again, now raise if it fails (check=True).
+        inspect_result = subprocess.run(docker_network_inspect_cmd, check=True,
+                                        stdout=subprocess.PIPE)
+
+    inspect = json.loads(inspect_result.stdout)
```
ghost: Once again, sanity checks might be useful here (e.g. check that `inspect is not None`)

tests/environment.py.saved
```diff
@@ -0,0 +1,173 @@
+###############################################################
```
ghost: Is this a backup file generated by your IDE? If so, I don't think it should be in version control ;)

tests/sniffer.py
```diff
@@ -0,0 +1,46 @@
...
+            os.makedirs(os.path.join(opts.tcpdump_dir, 'logs'), exist_ok=True)
+            outputfile = os.path.join(opts.tcpdump_dir, outputfile_basename) + ".pcap"
+            command = ["tcpdump", "-i", self.interface, "-w", outputfile]
+            self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
+            # tcpdump takes a while to start up. Wait for the appropriate output before continuing.
+            # poll() so we exit the loop if tcpdump terminates for any reason.
+            while not self.tcpdump_proc.poll():
+                line = self.tcpdump_proc.stderr.readline()
+                debug(line.decode()[:-1])  # strip off newline
+                if line.startswith(b"tcpdump: listening on " + self.interface.encode()):
```
ghost: Q: if you have your language set to French, would you get tcpdump output like "tcpdump: en train de écouter sur interface br-lan")? Because if so, this code has the potential to block forever for some people

ghost (approved):
Looks good in general, though there are some minor issues here and there you might want to look at

tests/environment.py
arnout: Not really PEP8 naming but probably a good idea indeed.

tests/environment.py
arnout: Running commands on the device.

tests/environment.py
arnout: I prefer to do keep it in a single file until we start doing the boardfarm stuff. I expect that to require moving stuff around between the abstract classes and the implementations, and that's easier if it's in a single file. Well, not by much, but still.

tests/environment.py
arnout: True, but that wasn't there before and there's already plenty in this PR :-)

tests/environment.py
arnout: json.loads will raise an exception if it's not JSON, and the lookups will raise KeyError. So I don't see why we should bother.

tests/sniffer.py
arnout: Good point, but not for this PR.

tests/environment.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def _docker_wait_for_log(container: str, program: str, regex: str, timeout: float) -> bool:
+    logfilename = os.path.join(rootdir, 'logs', container, 'beerocks_{}.log'.format(program))
+    deadline = time.monotonic() + timeout
+    try:
+        while True:
+            with open(logfilename) as logfile:
+                for line in logfile.readlines():
+                    if re.search(regex, line):
```
VladyslavTupikin: This implementation does not include a case when logfile contains 2 or more same words. It just takes the first found word.

arnout: Well, that feature is something that you added, no?

rmelotte (approved):
LGTM

PR #1032, opened 2020-03-26
documentation: prplMesh to boarfarm integration

abelog:
## Description

This PR represents proposal of way to integrate, organise and move current `test_flows.py` functionality to [boardfarm](https://github.com/mattsm/boardfarm/) project.
As tests and devices we test are not yet covered by this project, we have to implement them on our own.

## Contains:
- Classes organisation inside of `boardfarm`
- Methods, fields
- Typical set up schematic
- General single test flow

Main goal of this PR is to gather feedback, agreed on and have a plan (and documentation!) before moving to actual implementation.

## Not covered
Particular tests implementation. Test suites organisation.

**NOTE**: This PR is work in progress. But any suggestions are highly welcome.

### DIFF
--- a/documentation/images/plantuml/boardfarm_prplMesh_classes.puml
+++ b/documentation/images/plantuml/boardfarm_prplMesh_classes.puml
@@ -1,0 +1,155 @@
+@startuml
+package boardfarm_generic <<Frame>> {
+
+    class LinuxDevice {
+    }
+
+    class OpenWrtRouter {
+        string prompt
+        string lan_iface
+        string wan_iface
+        IPv4Network lan_network
+        IPv4Address lan_gateway
+
+        __init__()
+        install_package()
+        wait_for_boot()
+        get_wan_iface()
+        get_wan_proto()
+        config_wan_proto()
+        enable_ssh()
+        uci_allow_wan_ssh()
+    }
+}
+
+package prplMesh_devices <<Frame>> {
+
+    class prplWRT {
+        string model
+        IPv4Network wan_network
+        IPv4Address wan_gateway
+
+        check_status()
+    }
+
+    class prplMesh_docker {
+        string model
+        string wan_iface
+        string lan_iface
+        IPv4Network lan_network
+        IPv4Address lan_gateway
+        IPv4Network wan_network
+        IPv4Address wan_gateway
+
+        check_status()
+    }
+
+    class sniffer {
+        int listening_ports
+        string listening_proto
+        string iface
+
+        catch_message()
+        set_port()
+        set_proto()
+        start_monitor()
+        stop_monitor()
+    }
+
+    class wired_sniffer {
+        string model
+
+        __init__()
+    }
+
+    class wireless_sniffer {
+        string model
+        uint channel
+
+        __init__()
+        set_ssid()
+        set_bsid()
+        get_channel()
+        get_bandwith()
+    }
+}
+
+package prplMesh_entities <<Frame>> {
+
+    class prplMesh_agent {
+        int port
+        IPv4Address ip
+
+        __init__()
+        get_radio()
+        get_status()
+        check_log()
+    }
+
+    class prplMesh_radio {
+        string mac
+        int status
+
+        get_mac()
+        get_status()
+        get_txpower()
+        get_rxpower()
+    }
+
+    class prplMesh_radio_dummy {
+
+        check_log()
+        get_mac()
+        get_status()
+        get_txpower()
+        get_rxpower()
+    }
+
+    class prplMesh_radio_hw {
+
+        get_mac()
+        get_status()
+        get_txpower()
+        get_rxpower()
+    }
+
+    class prplMesh_controller {
+        int port
+        IPv4Address ip
+
+        __init__()
+        get_alid()
+        get_status()
+        check_log()
+    }
+}
+
+package tests {
+
+    class CAPI {
+        string dest_alid
+        string payload
+
+        send()
+        set_alid()
+        get_alid()
+        set_payload()
+        get_payload()
+    }
+}
+
+prplMesh_docker --* prplMesh_controller
+prplMesh_docker --* prplMesh_agent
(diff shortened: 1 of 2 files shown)


VladyslavTupikin (approved):

ghost:
So, in this diagram, the OpenWrtRouter class inherits from the rax40 class.
Isn't this backwards? Or are we saying that every OpenWrtRouter is a kind of rax40?

abelog:
It's backwards, you are right. Fixed!

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
```diff
@@ -0,0 +1,165 @@
...
+    wait_for_network()
+    get_memfree()
+    start_tinyproxy()
+    take_lock()
+    release_lock()
+    perform_curl()
+}
+
+class OpenWrtRouter {
+    string promt
```
arnout:
```suggestion
    string prompt
```
?

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
```diff
@@ -0,0 +1,165 @@
...
+    IPv4Network wan_network
+    IPv4Address wan_gateway
+    prplMesh_agent [..]
+
+    kill_console_at_exit()
+    check_logs()
+    check_status()
+}
+
+class backhaul_sniffer {
```
arnout: Although the terms fronthaul_sniffer and backhaul_sniffer are technically correct, I find them confusing. I would prefer to call them wired_sniffer and wireless_sniffer.

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
```diff
@@ -0,0 +1,165 @@
...
+    string model
+
+    __init__()
+    start_monitor()
+    stop_monitor()
+    catch_message()
+    set_ssid()
+    set_bsid()
+    get_channel()
+    get_bandwith()
```
arnout: AFAIU, the wireless sniffer should use monitor mode and therefore it should be configured with the channel to monitor, not with ssid and bssid.

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
```diff
@@ -0,0 +1,165 @@
...
+}
+
+class prplMesh_agent {
+    int port
+    IPv4Address ip
+    prplMesh_repeater[..]
```
arnout: This is already represented with the relationship `*--` below, so no need to repeat it here.

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
```diff
@@ -0,0 +1,165 @@
+@startuml
+class LinuxDevice {
+    string tftp_dir
```
arnout: I think the UML file is there to convey a message, not to have a one-to-one description of the code - the code is better at doing that :-). So, I think only the members that are really interesting (i.e. will be used in our tests) should be mentioned here.

arnout (commented):
I'm always a bit ambivalent about including UML files in the repo, because they are always inconsistent with reality. So I would be more inclined to put it on the wiki, since there things are expected to be inconsistent :-). That said, the whole point of having UML files is to be able to discuss the architecture, and that's easier in a PR than on the wiki.

Maybe an intermediate solution is to handle it in an issue.

Anyway, about the architecture proposal: looks OK to me in general. I think it should be merged with the UML file that I created to get an overall picture of where we are going.

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
abelog: Actually, difference between wireless and wired sniffer is in "listener" set up: 
- For wired we just have to call `tcpdump` and point ports + iface to listen to (probably setting proto will also make sense), just to minimize garbage amount and ease search.
- For wireless we have to additionally check monitor capability, put WLAN iface to monitor mode.

Most part of methods, like some abstract `wait_for_capability_report` will be same for both sniffers. So, we even may split it in 3 classes: 
- generic sniffer (has methods to work with `tcpdump`, which are common)
- wired sniffer, extends "generic sniffer", extends with constructor to bring wired `tcpdump`
- wireless sniffer, extends "generic sniffer" with wireless monitor set up methods + constructor

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
abelog: I wanted to use kind of "filter" (look just into frames with specific SSID of BSSID during waiting for some action)  with `tcpdump` to reduce amount of messages to search through.

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
abelog: Fixed in update! I have left only crucial fields and methods in boardfarm classes.

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
abelog: Fixed!

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
abelog: Fixed! Thanks for paying attention.

abelog:
Current state of diagram as image:
image

documentation/images/plantuml/boardfarm_prplMesh_classes.puml
abelog: Updated UML to demonstrate proposal of split up.

arnout (approved):
I don't think it makes sense to discuss more at the level of a UML file, let's just start implementing this and hopefully update the UML file if it turns out a little different. For sure, there is nothing here that is conflicting with what I have in my mind.

PR #1034, opened 2020-03-26
Agent: Does not send operating channel report

### LINKED ISSUE #1033: [BUG] Agent: Does not send operating channel report
**Describe the bug** -
When agent recieves empty channel selection message it does not send channel report message to the controller. Issue found in [this case](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/son_slave_thread.cpp#L4642). 

## Steps to reproduce

1. Build
2. Run controller & agent
3. Send empty channel selection request
4. Check controller logs - it does not contain OPERATING_CHANNEL_REPORT message.

## Expected behavior

Agent should send  OPERATING_CHANNEL_REPORT message when recieved empty channel selection request.

## Target

- Host: Ubuntu 18.04
- Platform: Linux DUMMY mode with ZMQ
- How you built prplMesh - ./docker-builder
- How you ran prplMesh - tests/test_flow.py

VladyslavTupikin:
Agent does not send operating channel report message
when recieve empty channel selection request.

Add in case which process empty channel request message
print in the log and add call send_operating_channel_report()
function.

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -4657,10 +4657,6 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
     int power_limit           = 0;
     bool power_limit_received = channel_selection_get_transmit_power_limit(cmdu_rx, power_limit);
 
-    // No preferences or power limit for current ruid
-    if (channel_preferences.empty() && !power_limit_received)
-        return true;
-
     // According to design only Resticted channels should be included in channel selection request
     auto switch_required = channel_selection_current_channel_restricted();
     beerocks::message::sWifiChannel channel_to_switch;


abelog (approved):

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4639,8 +4639,11 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
     bool power_limit_received = channel_selection_get_transmit_power_limit(cmdu_rx, power_limit);
 
     // No preferences or power limit for current ruid
-    if (channel_preferences.empty() && !power_limit_received)
```
tomereli, line 4661: This is incorrect (was a bug also before this change) - I missed it before, but from what I can understand the specification says that an empty channel preference in the channel selection request is valid, and should be answered with a channel selection response first, and only then an operating channel report.

What you did here is send the operating channel report without a channel selection response, which is wrong.
The solution is to still go through all the stages even for empty channel preference. We should delete the previously stored preference if any, send a channel selection response, and only then send the operating channel report.

@arnout please correct me if I'm wrong.

tomereli (changes requested):

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: > This is incorrect (was a bug also before this change) - I missed it before, but from what I can understand the specification says that an empty channel preference in the channel selection request is valid, and should be answered with a channel selection response first, and only then an operating channel report.

In fact, an empty channel preference indicates that all channels are equally preferred: " If a Channel Selection Request message contains zero Channel Preference TLVs, the Multi-AP Agent shall infer that the Multi-AP Controller is indicating the highest preference (preference score 15) for all channels and operating classes supported by all of the Multi-AP Agent’s radios."

In other words, a channel selection request without any TLVs is perfectly valid, and indicates that the agent can choose whatever channel it likes and can use the maximum transmit power.

> What you did here is send the operating channel report without a channel selection response, which is wrong.
> The solution is to still go through all the stages even for empty channel preference. We should delete the previously stored preference if any, send a channel selection response, and only then send the operating channel report.

Deletion is already done by channel_selection_get_channel_preference(). However, that is not correct. Instead, it should copy the channel preferences as reported by ACS (which would be preference 15 for all channels if no ACS is done), and then updating those preferences based on the CMDU. Also, it shouldn't add to the channel preferences, but only update the preference value for all the channels mentioned in the channel preferences as reported by ACS.

Thinking a bit more about it, I think that we should reconsider the work that @itayx and @kantera800 are doing now... Maybe we should keep the list of supported channels in son_slave as is, so not make a distinction between supported channels and channel preferences. When we receive a channel selection request, we update the preference value of all channels. And since we already set the preference values to our own preference, we should take care that it's own going to lower preference (and report DECLINE_VIOLATES_MOST_RECENTLY_REPORTED_PREFERENCES if the controller specified a higher preference than what we reported to it or if it leaves no operable channel). Then we just go over the channels list and take the one with the highest preference. 

And at some point we'll have to add DFS support - it seems we currently only select non-DFS channels.

And then for the bandwidths... Remember that each "channel preference" is actually a preference for a channel+bandwidth combination. So we should probably prefer a channel with the highest bandwidth if there are several channels with the same preference... And probably the same for Tx power limit (I assume you'll get better performance with higher Tx power).

Bottom line, it's going to get a little bit more complicated than this pull request...

arnout:
For a quick win, however, I think we can just remove the condition entirely. If channel_preferences is empty, then the current channel is not restricted so switch_required will be false, and we fall in the `!switch_required && !power_limit_received` condition below.

tomereli (approved):
LGTM, but please run the channel selection tests in the CI before merging, to see we didn't miss anything

PR #1036, opened 2020-03-26
CODINGSTYLE: add object types section

morantr:
Add a standard for objects and types definition and usage, to try and minimize
the number of review comments related to it.

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -14,13 +14,57 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
 
 ### Naming convention
 
-* Almost everything uses `lower_snake_case`.
+* Local variables uses `lower_snake_case`.
 * `#define` uses `UPPER_SNAKE_CASE`.
 * Everything is in a namespace.
 * Therefore, there are no prefixes.
-* Private data members have an "m" prefix: `m_data`.
-* Getter functions use the name without "m": `data()`.
-* Setter functions use the "set" prefix: `set_data(value)`.
+
+#### Structs
+
+Structs shall only be used for "plain-old-data" POD type with no logic or any functions except a data member initialization function (constructor).
+Structs shall not have protected or private members.
+Structs shall not inherit, not shall others inherit from it.
+
+* Struct definition shall use camel case with `s` prefix: `sSomeStruct`.
+* Struct members and initialization function shall use `lower_snake_case` with no prefix.
+* Struct shall not be defined with `typedef` keyword on C++ code.
+
+```cpp
+struct sSomeStruct {
+    int x_foo;
+    int y_bar;
+};
+```
+
+#### Classes
+
+Classes shall be used to create complex object with functions and logic in it.
+
+* Class definition shall use Pascal Case: `SomeClass`.
+* Class members and functions shall use `lower_snake_case`. 
+* Class data members have an `m_` prefix: `m_data`.
+
+```cpp
+class SomeClass {
+public:
+    SomeClass();
+    ~SomeClass()
+
+    set_foo(int foo);
+    int calculate_something(int param_1, int param_2);
+private:
+    int x_foo;
+    int y_bar;
+};
+```
+
+#### Other types
+
+Any other type shall be defined with `lower_snake_case` with `_t` suffix:
+
+```cpp
+typedef std::list<std::string> string_list_t;
+```
 
 ### Coding patterns
 


CODINGSTYLE.md
```diff
@@ -14,13 +14,57 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
-* Getter functions use the name without "m": `data()`.
-* Setter functions use the "set" prefix: `set_data(value)`.
+
+### Objects types
+
+#### Structs
+
+Structs shall only be used for "plain-old-data" POD type with no logic or any functions except a data member initialization function (constructor).
+
+* Struct definition shall use camel case with `s` prefix: `sSomeStruct`.
```
arnout, line 28: We discussed this a bit more than half a year ago. The code currently has roughly 50% sCamelCase and 50% snake_case. Back then, we decided that we would use snake_case (except to keep consistency within the same file if it is already using sCamelCase). Is there any reason to override that decision now?

CODINGSTYLE.md
```diff
@@ -14,13 +14,57 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
 * Therefore, there are no prefixes.
-* Private data members have an "m" prefix: `m_data`.
-* Getter functions use the name without "m": `data()`.
-* Setter functions use the "set" prefix: `set_data(value)`.
+
+### Objects types
+
+#### Structs
+
+Structs shall only be used for "plain-old-data" POD type with no logic or any functions except a data member initialization function (constructor).
```
arnout, line 24:
```suggestion
Structs shall only be used for "plain-old-data" POD type with no logic or any functions except a data member initialization function (constructor).
Structs shall not have protected or private members.
Structs shall not inherit, not shall others inherit from it.
```

CODINGSTYLE.md
```diff
@@ -14,13 +14,57 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
+    int x_foo;
+    int y_bar;
+};
+```
+
+#### Classes
+
+Classes shall be used to create complex object with functions and logic in it.
+
+* Class definition shall use Pascal Case: `SomeClass`.
```
arnout, line 43: Same here: we decided before to use lower snake case.

CODINGSTYLE.md
```diff
@@ -14,13 +14,57 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
+    int calculate_something(int param_1, int param_2);
+private:
+    int x_foo;
+    int y_bar;
+};
+```
+
+#### Other types
+
+Any other type shall be defined with `lower_snake_case` with `_t` suffix:
```
arnout, line 63: For this, no decision had been taken before. However, we currently have about 40 typedefs (not counting the 'typedef struct' cases which shouldn't be typedefs to begin with), and only 14 with `_t`. That said, looking at the ones without `_t`, they are either just shorthand for class names in other namespaces, or they are already violating the naming convention in a thousand other ways (I'm looking at you, bml...). So I think this is definitely a good rule.

CODINGSTYLE.md
```diff
@@ -14,13 +14,57 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
-* Almost everything uses `lower_snake_case`.
+* Local variables uses `lower_snake_case`.
 * `#define` uses `UPPER_SNAKE_CASE`.
 * Everything is in a namespace.
 * Therefore, there are no prefixes.
-* Private data members have an "m" prefix: `m_data`.
-* Getter functions use the name without "m": `data()`.
-* Setter functions use the "set" prefix: `set_data(value)`.
+
+### Objects types
```
arnout: The text below is really still about naming convention. Even "when to use struct and when to use class" can be considered naming convention. So I wouldn't add this subtitle and just make the rest subsections of Naming convention.

arnout (commented):
I can't approve this PR, since it's actually a change in coding style / naming convention that was not discussed before. I'm personally a fan of the naming convention we had, because it's simple. And I was the one who wrote it down to begin with :-).

So please open a vote on #democracy before merging this PR.

CODINGSTYLE.md
morantr: The main reason I did it now is since you gave a comment about it on another PR of mine. 😄 
But besides that, I think that it's missing and we need a convention for that.
Using the same convention for type and variables name is annoying IMO since usually we would want to give the same name of the type to the variable itself.

Bad IMO:
```cpp
class some_class {
    // code
};

// This is annoying since it forces me to be creative and give the variable a different name from the type
some_class some_class_obj;
```

Good IMO:
```cpp
class SomeClass {
    // code
};

// Perfect IMO :)
SomeClass some_class;
```

CODINGSTYLE.md
morantr: good idea.

CODINGSTYLE.md
morantr: Ok, Removed the title.

CODINGSTYLE.md
morantr: thanks

CODINGSTYLE.md
tomereli: I tend to agree with @morantr on this one now, at least new classes should use camle case for the class definition, but NOT for struct definition which is what @arnout commented on here - https://github.com/prplfoundation/prplMesh/pull/1019/commits/afd24cfb1715da85219874a09889081d6ba58c6f#r398541333

CODINGSTYLE.md
tomereli: Do you remember why we chose this?

CODINGSTYLE.md
```diff
@@ -14,13 +14,57 @@ In these cases, it can be overridden by surrounding it with `clang-format-off` a
...
+Structs shall not have protected or private members.
+Structs shall not inherit, not shall others inherit from it.
+
+* Struct definition shall use camel case with `s` prefix: `sSomeStruct`.
+* Struct members and initialization function shall use `lower_snake_case` with no prefix.
+* Struct shall not be defined with `typedef` keyword on C++ code.
+
+```cpp
+struct sSomeStruct {
+    int x_foo;
```
tomereli, line 34: This is confusing, why did you call it `x_foo` and `y_foo` and not just `x` and `y`?

tomereli (commented):
Same like Arnout said - all this has to be voted first on #prplmesh-democracy

CODINGSTYLE.md
morantr: To emphasize it is a lower snake case.

tomereli (approved):

arnout (approved):
Results from #prplmesh-democracy poll:

* Coding convention for classes: 9 CamelCase , 1 snake_case
* Naming convention for structs: 6 sPrefixedCamelCase , 1 snake_case, 1 CamelCase

Conclusion: this PR reflects the coding style we prefer.

PR #1039, opened 2020-03-28
logging: change default log suffix to .txt

tomereli:
File extention .log is not standard as .txt which is what the logs
contain, which makes default text editors not to understand that these
are regular text files.
An extremely annoying example is the ownCloud web GUI which doesn't know
how to show .log files therefore we need to download logs in order to
view them, which makes analyzing issues take much longer.
Therefore, change the default log suffix to .txt.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,7 +118,7 @@ set(BEEROCKS_ENABLE_ARP_MONITOR 1)
 # Logging configuration
 set(BEEROCKS_LOG_FILES_ENABLED      "true")
 set(BEEROCKS_LOG_FILES_PATH         "${TMP_PATH}/logs")
-set(BEEROCKS_LOG_FILES_SUFFIX       ".log")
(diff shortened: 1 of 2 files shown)


vitalybu:
I'm not sure this change is necessary. 
The log suffix is quite standard for log files, and it can potentially break prplMesh integration with various platforms.
If the motivation here is to be able to preview the files in ownCloud's web interface, perhaps it's better to configure ownCloud to provide "text/plain" mime type for *.log files:
https://doc.owncloud.org/server/9.0/admin_manual/configuration_mimetypes/mimetype_mapping.html

arnout:
> If the motivation here is to be able to preview the files in ownCloud's web interface, perhaps it's better to configure ownCloud to provide "text/plain" mime type for *.log files:

Done. Thanks for the hint, Vitaly.

@tomereli please close if this resolves the issue for you.

tomereli:
Ok, not needed.

PR #1043, opened 2020-03-29
bwl: refactor read_supported_channels to use get_radio_info

### LINKED ISSUE #1024: [TASK] Refactor read_supported_channels to use nl80211_client::get_radio_info
Refactor ap_wlan_hal_nl80211::read_supported_channels to use nl80211_client::get_radio_info.

PR #1043

itayx:
The function read_supported_channels reads the supported channels from the
hostap's GET_HW_FEATURES output. This only works using a specific patch to the hostap library.
Therefore using the NL to get the supported channels is the more correct path.

Refactor the function read_supported_channels in both
ap_wlan_hal_dwpal and ap_wlan_hal_nl80211 to use the get_radio_info
in the nl80211_client


common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,65 +1527,55 @@ bool ap_wlan_hal_dwpal::read_acs_report()
 
 bool ap_wlan_hal_dwpal::read_supported_channels()
 {
-    char *reply = nullptr;
-    if (!dwpal_send_cmd("GET_HW_FEATURES", &reply)) {
```
kantera800, line 1531: I think that reusing read_supported_channels is a good Idea, but i suggest that you keep the "get _hw_features" implementation and move it to a private function. 
for example `get_supported_channels_from_hostapd`

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,65 +1527,55 @@ bool ap_wlan_hal_dwpal::read_acs_report()
...
 
-    size_t replyLen = strnlen(reply, HOSTAPD_TO_DWPAL_MSG_LENGTH);
-    for (size_t i = 0; i < replyLen; i++) {
-        if (reply[i] == ')') {
-            reply[i] = ' ';
-        }
+    // Clear the supported channels vector
+    m_radio_info.supported_channels.clear();
+    // Resize the supported channels vector
+    m_radio_info.supported_channels.resize(MAX_SUPPORTED_20M_CHANNELS);
```
kantera800: It would feel more comfortable to use`MAX_SUPPORTED_CHANNELS`

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,65 +1527,55 @@ bool ap_wlan_hal_dwpal::read_acs_report()
...
+    bwl::nl80211_client::radio_info radio_info;
```
kantera800, line 1533: i suggest moving all "m_nl80211_client->get_radio_info" related code to a private function
for example `get_supported_channels_from_nl80211`

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -451,122 +451,49 @@ bool ap_wlan_hal_nl80211::read_supported_channels()
     m_radio_info.supported_channels.clear();
     // Resize the supported channels vector
     m_radio_info.supported_channels.resize(MAX_SUPPORTED_20M_CHANNELS);
```
kantera800: same comment here for the  _20M_

kantera800 (changes requested):

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
kantera800: same comment here regarding the _20M_

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: The value of `MAX_SUPPORTED_20M_CHANNELS` is defined to `beerocks::message::SUPPORTED_CHANNELS_LENGTH` which is the current size to use, since that is the size allocated in the outgoing message.
On the other hand, there is a `MAX_SUPPORTED_CHANNELS`, which is set to `(MAX_SUPPORTED_20M_CHANNELS*4)`
Which is larger then the amount I wish to allocate

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
itayx: I've explained in the first comment, why we need to use `MAX_SUPPORTED_20M_CHANNELS`

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: I disagree - in general we don't keep dead code, we have git history for that.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,65 +1527,55 @@ bool ap_wlan_hal_dwpal::read_acs_report()
 
 bool ap_wlan_hal_dwpal::read_supported_channels()
 {
-    char *reply = nullptr;
-    if (!dwpal_send_cmd("GET_HW_FEATURES", &reply)) {
-        LOG(ERROR) << __func__ << " failed";
-        return false;
-    }
+    LOG(TRACE) << __func__ << " for interface: " << get_radio_info().iface_name;
```
tomereli: You don't need `__func__` - it is part of the `LOG(TRACE)` I assume (nitpick)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Why vector and not map as we agreed?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Why? I think its redundant.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,65 +1527,55 @@ bool ap_wlan_hal_dwpal::read_acs_report()
...
-    for (size_t i = 0; i < replyLen; i++) {
-        if (reply[i] == ')') {
-            reply[i] = ' ';
-        }
+    // Clear the supported channels vector
+    m_radio_info.supported_channels.clear();
+    // Resize the supported channels vector
+    m_radio_info.supported_channels.resize(MAX_SUPPORTED_20M_CHANNELS);
+    bwl::nl80211_client::radio_info radio_info;
+    bool ret = false;
```
tomereli:
```suggestion
auto ifname = get_radio_info().iface_name;
if (!m_nl80211_client->get_radio_info(ifname, radio_info)) {
    LOG(ERROR) << "Failed to get radio info from nl80211";
    return false;
}
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,65 +1527,55 @@ bool ap_wlan_hal_dwpal::read_acs_report()
...
-    while (lineMsg != NULL) {
-        size_t lineLen = strnlen(lineMsg, HOSTAPD_TO_DWPAL_MSG_LENGTH);
-
-        // TODO: fieldsToParse.field should be contained in one struct and struct size should be
-        //       passed to the dwpal_string_to_struct_parse - a security fix for multiline reply
-        if (dwpal_string_to_struct_parse(lineMsg, lineLen, fieldsToParse, sizeof(dfs_state)) ==
-            DWPAL_FAILURE) {
-            LOG(ERROR) << "DWPAL parse error ==> Abort";
-            return false;
+    int idx = 0;
```
tomereli: You should not use a vector. This function can be so much simpler if you just use the same data structure used by the nl80211 client implementation:
```
/**
 * Channels supported in this band (obtained from NL80211_BAND_ATTR_FREQS).
 * Map key is the channel number and map value is the channel information.
 */
 std::unordered_map<uint8_t, channel_info> supported_channels;
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,65 +1527,55 @@ bool ap_wlan_hal_dwpal::read_acs_report()
...
-        i++;
-
-        chan = -1;
-        memset(dfs_state, 0, sizeof(dfs_state));
-
-        lineMsg = strtok_s(NULL, &dmaxLen, "\n", &p2str);
     }
-
-    return true;
+    return ret;
```
tomereli: Don't return ret, return `true` or false. This is redundant.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -451,122 +451,49 @@ bool ap_wlan_hal_nl80211::read_supported_channels()
...
+    bwl::nl80211_client::radio_info radio_info;
+    bool ret = false;
+    if (!(ret = m_nl80211_client->get_radio_info(get_radio_info().iface_name, radio_info))) {
```
tomereli:
```suggestion
auto ifname = get_radio_info().iface_name;
if (!m_nl80211_client->get_radio_info(ifname, radio_info)) {
    LOG(ERROR) << "Failed to get radio info from nl80211";
    return false;
}
```

tomereli (changes requested):

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: If we decide that the new `get_radio_info` will fully replace the "current" GET_HW_FEATURES
then I agree with @tomereli 
But if there is a possibility that some other utility will need it,
as with the case of ap_wlan_hal_nl80211, where the `read_supported_channels` is used instead of the read_acs_report
Then we might still need the original code

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: by map i presume you are talking about the AP Radio Basic Capabilities
see #686
I plan on adding the AP Radio Basic Capabilities "map" in the next PR

and as for the vector,
Currently the supported_channels are stored in a vector since they have no regard to the bandwidth, they are translated later on (`ap_manager_thread` & `son_slave_thread`)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: Removed

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
itayx: Changed

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: Changed

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: Changed

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: This is wrong - supported channels must contain channel and bandwidth. Please use the same data structure defined in nl80211_client.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
kantera800: > But if there is a possibility that some other utility will need it,
> as with the case of ap_wlan_hal_nl80211, where the `read_supported_channels` is used

nl80211 bwl cannot be using GET_HW_FEATURES since its INTEL hostapd patch.



> I disagree - in general we don't keep dead code, we have git history for that.

then  lest remove it along with the ap_wlan_hal_nl80211 implementation and use `get_radio_info` implementation for  both BWLs

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
kantera800: so we could easily switch back to Hostapd implementation in the future, but since we are not keeping it then it doesn't matter.

itayx:
@tomereli & @kantera800 
I took into consideration both your notes, and I agree with tomer
The new implementation of get_radio_info needs to replace the current read_supported_channels 
The only issue is that I change the functionality of the read_supported_channels by changing the way the m_radio_info.supported_channels are stored since the code is used in other places other then just during the join operation

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,64 +1527,57 @@ bool ap_wlan_hal_dwpal::read_acs_report()
...
+    if (!m_nl80211_client->get_radio_info(ifname, radio_info)) {
+        LOG(TRACE) << "Failed to get channels info from nl80211";
+        return false;
+    }
+    std::vector<bwl::WiFiChannel> supported_channels;
+    for (auto const &band : radio_info.bands) {
+        for (auto const &pair : band.supported_channels) {
+            auto &channel_info = pair.second;
+            for (auto bw : channel_info.supported_bandwidths) {
+
```
tomereli: remove blank line

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1527,64 +1527,57 @@ bool ap_wlan_hal_dwpal::read_acs_report()
...
+                bwl::WiFiChannel channel;
+                channel.channel = channel_info.number;
+                channel.tx_pow  = channel_info.tx_power;
+                channel.is_dfs  = channel_info.is_dfs;
+                switch (bw) {
```
tomereli: Please change this to use `beerocks::utils::convert_bandwidth_to_int()`

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -445,129 +445,58 @@ bool ap_wlan_hal_nl80211::read_acs_report()
...
+        for (auto const &pair : band.supported_channels) {
+            auto &channel_info = pair.second;
+            for (auto bw : channel_info.supported_bandwidths) {
+
+                bwl::WiFiChannel channel;
+                channel.channel = channel_info.number;
+                channel.tx_pow  = channel_info.tx_power;
+                channel.is_dfs  = channel_info.is_dfs;
+                switch (bw) {
+                case beerocks::eWiFiBandwidth::BANDWIDTH_20:
```
tomereli: Same here - `beerocks::utils::convert_bandwidth_to_int`

tomereli (approved):
Small changes but nothing I need to re-review so approving. Please fix before merging.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
itayx: changed

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: changed

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: removed

tomereli (changes requested):
Please add one more commit which removes `read_supported_channels` from the BWL APIs and instead read it as part of `refresh_radio_info`.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: As agreed currently the supported_channels will be stored as a vector
and in the son_slave_thread where ap radio basic capabilities tlv is created the channels will be stored as a map of channels and operating classes

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: Currently all read_supported_channels (dwpal & nl80211) will use the
`m_nl80211_client->get_radio_info` instead of the old `GET_HW_FEATURES` & the old NL operation

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: With the new changes the supported channels have the channel number, the supported bandwidth, supported tx power & whether the channel is dfs or not

The supported channels are stored in a vector, and later on sent to the son_slave_thread where we build with them the ap radio basic capabilities

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: In the new changes
both BWLs (dwpal & nl80211) use the `get_radio_info` implementation

tomereli (approved):

kantera800 (approved):

PR #1044, opened 2020-03-30
build: change CMAKE REVISION variable to PRPLMESH_REVISION

tomereli:
Building IPK from tools/builder/openwrt/scripts/build.sh is done as
follows:
make package/feeds/prpl/prplmesh/prepare USE_SOURCE_DIR="/home/openwrt/prplMesh" V=s
make package/feeds/prpl/prplmesh/compile V=sc -j"$(nproc)"

The prplMesh revision comes out wrong since the REVISION variable is
passed from the OpenWRT package Makefile AND is exported by the OpenWRT
build system anyway.
This is simply confusing since the revision showing in the logs, and in
the config/version file, are wrong.

In order to fix this, the feed Makefile needs to be modified not to pass
REVISION when USE_SOURCE_DIR is defined, but in addition, prplmesh build
system should use a unique variable for the revision otherwise it will get
the REVISION exported by the OpenWRT build system.

This commit changes REVISION to PRPLMESH_REVISION so that it will use
the revision passed from the OpenWRT feed Makefile.
Also, since this commit requires a change in the feed, update it as well
in the build script - relevant commit in the feed:
https://git.prpl.dev/prplmesh/iwlwav/-/commit/6265989f2c47f7b42f64d25f455443b87902582e.

### DIFF
--- a/agent/CMakeLists.txt
+++ b/agent/CMakeLists.txt
@@ -1,6 +1,6 @@
 add_definitions(-DBEEROCKS_VERSION="${prplmesh_VERSION}")
 add_definitions(-DBEEROCKS_BUILD_DATE="${BUILD_DATE}")
-add_definitions(-DBEEROCKS_REVISION="${REVISION}")
+add_definitions(-DBEEROCKS_REVISION="${PRPLMESH_REVISION}")
 
 add_definitions(-DBEEROCKS_BIN_PATH="${INSTALL_PATH}/bin/")
 add_definitions(-DBEEROCKS_CONF_PATH="${INSTALL_PATH}/config/")
--- a/cmake/multiap-helpers.cmake
+++ b/cmake/multiap-helpers.cmake
@@ -29,12 +29,12 @@ if (${FORCE_COLORED_OUTPUT})
     endif ()
 endif ()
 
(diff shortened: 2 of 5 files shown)


vitalybu (approved):

arnout (approved):
agent/CMakeLists.txt and controller/CMakeLists.txt have `add_definitions(-DBEEROCKS_REVISION="${REVISION}")` - that needs to be fixed up as well. In general, if you change something like this, do `git grep -w REVISION` to verify you caught everything :-)

tomereli:
> git grep -w REVISION

Wow how did I miss that? This deserved a "changes requested".

arnout:
> Wow how did I miss that? This deserved a "changes requested".

No it doesn't, because github review workflow sucks a bit. If I put it as changes requested, I *have* to re-review it. But I don't want to, that would be just a waste of time. So even though there are changes that really need to be made, I trust you to make them (correctly) without requiring re-review, so I approve already.

If that needs to be made more explicit, we could use the "don't merge" label for that. But I think it is sufficient to manually set the ready for merge label - I assume you wouldn't add that label without checking the reviews.

PR #1049, opened 2020-03-30
Run tests with tcpdump in CI

arnout:
Enable the tcpdump option of test_flows.py when running in CI.
This makes is easier to debug issues in CI, e.g. #1029.

This requires a fix taken from #1031 to make it possible to start tcpdump before test_gw_repeater.sh is called.
Cherrypicking just that single patch was a bit tricky, so I've taken along two other patches as well.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -101,7 +101,7 @@ run-tests:
   script:
     # Try to pull from dockerhub, if it fails, the run should fail
     - tools/docker/image-pull.sh
-    - tests/test_flows.py -v -u $RANDOM
+    - tests/test_flows.py -v -u $RANDOM --tcpdump
 
   artifacts:
     paths:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -23,7 +23,7 @@ flake8:
   stage: build
   image: pipelinecomponents/flake8:f087c4c
   variables:
-      FLAKE8_FILES: "tests/capi.py"
+      FLAKE8_FILES: "tests/capi.py tests/opts.py"
   script:
     - flake8 --verbose $FLAKE8_FILES
 
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -16,6 +16,7 @@ from typing import Dict
 import json
 
 from capi import tlv, UCCSocket
+from opts import debug, err, message, opts, status
 
 '''Regular expression to match a MAC address in a bytes string.'''
 RE_MAC = rb"(?P<mac>([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})"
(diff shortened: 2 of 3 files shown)


tests/test_flows.py
```diff
@@ -81,8 +81,21 @@ def tcpdump_start(self):
         if opts.tcpdump:
             outputfile = os.path.join(self.rootdir, 'logs', 'test_{}.pcap'.format(self.running))
             debug("Starting tcpdump, output file {}".format(outputfile))
-            inspect = json.loads(subprocess.check_output(('docker', 'network', 'inspect',
-                                                          'prplMesh-net-{}'.format(self.opts.unique_id))))
+            docker_network = 'prplMesh-net-{}'.format(self.opts.unique_id)
```
tomereli: It will be more appropriate to move this to a separate method.

tests/test_flows.py
```diff
@@ -81,8 +81,21 @@ def tcpdump_start(self):
...
+            docker_network_inspect_cmd = ('docker', 'network', 'inspect', docker_network)
+            inspect_result = subprocess.run(docker_network_inspect_cmd, stdout=subprocess.PIPE)
+            if inspect_result.returncode != 0:
+                # Assume network doesn't exist yet. Create it.
+                # This is normally done by test_gw_repeater.sh, but we need it earlier to be able to
+                # start tcpdump
+                # Raise an exception if it fails.
+                subprocess.run(('docker', 'network', 'create', docker_network), check=True,
+                               stdout=subprocess.DEVNULL)
+                # Inspect again, now raise if it fails.
```
tomereli: I might be missing something here, but where is the `raise Exception` ?

tests/test_flows.py
```diff
@@ -81,8 +81,21 @@ def tcpdump_start(self):
...
+                inspect_result = subprocess.run(docker_network_inspect_cmd, check=True,
+                                                stdout=subprocess.PIPE)
+
+            inspect = json.loads(inspect_result.stdout)
             prplmesh_net = inspect[0]
```
tomereli: This as well should be now a separate function IMO.

.gitlab-ci.yml
```diff
@@ -101,7 +101,7 @@ run-tests:
   script:
     # Try to pull from dockerhub, if it fails, the run should fail
     - tools/docker/image-pull.sh
-    - tests/test_flows.py -v -u $RANDOM
+    - tests/test_flows.py -v -u $RANDOM --tcpdump
```
tomereli, line 104: Can we make tcpdump enabled by default? Any reason not to?

tomereli (approved):
Minor comments but looking good so approving

tests/test_flows.py
arnout: It's in the `check = True`.

.gitlab-ci.yml
arnout: Well, it will be once #1020 matures.

tests/test_flows.py
arnout: Good point, but it will make rebase difficult so I'll do that later.

arnout: Actually, I needed to make some other changes anyway that will lead to conflicts during the rebase, so I did it now after all.

rmelotte (approved):

PR #1050, opened 2020-03-30
Feature/ci upload artifacts to owncloud 2

rmelotte:
When a pipeline succeeds on the master branch, upload the artifacts to owncloud: https://ftp.essensium.com/owncloud/index.php/apps/files/?dir=/artifacts&fileid=735563.

To make sure there is always a "latest" folder available with all the artifacts, they are not directly uploaded to the latest folder. They are first uploaded to a temporary folder, then on success the folder replaces the "latest" one.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,3 +1,8 @@
+stages:
+  - build
+  - test
+  - upload
+
 .in-prplmesh-builder:
   image:
     name: prplfoundationinc/prplmesh-builder:ubuntu-18.04
@@ -82,6 +87,21 @@ build-in-docker-static:
   extends: build-in-docker
   variables:
     EXTRA_CMAKE_FLAGS: "-DBUILD_SHARED_LIBS=OFF"
+
+upload-artifacts:
+  stage: upload
+  extends: .in-prplmesh-builder
+  script:
+    - ln -s "$CI_OWNCLOUD_PASSWORD" ~/.netrc
+    - sed -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
+    - ci/upload_to_owncloud.sh -v --overwrite-remote-path "artifacts/$CI_JOB_ID" artifacts/latest ./build
+  rules:
+    - if: $CI_COMMIT_BRANCH == "master"
+  dependencies:
+    - build-in-docker
+    - build-for-netgear-rax40
+    - build-for-glinet-b1300
+    - build-for-turris-omnia
 
 .run-test-in-docker:
   stage: test
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -95,8 +95,6 @@ upload-artifacts:
     - ln -s "$CI_OWNCLOUD_PASSWORD" ~/.netrc
     - sed -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
     - ci/upload_to_owncloud.sh -v --overwrite-remote-path "artifacts/$CI_JOB_ID" artifacts/latest ./build
-  rules:
-    - if: $CI_COMMIT_BRANCH == "master"
   dependencies:
     - build-in-docker
     - build-for-netgear-rax40
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ Alternatively, [tools/maptools.py](tools/README.md) can be used to build and ins
 
 Note - to build and run with Docker, see provided [building with docker](tools/docker/README.md)
 
+Builds for the master branch are also available through the following link: https://ftp.essensium.com/owncloud/index.php/s/6ob8ka1zYORDOVA .
+
 ## Running Instructions
 
 Once built, prplMesh controller, agent and framework can be started using `prplmesh_utils.sh`:
--- a/ci/upload_to_owncloud.sh
+++ b/ci/upload_to_owncloud.sh
@@ -13,10 +13,13 @@ rootdir=$(realpath "$scriptdir/../")
 . "$rootdir/tools/functions.sh"
 
 usage() {
-    echo "usage: $(basename "$0") [-hv] <user> <password> <remote-path> <local-path>"
+    echo "usage: $(basename "$0") [-hv] [--overwrite-remote-path <temp-path>] <user> <password> <remote-path> <local-path>"
     echo "  options:"
     echo "      -h|--help - display this help."
-    echo "      -v|--verbose - enable verbosity"
+    echo "      -v|--verbose - enable verbosity."
+    echo "      --overwrite-remote-path - replace any existing folder on remote-path with local-path."
+    echo "                      The first argument is the path to a temporary folder to use."
+    echo "                      The temporary folder is first emptied."
     echo "      -u|--url - webDAV URL (default https://ftp.essensium.com/owncloud/remote.php/dav/files/USERNAME)"
     echo
     echo "  positional arguments:"
@@ -30,7 +33,7 @@ usage() {
 }
 
 main() {
-    if ! OPTS=$(getopt -o 'hvu:' --long help,verbose,url: -n 'parse-options' -- "$@"); then
+    if ! OPTS=$(getopt -o 'hvu:' --long help,verbose,overwrite-remote-path:,url: -n 'parse-options' -- "$@"); then
         echo echo "Failed parsing options." >&2; usage; exit 1
     fi
 
@@ -40,6 +43,7 @@ main() {
         case "$1" in
             -h | --help)            usage; exit 0;;
             -v | --verbose)         export VERBOSE=true; QUIET=; shift;;
+            --overwrite-remote-path)REMOTE_TEMP_PATH="$2"; shift 2;;
             -u | --owncloud-url)    OWNCLOUD_URL="$2"; shift 2;;
             -- ) shift; break ;;
             * ) err "unsupported argument $1"; usage; exit 1;;
@@ -50,6 +54,17 @@ main() {
     remote_path="$1"; shift
     [ -n "$1" ] || { usage; err "Missing local-path"; exit 1; }
     local_path="$1"; shift
+    [ -z "${REMOTE_TEMP_PATH-unset}" ] && { usage; err "Temporary remote path is set but empty!"; exit 1; }
+
+    if [ -n "$REMOTE_TEMP_PATH" ] ; then
+        UPLOAD_URL="$OWNCLOUD_URL/$user/$REMOTE_TEMP_PATH"
+        info "Using temporary path on remote: \"$UPLOAD_URL\""
+        curl ${QUIET:+-s -S} -f -n -X DELETE "$UPLOAD_URL/" > /dev/null 2>&1
+        curl ${QUIET:+-s -S} -f -n -X MKCOL "$UPLOAD_URL/"
+    else
+        UPLOAD_URL="$OWNCLOUD_URL/$user/$remote_path"
+    fi
+
     OWNCLOUD_BROWSE_URL="https://ftp.essensium.com/owncloud/index.php/apps/files/?dir=$remote_path/$(basename "$local_path")"
 
     info "upload $local_path to $OWNCLOUD_BROWSE_URL/$remote_path/$(basename "$local_path") (using user $user)"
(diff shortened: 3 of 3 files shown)


ci/upload_to_owncloud.sh
```diff
@@ -13,10 +13,13 @@ rootdir=$(realpath "$scriptdir/../")
 . "$rootdir/tools/functions.sh"
 
 usage() {
-    echo "usage: $(basename "$0") [-hv] <user> <password> <remote-path> <local-path>"
+    echo "usage: $(basename "$0") [-hv] [--overwrite-remote-path <temp-path>] <user> <password> <remote-path> <local-path>"
     echo "  options:"
     echo "      -h|--help - display this help."
-    echo "      -v|--verbose - enable verbosity"
+    echo "      -v|--verbose - enable verbosity."
+    echo "      --overwrite-remote-path - replace any existing folder on remote-path with local-path."
```
tomereli, line 20: This option is unclear to me. I would expect this to be a true/false value, which means whether or not to overwrite files and folders in the remote if they already exist.
I understand why you would want to first upload to a temporary location - so that we won't get to a case where there is no prplmesh.ipk in owncloud, which is great, but I think you should use a hard coded temporary path and keep the option simple for the user.

.gitlab-ci.yml
```diff
@@ -83,6 +88,21 @@ build-in-docker-static:
...
 
+upload-artifacts:
+  stage: upload
+  extends: .in-prplmesh-builder
+  script:
+    - ln -s "$CI_OWNCLOUD_PASSWORD" ~/.netrc
+    - sed -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
+    - ci/upload_to_owncloud.sh -v --overwrite-remote-path "artifacts/$CI_JOB_ID" artifacts/latest ./build
+  rules:
+    - if: $CI_COMMIT_BRANCH == "master"
```
tomereli: Perhaps we should keep the artifacts in owncloud even for non-master builds, and in addition overwrite `artifacts/latest for master builds?

README.md
```diff
@@ -64,6 +64,8 @@ Alternatively, [tools/maptools.py](tools/README.md) can be used to build and ins
 
 Note - to build and run with Docker, see provided [building with docker](tools/docker/README.md)
 
+Builds for the master branch are also available through the following link: https://ftp.essensium.com/owncloud/index.php/s/6ob8ka1zYORDOVA .
```
tomereli, line 67: I think this does not belong in this README - should be in the WiKi instead IMO.

tomereli (approved):
Looks fine to me, some minor comments but nothing critical. If you tested directory upload to owncloud works fine with this, I don't see a reason why this shouldn't be merged soon.

ci/upload_to_owncloud.sh
rmelotte: I agree with you the option is obscure, but we cannot use a hardcoded path as CI jobs run concurrently (you could end up having 2 pipelines uploading to the same folder at the same time...).

.gitlab-ci.yml
rmelotte: The annoying thing here is that we have to do a move at the end to make sure the "latest" folder always points to something valid, which means we won't keep older artifacts for the master branch (ideally we would use and update a symlink, but from a quick look it doesn't seem possible with webdav).
We could upload and keep the artifacts for non-master branches regardless, but I found it weird to keep those only and not the ones from master.

ci/upload_to_owncloud.sh
```diff
@@ -53,11 +53,15 @@ main() {
...
-            realpath {} --relative-to="$(dirname "$local_path")" \; | \
+            realpath {} --relative-to="$(dirname "$local_path")" \; | {
+                error=0
                 while read -r -s dir; do
                     printf . # show progress
                     dbg "Create directory: $dir"
-                    curl ${QUIET:+ -s -S} -f -n -X MKCOL "$OWNCLOUD_URL/$user/$remote_path/$dir"
+                    curl ${QUIET:+ -s -S} -f -n -X MKCOL "$OWNCLOUD_URL/$user/$remote_path/$dir" || {
+                        error="$?"
+                        err "Failed to create dir: $remote_path/$dir/ (error $error)"
```
arnout, line 79: Does it make sense to continue this iteration if the directory creation did not succeed?

ci/upload_to_owncloud.sh
```diff
@@ -50,6 +50,7 @@ main() {
     remote_path="$1"; shift
     [ -n "$1" ] || { usage; err "Missing local-path"; exit 1; }
     local_path="$1"; shift
+    OWNCLOUD_BROWSE_URL="https://ftp.essensium.com/owncloud/index.php/apps/files/?dir=$remote_path/$(basename "$local_path")"
```
arnout, line 68: This can't be right - just below, the remote_path and local_path are added to it again. 

This definition should move down to where OWNCLOUD_BROWSE_URL is defined the first time.

Please use the public URL, so people who have no owncloud account can also click the link: `https://ftp.essensium.com/owncloud/index.php/s/ketl3eFhIgwcweZ?path=`. Note that this requires `$remote_path` to *not* include the certification/ bit.

ci/upload_to_owncloud.sh
arnout: We can generate a temporary name - using uuidgen, pwgen, `dd if=/dev/urandom bs=24 count=1 | sha256sum`, whatever. With 20 random bytes and a halfway decent RNG the chance of conflicts is about as large as that the universe is destroyed _right_this_instant_ :-)

ci/upload_to_owncloud.sh
```diff
@@ -50,6 +54,17 @@ main() {
     remote_path="$1"; shift
     [ -n "$1" ] || { usage; err "Missing local-path"; exit 1; }
     local_path="$1"; shift
+    [ -z "${REMOTE_TEMP_PATH-unset}" ] && { usage; err "Temporary remote path is set but empty!"; exit 1; }
```
arnout, line 57: Just to be clear - I think we should *always* upload to a randomly generated REMOTE_TEMP_PATH. Which means, no need for the option since we'll always use it.

ci/upload_to_owncloud.sh
```diff
@@ -50,6 +54,17 @@ main() {
...
+
+    if [ -n "$REMOTE_TEMP_PATH" ] ; then
+        UPLOAD_URL="$OWNCLOUD_URL/$user/$REMOTE_TEMP_PATH"
```
arnout, line 60: I think we should put the temp stuff together in one directory, so we can easily clean it up if anything is left behind from crashed scripts:
```suggestion
        UPLOAD_URL="$OWNCLOUD_URL/$user/temp_upload/$REMOTE_TEMP_PATH"
```

.gitlab-ci.yml
```diff
@@ -83,6 +88,21 @@ build-in-docker-static:
   variables:
     EXTRA_CMAKE_FLAGS: "-DBUILD_SHARED_LIBS=OFF"
 
+upload-artifacts:
+  stage: upload
+  extends: .in-prplmesh-builder
+  script:
+    - ln -s "$CI_OWNCLOUD_PASSWORD" ~/.netrc
+    - sed -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
```
arnout, line 96: This weirdness warrants some comment to explain what is going on. Even I am not sure I understand it - you symlink a file (which I suppose is some file that exists on the runner) and then you do a sed -i which IIUC makes a copy of the file...

.gitlab-ci.yml
arnout: I would upload everything to a folder with the commit hash in its name, and just upload a second time to latest only for master.

.gitlab-ci.yml
```diff
@@ -83,6 +88,21 @@ build-in-docker-static:
...
+    - ci/upload_to_owncloud.sh -v --overwrite-remote-path "artifacts/$CI_JOB_ID" artifacts/latest ./build
+  rules:
+    - if: $CI_COMMIT_BRANCH == "master"
+  dependencies:
+    - build-in-docker
+    - build-for-netgear-rax40
+    - build-for-glinet-b1300
+    - build-for-turris-omnia
```
arnout, line 102: I think there's a bug in the commit message: "because we don't want the job to start if
an unrelated job from a previous stage failed" -> because we *do* want the job to start if *only* an unrelated job from a previous stage failed?

README.md
arnout: I would think it is very normal for a README to specify where you can download binaries.

.gitlab-ci.yml
```diff
@@ -95,8 +95,6 @@ upload-artifacts:
     - ln -s "$CI_OWNCLOUD_PASSWORD" ~/.netrc
     - sed -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
     - ci/upload_to_owncloud.sh -v --overwrite-remote-path "artifacts/$CI_JOB_ID" artifacts/latest ./build
-  rules:
-    - if: $CI_COMMIT_BRANCH == "master"
```
arnout: Nice to have this commit for testing. However, you should have made sure it doesn't have your Sob. That way, the DCO check will fail and you can be sure it won't get merged accidentally.

arnout (approved):
As usual: important comments, but no re-review required.

README.md
tomereli: OK, I actually used it now in another script so it is important that it is documented in the repository.

ci/upload_to_owncloud.sh
rmelotte: The idea behind a configurable temporary name was to be able to use the job ID in CI so that we can easily identify the jobs if needed, but it's not that useful for now so ok I'll use uuidgen.

.gitlab-ci.yml
rmelotte: That's the other option, but I couldn't convince myself to upload the same things twice in a row... :upside_down_face: 
Actually it's not mentioned anywhere in the owncloud documentation, but webdav has a COPY method, so if owncloud accepts it we have a solution :-) 
http://webdav.org/specs/rfc4918.html#METHOD_COPY

ci/upload_to_owncloud.sh
rmelotte: Right, I broke it when rebasing.
The thing is that the URL is not going to be the same for the different folders (the artifacts folder is not part of the prplmesh folder).
What I propose is to just drop this URL, and add links to the relevant directories in the README.md.
What do you think?

tomereli: I really prefer not to drop the URL since it is printed to stdout as part of the cron job and the idea is that when we get the mail we can simply click the link to get to the results.
So what I propose is to add this print to  run_easymesh_cert.sh - I can do that in my followup PR if this is agreed by you and @arnout .

.gitlab-ci.yml
rmelotte: It's not a bug, the commit message correspond to Gitlab CI's behavior.
I wanted the upload to only be done if all the previous job for the pipeline succeeded (so that the `latest` folder on owncloud always corresponds to a build passing all the Gitlab CI tests). That's how downloading the "latest" artifacts using Gitlab API is supposed to work (from the documentation).

Since builds on the master branch are supposed to always be passing anyway, it's arguably not /that/ useful.

arnout: OK, then maybe "because we don't want the job to start if any test from the previous stage failed". The "unrelated" was confusing.

.gitlab-ci.yml
ghost: For webdav, a symlink looks roughly like this:
latest.html:
```
<html>
<head>
<meta http-equiv="refresh" content="2;url=http://example.com/ACTUAL_LATEST_DIRECTORY" />
<title>Latest version of test results</title>
</head>
<body>
Click <a href="http://example.com/ACTUAL_LATEST_DIRECTORY">here</a> to go to the latest results, if your web browser does not redirect you automatically.
</body>
</html>
```

Actually, using a `.htaccess` file would be even better, but I don't know if that's an option with owncloud

rmelotte:
Superseded by https://github.com/prplfoundation/prplMesh/pull/1059 .

PR #1052, opened 2020-03-31
run_easymesh_cert: download from owncloud by default and some more features and fixes

tomereli:
Enhancements to download_ipk.sh and run_easymesh_cert.sh:
- Add download from owncloud support (default)
- Add option to skip prplmesh upgrade and make it optional
- Add option to run certification for a specific device

Needs to be merged after #1050

### DIFF
--- a/ci/owncloud/owncloud_definitions.sh
+++ b/ci/owncloud/owncloud_definitions.sh
@@ -108,6 +108,40 @@ browse_url() {
     echo "$browse_url"
 }
 
+download_url() {
+    # Calculate and echo the download url for a relative path
+    #
+    # $1 base url name - "artifacts" or "certification"
+    # $2 relative path from the base_url
+    #
+    # Uses:
+    #   OWNCLOUD_PUBLIC_URLS
+    local relative_path filename base_url browse_url
+    base_url="${OWNCLOUD_PUBLIC_URLS[$1]}"
+    relative_path="${2%/*}"
+    filename="${2##*/}"
+
+    download_url="$base_url/download?path="
+    for dir in $(echo "$relative_path" | tr "/" "\\n"); do
+        download_url="$download_url%2F$dir"
+    done
+
+    echo "$download_url&files=$filename"
+}
+
+download_artifact() {
+    # Downloads a file from the artifacts public folder
+    #
+    # $1 relative path from the artifacts folder
+    # $2 local path including filename
+    local base_url download_path
+
+    download_url="$(download_url artifacts "$1")"
+    download_path=${2-$(basename "$1")}
+    # curl options - fail on error (-f), follow redirect (-L)
+    curl ${QUIET:+ -s -S} -f -L "$download_url" --output "$download_path"
+}
+
 OWNCLOUD_URL="https://ftp.essensium.com/owncloud/remote.php/dav/files"
 
 declare -A OWNCLOUD_PUBLIC_URLS
--- a/ci/owncloud/owncloud_definitions.sh
+++ b/ci/owncloud/owncloud_definitions.sh
@@ -1,11 +1,7 @@
 #!/bin/bash
 
-scriptdir="$(cd "${0%/*}" && pwd)"
-rootdir=$(realpath "$scriptdir/../../")
-
-# shellcheck source=../tools/functions.sh
-. "$rootdir/tools/functions.sh"
-
+# shellcheck source=../../tools/functions.sh
+. "$(dirname "${BASH_SOURCE[0]}")/../../tools/functions.sh"
 
 create_dir() {
     # create dir $1 in $OWNCLOUD_URL/$user/.
--- a/ci/owncloud/owncloud_definitions.sh
+++ b/ci/owncloud/owncloud_definitions.sh
@@ -92,7 +92,32 @@ copy() {
     return "$status"
 }
 
+browse_url() {
+    # Calculate and echo the browse url for a relative path
+    #
+    # $1 base url name - "artifacts" or "certification"
+    # $2 relative path from the base_url
+    #
+    # Uses:
+    #   OWNCLOUD_PUBLIC_URLS
+    local relative_path base_url browse_url
+    base_url="${OWNCLOUD_PUBLIC_URLS[$1]}"
+    relative_path="$2"
+
+    browse_url="$base_url?path="
+    for dir in $(echo "$relative_path" | tr "/" "\\n"); do
+        browse_url="$browse_url%2F$dir"
+    done
+
+    echo "$browse_url"
+}
+
 OWNCLOUD_URL="https://ftp.essensium.com/owncloud/remote.php/dav/files"
+
+declare -A OWNCLOUD_PUBLIC_URLS
+OWNCLOUD_PUBLIC_URLS["certification"]="https://ftp.essensium.com/owncloud/index.php/s/ketl3eFhIgwcweZ"
+OWNCLOUD_PUBLIC_URLS["artifacts"]="https://ftp.essensium.com/owncloud/index.php/s/pzpSry7KOskL4Wi"
+
 user=$(awk '/ftp.essensium.com/{getline; print $4}' ~/.netrc)
 
 if [ ! "$#" = 0 ] ; then
--- a/ci/owncloud/owncloud_definitions.sh
+++ b/ci/owncloud/owncloud_definitions.sh
@@ -150,6 +150,7 @@ OWNCLOUD_PUBLIC_URLS["artifacts"]="https://ftp.essensium.com/owncloud/index.php/
 
 user=$(awk '/ftp.essensium.com/{getline; print $4}' ~/.netrc)
 
-if [ ! "$#" = 0 ] ; then
+# Check if script is being sourced
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     "$@"
 fi
--- a/ci/owncloud/upload_to_owncloud.sh
+++ b/ci/owncloud/upload_to_owncloud.sh
@@ -70,7 +70,7 @@ upload() {
 
 main() {
     if ! OPTS=$(getopt -o 'hvu:' --long help,verbose,url: -n 'parse-options' -- "$@"); then
-        echo echo "Failed parsing options." >&2; usage; exit 1
+        echo "Failed parsing options." >&2; usage; exit 1
     fi
 
     eval set -- "$OPTS"
--- a/ci/owncloud/upload_to_owncloud.sh
+++ b/ci/owncloud/upload_to_owncloud.sh
@@ -33,6 +33,40 @@ usage() {
     echo ""
 }
 
+upload() {
+    # Recursively upload directory from local path to remote path
+    #
+    # $1 - local path to directory
+    # $2 - remote path to upload to
+
+    local local_path remote_path
+    local_path="$1"
+    remote_path="$2"
+
+    create_dir "$remote_path"
+    find "$local_path" -type d -exec \
+            realpath {} --relative-to="$(dirname "$local_path")" \; | {
+                error=0
+                while read -r -s dir; do
+                    printf . # show progress
+                    create_dir "$remote_path/$dir" || {
+                        error="$?"
+                        continue
+                    }
+                    printf . # show progress
+                    # get the list of files to upload in the format <file>,<file>,...,<file>
+                    files=$(find "$(dirname "$local_path")/$dir/" -type f -maxdepth 1 -print0 | tr '\0' ',' | sed 's/,$//')
+                    dbg "$files"
+                    [ -n "$files" ] && {
+                        upload_files "$remote_path/$dir" "$files" || {
+                            error="$?"
+                        }
+                    }
+                done
+                echo
+                return $error
+            }
+}
 
 main() {
     if ! OPTS=$(getopt -o 'hvu:' --long help,verbose,url: -n 'parse-options' -- "$@"); then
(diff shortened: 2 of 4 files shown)


ci/owncloud/owncloud_definitions.sh
```diff
@@ -117,6 +113,8 @@ OWNCLOUD_PUBLIC_URLS["artifacts"]="https://ftp.essensium.com/owncloud/index.php/
 
 user=$(awk '/ftp.essensium.com/{getline; print $4}' ~/.netrc)
 
-if [ ! "$#" = 0 ] ; then
-    "$@"
-fi
+[[ "$_" != "$0" ]] || {
+    if [ ! "$#" = 0 ] ; then
```
rmelotte: >In addition, do run the $@ commands only if the script is not sourced,
since this produced another warning when sourced by
run_easymesh_cert.sh.

This is what this specific line was supposed to do, so if it doesn't work, the new condition should replace this one.
What doesn't work/what warning do you get without the `[[ "$_" != "$0" ]] || {` part?

A downside of this last condition is that you can no longer call the script with anything in front of the script invocation (try `bash ci/owncloud/owncloud_definitions.sh copy a b` or even `HELLO=WORLD ./ci/owncloud/owncloud_definitions.sh copy a b`.

tools/download_ipk.sh
```diff
@@ -1,34 +1,37 @@
...
 # Copyright (c) 2020 Tomer Eliyahu (Intel)
 # This code is subject to the terms of the BSD+Patent license.
 # See LICENSE file for more details.
 ###############################################################
+scriptdir="$(cd "${0%/*}" && pwd)"
+rootdir=$(realpath "$scriptdir/../")
+
+# shellcheck source=functions.sh
+. "$rootdir/tools/functions.sh"
+# shellcheck source=ci/owncloud/owncloud_definitions.sh
```
rmelotte: The path looks wrong.

tools/download_ipk.sh
```diff
@@ -1,34 +1,37 @@
...
+. "$rootdir/ci/owncloud/owncloud_definitions.sh"
 
 usage() {
-    echo "usage: $(basename "$0") [-hbd]"
+    echo "usage: $(basename "$0") [-hbdrp]"
```
rmelotte: Only `h` and `v` are defined as short options.

ci/run_easymesh_cert.sh
```diff
@@ -12,8 +12,10 @@
 scriptdir="$(cd "${0%/*}" && pwd)"
 rootdir=$(realpath "$scriptdir/../")
 
-# shellcheck source=functions.sh
+# shellcheck source=../tools/functions.sh
 . "$rootdir/tools/functions.sh"
+# shellcheck source=ci/owncloud/owncloud_definitions.sh
```
rmelotte, line 17: The path is wrong, it has to be relative to the script directory.

rmelotte (approved):
Only small comments (except maybe for the `"$_" != "$0"` condition) so approving

ci/owncloud/owncloud_definitions.sh
tomereli: Actually I tested it again and apparently we don't need the extra `[[ "$_" != "$0" ]]`, so removing.

ci/run_easymesh_cert.sh
```diff
@@ -20,12 +20,12 @@ usage() {
     echo "  options:"
     echo "      -h|--help - display this help."
     echo "      -v|--verbose - set verbosity (ucc logs also redirected to stdout)"
-    echo "      -b|--branch - prplmesh branch to use (default master latest)"
     echo "      -o|--log-folder - path to put the logs to (make sure it does not contain previous logs)."
     echo "      -e|--easymesh-cert - path to easymesh_cert repository (default ../easymesh_cert)"
     echo "      -s|--ssh - target device ssh name (defined in ~/.ssh/config). (default: $TARGET_DEVICE_SSH)"
     echo "      --owncloud-upload - whether or not to upload results to owncloud. (default: false)"
     echo "      --owncloud-path - relative path in the owncloud server to upload results to. (default: $OWNCLOUD_PATH)"
+    echo "      --skip-upgrade - skip prplmesh upgrade for netgear-rax40 device (default: false)"
```
arnout:
```suggestion
    echo "      --skip-upgrade - skip prplmesh upgrade (default: false)"
```

ci/run_easymesh_cert.sh
```diff
@@ -70,28 +87,19 @@ main() {
...
+
     info "Logs location: $LOG_FOLDER"
     info "Tests to run: $TESTS"
+    info "Device: $DEVICE"
 
-    info "download latest ipk from branch $BRANCH"
-    "$TOOLS_PATH"/download_ipk.sh --branch "$BRANCH" ${VERBOSE:+ -v} || {
-        err "Failed to download prplmesh.ipk, abort"
-        exit 1
+    [ "$DEVICE" = "netgear-rax40" ] && [ "$UPGRADE_PRPLMESH" = "true" ] && {
```
arnout: I don't like hardcoding netgear-rax40 here... Which device do you want to exclude here? If it is Ubuntu, the exclude that one specifically. If it is UGW/RDKB, hm, I don't know...

ci/run_easymesh_cert.sh
```diff
@@ -90,9 +105,9 @@ main() {
 
     info "Logs location: $LOG_FOLDER"
     info "Tests to run: $TESTS"
-    info "Device: $DEVICE"
+    info "Device: $TARGET_DEVICE"
 
-    [ "$DEVICE" = "netgear-rax40" ] && [ "$UPGRADE_PRPLMESH" = "true" ] && {
+    [ "$UPGRADE_PRPLMESH" = "true" ] && {
```
arnout, line 113: Ah, well, never mind then - but then you shouldn't have added that condition in the previous commit :-)

Don't bother changing it.

ci/owncloud/owncloud_definitions.sh
```diff
@@ -1,11 +1,7 @@
 #!/bin/bash
 
-scriptdir="$(cd "${0%/*}" && pwd)"
-rootdir=$(realpath "$scriptdir/../../")
-
-# shellcheck source=../tools/functions.sh
-. "$rootdir/tools/functions.sh"
-
+# shellcheck source=../../tools/functions.sh
+. "$(dirname "${BASH_SOURCE[0]}")/../../tools/functions.sh"
```
arnout, line 4: Excellent!

ci/owncloud/owncloud_definitions.sh
```diff
@@ -105,6 +105,33 @@ browse_url() {
...
 }
 
+download_url() {
+    # Calculate and echo the download url for a file
+    #
+    # $1 relative path to the file from the base_url
+    # $2 base_url
+    local download_url
+
+    download_url="$(browse_url "$(dirname "$1")" "$2"/download)&files=$(basename "$1")"
```
arnout: I believe this should be
```suggestion
    download_url="$(browse_url "${1%%/*}" "$2"/download)&files=${1#*/}"
```
i.e. break off the first component of the path only, not the last component.

ci/owncloud/owncloud_definitions.sh
tomereli: So apparently this is a buggy way to do this, since it doesn't work when sourcing the script from a different script, I changed to use ${BASH_SOURCE[0]} = $0 condition to check if the script is being sourced.

ci/owncloud/owncloud_definitions.sh
```diff
@@ -92,7 +92,29 @@ copy() {
     return "$status"
 }
 
+browse_url() {
+    # Calculate and echo the browse url for a relative path
+    #
+    # $1 relative path from the base_url
+    # $2 base_url
```
arnout: (nitpick) rather than base_url, IMO it would make more sense to pass in "certification" or "artifacts" and dereference OWNCLOUD_PUBLIC_URLS from within this function. In that case, I'd also pass it as the first argument, and the relative path second. It looks nicer that way IMO.

ci/owncloud/owncloud_definitions.sh
```diff
@@ -105,6 +105,33 @@ browse_url() {
...
+    # Downloads a file from the artifacts public folder
+    #
+    # $1 relative path from the artifacts folder
+    # $2 local path including filename
+    # Uses:
+    #   OWNCLOUD_PUBLIC_URLS["artifacts"]
+    local base_url download_path
+
+    base_url="${OWNCLOUD_PUBLIC_URLS["artifacts"]}"
+    download_url="$(download_url "$1" "$base_url")"
```
arnout: So, with my earlier suggestion, this would become
```suggestion
    local download_path

    download_url="$(download_url "artifacts" "$1")"
```

tools/download_ipk.sh
```diff
@@ -1,34 +1,37 @@
...
 
 usage() {
-    echo "usage: $(basename "$0") [-hbd]"
+    echo "usage: $(basename "$0") [-hv]"
     echo "  options:"
     echo "      -h|--help - show this help menu"
     echo "      -v|--verbose - add verbosity"
     echo "      --branch  - branch to use (default master)"
     echo "      --device  - device to use (default netgear-rax40)"
+    echo "      --remote  - remote to use - owncloud or gitlab (default owncloud)"
```
arnout, line 23: Instead of specifying remote, I'd simply infer it from the branch name. If branch is master, use owncloud, else use gitlab.

Note BTW that AFAIU, the artifacts are *always* uploaded to owncloud, even for branches. It's just that the "latest" only refers to master. If we need a different one, we'd need to find in which directory the latest for that branch was uploaded.

ci/owncloud/owncloud_definitions.sh
```diff
@@ -140,6 +140,7 @@ OWNCLOUD_PUBLIC_URLS["artifacts"]="https://ftp.essensium.com/owncloud/index.php/
 
 user=$(awk '/ftp.essensium.com/{getline; print $4}' ~/.netrc)
 
-if [ ! "$#" = 0 ] ; then
+# Check if script is being sourced
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
```
arnout, line 154: Oh yes, this works as well!

ci/owncloud/upload_to_owncloud.sh
```diff
@@ -28,8 +28,24 @@ usage() {
     echo ""
 }
 
+create_dirs() {
+    # Recursively create directory path in remote
+    # $1 remote path to create
+    local path=
+    for dir in $(echo "$1" | tr "/" "\\n"); do
```
arnout: Doesn't shellcheck complain about this?

I think a more typical pattern would be
```suggestion
    echo "$1" | tr / '\n' | while read -r -s dir; do
```

arnout (approved):
These scripts are becoming waaay to complicated for shell :-)

TIL: when you write a script, do it in Python right away, even if you think it will be a simple script.

arnout:
I'm not sure why run-tests failed... It's not one of the randomly occuring issues...

ci/run_easymesh_cert.sh
rmelotte: You marked it as resolves but Github still shows it as `ci/owncloud/owncloud_definitions.sh` instead of `owncloud/owncloud_definitions.sh`.
Maybe you didn't push it yet?

ci/owncloud/upload_to_owncloud.sh
```diff
@@ -28,8 +28,24 @@ usage() {
...
+        if [ -z "$path" ]; then
+            create_dir "$dir"
+            path="$dir"
+        else
+            create_dir "$path/$dir"
+            path="$path/$dir"
+        fi
+    done
```
rmelotte: You can also loop over $1 directly if you set IFS to '/':
```suggestion
    local path IFS
    path=
    IFS='/'
    for dir in $1; do
        if [ -z "$path" ]; then
            create_dir "$dir"
            path="$dir"
        else
            create_dir "$path/$dir"
            path="$path/$dir"
        fi
    done
    unset IFS
```
(`unset IFS` is only there "just in case" the shell do not implement `local` i.e. it's not needed for bash)

ci/owncloud/upload_to_owncloud.sh
```diff
@@ -28,8 +28,24 @@ usage() {
     echo ""
 }
 
+create_dirs() {
+    # Recursively create directory path in remote
```
rmelotte, line 33: Nice!

ci/owncloud/upload_to_owncloud.sh
```diff
@@ -17,6 +17,7 @@ usage() {
     echo "      -h|--help - display this help."
     echo "      -v|--verbose - enable verbosity"
     echo "      -u|--url - webDAV URL (default https://ftp.essensium.com/owncloud/remote.php/dav/files/USERNAME)"
+    echo "      --direct - direct upload without first uploading to a temporary directory (default false)"
```
rmelotte, line 20: I still don't really understand why you'd prefer to upload directly there, but ok.

Not ok though: the help message says the default is false, but you set it to true.

ci/run_easymesh_cert.sh
```diff
@@ -121,16 +121,16 @@ main() {
...
     if [ -n "$OWNCLOUD_UPLOAD" ]; then
         if is_prplmesh_device "$TARGET_DEVICE"; then
-            REMOTE_PATH="$TARGET_DEVICE/$(basename "$LOG_FOLDER")"
+            REMOTE_PATH="$TARGET_DEVICE"
         else
-            REMOTE_PATH="certified/$TARGET_DEVICE/$(basename "$LOG_FOLDER")"
+            REMOTE_PATH="certified/$TARGET_DEVICE"
         fi
         info "Uploading $LOG_FOLDER to $OWNCLOUD_PATH/$REMOTE_PATH"
-        "$scriptdir"/owncloud/upload_to_owncloud.sh "$OWNCLOUD_PATH/$REMOTE_PATH" "$LOG_FOLDER" || {
```
rmelotte: Ok here you explain why you want the "direct" option.
Note that even if you do partial uploads to a temporary directory, you can still move the incomplete folder at the end if you want.

IMO, if the default behavior doesn't correspond to what you expect (i.e. it uploads by replacing the remote path you specified, instead of uploading /inside/ the remote path) we should change it instead of adding the `direct` option as a workaround.

To me you can merge it as it is now though.

rmelotte (approved):
Nothing blocking :+1:

ci/owncloud/upload_to_owncloud.sh
tomereli: It is mainly to avoid the dir/dir issue.. and it is confusing and unnecessary to upload to a temp location when uploading logs - if the upload fails at some point, the logs we did succeed to upload are still interesting, so direct upload IMO is the correct approach in this case.

ci/run_easymesh_cert.sh
tomereli: I didn't want to change it so not to cause issues with the CI uploads. I think doing it in steps, and later removing the direct option is better.

tools/download_ipk.sh
tomereli: Still I'd like the flexibility of downloading from gitlab, so I'd prefer to still have a `--remote` option.

ci/owncloud/upload_to_owncloud.sh
tomereli: IMO `for` loop is easier to read, why is while preferred?

ci/owncloud/upload_to_owncloud.sh
tomereli: Done

ci/owncloud/upload_to_owncloud.sh
tomereli: Actually @rmelotte 's suggestion to use `IFS` is the best IMO, so changed to that one.

ci/owncloud/upload_to_owncloud.sh
tomereli: The default is empty, which means it is not true. I will add a default so it is clearer.

PR #1055, opened 2020-03-31
Feature/ci prevent new cppcheck issues

rmelotte:
With the huge amount of issues cppcheck reports already it will take a lot of time to solve them all, and it could be useful to still prevent new issues like we do for PEP8.

Since we've never been block by "less severe" cppcheck issues before, this means we will potentially have to tweak cppcheck from time to time for issues we think are not relevant, and PR might get block by it in the meantime.


rmelotte:
~~We could also make the job "allowed to fail" for some time, until we are confident it only reports relevant issues.~~
As @arnout mentioned during the sync meeting, making the job "allowed to fail" will encourage people to **not** check its status, which means it won't be taken into account in the end.

tomereli (approved):

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# shellcheck source=../../tools/function.sh
+. "${rootdir}/tools/functions.sh"
+
+
+if [ ! -f "$rootdir/cppcheck_results.txt" ] ; then
+    err "run $rootdir/tools/docker/static-analysis/cppcheck.sh first"
+fi
+
+mapfile -t issues < <(grep -E ".+:[0-9]+:[0-9]+: .+: .+ \[.+\]$"  "$rootdir/cppcheck_results.txt")
```
arnout: TIL: the mapfile command. Interesting!

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -0,0 +1,26 @@
...
+
+scriptdir="$(cd "${0%/*}" && pwd)"
+rootdir="${scriptdir%/*/*}"
+
+# shellcheck source=../../tools/function.sh
+. "${rootdir}/tools/functions.sh"
+
+
+if [ ! -f "$rootdir/cppcheck_results.txt" ] ; then
+    err "run $rootdir/tools/docker/static-analysis/cppcheck.sh first"
```
arnout: Missing exit after printing the error.

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -0,0 +1,26 @@
...
+fi
+
+mapfile -t issues < <(grep -E ".+:[0-9]+:[0-9]+: .+: .+ \[.+\]$"  "$rootdir/cppcheck_results.txt")
+
+status=0
+for i in "${issues[@]}"
```
arnout, line 16: (nitpick) As happy as I am with learning the mapfile command, here it really doesn't bring anything over the traditional (and POSIX)

```
grep -E ... | \
    while read -r i; do
```

arnout (approved):
I will approve this PR, but I'm afraid we're going to revert it pretty quickly...

The problem is that the issue matching happens based on the line numbers. For example, if we add an include to db.cpp, all the errors that it currently gives will shift by one line. So we need to add an 800-line diff to cppcheck_existing_issues.txt to handle that... The result will be that people blindly overwrite cppcheck_existing_issues.txt so you still don't see a new issue.

To work around it, we could use a different output format where the code is on the same line as the rest, and where we don't have line and column numbers.

For now though, I'd merge as is. We can still fix it if it turns out to be an actual problem.

rmelotte:
Indeed the line number is a pretty annoying issue, I'm glad you mentioned it.
In the end it was easy to change the format of the reported issues since we're doing this in a separate job anyway.
So I did exactly what you suggested: removed line and column numbers and added the code itself.

PR #1057, opened 2020-03-31
Bugfix/netgear rax40 firmware

rmelotte:
As reported [here](https://git.prpl.dev/prplwrt/feed-intel/issues/2), building by selecting only the target and prplMesh doesn't produce a working image, so for the rax40, use the whole profile file instead.


tomereli (approved):

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -81,7 +81,10 @@ RUN echo "src-git prpl ${PRPL_FEED}" >> feeds.conf.default \
     && echo "CONFIG_TARGET_${TARGET_SYSTEM}_${SUBTARGET}=y" >> .config \
     && echo "CONFIG_TARGET_${TARGET_SYSTEM}_${SUBTARGET}_${TARGET_PROFILE}=y" >> .config \
     && echo "CONFIG_PACKAGE_prplmesh${PRPLMESH_VARIANT}=y" >> .config \
-    && make defconfig \
+    && if [ -n "$INTEL_FEED" ] ; then \
```
arnout: I think it would make a whole lot more sense to use
```suggestion
    && if [ "$TARGET_PROFILE" = "DEVICE_NETGEAR_RAX40" ] ; then \
```

tools/docker/builder/openwrt/build.sh
```diff
@@ -111,7 +111,7 @@ main() {
             TARGET_PROFILE=DEVICE_NETGEAR_RAX40
             PRPLMESH_VARIANT="-dwpal"
             PRPL_FEED="https://git.prpl.dev/prplmesh/iwlwav.git^6265989f2c47f7b42f64d25f455443b87902582e"
-            INTEL_FEED="https://git.prpl.dev/prplmesh/feed-intel.git^e3eca4e93286eb4346f0196b2816a3be97287482"
+            INTEL_FEED="https://git.prpl.dev/prplmesh/feed-intel.git^6323abee35f2d8cd6263656e9226176cf1bcfca7"
```
arnout: Seems to me that this should be done in the same commit that updates the prpl feed (nitpick).

arnout (approved):

tools/docker/builder/openwrt/Dockerfile
rmelotte: Indeed now that we have it it makes a lot more sense to use it.
Thanks :+1:

PR #1059, opened 2020-03-31
Feature/ci upload artifacts to owncloud 3

rmelotte:
As mentioned during the sync meeting, I changed so many things from https://github.com/prplfoundation/prplMesh/pull/1050 that I decided to create a whole new branch and PR.

I think I addressed all review comments, and made it so that artifacts are uploaded to owncloud for every branch (first in a temporary folder, then moving them into their final location).
When on the master branch, an additional copy is made to the "latest" folder.

The artifacts URL has also changed, as the directory is now back to `prplmesh/artifacts` (I figured out how to add it to the robot account home).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,3 +1,8 @@
+stages:
+  - build
+  - test
+  - upload
+
 .in-prplmesh-builder:
   image:
     name: prplfoundationinc/prplmesh-builder:ubuntu-18.04
@@ -82,6 +87,31 @@ build-in-docker-static:
   extends: build-in-docker
   variables:
     EXTRA_CMAKE_FLAGS: "-DBUILD_SHARED_LIBS=OFF"
+
+upload-artifacts:
+  stage: upload
+  extends: .in-prplmesh-builder
+  script:
+    # the owncloud password is stored in the file pointed by "$CI_OWNCLOUD_PASSWORD".
+    # We can only store the password in it, since spaces are not supported in Gitlab masked variables.
+    # We symlink the file to .netrc and pre-pend to it to avoid manipulating the password itself
+    # on the command line.
+    - ln -s "$CI_OWNCLOUD_PASSWORD" ~/.netrc
+    - sed --follow-symlinks -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
+    - ci/owncloud/upload_to_owncloud.sh -v "artifacts/$CI_COMMIT_SHA-$CI_JOB_ID" ./build
+    - |
+      if [ $CI_COMMIT_BRANCH = "master" ] ; then
+          echo "Updating the 'latest' folder"
+          # Copying takes a lot of time, and we want the "latest" folder to be updated
+          # "as fast as possible", so we make a copy and move it:
+          ci/owncloud/owncloud_definitions.sh copy "artifacts/$CI_COMMIT_SHA-$CI_JOB_ID" "temp_uploads/latest_$CI_COMMIT_SHA-$CI_JOB_ID"
+          ci/owncloud/owncloud_definitions.sh move "temp_uploads/latest_$CI_COMMIT_SHA-$CI_JOB_ID" "artifacts/latest"
+      fi
+  needs:
+    - build-in-docker
+    - build-for-netgear-rax40
+    - build-for-glinet-b1300
+    - build-for-turris-omnia
 
 .run-test-in-docker:
   stage: test
--- a/ci/upload_to_owncloud.sh
+++ b/ci/upload_to_owncloud.sh
@@ -75,6 +75,26 @@ upload_files() {
     return "$status"
 }
 
+move() {
+    # move $1 to $2 on remote.
+    # if $2 already exists, it's completely overwritten.
+    #
+    # uses:
+    #   OWNCLOUD_URL
+    #   user
+    local src dst status
+    src="$1"
+    dst="$2"
+    status=0
+    dbg "Moving \"$src\" to \"$dst\" on the remote"
+    curl ${QUIET:+-s -S} -f -n -X MOVE --header \
+         "Destination: $OWNCLOUD_URL/$user/$dst" "$OWNCLOUD_URL/$user/$src" || {
+        status="$?"
+        err "Failed to move $src to $dst (error $status)"
+    }
+    return "$status"
+}
+
 main() {
     if ! OPTS=$(getopt -o 'hvu:' --long help,verbose,url: -n 'parse-options' -- "$@"); then
         echo echo "Failed parsing options." >&2; usage; exit 1
--- a/ci/upload_to_owncloud.sh
+++ b/ci/upload_to_owncloud.sh
@@ -137,12 +137,29 @@ main() {
     local_path="$1"; shift
 
     info "upload $local_path to $OWNCLOUD_BROWSE_URL/$remote_path/$(basename "$local_path") (using user $user)"
+
+    if ! command -v uuidgen > /dev/null ; then
+        err "You need uuidgen to use this script. Please install it and try again."
+        exit 1
+    fi
+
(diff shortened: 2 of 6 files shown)


ci/upload_to_owncloud.sh
```diff
@@ -53,11 +53,16 @@ main() {
...
-            realpath {} --relative-to="$(dirname "$local_path")" \; | \
+            realpath {} --relative-to="$(dirname "$local_path")" \; | {
+                error=0
                 while read -r -s dir; do
                     printf . # show progress
                     dbg "Create directory: $dir"
-                    curl ${QUIET:+ -s -S} -f -n -X MKCOL "$OWNCLOUD_URL/$user/$remote_path/$dir"
+                    curl ${QUIET:+ -s -S} -f -n -X MKCOL "$OWNCLOUD_URL/$user/$remote_path/$dir" || {
+                        error="$?"
+                        err "Failed to create dir: $remote_path/$dir/ (error $error) skipping."
```
tomereli: It depends on the error - but if the directory exists we might still want to upload the files, don't we? In this case, perhaps we should not `continue`.

tools/docker/builder/Dockerfile
```diff
@@ -32,6 +32,7 @@ RUN apt-get update && apt-get install -y \
     python-yaml \
     python3 \
     python3-yaml \
+    uuid-runtime \
```
tomereli, line 35: I assume this should not only be used by docker - so please add it to the main readme as well

ci/upload_to_owncloud.sh
```diff
@@ -137,12 +137,29 @@ main() {
     local_path="$1"; shift
 
     info "upload $local_path to $OWNCLOUD_BROWSE_URL/$remote_path/$(basename "$local_path") (using user $user)"
+
+    if ! command -v uuidgen > /dev/null ; then
```
tomereli, line 61: So now we always upload to a temporary location before moving it to the final location? Isn't it wastefull? Uploading a full logs diretocy of 20+ tests takes A LOT of time as it is, I think this feature should be optional unless you tested it with a big folder with lots of files and found out that the move part is significantly faster than the upload part.

tomereli (approved):
One major comment about the potentially very long time to upload log folders - so please test that the move operation is negligible or make this optional.

ci/upload_to_owncloud.sh
rmelotte: If the directory already exists it means something bad happened, since we're always uploading to a temporary folder that is supposed to be empty.

tomereli: I would like to revisit this - when used by the `run_easymesh_cert` it is practically guaranteed that the directory will not exist so the whole temporary folder solution is not needed in this case which is why I suggest to make it optional.

ci/upload_to_owncloud.sh
rmelotte: Yes we always upload to a temporary location first, then do a move.

If the destination folder doesn't exist the move is instant since it's just a rename, so there shouldn't be any significant overhead.

When the folder already exists however (and it does for the `latest` folder), it takes much more time.
If we didn't upload to a temporary folder first, we would anyway have to remove the destination folder first before uploading, so I'm not convinced it would take less time.


To summarize: 
I think the move is annoying for the `latest` folder, because I assumed it would almost instantly replace the folder (we want it to contain valid artifacts at any time) and it doesn't, but it has nothing to do with the fact that we use temporary folders.
To improve this we could try [Frederik's suggestion](https://github.com/prplfoundation/prplMesh/pull/1050#discussion_r400861726) and see if we can use some kind of link instead of a plain folder.

arnout: I like Frederik's suggestion. So with that, we could say that we *require* the target folder not to exist, right?

That said, I think we should merge it as is now, even if the copy takes some time, and fix later. For Tomer's use case the temporary location shouldn't hurt, and this way we at least have something working for the "latest".

ci/upload_to_owncloud.sh
```diff
@@ -29,6 +29,52 @@ usage() {
...
+    #   OWNCLOUD_URL
+    #   user
+    local dir delete status
+    dir="$1"
+    delete="$2"
+    status=0
+
+    dbg "Create directory: $dir"
+    if [ "$delete" = true ] ; then
+        curl ${QUIET:+-s -S} -f -n -X DELETE "$OWNCLOUD_URL/$user/$dir" > /dev/null 2>&1
```
arnout: Given the number of times `curl ${QUIET:+-s -S} -f -n -X ` is repeated, it might make sense to make a wrapper for that as well (nitpick).

tools/docker/builder/Dockerfile
arnout: Since the script will tell you to install uuidgen if you don't have it, I don't think it's needed to add it to the main readme. Anyway, nobody outside of the maintainer team is going to use that script.

.gitlab-ci.yml
```diff
@@ -83,6 +88,30 @@ build-in-docker-static:
   variables:
     EXTRA_CMAKE_FLAGS: "-DBUILD_SHARED_LIBS=OFF"
 
+upload-artifacts:
+  stage: upload
+  extends: .in-prplmesh-builder
+  script:
+    # the owncloud password is stored in the file pointed by "$CI_OWNCLOUD_PASSWORD".
+    # We can only store the password in it, since spaces are not supported in Gitlab masked variables.
+    # We symlink the file to .netrc and pre-pend to it to avoid manipulating the password itself
```
arnout, line 97: I still don't understand why we symlink instead of copy the file...

.gitlab-ci.yml
```diff
@@ -83,6 +88,30 @@ build-in-docker-static:
...
+    - ln -s "$CI_OWNCLOUD_PASSWORD" ~/.netrc
+    - sed --follow-symlinks -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
+    - ci/owncloud/upload_to_owncloud.sh -v "artifacts/$CI_COMMIT_SHA-$CI_JOB_ID" ./build
+    - |
+      if [ $CI_COMMIT_BRANCH = "master" ] ; then
+          echo "Updating the 'latest' folder"
+          # Copying takes a lot of time, and we want the "latest" folder to be updated
+          # "as fast as possible", so we make a copy and move it:
+          ci/owncloud/owncloud_definitions.sh copy "artifacts/$CI_COMMIT_SHA-$CI_JOB_ID" "temp_uploads/latest_$CI_COMMIT_SHA-$CI_JOB_ID"
+          ci/owncloud/owncloud_definitions.sh move "temp_uploads/latest_$CI_COMMIT_SHA-$CI_JOB_ID" "artifacts/latest"
```
arnout, line 108: Nice!

arnout (approved):
Let's get this merged quickly.

ci/upload_to_owncloud.sh
rmelotte: With a link, we would be able to make it point to something that is valid at any point in time, so it won't matter much anymore if destination folder already exist or not.

I agree it should be good enough for now.

.gitlab-ci.yml
rmelotte: There is no particular reason for a symlink over a copy. 
If it feels more "natural", I can make a copy instead.
Is there a particular advantage to making a copy?

ci/upload_to_owncloud.sh
rmelotte: @tomereli Like I said in another comment if the target directory does not exist, using a temporary directory shouldn't make a noticeable difference.
Did you test it on your side?
Do you still want to revisit it?
If not I'll add the "ready for merge" tag.

tomereli: Go ahead please, if we will need to, we will revisit this in the future.

PR #1060, opened 2020-03-31
[Task] Compute MediaType for wireless ifaces in Device Information TLV

### LINKED ISSUE #700: [TASK] Add Device Information TLV to Topology Response message
## Description

Add Device Information TLV to Topology Response message.
- [x] Include information about wired local interfaces (#705)
- [x] Include information about wireless local interfaces (#1060). 

## Exit criteria

Topology Response message contains Device Information TLV with real data. 

## Detailed description

For a detailed description, see task #525 (this task has been divided into smaller tasks to be managed more easily).

mariomaz:
Compute MediaType for wireless ifaces in Device Information TLV of Topology Response message.


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -133,8 +135,55 @@ static bool get_media_type(const std::string &interface_name,
...
+        bwl::nl80211_client::radio_info radio_info;
+
+        if (client->get_radio_info(interface_name, radio_info)) {
+            if (!radio_info.bands.empty()) {
+                bwl::nl80211_client::band_info band_info = radio_info.bands.at(0);
+
+                beerocks::eFreqType frequency_band     = band_info.get_frequency_band();
+                beerocks::eWiFiBandwidth max_bandwidth = band_info.get_max_bandwidth();
+
+                if ((beerocks::eFreqType::FREQ_24G == frequency_band) &&
```
rmelotte: This block is really hard to read.
Maybe you could first group by frequency, then by bandwidth.
You could also maybe use a switch for the bandwidth
(nitpick)

rmelotte (approved):
:+1:

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/eMediaType.h
```diff
@@ -30,6 +30,7 @@ enum eMediaType: uint16_t {
     IEEE_802_11AC_5_GHZ = 0x105,
     IEEE_802_11AD_60_GHZ = 0x106,
     IEEE_802_11AF = 0x107,
+    IEEE_802_11AX = 0x108,
```
tomereli, line 33: This should be squashed to the previous commit. Generally, when the change is small, we prefer to update the autogenerated files in the same commit as the one which changes the yaml.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -133,8 +135,55 @@ static bool get_media_type(const std::string &interface_name,
...
+         * 3 IEEE 802.11n (2.4 GHz)      2.4 GHz    40 MHz
+         * 4 IEEE 802.11n (5 GHz)        5 GHz      40 MHz
+         * 5 IEEE 802.11ac (5 GHz)       5 GHz      80 MHz
+         * 6 IEEE 802.11ad (60 GHz)      60 GHz     160 MHz      Not supported
+         * 7 IEEE 802.11af (whitespace)                          Not supported
+         * 8 IEEE 802.11ax (2.4 GHz)     2.4 GHz    160 MHz      Not included in Table 6-12—Media type (intfType), WiFi6 is specified to use 0x0108 in R2
+         * 8 IEEE 802.11ax (5 GHz)       5 GHz      160 MHz      Not included in Table 6-12—Media type (intfType), WiFi6 is specified to use 0x0108 in R2
+         */
+
+        std::unique_ptr<bwl::nl80211_client> client =
```
tomereli: I think this is wrong to put this here. The backhaul manager already maintains a radio info map - `std::unordered_map<sMacAddr, sRadioInfo> m_radio_info_map;` which is first filled in `BACKHAUL_ENABLE` message from the son slave, and then on every event - client association, VAPS list update, etc.

A design which will better fit prplmesh architecture is to read all this information in the BWL when the radio info is built, then the backhaul manager will simply have it for every slave.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: Perhaps this block can be moved to a separate function `max_bandwidth_to_media_type`? (nitpick)

tomereli (changes requested):
One major comment - reading radio info should only be done by the BWL backend. We read from hostapd whatever we can, and supplement what's missing with the nl80211_client.
It makes the code easier to maintain and allows us to completely remove the dummy implmentation for the nl80211_client which I think is redundant now that we have the nl80211_client member in the BWL backend base class .

So, I suggest to update `refresh_radio_info` to read the missing values using the nl80211_client and use it eventually in the backhaul manager by parsing the `m_radio_info_map`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

framework/tlvf/AutoGenerated/include/tlvf/ieee_1905_1/eMediaType.h
mariomaz: Not squashed because the change is not so small after having addressed all the comments in the review.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

tomereli (approved):

PR #1061, opened 2020-03-31
check rdkb wlan task state before handling steering socket remove event

kantera800:
Currently if any controller socket is disconnected an additional
STEERING_REMOVE_SOCKET event is sent to rdkb wlan task.
In case there are no actual steering listeners to remove this event
results in a failure and prints several error messages.
This happens each time any 3rd party API is triggered via BML and
fills the controller log file with unnecessary error messages.

Solution is to check if steering listener socket even exists before
trying to remove it.

### DIFF
--- a/controller/src/beerocks/master/tasks/rdkb/rdkb_wlan_task.cpp
+++ b/controller/src/beerocks/master/tasks/rdkb/rdkb_wlan_task.cpp
@@ -377,16 +377,18 @@ void rdkb_wlan_task::handle_event(int event_type, void *obj)
     case STEERING_REMOVE_SOCKET: {
         if (obj) {
             auto event_obj = (listener_general_register_unregister_event *)obj;
-            TASK_LOG(DEBUG) << "STEERING_REMOVE_SOCKET event was received";
+            if (is_bml_rdkb_wlan_listener_socket(event_obj->sd)) {
+                TASK_LOG(DEBUG) << "STEERING_REMOVE_SOCKET event was received";
 
-            if (!set_bml_rdkb_wlan_events_update_enable(event_obj->sd, false)) {
-                TASK_LOG(ERROR) << "fail in set_bml_rdkb_wlan_events_update_enable";
-            }
+                if (!set_bml_rdkb_wlan_events_update_enable(event_obj->sd, false)) {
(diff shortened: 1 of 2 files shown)


arnout (approved):
Would be nice to have a generic way to do this, but OK.

tomereli (approved):

kantera800:
@morantr added valuable (but offline) comment that it better to check if  socket exists.
made the changes and waiting for his approval.

controller/src/beerocks/master/tasks/rdkb/rdkb_wlan_task.cpp
```diff
@@ -377,16 +377,18 @@ void rdkb_wlan_task::handle_event(int event_type, void *obj)
...
 
-            if (!set_bml_rdkb_wlan_events_update_enable(event_obj->sd, false)) {
-                TASK_LOG(ERROR) << "fail in set_bml_rdkb_wlan_events_update_enable";
-            }
+                if (!set_bml_rdkb_wlan_events_update_enable(event_obj->sd, false)) {
+                    TASK_LOG(ERROR) << "fail in set_bml_rdkb_wlan_events_update_enable";
+                }
 
-            remove_bml_rdkb_wlan_socket(event_obj->sd);
+                remove_bml_rdkb_wlan_socket(event_obj->sd);
```
morantr, line 387: Actually, I don't think this is necessary because it is already done in `remove_bml_rdkb_wlan_socket()`.
So I would recommend removing the function you added and remove also `set_bml_rdkb_wlan_events_update_enable()` since there is no point to set the 'events_update' flag if you remove the socket afterward on `remove_bml_rdkb_wlan_socket()`.
The only thing needs to be done here is to add return value (bool) to `remove_bml_rdkb_wlan_socket()` that indicates if the socket has been removed. If it has been removed, print the message to the log.

morantr (approved):
I think it can be better, but approving in order to no block this PR.

controller/src/beerocks/master/tasks/rdkb/rdkb_wlan_task.cpp
kantera800: This is a good point  , but then all steering APIs needs to be retested and this is not the purpose of this PR.
What i think needs to be done is to refactor RDKB task or to consider removing it completely (maybe move the implementation to rdkb wave api repo)

PR #1062, opened 2020-04-01
coding style: use std algorithm

RanRegev:
added a recommendation to use std algorithms as much as possible, including the rational behind it.

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -141,6 +141,18 @@ Example:
     }
 ```
 
+Prefer to use std algorithms as much as possible.
+Get to know std algorithms.
+std algorithms are better than hand crafted loops.
+std algorithms are thoroughly tested, well documented, and do exactly as expected.
+std algoritms are also well documented.
+```cpp
+std::transform(in.begin(), in.end(), out.begin(), 
+               [](/*...*/) -> /*...*/ {
+                   /*...*/
+               } );
+```
+
 Prefer to use ranged loops if at all possible:
 
 ```cpp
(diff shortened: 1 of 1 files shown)


CODINGSTYLE.md
```diff
@@ -140,6 +140,14 @@ Example:
         return false;
     }
 ```
+Prefer to use std algorithms as much as possible.
```
arnout, line 144: Nitpick: add a newline above

CODINGSTYLE.md
```diff
@@ -140,6 +140,14 @@ Example:
...
+Get to know std algorithms.
+std algorithms are better than hand crafted loops.
+std algorithms are thoughful, and tested, and do as expected.
+And std algoritms are also well documented.
+```cpp
+std::transform( in.begin(), in.end(), out.begin() [/*...*/](/*...*/){/*...*/} );
```
arnout: The example should satisfy the coding style, and have correct lamdba syntax:
```suggestion
std::transform(in.begin(), in.end(), out.begin(), 
               [](/*...*/) -> /*...*/ {
                   /*...*/
               } );
```

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
+Comments have the sole purpose of keeping the code maintainable.
+A good comment is a one that you, the writer of the comment, can understand after a while, after getting out of the context of writing it.
+Comments are written for future readers, it may be the future you, it may be someone else - 
+in this context in mind write your comments.
+Comments normaly explains the *why* and not the *what*.
+
+### Simple Comments
+It is useless to state the obvious:
+```cpp
+// checkig that the poinnter is not null
```
arnout: Comments with spelling errors are bad as well, but that's not what we want to say here :-)
```suggestion
// checking that the pointer is not null
```

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
+// in case of exiting pointer pointer we need to ...
+// and in case it is null we need to ...
+if (ptr){
+} else {
+}
+```
+
+### Spanning Comments
+There are cases that writing a comment accross few segments of the code, usually within a function's body, makes code understanding much easier.
+Sometimes it is hard to see behind the machincs of the code its intent.
```
arnout:
```suggestion
Sometimes it is hard to see the intent behind the mechanics of the code.
```

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
+// in this section we have to gather information from few resources:
+// A, B and C.
+// The best would be to have all of them, but it is also legal and
+// possible to progress with A and C only.
+// therefore we try to get A, then we try to get C,
+// and upon success we try to get B as well.
+// the action is taken at the end.
+
+// get A //
+/* 10 lines of complex machinary, iterating on a complex data structure */
```
arnout:
```suggestion
/* 10 lines of complex machinery, iterating on a complex data structure */
```

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
+```cpp
+// in this section we have to gather information from few resources:
+// A, B and C.
+// The best would be to have all of them, but it is also legal and
+// possible to progress with A and C only.
+// therefore we try to get A, then we try to get C,
+// and upon success we try to get B as well.
+// the action is taken at the end.
+
+// get A //
```
arnout: I'm not so sure about this `//` at the end. We don't do that anywhere at the moment, and I don't think the visual attention grabbing is sufficient to bother with it.

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
+// and in case it is null we need to ...
+if (ptr){
+} else {
+}
+```
+
+### Spanning Comments
+There are cases that writing a comment accross few segments of the code, usually within a function's body, makes code understanding much easier.
+Sometimes it is hard to see behind the machincs of the code its intent.
+Adding descriptive comments that "tells the story" makes it much easier to follow the code structure and logic even before reading the code itself. Example:
```
arnout: I really liked this part of your original comment, and it's missing here:

```
Sometimes the code is not trivial even if it doesn't do tricky things. 
It just has its logic that might be buried in a few lines of technical details that have nothing to do with the logic itself, like error exits.
In these cases it is more maintainable to comment the different sections in the function and their logic.
A small example can be seen [here](https://github.com/prplfoundation/prplMesh/blob/20388c47cf61f9419576fa223fd140e17adcb67d/agent/src/beerocks/slave/son_slave_thread.cpp#L4373-L4405).
This else clause starts with a story about what is done in it. 
In this case not what would you expect. 
It also states why it is fine and suggests future changes.
Later in the code the promise that was taken in the initial comment is implemented.
It is useful to read the first comment, then look forward in the code for more comments that support the story on top, and only then dive into the code to ensure it does what it says it does.
```

(I did minor editing of this text to make it fit with what you expect in documentation, e.g. avoid "I").

I think the code section that you refer to is a lot better than the example below, because it is real (and still simple enough).

arnout (approved):
Thanks for this!

CODINGSTYLE.md
```diff
@@ -140,6 +140,14 @@ Example:
         return false;
     }
 ```
+Prefer to use std algorithms as much as possible.
+Get to know std algorithms.
+std algorithms are better than hand crafted loops.
+std algorithms are thoughful, and tested, and do as expected.
```
tomereli: Is it a misspell of `thoughtful`? If so, what does it mean? I think a better way to say this is:
```
std algorithms are thoroughly tested, well documented, and do exactly as expected.
```

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
 ```
 
+## Comments
+Comments are important part of the code base.
+Comments are not compiled, nor have any influence on run time or any machinary.
+Comments are for humans, for developers, for us.
+Comments have the sole purpose of keeping the code maintainable.
+A good comment is a one that you, the writer of the comment, can understand after a while, after getting out of the context of writing it.
+Comments are written for future readers, it may be the future you, it may be someone else - 
+in this context in mind write your comments.
```
tomereli:
```suggestion
write your comments with this in mind.
```

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
+Comments normaly explains the *why* and not the *what*.
```
tomereli: Good!

tomereli:
```suggestion
Comments normally explain the *why* and not the *what*.
```

CODINGSTYLE.md
```diff
@@ -204,6 +204,70 @@ In the following example, the empty string is the sentinel value indicating fail
...
+
+### Simple Comments
+It is useless to state the obvious:
+```cpp
+// checkig that the poinnter is not null
+if (ptr){
+}
+```
+
+However, it is extremely helpful to state our expectations
```
tomereli:
```suggestion
However, it is extremely helpful to state our expectations:
```

tomereli (approved):

PR #1065, opened 2020-04-01
bwl: update basic radio capabilities

### LINKED ISSUE #1064: [TASK] Update Radio Basic Capabilities TLV using the new supported_channels
## Description
After Task #1024 
The supported_channels are needed to created the Radio Basic Capabilities TLV
To do so they need to be sent to the son_slave_thread, where they will be stored locally
Then the son_slave_thread needs to sent the stored hardware_supported_channels to the son_master_thread as part of the Radio Basic Capabilities TLV

## Exit criteria

When constructing the Radio Basic Capabilities TLV, use the hardware_supported_channels got from the new supported_channels

itayx:
The supported_channels are needed to created the Radio Basic Capabilities TLV
To do so they need to be sent to the son_slave_thread, where they will be stored locally
Then the son_slave_thread needs to sent the stored hardware_supported_channels to the son_master_thread as part of the Radio Basic Capabilities TLV

### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1569,6 +1569,9 @@ void ap_manager_thread::handle_hostapd_attached()
                 beerocks::message::VHT_MCS_SET_SIZE, notification->params().vht_mcs_set);
 
     // Copy the channels supported by the AP
+    auto tuple_supported_channels = notification->supported_channels(0);
+    copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
+                        &std::get<1>(tuple_supported_channels));
     copy_radio_channels(ap_wlan_hal->get_radio_info().preferred_channels,
                         notification->params().preferred_channels);
 
@@ -1588,6 +1591,8 @@ void ap_manager_thread::handle_hostapd_attached()
     LOG(INFO) << " vht_capability = " << std::hex << ap_wlan_hal->get_radio_info().vht_capability;
     LOG(INFO) << " preferred_channels = " << std::endl
               << get_radio_channels_string(ap_wlan_hal->get_radio_info().preferred_channels);
+    LOG(INFO) << " supported_channels = " << std::endl
+              << get_radio_channels_string(ap_wlan_hal->get_radio_info().supported_channels);
 
     // Send CMDU
     message_com::send_cmdu(slave_socket, cmdu_tx);
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -39,7 +39,7 @@ static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_
     auto radio_channels = ap_wlan_hal->get_radio_info().supported_channels;
 
     // Copy the channels
-    for (uint i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH && i < radio_channels.size();
+    for (uint i = 0; i < beerocks::message::RADIO_CHANNELS_LENGTH && i < radio_channels.size();
          i++) {
 
         supported_channels[i].channel        = radio_channels[i].channel;
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1512,6 +1512,7 @@ void ap_manager_thread::handle_hostapd_attached()
 {
     LOG(DEBUG) << "handling enabled hostapd";
 
+    ap_wlan_hal->read_supported_channels();
     if (acs_enabled) {
         LOG(DEBUG) << "retrieving ACS report";
         int read_acs_attempt = 0;
@@ -1527,7 +1528,7 @@ void ap_manager_thread::handle_hostapd_attached()
             usleep(ACS_READ_SLEEP_USC);
         }
     } else {
-        ap_wlan_hal->read_supported_channels();
+        ap_wlan_hal->update_preference_channels_from_supported_channels();
     }
 
     auto notification =
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -33,8 +33,9 @@ using namespace beerocks::net;
 /////////////////////////// Local Module Functions ///////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-static void copy_radio_channels(std::vector<bwl::WiFiChannel> radio_channels,
-                                beerocks::message::sWifiChannel dest_channels[])
+static void copy_radio_channels(
+    std::array<bwl::WiFiChannel, beerocks::message::RADIO_CHANNELS_LENGTH> radio_channels,
+    beerocks::message::sWifiChannel dest_channels[])
 {
 
     // Copy the channels
@@ -51,7 +52,8 @@ static void copy_radio_channels(std::vector<bwl::WiFiChannel> radio_channels,
     }
 }
 
-static std::string get_radio_channels_string(std::vector<bwl::WiFiChannel> radio_channels)
+static std::string get_radio_channels_string(
+    std::array<bwl::WiFiChannel, beerocks::message::RADIO_CHANNELS_LENGTH> radio_channels)
 {
     std::ostringstream os;
     for (auto val : radio_channels) {
@@ -784,9 +786,9 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             LOG(ERROR) << "Failed building message!";
             return false;
         }
-        auto tuple_supported_channels = response->supported_channels_list(0);
-        copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
-                            &std::get<1>(tuple_supported_channels));
+        auto tuple_preferred_channels = response->preferred_channels_list(0);
+        copy_radio_channels(ap_wlan_hal->get_radio_info().preferred_channels,
+                            &std::get<1>(tuple_preferred_channels));
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
         break;
@@ -1110,9 +1112,9 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
                 LOG(ERROR) << "Failed building message!";
                 return false;
             }
-            auto tuple_supported_channels = notification->supported_channels_list(0);
-            copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
-                                &std::get<1>(tuple_supported_channels));
+            auto tuple_preferred_channels = notification->preferred_channels_list(0);
+            copy_radio_channels(ap_wlan_hal->get_radio_info().preferred_channels,
+                                &std::get<1>(tuple_preferred_channels));
             fill_cs_params(notification->cs_params());
             acs_completed_vap_update = true;
         } else {
@@ -1567,8 +1569,8 @@ void ap_manager_thread::handle_hostapd_attached()
                 beerocks::message::VHT_MCS_SET_SIZE, notification->params().vht_mcs_set);
 
     // Copy the channels supported by the AP
-    copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
-                        notification->params().supported_channels);
+    copy_radio_channels(ap_wlan_hal->get_radio_info().preferred_channels,
+                        notification->params().preferred_channels);
 
     LOG(INFO) << "send ACTION_APMANAGER_JOINED_NOTIFICATION";
     LOG(INFO) << " iface = " << ap_wlan_hal->get_iface_name();
@@ -1584,8 +1586,8 @@ void ap_manager_thread::handle_hostapd_attached()
     LOG(INFO) << " ht_capability = " << std::hex << ap_wlan_hal->get_radio_info().ht_capability;
     LOG(INFO) << " vht_supported = " << ap_wlan_hal->get_radio_info().vht_supported;
     LOG(INFO) << " vht_capability = " << std::hex << ap_wlan_hal->get_radio_info().vht_capability;
-    LOG(INFO) << " supported_channels = " << std::endl
-              << get_radio_channels_string(ap_wlan_hal->get_radio_info().supported_channels);
+    LOG(INFO) << " preferred_channels = " << std::endl
+              << get_radio_channels_string(ap_wlan_hal->get_radio_info().preferred_channels);
 
     // Send CMDU
     message_com::send_cmdu(slave_socket, cmdu_tx);
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -33,30 +33,28 @@ using namespace beerocks::net;
 /////////////////////////// Local Module Functions ///////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
-                                          beerocks::message::sWifiChannel supported_channels[])
+static void copy_radio_channels(std::vector<bwl::WiFiChannel> radio_channels,
+                                beerocks::message::sWifiChannel dest_channels[])
 {
-    auto radio_channels = ap_wlan_hal->get_radio_info().supported_channels;
 
     // Copy the channels
     for (uint i = 0; i < beerocks::message::RADIO_CHANNELS_LENGTH && i < radio_channels.size();
          i++) {
 
-        supported_channels[i].channel        = radio_channels[i].channel;
-        supported_channels[i].noise          = radio_channels[i].noise;
-        supported_channels[i].tx_pow         = radio_channels[i].tx_pow;
-        supported_channels[i].bss_overlap    = radio_channels[i].bss_overlap;
-        supported_channels[i].is_dfs_channel = radio_channels[i].is_dfs;
-        supported_channels[i].channel_bandwidth =
+        dest_channels[i].channel        = radio_channels[i].channel;
+        dest_channels[i].noise          = radio_channels[i].noise;
+        dest_channels[i].tx_pow         = radio_channels[i].tx_pow;
+        dest_channels[i].bss_overlap    = radio_channels[i].bss_overlap;
+        dest_channels[i].is_dfs_channel = radio_channels[i].is_dfs;
+        dest_channels[i].channel_bandwidth =
             uint8_t(beerocks::utils::convert_bandwidth_to_enum(radio_channels[i].bandwidth));
     }
 }
 
-static std::string
-get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
+static std::string get_radio_channels_string(std::vector<bwl::WiFiChannel> radio_channels)
 {
     std::ostringstream os;
-    for (auto val : ap_wlan_hal->get_radio_info().supported_channels) {
+    for (auto val : radio_channels) {
         if (val.channel > 0) {
             os << " ch = " << int(val.channel) << " | dfs = " << int(val.tx_pow)
                << " | tx_pow = " << int(val.is_dfs) << " | noise = " << int(val.noise)
@@ -787,7 +785,8 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             return false;
         }
         auto tuple_supported_channels = response->supported_channels_list(0);
-        copy_radio_supported_channels(ap_wlan_hal, &std::get<1>(tuple_supported_channels));
+        copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
+                            &std::get<1>(tuple_supported_channels));
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
         break;
@@ -1112,7 +1111,8 @@ bool ap_manager_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t ev
                 return false;
             }
             auto tuple_supported_channels = notification->supported_channels_list(0);
-            copy_radio_supported_channels(ap_wlan_hal, &std::get<1>(tuple_supported_channels));
+            copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
+                                &std::get<1>(tuple_supported_channels));
             fill_cs_params(notification->cs_params());
             acs_completed_vap_update = true;
         } else {
@@ -1567,7 +1567,8 @@ void ap_manager_thread::handle_hostapd_attached()
                 beerocks::message::VHT_MCS_SET_SIZE, notification->params().vht_mcs_set);
 
     // Copy the channels supported by the AP
-    copy_radio_supported_channels(ap_wlan_hal, notification->params().supported_channels);
+    copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
+                        notification->params().supported_channels);
 
     LOG(INFO) << "send ACTION_APMANAGER_JOINED_NOTIFICATION";
     LOG(INFO) << " iface = " << ap_wlan_hal->get_iface_name();
@@ -1584,7 +1585,7 @@ void ap_manager_thread::handle_hostapd_attached()
     LOG(INFO) << " vht_supported = " << ap_wlan_hal->get_radio_info().vht_supported;
     LOG(INFO) << " vht_capability = " << std::hex << ap_wlan_hal->get_radio_info().vht_capability;
     LOG(INFO) << " supported_channels = " << std::endl
-              << get_radio_supported_channels_string(ap_wlan_hal);
+              << get_radio_channels_string(ap_wlan_hal->get_radio_info().supported_channels);
 
     // Send CMDU
     message_com::send_cmdu(slave_socket, cmdu_tx);
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -1789,7 +1789,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         notification_out->cs_params()     = notification_in->cs_params();
         auto tuple_in_supported_channels  = notification_in->supported_channels_list(0);
         auto tuple_out_supported_channels = notification_out->supported_channels(0);
-        std::copy_n(&std::get<1>(tuple_in_supported_channels), message::SUPPORTED_CHANNELS_LENGTH,
+        std::copy_n(&std::get<1>(tuple_in_supported_channels), message::RADIO_CHANNELS_LENGTH,
                     &std::get<1>(tuple_out_supported_channels));
         send_cmdu_to_controller(cmdu_tx);
         send_operating_channel_report();
@@ -2239,7 +2239,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         }
 
         auto tuple_supported_channels = response->supported_channels_list(0);
-        std::copy_n(&std::get<1>(tuple_supported_channels), message::SUPPORTED_CHANNELS_LENGTH,
+        std::copy_n(&std::get<1>(tuple_supported_channels), message::RADIO_CHANNELS_LENGTH,
                     hostap_params.supported_channels);
 
         // build channel preference report
@@ -3258,7 +3258,7 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             break;
         }
 
-        std::copy_n(hostap_params.supported_channels, message::SUPPORTED_CHANNELS_LENGTH,
+        std::copy_n(hostap_params.supported_channels, message::RADIO_CHANNELS_LENGTH,
                     &std::get<1>(tuple_supported_channels));
 
         // Send the message
@@ -3383,7 +3383,7 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             return false;
         }
 
-        std::array<beerocks::message::sWifiChannel, beerocks::message::SUPPORTED_CHANNELS_LENGTH>
+        std::array<beerocks::message::sWifiChannel, beerocks::message::RADIO_CHANNELS_LENGTH>
             supported_channels;
         std::copy_n(std::begin(hostap_params.supported_channels), supported_channels.size(),
                     supported_channels.begin());
@@ -4500,7 +4500,7 @@ beerocks::message::sWifiChannel slave_thread::channel_selection_select_channel()
         if (preference.channels.empty()) {
             continue;
         }
-        for (uint8_t i = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+        for (uint8_t i = 0; i < beerocks::message::RADIO_CHANNELS_LENGTH; i++) {
             const auto &channel  = hostap_params.supported_channels[i];
             auto operating_class = wireless_utils::get_operating_class_by_channel(channel);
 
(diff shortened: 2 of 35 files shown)


tomereli:
I don't like where this is going - reading supported channels should be done as part of `refresh_radio_info`. Please make sure to follow this direction.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -27,6 +27,9 @@ cACTION_APMANAGER_JOINED_NOTIFICATION:
   _type: class
   params: sNodeHostap
   cs_params: sApChannelSwitch
+  supported_channels_list:
+    _type: beerocks::message::sWifiChannel 
+    _length: [ "beerocks::message::SUPPORTED_CHANNELS_LENGTH" ]
```
tomereli: Since it is the last member, you don't need to specify the length, it can be dynamic.
Also - no need to say its a list, the 's' in the channels is enough.
So I would change it to:
```
supported_channels:
  _type: beerocks::message::sWifiChannel
  _length: [ ]
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -53,14 +53,20 @@ static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_
 }
 
 static std::string
```
tomereli: Please use "." in the commit messages - don't leave paragraphs hanging without an ending :)

tomereli: Also don't use "hardware supported channels" - the name is "supported channels".

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1510,6 +1516,11 @@ void ap_manager_thread::handle_hostapd_attached()
 {
     LOG(DEBUG) << "handling enabled hostapd";
 
+    ap_wlan_hal->read_supported_channels();
```
tomereli, line 1515: This is not what I intended - the supported channels should stay part of the radio info, no need for a new API.
Instead, in `base_wlan_hal_dwpal` initialization, read the supported channels from NL once. Then there is no change needed in the upper layer - we will still use `get_radio_info().supported_channels`.
In addition, the current use of the supported channels which are overrided by acs_report is wrong and it should be saving the report into `preferred_channels` under the same `m_radio_info` member.

So all this commit and probably the following commits should be changes, so I'll ask for changes.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1569,8 +1586,9 @@ void ap_manager_thread::handle_hostapd_attached()
     LOG(INFO) << " vht_center_frequency = " << ap_wlan_hal->get_radio_info().vht_center_freq;
     LOG(INFO) << " current bw = " << ap_wlan_hal->get_radio_info().bandwidth;
     LOG(INFO) << " supported_channels = " << std::endl
-              << get_radio_supported_channels_string(ap_wlan_hal);
-
+              << get_radio_supported_channels_string(notification->params().supported_channels);
+    LOG(INFO) << " hw_supported_channels = " << std::endl
```
tomereli: The terminology should be as follows:
supported channels - these are the supported channels by the radio we read from NL
preferred channels - these are the preferred channels we get from the ACS report

tomereli (changes requested):

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
arnout: :-1: dynamic length is very inconvenient because it is expressed in bytes, not in number of elements. So an explicit length is actually preferable.

(except if I'm mistaken and it really is expressed in number of elements)

agent/src/beerocks/slave/ap_manager_thread.cpp
arnout: > In addition, the current use of the supported channels which are overrided by acs_report is wrong

I think the idea is to do this step by step, i.e. adding the preferred_channels later. This update is already a step in the right direction, and it is a necessary step.

That said, introducing the name `hw_supported_channels` is wrong, of course.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
arnout: Let me modify that statement: *unkown length* is annoying. *dynamic length* is good:
```suggestion
supported_channels_length: uint8_t
supported_channels:
  _type: "beerocks::message::sWifiChannel"
  _length: [ supported_channels_length ]
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -53,14 +53,20 @@ static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_
 }
 
 static std::string
-get_radio_supported_channels_string(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal)
+get_radio_supported_channels_string(beerocks::message::sWifiChannel supported_channels[])
```
arnout: I don't really see a reason to change this function. Or is it because you used it to verify that the way you constructed the channel list was correct?

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1510,6 +1516,11 @@ void ap_manager_thread::handle_hostapd_attached()
 {
     LOG(DEBUG) << "handling enabled hostapd";
 
+    ap_wlan_hal->read_supported_channels();
+    beerocks::message::sWifiChannel
+        hw_supported_channels[beerocks::message::SUPPORTED_CHANNELS_LENGTH];
```
arnout: This tendency to use arrays of static size makes me want to puke... What's wrong with good old std::vector?

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1561,6 +1572,12 @@ void ap_manager_thread::handle_hostapd_attached()
     // Copy the channels supported by the AP
     copy_radio_supported_channels(ap_wlan_hal, notification->params().supported_channels);
 
+    auto supported_channels_list = &std::get<1>(notification->supported_channels_list(0));
+    for (uint i = 0, j = 0; i < beerocks::message::SUPPORTED_CHANNELS_LENGTH; i++) {
+        if (hw_supported_channels[i].channel != 0)
+            supported_channels_list[j++] = hw_supported_channels[i];
+    }
```
arnout: This is silly... we first copy into a temporary array and then copy again into another array. Why not simply copy from ap_wlan_hal here?
```suggestion
    uint i = 0;
    for (const auto &channel : ap_wlan_hal->get_radio_info().supported_channels) {
        if (channel.channel != 0) {
            supported_channels[i].channel        = channel.channel;
            ...
        }
    }
```

You could refactor that into a second copy_supported_channels function since it will probably be used elsewhere as well.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
tomereli: I wasn't aware, you are correct. IMO the length should return the number of elements and not the size of them, @LiorAmram what do you think?

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -200,6 +200,7 @@ class slave_thread : public beerocks::socket_thread {
     sSlaveBackhaulParams backhaul_params;
     beerocks_message::sNodeHostap hostap_params;
     beerocks_message::sApChannelSwitch hostap_cs_params;
+    std::vector<beerocks::message::sWifiChannel> hardware_supported_channels;
```
arnout: Apparently the message got lost somewhere: there is no such thing as hardware supported channels. There are supported channels, and there are channel preferences. What needs to be added is channel preferences, and ACS needs to use that instead of supported_channels.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
LiorAmram: Unknown length means there is no explicit member on the buffer that indicates what is the length or the number of elements. The only way to figure it out is parsing element by element until end of buffer is reached.
(Sidenote: When building or parsing the class, there is an internal private size element that indicates the **number** of elements in the list. However, it can't be accessed from outside the class as far as I remember).

So Unknown length isn't expressed in bytes, it's actually isn't expressed at all.

When the discussion in whether to use a dynamic list (with length variable) or Unknown length list (without length variable), I prefer using variable-length lists whenever possible.

However, in this code, there is another issue: the specified length is a const, which means all elements of the list will be allocated in advance. Is it the desired output? If not, it might be preferable to define a length variable to supported_channels.

arnout (commented):
In addition to the comments I made: I would expect supported_channels to be removed from sNodeHostap as well. Is there any reason to keep it there?

Not marking as changes requested because I don't know if I'll still do the re-review.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
LiorAmram: @arnout a small fix to your suggestion:
```suggestion
supported_channels_length: 
  _type: uint8_t
  _length_var: True
supported_channels:
  _type: "beerocks::message::sWifiChannel"
  _length: [ supported_channels_length ]
```

Setting `_length_var: True` is necessary for tlvf to update this variable when allocating new elements to the list. Otherwise, it's just another member on the buffer.

kantera800 (changes requested):
I saw couple valuable comments in this PR that need to be addressed.
IMO this is what is still missing:
1.  in `handle_hostapd_attached() `- when there is no "acs report" need to `update_preference_channels_from_supoorted_channels()` instead of `read_supported_channels()`
2. need to add controller commit that removes `set_hostap_supported_channels()` from `handle_intel_slave_join()` since only `add_hostap_supported_operating_class()` will handle this from now on.
3. verify changes in 2 scenarios :  
		a. Radio configured as Auto channel
		b. Radio configured as Static channel
	print "supported_channels" after autoconfig_wsc_parse_radio_caps().
	Check that in both scenarios prints are the same.
	add the result to PR description.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
itayx: The conversation is moot, the supported_channels was moved to the sNodeHostap

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: Yes, to help with debugging

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: Changed where I could to use vector instead

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: the point is moot,
The supported_channels are now stored as part of the RadioInfo struct

agent/src/beerocks/slave/son_slave_thread.h
itayx: Changed the previously  called `supported_channels` to `preferred_channels` throughout the codebase

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: Re worded most of the commit messages
The new commits now use "supported channels" and "preferred channels"

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: Changed

itayx:
This is the result of the print in `master_thread::autoconfig_wsc_parse_radio_caps` after the changes made in this PR

I checked the print two times, one with the radio channel set to auto
and one with the radio channel set to a static value (5G: 36, 2.4G: 5)

## While on auto-channel:
> operating_class=115
maximum_transmit_power_dbm=30
channel list={ 36 40 44 48 }
statically_non_operable_channel_list={  }
operating_class=116
maximum_transmit_power_dbm=30
channel list={ 36 44 }
statically_non_operable_channel_list={  }
operating_class=117
maximum_transmit_power_dbm=30
channel list={ 40 48 }
statically_non_operable_channel_list={  }
operating_class=124
maximum_transmit_power_dbm=30
channel list={ 149 153 157 161 }
statically_non_operable_channel_list={  }
operating_class=125
maximum_transmit_power_dbm=30
channel list={ 149 153 157 161 165 169 }
statically_non_operable_channel_list={ 169  }
operating_class=126
maximum_transmit_power_dbm=30
channel list={ 149 157 }
statically_non_operable_channel_list={  }
operating_class=127
maximum_transmit_power_dbm=30
channel list={ 153 161 }
statically_non_operable_channel_list={  }
operating_class=81
maximum_transmit_power_dbm=30
channel list={ 1 2 3 4 5 6 7 8 9 10 11 12 13 }
statically_non_operable_channel_list={ 12 13  }
operating_class=83
maximum_transmit_power_dbm=30
channel list={ 1 2 3 4 5 6 7 8 9 }
statically_non_operable_channel_list={  }

## while on static channel:
> operating_class=81
maximum_transmit_power_dbm=30
channel list={ 1 2 3 4 5 6 7 8 9 10 11 12 13 }
statically_non_operable_channel_list={ 12 13  }
operating_class=83
maximum_transmit_power_dbm=30
channel list={ 1 2 3 4 5 6 7 8 9 }
statically_non_operable_channel_list={  }
operating_class=115
maximum_transmit_power_dbm=30
channel list={ 36 40 44 48 }
statically_non_operable_channel_list={  }
operating_class=116
maximum_transmit_power_dbm=30
channel list={ 36 44 }
statically_non_operable_channel_list={  }
operating_class=117
maximum_transmit_power_dbm=30
channel list={ 40 48 }
statically_non_operable_channel_list={  }
operating_class=124
maximum_transmit_power_dbm=30
channel list={ 149 153 157 161 }
statically_non_operable_channel_list={  }
operating_class=125
maximum_transmit_power_dbm=30
channel list={ 149 153 157 161 165 169 }
statically_non_operable_channel_list={ 169  }
operating_class=126
maximum_transmit_power_dbm=30
channel list={ 149 157 }
statically_non_operable_channel_list={  }
operating_class=127
maximum_transmit_power_dbm=30
channel list={ 153 161 }

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: Changed all misuse of the get_radio_info().supported_channels to now use the new preferred_channels
Also changed it so now the base_wlan_hal_dwpal's refresh_radio_info will save the supported channels and added a function called `update_preference_channels_from_supoorted_channels` what will copy the supported channels to the preferred_channels in case the acs is not enabled

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -2101,9 +2101,6 @@ bool master_thread::handle_intel_slave_join(
     database.set_node_ipv4(radio_mac, bridge_ipv4);
     database.set_node_manufacturer(radio_mac, "Intel");
 
-    database.set_hostap_supported_channels(radio_mac, notification->hostap().supported_channels,
```
kantera800, line 2104: @arnout already mentioned it, 
please remove supported_channels from sNodeHostap seems like this is the only place its used in the controller

also, use get_hostap_supported_channels_string() instead of using this code  (line 2083)
```
        for (auto supported_channel : notification->hostap().supported_channels) {
            if (supported_channel.channel > 0) {
                LOG(DEBUG) << "supported_channel = " << int(supported_channel.channel);
            }
        }
```

itayx: with regard to this comment and the one made by @arnout
The supported channels still need to be part of the sNodeHostap so they will be sent to the son_slave_thread as part of the `ACTION_APMANAGER_JOINED_NOTIFICATION` message

because of this when the radio capabilities are sent as part of the `cACTION_CONTROL_SLAVE_JOINED_NOTIFICATION` message and can be disregard

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2240,7 +2240,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
 
         auto tuple_supported_channels = response->supported_channels_list(0);
```
kantera800: `preferred_channels = supported_channels`  is confusing 
if we consider ACS report as preferred channels , why not change the CMDU parameter as well .
the code will be more understandable:

```
auto tuple_preferred_channels = response->preferred_channels_list(0);
        std::copy_n(&std::get<1>(tuple_supported_channels), message::SUPPORTED_CHANNELS_LENGTH,
                    hostap_params.preferred_channels);
```

itayx: Because that will mean an overhaul of changes throughtout the code base to change all `supported_channels` to `preferred_channels`

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3258,7 +3258,7 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
             break;
         }
 
-        std::copy_n(hostap_params.supported_channels, message::SUPPORTED_CHANNELS_LENGTH,
+        std::copy_n(hostap_params.preferred_channels, message::SUPPORTED_CHANNELS_LENGTH,
                     &std::get<1>(tuple_supported_channels));
```
kantera800: same here 
`preferred_channels = supported_channels` is  confusing

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
```diff
@@ -318,6 +318,14 @@ class ap_wlan_hal : public virtual base_wlan_hal {
      */
     virtual bool read_supported_channels() = 0;
 
+    /*!
+     * Copy the supported channls to the preferred channls
+     * On successful completion the preferred channls can be retrieved
```
kantera800:
```suggestion
     * On successful completion the preferred channels can be retrieved 
```

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.h
```diff
@@ -67,6 +67,7 @@ class ap_wlan_hal_nl80211 : public base_wlan_hal_nl80211, public ap_wlan_hal {
     virtual bool restricted_channels_get(char *channel_list) override;
     virtual bool read_acs_report() override;
     virtual bool read_supported_channels() override;
+    virtual bool update_preference_channels_from_supoorted_channels() override;
```
kantera800:
```suggestion
    virtual bool update_preferred_channels_from_supported_channels() override;
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -785,9 +785,9 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             LOG(ERROR) << "Failed building message!";
             return false;
         }
-        auto tuple_supported_channels = response->supported_channels_list(0);
-        copy_radio_channels(ap_wlan_hal->get_radio_info().supported_channels,
-                            &std::get<1>(tuple_supported_channels));
+        auto tuple_preferred_channels = response->supported_channels_list(0);
+        copy_radio_channels(ap_wlan_hal->get_radio_info().preferred_channels,
+                            &std::get<1>(tuple_preferred_channels));
```
kantera800: same here
`preferred_channels = supported_channels` is confusing
IMO you need to use preferred_channels in all ACS related CMDUs  as a preparation commit.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1510,6 +1510,9 @@ void ap_manager_thread::handle_hostapd_attached()
 {
     LOG(DEBUG) << "handling enabled hostapd";
 
+    //the supported channels must always be read
```
kantera800: you can remove this comment since the code is self explained

controller/src/beerocks/master/son_master_thread.cpp
kantera800: > with regard to this comment and the one made by @arnout
> The supported channels still need to be part of the sNodeHostap so they will be sent to the son_slave_thread as part of the `ACTION_APMANAGER_JOINED_NOTIFICATION` message
> 
> because of this when the radio capabilities are sent as part of the `cACTION_CONTROL_SLAVE_JOINED_NOTIFICATION` message and can be disregard

since its the only place its used lets just add it there

```
cACTION_APMANAGER_JOINED_NOTIFICATION:
  _type: class
  params: sNodeHostap
  cs_params: sApChannelSwitch
  supported_channels_list:
    _type: beerocks::message::sWifiChannel
    _length: [ "beerocks::message::SUPPORTED_CHANNELS_LENGTH" ]
```
and then you can remove supported_channels from the sNodeHostap.

Another thing , 
notification->hostap().supported_channels is still used (line 2083 area)
use `get_hostap_supported_channels_string()` instead

common/beerocks/bwl/common/base_wlan_hal.cpp
```diff
@@ -31,6 +31,7 @@ base_wlan_hal::base_wlan_hal(HALType type, std::string iface_name, IfaceType ifa
     }
 
     // Initialize complex containers of the radio_info structure
+    m_radio_info.preferred_channels.resize(128 /* TODO: Get real value */);
```
tomereli: TODO should be accompanied with an issue URL which should fix it.

tomereli: Also, the commit message is really short.

Suggestion:

>Preparative commit.
Currently, the supported channels used in prplMesh per radio are the channels read from ACS (Automatic channel scan) report. This is wrong, the ACS report holds the **prefered** channels, not the supported channels.
The supported channels should be read from the driver via NL, and a new preferred channels vector should be added to hold the ACS results. This will be done in followup commits, for now, just add empty preferred_channels to the radio info structure.

tomereli: Not specifically related to this commit - but both the preferred channels and the supported channels are vectors - so why do you resize them to 128? If the number of elements is 128 - then just use `std::array`. This should be done in a separate, preparative commit.

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_common.h
```diff
@@ -378,19 +378,26 @@ typedef struct sNodeHostap {
     uint32_t vht_capability;
     uint8_t vht_mcs_set[beerocks::message::VHT_MCS_SET_SIZE];
     char driver_version[beerocks::message::WIFI_DRIVER_VER_LENGTH];
+    beerocks::message::sWifiChannel preferred_channels[beerocks::message::SUPPORTED_CHANNELS_LENGTH];
```
tomereli: The commit message does not follow the CONTRIBUTING.md.
Suggestion:

>btlvf: add preferred_channels to sNodeHostap
>
> Preparative commit.
> The supported_channels currently contains the ACS report channels, which are actually the preferred channels. Add preferred_channels to the sNodeHostap struct to be used instead of the currently supported_channels list.

2.

agent/src/beerocks/slave/son_slave_thread.cpp
tomereli: So this should be the first commit in this PR - rename supported_channels to preferred_channels. What you did here is simply unacceptable and VERY confusing.

tomereli (changes requested):
This PR is wrong and confusing as hell. You should build a new one which does this with the following commits:
1. rename supported_channels -> prefered_channels
2. change prefered_channels to use std::array
3. add supported_channels using nl

common/beerocks/bwl/common/base_wlan_hal.cpp
itayx: Changed the vector to an array for both `supported_channels` and `preferred_channels`

common/beerocks/tlvf/AutoGenerated/include/beerocks/tlvf/beerocks_message_common.h
itayx: Changed the commit messages to follow the instructions in `CONTRIBUTING.md`

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: Changed all ACS related variables and CMDUs to use `preferred_channels`

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.h
itayx: Fixed typos

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
itayx: Fixed typos

agent/src/beerocks/slave/son_slave_thread.cpp
itayx: Changed the all the ACS related variables to use `preferred_channels` instead

agent/src/beerocks/slave/son_slave_thread.cpp
itayx: Added a commit that refactors all the ACS related variables and messages to use `preferred_channels`

controller/src/beerocks/master/son_master_thread.cpp
itayx: Moved the supported_channels to outside the sNodeHostap
And changed all previews missuses of `supported_channels` to `preferred_channels`

agent/src/beerocks/slave/ap_manager_thread.cpp
itayx: Removed self-explanatory unneeded comments

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -43,7 +43,7 @@ enum eStructsConsts {
     VERSION_LENGTH                = 16,
     NODE_NAME_LENGTH              = 32,
     IFACE_NAME_LENGTH             = 32 + 4, //need extra 1 byte for null termination + alignment
-    SUPPORTED_CHANNELS_LENGTH     = 64,     //support upto # channels, every channel item is 32-bit
+    RADIO_CHANNELS_LENGTH         = 128,    //support upto # channels, every channel item is 32-bit
```
tomereli, line 46: 1. Why change to 128 by default?
2. why change at all? it should be changed to supported channels length and when you add the preferred channels length we will use a new enum.
3. The name is bad - NUM_SUPPORTED_CHANNELS and NUM_PREFERRED_CHANNELS is better.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -33,30 +33,28 @@ using namespace beerocks::net;
 /////////////////////////// Local Module Functions ///////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-static void copy_radio_supported_channels(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
-                                          beerocks::message::sWifiChannel supported_channels[])
+static void copy_radio_channels(std::vector<bwl::WiFiChannel> radio_channels,
```
tomereli: I think a better name now should be just `copy_channels` cause thats all it does:
```
static void copy_channels(std::vector<bwl::WiFiChannel> &from,
                                         beerocks::message::sWiFiChannel to[]);
```
(nitpick)

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1569,6 +1569,9 @@ void ap_manager_thread::handle_hostapd_attached()
                 beerocks::message::VHT_MCS_SET_SIZE, notification->params().vht_mcs_set);
 
     // Copy the channels supported by the AP
```
tomereli, line 1572:
```suggestion
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1551,8 +1551,8 @@ bool ap_wlan_hal_dwpal::read_supported_channels()
         }
     }
 
-    m_radio_info.preferred_channels.insert(m_radio_info.preferred_channels.begin(),
-                                           supported_channels.begin(), supported_channels.end());
+    m_radio_info.supported_channels.insert(m_radio_info.supported_channels.begin(),
```
tomereli, line 1554: This looks like a bug - isn't it?
```suggestion
    m_radio_info.supported_channels.insert(m_radio_info.supported_channels.begin(),
                                           supported_channels.begin(), supported_channels.end());
```

tomereli: Now that I look at it again - this is part of a functino which shouldn't exist in this commit - `bool ap_wlan_hal_dwpal::read_supported_channels()` - I'm confused, didn't the previous commit replace all occurrences of supported channels with preferred channels?

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -469,7 +469,7 @@ bool ap_wlan_hal_nl80211::read_supported_channels()
         }
     }
 
-    m_radio_info.preferred_channels.insert(m_radio_info.preferred_channels.begin(),
+    m_radio_info.supported_channels.insert(m_radio_info.supported_channels.begin(),
```
tomereli, line 472: Here as well - the `read_supported_channels()` should not exist in this commit.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -25,6 +25,9 @@ cACTION_APMANAGER_4ADDR_STA_JOINED:
  
 cACTION_APMANAGER_JOINED_NOTIFICATION:
   _type: class
+  supported_channels:
```
tomereli, line 28: This should be in a separate commit with the autogenerated files changes (nitpick)

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1512,6 +1512,7 @@ void ap_manager_thread::handle_hostapd_attached()
 {
     LOG(DEBUG) << "handling enabled hostapd";
 
+    ap_wlan_hal->read_supported_channels();
```
tomereli, line 1515: Why do we read the supported channels here? we anyway read it from NL so why every time we attach to hostapd? This should be done only once.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1527,7 +1528,7 @@ void ap_manager_thread::handle_hostapd_attached()
             usleep(ACS_READ_SLEEP_USC);
         }
     } else {
-        ap_wlan_hal->read_supported_channels();
+        ap_wlan_hal->update_preference_channels_from_supported_channels();
```
tomereli, line 1531: No - the logic of how we choose the channel preference should be abstracted by the BWL.
So instead we should have a `read_preferred_channels()` API in BWL which **internally** will use the radio supported channels **we read at boot** if ACS is disabled.
So - this whole commit should be removed, and replaced with a simple clear API to read preferred channels exactly like the API to read supported channels.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -474,6 +474,12 @@ bool ap_wlan_hal_nl80211::read_supported_channels()
     return true;
 }
 
+bool ap_wlan_hal_nl80211::update_preference_channels_from_supported_channels()
+{
+    m_radio_info.preferred_channels = m_radio_info.supported_channels;
```
tomereli, line 479: So you already do this here - so the API is very simple.. Just use `m_radio_info.supported_channels` if ACS is disabled.

tomereli (changes requested):
Better, but still a lot of comments, some major, so asking for re-review.

PR #1066, opened 2020-04-01
Add nl80211 get interface command

### LINKED ISSUE #1023: [TASK] Add NL80211_CMD_GET_INTERFACE command
Add a new method to nl80211_client to obtain information about a given interface through NL. This method can later be used to obtain interface index, type, address and SSID; as the `iw` command line tool does:

```
root@OpenWrt:~# iw wlan2.0 info
Interface wlan2.0
	ifindex 18
	wdev 0x100000002
	addr 02:9a:96:fb:59:22
	ssid prplMesh
	type AP
	wiphy 1
```

This command can also be used to find out if a given interface name is a wireless interface. It works both for radio interfaces and VAP interfaces.

mariomaz:
See #1023

### DIFF
--- a/common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
+++ b/common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
@@ -9,12 +9,26 @@
 #include "nl80211_client_dummy.h"
 
 #include <bcl/beerocks_utils.h>
+#include <bcl/network/network_utils.h>
 
 namespace bwl {
 
 nl80211_client_dummy::nl80211_client_dummy() {}
 
 nl80211_client_dummy::~nl80211_client_dummy() {}
+
+bool nl80211_client_dummy::get_interface_info(const std::string &interface_name,
+                                              interface_info &interface_info)
+{
+    interface_info.name  = interface_name;
+    interface_info.index = 1;
+    interface_info.addr  = beerocks::net::network_utils::mac_from_string("01:02:03:04:05:06");
+    interface_info.ssid  = "prplmesh";
+    interface_info.type  = 3; // NL80211_IFTYPE_AP
+    interface_info.wiphy = 2;
+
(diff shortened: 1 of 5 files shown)


tomereli:
> See #1023

Note that using `hub` command you can convert the issue to a PR from the command line:
`hub pull-request 1023`.

common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
```diff
@@ -9,13 +9,27 @@
...
 
 nl80211_client_dummy::~nl80211_client_dummy() {}
 
+bool nl80211_client_dummy::get_interface_info(const std::string &interface_name,
+                                              interface_info &interface_info)
+{
+    interface_info.name  = interface_name;
+    interface_info.index = 1;
+    interface_info.addr  = beerocks::net::network_utils::mac_from_string("01:02:03:04:05:06");
+    interface_info.ssid  = "prplmesh";
```
tomereli, line 22: This is a bit problematic. In dummy mode we update the ssid in the `m_radio_info.available_vaps[vap_id].ssid`, so this doesn't reflect that.
Actually I don't understand now why we need a dummy nl80211_client at all - the nl80211_client member is now in the respective BWL base class backend - `base_bwl_wlan_hal_dwapl` and `base_bwl_wlan_hal_nl80211`, but we don't use it in `base_bwl_wlan_hal_dummy` - and we just implement the APIs that we need without it in the dummy implementation.

Seeing how it is now implemented, I think the whole nl80211 client dummy implementation is redundant and potentially confusing.

tomereli (approved):
LGTM. My only comment is regarding the nl80211_dummy implementation which I think should be removed altogether. It can be a separate PR or the last commit in this one, if you agree with me.

rmelotte (approved):

PR #1068, opened 2020-04-01
Create sniffer as a separate module

arnout:
This is yet another spinoff of #1031.
It is here to allow @Galvien to rebase #1020 on top of it and put the new tshark features in the sniffer class.

Also, at @Galvien's request, make test_flows.py PEP8-compliant now.

### DIFF
--- a/tests/opts.py
+++ b/tests/opts.py
@@ -10,6 +10,7 @@ class opts:
     '''Static class that encodes the global options.'''
     verbose = False
     tcpdump = False
+    tcpdump_dir = ''
     stop_on_failure = False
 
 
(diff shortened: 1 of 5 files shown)


tests/opts.py
```diff
@@ -10,6 +10,7 @@ class opts:
     '''Static class that encodes the global options.'''
     verbose = False
     tcpdump = False
+    tcpdump_dir = ''
     stop_on_failure = False
```
Galvien, line 15: Is this common practice? while it's convenient to keep these options on a static class I would assume they would be stored in either a `tests.config` , a json file file or something of that sort.

Galvien (approved):

tomereli (approved):

tests/opts.py
arnout: A static class is a somewhat pythonesque way of organising global variables, yes. It has the advantage that you don't need the `global` keyword.

Issue #461, opened 2019-10-31
[BUG] tools: docker: the docker inspect returns a wrong IP


rmelotte:
At the moment, if no IP was specified when calling `run.sh`, a "random" IP is generated based on the docker network subnet. 
Because we don't specify the IP to docker itself when creating the container (the IP is passed to the entrypoint script, but not to docker through the `--ip` option), docker generates itself another one.

The docker-generated IP is never used by the container as the entrypoint script calls `ip addr flush eth0`.  This leads to a confusing situation where if you ask docker for the container IP (through `docker inspect`), it will return the wrong IP.

As a consequence, scripts that need the container IP (e.g. `test_flows.sh`) need to know the bridge name to be able to retrieve the IP "manually".

Here is a proposal to fix this:
- if no IP was given when calling `run.sh`, do not generate a "random" IP but let docker generate it
This would avoid the need to pass the IP to the entrypoint script, which can get the IP used by looking at `eth0`.
- if an IP is passed when calling `run.sh`, use `--ip` when calling `docker container run` . This will only work if the IP is on the same subnet as the docker network *and* the docker network was assigned a subnet manually (otherwise docker will refuse to use the IP; this might have changed with "recent" docker versions).

Letting docker generate an IP is probably a better idea anyway, since it will pick one that does not conflict with other docker containers (but it will not check that it does not overlap with non-docker networks).

PR #1077, opened 2020-04-01
Support running test flows on WSL

vitalybu:
The purpose of this PR is to add support for running the test flows on WSL.
Docker for Windows is currently unable to forward traffic from the host to Linux based containers.
To workaround this limitation, when running on WSL, we expose the UCC ports to the host machine.
Once the ports are exposed, the `test_flows.py` script opens the CAPI sockets to the localhost and is able to communicate with the listeners in the containers.

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -11,6 +11,7 @@ import os
 import re
 import subprocess
 import sys
+import platform
 import time
 from typing import Dict
 import json
@@ -176,8 +177,18 @@ class TestFlows:
         with open(os.path.join(self.installdir, 'config', config_file_name)) as config_file:
             ucc_port = re.search(r'ucc_listener_port=(?P<port>[0-9]+)', config_file.read()).group('port')
 
-        device_ip_output = self.docker_command(device, 'ip', '-f', 'inet', 'addr', 'show', self.bridge_name)
-        device_ip = re.search(r'inet (?P<ip>[0-9.]+)', device_ip_output.decode('utf-8')).group('ip')
+        # On WSL, connect to the locally exposed container port
+        if self.on_wsl:
+            published_port_output = subprocess.check_output(
+                ["docker", "port", device, ucc_port]).decode('utf-8').split(":")
+            device_ip = published_port_output[0]
+            ucc_port = int(published_port_output[1])
+        else:
+            device_ip_output = self.docker_command(
+                device, 'ip', '-f', 'inet', 'addr', 'show', self.bridge_name)
+            device_ip = re.search(
+                r'inet (?P<ip>[0-9.]+)', device_ip_output.decode('utf-8')).group('ip')
+
         return UCCSocket(device_ip, ucc_port)
 
     def init(self):
@@ -186,6 +197,7 @@ class TestFlows:
         self.gateway = 'gateway-' + self.opts.unique_id
         self.repeater1 = 'repeater1-' + self.opts.unique_id
         self.repeater2 = 'repeater2-' + self.opts.unique_id
+        self.on_wsl = "microsoft" in platform.uname()[3].lower()
         if not self.opts.skip_init:
             self.tcpdump_start()
             try:
@@ -223,6 +235,14 @@ class TestFlows:
     def _check_log_internal(self, device: str, program: str, regex: str, start_line: int):
         '''Search for regex in logfile for program on device.'''
         logfilename = os.path.join(self.rootdir, 'logs', device, 'beerocks_{}.log'.format(program))
+
+        # WSL doesn't support symlinks on NTFS, so resolve the symlink manually
+        if self.on_wsl:
+            logfilename = os.path.join(
+                self.rootdir, 'logs', device,
+                subprocess.check_output(["tail", "-2", logfilename]).
+                decode('utf-8').rstrip(' \t\r\n\0'))
+
         with open(logfilename) as logfile:
             for (i, v) in enumerate(logfile.readlines()):
                 if i <= start_line:
--- a/tests/test_gw_repeater.sh
+++ b/tests/test_gw_repeater.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 ###############################################################
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 # Copyright (c) 2019 Tomer Eliyahu (Intel)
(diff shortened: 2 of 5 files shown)


tools/docker/run.sh
```diff
@@ -107,7 +110,7 @@ main() {
                 -e INSTALL_DIR=${installdir}
                 --privileged
                 --network ${NETWORK}
-                --expose ${PORT}
+                --expose ${PORT} ${PUBLISH}
```
tomereli: Any reason why not to always publish all ports?
What is the difference between exposing port and publishing it?

tomereli (approved):
@arnout or @rmelotte should also review this.

tools/docker/run.sh
vitalybu: You don't always want to publish the ports to the host.
Sometimes you only want to expose certain ports for container-to-container communication.

From the official docker documentation:

> The EXPOSE instruction defines the initial incoming ports that provide services. These ports are available to processes inside the container. An operator can use the --expose option to add to the exposed ports.
> 
> To expose a container’s internal port, an operator can start the container with the -P or -p flag. The exposed port is accessible on the host and the ports are available to any client that can reach the host.

tests/test_flows.py
```diff
@@ -223,6 +234,14 @@ def init(self):
     def _check_log_internal(self, device: str, program: str, regex: str) -> bool:
         '''Search for regex in logfile for program on device.'''
         logfilename = os.path.join(self.rootdir, 'logs', device, 'beerocks_{}.log'.format(program))
+
+        # WSL doesn't support symlinks on NTFS, so resolve the symlink manually
```
rmelotte, line 239: What does WSL use NTFS for? (I know nothing about WSL)

tools/docker/runner/Dockerfile
```diff
@@ -39,4 +39,4 @@ ADD start-controller /usr/bin
 ADD start-agent /usr/bin
 ADD start-controller-agent /usr/bin
 
-ENTRYPOINT ["/root/entrypoint.sh"]
+ENTRYPOINT ["/bin/bash", "/root/entrypoint.sh"]
```
rmelotte: Why do you want to explicitly call bash instead of relying on the shebang?

tools/docker/runner/entrypoint.sh
```diff
@@ -14,6 +14,12 @@ run() {
 bridge_ip="$1"; shift
 base_mac="$1"; shift
 
+# If the requested bridge ip is 0.0.0.0, use the ip address that
+# was allocated by the daemon to this container (from eth0)
+if [[ ${bridge_ip} == 0.0.0.0* ]]; then
+    bridge_ip="$(ip addr show dev eth0 | grep inet | awk '{print $2}')"
+fi
```
rmelotte, line 21: I proposed to do this a while ago, so if you feel like it this could be done even for linux : https://github.com/prplfoundation/prplMesh/issues/461

tools/docker/run.sh
```diff
@@ -64,7 +65,8 @@ main() {
             -n | --name)        NAME="$2"; shift; shift ;;
             -u | --unique-id)   UNIQUE_ID="$2"; shift; shift ;;
             -t | --tag)         TAG=":$2"; shift; shift ;;
-            -p | --port)        PORT=":$2"; shift; shift ;;
+            -p | --port)        PORT="$2"; shift; shift ;;
+            -P | --publish)     PUBLISH="-P"; shift ;;
```
rmelotte: From `man docker-run`:
>       -P, --publish-all=true|false
>       Publish all exposed ports to random ports on the host interfaces. The default is false.

Are you sure you want to expose them to random ports on the host?

If not you could use this syntax to control what the ports are mapped to:
>       -p, --publish ip:[hostPort]:containerPort | [hostPort:]containerPort
>          Publish a container's port, or range of ports, to the host.

**More importantly**: I don't think you need them to be published externally, do you?
If not, publish them to "localhost" on the host instead of "0.0.0.0" (which is what "-P" does, at least under Linux).

tests/test_gw_repeater.sh
```diff
@@ -27,6 +27,35 @@ usage() {
...
+check_wsl() {
+    # Do nothing for non-WSL environments
+    if ! grep -q Microsoft /proc/version; then return; fi
+
+    status "Running in WSL"
+    
+    # Read GW & Repeater UCC ports
+    local GW_UCC_PORT=$(grep ucc_listener_port \
+    ${rootdir}/build/install/config/beerocks_controller.conf | \
+    awk -F'[= ]' '{ print $2 }')
```
rmelotte: I know there are already many shellcheck issues in this script, but let's not introduce more
```suggestion
    local GW_UCC_PORT
    GW_UCC_PORT="$(grep ucc_listener_port \
    "${rootdir}/build/install/config/beerocks_controller.conf" | \
    awk -F'[= ]' '{ print $2 }')"
```

tests/test_gw_repeater.sh
```diff
@@ -27,6 +27,35 @@ usage() {
...
+
+    local RP_UCC_PORT=$(grep ucc_listener_port \
+    ${rootdir}/build/install/config/beerocks_agent.conf | \
+    awk -F'[= ]' '{ print $2 }')
```
rmelotte, line 51: Same here

rmelotte (approved):
Not requesting changes to not block you, but I think you'll want to at least address the comment on using `-P` with `docker run`.

tests/test_flows.py
vitalybu: Actually it's a combination of issues. WSL doesn't use NTFS directly, it does implement a somewhat Linux compatible filesystem.
The problem is with Docker for Windows. There's a known bug that prevents the daemon (that actually runs on a Hyper-V VM) from mounting non-NTFS volumes into the containers.
Since we map the `build` and `tmp` folders into the `prplmesh-runner`, all the logs are eventually created on an NTFS filesystem that doesn't support symlinks.

tools/docker/runner/Dockerfile
vitalybu: Not sure actually :)
I'll revert and update the `entrypoint.sh` to `#!/bin/bash`

tools/docker/runner/entrypoint.sh
vitalybu: Maybe we can skip generating IP addresses at all.
Not sure why we need if the daemon does it anyway.

tools/docker/run.sh
vitalybu: Publish just makes these ports accessible to the host.
I do prefer random ports (which we can later read by `docker port`) instead of managing them myself.

tests/test_flows.py
rmelotte: Ok, thanks for the explanation :-)

tools/docker/runner/entrypoint.sh
rmelotte: Yes exactly.

tools/docker/run.sh
rmelotte: `--publish` (which is not the option you used) will make it accept connections from whatever IP you specified.
`--publish-all` will make it accept connections from anywhere.

Still from `man docker-run`:
>When set to true publish all exposed ports to the host interfaces. The default is false. If the operator uses -P (or -p) then Docker will make the exposed port accessible on the host and
       **the ports will be available to any client that can reach the host**. When using -P, Docker will bind any exposed port to a random port on the host within an ephemeral port range defined by
>       /proc/sys/net/ipv4/ip_local_port_range. To find the mapping between the host ports and the exposed ports, use docker port(1).

In other words, you _have_ to use `--publish 127.0.0.1::<capi_port>` unless you want the CAPI listener to be available to anyone on your local network.

(you will still have random ports on the host if you specify the IP and the container port only)

tests/test_flows.py
ghost: Actually, NTFS does support symlinks (since NTFS version 3.0, which was released together with Windows 2000). Windows API support (e.g. OpenFile) for symlinks was added to Windows in Windows Vista, though linux (using ntfs-3g) could create these symlinks on NTFS even before then.
You can create those symlinks on Windows with the `mklink` command.
Annoyingly, IIRC you always need administrator privileges to create symlinks on windows though.

ghost: Actually, it seems my information is outdated: https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/

vitalybu: Symlinks work fine when created from Windows using `mklink`, but when created in WSL, they appear as regular files on NTFS.

tools/docker/run.sh
vitalybu: Updated the script to allow publishing ports on the localhost only.

tests/test_gw_repeater.sh
vitalybu: Done. Also fixed all shellcheck issues while I'm at it.

PR #1078, opened 2020-04-02
[rdkb client steering BML] consider hwm SNR 0 as soft block disabled for  

kantera800:
Using rdkb client steering BMLs, 3rd party application can set high probe SNR value to 0 in order to
disable the soft block and to allow the connection.

Currently, when it's set to 0, no client can connect.
Solution is when snrProbeHWM=0, we will set sta_softblock remove=1


controller/src/beerocks/bml/rdkb/internal/bml_rdkb_internal.cpp
```diff
@@ -132,7 +132,9 @@ int bml_rdkb_internal::steering_client_set(uint32_t steeringGroupIndex, const BM
         request->config().snrHighXing      = config->snrHighXing;
         request->config().snrLowXing       = config->snrLowXing;
         request->config().authRejectReason = config->authRejectReason;
-        request->remove()                  = 0;
+        if (request->config().snrProbeHWM > 0) {
+            request->remove() = 0;
+        }
```
morantr, line 137: Why specifically snrProbeHWM? Why not other parameters?
I don't understand why it is needed, the API says that if assigning nullptr to the argument `BML_STEERING_CLIENT_CONFIG *config` it means a request for removal.

yaniv-yitzhak: This is how the API is used, softblock is removed when *config is NULL OR if snrProbeHWM=0.
This is NOT a WA, we need to changes the API description as well.

yaniv-yitzhak (approved):

controller/src/beerocks/bml/rdkb/internal/bml_rdkb_internal.cpp
kantera800: > This is how the API is used, softblock is removed when *config is NULL OR if snrProbeHWM=0.
> This is NOT a WA, we need to changes the API description as well.

Please provide new API description and I will add it as part of this pr

yaniv-yitzhak: "   * Call this function to add/modify/remove per-client configuration config of client_mac.
     * To remove a client configuration call with NULL as config or snrProbeHWM=0."

morantr (approved):

adam1985d (approved):

PR #1079, opened 2020-04-02
bwl: dwpal: fix missing channel bandwidth

itayx:
As part of the DCS's channel scan, the channel bandwidth for each
neighboring AP is retrieved from the NL message.
The value is stored as part of the Information Elements or IES
in the vht & ht fields.

During scan dump, add check to both Information Elements and IES.

### DIFF
--- a/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
@@ -394,7 +394,7 @@ static bool translate_nl_data_to_bwl_results(sChannelScanResults &results,
     }
 
     //get information elements from information-elements-buffer or from beacon
-    if (bss[NL80211_BSS_BEACON_IES]) {
+    if (bss[NL80211_BSS_BEACON_IES] || bss[NL80211_BSS_INFORMATION_ELEMENTS]) {
         enum nl80211_bss ies_index = (bss[NL80211_BSS_INFORMATION_ELEMENTS])
                                          ? NL80211_BSS_INFORMATION_ELEMENTS
                                          : NL80211_BSS_BEACON_IES;


common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -50,13 +50,11 @@ nl80211_client_impl::nl80211_client_impl(std::unique_ptr<nl80211_socket> socket)
...
-        bool width_40;
-        bool width_80;
-        bool width_80_80;
-        bool width_160;
-    } ctx{};
+    int last_band    = -1;
+    bool width_40    = false;
+    bool width_80    = false;
+    bool width_80_80 = false;
+    bool width_160   = false;
```
vitalybu: Any reason to define these here?
As far as I can see, they are used only when processing the netlink response.

common/beerocks/bwl/shared/netlink_socket.cpp
```diff
@@ -101,9 +101,7 @@ bool netlink_socket::send_receive_msg(std::function<bool(struct nl_msg *msg)> ms
     // function that calls the capturing lambda function.
     static thread_local std::function<int(struct nl_msg * msg, void *arg)> nl_handler_cb_wrapper;
     nl_handler_cb_wrapper = [&](struct nl_msg *msg, void *arg) -> int {
-        if (!msg_handle(msg)) {
-            LOG(ERROR) << "User's netlink handler function failed!";
-        }
+        msg_handle(msg);
         return NL_SKIP;
```
vitalybu: Why not return `NL_OK` in case of error?
A handler can also return `false` if it doesn't know how to "handle" the incoming message at all.
Maybe there's another registered handler that will manage to handle the incoming message.

vitalybu (changes requested):
Beside the review comments, there's a duplicated commit and a merge of another branch.

tomereli (changes requested):
Please remove the merge commit.

common/beerocks/bwl/shared/netlink_socket.cpp
itayx: This was added to my PR but is part of #1053

common/beerocks/bwl/shared/nl80211_client_impl.cpp
itayx: This was added to my PR but is part of #1053

vitalybu (approved):

tomereli (approved):

PR #1091, opened 2020-04-03
hotfix:test_channel_selection

VladyslavTupikin:
Corrected AL-MAC in channel selection request.

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -537,7 +537,7 @@ class TestFlows:
                     0x8006,
                     tlv(0x8B, 0x005F, '{} {}'.format(self.mac_repeater1_wlan0, payload_wlan0)),
                     tlv(0x8D, 0x0007, '{} 0x{:2x}'.format(self.mac_repeater1_wlan0, tp20dBm)),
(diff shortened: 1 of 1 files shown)


tomereli (approved):

rmelotte:
I didn't look into in in details but: how is it possible that the test doesn't fail before this commit?
If sending the test payload to the wrong MAC address doesn't change the test result at all, it sounds like there is something wrong with the test itself.

@VladyslavTupikin any idea?

arnout:
> I didn't look into in in details but: how is it possible that the test doesn't fail before this commit?
> If sending the test payload to the wrong MAC address doesn't change the test result at all, it sounds like there is something wrong with the test itself.

There is something wrong with the test itself. We currently don't check at all if the message is taken into account by the agent, i.e. if it switches channels to one of the channels specified in the message. So a missing message for one of the radios is not noticed.

What this mistake does expose is that we don't detect that there are two conflicting channel preference TLVs. But I don't know if we need to handle that case gracefully anyway. We don't crash, which IMO is sufficient.

arnout:
Oh and BTW I accidentally fixed this while rebasing #1031, so maybe we can close this.

VladyslavTupikin:
Fixed in #1031

PR #1093, opened 2020-04-05
Feature/channel pool validation

itayx:
Currently the DCS's set_channel_scan_pool uses a static list of supported channels for verifications to see if the channels provided in the channel pool are supported.

The DB has a list known as hostap_supported_channels, this is filled with channels provided by the basic radio capabilities

The DCS's set_channel_scan_pool need to validate it's channel pool against this supported channels list

### DIFF
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -2291,6 +2291,24 @@ int db::get_channel_scan_dwell_time_msec(const sMacAddr &mac, bool single_scan)
         .dwell_time_msec;
 }
 
+bool db::is_channel_scan_pool_supported(const sMacAddr &mac,
+                                        const std::unordered_set<uint8_t> &channel_pool)
+{
+    auto supported_channels = get_hostap_supported_channels(network_utils::mac_to_string(mac));
+    for (const auto &channel : channel_pool) {
+        auto found_channel =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&channel](const beerocks::message::sWifiChannel &supported_channel) {
+                             return supported_channel.channel == channel;
+                         });
+        if (found_channel == supported_channels.end()) {
+            LOG(ERROR) << "channel #" << int(channel) << " is not supported";
+            return false;
+        }
+    }
+    return true;
+}
+
 bool db::set_channel_scan_pool(const sMacAddr &mac, const std::unordered_set<uint8_t> &channel_pool,
                                bool single_scan)
 {
@@ -2300,22 +2318,9 @@ bool db::set_channel_scan_pool(const sMacAddr &mac, const std::unordered_set<uin
         return false;
     }
 
-    //Validate new dcs channel pool with hostap_supported_channels list
-    //This is a workaround
-    static uint8_t supported_channels_array[] = {
-        1,  2,   3,   4,   5,   6,   7,   8,   9,   10,  11,  36,  40,  44,  48,  52,  56,  60,
-        64, 100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144, 149, 153, 157, 161, 165};
-    static size_t supported_channels_array_size =
-        sizeof(supported_channels_array) / sizeof(uint8_t);
-    std::set<uint8_t> supported_channels(supported_channels_array,
-                                         supported_channels_array + supported_channels_array_size);
-    for (auto channel : channel_pool) {
-        if (supported_channels.find(channel) == supported_channels.end()) {
-            //even if one channel is not supported reject all pool
-            LOG(ERROR) << "channel #" << int(channel)
-                       << " is invalid, setting channel pool failed !";
-            return false;
-        }
+    if (!is_channel_scan_pool_supported(mac, channel_pool)) {
+        LOG(ERROR) << "setting channel pool failed, one of the channels is not supported!";
+        return false;
     }
 
     (single_scan ? hostap->single_scan_config : hostap->continuous_scan_config).channel_pool =
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -513,6 +513,17 @@ public:
                                bool single_scan);
 
     /**
+     * @brief Validate the channel scan pool
+     * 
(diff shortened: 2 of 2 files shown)


controller/src/beerocks/master/db/db.cpp
```diff
@@ -2299,16 +2302,14 @@ bool db::set_channel_scan_pool(const sMacAddr &mac, const std::unordered_set<uin
...
-                                         supported_channels_array + supported_channels_array_size);
+    auto supported_channels = get_hostap_supported_channels(network_utils::mac_to_string(mac));
     for (auto channel : channel_pool) {
-        if (supported_channels.find(channel) == supported_channels.end()) {
+        auto invalid_channel =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                             [&channel](const beerocks::message::sWifiChannel &supported_channel) {
+                                 return supported_channel.channel == channel;
+                             });
+        if (invalid_channel == supported_channels.end()) {
```
kantera800: please use separate method instead  `is_channel_scan_pool_supported()` 
also, no need to scan all pool , just  return false for the first channel that is not supported

itayx: added `is_channel_scan_pool_supported()`

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2289,6 +2290,24 @@ int db::get_channel_scan_dwell_time_msec(const sMacAddr &mac, bool single_scan)
         .dwell_time_msec;
 }
 
+bool db::is_channel_scan_pool_supported(const sMacAddr &mac,
+                                        const std::unordered_set<uint8_t> &channel_pool)
+{
+    auto supported_channels = get_hostap_supported_channels(network_utils::mac_to_string(mac));
+    for (auto channel : channel_pool) {
```
CoralMalachi:
```suggestion
    for (const auto &channel : channel_pool) {
```

As far as I understand, you are only reading the `channel` value and you have no attempt to modify it, so it preferred to make it a read-only value.

CoralMalachi (approved):

controller/src/beerocks/master/db/db.cpp

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2289,6 +2290,24 @@ int db::get_channel_scan_dwell_time_msec(const sMacAddr &mac, bool single_scan)
...
+{
+    auto supported_channels = get_hostap_supported_channels(network_utils::mac_to_string(mac));
+    for (const auto &channel : channel_pool) {
+        auto found_channel =
+            std::find_if(supported_channels.begin(), supported_channels.end(),
+                         [&channel](const beerocks::message::sWifiChannel &supported_channel) {
+                             return supported_channel.channel == channel;
+                         });
+        if (found_channel == supported_channels.end()) {
+            LOG(ERROR) << "channel #" << int(channel) << " is not supported";
```
rmelotte, line 2305: To me it's a bit weird that a function called `is_channel_scan_pool_supported` logs something (especially an error) if a channel is not supported. Maybe it would be best to leave that to the caller (or to change the level to INFO or DEBUG).

Aren't there any case where you would want to call `is_channel_scan_pool_supported` without logging an error?

(nitpick)

controller/src/beerocks/master/db/db.h
```diff
@@ -512,6 +512,17 @@ class db {
     bool set_channel_scan_pool(const sMacAddr &mac, const std::unordered_set<uint8_t> &channel_pool,
                                bool single_scan);
 
+    /**
+     * @brief Validate the channel scan pool
+     * 
+     * @param mac:          MAC address of radio
+     * @param channel_pool: Channel pool of channel scan
+     * @return true if pool is valid
+     * @return false if pool is invalid
```
rmelotte, line 521: To me the function name is pretty explicit about what is being done already, but the documentation is so vague that it doesn't bring anything useful. 
What kind of validation does it do?

For example:

```suggestion
    /**
     * @brief Validate the channel scan pool.
     * 
     * A pool is considered valid if all the channels it contains are supported by the radio.
     * 
     * @param mac:          MAC address of the radio
     * @param channel_pool: Channel pool to validate
     * @return true if channel_pool is valid, false otherwise.
```

(nitpick)

rmelotte (approved):
Only nitpicks :+1:

controller/src/beerocks/master/db/db.cpp
itayx: I completely understand
And I think I'll change it to `INFO`
But I still want the log, that way there is a log of which sent channel isn't supported
Since the function doesn't return the unsupported channel, the only way to know which is via the log

kantera800 (approved):

kantera800:
approved , but don't merge it yet.
**Wait for the NO_IR issue to be solved.**

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1248,8 +1248,9 @@ std::string db::get_hostap_supported_channels_string(const std::string &radio_ma
     for (const auto &val : supported_channels) {
         if (val.channel > 0) {
             os << " ch = " << int(val.channel) << " | dfs = " << int(val.is_dfs_channel)
-               << " | tx_pow = " << int(val.tx_pow) << " | noise = " << int(val.noise)
-               << " [dbm] | bss_overlap = " << int(val.bss_overlap) << std::endl;
+               << " | tx_pow = " << int(val.tx_pow) << " | noise = " << int(val.noise) << " [dbm]"
+               << " | bss_overlap = " << int(val.bss_overlap)
+               << " | bw = " << int(val.channel_bandwidth) << std::endl;
```
tomereli: This should be a separate commit (nitpick)

tomereli (approved):

controller/src/beerocks/master/db/db.cpp
itayx: This was moved to #1176 
I will remove it from here once that PR is merged

controller/src/beerocks/master/db/db.h
itayx: Updated the doxygen documentation

PR #1096, opened 2020-04-06
Dev/bap wps onboarding

tomereli:
Add support bAPs for allowing other agents to onboard the prplMesh agent using wireless backhaul:
- Use Intel hostapd mesh mode instead of multi_ap for VAPs configuration
- Add WPS PBC implementation using hostapd upstream support
- Remove backhaul vaps unused code

PR in prplwrt iwlwav - https://git.prpl.dev/prplmesh/iwlwav/-/merge_requests/12


agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -169,6 +169,27 @@ bool agent_ucc_listener::send_cmdu_to_destination(ieee1905_1::CmduMessageTx &cmd
...
+    }
+    auto method = params.find("wpsconfigmethod");
+    if (method == params.end()) {
+        err_string = "missing method parameter";
+        return false;
+    }
+    if (method->second != "PBC") {
+        err_string = "invalid WpsConfigMethod, supporting only PBC";
+        return false;
+    }
```
arnout: (nitpick) I think it's better to do this in beerocks_ucc_listener. We do that for the other commands as well.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
```diff
@@ -131,3 +131,6 @@ cACTION_BACKHAUL_CLIENT_DISCONNECTED_NOTIFICATION:
   iface_mac: sMacAddr
   client_mac: sMacAddr
   bssid: sMacAddr
+
+cACTION_BACKHAUL_START_WPS_PBC_REQUEST:
+  _type: class
```
arnout: Newline at end of file.

More importantly though: the band will definitely need to be part of this message. Or perhaps the radio_uid. Or the bssid. For sure, *something* :-)

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -169,6 +169,27 @@ bool agent_ucc_listener::send_cmdu_to_destination(ieee1905_1::CmduMessageTx &cmd
...
 }
 
+bool agent_ucc_listener::handle_start_wps_registration(
+    std::unordered_map<std::string, std::string> &params, std::string &err_string)
+{
+    auto band = params.find("band");
+    if (band == params.end()) {
+        err_string = "missing band parameter";
+        return false;
+    }
```
arnout: In my version of the test plan, there is an `interface` parameter, not a `band` parameter...

But... This is f**ing marvelous... UCC indeed uses the band parameter, not the interface parameter. So the test plan is apparently just informational :-(

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3438,4 +3438,30 @@ bool backhaul_manager::get_neighbor_links(
     return true;
 }
 
+bool backhaul_manager::start_wps_pbc(bool is_5ghz)
```
arnout: I would like to avoid (new) boolean values for the band, because at some point we might support 60GHz, and I'm not sure if 6GHz is going to be treated as a separate band or not.

So, prefer to use an enum or maybe even bitmask here.

That said, I think this `band` is just an artefact of UCC. I think in the high-level API, we'll want to be able to start PBC either on all radios or on a single radio, or maybe a list of radios. So I think it's better that we resolve the radio(s) in the ucc listener and that backhaul_manager::start_wps_pbc takes the radio(s) as argument.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -186,8 +186,7 @@ bool agent_ucc_listener::handle_start_wps_registration(
         err_string = "invalid WpsConfigMethod, supporting only PBC";
         return false;
     }
-    // TODO trigger WPS PBC on the selected band
-    return true;
+    return m_backhaul_manager_ctx.start_wps_pbc(band->second == "5G");
```
arnout: If it returns false, err_string should be set to something.

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
```diff
@@ -346,6 +346,15 @@ class ap_wlan_hal : public virtual base_wlan_hal {
...
     virtual bool generate_connected_clients_events() = 0;
+
+    /*!
+     * Start WPS PBC procedure.
+     * If backhaul VAP is configured (system operational),
+     * start WPS PBC on the active bAP, otherwise, start it on the bSTA. 
+     * 
+     * @return true on success or false on error
+     */
+    virtual bool start_wps_pbc() = 0;
```
arnout: In general, it is possible (and likely) that WPS is started on several radios simultaneously. In that case, we need to make sure that they use the same UUID. I think the only way to do that is to generate the UUID from prplmesh and pass it here.

Note that we don't need to implement generating the UUID yet - that would be an M3 feature. However, for the bwl it is better to have the interface as complete as possible already from the start - bwl changes are annoying for integrators. So, please allow a UUID as an optional parameter here.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -342,6 +342,12 @@ bool ap_wlan_hal_dummy::get_vap_enable(const std::string &iface_name, bool &enab
 
 bool ap_wlan_hal_dummy::generate_connected_clients_events() { return true; }
 
+bool ap_wlan_hal_dummy::start_wps_pbc()
+{
+    // TODO - simulate client connected?
```
arnout: For dummy, the minimal implementation is to write something to the log. Then you don't need a todo.

arnout (changes requested):
Looks good to me so far. I think actually this is enough for a PR to get merged right away, even if the dwpal implementation is incomplete or doesn't work yet.

That said, I would like to re-review the final version, so changes requested.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_backhaul.yaml
tomereli: Ehmm, why? The backhaul manager sends this message to a single radio agent, the one which should start WPS, so radio_uid is not needed.

tomereli: Fixed the newline.
Add radio id and bssid on which to trigger wps.

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
tomereli: I don't understand - the hostapd API doesn't support adding UUID, so where will it be used? Also what is the UUID type? I think this should be done in a separate PR anyway.

tomereli:
Re-opened since with this, autoconfiguration seems to work on the backhaul VAP, compared to https://github.com/prplfoundation/prplMesh/pull/1168 which fails. Will rebase this PR to see if the reason is something with the recent prplMesh changes.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3592,4 +3592,45 @@ bool backhaul_manager::add_ap_he_capabilities(const sRadioInfo &radio_info)
...
+    auto it =
+        std::find_if(slaves_sockets.begin(), slaves_sockets.end(),
+                     [&](std::shared_ptr<sRadioInfo> slave) { return slave->freq_type == freq; });
+
+    if (it == slaves_sockets.end()) {
+        LOG(ERROR) << "couldn't find slave for freq " << int(freq);
+        return std::string();
+    }
+
+    auto slave = *it;
```
vitalybu, line 3606: Why do you need another variable in the way?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -47,6 +47,22 @@ class backhaul_manager : public btl::transport_socket_thread {
 
     // For agent_ucc_listener
     const std::string &get_controller_bridge_mac() { return controller_bridge_mac; }
+    /**
+     * @brief get radio mac (ruid) of registered slave based on frequency type
+     * 
+     * @param freq radio frequency to search
+     * @return radio mac of the found slave if found, otherwise empty string
+     */
+    const std::string freq_to_radio_mac(enum eFreqType freq) const;
```
vitalybu, line 56:
```suggestion
    const std::string freq_to_radio_mac(eFreqType freq) const;
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -946,10 +875,34 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+            return false;
+        }
+        const auto &vaps = ap_wlan_hal->get_radio_info().available_vaps;
+        auto it          = std::find_if(
+            vaps.begin(), vaps.end(), [&](const std::pair<int, bwl::VAPElement> &element) {
+                return network_utils::mac_from_string(element.second.mac) == request->bssid();
+            });
+        if (it == vaps.end()) {
+            LOG(ERROR) << "Failed to find VAP with bssid " << request->bssid();
+            return false;
```
vitalybu, line 897: I don't think you should return `false` here (it's for critical errors only).
If there's no response message, just `break`.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -946,10 +875,34 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        }
+        auto vap_id   = it->first;
+        auto vap_name = utils::get_iface_string_from_iface_vap_ids(
+            ap_wlan_hal->get_radio_info().iface_name, vap_id);
+
+        ap_wlan_hal->start_wps_pbc(vap_name);
```
vitalybu, line 903: This call can also fail

common/beerocks/bwl/econet/ap_wlan_hal_econet.h
```diff
@@ -43,8 +43,9 @@ class ap_wlan_hal_dummy : public base_wlan_hal_dummy, public ap_wlan_hal {
     virtual bool sta_deauth(int8_t vap_id, const std::string &mac, uint32_t reason = 0) override;
     virtual bool sta_bss_steer(const std::string &mac, const std::string &bssid, int chan,
                                int disassoc_timer, int valid_int) override;
-    virtual bool update_vap_credentials(
-        std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list) override;
+    virtual bool
+    update_vap_credentials(std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list,
```
vitalybu, line 47: Forgot the `start_wps_pbc` override:
```c++
virtual bool start_wps_pbc(const std::string &iface_name) override;
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,8 +934,99 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
     return true;
 }
 
+static bool set_vap_multiap_mode(std::vector<std::string> &vap_hostapd_config,
+                                 WSC::eWscVendorExtSubelementBssType bss_type,
+                                 std::list<bwl::backhaul_vap_list_element_t> &backhaul_vaps_list)
+{
+    std::string bssid, ifname;
```
vitalybu, line 941: (nitpick)
```suggestion
    std::string bssid;
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,8 +934,99 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    if (!hostapd_config_get_value(vap_hostapd_config, "bssid", bssid)) {
+        LOG(ERROR) << "Failed to get VAP BSSID";
+        return false;
+    }
+    if (!hostapd_config_get_value(vap_hostapd_config, "bss", ifname)) {
```
vitalybu, line 946: (nitpick)
```suggestion

    std::string ifname;
    if (!hostapd_config_get_value(vap_hostapd_config, "bss", ifname)) {
```

tomereli:
Obsolete - replaced by #1168

PR #1102, opened 2020-04-06
DCS: Feature/add scan trigger response error handling in dcs task

### LINKED ISSUE #1135: [BUG] Controller: DCS task doesn't handle scan-trigger-response with error
**Describe the bug** -
The DCS task triggers a scan and waits for the SCAN_TRIGGERED event.
If the operation of triggering the scan fails - the task doesn't change state to scan-abort (and then IDLE and retry scan) immediately.
Instead, the task handles a SCAN_TRIGGERED timeout error when the timeout is reached.

## Steps to reproduce

For easy reproduce:
change monitor_thread implementation to always return an error (and not trigger the scan).
build, deploy and trigger a scan:
`/opt/prplmesh/bin/beerocks_cli -c "bml_start_dcs_single_scan <ap-2.4-mac> 30 1,2,3,4,5,6,7,8,9,10,11"`

## Expected behavior

DCS task should handle failure to trigger the scan and not timeout of the scan-triggered event

adam1985d:
Add scan-trigger-response with failure handling for the DCS task.

In the case where scan-trigger operation failed, the DCS task would only fail due to timeout on the scan-triggered event, even though the failure was already received by the son_master in the ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE CMDU.

Added SCAN_TRIGGER_FAILED event to DCS task and in the case of failure to trigger a scan, the son_manager will push that event to the task. The DCS task will abort the scan and set the last_scan_error_code to CHANNEL_SCAN_INTERNAL_FAILURE.

Fixes #1135 

Signed-off-by: Adam Dov adamx.dov@intel.com

### DIFF
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -3551,6 +3551,33 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
     }
     case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE: {
         LOG(TRACE) << "ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE for mac " << hostap_mac;
+        auto response =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE>();
+        if (!response) {
+            LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE failed";
+            return false;
+        }
+        //get the mac from hostap_mac
+        auto radio_mac = network_utils::mac_from_string(hostap_mac);
+
+        if (!response->success()) {
+            LOG(ERROR) << "failed to trigger scan on radio(" << radio_mac
+                       << "): sending notification to DCS task";
+
+            dynamic_channel_selection_task::sScanEvent new_event;
+            new_event.radio_mac = radio_mac;
+            auto taskId         = database.get_dynamic_channel_selection_task_id(radio_mac);
+            if (taskId == -1) {
+                LOG(ERROR) << "no task for the requested mac (" << radio_mac << ") could be found!";
+                break;
+            }
+
+            tasks.push_event(taskId,
+                             int(dynamic_channel_selection_task::eEvent::SCAN_TRIGGER_FAILED),
+                             static_cast<void *>(&new_event));
+            break;
+        }
         break;
     }
     case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE: {
@@ -3574,7 +3601,7 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
         new_event.radio_mac = radio_mac;
         auto taskID         = database.get_dynamic_channel_selection_task_id(radio_mac);
         if (taskID == -1) {
-            LOG(ERROR) << "no task for the requesed mac (" << radio_mac << ") could be found!";
+            LOG(ERROR) << "no task for the requested mac (" << radio_mac << ") could be found!";
         } else {
             tasks.push_event(taskID, (int)dynamic_channel_selection_task::eEvent::SCAN_TRIGGERED,
                              (void *)&new_event);
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -3558,18 +3558,18 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
             LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE failed";
             return false;
         }
-        //get the mac from hostap_mac
         auto radio_mac = network_utils::mac_from_string(hostap_mac);
 
         if (!response->success()) {
-            LOG(ERROR) << "failed to trigger scan on radio(" << radio_mac
+            LOG(ERROR) << "Failed to trigger scan on radio (" << radio_mac
                        << "): sending notification to DCS task";
 
             dynamic_channel_selection_task::sScanEvent new_event;
             new_event.radio_mac = radio_mac;
             auto taskId         = database.get_dynamic_channel_selection_task_id(radio_mac);
             if (taskId == -1) {
-                LOG(ERROR) << "no task for the requested mac (" << radio_mac << ") could be found!";
+                LOG(ERROR) << "No DCS task for the requested mac (" << radio_mac
+                           << ") is currently running!";
                 break;
             }
 
@@ -3601,7 +3601,8 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
         new_event.radio_mac = radio_mac;
         auto taskID         = database.get_dynamic_channel_selection_task_id(radio_mac);
         if (taskID == -1) {
-            LOG(ERROR) << "no task for the requested mac (" << radio_mac << ") could be found!";
+            LOG(ERROR) << "No DCS task for the requested mac (" << radio_mac
+                       << ") is currently running!";
         } else {
             tasks.push_event(taskID, (int)dynamic_channel_selection_task::eEvent::SCAN_TRIGGERED,
                              (void *)&new_event);
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -269,6 +269,15 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
     }
     case eEvent::SCAN_TRIGGER_FAILED: {
         TASK_LOG(DEBUG) << "SCAN_TRIGGER_FAILED received";
+        if (fsm_in_state(eState::WAIT_FOR_SCAN_TRIGGERED)) {
+            auto scan_trigger_failed_event = reinterpret_cast<sScanEvent *>(obj);
+            event_handled                  = true;
+            TASK_LOG(DEBUG) << "SCAN_TRIGGER_FAILED handled on:"
+                            << scan_trigger_failed_event->radio_mac.oct;
+            m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
+            clear_pending_events();
+            fsm_move_state(eState::ABORT_SCAN);
+        }
         break;
     }
     case eEvent::SCAN_TRIGGERED: {
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -267,6 +267,10 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
         m_is_single_scan_pending = true;
         break;
     }
+    case eEvent::SCAN_TRIGGER_FAILED: {
+        TASK_LOG(DEBUG) << "SCAN_TRIGGER_FAILED received";
+        break;
+    }
     case eEvent::SCAN_TRIGGERED: {
         TASK_LOG(DEBUG) << "SCAN_TRIGGERED received";
         if (fsm_in_state(eState::WAIT_FOR_SCAN_TRIGGERED)) {
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -272,8 +272,8 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
         if (fsm_in_state(eState::WAIT_FOR_SCAN_TRIGGERED)) {
             auto scan_trigger_failed_event = reinterpret_cast<sScanEvent *>(obj);
             event_handled                  = true;
-            TASK_LOG(DEBUG) << "SCAN_TRIGGER_FAILED handled on:"
-                            << scan_trigger_failed_event->radio_mac.oct;
+            TASK_LOG(WARNING) << "failed to trigger a scan on:"
+                              << scan_trigger_failed_event->radio_mac.oct << ", aborting scan";
             m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
             clear_pending_events();
             fsm_move_state(eState::ABORT_SCAN);
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
@@ -39,6 +39,7 @@ public:
 #define FOREACH_DCS_EVENT(EVENT)                                                                   \
     EVENT(INVALID_EVENT)                                                                           \
     EVENT(TRIGGER_SINGLE_SCAN)                                                                     \
+    EVENT(SCAN_TRIGGER_FAILED)                                                                     \
     EVENT(SCAN_TRIGGERED)                                                                          \
     EVENT(SCAN_RESULTS_READY)                                                                      \
     EVENT(SCAN_RESULTS_DUMP)                                                                       \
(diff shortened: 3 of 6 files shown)


controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -39,6 +39,7 @@ class dynamic_channel_selection_task : public task, public std::enable_shared_fr
 #define FOREACH_DCS_EVENT(EVENT)                                                                   \
     EVENT(INVALID_EVENT)                                                                           \
     EVENT(TRIGGER_SINGLE_SCAN)                                                                     \
+    EVENT(SCAN_TRIGGER_FAILED)                                                                     \
```
vitalybu, line 42: Maybe describe the commit message in what cases this event is fired

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -252,6 +252,15 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
     }
     case eEvent::SCAN_TRIGGER_FAILED: {
         TASK_LOG(DEBUG) << "SCAN_TRIGGER_FAILED received";
+        if (fsm_in_state(eState::WAIT_FOR_SCAN_TRIGGERED)) {
+            auto scan_trigger_failed_event = reinterpret_cast<sScanEvent *>(obj);
+            event_handled                  = true;
+            TASK_LOG(DEBUG) << "SCAN_TRIGGER_FAILED handled on:"
+                            << scan_trigger_failed_event->radio_mac.oct;
```
vitalybu: That's not a very informative message, and in any case I think it should be printed at least with an INFO (if not WARNING) level.
Consider changing the message to something like:
"Triggering scan failed on <MAC>, aborting scan."

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -252,6 +252,15 @@ void dynamic_channel_selection_task::handle_event(int event_type, void *obj)
     }
     case eEvent::SCAN_TRIGGER_FAILED: {
         TASK_LOG(DEBUG) << "SCAN_TRIGGER_FAILED received";
+        if (fsm_in_state(eState::WAIT_FOR_SCAN_TRIGGERED)) {
+            auto scan_trigger_failed_event = reinterpret_cast<sScanEvent *>(obj);
```
vitalybu, line 273: Is there a chance that we'll received a `SCAN_TRIGGER_FAILED` event not from the radio we're waiting for?
You have `radio_mac` in the event, but you don't check it...

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3551,6 +3551,33 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
     case beerocks_message::ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE: {
         LOG(TRACE) << "ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE for mac " << hostap_mac;
+        auto response =
+            beerocks_header
+                ->addClass<beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE>();
+        if (!response) {
+            LOG(ERROR) << "addClass cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE failed";
+            return false;
+        }
+        //get the mac from hostap_mac
```
vitalybu:
```suggestion

```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3551,6 +3551,33 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+        auto radio_mac = network_utils::mac_from_string(hostap_mac);
+
+        if (!response->success()) {
+            LOG(ERROR) << "failed to trigger scan on radio(" << radio_mac
```
vitalybu:
```suggestion
            LOG(ERROR) << "Failed to trigger scan on radio (" << radio_mac
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3551,6 +3551,33 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+                       << "): sending notification to DCS task";
+
+            dynamic_channel_selection_task::sScanEvent new_event;
+            new_event.radio_mac = radio_mac;
+            auto taskId         = database.get_dynamic_channel_selection_task_id(radio_mac);
+            if (taskId == -1) {
+                LOG(ERROR) << "no task for the requested mac (" << radio_mac << ") could be found!";
```
vitalybu:
```suggestion
                LOG(ERROR) << "No DCS task for the requested mac (" << radio_mac << ") is currently running!";
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3551,6 +3551,33 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+                break;
+            }
+
+            tasks.push_event(taskId,
+                             int(dynamic_channel_selection_task::eEvent::SCAN_TRIGGER_FAILED),
```
vitalybu, line 3577: When do we actually fire the `SCAN_TRIGGERED` event?
If `response->success() == true` no event is fired to the task.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3574,7 +3601,7 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
         new_event.radio_mac = radio_mac;
         auto taskID         = database.get_dynamic_channel_selection_task_id(radio_mac);
         if (taskID == -1) {
-            LOG(ERROR) << "no task for the requesed mac (" << radio_mac << ") could be found!";
+            LOG(ERROR) << "no task for the requested mac (" << radio_mac << ") could be found!";
```
vitalybu:
```suggestion
            LOG(ERROR) << "No DCS task for the requested mac (" << radio_mac << ") is currently running!";
```

controller/src/beerocks/master/son_master_thread.cpp
vitalybu: Just noticed that `SCAN_TRIGGERED` event is fired only after receiving the `ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE` message.
Is this intentional? Seems a bit odd.

vitalybu (changes requested):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
adam1985d: actually it is any failure to trigger a scan.
we don't have a specific case where we experience such failure - but we should protect from any future failure to trigger a scan

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: at the beginning of the `handle_event` function, we filter-out events on radios that are not handled by this task
```
auto event_radio_mac = *reinterpret_cast<sMacAddr *>(obj);
if (m_radio_mac != event_radio_mac) {
    TASK_LOG(ERROR) << "event : " << s_ar_events[int(eEvent(event_type))]
                    << " in state: " << s_ar_states[int(m_fsm_state)]
                    << " for mac: " << event_radio_mac;
    if (m_dcs_waiting_for_event != eEvent::INVALID_EVENT) {
        TASK_LOG(DEBUG) << "calling wait_for_event for "
                        << s_ar_events[int(m_dcs_waiting_for_event)];
        wait_for_event(int(m_dcs_waiting_for_event));
    }
    return;
}
```

itayx (approved):
only nitpick is to split yaml from auto-generated code

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: changed to a `WARNING` level and refactored the log-print message.

controller/src/beerocks/master/son_master_thread.cpp
adam1985d: done

controller/src/beerocks/master/son_master_thread.cpp
adam1985d: done

controller/src/beerocks/master/son_master_thread.cpp
adam1985d: done

controller/src/beerocks/master/son_master_thread.cpp
adam1985d: the `ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION` is the trigger for sending `SCAN_TRIGGERED` event to the task.
this notification is received from the NL when the scan is actually started.

note that the `ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE` returns whether the BWL managed to request the scan from the NL - and not the actual scan start.

controller/src/beerocks/master/son_master_thread.cpp
adam1985d: not part of the changes done - but fixed anyways :-)

kantera800 (commented):
good commit messages.
but, I don't like were this PR is headed.
Remember that soon we need to align DCS according to easymesh spec.
Here added more VS messages that complicates things even more.

First (and maybe i'm wrong) i don't understand why not wait for an event from BWL (SCAN_TRIGGERED_NOTIFICATION, SCAN_ABORT_NOTIFICATION ...) or let the  timeout handle the failure.

Secondly , if you want to handle more error cases i would replace all these notification/responses with 1 CMDU (SCAN_ABORT) that will hold the error type and handle it in the DCS task.

This would save a lot of code space, remove the following VS CMDUs  and be less complicated.
ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE
ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE

will be left with:
ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION
ACTION_CONTROL_CHANNEL_SCAN_RESULTS_NOTIFICATION
ACTION_CONTROL_CHANNEL_SCAN_FINISHED_NOTIFICATION
ACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION

@vitalybu @arnout I would appreciate you opinion as well.

adam1985d:
> good commit messages.
> but, I don't like were this PR is headed.
> Remember that soon we need to align DCS according to easymesh spec.
> Here added more VS messages that complicates things even more.
> 
> First (and maybe i'm wrong) i don't understand why not wait for an event from BWL (SCAN_TRIGGERED_NOTIFICATION, SCAN_ABORT_NOTIFICATION ...) or let the timeout handle the failure.
> 
> Secondly , if you want to handle more error cases i would replace all these notification/responses with 1 CMDU (SCAN_ABORT) that will hold the error type and handle it in the DCS task.
> 
> This would save a lot of code space, remove the following VS CMDUs and be less complicated.
> ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE
> ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE
> 
> will be left with:
> ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION
> ACTION_CONTROL_CHANNEL_SCAN_RESULTS_NOTIFICATION
> ACTION_CONTROL_CHANNEL_SCAN_FINISHED_NOTIFICATION
> ACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION
> 
> @vitalybu @arnout I would appreciate you opinion as well.

1. when you commit regarding a specific commit - you can commit inside a review of the specific commit - this references the commit you comment about automatically
2. I didn't add any new CMDUs - just implemented handling of existing one
3. We shouldn't confuse between `failed-to-perform-requested-operation` (in this case **trigger**-a-scan and `something-went-wrong-during-execution` which is what you refer to.

To conclude - the timeout is a guard against `something-went-wrong-and-we-are-not-aware-of` while the case here is `we-did-not success-to-complete-the-operation`

kantera800:
> 
> 1. when you commit regarding a specific commit - you can commit inside a review of the specific commit - this references the commit you comment about automatically

Its not related to any specific commit its the general ERROR handling that i'm concerned of. 
If you areI prefer to to remove them

kantera800:
> 
> 1. when you commit regarding a specific commit - you can commit inside a review of the specific commit - this references the commit you comment about automatically

There is no specific commit, in fact they are very good.
Its just that i would prefer not to make changes in VS messages unless we have a bug.
But, if you already touched this area IMO you should have taken a different approach (see my suggestion above)

tomereli:
> good commit messages.
> but, I don't like were this PR is headed.
> Remember that soon we need to align DCS according to easymesh spec.
> Here added more VS messages that complicates things even more.
> 
> First (and maybe i'm wrong) i don't understand why not wait for an event from BWL (SCAN_TRIGGERED_NOTIFICATION, SCAN_ABORT_NOTIFICATION ...) or let the timeout handle the failure.
> 
> Secondly , if you want to handle more error cases i would replace all these notification/responses with 1 CMDU (SCAN_ABORT) that will hold the error type and handle it in the DCS task.
> 
> This would save a lot of code space, remove the following VS CMDUs and be less complicated.
> ACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_RESPONSE
> ACTION_CONTROL_CHANNEL_SCAN_DUMP_RESULTS_RESPONSE
> 
> will be left with:
> ACTION_CONTROL_CHANNEL_SCAN_TRIGGERED_NOTIFICATION
> ACTION_CONTROL_CHANNEL_SCAN_RESULTS_NOTIFICATION
> ACTION_CONTROL_CHANNEL_SCAN_FINISHED_NOTIFICATION
> ACTION_CONTROL_CHANNEL_SCAN_ABORT_NOTIFICATION
> 
> @vitalybu @arnout I would appreciate you opinion as well.

The problem is that there is no specific message in the standard which allows a controller to trigger scan on an AP, so I don't know what we can do besides using VS messages... @arnout ?

tomereli (approved):
LGTM. I think the direction is OK since triggering scan is vendor specific (no EasyMesh way to request this specifically).

kantera800:
> no EasyMesh way to request this specifically

@tomereli you are not correct , see easymesh r2 spec

17.1.33 Channel Scan Request message format
The following TLV shall be included in this message:
• One Channel Scan Request TLV (see 17.2.39).
17.1.34 Channel Scan Report message format
The following TLV shall be included in this message:
• One Timestamp TLV (see 17.2.41).
• One or more Channel Scan Result TLVs (see 17.2.40)

tomereli:
> > no EasyMesh way to request this specifically
> 
> @tomereli you are not correct , see easymesh r2 spec
> 
> 17.1.33 Channel Scan Request message format
> The following TLV shall be included in this message:
> • One Channel Scan Request TLV (see 17.2.39).
> 17.1.34 Channel Scan Report message format
> The following TLV shall be included in this message:
> • One Timestamp TLV (see 17.2.41).
> • One or more Channel Scan Result TLVs (see 17.2.40)

I didn't realize this! In this case you are correct and we should fix this to use the specification messaging. However, since this is only modifying existing VS messages and not creating new ones, I say lets merge this and add a task for Adam to align this to the EasyMesh R2 specification. What do you think?

kantera800:
> > > no EasyMesh way to request this specifically
> > 
> > 
> > @tomereli you are not correct , see easymesh r2 spec
> > 17.1.33 Channel Scan Request message format
> > The following TLV shall be included in this message:
> > • One Channel Scan Request TLV (see 17.2.39).
> > 17.1.34 Channel Scan Report message format
> > The following TLV shall be included in this message:
> > • One Timestamp TLV (see 17.2.41).
> > • One or more Channel Scan Result TLVs (see 17.2.40)
> 
> I didn't realize this! In this case you are correct and we should fix this to use the specification messaging. However, since this is only modifying existing VS messages and not creating new ones, I say lets merge this and add a task for Adam to align this to the EasyMesh R2 specification. What do you think?

ok lets merge it .
created task for the easymesh R2 alignment #1152

vitalybu (approved):

PR #1104, opened 2020-04-07
boardfarm: Run test_initial_ap_config for dockerized device

MuteSpirit:
* Start PrplMeshDocker device
* Add draft of test setup config for dummy setup (controller + 2 agents)
  and for setup with single RAX40
* tools/docker/run.sh: fix use network name passed via cmdline args
  instead of hard-coded name

### DIFF
--- a/tests/bft.sh
+++ b/tests/bft.sh
@@ -1,0 +1,6 @@
+#!/bin/sh
+scriptdir=$(dirname $(readlink -f $0))
+bf_dir=$(realpath "$scriptdir/../../boardfarm")
+bfp_dir=$scriptdir
+
+PYTHONPATH="$bfp_dir" python3 "$bf_dir/bft" "$@"
--- a/tests/boardfarm_prplmesh/devices/prplmesh_docker.py
+++ b/tests/boardfarm_prplmesh/devices/prplmesh_docker.py
@@ -1,0 +1,49 @@
+#! /usr/bin/env python3
+#
+# Work-In-Progress. Non-working code yet
+#
+
+from boardfarm.devices import debian
+
+import os
+import subprocess
+from pprint import pprint
+from environment import ALEntriryDocker
+
+# rootdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../..'))
+
+
+class PrplMeshDocker(debian.DebianBox):
+    '''
+    Dockerized prplMesh device containing controller or agent
+    '''
+    model = ('prplmesh_docker')
+    name = "prplmesh_docker"
+    ptyproc = None
+
+    def __init__(self, *args, **kwargs):
+        name = kwargs.pop("name", "gateway")
+        net = kwargs.pop("net", "prplMesh-net")
+
+        if "controller" in kwargs:
+            controller_data = kwargs.pop("controller", {})
+            ucc_port = controller_data.get("ucc_port", 8002)
+            subprocess.check_call((os.path.join(rootdir, "tools", "docker", "run.sh"),
+                                   "--verbose", "--detach", "--force", 
+                                   "--expose {ucc_port} --publish 127.0.0.1::{ucc_port} --ipaddr 0.0.0.0".format(ucc_port=ucc_port),
+                                   "--name", name,
+                                   "--mac", "00:11:22:33",
+                                   "start-controller-agent")
+            self.controller = ALEntityDocker(name, is_controller=True)
+
+        if "agent" in kwargs:
+            agent_data = kwargs.pop("agent", {})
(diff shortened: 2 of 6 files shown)


tests/boardfarm_prplmesh/tests/fake.py
```diff
@@ -0,0 +1,8 @@
+from boardfarm.tests import rootfs_boot
+
+
+class Fake(rootfs_boot.RootFSBootTest):
+    '''lsmod shows loaded kernel modules.'''
```
arnout, line 5: This doesn't seem right :-)

tests/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,49 @@
...
+class PrplMeshDocker(debian.DebianBox):
+    '''
+    Dockerized prplMesh device containing controller or agent
+    '''
+    model = ('prplmesh_docker')
+    name = "prplmesh_docker"
+    ptyproc = None
+
+    def __init__(self, *args, **kwargs):
+        name = kwargs.pop("name", "gateway")
```
arnout, line 25: Why pop and not get?

tests/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,49 @@
...
+        net = kwargs.pop("net", "prplMesh-net")
+
+        if "controller" in kwargs:
+            controller_data = kwargs.pop("controller", {})
+            ucc_port = controller_data.get("ucc_port", 8002)
+            subprocess.check_call((os.path.join(rootdir, "tools", "docker", "run.sh"),
+                                   "--verbose", "--detach", "--force", 
+                                   "--expose {ucc_port} --publish 127.0.0.1::{ucc_port} --ipaddr 0.0.0.0".format(ucc_port=ucc_port),
+                                   "--name", name,
+                                   "--mac", "00:11:22:33",
```
arnout, line 35: I would actually like to get rid of this option and just rely on randomly generated MAC addresses.

tests/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,49 @@
...
+                                   "start-controller-agent")
+            self.controller = ALEntityDocker(name, is_controller=True)
+
+        if "agent" in kwargs:
```
arnout, line 39: This should be just an else. IMO this should only start a single container, not two.

tests/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,49 @@
...
+    ptyproc = None
+
+    def __init__(self, *args, **kwargs):
+        name = kwargs.pop("name", "gateway")
+        net = kwargs.pop("net", "prplMesh-net")
+
+        if "controller" in kwargs:
+            controller_data = kwargs.pop("controller", {})
+            ucc_port = controller_data.get("ucc_port", 8002)
+            subprocess.check_call((os.path.join(rootdir, "tools", "docker", "run.sh"),
```
arnout, line 31: I wonder if the run script still gives much added value. We already create the network externally, and most of the options are passed more or less unchanged to docker...

tests/bft.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+scriptdir=$(dirname $(readlink -f $0))
+bf_dir=$(realpath "$scriptdir/../../boardfarm")
+bfp_dir=$scriptdir
+
+PYTHONPATH="$bfp_dir" python3 "$bf_dir/bft" "$@"
```
arnout, line 6: You need to extend the PYTHONPATH if it exists already.

Also, better use exec.
```suggestion
if [ -n "$PYTHONPATH" ]; then
   PYTHONPATH="$bfp_dir:$PYTHONPATH"
else
   PYTHONPATH="$bfp_dir"
fi
export PYTHONPATH


exec python3 "$bf_dir/bft" "$@"
```

tools/docker/run.sh
```diff
@@ -89,10 +89,10 @@ main() {
 
     [ -z "$IPADDR" -a -n "$NETWORK" ] && {
         dbg "Generate random IP for container $NAME for network $NETWORK"
-        IPADDR=$(generate_container_random_ip $NETWORK)
+        IPADDR=$(generate_container_random_ip ${NETWORK})
```
arnout, line 92: Same for the IP addresses: it would be better to rely on the docker-generated IP addresses instead of trying to make our own.

arnout:
Replaced by #1136.

PR #1106, opened 2020-04-07
scripts: tools/docker/run.sh: fix usage docker network name passed via cmdline arg

MuteSpirit:
Script "tools/docker/run.sh" assumes that gateway/repeater container
will be attached to Docker network named "prplMesh-net-${UNIQUE_ID}" and
ignore network name passed via cmdline argument.

For real boards usage purposes it's need to be able pass custom Docker
network name to provide connectivity of gateway container with choosen
board.

Fix assignement network name with value from args if the last is exist.
Fix checking the presense of created container. Current solution does
not correctly match, for example, "test" name if
"test-and-the-rest-name" is already present.

Signed-off-by: Ivan Efimov <i.efimov@inango-systems.com>

### DIFF
--- a/tools/docker/run.sh
+++ b/tools/docker/run.sh
@@ -44,7 +44,7 @@ generate_container_random_ip() {
 }
 
 gateway_netid_length() {
-    docker network inspect prplMesh-net-${UNIQUE_ID} --format "{{(index .IPAM.Config 0).Subnet}}" | sed -rn 's/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}(\/[0-9]{2})$/\1/p'
+    docker network inspect "$1" --format "{{(index .IPAM.Config 0).Subnet}}" | sed -rn 's/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}(\/[0-9]{2})$/\1/p'
 }
 
 main() {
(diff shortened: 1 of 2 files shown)


tools/docker/run.sh
```diff
@@ -80,19 +80,19 @@ main() {
         run ${scriptdir}/image-build.sh
     }
 
-    NETWORK=prplMesh-net-${UNIQUE_ID}
+    NETWORK=${NETWORK:-prplMesh-net-${UNIQUE_ID}}
```
vitalybu:
```suggestion
    NETWORK="${NETWORK:-prplMesh-net-${UNIQUE_ID}}"
```

tools/docker/run.sh
```diff
@@ -80,19 +80,19 @@ main() {
...
+        dbg "Network ${NETWORK} does not exist, creating..."
         run docker network create ${NETWORK} >/dev/null 2>&1
-        echo "network $NETWORK" >> "${scriptdir}/.test_containers"
+        echo "network ${NETWORK}" >> "${scriptdir}/.test_containers"
     }
 
     [ -z "$IPADDR" -a -n "$NETWORK" ] && {
         dbg "Generate random IP for container $NAME for network $NETWORK"
-        IPADDR=$(generate_container_random_ip $NETWORK)
+        IPADDR=$(generate_container_random_ip ${NETWORK})
```
vitalybu: shellcheck SC2086:
```suggestion
        IPADDR="$(generate_container_random_ip "${NETWORK}")"
```

vitalybu (approved):
Approving as my only comments are shellcheck related nitpicks.
But it would be great if you could add another commit fixing all shellcheck issues.

tomereli (approved):
I removed the ready for merge since it looks like there are shellcheck issues - so please fix those before you add the label back. Otherwise LGTM.

tools/docker/run.sh
MuteSpirit: Fixed

tools/docker/run.sh
MuteSpirit: fixed

PR #1107, opened 2020-04-07
agent: backhaul_manager: get_media_type(): return true on success

### LINKED ISSUE #1097: [BUG] No Topology Response message found
**Describe the bug** -
I found out that if I rebase my side branch on the master, the test I'm working on [4.6.2] is flailing with the following error:

image

it is important to note that before the last rebase the test passed, and the rebase did not ask for any merge conflicts.
Therefore, I have decided to check if the master is broken, and I found out that the same error occurs in every test which involves triggering of the CTT Controller to send Topology Query to MAUT:

image
https://gitlab.com/prpl-foundation/prplMesh/-/jobs/500319964

## Steps to reproduce
run test 4.4.2 on the master.
You can also take a look at my logs: https://gitlab.com/prpl-foundation/prplMesh/pipelines/133287916

## Expected behavior

The test fails with the error message : 
No Topology Response message found

### Casuse of the bug
The last time we ran the certification tests on master version was on `ce6709cae467e48c754f47a6068a3454f0906c0f` commit,  and the test passed.
After that, the following PRs were merged to master: 

- #1053 
- #1060
- #1068 
- #1043
- #1029

So one of them must be the cause of the problem.

arnout:
The get_media_type() function sets result to false by default. In the
IEEE_802_11, it wasn't set to true in the success case.

Set result to true before breaking out of the loop.

Fixes: #1097

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -3439,7 +3439,7 @@ bool backhaul_manager::get_media_type(const std::string &interface_name,
                     if ((std::get<0>(tuple) == radio_info.frequency_band) &&
                         (std::get<1>(tuple) == radio_info.max_bandwidth)) {
                         media_type = std::get<2>(tuple);
-
+                        result     = true;
                         break;
                     }
                 }


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3439,7 +3439,7 @@ bool backhaul_manager::get_media_type(const std::string &interface_name,
                     if ((std::get<0>(tuple) == radio_info.frequency_band) &&
                         (std::get<1>(tuple) == radio_info.max_bandwidth)) {
                         media_type = std::get<2>(tuple);
-
+                        result     = true;
```
tomereli, line 3442: Why can't we just return true in these cases?

tomereli (approved):

rmelotte (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: It's the multiple exit points vs. single exit point discussion, which I'm not sure we put to a vote.

arnout:
Mario did the same fix in parallel and his version is slightly better, so I'm closing this PR.

Issue #1071, opened 2020-04-01
[TASK] Port test_initial_ap_config to boardfarm for dockerized device


abelog:
### Description
As part of migration to boardfarm testing, we need to port `test_initial_ap_config` of `test_flows.py` to boardfarm for "dummy" docker device.

### Exit criteria
In order to complete the task, we have to complete:

- [ ] Integrate all classes in environment.py with boardfarm classes
- [x] Turn this into a boardfarm plugin
- [x] Make sure that `RadioDocker.wait_for_log` still works
- [x] Create a prplMesh testsuite for boardfarm
- [x] Create InitialApConfig test case for boardfarm
- [ ] Run bft from gitlab-ci.yml
- [ ] Document the boardfarm integration in documentation/boardfarm-testing.md based on the UML classes structure and relations in #1032.

oponomar:
To be able use boardfarm without installation it is necessary to add __init__.py to make Python treat the boardfarm directory contains packages for import.
Created PR to prplfoundation/boardfarm:
https://github.com/prplfoundation/boardfarm/pull/10

Updated main PR: #1136

PR #1108, opened 2020-04-07
tests: boardfarm_plugins: add boardfarm tests template

### LINKED ISSUE #1080: [TASK] Boardfarm test template
## Description

As part of migration to Boardfarm testing we need to know how to place Boarfarm tests, test setup configs, devices classes and so on in prplmesh repo

Parent EPIC is #1074 

## Exit criteria

- There are next code inside prplmesh repo:
  - prplmesh Boardfarm test suite
  - prplmesh initial_ap_config test case (but without any test, it just passes)
  - Boardfarm device class for prplMesh docker
  - Boardfarm test setup config for DUT in docker
  - README file with info needed for developers who will migrate `test_flows.py` to Boardfarm

## Detailed description

Use the same folder structure as in Boardfarm repo
Use "boardfarm.plugins" mechanism to show Boardfarm the tests inside prplmesh repo
Use stub test that's just run and return success
Use dockerized DUT in test setup config

taurha:
Added boardfarm plugin's usage description in README.md.
According to plugin's behavior, it was created example of plugin,
which contains suggested file hierarchy, common necessary files,
with stub test to demonstrate its workflow.

Signed-off-by: Oleksii Ponomarenko <o.ponomarenko@inango-systems.com>

### DIFF
--- a/tests/boardfarm_plugins/README.md
+++ b/tests/boardfarm_plugins/README.md
@@ -1,0 +1,182 @@
+<!--
+SPDX-License-Identifier: BSD-2-Clause-Patent
+Copyright (c) 2020 the prplMesh contributors
+This code is subject to the terms of the BSD+Patent license.
+See LICENSE file for more details.
+-->
+
+### Prerequisites
+
+To obtain prplMesh and boardfarm sources a Google repo tool is used.
+Refer to its [README](https://gerrit.googlesource.com/git-repo/+/refs/heads/master/README.md) for detailed info.
+
+### Boardfarm installation
+
+1. To obtain sources using previously mentioned repo tool run:
+```
+repo init -u https://github.com/prplfoundation/prplMesh-manifest.git
+repo sync
+repo forall -p -c 'git checkout $REPO_RREV' # (optional)
+```
+
+2. Install prplMesh to your host using instruction provided by its [README](https://github.com/prplfoundation/prplMesh/blob/master/README.md). 
+
+3. Install additional python packages:
+```
+cd boardfarm
+pip3 install -r requirements.txt
+```
+
+4. Install boardfarm to your host(optional):
+```
+python3 setup.py install --user
+```
+If you consider not to install boardfarm to your host and use binary provided in sources, then you may encounter additional warnings in boardfarm log caused by files not included in instalation package.
+
+### Boardfarm plugin dir structure:
+```
+boardfarm_*     <--- "boardfarm_" prefix is mandatory for boardfarm plugin name
+    |
+    +- __init__.py
+    +- boardfarm_config.json
+    +- devices
+    |   |
+    |   +- __init__.py
+    |   +- sample_device_0.py
+    |   +- sample_device_1.py
+    +- tests
+    |   |
+    |   +- __init__.py
+    |   +- sample_test_0.py
+    |   +- sample_test_1.py
+    |   +- ...
+    +- testsuites.cfg
+```
+
+### Custom device description
+If the tested device is abcent in `boardfarm/boardfarm/devices` dir we can specify new one by creating `custom_config.json` file sort of:
+```json
+{
+    "prplmesh_dummy": {
+        "name": "prplMesh-controller",
+        "board_type": "prplmesh_docker",
+        "role": "controller",
+        "delay": 7,
+        "conn_cmd": "",
+        "connection_type": "local_cmd",
+        "devices": [
+            {
+                "name": "repeater1",
+                "type": "prplMesh_dut",
+                "role": "agent",
+                "delay": 7,
+                "conn_cmd": "",
+                "radios": [
+                    {
+                        "type": "radio_dummy",
+                        "iface": "wlan0"
+                    },
+                    {
+                        "type": "radio_dummy",
+                        "iface": "wlan2"
+                    }
+                ]
+            },
+            {
+                "name": "repeater2",
+                "type": "prplMesh_dut",
+                "role": "agent",
+                "delay": 7,
+                "conn_cmd": "",
+                "radios": [
+                    {
+                        "type": "radio_dummy",
+                        "iface": "wlan0"
+                    },
+                    {
+                        "type": "radio_dummy",
+                        "iface": "wlan2"
+                    }
+                ]
+            }
+        ]
+    }
+}
+
+```
+Mandatory fields for parent device are:
+* `name` is any unique name at all
+* `board_type` is a name of existing device class
+* `devices` is a list of devices that will be added, must be present even it's empty(i.e. `"devices": []`).
+Mandatory fields for child devices(those are listed in **devices** section of parent one):
+* `name` is any unique name at all
+* `type` is a name of existing device class
+
+When running boardfarm we can specify that custom config with in following manner:
+```
+-c /path/to/custom_config.json
+```
+
+### Devices
+There are several devices already existing in `<boardfarm-repo-root>/boardfarm/devices`, but if we created `custom_config.json` which uses previously unknown device, then we must implement its class in plugin dir `boardfarm_*/devices/<device-name>.py`.
+As template for the device we can use `<boardfarm-repo-root>/devices/base.py`.
+
+### Test suites
+Custom test suite should be present in root of boardfarm plugin dir and its name must match `testsuites*.cfg`.
+Suite names are specified in brackets [], test names must match test classes names defined in the files at the "boardfarm_*/tests/" directory.
+Sample contents:
+```
+[suite-name]
+TestCaseName0
+
+[another-suite-name]
+TestCaseName0
+TestCaseName1
+TestCaseName2
+```
+
+### Test file contents:
+Each test must be presented as class, for test file template please refer to:
+`boardfarm/tests/bft_base_test.py`
+Below you can see mandatory class members:
+```
+class SampleTest(BftBaseTest):
+    def runTest(self):
+        '''Main test logic must be implemented here.'''
+        print('{}: Running...'.format(self.__class__.__name__))
+
+    def skipTest(self):
+        '''Method invoked if DUT doesn't respond.
+
+        If the DUT's consoles don't respond, generally the device is considered as 'dead'.
+        Here we can provide info about possible reasons and suggestions.
+        '''
+        print('{}: Skipping...'.format(self.__class__.__name__))
+
+    def recover(self):
+        '''Method invoked if test fails with an unexpected exception type.
+
+        According to boardfarm sources, it is considered as deprecated,
+        but it is suggested to leave it as a stub to prevent possible faults.
+        '''
+        print('{}: Recovering...'.format(self.__class__.__name__)
+
+    @classmethod
+    def teardown_class(cls):
+        '''Teardown method, optional for boardfarm tests.'''
+        print('{}: Teardown...'.format(cls.__class__.__name__))
+```
+For operational example please refer to **boardfarm_prplmesh** plugin.
+
+### Boardfarm usage
+
+To run boardfarm sample you can use following command:
+```
+PYTHONPATH=/path/to/module/dir bft -c /path/to/boardfarm_config.json --testsuite <test-suite-name> -n <device-name>
+```
+or invoke provided script:
+```
+/bin/bash run_bf.sh
+```
+It will lauch *boardfarm_prplmesh* plugin.
+By default, all test logs will be stored in **results** dir.
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/__init__.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/__init__.py
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_base.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_base.py
@@ -1,0 +1,49 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+import pexpect
+
+from boardfarm.devices import linux
+
+
+class PrplMeshBase(linux.LinuxDevice):
+    """PrplMesh abstract device."""
+
+    def _run_shell_cmd(self, cmd: str = "", args: list = None, timeout: int = 30):
+        """Wrapper that executes command with specified args on host machine and logs output."""
+
+        res = pexpect.run(cmd, args=args, timeout=timeout, encoding="utf-8")
+        entry = " ".join((cmd, " ".join(args)))
+        self.log_calls += entry
+        self.log += "$ " + entry + "\r\n" + res
+
+    def check_status(self):
+        """Method required by boardfarm.
+
+        It is used by boardfarm to indicate that spawned device instance is ready for test
+        and also after test - to insure that device still operational.
+        """
+        pass
+
+    def close(self):
+        """Method required by boardfarm.
+
+        Purpose is to close connection to device's consoles.
+        """
+        pass
+
+    def isalive(self):
+        """Method required by boardfarm.
+
+        States that device is operational and its consoles are accessible.
+        """
+        pass
+
+    def touch(self):
+        """Method required by boardfarm.
+
+        Purpose is to keep consoles active, so they don't disconnect for long running activities.
+        """
+        pass
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
@@ -1,0 +1,81 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+import os
+import time
+
+from environment import ALEntityDocker
+from .prplmesh_base import PrplMeshBase
+
+rootdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../..'))
+
+
+class PrplMeshDocker(PrplMeshBase):
+    """Dockerized prplMesh device."""
+
+    model = ("prplmesh_docker")
+    agent_entity = None
+    controller_entity = None
+
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+        config = kwargs.get("config", kwargs)
+
+        # List of device's consoles test can interact with
+        self.consoles = [self]
+
+        # Getting unic ID to distinguish devices and network they belong to
+        self.unique_id = os.getenv("SUDO_USER", os.getenv("USER", ""))
+
+        self.name = "-".join((config.get("name", "prplmesh_docker"), self.unique_id))
+        self.role = config.get("role", "agent")
+        self.cleanup_cmd = config.get("cleanup_cmd", None)
+        self.conn_cmd = config.get("conn_cmd", None)
+        self.delay = config.get("delay", 7)
+        self.docker_network = config.get("docker_network",
+                                         "prplMesh-net-{}".format(self.unique_id))
+
+        docker_cmd = os.path.join(rootdir, "tools", "docker", "run.sh")
+        docker_args = ["--verbose", "--detach", "--force", "--name", self.name,
+                       "--network", self.docker_network]
+
+        if self.role == "controller":
+            # Spawn dockerized controller
+            docker_args.append("start-controller-agent")
+            self._run_shell_cmd(docker_cmd, docker_args)
+
+            time.sleep(self.delay)
+            self.controller_entity = ALEntityDocker(self.name, is_controller=True)
+        else:
+            # Spawn dockerized agent
+            docker_args.append("start-agent")
+            self._run_shell_cmd(docker_cmd, docker_args)
+
+            time.sleep(self.delay)
+            self.agent_entity = ALEntityDocker(self.name, is_controller=False)
+
+        self.check_status()
+
+    def __del__(self):
+        if self.cleanup_cmd:
+            self._run_shell_cmd(self.cleanup_cmd)
+
+    def check_status(self):
+        """Method required by boardfarm.
+
+        It is used by boardfarm to indicate that spawned device instance is ready for test
+        and also after test - to insure that device still operational.
+        """
+        self._run_shell_cmd(os.path.join(rootdir, "tools", "docker", "test.sh"),
+                            ["-v", "-n", self.name])
+
+    def isalive(self):
+        """Method required by boardfarm.
+
+        States that device is operational and its consoles are accessible.
+        """
+        return True
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
@@ -1,0 +1,9 @@
+{
+    "prplmesh_docker": {
+        "name": "dockerized_device",
+        "board_type": "prplmesh_docker",
+        "role": "controller",
+        "conn_cmd": "",
+        "devices": []
+    }
+}
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
@@ -1,0 +1,17 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+from boardfarm.tests import bft_base_test
+
+
+class InitialApConfig(bft_base_test.BftBaseTest):
+    """PrplMesh sample test case, no actual testing, it just passes."""
+
+    def runTest(self):
+        """Main test logic must be implemented here."""
+
+    @classmethod
+    def teardown_class(cls):
+        """Teardown method, optional for boardfarm tests."""
--- a/tests/run_bf.sh
+++ b/tests/run_bf.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+scriptdir=$(dirname "$(readlink -f "${0}")")
+bf_plugins_dir=${scriptdir}/boardfarm_plugins
+bf_dir=$(realpath "${scriptdir}/../../boardfarm")
+
+if [ -n "${PYTHONPATH}" ]; then
+   PYTHONPATH="${bf_dir}:${bf_plugins_dir}:${scriptdir}:${PYTHONPATH}"
+else
+   PYTHONPATH="${bf_dir}:${bf_plugins_dir}:${scriptdir}"
+fi
+export PYTHONPATH
+
+exec python3 "${bf_dir}"/bft -c "${bf_plugins_dir}"/boardfarm_prplmesh/prplmesh_config.json -n prplmesh_docker -x test_flows
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -27,6 +27,7 @@ This file lists all contributors.
 - Mario Maz
 - MaxLinear
 - Moran Shoeg
+- Oleksii Ponomarenko
 - Ostap Slyusar
 - Pavel Levitsky
 - Ram Marzin
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/__init__.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/__init__.py
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/__init__.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/__init__.py
@@ -1,0 +1,4 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/testsuites.cfg
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/testsuites.cfg
@@ -1,0 +1,7 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+[test_flows]
+InitialApConfig


tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,113 @@
...
+        "powerip": "<powerserver>",
+        "powerport": "<outlet>"
+    }
+}
+```
+Where:
+* `name` is any unique name at all
+* `model` is a descriptive model name
+* `conn_cmd` is the command to run to connect to the board
+* `devices` is an array of devices that will be added, must have a WAN device at minimum
```
arnout: Is a WAN device really required? Why?

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+### Devices
+There're bunch of devices already exist in `<boardfarm-repo-root>/boardfarm/devices`,
+but if we created custom `boardfarm_config.json` wich uses previously unknown device,
+then we must implement its class in plugin dir `boardfarm_*/devices/<device-name>.py`.
+As template for the device we can use `<boardfarm-repo-root>/devices/base.py`.
+
+### Test suites
+Custom test suite shoud be presetn in root of boardfarm plugin dir and its name must
```
arnout:
```suggestion
Custom test suite should be present in root of boardfarm plugin dir and its name must
```

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,113 @@
+### Boardfarm plugin dir structure:
```
arnout, line 36: I realize we don't have it in any other markdown file, but we really should have a copyright header:

```suggestion
<!--
SPDX-License-Identifier: BSD-2-Clause-Patent
Copyright (c) 2020 the prplMesh contributors
This code is subject to the terms of the BSD+Patent license.
See LICENSE file for more details.
-->

### Boardfarm plugin dir structure:
```

Note that this would be the first time we use that "the prplMesh contributors" approach. If you feel uncomfortable with it, you can also just use your own name.

arnout (commented):
To be honest, I see little added value of a boardfarm_sample_plugin in the repository. Instead I think we should work directly towards starting the prplmesh docker containers from boardfarm, like is done in #1104.

So I'd rework this PR to just add the README file.

arnout:
> To be honest, I see little added value of a boardfarm_sample_plugin in the repository. Instead I think we should work directly towards starting the prplmesh docker containers from boardfarm, like is done in #1104.

So, to be clear: it *is* useful to have an "empty" plugin and a simple test that always passes. However, it is not useful to call it `boardfarm_sample_plugin`. Instead, it should already be the prplmesh plugin, but with stub implementations.

To clarify this, I've rephrased the exit criteria of #1080 as follows:

- prplmesh Boardfarm test suite
- prplmesh initial_ap_config test case (but without any test, it just passes)
- Boardfarm device class for prplMesh docker
- Boardfarm test setup config for DUT in docker
- README file with info needed for developers who will migrate `test_flows.py` to Boardfarm

> So I'd rework this PR to just add the README file.

And this I said simple because I think the README is ready and we can merge it already. I prefer to merge things quickly instead of waiting for weeks for a super-PR to be finished.

If you can already reach the rest of the exit criteria in e.g. half a day, then by all means do it in the same PR.

tests/boardfarm_plugins/README.md
taurha: > Is a WAN device really required? Why?

By the moment, README.md is still in progress and that list of parameters may change when all acceptance criteria will be completed.

tests/boardfarm_plugins/README.md
taurha: Got it.
"the prplMesh contributors" is OK.

taurha:
> I think the README is ready and we can merge it already

It is still in progress as I want to insure that info provided here will be relevant when all other acceptance criteria will be reached.
I'll try to make it done ASAP.

arnout:
> > I think the README is ready and we can merge it already
> 
> It is still in progress as I want to insure that info provided here will be relevant when all other acceptance criteria will be reached.
> I'll try to make it done ASAP.

Understood. In that case, don't hurry, and continue as you were planning.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+```
+3. Install boardfarm to your host:
+```
+sudo python3 setup.py install
+```
+4. Copy missing *html* templates to installed boardfarm dir, example:
+```
+sudo cp -r boardfarm/html /usr/local/lib/python3.6/dist-packages/boardfarm-1.1.1-py3.6.egg/boardfarm/
+```
+(They were missed in setup script)
```
MuteSpirit: Maybe include into `boardfarm_hotfix.patch` also changes in `setup.py` script to fix installation `html` files?
They maybe installed as [additional data files](https://docs.python.org/3/distutils/setupscript.html#installing-additional-files)

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Boardfarm plugin dir structure:
+```
+boardfarm_*                 <--- that name template is mandatory for boardfarm plugin
```
MuteSpirit: I suggest to make comment more strict, i.e.
`"boardfarm_" prefix is mandatory for boardfarm plugin name`

tests/boardfarm_plugins/run_bf.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
MuteSpirit: Copyright is missing

tests/boardfarm_plugins/run_bf.sh
```diff
@@ -0,0 +1,15 @@
...
+scriptdir=$(dirname $(readlink -f ${0}))
+bf_plugin_dir=${scriptdir}
```
MuteSpirit: `bf_plugin_dir` -> `bf_plugins_dir`
Reason: `tests/boardfarm_plugins/` is folder for many plugins

tests/boardfarm_plugins/run_bf.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+scriptdir=$(dirname $(readlink -f ${0}))
```
MuteSpirit: Please fix shellcheck issue:
```
In test.sh line 2:
scriptdir=$(dirname $(readlink -f ${0}))
                    ^-----------------^ SC2046: Quote this to prevent word splitting.
                                  ^--^ SC2086: Double quote to prevent globbing and word splitting.
```

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+        "role": "controller",
+        "delay": 7,
+        "conn_cmd": "",
+        "cleanup_cmd": "docker container rm -f gateway",
+        "connection_type": "local_cmd",
+        "devices": [
+            {
+                "name": "repeater1",
+                "type": "prplmesh_docker",
+                "role": "agent",
```
MuteSpirit, line 71: As far as i understand 
1) we are going to use similar test setup config but set another types: `prplmesh_docker` -> `prplwrt`, `radio_dummy` -> `radio_hw` and etc.
2) one container (or one board) may run both controller and agent
See peace of `test_gw_repeater.sh` for run gateway:
```
    [ "$START_GATEWAY" = "true" ] && {
        "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} ${GW_EXTRA_OPT} \
            start-controller-agent -d -n "${GW_NAME}" -m 00:11:22:33 -- "$@"
    }
```

So i think that ability to use only single **role** for device is not correct and it's needed to have ability to say that in docker (or board) we want to run both controller and agent

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+}
+
+```
+Mandatory fields for parent device are:
+* `name` is any unique name at all
+* `board_type` is a name of existing device class
+* `devices` is a list of devices that will be added, must have at least one device.
+Mandatory fields for child devices(those are listed in **devices** section of parent one):
+* `name` is any unique name at all
+* `board_type` is a name of existing device class
```
MuteSpirit: For devices it's needed to provide `type` instead of `board_type`

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+we can specify new one by creating `custom_config.json` file sort of:
+```json
+{
+    "prplmesh_dummy": {
+        "name": "gateway",
+        "board_type": "prplmesh_docker",
+        "role": "controller",
+        "delay": 7,
+        "conn_cmd": "",
+        "cleanup_cmd": "docker container rm -f gateway",
```
MuteSpirit: You incapsulates in class `PrplMeshDocker` that controller/agent will run inside docker container.
So i suggest to execute command for stop docker container implicitly in class `PrplMeshDocker` and avoid mention `docker container rm` in test setup config file.

There will be one more advantage in this solution - we won't duplicate container name in fields `name` and `cleanup_cmd`

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+```
+
+### Devices
+There're bunch of devices already exist in `<boardfarm-repo-root>/boardfarm/devices`,
+but if we created `custom_config.json` which uses previously unknown device,
+then we must implement its class in plugin dir `boardfarm_*/devices/<device-name>.py`.
+As template for the device we can use `<boardfarm-repo-root>/devices/base.py`.
+
+### Test suites
+Custom test suite shoud be presetn in root of boardfarm plugin dir and its name must
```
MuteSpirit: Fix the typo: `presetn` -> `present`

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+match `testsuites*.cfg`.
+Suite names are specified in brackets [], test names must match tests found
```
MuteSpirit: I suggest to edit phrase next way:
`test names must match test classes names defined in the files at the "boardfarm_*/tests/" directory`

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,126 @@
+#!/usr/bin/env python3
+
```
MuteSpirit: Copyright missing

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,126 @@
...
+        if self.cleanup_cmd:
+            self._run_shell_cmd(self.cleanup_cmd)
+
+        docker_run_cmd = os.path.join(rootdir, "tools", "docker", "run.sh") \
+                        + " --verbose --detach --force --name {}".format(self.name) \
+                        + " --mac {}".format(self.mac)
+
+        if self.role == "controller":
+            # Spawn dockerized controller
+            self._run_shell_cmd(docker_run_cmd + " start-controller-agent")
```
MuteSpirit: Command `start-controller-agent` starts controller + local agent instead of alone controller

tests/boardfarm_plugins/run_bf.sh
```diff
@@ -0,0 +1,15 @@
...
+prplMesh_dir=$(realpath "${scriptdir}/..")
+
+if [ -n "${PYTHONPATH}" ]; then
+   PYTHONPATH="${bf_dir}:${bf_plugin_dir}:${prplMesh_dir}:${PYTHONPATH}"
+else
+   PYTHONPATH="${bf_dir}:${bf_plugin_dir}:${prplMesh_dir}"
+fi
+export PYTHONPATH
+
+exec bft -c boardfarm_sample_plugin/prplmesh_config.json -n prplmesh_dummy -x sample-suite
```
MuteSpirit: Does i correctly understand that if you pass args (`-c`, `-n`, `-x`) to `bft` here then you are not planning that `run_bf.sh` will be used as Boardfarm run script?

MuteSpirit (changes requested):

tests/boardfarm_plugins/README.md
taurha: > 1. we are going to use similar test setup config but set another types: `prplmesh_docker` -> `prplwrt`, `radio_dummy` -> `radio_hw` and etc.

Yeap, right.

> So i think that ability to use only single **role** for device is not correct and it's needed to have ability to say that in docker (or board) we want to run both controller and agent

I'll fix that and add `"role": "both"` for cases when we need both roles at single device.

tests/boardfarm_plugins/run_bf.sh
taurha: > Does i correctly understand that if you pass args (`-c`, `-n`, `-x`) to `bft` here then you are not planning that `run_bf.sh` will be used as Boardfarm run script?

Script's current version is supposed to be used only for demonstration of sample plugin.
When in #1071 there will be added operational testsuite where I'll provide a bit more advanced launcher script.

tests/boardfarm_plugins/README.md
taurha: @MuteSpirit , I've checked `test_gw_repeater.sh` code - option `start-controller-agent` is the only available to start controller.
As `test_flows.py` apparently using only these two options: `start-controller-agent` and `start-agent` - I think, we should maintain only these ones so far..

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -56,11 +62,16 @@ def __init__(self, *args, **kwargs):
...
+        elif:
             # Spawn dockerized agent
             self._run_shell_cmd(docker_run_cmd + " start-agent")
             time.sleep(config.get("delay", 5))
             self.al_entity = ALEntityDocker(self.name, is_controller=False)
+        else:
+            # Spawn dockerized controller
+            self._run_shell_cmd(docker_run_cmd + " start-controller-agent")
+            time.sleep(config.get("delay", 5))
+            self.al_entity = ALEntityDocker(self.name, is_controller=True)
```
taurha: These changes were made by mistake, removed them in next fixup commit.

.gitignore
```diff
@@ -7,6 +7,7 @@ _GO_KW
 .kwps
 kw_reports
 __pycache__/
+*.pyc
```
tomereli: This should be in a separate commit.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
+<!--
+SPDX-License-Identifier: BSD-2-Clause-Patent
+Copyright (c) 2020 the prplMesh contributors
+This code is subject to the terms of the BSD+Patent license.
+See LICENSE file for more details.
+-->
+
+### Boardfarm installation
+
+1. To obtain boardfarm sources alongside prplMesh ones run:
```
tomereli: I suggest to also add a Prerequisites section which will explain that google repo needs to be downloaded - you can refer to its README.md - https://gerrit.googlesource.com/git-repo/+/refs/heads/master/README.md

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+```
+repo init -u https://github.com/prplfoundation/prplMesh-manifest.git
```
tomereli, line 17: I actually think that we should have a single manifest in prpl and use different branches for different projects where needed, so I would change that to just 'manifest.git' (nitpick)

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+repo sync
+```
+
+2. Apply provided patch(*boardfarm_hotfix.patch*) to fix boardfarm issue with log writing:
```
tomereli: In this case I suggest we fork boardfarm and fix it in our fork. Users should not be applying patches.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+Mandatory fields for child devices(those are listed in **devices** section of parent one):
+* `name` is any unique name at all
+* `board_type` is a name of existing device class
+When running boardfarm we can specify that custom config with in following manner:
+```
+-c /path/to/custom_config.json
+```
+
+### Devices
+There're bunch of devices already exist in `<boardfarm-repo-root>/boardfarm/devices`,
```
tomereli:
```suggestion
There are several devices already existing in `<boardfarm-repo-root>/boardfarm/devices`,
```

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Test file contents:
+Each test must be presented as class, for test file template please refer to:
+`boardfarm/tests/bft_base_test.py`
+Above you can see mandatory class members:
+```
+class SampleTest(BftBaseTest):
+    def runTest(self):
+        '''
+        Main test logic must be implemented here.
```
tomereli: Is this according to our python coding style?

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,126 @@
...
+            self._run_shell_cmd(self.cleanup_cmd)
+
+        docker_run_cmd = os.path.join(rootdir, "tools", "docker", "run.sh") \
+                        + " --verbose --detach --force --name {}".format(self.name) \
+                        + " --mac {}".format(self.mac)
+
+        if self.role == "controller":
+            # Spawn dockerized controller
+            self._run_shell_cmd(docker_run_cmd + " start-controller-agent")
+            time.sleep(config.get("delay", 5))
```
tomereli: Delay should be an optional parameter. Why always delay 5 seconds?

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,126 @@
...
+            self.al_entity = ALEntityDocker(self.name, is_controller=False)
+    
+        self.check_device_status()
+
+    def __del__(self):
+        if self.cleanup_cmd:
+            self._run_shell_cmd(self.cleanup_cmd)
+
+    def _get_rand_mac(self, oct_amount: int):
```
tomereli: IMO this should go to a different utils (static) class

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,126 @@
...
+        The reason of that behavior wasn't clarified so far, so a temporary empiric solution
+        is provided.
+        """
+        # mac="{:02x}".format(random.randint(0, 255))
+        mac="00"
+        for i in range(1, oct_amount):
+            mac += ":{:02x}".format(random.randint(0, 255))
+        return mac
+
+    def _run_shell_cmd(self, cmd: str = ""):
```
tomereli: This should also be in a common class, or in the base class.

tests/boardfarm_plugins/boardfarm_sample_plugin/prplmesh_config.json
```diff
@@ -0,0 +1,49 @@
...
+        "delay": 7,
+        "conn_cmd": "",
+        "cleanup_cmd": "docker container rm -f gateway",
+        "connection_type": "local_cmd",
+        "devices": [
+            {
+                "name": "repeater1",
+                "type": "prplmesh_docker",
+                "role": "agent",
+                "delay": 7,
```
tomereli: So we have the delay here - what is it used for?

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
```diff
@@ -0,0 +1,41 @@
+
+import boardfarm
+
+from boardfarm.tests import bft_base_test
```
tomereli: All this file changes should be in a separate commit (nitpick)

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -57,11 +57,16 @@ def __init__(self, *args, **kwargs):
                         + " --verbose --detach --force --name {}".format(self.name) \
                         + " --mac {}".format(self.mac)
 
-        if self.role == "controller":
```
tomereli: Controller alone doesn't need to be supported IMO, we don't support running a standalone controller without a local agent in prplmesh anyway.

tests/boardfarm_plugins/README.md
arnout: Renaming an existing repository is not something that is advisable unless there's a very good reason for it.

tests/boardfarm_plugins/README.md
arnout: :+1: 
The original assumption was that we would be able to use boardfarm as is. If not, we need to fork it.

That also allows us to create an upstream PR.

.gitignore
arnout: Actually, this shouldn't even be there... This would only be needed for python2 because python3 compiled files are under `__pycache__`.

tests/boardfarm_plugins/README.md
taurha: > Is this according to our python coding style?

Actually no, but such naming style is the part of boardfarm plugin requirements.
(i.e. it is mandatory)

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
taurha: > Delay should be an optional parameter. Why always delay 5 seconds?

We can rewrite that value using setup description in `.json` config.
I saw that delay hardcoded in `tests/environment.py` (7sec), so I've thought that it is reasonable and added it here(slightly decreased).

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
taurha: > IMO this should go to a different utils (static) class

It contains a "workaround"(constant first octet), so I can't state that method for general use.
Perhaps, I'll leave that "workaround" here and move that function to separate utils class.
Is that acceptable?

tests/boardfarm_plugins/boardfarm_sample_plugin/prplmesh_config.json
taurha: > So we have the delay here - what is it used for?

That delay was hardcoded in `tests/environment.py: launch_environment_docker()` and it's using as timeout between starting of dockerized device and creating corresponding ALEntity object.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+### Custom device description
+If the tested device is abcent in `boardfarm/boardfarm/devices` dir
+we can specify new one by creating `custom_config.json` file sort of:
+```json
+{
+    "prplmesh_dummy": {
+        "name": "gateway",
+        "board_type": "prplmesh_docker",
```
arnout, line 62: AFAIU, this is supposed to be the DUT, and the "devices" section below is for additional stuff that is connected to the DUT and interacts with it (although it seems the boundary is a bit vague, it also seems to represent parts of the DUT e.g. wan and lan).

Also AFAIU, we were planning to run only agent on the physical DUT and always use a docker for the controller.

So, I think the DUT would typically not be the gateway with controller role, but rather a repeater with agent role.

However, since this is just an example, it doesn't matter much.

tests/boardfarm_plugins/README.md
arnout: I think we more or less decided that we would use repeater1 as the DUT and always use a docker for the controller - at least to start with. The reasoning behind this is that the controller functionality anyway gets tested even if it runs in docker, and it shouldn't make a difference whether it's in docker or on an actual device.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+```
+Mandatory fields for parent device are:
+* `name` is any unique name at all
+* `board_type` is a name of existing device class
+* `devices` is a list of devices that will be added, must have at least one device.
+Mandatory fields for child devices(those are listed in **devices** section of parent one):
+* `name` is any unique name at all
+* `type` is a name of existing device class
+When running boardfarm we can specify that custom config with in following manner:
```
arnout, line 115: (nitpick) this should be a new paragraph.
```suggestion

When running boardfarm we can specify that custom config with in following manner:
```

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+[another-suite-name]
+TestCaseName0
+TestCaseName1
+TestCaseName2
+```
+
+### Test file contents:
+Each test must be presented as class, for test file template please refer to:
+`boardfarm/tests/bft_base_test.py`
+Above you can see mandatory class members:
```
arnout:
```suggestion
Below you can see mandatory class members:
```
at least I think that's what you mean?

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+    |   +- __init__.py
+    |   +- sample_test_0.py
+    |   +- sample_test_1.py
+    |   +- ...
+    +- testsuites.cfg
+```
+
+### Custom device description
+If the tested device is abcent in `boardfarm/boardfarm/devices` dir
+we can specify new one by creating `custom_config.json` file sort of:
```
arnout: Please refer to CODINGSTYLE.md for the coding style of markdown documents:

> Lines in markdown documents are not wrapped, but every sentence should start on a new line.
> Wrapping in text files is annoying, because adding a single word can cause the entire paragraph to be rewrapped.
> On the other hand, putting everything on one line is not readable either.
> As a compromise, no fixed wrapping is done.
> Instead, each sentence should be started on a new line.
> This way, the diff will only show the sentences that are really affected by the change.
> For very long sentences, subsentences may also be placed on a separate line.

Please apply this to the entire document.

tests/boardfarm_plugins/README.md
tomereli: I created this repository and it is not used anywhere yet, so if we ever want to rename it, now's the time. Its not terribly important but normally the manifest repository is just called manifest, don't know why I chose `prplMesh-manifest` :/

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
tomereli: Yes, I think that we should use the values from the config file.

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
tomereli: Yes, that's acceptable. Also leaving it as is acceptable, your call :)

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+### Test file contents:
+Each test must be presented as class, for test file template please refer to:
+`boardfarm/tests/bft_base_test.py`
+Above you can see mandatory class members:
+```
+class SampleTest(BftBaseTest):
+    def runTest(self):
+        '''
+        Main test logic must be implemented here.
+        '''
```
arnout: As @tomereli commented:
```suggestion
    def runTest(self):
        '''Main test logic must be implemented here.'''
```

Cfr. [PEP8 documentation strings](https://www.python.org/dev/peps/pep-0008/#documentation-strings)

arnout (changes requested):
Unfinished, but I'll want to re-review afterwards so already marking as Changes Requested.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+        print('{}: Running...'.format(self.__class__.__name__))
+
+    def skipTest(self):
+        '''
+        Method invoked if DUT's consoles don't respond,
+        generally considering device as 'dead'.
+        Here we can provide info about possible reasons and suggestions.
+        '''
```
arnout:
```suggestion
        '''Method invoked if DUT doesn't respond.
        
        If the DUT's consoles don't respond, generally the device is considered as 'dead'.
        Here we can provide info about possible reasons and suggestions.
        '''
```

Also for the other docstrings please follow PEP8 coding style.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+        but it is suggested to leave it as a stub to prevent possible faults.
+        '''
+        print('{}: Recovering...'.format(self.__class__.__name__)
+
+    @classmethod
+    def teardown_class(cls):
+        '''
+        Teardown method, optional for boardfarm tests.
+        Its 'strange' declaration comparing to other test class methods
+        caused only to boardfarm workflow
```
arnout: What does this mean?

Are you referring to the `@classmethod` bit? There's nothing strange about that, it's a decorator. [Search the web for some explanation](https://www.tutorialsteacher.com/python/classmethod-decorator)

tests/boardfarm_plugins/boardfarm_sample_plugin/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+SPDX-License-Identifier: BSD-2-Clause-Patent
+SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+This code is subject to the terms of the BSD+Patent license.
+See LICENSE file for more details.
+"""
```
arnout: (nitpick) if you do it like this, it will be the module (or in this case, package) docstring. We generally don't want to use the license header for the docstring; instead, we'd want an actual docstring. Therefore:
```suggestion
# SPDX-License-Identifier: BSD-2-Clause-Patent
# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
# This code is subject to the terms of the BSD+Patent license.
# See LICENSE file for more details.
```

tests/boardfarm_plugins/boardfarm_hotfix.patch
```diff
@@ -0,0 +1,34 @@
...
+                     line = line[1:]
+                 if line.startswith(' ['):
+                     line = line[1:]
+-                ts, text = line.split(']', 1)
++                text = line.split(']', 1)
+                 combined_list.append({
+-                    "time": float(ts[1:-1]),
++                    "time": 0,
+                     "text": str(text),
+                     "name": name
```
arnout: This doesn't seem right - I think it will replace one bug with another one. I guess the problem was that this would do the wrong thing in case `line` didn't contain the ` [<timestamp>]` part. So I guess the correct fix would be:
```
if line.startswith(' ['):
    line = line[1:]
    ts, text = line.split(']', 1)
    timestamp = float(ts[1:-1])
else:
    line = text
    timestamp = 0.0

combined_list.append({
    "time": timestamp,
    "text": str(text),
...
```

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+SPDX-License-Identifier: BSD-2-Clause-Patent
+SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+This code is subject to the terms of the BSD+Patent license.
+See LICENSE file for more details.
+"""
```
arnout: Same comment.

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python3
+"""
+SPDX-License-Identifier: BSD-2-Clause-Patent
+SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+This code is subject to the terms of the BSD+Patent license.
+See LICENSE file for more details.
+"""
```
arnout: Same comment, should be comment rather than docstring.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,182 @@
...
+        '''
+        print('{}: Teardown...'.format(cls.__class__.__name__))
+```
+For operational example please refer to **boardfarm_sample_plugin** code.
```
arnout: As a wrote before: I don't think it makes sense to add a `boardfarm_sample_plugin`. Instead, add the `boardfarm_prplmesh` plugin. We start out with the sample code, and replace that with the test_flows code as we continue doing the boardfarm conversion.

Once the boardfarm conversion is done, there is really no reason to keep a sample plugin in our repository. So it's better to work on the *actual* plugin right away (but fill it in with sample code to start with).

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
```diff
@@ -0,0 +1,42 @@
...
+        print('{}: Running...'.format(self.__class__.__name__))
+
+    def skipTest(self, msg: str = ""):
+        '''
+        Method invoked if DUT's consoles don't respond,
+        generally considering device as 'dead'(in normal workflow).
+        Also, it can be used to provide info about possible failure reasons.
+        '''
+        print('{}: Skipping...'.format(self.__class__.__name__))
+        raise boardfarm.exceptions.SkipTest(msg)
```
arnout: Do we need to implement this method or is it optional?

If it is optional, I would just remove it. It's mentioned in the README, that's enough.

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def recover(self):
+        '''
+        Method invoked if test fails with an unexpected exception type.
+        According to boardfarm sources, it is considered as deprecated,
+        and it is suggested to leave it as a stub.
+        '''
+        print('{}: Recovering...'.format(self.__class__.__name__))
```
arnout: Same here: if there is a default implementation in BftBaseTest, there's no need to implement it here.

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
```diff
@@ -0,0 +1,42 @@
...
+        generally considering device as 'dead'(in normal workflow).
+        Also, it can be used to provide info about possible failure reasons.
+        '''
+        print('{}: Skipping...'.format(self.__class__.__name__))
+        raise boardfarm.exceptions.SkipTest(msg)
+
+    def recover(self):
+        '''
+        Method invoked if test fails with an unexpected exception type.
+        According to boardfarm sources, it is considered as deprecated,
```
arnout: If it is deprecated, what is the "new" way of recovering from a hard failure?

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
```diff
@@ -0,0 +1,42 @@
+
+import boardfarm
+
+from boardfarm.tests import bft_base_test
+
+
+class InitialApConfig(bft_base_test.BftBaseTest):
+    '''
+    PrplMesh sample test case, no actual testing, it just passes
+    '''
```
arnout: Docstring coding style:
```suggestion
    '''prplMesh sample test case, no actual testing, it just passes.'''
```

tests/boardfarm_plugins/boardfarm_sample_plugin/testsuites.cfg
```diff
@@ -0,0 +1,7 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+[sample-suite]
```
arnout: Can we give a better name? E.g. prplmesh-testflows? Or just testflows?

tests/boardfarm_plugins/boardfarm_sample_plugin/prplmesh_config.json
```diff
@@ -0,0 +1,49 @@
+{
+    "prplmesh_dummy": {
+        "name": "gateway",
+        "board_type": "prplmesh_docker",
+        "role": "controller",
+        "delay": 7,
+        "conn_cmd": "",
+        "cleanup_cmd": "docker container rm -f gateway",
+        "connection_type": "local_cmd",
```
arnout: Please limit the contents of this file to what is actually used/required. I believe that would be just cleanup_cmd and connection_type.

tests/boardfarm_plugins/boardfarm_sample_plugin/prplmesh_config.json
```diff
@@ -0,0 +1,49 @@
...
+        "devices": [
+            {
+                "name": "repeater1",
```
arnout: Although the README claims that devices have to be declared, is that really true?

tests/boardfarm_plugins/run_bf.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+scriptdir=$(dirname "$(readlink -f "${0}")")
+bf_plugins_dir=${scriptdir}
+bf_dir=$(realpath "${scriptdir}/../../../boardfarm/boardfarm")
```
arnout: I would put this script one directory higher, in `tests/run_bf.sh`. It will be the typical entrypoint for the developer who wants to run tests, so each subdirectory is more typing.

tests/boardfarm_plugins/run_bf.sh
```diff
@@ -0,0 +1,19 @@
...
+prplMesh_dir=$(realpath "${scriptdir}/..")
+
+if [ -n "${PYTHONPATH}" ]; then
+   PYTHONPATH="${bf_dir}:${bf_plugins_dir}:${prplMesh_dir}:${PYTHONPATH}"
+else
+   PYTHONPATH="${bf_dir}:${bf_plugins_dir}:${prplMesh_dir}"
+fi
+export PYTHONPATH
+
+exec bft -c "${bf_plugins_dir}"/boardfarm_sample_plugin/prplmesh_config.json -n prplmesh_dummy -x sample-suite
```
arnout: Does the boardfarm setup really have to be run? Can't we skip it by using
```suggestion
exec ${bf_dir}/bft -c "${bf_plugins_dir}"/boardfarm_sample_plugin/prplmesh_config.json -n prplmesh_dummy -x sample-suite
```

arnout (changes requested):
The `__init__.py` in the tests subdirectory is missing the copyright header.

I think it would be better to split this PR into a couple of commits: 
1. README 
2. add plugin but without devices or tests
3. add test + testsuite
4. add device
5. add run_bft.sh

However, I don't think it's worth spending too much time on splitting here. That said, please *do* spend some time on writing a good commit message.

I haven't reviewed the `prplmesh_docker.py` file yet, but I wonder if it really belongs to this PR. Indeed, it doesn't get used in the sample test, so it's dead code for the moment.

tests/boardfarm_plugins/README.md
taurha: > What does this mean?
> 
> Are you referring to the `@classmethod` bit? There's nothing strange about that, it's a decorator. [Search the web for some explanation](https://www.tutorialsteacher.com/python/classmethod-decorator)

I considered `teardown_class` function strange because I don't understand why it was implemented as 'static' one.
Anyway, It seems like that comment is pointless and I'll remove it leaving only function description.

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
oponomar: > Do we need to implement this method or is it optional?
> 
> If it is optional, I would just remove it. It's mentioned in the README, that's enough.

All four methods mentioned in the README are mandatory.

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
taurha: > If it is deprecated, what is the "new" way of recovering from a hard failure?

There's no "new" way for current version(v1.1.1), warning message states that it will be provided in future versions.

tests/boardfarm_plugins/run_bf.sh
taurha: > Does the boardfarm setup really have to be run? Can't we skip it by using

We need install boardfarm to run it.
I've spent some time to investigate how we can bypass that limitation - it requires changes of boardfarm sources, which may be very time consuming.

tests/boardfarm_plugins/boardfarm_sample_plugin/prplmesh_config.json
taurha: > Although the README claims that devices have to be declared, is that really true?

With no devices boardfarm will raise exception, so we need specify at least one device.

tests/boardfarm_plugins/boardfarm_hotfix.patch
taurha: > This doesn't seem right - I think it will replace one bug with another one. I guess the problem was that this would do the wrong thing in case `line` didn't contain the ` [<timestamp>]` part. So I guess the correct fix would be:
> 
> ```
> if line.startswith(' ['):
>     line = line[1:]
>     ts, text = line.split(']', 1)
>     timestamp = float(ts[1:-1])
> else:
>     line = text
>     timestamp = 0.0
> 
> combined_list.append({
>     "time": timestamp,
>     "text": str(text),
> ...
> ```

Added that solution to PR:
https://github.com/prplfoundation/boardfarm/pull/8

taurha:
Updated prplfoundation/boardfarm fork with latest upstream and added PR with changes previously proposed by boardfarm_hotfix.patch (it is no longer necessary, removed it from current PR):
https://github.com/prplfoundation/boardfarm/pull/8

tests/boardfarm_plugins/README.md
taurha: Added PR to boardfarm fork with changes proposed in patch:
https://github.com/prplfoundation/boardfarm/pull/8

tests/boardfarm_plugins/boardfarm_sample_plugin/devices/prplmesh_docker.py
taurha: > This should also be in a common class, or in the base class.

Added `PrplMeshBase` parent class, added common methods to it.
(`PrplMeshDocker` inherits it)

tests/boardfarm_plugins/boardfarm_sample_plugin/tests/initial_ap_config.py
taurha: > All this file changes should be in a separate commit (nitpick)

I changed commit history according @arnout 's suggestions:
1. add README
2. add plugin but without devices or tests
3. add test + testsuite
4. add base class
5. add docker device
6. move common functions from class methods to separate file
6. add run_bft.sh

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,174 @@
...
+SPDX-License-Identifier: BSD-2-Clause-Patent
+Copyright (c) 2020 the prplMesh contributors
+This code is subject to the terms of the BSD+Patent license.
+See LICENSE file for more details.
+-->
+
+### Prerequisites
+
+To obtain prplMesh and boardfarm sources a Google repo tool is used.
+Refer to its README for detailed info:
```
tomereli:
```suggestion
Refer to its [README](https://gerrit.googlesource.com/git-repo/+/refs/heads/master/README.md) for detailed info.
```

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,174 @@
...
+https://gerrit.googlesource.com/git-repo/+/refs/heads/master/README.md
```
tomereli:
```suggestion
```

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+### Boardfarm installation
+
+1. To obtain sources using previously mentioned repo tool run:
+```
+repo init -u https://github.com/prplfoundation/prplMesh-manifest.git
+repo sync
+repo forall -p -c 'git checkout $REPO_RREV'
```
tomereli: This is actually optional, it is perfectly fine to work in detached head (which is what repo is doing by default when its saying "no branch") - bottom line, you should add a comment here saying this is optional (nitpick)

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+2. Install boardfarm to your host:
+```
+sudo python3 setup.py install
```
tomereli: 1. You need to add a 'cd boardfarm` before this commant
2. Is there a way to install it without sudo?

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -0,0 +1,17 @@
...
+
+class InitialApConfig(bft_base_test.BftBaseTest):
+    """PrplMesh sample test case, no actual testing, it just passes."""
+
+    def runTest(self):
+        """Main test logic must be implemented here."""
+
+    @classmethod
+    def teardown_class(cls):
+        """Teardown method, optional for boardfarm tests."""
```
tomereli, line 17: Same here - the whole commit looks like it was "cut" in the middle. Something wrong with rebase?

tests/boardfarm_plugins/boardfarm_prplmesh/__init__.py
```diff
@@ -0,0 +1,4 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
```
tomereli, line 1: Are these empty intentionally?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -5,12 +5,12 @@
 
 import logging
 import os
-import random
 import sys
 import time
 
 from environment import ALEntityDocker
 from .prplmesh_base import PrplMeshBase
+from utils import get_rand_mac
```
tomereli: I suggest to add a class with static functions, then you can just import the class when using one of the functions instead of importing each function on its own.

tomereli (approved):
Still have some comments, but nothing which requires re-review, so approving.

tests/boardfarm_plugins/README.md
taurha: > 1. You need to add a 'cd boardfarm` before this commant
> 2. Is there a way to install it without sudo?

1. Indeed
2. No, because we need permissions to write in `/usr/local/lib/python3.6/dist-packages` where boardfarm will be installed.

tests/boardfarm_plugins/boardfarm_prplmesh/__init__.py
taurha: > Are these empty intentionally?

Yes, all `__init__.py` were added for only purpose: mark any `*.py` file at same directory level as python module.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
taurha: > Same here - the whole commit looks like it was "cut" in the middle. Something wrong with rebase?

Actually - no.
I've splitted all commits to atomic units as were requested by @arnout.
In that particular case - the dummy testsuite was added, its just passes.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
taurha: > I suggest to add a class with static functions, then you can just import the class when using one of the functions instead of importing each function on its own.

Indeed.
But as mentioned @arnout in his [comment](https://github.com/prplfoundation/prplMesh/pull/1136#discussion_r408968971) to #1136 draft PR - that function will become useless soon, so I'll remove it here at all.

tests/boardfarm_plugins/README.md
arnout: Actually, you *can* install without sudo, you just need an additional option `--user`:
```suggestion
python3 setup.py install --user
```

tests/boardfarm_plugins/run_bf.sh
arnout: No we don't.

The only thing that is needed is to install the requirements. So `pip3 install -r requirements.txt` is sufficient. Or alternatively, what I did, install the required packages using the distro package manager.

tests/boardfarm_plugins/boardfarm_sample_plugin/prplmesh_config.json
taurha: > Although the README claims that devices have to be declared, is that really true?

I've rechecked that option - we can specify empty device list `"devices": [],` and that actually works, but `"devices": []` list, even if it's empty, must be present anyway
Added note about that to README.

tests/boardfarm_plugins/README.md
arnout: ^^^ You can install without sudo

In addition, I really don't think it is necessary to install it. It is sufficient to install the prerequisites with `pip3 install -r requirements.txt`. This is important because installing it means that every change to boardfarm also requires re-install. Don't tell me it's not possible, I tried it :-). It does require some changes, as indicated elsewhere.

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,175 @@
...
+```
+-c /path/to/custom_config.json
+```
+
+### Devices
+There are several devices already existing in `<boardfarm-repo-root>/boardfarm/devices`, but if we created `custom_config.json` which uses previously unknown device, then we must implement its class in plugin dir `boardfarm_*/devices/<device-name>.py`.
+As template for the device we can use `<boardfarm-repo-root>/devices/base.py`.
+
+### Test suites
+Custom test suite should be present in root of boardfarm plugin dir and its name mustmatch `testsuites*.cfg`.
```
arnout:
```suggestion
Custom test suite should be present in root of boardfarm plugin dir and its name must match `testsuites*.cfg`.
```

tests/boardfarm_plugins/README.md
```diff
@@ -0,0 +1,175 @@
...
+        but it is suggested to leave it as a stub to prevent possible faults.
+        '''
+        print('{}: Recovering...'.format(self.__class__.__name__)
+
+    @classmethod
+    def teardown_class(cls):
+        '''Teardown method, optional for boardfarm tests.'''
+        print('{}: Teardown...'.format(cls.__class__.__name__))
+```
+For operational example please refer to **boardfarm_prplmesh** plugin.
```
arnout, line 169: :+1:

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_base.py
```diff
@@ -0,0 +1,49 @@
...
+
+import pexpect
+
+from boardfarm.devices import debian
+
+
+class PrplMeshBase(debian.DebianBox):
+    """PrplMesh abstract device."""
+
+    def _run_shell_cmd(self, cmd: str = "", args: list = None, timeout: int = 30):
```
arnout, line 14: (nitpick) For completeness, you could add **kwargs here, which are passed unchanged to pexpect.run.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_base.py
```diff
@@ -0,0 +1,49 @@
...
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+import pexpect
+
+from boardfarm.devices import debian
+
+
+class PrplMeshBase(debian.DebianBox):
```
arnout: What does DebianDevice give that LinuxDevice doesn't?

I tried with LinuxDevice and so far at least, it works.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
```diff
@@ -0,0 +1,9 @@
+{
+    "prplmesh_docker": {
+        "name": "dockerized_device",
+        "board_type": "prplmesh_docker",
+        "role": "controller",
+        "conn_cmd": "",
```
arnout, line 6: I believe this is only needed for DebianBox, not for LinuxDevice...

tests/run_bf.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+scriptdir=$(dirname "$(readlink -f "${0}")")
+bf_plugins_dir=${scriptdir}/boardfarm_plugins
+bf_dir=$(realpath "${scriptdir}/../../boardfarm/boardfarm")
```
arnout:
```suggestion
bf_dir=$(realpath "${scriptdir}/../../boardfarm")
```

tests/run_bf.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [ -n "${PYTHONPATH}" ]; then
+   PYTHONPATH="${bf_dir}:${bf_plugins_dir}:${scriptdir}:${PYTHONPATH}"
+else
+   PYTHONPATH="${bf_dir}:${bf_plugins_dir}:${scriptdir}"
+fi
+export PYTHONPATH
+
+exec bft -c "${bf_plugins_dir}"/boardfarm_prplmesh/prplmesh_config.json -n prplmesh_docker -x test_flows
```
arnout:
```suggestion
exec ${bf_dir}/bft -c "${bf_plugins_dir}"/boardfarm_prplmesh/prplmesh_config.json -n prplmesh_docker -x test_flows
```

With those two changes, it works for me without calling `setup.py`.

arnout (changes requested):

arnout:
Sorry, I pushed "Submit" before writing my overall commit.

There's just two changes I'd still like to see:
1. Modify the run_bf script so it uses boardfarm directly without requiring install. Note that it is not strictly necessary to modify README.md - it's still convenient if boardfarm is installed.
2. Check if LinuxDevice works as well as DebianBox, and if not, add an explanation somewhere (e.g. in a comment in PrplMeshBase) what we use from DebianBox.

arnout:
Oh, and there's one more thing: you need to add yourself to AUTHORS.md

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_base.py
taurha: > I tried with LinuxDevice and so far at least, it works.

I've checked - no difference when using `LinuxDevice` instead of `DebianBox`.
Replaced `DebianBox` by `LinuxDevice`.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
taurha: > I believe this is only needed for DebianBox, not for LinuxDevice...

Unfortunately - no.
Field `conn_cmd` is mandatory as it has no default value.
There's no dependency on device class, it's only boardfarm parser's issue.
When boardfarm attempts to parse custom JSON config without that field - exception will be thrown.

oponomar:
> 1. Modify the run_bf script so it uses boardfarm directly without requiring install. Note that it is not 
strictly necessary to modify README.md - it's still convenient if boardfarm is installed.

Updated run_bf script, so now it uses boardfarm local sources instead of installed module.

> 2. Check if LinuxDevice works as well as DebianBox, and if not, add an explanation somewhere (e.g. in a comment in PrplMeshBase) what we use from DebianBox.

I checked and there's no difference of using `LinuxDevice` instead of `DebianBox`, so I changed parent of `PrplMeshBase` class to  `LinuxDevice` as it`s higher in class hierarchy.

taurha:
> 1. Modify the run_bf script so it uses boardfarm directly without requiring install. Note that it is not 
strictly necessary to modify README.md - it's still convenient if boardfarm is installed.

Updated run_bf script, so now it uses boardfarm local sources instead of installed module.

> 2. Check if LinuxDevice works as well as DebianBox, and if not, add an explanation somewhere (e.g. in a comment in PrplMeshBase) what we use from DebianBox.

I checked and there's no difference of using `LinuxDevice` instead of `DebianBox`, so I changed parent of `PrplMeshBase` class to  `LinuxDevice` as it`s higher in class hierarchy.

MuteSpirit (approved):

arnout (approved):
Let's get this merged!

PR #1110, opened 2020-04-07
gitignore: add .vscode to .gitignore

adam1985d:
added .vscode to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ __pycache__/
 .quilt_checked
 .source_dir
 ipkg-*/
+.vscode


tomereli (changes requested):
IMOH this shouldn't be merged. We should add the .vscode to source control since we can all benefit from a single configuration.

vitalybu (changes requested):
This is not acceptable.
We don't add IDE specific files to the `.gitignore` file.
You can manage a global `.gitignore` for your specific needs on your development machine:
https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

Issue #1085, opened 2020-04-02
[BUG] agent: backhaul: 1905 forward on fail mechanism


RanRegev:
**Description**
Currently 1905 messages are handled in the backhaul in this manner:
calling a ```bool handle_1905_xxx()``` function and looking on the return value:
when true: the function handled the message, processing terminates.
when false: the function didn't handle the message, processing continues to agents.

The problem is that inside these functions, if error occurs they return `false`, therefore processing, in an error flow, is forwarded to the agents.

tomereli:
@morantr please comment

morantr:
Yes, it is bad implementation. Fortunately, I'm working on a unified Agent so this whole message forwarding mechanism will be removed.
I suggest closing this issue.

arnout:
Let's instead say that this will be fixed by #1082.

PR #1114, opened 2020-04-07
Feature/implement certification 4.7.9 i

RanRegev:
This PR contains the entire agent flow to pass certification test 4.7.9 with ETH backhaul.
It added two tlvs (beacon query and response), it takes care for sending the request only to the correct radio, and it handles the response.
The result of the full flow is a beacon measurements response, 1905 format, received at the controller. The data from the message is NOT stored or processed in any way.


agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4405,6 +4408,12 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
     }
 }
 
+
+bool slave_thread::handle_beacon_metrics_query_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    return false;
+}
+
```
vitalybu: Looking at the [entire commit](https://github.com/prplfoundation/prplMesh/pull/1114/commits/0bd383714bbdc6409d1d28366612f1f14ac706ec), it seems like there are far more changes then described in the commit messages.
If you decided to reorder the `switch/case` alphabetically, at least do it in a separate commit.

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -248,6 +248,7 @@ class slave_thread : public beerocks::socket_thread {
     bool handle_multi_ap_policy_config_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
     bool handle_client_association_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
     bool handle_client_steering_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
+    bool handle_beacon_metrics_query_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
```
vitalybu, line 251: Any reason to not pass the `cmdu_rx` as a const reference?
I know that other functions do not, but maybe they should...

framework/tlvf/tlvf_conf.yaml
```diff
@@ -10,30 +10,31 @@ include_yaml_path: {
   # "wfa_map/",
   "tlvf/wfa_map/eTlvTypeMap.yaml",
   "tlvf/wfa_map/tlvApCapability.yaml",
-  "tlvf/wfa_map/tlvChannelPreference.yaml",
-  "tlvf/wfa_map/tlvSupportedService.yaml",
+  "tlvf/wfa_map/tlvApHeCapabilities.yaml",
```
vitalybu, line 13: This commit should be split into 2 commits:
1. Reorder the lines alphabetically
2. Add your yaml file

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4408,10 +4409,58 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
     }
 }
 
-
-bool slave_thread::handle_beacon_metrics_query_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+bool slave_thread::handle_beacon_metrics_query_request(Socket *sd,
+                                                       ieee1905_1::CmduMessageRx &cmdu_rx)
```
vitalybu, line 4445: Did you run `clang_format`?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1652,7 +1654,8 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
         // Handle the CMDU message. If the message was processed locally
         // (by the Backhaul Manager), this function will return 'true'.
         // Otherwise, it should be forwarded to the slaves.
-        if (handle_1905_1_message(cmdu_rx, src_mac)) {
+        Socket *destinationAP = nullptr;
+        if (handle_1905_1_message(cmdu_rx, src_mac, destinationAP)) {
```
arnout, line 1685: I think this makes things overly complicated. We can just as well do the swap + forward_cmdu_to_uds directly from within the handle_1905_1_message function.

In fact, we can very easily solve #1085 by moving the forwarding entirely into handle_1905_1_message. Almost all the `return false` cases that are in there are wrong - they are error conditions where no forwarding should be done.

And if it turns our there are several places where forwarding needs to be done, it can easily be handled by a version of forward_cmdu_to_uds() that doesn't take the destination argument.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2885,6 +2901,72 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+
+        if (radioMac) {
+            LOG(DEBUG) << "found the mac of the radio that has the sattion. radio: " << *radioMac << "; station: " << requiredMac;
+            auto it = std::find_if(slaves_sockets.begin(), slaves_sockets.end(),
+                                   [&radioMac](const std::shared_ptr<SSlaveSockets> slave) {
+                                       return network_utils::mac_from_string(slave->radio_mac) == *radioMac;
+                                   });
+            if (it != slaves_sockets.end()) {
+                LOG(DEBUG) << "found the socket to send to the agent that will send to:" << *radioMac << " for STA: " << requiredMac;
+                forward_to = (*it)->slave;
```
arnout: This would be a lot simpler if the sRadioInfo and the SSlaveSockets structures were merged...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2885,6 +2901,72 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+                LOG(DEBUG) << "didn't find the socket to send to :" << *radioMac << "; this is strange";
+            }
+        }
+        else
+        {
+            LOG(DEBUG) << "couldn't find any agent for the requested mac: " << requiredMac; 
+        }
+    } else {
+        LOG(ERROR) << "handle_1905_beacon_metrics_query should handle only tlvBeaconMetrics, but got something else: " <<
+            std::hex << (uint16_t)cmdu_rx.getMessageType();
```
arnout: The message type and the TLV type are distinct. So the correct error message would be:
```
        LOG(ERROR) << "getClass wfa_map::tlvBeaconMetricsQuery failed";
```
or something like that.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4405,6 +4409,80 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+        return false;
+    }
+
+    // get the correct type of the message
+    auto beaconMetricsQuery = cmdu_rx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+
+    if (!beaconMetricsQuery) {
+        LOG(ERROR) << "tlvBeaconMetricsQuery missing - ignoring beacon metrics query message";
+        return false;
+    }
```
arnout: This is part of checking the input, so it makes more sense to do that *before* creating request_out.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4405,6 +4409,80 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+    */
+
+    measurement_request.expected_reports_count = 1;
+
+    measurement_request.rand_ival = beerocks::BEACON_MEASURE_DEFAULT_RANDOMIZATION_INTERVAL;
+    measurement_request.sta_mac   = beaconMetricsQuery->associated_sta_mac();
+
+    message_com::send_cmdu(monitor_socket, cmdu_tx);
+
+    // where is the response???
```
arnout: The response doesn't come here. The flow is as follows:
```
controller -> agent: Beacon Metrics Query
agent -> controller: ACK
agent -> station: Beacon Metrics Query
station -> station: scan for beacon specified in query
station -> agent: Beacon Metrics Report
agent -> controller: Beacon Metrics Response
```

(In case this wasn't clear yet: beacon metrics is part of 802.11k.)

In other words, we have to "wait" for the station to send its report. But of course we're not going to wait here. Instead, there will be an event generated by the monitor when the station sends the report, and the monitor thread generates the response message when it handles that event.

There is BTW an interesting additional problem here: the query doesn't necessarily come from the controller, an agent is also allowed to perform a beacon metrics query on another agent. So you would need to store the requesting AL-MAC address to be able to respond to the correct entity. Fortunately, this isn't tested in certification, so we can get away with always responding to the controller.

framework/transport/ieee1905_transport/ieee1905_transport_local_bus.cpp
```diff
@@ -81,7 +81,7 @@ void Ieee1905Transport::handle_local_bus_cmdu_tx_message(CmduTxMessage &msg)
         }
 
         if (msg.metadata()->msg_type != ntohs(ch->messageType)) {
-            MAPF_WARN("CmduTxMessage messageType mismatch.");
+            MAPF_WARN("CmduTxMessage messageType mismatch: cmdu: " << std::hex << msg.metadata()->msg_type << " packet: " << ntohs(ch->messageType));
```
arnout: (nitpick) if it's hex, please add 0x.

tests/test_flows.py
```diff
@@ -733,13 +733,14 @@ def test_client_capability_query(self):
                         .format(sta_mac2 = sta_mac2,
                                 mac_repeater1_wlan0 = self.mac_repeater1_wlan0))
 
-    def test_client_association_dummy(self):
-        sta_mac = "11:11:33:44:55:66"
+    def test_client_association_dummy(self, sta_mac="11:11:33:44:55:66"):
```
arnout: These changes are no longer necessary.

tests/test_flows.py
```diff
@@ -978,6 +980,23 @@ def test_topology(self):
...
+        repeater = self.repeater1
+        mac_repeater = self.mac_repeater1
+
+        debug("Connect dummy STA to wlan0")
+        self.send_bwl_event(repeater, "wlan0", "EVENT AP-STA-CONNECTED {}".format(sta_mac))
+
+        # send beacon query request
+        debug("Sending beacon report query to repeater:")
+        request='{mac} 0x73 0xFFFFFFFFFFFF 0x02 0x00 0x01 0x03 0x73 0x24 0x30'.format(mac=sta_mac)
+        debug(request)
```
arnout: Maybe it would be better to extend dev_send_1905 to always print the command if debug is enabled.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4405,6 +4409,80 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+    /*
+	measurement_request.measurement_mode		= beerocks::MEASURE_MODE_PASSIVE;
+	measurement_request.duration				= beerocks::BEACON_MEASURE_DEFAULT_PASSIVE_DURATION;
+    */
+
+    measurement_request.expected_reports_count = 1;
+
+    measurement_request.rand_ival = beerocks::BEACON_MEASURE_DEFAULT_RANDOMIZATION_INTERVAL;
+    measurement_request.sta_mac   = beaconMetricsQuery->associated_sta_mac();
+
```
arnout, line 4524: You will also need to set:
- op_class
- repeats (to 0 or 1 I guess)
- parallel (to 0)
- enable (I think to 0)
- request (I think to 1)
- report (I think to 0)
- mandatory_duration (I think to 0)
- use_optional_ssid, ssid (based on ssid in query)
- use_optional_ap_ch_report (based on channel reports in query)
- use_optional_req_elements, req_elements (based on elements list in query)
- use_optional_wide_band_ch_switch
- new_ch_width
- new_ch_center_freq_seg_0
- new_ch_center_freq_seg_1

While I'm developing, I usually copy the relevant part of the struct (or in this case, of the yaml) into a comment in the code to be sure I don't forget anything.

There may be some fields there that are actually not used, e.g. new_ch_width etc. don't seem to be used. I would add a separate commit that removes that stuff before you begin.

Finally, we also have to think what to do with the reporting detail field. We might have to add it to the struct.

ghost:
Could you fix the PR title, and add a description?
Thanks :-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: So I did that -> #1131

arnout:
> Could you fix the PR title, and add a description?

That's not necessary for a draft PR.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Remember that this is a draft PR, so you can't reliably give feedback on commit messages, commit splitting or commit ordering.

agent/src/beerocks/slave/son_slave_thread.h
RanRegev: Indeed. sometimes there is a tense between doing things _correctly_ vs. _uniformly_.
The tense rises from the fact that both have positive implications.
The "spirit" that I got from this project up till now is that uniform wins.

framework/tlvf/tlvf_conf.yaml
RanRegev: 👍

agent/src/beerocks/slave/son_slave_thread.cpp
RanRegev: I run it fro time to time, not necessarily with each commit.
I think we should consider adding clang-format _per compilation_.
i.e. automatically format only the files that were changed.
I guess currently it is not done this way since formatting and compiling are done in two different environments.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: I agree, this forwarding mechanism is error prone, mostly because the return value true/false does not indicates success or fail in this case, but rather it signals the caller function what it should do next.
I think @morantr is working on rewriting this area.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: I took it.
But it didn't simplify much the search process.

agent/src/beerocks/slave/son_slave_thread.cpp
RanRegev: My

> // where is the response???

And your _detailed_ answer is the reason I wanted to postponed the pull-request.
For me, it was just a kind of a reminder to handle the response. And also to notify myself that I'm not familiar with the flow in the code that handles responses. The comment is shown in diffs, I see it whenever I visit the function, and in general it is my own personal way to remind myself that there is more to do in this task.
For you, on the other hand, it indicates that the writer (me) does not understand at ALL what is going on here (hence your detailed response). 
:-)

tests/test_flows.py
RanRegev: I vote yes!

agent/src/beerocks/slave/son_slave_thread.cpp
RanRegev: The way it is done today is error prone and I hope it will be fixed (the sooner the better).
The main problem is this:
We have two languages - 1905 and vendor specific - that in general describe the same (this is not the problem, wait.)
We decided to translate from one to the other when implementing certain flows. This is also fine, we are reusing existing implementation, we talk the language that we feel comfortable with and translate only at the edges (certification tests in this case).
However, writing the translation deep inside the code of specific implementation of specific flow is wrong. The translation has nothing to do with the flow that _uses_ it.
The correct way to do it is to have a separate library (normally named `bridge` or alike) that does the conversions back and forth, as needed, between two different domains.
The goal is to have the developer focusing on the logic of the flow and not on domains translations. e.g.:
```
beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST msg;
ieee1905_1::CmduMessageRx rx // the given 1905 message;

// load vendor specific from 1905 message
// operator<<() in this case is a free function taking 
// objects from both domains and converts between them
msg << rx; 

```

framework/transport/ieee1905_transport/ieee1905_transport_local_bus.cpp
RanRegev: This log:

> Received ACK_MESSAGE, mid=0x8000

Was created by this code
`LOG(DEBUG) << "Received ACK_MESSAGE, mid=" << std::hex << int(mid);`

On the other hand, the [standard](https://en.cppreference.com/w/cpp/io/manip/hex) says nothing about auto adding 0x. So I guess LOG does it.
But yet on another hand, you asked to add 0x.
All in all - I'm confused
:-)

arnout: I was confused for a while as well... But it turns out that easylogging++ already does a `showbase` on the stream, so no need for an explicit `0x`. Sorry for the confusion!

agent/src/beerocks/slave/son_slave_thread.cpp
RanRegev: I'll see how I squash, unite, remove or whatever some of the commits.
The goal - clean history.
I chose to develop without too much care about the amount of commits, changing back and forth and so on. I just worked on the code and let git save the changes for me.
I'll fix it into another branch before merging to master.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: I see what you mean.
They are two different things: 
a. the message type 
b. the tlv it carries with it.
However, they always come together.
I'll fix the LOG.

RanRegev:
After fixing the commit history, this PR is closed.
The active one is PR #1172

PR #1115, opened 2020-04-07
CODINGSTYLE.md: Add multiple vs single point of exit

tomereli:
Based on the poll done in #prplmesh-democracy on slack, the majority
prefers to use multiple points of exit from functions.
Add a section for this in CODINGSTYLE.md which explains the pros and
cons of using single point of exit vs multiple points of exit, with
examples and which pattern we expect contributors to use.

### DIFF
--- a/CODINGSTYLE.md
+++ b/CODINGSTYLE.md
@@ -196,6 +196,93 @@ In the following example, the empty string is the sentinel value indicating fail
 std::string parse_bss_info(const std::string &bss_info_str, db::bss_info_conf_t &bss_info_conf);
 ```
 
+Prefer to use multiple points of exit from functions.  
+One might argue that it is less confusing if there is only a single exit point, but it leads to more nesting and therefore more indentation which makes the code less readable.  
+Another advantage of using multiple points of exit is that you can check preconditions and exit early at the start of a function, so that you know in the body of the function that certain conditions are true, without the entire body of the function being indented 5 miles off to the right.  
+This usually minimizes the number of scopes you have to worry about, which makes code much easier to follow.  
+Multiple points of exit means that you can exit anywhere you please.  
+This used to be more confusing in the old days, but now that we have syntax-coloring editors and compilers that detect unreachable code, it's a lot easier to deal with.
+
+Avoid using `goto` - The goto failure pattern is a corollary of the multiple exit points.  
+It applies when there is cleanup to do, but since we use C++, there should never be any cleanup to do (it should all be handled by automatic variables).
+
+Examples:
+
+Good - multiple points of exit:
+
+```cpp
+bool func()
+{
+    if (something) {
+        // do something
+        return true;
+    } else if (something else) {
+        // do something else
+        return false;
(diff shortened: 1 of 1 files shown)


CODINGSTYLE.md
```diff
@@ -196,6 +196,75 @@ In the following example, the empty string is the sentinel value indicating fail
...
+    if (something) {
+        // do something
+        return true;
+    } else if (something else) {
+        // do something else
+        return false;
+    } else {
+        return false;
+    }
+}
```
vitalybu: Your function returns an `int` but internally use return `true`/`false` :)
Consider adding another good example:
```cpp
bool func(int arg1)
{
    // validate arguments
    if (arg1 < 0) {
        return false;
    }

    // do something

    return true;
}
```

The idea is to not use an `else` clause in this case to reduce indentation levels.

CODINGSTYLE.md
```diff
@@ -196,6 +196,75 @@ In the following example, the empty string is the sentinel value indicating fail
...
+  // do something
+  if (some condition)
+    goto failure;
+  }
+  result = true; // success
+  goto success;
+failure:
+  cleanup();
+success:
+  return result;
```
vitalybu: Be consistent in your examples. Either use 2 or 4 (preferably) spaces for indentation.

vitalybu (changes requested):

vitalybu (approved):

CODINGSTYLE.md
```diff
@@ -196,6 +196,90 @@ In the following example, the empty string is the sentinel value indicating fail
 std::string parse_bss_info(const std::string &bss_info_str, db::bss_info_conf_t &bss_info_conf);
 ```
 
+Prefer to use multiple points of exit from functions.
+One might argue that it is less confusing if there is only a single exit point, but it leads to more nesting and therefore more indentation which makes the code less readable.
+Another advantage of using multiple points of exit is that you can check preconditions and exit early at the start of a function, so that you know in the body of the function that certain conditions are true, without the entire body of the function being indented 5 miles off to the right. This usually minimizes the number of scopes you have to worry about, which makes code much easier to follow.
```
rmelotte: Please split this to have one sentence per line, it's in CODINGSTYLE.md already :-) 
>Lines in markdown documents are not wrapped, but every sentence should start on a new line.

CODINGSTYLE.md
```diff
@@ -196,6 +196,90 @@ In the following example, the empty string is the sentinel value indicating fail
...
+{
+    if (something) {
+        // do something
+        return true;
+    } else if (something else) {
+        // do something else
+        return false;
+    } else {
+        return false;
+    }
```
rmelotte, line 224: IMO this doesn't really show your point of keeping the indentation level low or checking arguments early, so maybe we should keep only the other example.
On the other hand, it shows what having multiple points of exit _means_, I guess that's why you did it like this.

rmelotte (approved):

PR #1119, opened 2020-04-08
agent: backhaul_manager: use reference to map entry to persist changes

### LINKED ISSUE #1097: [BUG] No Topology Response message found

mariomaz:
When a ACTION_BACKHAUL_ENABLE message is received, the map entry in `m_radio_info_map` corresponding to the interface needs to be updated.
    
Map entry must be obtained by reference, not by value, or otherwise a copy of the entry is used and changes made to it are lost.

Fixes: #1097

### DIFF
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -1562,12 +1562,15 @@ void ap_manager_thread::handle_hostapd_attached()
     copy_radio_supported_channels(ap_wlan_hal, notification->params().supported_channels);
 
     LOG(INFO) << "send ACTION_APMANAGER_JOINED_NOTIFICATION";
+    LOG(INFO) << " iface = " << ap_wlan_hal->get_iface_name();
     LOG(INFO) << " mac = " << ap_wlan_hal->get_radio_mac();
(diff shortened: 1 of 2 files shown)


tomereli (approved):

CoralMalachi (approved):
https://gitlab.com/prpl-foundation/prplMesh/-/jobs/503377257
I see it solves the issue :)

PR #1120, opened 2020-04-08
Feature/ci certification tests manual jobs

### LINKED ISSUE #759: [TASK] CI: split certification tests per device
## Description

Instead of requiring the users to enter the DUT for each test they want to run, it would be more convenient to have a separate manual job for each device.

## Exit criteria

- There is a separate job to manually run the certification tests with each device.
- The Gitlab schedules can still choose what device to run the tests with

### LINKED ISSUE #761: [TASK] CI: split the certification tests into small groups
## Description

Splitting the certification tests into small groups of test has some advantages:
- if a test fail, you don't need to wait for many other tests to finish before being able to get the artifacts
- it's easier to later enable auto-retry if each job run only a small number of tests (if a test fail, no need to retry 20 others).

Related to https://github.com/prplfoundation/prplMesh/issues/759

## Exit criteria

- The tests are grouped into small groups of 1 to 6 tests
- The Gitlab schedules also runs the tests in small groups, but can still run all of the agent tests, or all of the controller tests with the target it chooses (without creating one schedule per group).
- The groups use Gitlab's [grouped jobs feature](https://docs.gitlab.com/ee/ci/pipelines.html#grouping-jobs)

### LINKED ISSUE #1007: [TASK] CI, find out variables used to run a test, when the test initiated from the web interface
**A way to trace back the values of variables, e.g. TEST_TO_RUN, when running CI pipelines from the web interface (https://gitlab.com/prpl-foundation/prplMesh/pipelines/new)**
- Looking at tests results and relying on the good will of the initiator (even if it is you) to remember what were the values, is not a good practice. You may be solving a non-problem or looking at output that does not match your expectations.

**Alternatives**
- Running from the command line has both the command line history AND a web presentation of the variables.
- Splitting the tests into non-variable commands. Each test on its own.

**Additional context**
The problem occurred to me the first time I tried to run a test.
I couldn't verify I ran it correctly.

rmelotte:
This is something I've quickly put together in the background to make it simpler for anyone to trigger certification tests.

With these changes:
- If the last commit of a pipeline contains the name of a test job (for example: `MAP-5.3.1:netgear-rax40`), the test is run automatically.
  If the test fails, the whole pipeline will be marked as failing.

  This will be very useful to trigger certification tests automatically before a pull request is merged. 
  If a developer is making changes that could affect a particular certification test(s), the test name(s) can just be mentioned in the commit description of the last commit, and the PR will be blocked until the certification test passes.

- It's no longer necessary to trigger a whole new pipeline just to run a certification test.
  Previously when multiple people were working at the same time, we had a lot of duplicate pipelines: one was created automatically by a push, and the other one was created only to run a test. This quickly slowed things down unnecessarily.

  From now on, every pipeline would have a list of tests that can be triggered from the web interface. In other words, even if the commits don't mention the test names, we could trigger a test for any branch at any time:
  image
 (the image is truncated to save space)

  With this, if we wanted/needed to, we could also easily allow any project member with the "developer" role to trigger certification tests on the master branch (using [environments](https://docs.gitlab.com/ee/ci/yaml/#protecting-manual-jobs-premium)), which AFAIK was not possible before and has been requested by multiple other developers in the past.

- It makes it a whole lot easier to understand what is being run (fixes https://github.com/prplfoundation/prplMesh/issues/1007)

- When running multiple tests, we would no longer have to wait until they are all finished before we can get the artifacts (the logs).
  It also makes it easier to make use of Gitlab's auto-retry mechanism.
  fixes https://github.com/prplfoundation/prplMesh/issues/761.

Also fixes https://github.com/prplfoundation/prplMesh/issues/759.

Keeping it as a draft for now, because the number of total jobs for active pipelines on Gitlab.com are potentially limited for free plans, and it could be a blocking issue.

### DIFF
--- a/ci/certification/dummy.yml
+++ b/ci/certification/dummy.yml
@@ -1,0 +1,238 @@
+.certification:dummy:
+  stage: certification-dummy
+  extends: .certification-generic
+  variables:
+    DEVICE_UNDER_TEST: "prplmesh"
+  needs:
+    - job: build-in-docker
+
+MAP-5.3.1:dummy:
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-5.3.1:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-5.4.1_ETH_FH24G:dummy:
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-5.4.1_ETH_FH24G:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-5.4.1_ETH_FH5GL:dummy:
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-5.4.1_ETH_FH5GL:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-5.4.1_ETH_FH5GH:dummy:
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-5.4.1_ETH_FH5GH:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-5.4.2_ETH_FH24G:dummy:
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-5.4.2_ETH_FH24G:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-5.4.2_ETH_FH5GL:dummy:
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-5.4.2_ETH_FH5GL:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-5.4.2_ETH_FH5GH:dummy:
+  extends: .certification:dummy
+  rules:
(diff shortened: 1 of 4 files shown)


ci/certification/dummy.yml
```diff
@@ -0,0 +1,1798 @@
+.certification:dummy:
+  stage: certification-dummy
+  extends: .certification-generic
+  variables:
+    DEVICE_UNDER_TEST: "prplmesh"
+  needs:
+    - job: build-in-docker
+
+MAP-4.2.1:dummy:
```
arnout: I think we shouldn't generate the agent tests for dummy, should we?

ci/certification/dummy.yml
```diff
@@ -0,0 +1,1798 @@
...
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-4.2.1:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
```
arnout: It really sucks that this can't move to .certification-generic... But indeed, seems to be a limitation of the gitlab syntax.

arnout (approved):
As mentioned, we will have to wait for the Gold license to come through before we merge this, otherwise we're going to go over 500 jobs pretty quickly.

However, since this is so useful, maybe you can pare it down to 30-ish jobs which are relevant at this time? That means:
- netgear only;
- 4.2.1
- 4.2.3
- 4.4.2
- 4.5.2
- 4.6.1
- 4.6.2
- 4.7.3
- 4.7.4
- 4.7.5
- 4.7.6
- 4.7.7
- 4.7.9
- 4.8.1
- 4.8.2
- 4.8.3
- 4.8.5
- 4.10.3
- 4.10.4
- 4.11.1

And one band is sufficient. If someone needs an additional band, they can update the file pretty easily.

ci/certification/dummy.yml
```diff
@@ -0,0 +1,238 @@
...
+
+MAP-5.3.1:dummy:
+  extends: .certification:dummy
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-5.3.1:dummy.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
```
ghost, line 17: This probably betrays my general gitlab ignorance, but:
Why allow failure? Don't we *want* a test failure to block the PR?

ghost (approved):
Nothing else really jumps out at me, so :+1:  looking good :)

ci/certification/dummy.yml
rmelotte: Allow failure is for the manual jobs only. I'm not sure what the reasoning behind it is, but manual jobs that have allow_failure=false make the whole pipeline appear as "blocked" and need the jobs to be triggered before the pipeline is reported as "passing".

For the jobs that run automatically (so if the test is mentioned in the commit message) PRs will block until the test passes (even after an automatic rebase).
So I think for PRs we should encourage mentioning every potentially affected test names in commits messages, and use the manual jobs for the other cases (sporadically check that a test still passes for example, or check the status of a not-yet-passing test).

arnout: > Allow failure is for the manual jobs only. I'm not sure what the reasoning behind it is, but manual jobs that have allow_failure=false make the whole pipeline appear as "blocked" and need the jobs to be triggered before the pipeline is reported as "passing".

That's a bit of historical accident...

Originally, the "manual" was meant to do block (part of) a pipeline and require manual intervention to launch it, as a kind of "approval" step. For example, you could have a job "review" that needs to be clicked by a human to allow the pipeline to succeed.

Using "manual" for the use case we have here is something that was only thought of afterwards, and they used the already existing "allow_failure" tag to mark it.

PR #1103, opened 2020-04-07
Deprecated shell scripts cleanup

MuteSpirit:
Delete deprecated shell scripts mentioned in task #915 and not generated during build

PR #1123, opened 2020-04-09
Fix shellcheck errors in scripts (part 1)

### LINKED ISSUE #915: [TASK] Perform shellcheck on shell scripts
Shell scripts are risky, but the `shellcheck` tool can help to find issues with it.
- [x] (merged) Delete deprecated scripts
- [x] (**all on review**) Fix existing shellcheck issues
          (number of shellcheck issues)
   - [x] ( 1) ./ci/owncloud/upload_to_owncloud.sh
   - [x] ( 1) ./ci/run_easymesh_cert.sh
   - [x] ( 2) ./docker-builder.sh
   - [x] ( 2) ./docker-runner.sh
   - [x] ( 4) ./framework/platform/bpl/uci/scripts/platform_daemon.sh
   - [x] ( 1) ./tests/openwrt/test_status.sh
   - [x] ( 2) ./tests/test_gw_repeater.sh
   - [x] ( 6) ./tools/docker/builder/openwrt/build.sh
   - [x] ( 1) ./tools/docker/builder/rdkb/toolchain/bitbake.sh
   - [x] ( 6) ./tools/docker/build.sh
   - [x] ( 1) ./tools/docker/clang-format.sh
   - [x] (18) ./tools/docker/image-build.sh
   - [x] (10) ./tools/docker/image-pull.sh
   - [x] ( 1) ./tools/docker/runner/entrypoint.sh
   - [x] ( 3) ./tools/docker/run.sh
   - [x] ( 4) ./tools/docker/static-analysis/cppcheck.sh
   - [x] ( 2) ./tools/docker/stop.sh
   - [x] ( 7) ./tools/docker/test.sh
   - [x] ( 3) ./tools/functions.sh
   - [x] ( 2) ./tools/klocwork/kw.sh

- [x] (Canceled) Add shellcheck to the builder docker image
- [x] (**on review**) Run shellcheck in gitlab CI.

## Total lists
**Note:** Lists below will be updated based on discussion
Removal candidates:
-`./build/common/beerocks/scripts/prplmesh_utils.sh`
-`./build/install/scripts/prplmesh_utils.sh`
-`./build/out/scripts/prplmesh_utils.sh`
-`./common/beerocks/scripts/dhcp.sh`
-`./docker-builder.sh`
-`./docker-runner.sh`
-`./tests/openwrt/test_status.sh`
-`./tools/maptools.sh`
-`./tools/scripts/prplMesh-platform-init.sh`
-`./tools/commands/deploy_rdkb.sh`
-`./tools/commands/deploy_ugw.sh`
-`./framework/platform/bpl/uci/scripts/platform_daemon.sh`

No action due to soon removal:
-`./tests/test_flows.sh`
-`./tests/test_gw_repeater.sh`

Scripts not mentioned in any of lists above will be fixed.

MuteSpirit:
scripts: fix shellcheck issues in tools/docker/image-pull.sh

Will contain one commit only after merge #1103

### DIFF
--- a/tests/test_gw_repeater.sh
+++ b/tests/test_gw_repeater.sh
@@ -38,7 +38,7 @@ check_wsl() {
     if ! grep -q Microsoft /proc/version; then return; fi
 
     status "Running in WSL"
-    
+
     # Read GW & Repeater UCC ports
     local GW_UCC_PORT
     GW_UCC_PORT=$(grep ucc_listener_port \
@@ -56,7 +56,7 @@ check_wsl() {
 
 main() {
     if ! OPTS=$(getopt -o 'hvd:fg:r:u:' --long help,verbose,rm,gateway-only,repeater-only,delay:,force,gateway:,repeater:,unique-id: -n 'parse-options' -- "$@"); then
-        err "Failed parsing options." >&2 ; usage; exit 1 ; 
+        err "Failed parsing options." >&2 ; usage; exit 1 ;
     fi
 
     eval set -- "$OPTS"
@@ -96,7 +96,7 @@ main() {
 
     [ "$START_GATEWAY" = "true" ] && {
         status "Start GW (Controller + local Agent)"
-        "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} ${GW_EXTRA_OPT} \
+        "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} "${GW_EXTRA_OPT[@]}" \
             start-controller-agent -d -n "${GW_NAME}" -- "$@"
     }
 
@@ -108,7 +108,7 @@ main() {
     [ "$START_REPEATER" = "true" ] && {
         for repeater in $REPEATER_NAMES; do
             status "Start Repeater (Remote Agent): $repeater"
-            "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} ${RP_EXTRA_OPT} \
+            "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} "${RP_EXTRA_OPT[@]}" \
                 start-agent -d -n "${repeater}" -- "$@"
         done
     }
@@ -118,14 +118,14 @@ main() {
 
     error=0
     [ "$START_GATEWAY" = "true" ] && report "GW operational" \
-        "${rootdir}"/tools/docker/test.sh "${VERBOSE_OPT}" -n "${GW_NAME}"
+        "${rootdir}"/tools/docker/test.sh ${VERBOSE_OPT} -n "${GW_NAME}"
 
 
     [ "$START_REPEATER" = "true" ] && {
         for repeater in $REPEATER_NAMES
         do
             report "Repeater $repeater operational" \
-            "${rootdir}"/tools/docker/test.sh "${VERBOSE_OPT}" -n "${repeater}"
+            "${rootdir}"/tools/docker/test.sh ${VERBOSE_OPT} -n "${repeater}"
         done
     }
 
--- a/tools/docker/image-pull.sh
+++ b/tools/docker/image-pull.sh
@@ -6,13 +6,14 @@
 # See LICENSE file for more details.
 ###############################################################
 
-scriptdir="$(cd "${0%/*}"; pwd)"
+scriptdir="$(cd "${0%/*}" || exit 1; pwd)"
 rootdir="${scriptdir%/*/*}"
 
-. ${rootdir}/tools/functions.sh
+# shellcheck source=../../tools/functions.sh
+. "${rootdir}/tools/functions.sh"
 
 usage() {
-    echo "usage: $(basename $0) [-hvbt]"
+    echo "usage: $(basename "$0") [-hvbt]"
     echo "  options:"
     echo "      -h|--help - show this help menu"
     echo "      -v|--verbose - verbosity on"
@@ -33,15 +34,16 @@ pull_and_tag() {
 }
 
 main() {
-    OPTS=`getopt -o 'hvb:t:' --long verbose,help,base-image,tag -n 'parse-options' -- "$@"`
-
-    if [ $? != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
+    if ! OPTS=$(getopt -o 'hb:t:' --long help,base-image,tag -n 'parse-options' -- "$@"); then
+        err "Failed parsing options." >&2
+        usage
+        exit 1
+    fi
 
     eval set -- "$OPTS"
 
     while true; do
         case "$1" in
-            -v | --verbose)         VERBOSE=true; shift ;;
             -h | --help)            usage; exit 0; shift ;;
             -b | --base-image)      IMAGE="$2"; shift ; shift ;;
             -t | --tag)             TAG=":$2"; shift ; shift ;;
@@ -61,7 +63,6 @@ main() {
     pull_and_tag runner "$base_image" "$TAG"
 }
 
-VERBOSE=false
 IMAGE="ubuntu:18.04"
 
-main $@
+main "$@"
--- a/tools/docker/image-pull.sh
+++ b/tools/docker/image-pull.sh
@@ -17,7 +17,6 @@ usage() {
     echo "      -h|--help - show this help menu"
     echo "      -v|--verbose - verbosity on"
     echo "      -b|--base-image - Base OS image to use (Dockerfile 'FROM')"
-    echo "      -n|--native - Use the same base OS image as the running system"
     echo "      -t|--tag - tag to add to prplmesh-builder and prplmesh-runner images"
 }
 
@@ -34,7 +33,7 @@ pull_and_tag() {
 }
 
 main() {
-    OPTS=`getopt -o 'hvnb:t:' --long verbose,help,base-image,tag -n 'parse-options' -- "$@"`
+    OPTS=`getopt -o 'hvb:t:' --long verbose,help,base-image,tag -n 'parse-options' -- "$@"`
 
     if [ $? != 0 ] ; then err "Failed parsing options." >&2 ; usage; exit 1 ; fi
 
@@ -45,11 +44,6 @@ main() {
             -v | --verbose)         VERBOSE=true; shift ;;
             -h | --help)            usage; exit 0; shift ;;
             -b | --base-image)      IMAGE="$2"; shift ; shift ;;
-            -n | --native)          IMAGE=$(
-                                        . /etc/os-release
-                                        distro="$(echo $NAME | awk '{print tolower($0)}')"
-                                        echo "$distro:$VERSION_ID"
-                                    ); shift ;;
             -t | --tag)             TAG=":$2"; shift ; shift ;;
             -- ) shift; break ;;
             * ) err "unsupported argument $1"; usage; exit 1 ;;
@@ -68,7 +62,6 @@ main() {
 }
 
 VERBOSE=false
-NATIVE=false
 IMAGE="ubuntu:18.04"
 
 main $@
--- a/tools/docker/runner/entrypoint.sh
+++ b/tools/docker/runner/entrypoint.sh
@@ -27,5 +27,5 @@ run ip link set      dev wlan2    up
 run ip address add   dev br-lan "$bridge_ip"
 run ip link set      dev br-lan   up
 
-cd ${INSTALL_DIR}
+cd "${INSTALL_DIR}" || exit 1
 exec /bin/bash "$@"
--- a/tools/functions.sh
+++ b/tools/functions.sh
@@ -1,3 +1,4 @@
+# shellcheck shell=sh
 ###############################################################
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 # SPDX-FileCopyrightText: 2019-2020 the prplMesh contributors (see AUTHORS.md)
@@ -5,10 +6,11 @@
 # See LICENSE file for more details.
 ###############################################################
 
-installdir="${rootdir}/build/install"
+# shellcheck disable=SC2034
+installdir="${rootdir:?}/build/install"
 
 dbg() {
-    if [ "$VERBOSE" = "true" ]; then echo "$(basename $0): $*"; fi
+    if [ "$VERBOSE" = "true" ]; then echo "$(basename "$0"): $*"; fi
 }
 
 status() {
--- a/tools/klocwork/kw.sh
+++ b/tools/klocwork/kw.sh
@@ -21,7 +21,7 @@ declare PLATFORM_TYPE
 ROOT_PATH=$(realpath "$(pwd)"/../../)
 PLATFORM_TYPE=$PRPLMESH_PLATFORM_TYPE
 
-echo platfrom identified: $PLATFORM_TYPE
+echo platform identified: "$PLATFORM_TYPE"
 
 # Remove old compilation script
 rm -rf _GO_KW
--- a/ci/owncloud/upload_to_owncloud.sh
+++ b/ci/owncloud/upload_to_owncloud.sh
@@ -6,10 +6,13 @@
 # See LICENSE file for more details.
 ###############################################################
 
-# shellcheck source=../../tools/functions.sh
-. "$(dirname "${BASH_SOURCE[0]}")/../../tools/functions.sh"
-# shellcheck source=./owncloud_definitions.sh
-. "$(dirname "${BASH_SOURCE[0]}")/owncloud_definitions.sh"
+scriptdir=$(cd "${0%/*}" || exit 1; pwd)
+rootdir="${scriptdir%/*/*}"
+
+# shellcheck source=tools/functions.sh
+. "${rootdir}/tools/functions.sh"
+# shellcheck source=ci/owncloud/owncloud_definitions.sh
+. "${rootdir}/ci/owncloud/owncloud_definitions.sh"
 
 usage() {
     echo "usage: $(basename "$0") [-hv] <user> <password> <remote-path> <local-path>"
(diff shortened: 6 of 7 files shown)


ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
 
-scriptdir="$(cd "${0%/*}" && pwd)"
-rootdir="${scriptdir%/*/*}"
+scriptdir=$(dirname "$(readlink -f "$0")")
```
arnout: Which shellcheck error are you solving here? The original is the typical way of getting the current directory, why change it?

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -1,9 +1,10 @@
...
+rootdir=$(realpath "$scriptdir/../..")
 
-# shellcheck source=../../tools/functions.sh
+# shellcheck source=tools/functions.sh
```
arnout: The shellcheck invocation that is used inside an IDE relies on a path that is relative to the script itself, not to the root directory.

Fortunately, you can simply use `shellcheck -P SCRIPTDIR` to allow paths relative to the script.

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -1,9 +1,10 @@
...
+# shellcheck disable=SC1091
```
arnout: No no no - make sure you include functions.sh in your shellcheck invocation and this won't be needed.

tools/docker/image-pull.sh
```diff
@@ -45,11 +46,7 @@ main() {
             -v | --verbose)         VERBOSE=true; shift ;;
             -h | --help)            usage; exit 0; shift ;;
             -b | --base-image)      IMAGE="$2"; shift ; shift ;;
-            -n | --native)          IMAGE=$(
-                                        . /etc/os-release
-                                        distro="$(echo $NAME | awk '{print tolower($0)}')"
-                                        echo "$distro:$VERSION_ID"
-                                    ); shift ;;
+            -n | --native)          NATIVE=true; shift; ;;
```
arnout: Which shellcheck error is this trying to fix? That should be part of the commit message, ideally.

If it is just a refactoring because you think it's cleaner that way, do that in a separate commit. That way, it's easier to see what is really the shellcheck fix.

However, I don't think anybody uses the native option - it's an idea I had a long time ago but it doesn't really work. So instead, just remove the option.

tools/functions.sh
```diff
@@ -1,14 +1,15 @@
+#!/bin/sh
```
arnout: This script is not meant to be executed, so [use a directive](https://github.com/koalaman/shellcheck/wiki/SC2148) `# shellcheck shell=sh`.

tools/functions.sh
```diff
@@ -1,14 +1,15 @@
...
 ###############################################################
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 # Copyright (c) 2019 Tomer Eliyahu (Intel)
 # This code is subject to the terms of the BSD+Patent license.
 # See LICENSE file for more details.
 ###############################################################
 
-installdir="${rootdir}/build/install"
+export installdir="${rootdir:?}/build/install"
```
arnout: We really don't want to export it. This is an artifact of it being sourced. So here we need an exception.

rootdir is expected to have been set by the calling script. Instead of giving it a default, add a condition that errors out if it is not set.

tools/docker/image-pull.sh
```diff
@@ -6,13 +6,14 @@
 # See LICENSE file for more details.
 ###############################################################
 
-scriptdir="$(cd "${0%/*}"; pwd)"
-rootdir="${scriptdir%/*/*}"
+scriptdir=$(dirname "$(readlink -f "$0")")
```
arnout: Again, please use
```suggestion
scriptdir="$(cd "${0%/*}" && pwd)"
```

tools/klocwork/kw.sh
```diff
@@ -21,7 +21,7 @@ declare PLATFORM_TYPE
 ROOT_PATH=$(realpath "$(pwd)"/../../)
 PLATFORM_TYPE=$PRPLMESH_PLATFORM_TYPE
 
-echo platfrom identified: $PLATFORM_TYPE
+echo platfrom identified: "$PLATFORM_TYPE"
```
arnout: While you're at it, fix the typo :-)

arnout (changes requested):
I think this is enough to merge. Well, it still needs a gitlab-ci update of course. You can use the same approach as the flake8 check, using [pipelinecomponents/shellcheck](https://hub.docker.com/r/pipelinecomponents/shellcheck). I don't think it's needed to add something similar as flake8-diff, review will make sure that modified scripts are shellchecked.

tools/functions.sh
MuteSpirit: 1) about `installdir` - agree. I'll add exception
2) about `rootdir`
`${rootdir:?}` does not define default value for variable but check that variable was set

Usage example
Add file `test.sh`
```
#!/bin/sh
. tools/functions.sh
echo "Hi"
```
and try execute it
```
$ sh test.sh 
test.sh: 9: tools/functions.sh: rootdir: parameter not set or null
```

ci/cppcheck/cppcheck-new-issues.sh
MuteSpirit: Reverted changes.

I found why we see different number of shellcheck issues:
1) Ubuntu 18.04 uses shellcheck 0.4.6
    To install newer version (0.7.0) it's needed to use snap
    `sudo snap install shellcheck`
2) I run shellcheck without args
    Now use next command:
         `shellcheck -P SCRIPTDIR -x <script-path>`
3) I reconfigured Syntastic plugin in my Vim to see the same results.
     I've added next in my `~/.vimrc`
          `let g:syntastic_sh_shellcheck_args = "-P SCRIPTDIR -x"`

ci/cppcheck/cppcheck-new-issues.sh
MuteSpirit: Agree.
Fixed used shellcheck command.

ci/cppcheck/cppcheck-new-issues.sh
MuteSpirit: Reverted my changes.

tools/docker/image-pull.sh
MuteSpirit: Add issues founded by shellcheck before refactoring into commit messages for all fixed scripts.
Create separate commit with removing "--native" option.

tools/functions.sh
MuteSpirit: Followed your advice

tools/docker/image-pull.sh
MuteSpirit: Eventually i did not get `scriptdir` via
```
scriptdir=$(dirname "$(readlink -f "$0")")
```
but followed shellcheck advice instead
```
    In tools/docker/image-pull.sh line 9:
                 ^----------^ SC2164: Use 'cd ... || exit' or 'cd ... || return' in case cd fails.
    
    Did you mean:
    scriptdir="$(cd "${0%/*}" || exit; pwd)"
```
and used
```
scriptdir="$(cd "${0%/*}" || exit 1; pwd)"
```

tools/klocwork/kw.sh
MuteSpirit: Fixed

MuteSpirit:
> I think this is enough to merge. Well, it still needs a gitlab-ci update of course. You can use the same approach as the flake8 check, using [pipelinecomponents/shellcheck](https://hub.docker.com/r/pipelinecomponents/shellcheck). I don't think it's needed to add something similar as flake8-diff, review will make sure that modified scripts are shellchecked.

I suggest to fix shellcheck issues in all shell scripts before enable this check on GitLab CI, otherwise developers besides me will have to fix shellcheck issues on any script change.

tools/docker/image-pull.sh
arnout: Well, I don't know why shellcheck advises that, because it's completely equivalent to
```
```
(which is what we use in most other places that get it right).

Anyway, doesn't matter much.

tools/functions.sh
arnout: Oh, I didn't know that. Excellent!

tools/klocwork/kw.sh
```diff
@@ -21,7 +21,7 @@ declare PLATFORM_TYPE
 ROOT_PATH=$(realpath "$(pwd)"/../../)
 PLATFORM_TYPE=$PRPLMESH_PLATFORM_TYPE
 
-echo platfrom identified: $PLATFORM_TYPE
+echo platform identified: "$PLATFORM_TYPE"
```
tomereli, line 24: The commit message is way too long - suggestion:

> scripts: kw.sh: fix shellcheck issue SC2086
>
> Fix typo + add Quoting to prevent word splitting / globbing.
>For more information:
>  https://www.shellcheck.net/wiki/SC2206 -- Quote to prevent word splitting/g...
>  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
>
> Signed-off-by: Ivan Efimov <i.efimov@inango-systems.com>

The same for all commit messages in this series.

tools/functions.sh
```diff
@@ -1,14 +1,16 @@
+# shellcheck shell=sh
 ###############################################################
 # SPDX-License-Identifier: BSD-2-Clause-Patent
 # SPDX-FileCopyrightText: 2019-2020 the prplMesh contributors (see AUTHORS.md)
 # This code is subject to the terms of the BSD+Patent license.
 # See LICENSE file for more details.
 ###############################################################
 
-installdir="${rootdir}/build/install"
```
tomereli, line 8: Here as well, the commit message is just a copy of the shellcheck output. It is not helpful IMO, please explain why you disabled SC2034 in the commit message and what syntax you used to set `installdir` - like you did here - https://github.com/prplfoundation/prplMesh/pull/1123#discussion_r407085029
You should also mention the quoting that was missing.

tools/docker/runner/entrypoint.sh
```diff
@@ -32,5 +32,5 @@ run ip link set      dev wlan2    up
 run ip address add   dev br-lan "$bridge_ip"
 run ip link set      dev br-lan   up
 
-cd ${INSTALL_DIR}
+cd "${INSTALL_DIR}" || exit 1
```
tomereli, line 30: Same comment as for previous commit messages.

tests/test_gw_repeater.sh
```diff
@@ -55,8 +55,8 @@ check_wsl() {
     # Docker fails to route traffic from the host to the containers
     # if the IP is anything but what was allocated by the daemon
     # when the container was created
-    GW_EXTRA_OPT="--expose ${GW_UCC_PORT} --publish 127.0.0.1::${GW_UCC_PORT} --ipaddr 0.0.0.0"
-    RP_EXTRA_OPT="--expose ${RP_UCC_PORT} --publish 127.0.0.1::${RP_UCC_PORT} --ipaddr 0.0.0.0"
+    GW_EXTRA_OPT=(--expose "${GW_UCC_PORT}" --publish 127.0.0.1::"${GW_UCC_PORT}" --ipaddr 0.0.0.0)
```
tomereli: The commit message explains that the shellcheck advice is not taken, but it doesn't explain what you did instead.

tools/docker/image-pull.sh
```diff
@@ -17,7 +17,6 @@ usage() {
     echo "      -h|--help - show this help menu"
     echo "      -v|--verbose - verbosity on"
     echo "      -b|--base-image - Base OS image to use (Dockerfile 'FROM')"
-    echo "      -n|--native - Use the same base OS image as the running system"
```
tomereli, line 20: Please drop this commit or fix the --native option - why isn't it usable? if indeed there is a bug, it should be fixed rather than removed.

tomereli (approved):
I think some of the commit messages should be revisited. I also think the --native option should not be removed. Other than that LGTM, so approving, but please fix those before you merge this.

tools/docker/image-pull.sh
```diff
@@ -6,18 +6,18 @@
 # See LICENSE file for more details.
 ###############################################################
 
-scriptdir="$(cd "${0%/*}"; pwd)"
+scriptdir="$(cd "${0%/*}" || exit 1; pwd)"
```
rmelotte, line 9: In other scripts we used this instead, it's more concise and the result is the same:

```suggestion
scriptdir="$(cd "${0%/*}" && pwd)"
```

Note that both solutions do silence the shellcheck issue, but do *not* exit the script if `cd` fails (but it's not something you have to fix IMO).

tools/klocwork/kw.sh
rmelotte: It looks like a big part of the commit message has nothing to do with the changes, so Tomer's comment is not a nitpick IMO, please address it.

rmelotte (approved):
LGTM overall, but please take Tomer's comments about commit messages into account.

tools/docker/image-pull.sh
arnout: > Please drop this commit or fix the --native option - why isn't it usable?

That was my suggestion. Obviously, the explanation should have been part of the commit message.

```
The --native option doesn't really work because:
- for starters, it only applies to Debian derivatives, because the Dockerfile uses APT;
- in addition, it only works with Ubuntu because the package names differ slightly;
- finally, it only works with Ubuntu 18.04 and a few versions very close to it, because of package incompatibilities.

In other words, we will anyway need a separate Dockerfile for any platform that is not Ubuntu 18.04 (or close to it, e.g. 18.10 does work). Therefore the --native option is next to useless.
```

Unless you know someone who actually uses this option, @tomereli ?

tomereli: All of Intel users are working on Ubuntu18.04 LTS server, so we used to use the -n option all the time. I haven't used it in a while since I didn't need to, so I'm not strictly against removing it - but I prefer to keep it unless its too much work.

arnout: On an Ubuntu 18.04 the -n option is a NOP, it will set IMAGE to its default value `ubuntu:18.04`.

tomereli: So it was probably used in the beginning when we used Ubuntu19 in our HyperV VMs - in that case I guess we can remove this.

tests/test_gw_repeater.sh
arnout: > The commit message explains that the shellcheck advice is not taken, but it doesn't explain what you did instead.

This is still the case. Also, in the end you *do* follow the shellcheck advice (use a bash array).

tests/test_gw_repeater.sh
```diff
@@ -55,8 +55,8 @@ check_wsl() {
...
+    RP_EXTRA_OPT=(--expose "${RP_UCC_PORT}" --publish 127.0.0.1::"${RP_UCC_PORT}" --ipaddr 0.0.0.0)
```
arnout: (nitpick)
```suggestion
    RP_EXTRA_OPT=(--expose "${RP_UCC_PORT}" --publish "127.0.0.1::${RP_UCC_PORT}" --ipaddr 0.0.0.0)
```
(I find it clearer to put the quotes on the outside.)

tests/test_gw_repeater.sh
```diff
@@ -101,7 +101,7 @@ main() {
 
     [ "$START_GATEWAY" = "true" ] && {
         status "Start GW (Controller + local Agent)"
-        "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} ${GW_EXTRA_OPT} \
+        "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} "${GW_EXTRA_OPT[@]}" \
```
arnout, line 99: Doesn't this still trigger a shellcheck error for VERBOSE_OPT?

tools/docker/image-pull.sh
arnout, line 20: Please update the commit message with the full explanation as mentioned in an earlier comment:
```
The --native option doesn't really work because:
- for starters, it only applies to Debian derivatives, because the Dockerfile uses APT;
- in addition, it only works with Ubuntu because the package names differ slightly;
- finally, it only works with Ubuntu 18.04 and a few versions very close to it, because of package incompatibilities.

In other words, we will anyway need a separate Dockerfile for any platform that is not Ubuntu 18.04 (or close to it, e.g. 18.10 does work). Therefore the --native option is next to useless.
```

tools/docker/image-pull.sh
```diff
@@ -33,15 +34,16 @@ pull_and_tag() {
...
+        err "Failed parsing options." >&2
+        usage
+        exit 1
+    fi
 
     eval set -- "$OPTS"
 
     while true; do
         case "$1" in
-            -v | --verbose)         VERBOSE=true; shift ;;
```
arnout, line 45: The removal of the verbose option is not mentioned in the commit message. It's probably OK to do it (I don't think it has any effect anyway), but in general we prefer such changes to be in a separate commit.

Don't bother doing it now though, just for future.

arnout (approved):
There are still a few comments from other reviewers that haven't been resolved. Nothing major though.

tools/klocwork/kw.sh
abelog: Commit messages are fixed.

tools/docker/image-pull.sh
abelog: If `scriptdir` is missing,  what is the point going on with script? It will just report missing functions of `functions.sh`.

tools/docker/image-pull.sh
abelog: Done!

tools/docker/image-pull.sh
arnout: > If `scriptdir` is missing, what is the point going on with script?

Excellent argument to prefer `&&`! The problem with `exit 1` is that an unwary reader may think it will exit the script, while it actually just exists the subshell created by `$(...)`. To exit the script you have to do
```suggestion
scriptdir="$(cd "${0%/*}" && pwd)" || exit 1
```
or `set -e`.

rmelotte: >The problem with exit 1 is that an unwary reader may think it will exit the script, while it actually just exists the subshell created by $(...)

Exactly. I now realize my comment was not clear enough: 
when I said "it's not something you have to fix IMO" what I meant was "you don't have to make the script exit on error" and **not** "you don't have to follow my suggestion at all".

PR #1131, opened 2020-04-10
agent: merge SSlaveSocket with sRadioInfo

arnout:
Some of the info is overlapping, and the map turns out to be not needed.

This PR was triggered by a review of #1114.

While we're at it, clean up some unused stuff.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1708,9 +1708,7 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
 
         soc->sta_iface.assign(request->sta_iface(message::IFACE_NAME_LENGTH));
         soc->hostap_iface.assign(request->hostap_iface(message::IFACE_NAME_LENGTH));
-        soc->ruid = request->ruid();
-        soc->operational_on_registration =
-            FSM_IS_IN_STATE(OPERATIONAL) || FSM_IS_IN_STATE(CONNECTED);
+        soc->ruid                 = request->ruid();
         soc->sta_iface_filter_low = request->sta_iface_filter_low();
         local_master              = request->local_master();
         local_gw                  = request->local_gw();
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1772,7 +1772,7 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
                     radio_info.supported_channels.begin());
 
         soc->ruid                  = request->iface_mac();
-        soc->radio_mac             = mac;
+        soc->radio_mac             = request->iface_mac();
         soc->freq_type             = (request->iface_is_5ghz() ? beerocks::eFreqType::FREQ_5G
                                                    : beerocks::eFreqType::FREQ_24G);
         soc->controller_discovered = false;
@@ -2269,7 +2269,7 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
             auto localInterfaceInfo = tlvDeviceInformation->create_local_interface_list();
 
             localInterfaceInfo->mac() =
-                front_iface ? network_utils::mac_from_string(soc->radio_mac)
+                front_iface ? soc->radio_mac
                             : network_utils::mac_from_string(soc->sta_wlan_hal->get_radio_mac());
 
             LOG(DEBUG) << "Added radio interface to tlvDeviceInformation: "
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1978,9 +1978,9 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
             return false;
         }
 
-        m_radio_info_map[msg->ruid()].vaps_list = msg->params();
+        m_radio_info_map[soc->radio_mac].vaps_list = msg->params();
         if (m_agent_ucc_listener) {
-            m_agent_ucc_listener->update_vaps_list(network_utils::mac_to_string(msg->ruid()),
+            m_agent_ucc_listener->update_vaps_list(network_utils::mac_to_string(soc->radio_mac),
                                                    msg->params());
         }
         break;
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1752,11 +1752,15 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<SSlaveSocke
             return false;
         }
 
-        auto &radio_info = m_radio_info_map[request->iface_mac()];
+        auto radio_info = m_radio_info_map[request->iface_mac()];
+        if (!radio_info) {
+            radio_info = std::make_shared<sRadioInfo>();
+            m_radio_info_map[request->iface_mac()] = radio_info;
+        }
 
-        radio_info.interface_name = request->ap_iface();
-        radio_info.frequency_band = request->frequency_band();
-        radio_info.max_bandwidth  = request->max_bandwidth();
+        radio_info->interface_name = request->ap_iface();
+        radio_info->frequency_band = request->frequency_band();
+        radio_info->max_bandwidth  = request->max_bandwidth();
 
         std::string mac = network_utils::mac_to_string(request->iface_mac());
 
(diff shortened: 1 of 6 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -106,24 +107,6 @@ class backhaul_manager : public btl::transport_socket_thread {
...
-        bool operational_on_registration = false;
-        eIfaceType slave_iface_type;
-        sMacAddr ruid;
-
-        std::shared_ptr<bwl::sta_wlan_hal> sta_wlan_hal;
-        Socket *sta_hal_ext_events = nullptr;
-        Socket *sta_hal_int_events = nullptr;
-    };
-
     SocketClient *master_discovery_socket = nullptr;
```
RanRegev, line 110: ok, we gave up this... (see later)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -259,15 +242,25 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+        eFreqType freq_type            = eFreqType::FREQ_UNKNOWN;
+        bool sta_iface_filter_low      = false;
+        bool slave_is_backhaul_manager = false;
+        bool controller_discovered     = false;
+
+        std::shared_ptr<bwl::sta_wlan_hal> sta_wlan_hal;
+        Socket *sta_hal_ext_events = nullptr;
+        Socket *sta_hal_int_events = nullptr;
+
         beerocks::eWiFiBandwidth max_bandwidth =
```
RanRegev, line 264: ... and we created this... (see later)

RanRegev:
... but, after removing and uniting structures, which is a good thing to do, my main concern is this:
We use structures as a place to hold pieces of data that have some relation to each other.
We query this structure from *outside* the structure.
This is functional / C-style programming: Having structures in one hand and functionality on the other hand - unrelated.
I prefer _encapsulating_ the functionality inside the structure. In this way, as a programmer, when I need to implement some logic with it, I can concentrate on the logic itself and not on the mechanics that brings me the required data.
Here is an example, from this specific code. In this case the collection `std::list<std::shared_ptr<sRadioInfo>> slaves_sockets` -
Looking for specific mac:
https://github.com/prplfoundation/prplMesh/blob/e60596a85554aaed08b9aebd9580726f1626e6c9/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp#L913-L917
Looking for specific Socket:
https://github.com/prplfoundation/prplMesh/blob/e60596a85554aaed08b9aebd9580726f1626e6c9/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp#L1653-L1657

We are seeking the collection for specific value from _outside_ the collection.
I truly prefer it this way:
```
slaves_sockets.find( socket ); // returns the structure with the given socket.
slaves_sockets.find_sta( sta_mac ); // returns the structure with the given STA mac.
```

So I can implement my logic very cleanly:

```
if (auto& s = slaves_sockets.find_sta( mac )){
/* do my stuff */
}
```

Note:
* `find_sta(sta_mac)` in current code (before this change at least) was few lines of massing around with the bits and bytes of two structures.
* implementing the logic _inside_ the structure also prevents copy/paste the functionality all over the code.

RanRegev (approved):
My comment on this PR is just an initial talk, via specific example, on a general direction.
I actually took this code and I'm using it in my branch.

arnout:
> ... but, after removing and uniting structures, which is a good thing to do, my main concern is this:
> We use structures as a place to hold pieces of data that have some relation to each other.
> We query this structure from _outside_ the structure.

I generally concur with your observation.
However, in this particular case, the `sRadioInfo` structure itself doesn't have many opportunities for abstraction. What you're asking for here is abstraction for the collection-of-sRadioInfo. But the collection-of-sRadioInfo is (at the moment) the backhaul_manager_thread class...

So you have two good suggestions:
- there should be functions to look things up in the collection-of-sRadioInfo rather than open-coding that every time;
- collection-of-sRadioInfo should be a class on its own rather than being mixed up with the backhaul_manager_thread state machine.

However, both of these changes would be for separate PRs. This PR is already 13 commits, which is bigger than a PR should be in the ideal world.

RanRegev:
> > ... but, after removing and uniting structures, which is a good thing to do, my main concern is this:
> > We use structures as a place to hold pieces of data that have some relation to each other.
> > We query this structure from _outside_ the structure.
> 
> I generally concur with your observation.
> However, in this particular case, the `sRadioInfo` structure itself doesn't have many opportunities for abstraction. What you're asking for here is abstraction for the collection-of-sRadioInfo. But the collection-of-sRadioInfo is (at the moment) the backhaul_manager_thread class...
> 
> So you have two good suggestions:
> 
> * there should be functions to look things up in the collection-of-sRadioInfo rather than open-coding that every time;
> * collection-of-sRadioInfo should be a class on its own rather than being mixed up with the backhaul_manager_thread state machine.
> 
> However, both of these changes would be for separate PRs. This PR is already 13 commits, which is bigger than a PR should be in the ideal world.

I agree, it shouldn't be done here.
My suggestion is more than "add functionality into the structure". 
It actually suggest to start thinking in encapsulated way - separate structures (and classes) from their immediate-that-is-what-I-need-now-for-this-flow approach. Rather, make classes more self contained, supplying abilities (services) to be used by various flows.

vitalybu (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -602,7 +602,7 @@ void backhaul_manager::after_select(bool timeout)
 }
 
 bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
-                                                     std::shared_ptr<SSlaveSockets> pSocket)
+                                                     std::shared_ptr<sRadioInfo> pSocket)
```
rmelotte, line 605: From a quick look it seems that pSocket is not stored or anything, in which case it's weird to take a shared_ptr instead of a reference as argument (but it was already the case before, so not specific to this PR).

rmelotte (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: You're saying that is should be passed as reference rather than pointer, right? Indeed, that is a fairly general shortcoming in the code currently: in many places, things are passed by shared_ptr just because they happen to be stored as shared_ptr, while really they should be passed as reference (sometimes const reference). But as you say: not for this PR.

morantr (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
rmelotte: Yes, that's exactly what I was trying to say :- )

RanRegev: I hope that slowly we'll start giving _meaning_ to the types we hold, pass or return.
In particular:
A pointer (shared, unique, raw) - the underlying object may exists or my *not*, and it is OK. In other word we have 0..1 objects behind this pointer.
A reference - we promise that we have _exactly_ one object behind it.
(therefore `A& foo(){ return *a; }` is extremely dangerous and has only rare valid cases)
There are of course many more meanings, like I own it, I transfer ownership, we both own it, you may change it, you may not change and so on.

tomereli: @RanRegev you got this backwards - we want to **pass** by reference to functions, not return reference from functions:
```
bool func()
{
  std::shared_ptr<sVar> var = std::make_shared...
  if (!var) {
    LOG(ERROR) << "Failed to allocate memory";
    return false; 
  }
  ...
  do_something_with_var(*var)
  ...
}

void do_something_with_var(sVar &var);

```

RanRegev: No, I didn't get it backward, I was talking generically.
What is "backward" here is 

> we want to *pass* by reference

You never "want to pass", you always pass "what was requested".
The caller does not set the requirements from the parameters, return value or constness, ever. It is the declaration of the function that does this.
And in this manner, the function declaration _states_ that it must have an object under the passed value by setting the parameter with a reference. It can also state "it is fine to have a non-object behind this, I know how to handle it" then it declares a pointer.
functions like:
```
void foo( A* a ){
if ( !a ) return;
/* code with a */
}
```
Are a mistake in the _declaration_.
If you know that your caller has a pointer to A (A*) and therefore you declare your function to request a pointer - this is a mistake. You are not obligated to the way your clients will use you. You are obligated to your needs.
In the example I gave it is better to declare:
```
void foo( A& a ) {
/* code with a */
}
```
and let the caller decide how to call:
```
if ( a ) {
foo( *a ) 
} else {
/* error flow */
}
``` 
C++ is a very declarative language, that let you control and emphasis many aspects of your intents and assumptions. I think we should use these abilities more.

arnout: @tomereli @RanRegev You're both saying the same thing: pass objects by reference, unless it is "optional" (i.e. may be nullptr).

PR #1134, opened 2020-04-12
DCS: Feature/limit dcs scans to one at a time

### LINKED ISSUE #1094: [TASK] DCS: Limit scans to one at a time to prevent NL messages overrun
## Description

When a scan is triggered on the 2 radios at the same time - the events of dump_results may overlap and the NL buffer size may exceed the expected size due to messages overlap (as described in #1001 )
The DCS can protect against simultaneous scans and prevent the issue from happening.
Moreover, the crash of the monitor thread in the case of one event-message-overlap is too strict since we see that the following events arrive successfully.

The root cause of the issue might be DWPAL implementation - and since we plan to move the monitor thread to use the mon_wlan_hal_nl80211 instead of DWPAL, and since there is no requirement to support scan on both radios simultaneously, the decision is to limit the DCS feature to scan one radio at a time.

As part of simulating the issue, we found that although the DCS task handles a scan-triggered event timeout, it should also handle a scan-trigger-response returning an error (which also can arrive much earlier than the timeout itself)

Fixes: #1001 

## Exit criteria

- [x] a single wrong-buffer-size as described in #1001 should not cause monitor thread crash
PR: #1100 
- [x] DCS task should handle scan-trigger-response with error
will be handled as part of #1135 
- [x] limit DCS tasks to a single scan on all radios at a time
PR: #1134 

## Detailed description

Currently, DCS tasks do not share resources and are completely independent. The solution suggested is for all DCS tasks to share a common "is-scan-in-progress" in the DB which that each task must check before starting a new scan.
Multi-threading must be taken into account in the solution provided.

adam1985d:
Added DB APIs to try-lock/unlock scan-permission.
The above-mentioned APIs are used by the DCS task instances to check if the task can trigger a scan or another task is already scanning.
This is to limit to a single scan at a time - independent of the number of DCS tasks created.

The tasks release the ownership if a scan is complete (including getting all of the dump-results from the BWL) or if a scan failed/aborted for any reason.

In the case that the DCS task performs a retry of the scan, the task will require to take ownership once again - this is to prevent starvation of other tasks.

Fixes: #1094

### DIFF
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -564,6 +564,24 @@ public:
     const std::list<sChannelScanResults> &get_channel_scan_results(const sMacAddr &mac,
                                                                    bool single_scan);
 
+    /**
+     * @brief Lock permission for channel scan
+     * 
+     * @param task_id: id of task requesting to lock the scan permission
+     * @return true on success
+     * @return false on failure (if failed to lock or locked by other task)
+     */
+    bool try_lock_scan_permission(int task_id);
+
+    /**
+     * @brief Unlock permission for channel scan
+     * 
+     * @param task_id: id of task requesting to unlock the scan permission
+     * @return true on success
+     * @return false on failure (if failed to release or unlocked by other task)
+     */
+    bool unlock_scan_permission(int task_id);
+
     //
     // CLI
     //
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -926,6 +926,9 @@ private:
 
     master_thread *m_master_thread_ctx = nullptr;
     const std::string m_local_bridge_mac;
+
+    // task_id==-1 means lock is free
+    int m_scan_locked_by_task_id = -1;
 };
 
 } // namespace son
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -117,6 +117,11 @@ void dynamic_channel_selection_task::work()
         break;
     }
     case eState::TRIGGER_SCAN: {
+        // get permission to scan
+        if (!database.try_lock_scan_permission(id)) {
+            break;
+        }
+
         LOG(TRACE) << "TRIGGER_SCAN, mac=" << m_radio_mac << ", scan_type is "
                    << ((m_is_single_scan) ? "single-scan" : "continuous-scan");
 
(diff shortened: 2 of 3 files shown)


controller/src/beerocks/master/db/db.h
```diff
@@ -564,6 +564,24 @@ class db {
...
                                                                    bool single_scan);
 
+    /**
+     * @brief Lock permission for channel scan
+     * 
+     * @param task_id: id of task requesting to lock the scan permission
+     * @return true on success
+     * @return false on failure (if failed to lock or locked by other task)
+     */
+    bool try_lock_scan_permission(int task_id);
```
kantera800, line 574: IMO should be just `lock_scan_permission` since its clear from the definition (nitpick)

kantera800 (approved):
good commit split.

controller/src/beerocks/master/db/db.h
adam1985d: the `try` is to differ from `lock-until-resource-is-free`.
we try to take ownership and if we fail we return false.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -117,6 +117,11 @@ void dynamic_channel_selection_task::work()
         break;
     }
     case eState::TRIGGER_SCAN: {
+        // get permission to scan
+        if (!database.try_lock_scan_permission(id)) {
+            break;
```
itayx, line 123: nitpick, add an error message for logging

itayx (approved):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: this was intentionally not added since it will flood the controller log
I added a log of lock and release

a combination of those 2 with the state-change log print can provide all possible behavior states.

tomereli (changes requested):
It is wrong to use locking terminology here since it is not real locking, you are counting on the fact that tasks are scheduled one after the other by the controller. You can't rely on that, or at least you shouldn't, since it is a very fragile design.

Instead, you should simply use a static member in the DCS task which will hold the currently scanning task id. Before starting a scan, it will check if the task ID is set, and if it is, it will simply fail.
Do not add APIs at all - use the member simply in the function.

Example:

```
bool start_scan(int id)
{
  if (m_scanning == -1) {
    m_scanning = id;
  } else {
    LOG(ERROR) << "scanning active by task " << int(id);
    return false;
  }
  // do scan
  return true
}
```

With this, I expect a single commit and not 5.

PR #1136, opened 2020-04-12
tests: boardfarm_plugins: port initial_ap_config

### LINKED ISSUE #1071: [TASK] Port test_initial_ap_config to boardfarm for dockerized device

taurha:
Ported test_initial_ap_config from tests/test_flow.py for dockerized
boardfarm device and created prplMesh testsuite.

During porting:
* environment.py contents were used "as is", there's
  no strong necessity to modify it;
* boardfarm logs contains only output of commands issued by
  its devices, i.e. all actual test logs still generated in:
  <prplMesh-repo-root>/logs/*
  (for that reason, RadioDocker.wait_for_log still works)
* boardfarm logs can be found in directory "results" created
  in same dir bft was started from;

Signed-off-by: Oleksii Ponomarenko <o.ponomarenko@inango-systems.com>

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
@@ -6,8 +6,10 @@
 import os
 import time
 
-from environment import ALEntityDocker
+import boardfarm
+from environment import ALEntityDocker, _get_bridge_interface
 from .prplmesh_base import PrplMeshBase
+from sniffer import Sniffer
 
 rootdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../..'))
 
@@ -58,6 +60,8 @@ class PrplMeshDocker(PrplMeshBase):
             time.sleep(self.delay)
             self.agent_entity = ALEntityDocker(self.name, is_controller=False)
 
+        self.wired_sniffer = Sniffer(_get_bridge_interface(self.docker_network),
+                                     True, boardfarm.config.output_dir)
         self.check_status()
 
     def __del__(self):
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
@@ -4,6 +4,17 @@
         "board_type": "prplmesh_docker",
         "role": "controller",
         "conn_cmd": "",
-        "devices": []
+        "devices": [
+            {
+                "name": "repeater1",
+                "type": "prplmesh_docker",
+                "conn_cmd": ""
+            },
+            {
+                "name": "repeater2",
+                "type": "prplmesh_docker",
+                "conn_cmd": ""
+            }
+        ]
     }
 }
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
@@ -4,13 +4,31 @@
 # See LICENSE file for more details.
 
 from boardfarm.tests import bft_base_test
+from utils import Utils
 
 
 class InitialApConfig(bft_base_test.BftBaseTest):
-    """PrplMesh sample test case, no actual testing, it just passes."""
+    """Check initial configuration on device."""
 
     def runTest(self):
-        """Main test logic must be implemented here."""
+        for dev in self.dev:
+            if dev.agent_entity:
+                dev.wired_sniffer.start(self.__class__.__name__)
+
+                Utils.check_log(dev.agent_entity.radios[0],
+                                r"WSC Global authentication success")
+                Utils.check_log(dev.agent_entity.radios[1],
+                                r"WSC Global authentication success")
+                Utils.check_log(dev.agent_entity.radios[0],
+                                r"KWA \(Key Wrap Auth\) success")
+                Utils.check_log(dev.agent_entity.radios[1],
+                                r"KWA \(Key Wrap Auth\) success")
+                Utils.check_log(dev.agent_entity.radios[0],
+                                r".* Controller configuration \(WSC M2 Encrypted Settings\)")
+                Utils.check_log(dev.agent_entity.radios[1],
+                                r".* Controller configuration \(WSC M2 Encrypted Settings\)")
+
+                dev.wired_sniffer.stop()
 
     @classmethod
     def teardown_class(cls):
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,0 +1,20 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+from typing import Union
+
+import environment as env
+
+
+class Utils(object):
+    """Collection of common functions used by tests."""
+
+    @staticmethod
+    def check_log(entity_or_radio: Union[env.ALEntity, env.Radio], regex: str,
+                  start_line: int = 0, timeout: float = 0.3) -> bool:
+        result, line, match = entity_or_radio.wait_for_log(regex, start_line, timeout)
+        if not result:
+            raise Exception
+        return result, line, match
--- a/tests/environment.py
+++ b/tests/environment.py
@@ -265,9 +265,8 @@ class VirtualAPDocker(VirtualAP):
         self.radio.send_bwl_event("EVENT AP-STA-DISCONNECTED {}".format(sta.mac))
 
 
-def _get_bridge_interface(unique_id):
+def _get_bridge_interface(docker_network):
     '''Use docker network inspect to get the docker bridge interface.'''
-    docker_network = 'prplMesh-net-{}'.format(unique_id)
     docker_network_inspect_cmd = ('docker', 'network', 'inspect', docker_network)
     inspect_result = subprocess.run(docker_network_inspect_cmd, stdout=subprocess.PIPE)
     if inspect_result.returncode != 0:
@@ -296,8 +295,8 @@ def _get_bridge_interface(unique_id):
 
 def launch_environment_docker(unique_id: str, skip_init: bool = False):
     global wired_sniffer
-    wired_sniffer = sniffer.Sniffer(_get_bridge_interface(unique_id),
-                                    opts.tcpdump, opts.tcpdump_dir)
+    iface = _get_bridge_interface('prplMesh-net-{}'.format(unique_id))
+    wired_sniffer = sniffer.Sniffer(iface, opts.tcpdump, opts.tcpdump_dir)
 
     gateway = 'gateway-' + unique_id
     repeater1 = 'repeater1-' + unique_id
--- a/tests/environment.py
+++ b/tests/environment.py
@@ -13,7 +13,7 @@ import subprocess
 import time
 
 from capi import UCCSocket
-from opts import debug, err
+from opts import opts, debug, err
 import sniffer
 
 
@@ -296,7 +296,8 @@ def _get_bridge_interface(unique_id):
 
 def launch_environment_docker(unique_id: str, skip_init: bool = False):
     global wired_sniffer
-    wired_sniffer = sniffer.Sniffer(_get_bridge_interface(unique_id))
+    wired_sniffer = sniffer.Sniffer(_get_bridge_interface(unique_id),
+                                    opts.tcpdump, opts.tcpdump_dir)
 
     gateway = 'gateway-' + unique_id
     repeater1 = 'repeater1-' + unique_id
--- a/tests/sniffer.py
+++ b/tests/sniffer.py
@@ -8,21 +8,23 @@
 import os
 import subprocess
 
-from opts import debug, err, opts, status
+from opts import debug, err, status
 
 
 class Sniffer:
     '''Captures packets on an interface.'''
-    def __init__(self, interface: str):
+    def __init__(self, interface: str, use_tcpdump: bool, tcpdump_log_dir: str):
         self.interface = interface
+        self.use_tcpdump = use_tcpdump
+        self.tcpdump_log_dir = tcpdump_log_dir
         self.tcpdump_proc = None
 
     def start(self, outputfile_basename):
         '''Start tcpdump if enabled by config.'''
-        if opts.tcpdump:
+        if self.use_tcpdump:
             debug("Starting tcpdump, output file {}.pcap".format(outputfile_basename))
-            os.makedirs(os.path.join(opts.tcpdump_dir, 'logs'), exist_ok=True)
-            outputfile = os.path.join(opts.tcpdump_dir, outputfile_basename) + ".pcap"
+            os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
+            outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
             command = ["tcpdump", "-i", self.interface, "-w", outputfile]
             self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
             # tcpdump takes a while to start up. Wait for the appropriate output before continuing.


taurha:
That PR should be reviewed first: #1108

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -0,0 +1,65 @@
...
+
+class InitialApConfig(bft_base_test.BftBaseTest):
+    '''
+    Check initial configuration on device
+    '''
+
+    def runTest(self):
+        '''
+        Main test logic must be implemented here.
+        '''
```
MuteSpirit: I think that such common comment should be replaced with description of test goals

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -0,0 +1,65 @@
...
+        regex_list = [
+            r"WSC Global authentication success",
+            r"KWA \(Key Wrap Auth\) success",
+            r".* Controller configuration \(WSC M2 Encrypted Settings\)"
+        ]
+        failed_regex = []
+
+        wired_sniffer = sniffer.Sniffer(_get_bridge_interface(self.dev.board.unique_id))
```
MuteSpirit: Function `_get_gridge_interface` expected that docker network will have name `prplMesh-net-{unique_id}`, but in real test setup network may have another name template.
So i suggest 
1) set docker network name in test setup config
2) modify `_get_gridge_interface` to be able pass it docker network name instead of unique id

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
```
arnout: This file doesn't have a main and isn't executable, so a shebang is useless.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+import subprocess
+import sys
+import time
+
+from boardfarm.devices import debian
+from environment import ALEntityDocker, _get_bridge_interface
+
+rootdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../..'))
+
+logging.basicConfig(stream=sys.stdout, format='%(message)s')
```
arnout: I believe bft already does this so it shouldn't be needed, no?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+logger = logging.getLogger('bft')
+logger.setLevel(logging.INFO)  # DEBUG, INFO, WARNING, ERROR, CRITICAL
+
+
+class PrplMeshDocker(debian.DebianBox):
+    '''
+    Dockerized prplMesh device
+    '''
```
arnout:
```suggestion
    '''Dockerized prplMesh device.'''
```

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+    model = ("prplmesh_docker")
+
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+        config = kwargs["config"] if kwargs.get("config", None) else kwargs
```
arnout:
```suggestion
        config = kwargs.get("config", kwargs)
```

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+logger.setLevel(logging.INFO)  # DEBUG, INFO, WARNING, ERROR, CRITICAL
+
+
+class PrplMeshDocker(debian.DebianBox):
+    '''
+    Dockerized prplMesh device
+    '''
+    model = ("prplmesh_docker")
+
+    def __init__(self, *args, **kwargs):
```
arnout: There is no call to the base class constructor anywhere, which seems weird to me...

I would actually expect to start with a call to the base class constructor, and then only handle the arguments which are specific to this class.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+
+        # Getting unic ID to distinguish devices and network they belong to
+        self.unique_id = os.getenv("SUDO_USER", os.getenv("USER", ""))
+
+        self.name = "-".join((config.get("name", "prplmesh_docker"), self.unique_id))
+        self.role = config.get("role", "agent")
+        self.cleanup_cmd = config.get("cleanup_cmd", None)
+        self.conn_cmd = config.get("conn_cmd", None)
+        self.mac = config.get("mac", self._get_rand_mac(oct_amount=4))
+        self.use_tcpdump = config.get("use_tcpdump", False)
```
arnout: Note that this option will go away (cfr. #1145) so don't bother adding it here. Just always assume true.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+        self.consoles = [self]
+
+        # Getting unic ID to distinguish devices and network they belong to
+        self.unique_id = os.getenv("SUDO_USER", os.getenv("USER", ""))
+
+        self.name = "-".join((config.get("name", "prplmesh_docker"), self.unique_id))
+        self.role = config.get("role", "agent")
+        self.cleanup_cmd = config.get("cleanup_cmd", None)
+        self.conn_cmd = config.get("conn_cmd", None)
+        self.mac = config.get("mac", self._get_rand_mac(oct_amount=4))
```
arnout: This will be removed (cfr. #1146) so don't bother adding it.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _run_shell_cmd(self, cmd: str = ""):
+        """
+        Runs specified command in host's shell and saves its output into device's log
+        """
+        res = subprocess.run(("/bin/bash", "-c", cmd), stdout=subprocess.PIPE,
+                             stderr=subprocess.STDOUT, encoding="utf-8")
+        self.log_calls += cmd
+        self.log += res.stdout
+        logger.debug(res.stdout)
```
arnout: Avoid using shell for subprocesses, instead use proper individual command-line options as a list. We do need a shell for commands specified in the config file, but there shouldn't be a need for those in our specific PrplMeshDocker class.

I think this wrapper for subprocess is not terribly useful because it's not called very often and it just adds the logging bit.

Also, instead of subprocess, I think it's better to use bft_pexpect_helper because it handles timeouts.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+        self.role = config.get("role", "agent")
+        self.cleanup_cmd = config.get("cleanup_cmd", None)
+        self.conn_cmd = config.get("conn_cmd", None)
+        self.mac = config.get("mac", self._get_rand_mac(oct_amount=4))
+        self.use_tcpdump = config.get("use_tcpdump", False)
+        self.docker_network = config.get("docker_network",
+                                         "prplMesh-net-{}".format(self.unique_id))
+
+        if self.cleanup_cmd:
+            self._run_shell_cmd(self.cleanup_cmd)
```
arnout: Feels weird to call this in the constructor...

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+
+        docker_run_cmd = os.path.join(rootdir, "tools", "docker", "run.sh") \
+            + " --verbose --detach --force --name {}".format(self.name) \
+            + " --mac {}".format(self.mac) \
+            + " --network {}".format(self.docker_network)
```
arnout: This should be a tuple instead of a string.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+
+        if self.role == "both":
+            # Spawn dockerized controller + agent in single device
+            self._run_shell_cmd(docker_run_cmd + " start-controller-agent")
+            time.sleep(config.get("delay", 5))
```
arnout: Instead of doing config.get every time, add the delay as an attribute and use self.delay.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+            self.al_entity = ALEntityDocker(self.name, is_controller=True)
+        else:
+            # Spawn dockerized agent
+            self._run_shell_cmd(docker_run_cmd + " start-agent")
+            time.sleep(config.get("delay", 5))
+            self.al_entity = ALEntityDocker(self.name, is_controller=False)
+
+        self.gw_iface = _get_bridge_interface(self.docker_network)
+
+        self.check_status()
```
arnout, line 65: It would be more appropriate to have this as a test instead of doing it in the constructor.

And that should be a separate commit.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+        if self.role == "both":
+            # Spawn dockerized controller + agent in single device
+            self._run_shell_cmd(docker_run_cmd + " start-controller-agent")
+            time.sleep(config.get("delay", 5))
+            self.al_entity = ALEntityDocker(self.name, is_controller=True)
+        elif self.role == "controller":
+            # Spawn dockerized controller
+            self._run_shell_cmd(docker_run_cmd + " start-controller")
+            time.sleep(config.get("delay", 5))
+            self.al_entity = ALEntityDocker(self.name, is_controller=True)
```
arnout: As mentioned, controller-only won't be supported. Alternatively, we could support it but with `start-controller-agent` instead of `start-controller`.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
```diff
@@ -0,0 +1,52 @@
...
+        "devices": [
+            {
+                "name": "repeater1",
+                "type": "prplmesh_docker",
+                "role": "agent",
+                "delay": 7,
+                "use_tcpdump": true,
+                "docker_network": "prplmesh-net",
+                "conn_cmd": "",
+                "radios": [
```
arnout: I would like to start out with a minimum config. It should have none of the options where we use the default (and conversely, the defaults should be what we need, e.g. delay).

The minimum config for now is just the top-level device without any subdevices.

Overall this needs to be split up into commits that add things in small steps. Now it's difficult to see why something is needed (and if it is really needed).

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -0,0 +1,146 @@
...
+        docker_run_cmd = os.path.join(rootdir, "tools", "docker", "run.sh") \
+            + " --verbose --detach --force --name {}".format(self.name) \
+            + " --mac {}".format(self.mac) \
+            + " --network {}".format(self.docker_network)
+
+        if self.role == "both":
+            # Spawn dockerized controller + agent in single device
+            self._run_shell_cmd(docker_run_cmd + " start-controller-agent")
+            time.sleep(config.get("delay", 5))
+            self.al_entity = ALEntityDocker(self.name, is_controller=True)
```
arnout: You should create two entities here: one for the agent and one for the controller.
```suggestion
            self.agent_entity = ALEntityDocker(self.name, is_controller=False)
            self.controller_entity = ALEntityDocker(self.name, is_controller=True)
```

For the other cases, we leave either agent_entity or controller_entity None.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -0,0 +1,68 @@
...
+            r"KWA \(Key Wrap Auth\) success",
+            r".* Controller configuration \(WSC M2 Encrypted Settings\)"
+        ]
+        failed_regex = []
+
+        wired_sniffer = sniffer.Sniffer(self.dev.board.gw_iface, True, boardfarm.config.output_dir)
+        wired_sniffer.start(self.__class__.__name__)
+
+        for radio in self.dev.board.al_entity.radios:
+            for regex in regex_list:
```
arnout: Although this looks like a nice refactoring at first sight, I don't really agree. I think it's harder to read this way than when it's explicitly separate checks, like is done in test_flows.

In general, try to take over the code from test_flows as literally as possible, and do any refactorings in a second stage.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -0,0 +1,68 @@
...
+        wired_sniffer.stop()
+
+        if failed_regex:
+            msg = ""
+            for entry in failed_regex:
+                msg += "Check for {}:{}: '{}' failed!\n".format(self.dev.board.al_entity.name,
+                                                                entry[0],
+                                                                entry[1]
+                                                                )
+            raise Exception(msg)
```
arnout: Is raising an exception the only way to indicate failure?

arnout (commented):
Lots of comments but nothing major. Keep up the good work!

I probably missed some things in the PrplMeshDocker class so expect more comments in the second review cycle.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
taurha: > I believe bft already does this so it shouldn't be needed, no?

Indeed, it's redundant snippet.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
taurha: > There is no call to the base class constructor anywhere, which seems weird to me...
> 
> I would actually expect to start with a call to the base class constructor, and then only handle the arguments which are specific to this class.

In our case, we using already existing boardfarm device as parent class, but when such parent class constructor sourced - we got bunch of exceptions from it.
It seems like existing boardfarm devices include additional functionality and requires some enhancement before we can use it as is - that is redundant for us.

So by inheriting existing boardfarm device, I just want to have access to parent methods that are absent in derived device class and mostly used by boardfarm internal logic.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
taurha: > Avoid using shell for subprocesses, instead use proper individual command-line options as a list. We do need a shell for commands specified in the config file, but there shouldn't be a need for those in our specific PrplMeshDocker class.
> 
> Also, instead of subprocess, I think it's better to use bft_pexpect_helper because it handles timeouts.

Applied in #1108.

I've already tried to use _bft_pexpect_helper_, but it's functionality seems to be in development and not works as expected. So instead I replaced _subprocess_ calls by _pexpect_ ones.

> I think this wrapper for subprocess is not terribly useful because it's not called very often and it just adds the logging bit.

IMO, without that wrapper the code will be inflated by similar snippets, so I would prefer to leave it to respect "Don't repeat yourself" principle.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
taurha: > It would be more appropriate to have this as a test instead of doing it in the constructor.
> 
> And that should be a separate commit.

That method is a "health check" and must be present in the device class:
it is used by boardfarm to indicate that spawned device instance is ready for test and also it is used after test to insure that device still operational.
So creating a testcase from that method will be wrong, as we have to run it after each test in boardfarm testsuite - that's a mess IMO.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
taurha: > Overall this needs to be split up into commits that add things in small steps. Now it's difficult to see why something is needed (and if it is really needed).

Applied to #1108.
Added minimal config with empty subdevices list.
(`devices` list should be present even if it is empty)

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
taurha: > Is raising an exception the only way to indicate failure?

Yes, `runTest()` methods doesn't have any return statement, so, according to boardfarm sources, the only way to indicate failure - throw exception.

MuteSpirit (approved):

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
arnout: Actually, AFAICS in the source code of boardfarm, this method is only called after a failed test, as a way to distinguish between a failing test and a failing device.

But regardless, your point still stands: it doesn't make much sense to implement this as a separate test. And it *does* make sense to check the status in the constructor, since indeed boardfarm assumes that the device is fully operational when the constructor finishes.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
```diff
@@ -4,6 +4,17 @@
         "board_type": "prplmesh_docker",
         "role": "controller",
         "conn_cmd": "",
-        "devices": []
+        "devices": [
+            {
+                "name": "repeater1",
+                "type": "prplmesh_docker",
```
arnout, line 10: Actually, I believe the idea was to have repeater1 as DUT, and the controller as a peripheral device. So it would make more sense to structure it as
```
{
    "prplmesh_docker": {
        "name": "repeater1",
        "role": "agent",
        ...
       "devices": [
          {
            "name": "gateway",
            "role": "controller",
...
```

But this doesn't make any practical difference, so (nitpick).

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -58,6 +60,8 @@ def __init__(self, *args, **kwargs):
             time.sleep(self.delay)
             self.agent_entity = ALEntityDocker(self.name, is_controller=False)
 
+        self.wired_sniffer = Sniffer(_get_bridge_interface(self.docker_network),
```
arnout, line 63: This is not correct - there should be a single, global, wired sniffer.

But then the problem, of course, is that you don't know the docker_network at the global scope.

So I think the solution is to only define the sniffer in the DUT specification. I'm not sure, however, how to distinguish the top-level device from the others... Maybe with an additional, explicit option in the JSON file? So in the JSON file we'd set `"wired_interface": "eth0"` to enable the sniffer on eth0. And we could have a special case `"wired_interface": "docker"` to use `self.docker_network`.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -4,13 +4,31 @@
...
 
 class InitialApConfig(bft_base_test.BftBaseTest):
-    """PrplMesh sample test case, no actual testing, it just passes."""
+    """Check initial configuration on device."""
 
     def runTest(self):
-        """Main test logic must be implemented here."""
+        for dev in self.dev:
+            if dev.agent_entity:
+                dev.wired_sniffer.start(self.__class__.__name__)
```
arnout, line 16: This should be refactored to be done automatically before each test. But that can be done later.

tests/utils.py
```diff
@@ -0,0 +1,20 @@
...
+from typing import Union
+
+import environment as env
+
+
+class Utils(object):
+    """Collection of common functions used by tests."""
+
+    @staticmethod
+    def check_log(entity_or_radio: Union[env.ALEntity, env.Radio], regex: str,
```
arnout, line 15: Personally, I much prefer the approach taken in current test_flows.py: define a base class that has these helper functions, and make the individual tests derive from that class. But I guess it's mostly a matter of taste. It does become important, however, if we ever need to keep any state - and I think we *will* need to do that.

arnout (approved):
All my comments are a little bit vague, so I'm OK with merging as-is and refactoring later. @oponomar please think about the idea of making a base class for the prplmesh tests and derive InitialApConfig from the base class, rather than introducing the Utils class.

tomereli (approved):

PR #1138, opened 2020-04-13
DCS: Feature/limit dcs scans to one at a time

### LINKED ISSUE #1094: [TASK] DCS: Limit scans to one at a time to prevent NL messages overrun

adam1985d:
Added `FINISH` state and moved the scan-cleanups to the `FINISH` state.

Added to the DCS class functions and a static data member that adds the capability to limit to a single scan at a time.
Added PRE_SCAN state in which the limitation of single-scan-at-a-time is enforced. When single-scan or continuous-scan conditions are met, the state is changed to PRE_SCAN and only when the no-scan-in-progress condition is met, the state is changed to `TRIGGER_SCAN` which triggers the scan.

The limit to a single scan at a time is independent of the number of DCS tasks created.

The tasks release the ownership of the scan in the new `FINISH` state as part of the scan-cleanup.

In the case that the DCS task performs a retry of the scan, the task will require to take ownership once again - this is to prevent starvation of other tasks.

Fixes: #1094

### DIFF
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -195,16 +195,17 @@ void dynamic_channel_selection_task::work()
         }
 
         m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_SUCCESS;
-        database.set_channel_scan_results_status(m_radio_mac, m_last_scan_error_code,
-                                                 m_is_single_scan);
-
-        database.set_channel_scan_in_progress(m_radio_mac, false, m_is_single_scan);
-        m_is_single_scan = false;
-        fsm_move_state(eState::IDLE);
+        fsm_move_state(eState::FINISH);
         break;
     }
     case eState::ABORT_SCAN: {
         LOG(ERROR) << "aborting scan for mac=" << m_radio_mac << ", last_scan_timestamp is not set";
+
+        fsm_move_state(eState::FINISH);
+        break;
+    }
+    case eState::FINISH: {
+        LOG(TRACE) << "finish scan for mac=" << m_radio_mac;
 
         database.set_channel_scan_results_status(m_radio_mac, m_last_scan_error_code,
                                                  m_is_single_scan);
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -103,15 +103,16 @@ void dynamic_channel_selection_task::work()
         if (m_is_single_scan_pending) {
             m_is_single_scan         = true;
             m_is_single_scan_pending = false;
+            LOG(DEBUG) << "single scan is pending, trigger single scan";
             fsm_move_state(eState::TRIGGER_SCAN);
         } else if (database.get_channel_scan_is_enabled(m_radio_mac)) {
             auto now = std::chrono::steady_clock::now();
 
             if (now > m_next_scan_timestamp_interval) {
                 m_is_single_scan = false;
+                LOG(DEBUG) << "interval condition is met, trigger scan";
                 fsm_move_state(eState::TRIGGER_SCAN);
                 m_last_scan_try_timestamp = now;
-                LOG(DEBUG) << "interval condition is met, trigger scan";
             }
         }
         break;
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -18,6 +18,9 @@
 std::string s_ar_states[] = {FOREACH_DCS_STATE(GENERATE_STRING)};
 std::string s_ar_events[] = {FOREACH_DCS_EVENT(GENERATE_STRING)};
 
+int dynamic_channel_selection_task::m_scanning_task_id =
+    -1; // Currently scanning task id. -1 means no scan in progress.
+
 dynamic_channel_selection_task::dynamic_channel_selection_task(db &database_,
                                                                ieee1905_1::CmduMessageTx &cmdu_tx_,
                                                                task_pool &tasks_,
@@ -115,6 +118,10 @@ void dynamic_channel_selection_task::work()
                 m_last_scan_try_timestamp = now;
             }
         }
+        break;
+    }
+    case eState::PRE_SCAN: {
+        LOG(TRACE) << "PRE_SCAN, mac=" << m_radio_mac;
         break;
     }
     case eState::TRIGGER_SCAN: {
@@ -435,3 +442,29 @@ void dynamic_channel_selection_task::fsm_move_state(eState new_state)
     m_fsm_state = new_state;
 }
 bool dynamic_channel_selection_task::fsm_in_state(eState state) { return m_fsm_state == state; }
+
+bool dynamic_channel_selection_task::start_scan()
+{
+    if (m_scanning_task_id != -1) {
+        return false;
+    }
+    // set scanning task id to current task id
+    m_scanning_task_id = id;
+    return true;
+}
+
+bool dynamic_channel_selection_task::finish_scan()
+{
+    if (m_scanning_task_id == -1) {
+        return true;
+    }
+
+    if (m_scanning_task_id != id) {
+        TASK_LOG(ERROR) << "finish scan is called for a task that isn't the scanning task";
+        return false;
+    }
+
+    // scan by current task is finished
+    m_scanning_task_id = -1;
+    return true;
+}
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -107,21 +107,25 @@ void dynamic_channel_selection_task::work()
             m_is_single_scan         = true;
             m_is_single_scan_pending = false;
             LOG(DEBUG) << "single scan is pending, trigger single scan";
-            fsm_move_state(eState::TRIGGER_SCAN);
+            fsm_move_state(eState::PRE_SCAN);
         } else if (database.get_channel_scan_is_enabled(m_radio_mac)) {
-            auto now = std::chrono::steady_clock::now();
-
-            if (now > m_next_scan_timestamp_interval) {
+            if (std::chrono::steady_clock::now() > m_next_scan_timestamp_interval) {
                 m_is_single_scan = false;
                 LOG(DEBUG) << "interval condition is met, trigger scan";
-                fsm_move_state(eState::TRIGGER_SCAN);
-                m_last_scan_try_timestamp = now;
+                fsm_move_state(eState::PRE_SCAN);
             }
         }
         break;
     }
     case eState::PRE_SCAN: {
-        LOG(TRACE) << "PRE_SCAN, mac=" << m_radio_mac;
+        if (start_scan()) {
+            LOG(TRACE) << "PRE_SCAN, mac=" << m_radio_mac << ", scan_type is "
+                       << ((m_is_single_scan) ? "single-scan" : "continuous-scan");
+            if (!m_is_single_scan) {
+                m_last_scan_try_timestamp = std::chrono::steady_clock::now();
+            }
+            fsm_move_state(eState::TRIGGER_SCAN);
+        }
         break;
     }
     case eState::TRIGGER_SCAN: {
@@ -213,6 +217,10 @@ void dynamic_channel_selection_task::work()
     }
     case eState::FINISH: {
         LOG(TRACE) << "finish scan for mac=" << m_radio_mac;
+
+        if (!finish_scan()) {
+            TASK_LOG(ERROR) << "scan abort event received for a task that isn't the scanning task";
+        }
 
         database.set_channel_scan_results_status(m_radio_mac, m_last_scan_error_code,
                                                  m_is_single_scan);
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
@@ -52,6 +52,7 @@ public:
 #define FOREACH_DCS_STATE(STATE)                                                                   \
     STATE(INIT)                                                                                    \
     STATE(IDLE)                                                                                    \
+    STATE(PRE_SCAN)                                                                                \
     STATE(TRIGGER_SCAN)                                                                            \
     STATE(WAIT_FOR_SCAN_TRIGGERED)                                                                 \
     STATE(WAIT_FOR_RESULTS_READY)                                                                  \
@@ -72,6 +73,8 @@ private:
     void dcs_wait_for_event(eEvent cs_event);
     void fsm_move_state(eState new_state);
     bool fsm_in_state(eState state);
+    bool start_scan();
+    bool finish_scan();
 
     beerocks::eChannelScanErrCode dcs_request_scan_trigger();
     beerocks::eChannelScanErrCode dcs_request_scan_dump();
@@ -91,6 +94,8 @@ private:
 
     bool m_is_single_scan_pending = false;
     bool m_is_single_scan         = false;
+
+    static int m_scanning_task_id; // Currently scanning task id. -1 means no scan in progress.
 };
 } //namespace son
 #endif
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
@@ -57,7 +57,8 @@ public:
     STATE(WAIT_FOR_RESULTS_READY)                                                                  \
     STATE(WAIT_FOR_RESULTS_DUMP)                                                                   \
     STATE(SCAN_DONE)                                                                               \
-    STATE(ABORT_SCAN)
+    STATE(ABORT_SCAN)                                                                              \
+    STATE(FINISH)
 
     // State ENUM
     enum class eState { FOREACH_DCS_STATE(GENERATE_ENUM) };
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
@@ -27,7 +27,7 @@ class dynamic_channel_selection_task : public task, public std::enable_shared_fr
 public:
     dynamic_channel_selection_task(db &database, ieee1905_1::CmduMessageTx &cmdu_tx_,
                                    task_pool &tasks_, sMacAddr radio_mac_);
-    virtual ~dynamic_channel_selection_task() {}
+    virtual ~dynamic_channel_selection_task() { finish_scan(); }
 
     struct sScanEvent {
         sMacAddr radio_mac;


controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -90,6 +91,8 @@ class dynamic_channel_selection_task : public task, public std::enable_shared_fr
 
     bool m_is_single_scan_pending = false;
     bool m_is_single_scan         = false;
+
+    static int m_scanning;
```
tomereli:
```suggestion
    static int m_scanning_task_id = -1; // Currently scanning task id. -1 means no scan in progress.
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -20,6 +20,7 @@
 // Helper MACROs for Enum/String generation
 #define GENERATE_ENUM(ENUM) ENUM,
 #define GENERATE_STRING(STRING) #STRING,
+#define NO_SCAN_IN_PROGRESS -1
```
tomereli: If you change `m_scanning` to `m_scanning_task_id`, no need for this define.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -206,6 +224,13 @@ void dynamic_channel_selection_task::work()
     case eState::ABORT_SCAN: {
         LOG(ERROR) << "aborting scan for mac=" << m_radio_mac << ", last_scan_timestamp is not set";
 
+        if (m_scanning != id) {
+            TASK_LOG(ERROR) << "scan abort event received for a task that isn't the scanning task";
+        } else {
+            // scan by current task is aborted
+            m_scanning = NO_SCAN_IN_PROGRESS;
```
tomereli:
```suggestion
            m_scanning_task_id = -1;
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -206,6 +224,13 @@ void dynamic_channel_selection_task::work()
     case eState::ABORT_SCAN: {
         LOG(ERROR) << "aborting scan for mac=" << m_radio_mac << ", last_scan_timestamp is not set";
 
+        if (m_scanning != id) {
```
tomereli:
```suggestion
        if (m_scanning_task_id != id) {
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -187,6 +195,16 @@ void dynamic_channel_selection_task::work()
...
+            TASK_LOG(ERROR) << "scan done event received for a task that isn't the scanning task";
+            m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
+            fsm_move_state(eState::ABORT_SCAN);
+            break;
+        }
+
+        // scan by current task is complete
+        m_scanning = NO_SCAN_IN_PROGRESS;
```
tomereli:
```suggestion
        m_scanning_task_id = -1; //mark no task is currently scanning
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -118,6 +120,12 @@ void dynamic_channel_selection_task::work()
         break;
     }
     case eState::TRIGGER_SCAN: {
+        if (m_scanning != NO_SCAN_IN_PROGRESS) {
+            break;
+        }
+        // set scanning task id to current task id
+        m_scanning = id;
```
tomereli:
```suggestion
        m_scanning_task_id = id;
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -18,6 +18,8 @@
 std::string s_ar_states[] = {FOREACH_DCS_STATE(GENERATE_STRING)};
 std::string s_ar_events[] = {FOREACH_DCS_EVENT(GENERATE_STRING)};
 
+int dynamic_channel_selection_task::m_scanning = NO_SCAN_IN_PROGRESS;
```
tomereli:
```suggestion
int dynamic_channel_selection_task::m_scanning = -1; //no scan in progress
```

tomereli (approved):
Small changes which I think you should do, but nothing that requires re-review.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
vitalybu:
```suggestion
    static int m_scanning_task_id;
```

adam1985d: static initialization is done inside the cpp file. added the comment as suggested

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
vitalybu: No need for this define, we use `-1` for invalid task id.
Maybe should add an `INVALID_TASK_ID` to the base class (but it's not part of this PR anyway).

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -118,6 +120,12 @@ void dynamic_channel_selection_task::work()
         break;
     }
     case eState::TRIGGER_SCAN: {
+        if (m_scanning != NO_SCAN_IN_PROGRESS) {
```
vitalybu: When moving to the `TRIGGER_SCAN` state from the `IDLE` state, we also initialize the `m_last_scan_try_timestamp` member. If now the task will block until the previous scan is complete, what impact does it have on this timeout?
Maybe it's better to remain in `IDLE` state? or even in `INIT`?

vitalybu (changes requested):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
adam1985d: changed `m_scanning` to `m_scanning_task_id`

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
adam1985d: done

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: good point, but I think it is better to move the `m_last_scan_try_timestamp` set point to where we actually trigger the scan - meaning the `TRIGGER_SCAN` state.
will move the set of `m_last_scan_try_timestamp` to `TRIGGER_SCAN` state - right after setting the `m_scanning_task_id`

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -108,23 +108,22 @@ void dynamic_channel_selection_task::work()
             LOG(DEBUG) << "single scan is pending, trigger single scan";
             fsm_move_state(eState::TRIGGER_SCAN);
         } else if (database.get_channel_scan_is_enabled(m_radio_mac)) {
-            auto now = std::chrono::steady_clock::now();
-
-            if (now > m_next_scan_timestamp_interval) {
+            if (std::chrono::steady_clock::now() > m_next_scan_timestamp_interval) {
```
vitalybu, line 112: I still think that there's no sense to move into the `TRIGGER_SCAN` state if you can't really trigger a scan.
The task should remain in `IDLE` until it's able to trigger a scan.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -195,15 +194,15 @@ void dynamic_channel_selection_task::work()
     case eState::SCAN_DONE: {
         LOG(TRACE) << "SCAN_DONE";
 
-        if (m_scanning != id) {
+        if (m_scanning_task_id != id) {
             TASK_LOG(ERROR) << "scan done event received for a task that isn't the scanning task";
             m_last_scan_error_code = beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
             fsm_move_state(eState::ABORT_SCAN);
```
vitalybu: Why do we `ABORT` the scan in this case? How can this even happen?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
adam1985d: removed the define

vitalybu (changes requested):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: please see the last fixup - it addresses both.
I will edit the commit message accordingly as part of the squash

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -457,4 +447,25 @@ void dynamic_channel_selection_task::fsm_move_state(eState new_state)
...
+bool dynamic_channel_selection_task::finish_scan()
+{
+    if (m_scanning_task_id != id) {
+        TASK_LOG(ERROR) << "finish scan is called for a task that isn't the scanning task";
+        return false;
+    }
+    // scan by current task is aborted
+    m_scanning_task_id = -1;
+    return true;
+}
```
vitalybu: Missing new line

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -194,16 +190,13 @@ void dynamic_channel_selection_task::work()
     case eState::SCAN_DONE: {
         LOG(TRACE) << "SCAN_DONE";
 
-        if (m_scanning_task_id != id) {
+        if (!finish_scan()) {
```
vitalybu: A better solution may be to add a new `FINISH` state, where you finalize everything.
Then you won't have to do this from multiple states.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -102,28 +102,24 @@ void dynamic_channel_selection_task::work()
         break;
     }
     case eState::IDLE: {
-        if (m_is_single_scan_pending) {
+        if (m_is_single_scan_pending && start_scan()) {
```
vitalybu: IMO It's better to call this function right before moving to the `TRIGGER_SCAN` state.
Something like this:
```c++
auto now = std::chrono::steady_clock::now();
if (m_is_single_scan_pending) {
    ...
} else if (database.get_channel_scan_is_enabled(m_radio_mac) {
    ...
} else {
    break;
}

if (start_scan()) {
    m_last_scan_try_timestamp = now;
    fsm_move_state(eState::TRIGGER_SCAN);
}
```

In this case you actually start the scan (`start_scan()`) only when all conditions are met and no other task is currently scanning.

vitalybu (changes requested):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: true - I intend to run the clang after the squash

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: but the solution you suggest updates the `m_last_scan_try_timestamp` for a single scan as well.
also, you don't want to set the data members only to then break due to start_scan() returning false

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: it shouldn't happen - but wanted to be extra safe here and not update the `m_scanning_task_id` and the results if for some reason this state is reached even though we don't own the scanning-permission

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
vitalybu: Also true, but can be solved by adding another state :)
`IDLE` -> `CAN_SCAN` -> `TRIGGER_SCAN` -> ... -> `FINISH`

The FSM will reach the `CAN_SCAN` state only when it's ready to scan.
It'll wait in this state until it can actually trigger a scan, and only then move to `TRIGGER_SCAN`.
`FINISH` will always do all the cleanups (including removing the id of the scanning task).

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: I agree - but this is more of refactoring the scan_done and abort_scan cleanups to be in a new state - and less of the actual add of protection.

I added this as part of the PR.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: it is more of `SHOULD_SCAN` rather than `CAN_SCAN` since the `start_scan()` itself is the condition of can-scan

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: good idea, steering task also use such a state

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: please re-review. provided a solution that is based on your suggestion.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: I added a PRE_SCAN state and FINISH state - please re-review

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -432,4 +449,25 @@ void dynamic_channel_selection_task::fsm_move_state(eState new_state)
...
+}
+
+bool dynamic_channel_selection_task::finish_scan()
+{
+    if (m_scanning_task_id != id) {
+        TASK_LOG(ERROR) << "finish scan is called for a task that isn't the scanning task";
+        return false;
+    }
+    // scan by current task is aborted
+    m_scanning_task_id = -1;
```
kantera800, line 476: I'm pretty sure that a task can be erased without going through finish_scan().
I'm wondering if this is the only place for clearing this flag.

maybe we should clear it in the destructor
` ~dynamic_channel_selection_task() { m_scanning_task_id = -1 }`

or 

```
void dynamic_channel_selection_task::handle_task_end()
{
  m_scanning_task_id = -1;
}
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -103,19 +106,28 @@ void dynamic_channel_selection_task::work()
         if (m_is_single_scan_pending) {
             m_is_single_scan         = true;
             m_is_single_scan_pending = false;
```
kantera800, line 108: (nitpick ) I know its not part of this PR but `m_is_single_scan_pending` and `m_is_single_scan_pending` confused me .
i prefer `m_is_manual_scan` and also would add a comment in code why do we need to treat manual scan differently from continues one .

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
vitalybu: That's a good point, but instead of setting `m_scanning_task_id = -1;`, you should just call `finish_scan();` from the destructor (as it also make checks that the current task if the owner of this flag).

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: will add it as a separate hotfix, but for discussion, we use the "single" terminology across the entire DCS feature implementation - why should the task name it differently?
we can change it to `m_handling_single_scan_request` if it would be more clear.

secondly - the `m_is_single_scan_pending` is a flag that indicates that a single scan request was made and haven't been treated yet.
the `m_is_single_scan` is used through the entire task to determine which scan params to read and which scan results to update. where do you think a comment regarding this should be added?

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: added a fixup commit that calls `finish_scan();` as part of the DTOR

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
kantera800: > will add it as a separate hotfix, but for discussion, we use the "single" terminology across the entire DCS feature implementation - why should the task name it differently?
> we can change it to `m_handling_single_scan_request` if it would be more clear.
>
In that case just adding "request" would do 
`m_single_scan_request` 
`m_single_scan_request_pending`

> 
> secondly - the `m_is_single_scan_pending` is a flag that indicates that a single scan request was made and haven't been treated yet.
> the `m_is_single_scan` is used through the entire task to determine which scan params to read and which scan results to update. where do you think a comment regarding this should be added?

i would add it to the header file

kantera800 (approved):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -432,4 +449,25 @@ void dynamic_channel_selection_task::fsm_move_state(eState new_state)
...
+        return false;
+    }
+    // set scanning task id to current task id
+    m_scanning_task_id = id;
+    return true;
+}
+
+bool dynamic_channel_selection_task::finish_scan()
+{
+    if (m_scanning_task_id != id) {
```
vitalybu, line 470: Since now you're also calling this method from the destructor, you should also check that `m_scanning_task_id != -1`, otherwise it will always be printed:
```suggestion
    if (m_scanning_task_id != -1 && m_scanning_task_id != id) {
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
vitalybu: Still missing

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -432,4 +449,25 @@ void dynamic_channel_selection_task::fsm_move_state(eState new_state)
...
+        TASK_LOG(ERROR) << "finish scan is called for a task that isn't the scanning task";
+        return false;
+    }
+    // scan by current task is aborted
```
vitalybu:
```suggestion
    
    // scan by current task is finished
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: I will add a separate check:
```
if (m_scanning_task_id == -1) {
    return true;
}
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: fixed

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
adam1985d: fixed

vitalybu (approved):

PR #1139, opened 2020-04-13
son_slave_thread: trigger measurement request on sta connection

vitalii-komisarenko:
At the moment Associated STA Link Metrics collection is triggered by controller.
It is a vendor-specific message and it is sent just after STA connection.

Vendor-specific approach has disadvantage of not supporting non prplMesh contollers
and it is just redundant since there is a non-VS message for doing similar work.

We would like to replace controller-agent communication with non-VS messages,
but we would like to do it in tiny steps in order to not break anything.

This commit triggers statistics collection on association without
removing any code. Old flows need to be removed/refactored in future commits.

Signed-off-by: Vitalii Komisarenko <vitalii.komisarenko@gmail.com>


CoralMalachi (approved):

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -394,6 +394,45 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
     return true;
 }
 
+bool slave_thread::start_monitoring(const beerocks_message::sClientMonitoringParams &params,
+                                    std::shared_ptr<beerocks_header> beerocks_header)
+{
+    LOG(DEBUG) << __func__;
```
tomereli: This is not very useful - we have a more comprehensive debug log below.
```suggestion
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -394,6 +394,45 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
     return true;
 }
 
+bool slave_thread::start_monitoring(const beerocks_message::sClientMonitoringParams &params,
+                                    std::shared_ptr<beerocks_header> beerocks_header)
```
tomereli: It is better to pass `beerocks_header` as reference. It is also in our [CODINGSTYLE.md](https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -394,6 +394,45 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
...
+{
+    LOG(DEBUG) << __func__;
+
+    std::string client_mac              = network_utils::mac_to_string(params.mac);
+    std::string client_bridge_4addr_mac = network_utils::mac_to_string(params.bridge_4addr_mac);
+    std::string client_ip               = network_utils::ipv4_to_string(params.ipv4);
+
+    LOG(DEBUG) << "START_MONITORING_REQUEST: mac=" << client_mac << " ip=" << client_ip
+               << " bridge_4addr_mac=" << client_bridge_4addr_mac
+               << " channel=" << std::to_string(params.channel)
```
tomereli: If you're printing channel, better also print the bandwidth. As a matter of fact, now it makes more sense to also print the operating class if its available in this context.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -394,6 +394,45 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
     return true;
 }
 
+bool slave_thread::start_monitoring(const beerocks_message::sClientMonitoringParams &params,
```
tomereli: It is great that you split the start monitoring to a separate function. However, this should be done in a separate commit.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -394,6 +394,45 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
...
+        }
+        request_out->mac() = params.mac;
+        message_com::send_cmdu(ap_manager_socket, cmdu_tx);
+    }
+
+    //notify monitor
+    auto request_out = message_com::create_vs_message<
+        beerocks_message::cACTION_MONITOR_CLIENT_START_MONITORING_REQUEST>(cmdu_tx,
+                                                                           beerocks_header->id());
+    if (request_out == nullptr) {
```
tomereli:
```suggestion
    if (!request_out) {
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2147,7 +2157,17 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
 
         send_cmdu_to_controller(cmdu_tx);
 
-        break;
+        beerocks_message::sClientMonitoringParams monitoring_params;
```
tomereli: Since you made start_monitoring a class method, all this can and should be done inside the function, to keep this switch case even shorter.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2147,7 +2157,17 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
...
+        monitoring_params.mac = notification_in->params().mac;
+        monitoring_params.bridge_4addr_mac =
+            network_utils::mac_from_string(backhaul_params.bridge_mac);
+        monitoring_params.ipv4    = network_utils::ipv4_from_string(network_utils::ZERO_IP_STRING);
+        monitoring_params.channel = backhaul_params.backhaul_channel; // TODO: unused?
```
tomereli: If it is indeed unused, please remove this in a separate commit.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2147,7 +2157,17 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
...
+        monitoring_params.vap_id  = notification_in->params().vap_id;
+        monitoring_params.is_ire  = 0;
```
tomereli: This looks wrong. I don't understand the logic completely (perhaps @morantr or @vitalybu can shed some light), but IRE stands for "Intelligent range extender", which means its a repeater (well, range extender differs from repeater in that it has 3 radios, meaning a dedicated radio for backhaul).

From the looks of it, it looks like the whole IRE VS messaging path here should be removed. It is sending a notification to the ap manager that the connected client is a 4addr client, and this whole path is now dead code as it seems. But this should be removed in a separate PR.

For now, I would just remove the is_ire from here completely (in a separate commit).

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -515,37 +554,8 @@ bool slave_thread::handle_cmdu_control_message(Socket *sd,
...
-            cmdu_tx, beerocks_header->id());
-        if (request_out == nullptr) {
-            LOG(ERROR) << "Failed building ACTION_MONITOR_CLIENT_START_MONITORING_REQUEST message!";
-            return false;
-        }
-        request_out->params() = request_in->params();
-        message_com::send_cmdu(monitor_socket, cmdu_tx);
-        break;
+        LOG(DEBUG) << "Call start_monitoring from ACTION_CONTROL_CLIENT_START_MONITORING_REQUEST";
+        return start_monitoring(request_in->params(), beerocks_header);
```
tomereli: What about stop_monitoring? Should be a separate API for that, and it should be called when a client disassociates, doesn't it?

tomereli (changes requested):

agent/src/beerocks/slave/son_slave_thread.cpp
vitalii-komisarenko: To be honest, I don't have much confidence in these values.
According to the logs from `./test_flows.py`, `ipv4` and `bridge_4addr_mac` are always zero.
Setting `is_ire` to non-zero value breaks `client_steering_dummy` test.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -394,6 +394,42 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
...
+            return false;
+        }
+        request_out->mac() = params.mac;
+        message_com::send_cmdu(ap_manager_socket, cmdu_tx);
+    }
+
+    //notify monitor
+    auto request_out = message_com::create_vs_message<
+        beerocks_message::cACTION_MONITOR_CLIENT_START_MONITORING_REQUEST>(cmdu_tx,
+                                                                           beerocks_header.id());
```
arnout: This is the only use of beerocks_header, so I don't think it makes much sense to pass the entire header.

For the new call, the id parameter can just be zero.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -394,6 +394,42 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
...
+        auto request_out = message_com::create_vs_message<
+            beerocks_message::cACTION_APMANAGER_CLIENT_IRE_CONNECTED_NOTIFICATION>(cmdu_tx);
+        if (!request_out) {
+            LOG(ERROR) << "Failed building ACTION_APMANAGER_CLIENT_IRE_CONNECTED_NOTIFICATION "
+                          "message!";
+            return false;
+        }
+        request_out->mac() = params.mac;
+        message_com::send_cmdu(ap_manager_socket, cmdu_tx);
+    }
```
arnout: This part is only relevant when handling the VS message, not when we trigger it autonomously. So I'd leave this outside of the function.

That also resolves Tomer's question about is_ire.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: That's exactly what Tomer is saying: the is_ire != 0 part should be removed entirely.

arnout (approved):
I believe this PR is good enough to be merged with small changes.

agent/src/beerocks/slave/son_slave_thread.cpp
vitalii-komisarenko: @tomereli , @arnout  Shall I remove the `ACTION_APMANAGER_CLIENT_IRE_CONNECTED_NOTIFICATION` , the list of IREs (`inserted_ire`) managed by ap manager and all the code that uses the list?

Or just remove `is_ire != 0` path from `ACTION_APMANAGER_CLIENT_ASSOCIATED_NOTIFICATION`?

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -757,7 +756,6 @@ bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
         }
 
         sta_node = mon_db.sta_add(sta_mac, vap_id);
-        sta_node->set_ipv4(sta_ipv4);
```
morantr, line 818: This commit will cause serious bugs. Please drop this commit.
It is not a good idea to remove code that is being used if you are not 100% sure what it does and how it works.

morantr (changes requested):
Sorry, but I can't approve it.
First, it would really be nice if there was a DR task on that issue mentioned in the PR description.

Second, the drop of the IP from the message it wrong. The monitor implements an arp sending mechanism that uses this IP which being sent by the controller. So removing it will cause unconnected measurements to fail.

I would implement it differently:
1. The start_monitoring does not need to be sent by the agent. You cant add STA-CONNECT handler to monitor bwl, so the monitor gets this event (and also the disconnect event) directly from bwl, and start/stop monitoring directly (which is basically add/remove the client from mon_db).
2. Add a message that updates the monitor with the IP address. The son_slave receives it from the platform manager so all you need to do is to forward it to the monitor and update the station IP add on mon_db.
3. Remove the start/stop_monitoring message from everywhere.

agent/src/beerocks/monitor/monitor_thread.cpp
tomereli: @morantr Please explain which bugs this commit will cause. The point of code review is also to teach newcomers and not only criticize their proposed changes.

agent/src/beerocks/monitor/monitor_thread.h
```diff
@@ -54,6 +54,18 @@ class monitor_thread : public beerocks::socket_thread {
     virtual std::string print_cmdu_types(const beerocks::message::sUdsHeader *cmdu_header) override;
 
 private:
+    /**
+     * @brief mutex to protect the list of STAs being monitored
+     */
+    std::mutex monitoring_sta_mutex;
```
vitalybu: What's that for?
Monitor runs in a single context..

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -653,6 +653,70 @@ bool monitor_thread::update_ap_stats()
...
+{
+    const std::lock_guard<std::mutex> lock(monitoring_sta_mutex);
+
+    if (!_start_monitoring_sta(sta_mac, set_bridge_4addr_mac, vap_id)) {
+        LOG(ERROR) << "_start_monitoring_sta with mac " << sta_mac << ", bridge_mac "
+                   << set_bridge_4addr_mac << ", vap_id " << vap_id << " failed!";
+        return false;
+    }
+
+    if (!_monitored_sta_set_ipv4(sta_mac, sta_ipv4)) {
```
vitalybu: Why a separate method for setting the IP address?
Can't you just pass it to `_start_monitoring_sta()`?

agent/src/beerocks/monitor/monitor_thread.h
vitalii-komisarenko: Just a habit to add mutexes. But since it is a single thread, I will remove it.

agent/src/beerocks/monitor/monitor_thread.cpp
vitalii-komisarenko: At the moment, I am moving in direction described by https://github.com/prplfoundation/prplMesh/pull/1139#pullrequestreview-398271772 .
BWL will give me `mac` and `vap_id` at the same time, but `ipv4` will arrive a bit later from the slave.

arnout:
I haven't reviewed yet, however:
- 28 commits is a bit too big to review in one shot. Would it be possible/easy to split into 2-3 PRs?
- We are still missing an update to tests/certification/passing_agent_tests.txt to add 4.7.7 to the passing tests.
- The last commit of the PR should trigger the 4.7.7 tests with the magic string in the commit message.

vitalii-komisarenko:
Replaced with #1290, #1289 and #1291.

PR #1141, opened 2020-04-13
transport: change interface map to use ifname instead of ifindex 

### LINKED ISSUE #1137: [BUG] ieee1905_transport: Not tracking WLAN VAP interfaces
**Describe the bug** -
When prplmesh is started, all WLAN VAP interfaces (wlan0.0, wlan2.0, etc.) are registered by transport using their ifindex.
During autoconfiguration, the VAP interfaces are removed and brought back up by the driver, causing their ifindex to change, therefore not to be tracked anymore by the transport.
This means that any 1905 messages received on the wireless bAPs are simply not handled by the ieee1905 transport service and therefore never get to their destination.

## Steps to reproduce

1. Build branch `dev/bap-wps-onboarding`
2. Run controller & agent
3. See in `/tmp/beerocks/logs/transport.log` that the interface index changes and are ignored

## Expected behavior

The ieee1905 transport should pick up any interface index changes and continue monitoring them.

Logs Attached -
[transport.log](https://github.com/prplfoundation/prplMesh/files/4468612/transport.log)

### LINKED ISSUE #1153: [TASK] CI: add a local shell wrapper for cppcheck
Same as ##1117 but for cppcheck

tomereli:
The ieee1905 transport keeps track of network interfaces using a map of
interfaces where the keys are the interface indexes.
Configuration of the transport to allow an interface to take part in the
ieee1905 networking is done by publishing an interface configuration
message on the bus, which is done once in the BTL for all network
interfaces under the bridge.
Some network interfaces indexes may change in the kernel as part of
autoconfiguration which removes the VAPs and brings them back on, which
causes the interfaces not to be tracked by the ieee1905 transport.

The simplest way to overcome this is to keep track of interface names
instead of indexes, as those will not change (and if they will, prplmesh
needs to be restarted).

In addition, fix a bunch of cppcheck issues. While at it, also add a cppcheck shell wrapper.

Fixes #1137 
Fixes #1153


framework/transport/ieee1905_transport/ieee1905_transport_local_bus.cpp
```diff
@@ -156,20 +159,36 @@ bool Ieee1905Transport::send_packet_to_local_bus(Packet &packet)
     return true;
 }
 
+static void copy_string(char *dst, const char *src, size_t dst_len)
+{
+    const char *src_end = std::find((char *)src, ((char *)src) + dst_len, '\0');
```
rmelotte: Why are you casting away the constness of src? 
(I'm surprised cppcheck doesn't report it)

tomereli: I have no idea - I just copied this from BCL, this is how its implemented there. I will fix it in both places

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport_messages.h
```diff
@@ -268,9 +269,10 @@ class InterfaceConfigurationMessage : public Message {
     };
 
     struct Interface {
-        uint32_t if_index = 0; // the linux interface index of the specified interface
-        uint32_t bridge_if_index =
-            0; // the interface index of the bridge to which this interface belong to (or zero if it does not belong to a bridge)
+        // the linux interface index of the specified interface
+        char ifname[IF_NAMESIZE];
+        // the interface index of the bridge to which this interface belong to (or zero if it does not belong to a bridge)
+        char bridge_ifname[IF_NAMESIZE];
```
arnout, line 278: As noted in https://github.com/prplfoundation/prplMesh/issues/1137#issuecomment-613941348 I'm not so sure that using ifname for interfaces is a good idea. However, I could live with it just to make progress.

That said, I think it is definitely *not* a good idea to use an interface name for the bridge. If the bridge gets deleted while prplMesh is running, I think we die a thousand deads anyway... On the other hand, if an interface does get renamed, the bridge is the most likely one to be renamed. And finally, for the issue you're trying to fix, there is no need at all for the bridge to be accessed by name rather than by index. So, I tend to prefer not to change it.

That said, if my vague plan goes through, this configuration message will be removed completely anyway, so it doesn't matter much.

common/beerocks/btl/btl_local_bus.cpp
```diff
@@ -106,30 +107,28 @@ bool transport_socket_thread::configure_ieee1905_transport_interfaces(
...
+    THREAD_LOG(DEBUG) << "adding bridge " << bridge_iface
+                      << " to ieee1905 transport, bridge iface=" << bridge_iface;
     for (const auto &iface : ifaces) {
-        int if_index = network_utils::linux_get_iface_index(iface);
-        interface_configuration_request_msg.metadata()->interfaces[n].if_index = if_index;
-        interface_configuration_request_msg.metadata()->interfaces[n].bridge_if_index =
-            bridge_if_index;
+        string_utils::copy_string(
+            interface_configuration_request_msg.metadata()->interfaces[n].ifname, iface.c_str(),
+            IF_NAMESIZE);
```
arnout: (nitpick) would be nice to add a variant of copy_string that takes a std::string as argument.

framework/transport/ieee1905_transport/test/ieee1905_transport_test.cpp
```diff
@@ -251,18 +251,10 @@ int main(int argc, char *argv[])
     while ((c = getopt(argc, argv, "b:i:u:x")) != -1) {
         switch (c) {
         case 'b':
-            ieee1905_bridge_if_index = if_nametoindex(optarg);
-            if (ieee1905_bridge_if_index == 0) {
-                fprintf(stderr, "unknown interface %s\n", optarg);
-                return false;
-            }
+            ieee1905_bridge_ifname = std::string(optarg, IF_NAMESIZE);
```
arnout: This will do a serious out-of-bounds access, because it copies exactly IF_NAMESIZE characters even if optarg is shorter!
```suggestion
            ieee1905_bridge_ifname = std::string(optarg);
```

framework/transport/ieee1905_transport/ieee1905_transport_local_bus.cpp
arnout: Instead of copying it from BCL, move it from BCL to mapf::utils where everyone can use it.

framework/transport/ieee1905_transport/ieee1905_transport.cpp
```diff
@@ -74,21 +74,20 @@ void Ieee1905Transport::run()
 
         // check for events on all active network interface sockets
         for (auto it = network_interfaces_.begin(); it != network_interfaces_.end(); ++it) {
-            unsigned int if_index   = it->first;
-            auto &network_interface = it->second;
+            auto &iface = it->second;
```
arnout: I prefer to have such a rename (`network_interface` -> `iface`) in a separate commit.

However, in this case I don't see a reason for the rename. The longer name is more descriptive and its meaning doesn't change, so why rename it?

framework/transport/ieee1905_transport/ieee1905_transport_netlink.cpp
```diff
@@ -70,11 +70,15 @@ int Ieee1905Transport::handle_netlink_message(struct nlmsghdr *msghdr)
     if (msghdr->nlmsg_type == RTM_NEWLINK || msghdr->nlmsg_type == RTM_DELLINK) {
         struct ifinfomsg *ifi = (struct ifinfomsg *)NLMSG_DATA(msghdr);
         bool is_active        = (ifi->ifi_flags & IFF_RUNNING) && (ifi->ifi_flags & IFF_UP);
-
+        std::string ifname    = if_index2name(ifi->ifi_index);
```
arnout, line 73: Note that the message is followed by the attributes of the interface, one of which is IFLA_IFNAME. So we could just parse those attributes instead of opening a socket (which if_index2name does).

(nitpick)

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
```diff
@@ -64,16 +64,20 @@ class Ieee1905Transport {
     // this table holds all the network interfaces that are to be used by
     // the transport.
     //
-    // interface index (if_index) is used as Key to the table
+    // interface name (ifname) is used as Key to the table
     struct NetworkInterface {
-        int fd =
-            -1; // the file descriptor of the socket bound to this interface (or -1 if inactive)
+        // the file descriptor of the socket bound to this interface (or -1 if inactive)
```
arnout: If you're anyway changing this, please also doxygenify it:
```suggestion
        /// the file descriptor of the socket bound to this interface (or -1 if inactive)
```

framework/transport/ieee1905_transport/test/ieee1905_transport_test.cpp
```diff
@@ -388,15 +388,15 @@ int main(int argc, char *argv[])
         if (!ieee1905_bridge_ifname.empty()) {
             using Flags = InterfaceConfigurationRequestMessage::Flags;
             strncpy(interface_configuration_request_msg.metadata()->interfaces[n].ifname,
-                    ieee1905_bridge_ifname.c_str(), IF_NAMESIZE);
+                    ieee1905_bridge_ifname.c_str(), IF_NAMESIZE - 1);
```
arnout: This seems to be an unrelated change...

If you're doing this, use copy_string like everywhere else. Because now it's actually still not correct since the terminating '\0' is not added.

arnout (approved):
Even though I think it is going in the wrong direction, the right direction is a much bigger change. So we can merge this IMO.

I still have a bunch of review comments which need to be addressed, but I don't need to re-review.

framework/transport/ieee1905_transport/ieee1905_transport_local_bus.cpp
tomereli: Done - wasn't so simple since I had to link BCL to mapf::common and update all callers to use the new API location, but its done now. We need to move everything that should be shared to mapf::common eventually - so that BCL will be merged into mapf::common.

common/beerocks/btl/btl_local_bus.cpp
tomereli: I don't see the benefit of such API if we're just saving the call to str->c_str().

framework/transport/ieee1905_transport/ieee1905_transport.cpp
tomereli: Reverted the change

framework/transport/ieee1905_transport/ieee1905_transport_netlink.cpp
tomereli: I think that parsing the attribute here is less readable than using indextoname, and the overhead of creating a socket is negligible.

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
tomereli: I had no idea this should be using `///` instead of `//`. Done.

framework/transport/ieee1905_transport/test/ieee1905_transport_test.cpp
tomereli: Changed to use copy_string.

vitalybu (approved):

PR #1145, opened 2020-04-14
Feature/Packet capture using tshark processing of tcpdump capture

### LINKED ISSUE #942: [TASK] Capture packets in test_flows.py using tshark

Galvien:
As a third attempt, this change allows easy, **fast** and simple analysis of CMDU messages passing through the container's bridge interface.

The general idea is that each running test is monitored by tcpdump, then saves the packet data into a pcap file, which is later quickly scanned by tshark.

In addition, a helper method extends the dictionary with direct, casted entries of more important values from the tree - for easier access and handling.

### DIFF
--- a/tests/opts.py
+++ b/tests/opts.py
@@ -9,7 +9,6 @@
 class opts:
     '''Static class that encodes the global options.'''
     verbose = False
-    tcpdump = False
     tcpdump_dir = ''
     stop_on_failure = False
 
--- a/tests/sniffer.py
+++ b/tests/sniffer.py
@@ -99,12 +99,14 @@ class Sniffer:
         self.tcpdump_log_dir = tcpdump_log_dir
         self.tcpdump_proc = None
         self.current_outputfile = None
+        self.checkpoint_frame_number = 0
 
     def start(self, outputfile_basename):
         '''Start tcpdump to outputfile.'''
         debug("Starting tcpdump, output file {}.pcap".format(outputfile_basename))
         os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
         self.current_outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
+        self.checkpoint_frame_number = 0
         command = ["tcpdump", "-i", self.interface, '-U', '-w', self.current_outputfile]
         self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
         # tcpdump takes a while to start up. Wait for the appropriate output before continuing.
@@ -126,7 +128,8 @@ class Sniffer:
         if not self.current_outputfile:
             err("get_packet_capture but no capture file")
             return []
-        tshark_command = ['tshark', '-r', self.current_outputfile, '-T', 'json']
+        tshark_command = ['tshark', '-r', self.current_outputfile, '-T', 'json',
+                          '-Y', 'frame.number >= {}'.format(self.checkpoint_frame_number)]
         tshark_result = subprocess.run(tshark_command, stdout=subprocess.PIPE,
                                        stderr=subprocess.PIPE)
         if tshark_result.returncode != 0:
@@ -139,6 +142,16 @@ class Sniffer:
         except json.JSONDecodeError as error:
             err("capture JSON decoding failed: %s".format(error))
             return []
+
+    def checkpoint(self):
+        '''Checkpoint the capture.
+
+        Any subsequent calls to get_packet_capture will only return packets capture after now.
+        '''
+        capture = self.get_packet_capture()
+        if capture:
+            self.checkpoint_frame_number = capture[-1].frame_number + 1
+        # else keep last checkpoint
 
     def stop(self):
         '''Stop tcpdump if it is running.'''
--- a/tests/sniffer.py
+++ b/tests/sniffer.py
@@ -13,32 +13,30 @@ from opts import debug, err, status
 
 class Sniffer:
     '''Captures packets on an interface.'''
-    def __init__(self, interface: str, use_tcpdump: bool, tcpdump_log_dir: str):
+    def __init__(self, interface: str, tcpdump_log_dir: str):
         self.interface = interface
-        self.use_tcpdump = use_tcpdump
         self.tcpdump_log_dir = tcpdump_log_dir
         self.tcpdump_proc = None
 
     def start(self, outputfile_basename):
-        '''Start tcpdump if enabled by config.'''
-        if self.use_tcpdump:
-            debug("Starting tcpdump, output file {}.pcap".format(outputfile_basename))
-            os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
-            outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
-            command = ["tcpdump", "-i", self.interface, "-w", outputfile]
-            self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
-            # tcpdump takes a while to start up. Wait for the appropriate output before continuing.
-            # poll() so we exit the loop if tcpdump terminates for any reason.
-            while not self.tcpdump_proc.poll():
-                line = self.tcpdump_proc.stderr.readline()
-                debug(line.decode()[:-1])  # strip off newline
-                if line.startswith(b"tcpdump: listening on " + self.interface.encode()):
-                    # Make sure it doesn't block due to stderr buffering
-                    self.tcpdump_proc.stderr.close()
-                    break
-            else:
-                err("tcpdump terminated")
-                self.tcpdump_proc = None
+        '''Start tcpdump to outputfile.'''
+        debug("Starting tcpdump, output file {}.pcap".format(outputfile_basename))
+        os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
+        outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
+        command = ["tcpdump", "-i", self.interface, "-w", outputfile]
+        self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
+        # tcpdump takes a while to start up. Wait for the appropriate output before continuing.
+        # poll() so we exit the loop if tcpdump terminates for any reason.
+        while not self.tcpdump_proc.poll():
+            line = self.tcpdump_proc.stderr.readline()
+            debug(line.decode()[:-1])  # strip off newline
+            if line.startswith(b"tcpdump: listening on " + self.interface.encode()):
+                # Make sure it doesn't block due to stderr buffering
+                self.tcpdump_proc.stderr.close()
+                break
+        else:
+            err("tcpdump terminated")
+            self.tcpdump_proc = None
 
     def stop(self):
         '''Stop tcpdump if it is running.'''
--- a/tests/sniffer.py
+++ b/tests/sniffer.py
@@ -23,7 +23,7 @@ class Sniffer:
         debug("Starting tcpdump, output file {}.pcap".format(outputfile_basename))
         os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
         outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
-        command = ["tcpdump", "-i", self.interface, "-w", outputfile]
+        command = ["tcpdump", "-i", self.interface, '-U', '-w', outputfile]
         self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
         # tcpdump takes a while to start up. Wait for the appropriate output before continuing.
         # poll() so we exit the loop if tcpdump terminates for any reason.
--- a/tests/sniffer.py
+++ b/tests/sniffer.py
@@ -6,8 +6,8 @@
 ###############################################################
 
 import os
+import json
 import subprocess
-
 from opts import debug, err, status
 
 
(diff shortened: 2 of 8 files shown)


Galvien:
@arnout Gitlab CI will always fail since this change completely removes the `--tcpdump` option.
I think we'll need to change that.

tomereli:
> @arnout Gitlab CI will always fail since this change completely removes the `--tcpdump` option.
> I think we'll need to change that.

So change what you need in .gitlab-ci.yml

tests/opts.py
```diff
@@ -9,7 +9,6 @@
 class opts:
     '''Static class that encodes the global options.'''
     verbose = False
```
tomereli, line 11: If you remove this section, you need to also update the README to mention that tcpdump is now a requirement for running tests even in docker. You should also explain how to fix the capabilities so that non sudo users can use tcpdump.

tests/sniffer.py
```diff
@@ -13,30 +13,33 @@
...
-                    # Make sure it doesn't block due to stderr buffering
-                    self.tcpdump_proc.stderr.close()
-                    break
-            else:
-                err("tcpdump terminated")
-                self.tcpdump_proc = None
+    def start(self, test_name):
+        '''Start tcpdump.'''
+        outputfile_basename = "test_" + test_name
+        debug("Starting tcpdump, output file {}.pcap".format(outputfile_basename))
```
tomereli, line 106: I don't think you should make the output file hard coded - why is this needed? You can just add another argument with default value (None) for the output file name - so that if it is None, you will use the test name.

tests/sniffer.py
```diff
@@ -26,7 +26,7 @@ def start(self, test_name):
         os.makedirs(os.path.join(opts.tcpdump_dir, 'logs'), exist_ok=True)
         outputfile = os.path.join(opts.tcpdump_dir, outputfile_basename) + ".pcap"
         self.outputfiles[test_name] = outputfile
-        command = ["tcpdump", "-i", self.interface, '-w', outputfile]
+        command = ["tcpdump", "-i", self.interface,'-U', '-w', outputfile]
```
tomereli: I don't understand whNot sure I understand that the `-U` flag is used for here:

From tcpdump manual:

>-U
--packet-buffered
If the -w option is not specified, or if it is specified but the --print flag is also specified, make the printed packet output ``packet-buffered''; i.e., as the description of the contents of each packet is printed, it will be written to the standard output, rather than, when not writing to a terminal, being written only when the output buffer fills.
If the -w option is specified, make the saved raw packet output ``packet-buffered''; i.e., as each packet is saved, it will be written to the output file, rather than being written only when the output buffer fills.
The -U flag will not be supported if tcpdump was built with an older version of libpcap that lacks the pcap_dump_flush(3PCAP) function.

Your commit message also doesn't explain what issue this flag is solving. Please elaborate.

tests/sniffer.py
```diff
@@ -41,6 +42,23 @@ def start(self, test_name):
...
 
+    def get_packet_capture(self, test_name):
+        if test_name in self.captures:
+            return self.captures[test_name]
+        tshark_command = ['tshark', '-r', self.outputfiles[test_name], '-T', 'json']
+        self.tshark_proc = subprocess.run(
+            tshark_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        try:
+            capture = json.loads(self.tshark_proc.stdout)
+            def ieee_filter(x): return 'ieee1905' in x['_source']['layers']
```
tomereli: I don't think this is appropriate - instead, we should capture only ieee1905 packets from the start when we use tcpdump:
```
tcpdump -i<iface> ether proto 0x893a
```

tests/sniffer.py
```diff
@@ -41,6 +42,23 @@ def start(self, test_name):
...
+            result = self._simplify_packets(filter(ieee_filter, capture))
+            if not result or len(result) == 0:
+                result = None
+            self.captures[test_name] = result
+            return self.captures[test_name]
+        except BaseException as ex:
+            return None
```
tomereli: Can you explain what kind of exception we're ignoring here?

tests/sniffer.py
```diff
@@ -59,6 +59,20 @@ def ieee_filter(x): return 'ieee1905' in x['_source']['layers']
         except BaseException as ex:
             return None
 
+    def _simplify_packets(self, packets):
```
tomereli: This should be reordered since you use this method in the previous commit.

tests/test_flows.py
```diff
@@ -365,10 +365,12 @@ def test_ap_capability_query(self):
 
         debug("Confirming ap capability query has been received on agent")
         self.check_log(env.agents[0], "AP_CAPABILITY_QUERY_MESSAGE")
-
+        capability_query = env.wired_sniffer.get_packets_for_cmdu_type(self.running,0x8001)
```
tomereli: The example should replace the check_log, not just add a call without testing it. Also I think you should add an API to search for specific packets like it is done in the UCC.

For example, this is how the UCC is checking for a specific topology response:

```
2020-04-14 04:08:41.506 - INFO - 
        ~~~~~ 4.1.1 Step 9 sniffer checks - Topology Response message ~~~~~

2020-04-14 04:08:41.512 - INFO - SNIFFER (192.168.250.5:9999) ---> sniffer_control_filter_capture,InFile,_MAP-4.1.1-STEP9,OutFile,MAP-4.1.1-STEP9,msgId,0x43c6,filterfile,mapfile,msgType,0x0003,nframes,1
2020-04-14 04:08:42.023 - INFO - SNIFFER (192.168.250.5:9999) <-- status,COMPLETE,FilterStatus,SUCCESS,FrameCount,1

```

I think the equivalent in python should be something like:

```
topology_response = env.wired_sniffer.get_packets(self.running, cmdu_type=0x0003, mid=0x43c6)
```
The API should also allow for more arguments - source and destination MACs so we can filter on ALid, and probably more that I can't think of now. Every parameter should be optional except for self.running.

tests/sniffer.py
```diff
@@ -73,6 +74,23 @@ def _simplify(x):
             return x
         return list(map(_simplify, packets))
 
+    def get_packets_for_cmdu_type(self, test_name, msg_type: int):
```
tomereli: I think we're missing an important argument here - message identifier (mid) - this one should be used to verify that we are looking at the correct cmdu. This should be optional parameter, so if its none, it should return all the CMDUs which match the message type, otherwise it would return a single CMDU or None if there is no message with this MID.

In addition, we must also add source and destination macs parameters, since this is the only way we can uniquely identify a message - source + destination ALid + mid.

tomereli (changes requested):
Very nice PR, good commits split. I have some comments so re-review is required.

tests/sniffer.py
```diff
@@ -7,36 +7,89 @@
...
+        return list(map(_simplify, packets))
+
+    def get_packets_for_cmdu_type(self, test_name, msg_type: int):
+        def filter_func(x): return x['msg_type'] == msg_type
+        try:
+            return list(filter(filter_func, self.get_packet_capture(test_name)))
+        except:
+            err("No packets captured for specified cmdu type: {}".format(msg_type))
+            return None
+        return list(filter(filter_func, self.get_packet_capture(test_name)))
```
ghost: I think you must not be familiar with list comprehensions in python. They're really useful. 
```python
def get_packets_for_cmdu_type(self, test_name :str, msg_type: int):
    return [x for x in self.get_packet_capture(test_name) if x['msg_type'] == msg_type]
```

tests/sniffer.py
```diff
@@ -7,36 +7,89 @@
...
-            else:
-                err("tcpdump terminated")
-                self.tcpdump_proc = None
+    def filter_packets_by_time(self, packets, start_time=None, end_time=time()):
+        def _get_time_epoch_for_packet(packet):
+            t = packet['time_epoch']
+            if start_time:
+                return t >= start_time and t <= end_time
+            return t <= end_time
+        return list(filter(_get_time_epoch_for_packet, packets))
```
ghost: Same here, I would do this with a list comprehension
```python
def filter_packets_by_time(self, packets, start_time=None, end_time=None):
    if not end_time:
        end_time = time()
    return [x for x in packets if x['time_epoch'] > start_time and x['time_epoch'] < end_time]
```
This also illustrates an important python pitfall: never use function calls for default arguments:
```
>>> def test_time(t=time()):
...     print(t)
...
>>> time()
1587480434.0399244
>>> test_time()
1587480426.4636228
>>> time()
1587480438.1276047
```
See how it appears we've gone back in time? That's because the default argument is evaluated when the python interpreter first sees the function, NOT when it's called!

tests/test_flows.py
```diff
@@ -366,10 +365,12 @@ def test_ap_capability_query(self):
...
+        
         debug("Confirming ap capability report has been received on controller")
         self.check_log(env.controller, "AP_CAPABILITY_REPORT_MESSAGE")
-
+        capability_report = env.wired_sniffer.get_packets_for_cmdu_type(self.running,0x8002)
```
ghost: Please remember to run `flake8` on your python scripts and fixing issues it notices. For example, here it would complain about not having a space after the comma.

tests/sniffer.py
arnout: When the outputfiles dict is removed, this change is not necessary.

tests/sniffer.py
arnout: Without -U, the output file will be buffered. That means we'd have to terminate tcpdump to see the last few packets in the output file. with the -U option, it is packet-buffered, i.e. whenever a packet is captured, the file is flushed. That way, when we read the file, we're sure that it contains everything that has been captured.

tests/sniffer.py
```diff
@@ -7,36 +7,89 @@
...
+from time import time
 
 
 class Sniffer:
     '''Captures packets on an interface.'''
+
     def __init__(self, interface: str):
         self.interface = interface
         self.tcpdump_proc = None
+        self.outputfiles = {}
```
arnout: There is no need for this dictionary. At any point, there is only one running capture, thus only one running outputfile. All the get_packet_capture() calls will be for that last outputfile.

Note that multiple captures in parallel is not possible because they're all capturing the same packets.

tests/sniffer.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, interface: str):
         self.interface = interface
         self.tcpdump_proc = None
         self.outputfiles = {}
+        self.captures = {}
```
arnout: Again, not needed: just use the latest one.

tests/sniffer.py
```diff
@@ -59,6 +59,20 @@ def ieee_filter(x): return 'ieee1905' in x['_source']['layers']
         except BaseException as ex:
             return None
 
+    def _simplify_packets(self, packets):
+        def _simplify(x):
+            x['src_mac'] = x['_source']['layers']['eth']['eth.src']
```
arnout: (nitpick) Instead of "simplifying", it would be nicer to convert to a brand new class Ieee1905Message. Or a namedtuple instead o a full class.

tests/sniffer.py
arnout: Yeah, there are multiple ways to go about it. Basically we want to check "was there a packet that matches these and these parameters".

I think actually the best way to do it would be to have a check_packet function in TestFlows itself that takes the filter as an argument. The filter can then check all the data it needs to.

arnout (changes requested):

arnout (approved):
I took over this PR and fixed all the issues that were reported in this one and in #1020.

Now it is perfect :-)

tomereli (approved):

tests/sniffer.py
```diff
@@ -23,7 +23,7 @@ def start(self, outputfile_basename):
         debug("Starting tcpdump, output file {}.pcap".format(outputfile_basename))
         os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
         outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
-        command = ["tcpdump", "-i", self.interface, "-w", outputfile]
+        command = ["tcpdump", "-i", self.interface, '-U', '-w', outputfile]
```
rmelotte: Maybe it would be worth adding the `-s` option with the maximum size of a wireless frame (2346).
The default size depends on the tcpdump version, but on my machine for example it's set to 262144, while we know the maximum size we'll ever capture can't exceed 2346 (and probably even 1516 in practice).

AFAIU tcpdump sets it so high for backward compatibility, but they recommend to set it to something smaller (from man tcpdump):

> You should limit snaplen to the smallest number that will  capture the protocol information you're interested in.

(it's a nitpick comment of course, we shouldn't have that much traffic for now, so we can still do that once we start sniffing wireless frames)

rmelotte (approved):
I didn't review the details, overall LGTM.

I only have a nitpick comment, but it will probably only become relevant once we start sniffing wireless interfaces, so I'll add the "ready for merge" tag anyway.

tests/sniffer.py
arnout: I believe snaplen only has an impact on memory usage, not so much on performance. So I'm not really worried about it. That said, it's a good suggestion.

BTW I intend to also drop tcpdump and replace it with tshark at some point - it's kind of stupid to require both to be installed.

PR #1146, opened 2020-04-14
Feature/docker use generated ip

### LINKED ISSUE #461: [BUG] tools: docker: the docker inspect returns a wrong IP

### LINKED ISSUE #927: [BUG] run-tests gets the same IP for 2 different devices

### LINKED ISSUE #1113: [TASK] docker: drop the hard-coded MAC addresses
## Description

The MAC addresses used in docker are currently hard-coded.

To increase test coverage and also because we don't need the MAC addresses to be hard-coded anymore (they are discovered in `test_flows.py`), we could just use random ones.

## Exit criteria

The MAC addresses are not hard-coded anymore in:
- `common/beerocks/scripts/prplmesh_utils.sh.in`
- `tools/docker/runner/entrypoint.sh`

rmelotte:


tests/test_gw_repeater.sh
```diff
@@ -55,8 +55,8 @@ check_wsl() {
     # Docker fails to route traffic from the host to the containers
     # if the IP is anything but what was allocated by the daemon
     # when the container was created
```
arnout: Is this comment still valid?

tools/docker/run.sh
```diff
@@ -61,7 +47,7 @@ main() {
             -h | --help)        usage; exit 0; shift ;;
             -d | --detach)      DETACH=true; shift ;;
             -f | --force)       FORCE=true; shift ;;
-            -i | --ipaddr)      IPADDR="$2"; shift; shift ;;
+            -i | --ipaddr)      IPADDR="--ip=$2"; shift; shift ;;
```
arnout: I don't believe anyone uses this option anyway, so let's just remove it.

tests/test_gw_repeater.sh
```diff
@@ -111,12 +111,10 @@ main() {
     }
 
     [ "$START_REPEATER" = "true" ] && {
-        index=0
         for repeater in $REPEATER_NAMES; do
             status "Start Repeater (Remote Agent): $repeater"
             "${rootdir}"/tools/docker/run.sh -u "${UNIQUE_ID}" ${VERBOSE_OPT} ${FORCE_OPT} ${RP_EXTRA_OPT} \
-                start-agent -d -n "${repeater}" -m aa:bb:cc:"$index$index" -- "$@"
-            index=$((index+1))
+            start-agent -d "$DELAY" -n "${repeater}" -- "$@"
```
arnout: Something funky going on here... The `-d` option is for detaching, not for DELAY. And also you removed the indentation.

Same for all other changes in this commit.

tools/docker/image-pull.sh
```diff
@@ -59,12 +59,12 @@ main() {
     dbg "TAG=$TAG"
     dbg "rootdir=$rootdir"
 
-    # The tag used in dockerhub is the base image with : converted to -
-    upstream_tag="$(echo "$IMAGE" | tr : -)"
-    info "Base docker image $IMAGE -> tag $tag"
+    # The registry used in Gitlab has the base image name appended with : converted to -
```
arnout, line 62: NACK this. At least for me, I didn't intend to drop dockerhub and switch to the gitlab registry. Instead, what I meant was to:
- dockerbuild images as part of CI;
- push them to the registry in each pipeline;
- push them to dockerhub on master.

So the image-pull.sh script wouldn't be affected. Instead, CI should not use image-pull.sh but pull from the registry "manually". Or, better yet, not pull anything explicitly and rely on automatic pulling, by passing the image name explicitly to run.sh.

But that's just my opinion, maybe dropping dockerhub is simpler and good enough.

arnout (approved):
Looks good to me so far. Is there any reason why this is still draft?

tools/docker/image-pull.sh
rmelotte: > Instead, what I meant was to:

    dockerbuild images as part of CI;
    push them to the registry in each pipeline;
    push them to dockerhub on master.

This is almost what I had in mind, only that last step would have been: tag the image with "latest" on Gitlab if we're on the master branch instead of pushing to dockerhub.
It doesn't really make a difference to me though, the only thing we need to push to dockerhub on master is a robot account/token if we don't already have one, but it shouldn't be a big deal.

One change that I propose to keep though, is to tag the images like `prplmesh-runner-ubuntu-18.04` instead of `prplmesh-runner:ubuntu-18
04`. 
On Gitlab since we're going to have hundreds of different tags (one for each pipeline) I think it makes things easier if the different variants are not mixed together under a single registry name.

tools/docker/run.sh
rmelotte: Ok, done.

tests/test_gw_repeater.sh
rmelotte: Oops, yes that's a mistake, thanks.

tests/test_gw_repeater.sh
rmelotte: To me it looks like it's no longer relevant, I removed it.

tools/docker/image-pull.sh
arnout: As discussed orally: we agree.

abelog (approved):

Issue #394, opened 2019-10-09
4.7.4,5,6 Per-AP Metrics Response


arnout:
The prplMesh agent on Turris Omnia with OpenWrt needs to pass the certification test case 4.7.4, 4.7.5 and 4.7.6 as defined by the Wi-Fi CERTIFIED EasyMesh (tm) Test Plan.

Since all these tests are very similar, they are collected into a single milestone entry.

- [x] #392 AP Metrics and STA Link Metrics and Traffic Stats for nl80211
- [x] #393 Agent handle Steering and Metric Reporting policy
- [x] #384 Pass flow 4.7.4 test in testbed
- [x] #385 Pass flow 4.7.5 test in testbed
- [x] #386 Pass flow 4.7.6 test in testbed

ghost:
Re-opening because 4.7.4-4.7.6 is not done yet, just the DR is

PR #1149, opened 2020-04-15
slave: monitor: add Associated STA Link Metrics Response

vitalii-komisarenko:
At the moment Associated STA Link Metrics are handled as
a part of *_HOSTAP_STATS_MEASUREMENT_* messages.
We would like to handle them separately.
    
This commit adds a flow for the Associated STA Link Metrics Response
that basically duplicates a subset of *_HOSTAP_STATS_MEASUREMENT_*.
It is a preparation for removing above-mentioned subset from Hostap Measurements.


vitalii-komisarenko:
Replaced by  PR #1149

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -179,11 +181,50 @@ void monitor_stats::process()
             sta_stats_msg.rx_load_percent   = sta_stats.rx_load_percent_curr;
             sta_stats_msg.stats_delta_ms    = sta_stats.delta_ms;
             sta_stats_msg.rx_rssi           = sta_stats.rx_rssi_curr;
+
+            beerocks_message::sBssidInfo bss_info;
+            bss_info.bssid                      = network_utils::mac_from_string(sta_mac);
```
vitalybu: `bssid` should be the MAC address of the AP not the STA.
Unless the field name is incorrect

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2088,6 +2095,56 @@ bool backhaul_manager::get_sta_bssid(const sMacAddr &sta_mac, sMacAddr &bssid)
     return false;
 }
 
+bool backhaul_manager::handle_associated_sta_link_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                                const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received ASSOCIATED_STA_LINK_METRICS_QUERY_MESSAGE , mid=" << std::dec
```
vitalybu, line 2130:
```suggestion
    LOG(DEBUG) << "Received ASSOCIATED_STA_LINK_METRICS_QUERY_MESSAGE, mid=" << std::dec
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2088,6 +2095,56 @@ bool backhaul_manager::get_sta_bssid(const sMacAddr &sta_mac, sMacAddr &bssid)
...
+        error_code_tlv->sta_mac() = mac->sta_mac();
+
+        LOG(DEBUG) << "Send a ASSOCIATED_STA_LINK_METRICS_RESPONSE_MESSAGE back to controller";
+        return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
+    } else {
+        LOG(DEBUG) << "client with mac address " << mac->sta_mac() << " connected to " << bssid;
+        //TODO
+        return false;
+    }
+}
```
```suggestion
    }
    
    LOG(DEBUG) << "client with mac address " << mac->sta_mac() << " connected to " << bssid;
    //TODO
    return false;
}
```
`//TODO` what?

Also, if you already found the `bssid` to which the STA is connected, it's better to forward the CMDU the correct slave from here. Otherwise (by returning `false`) the message will be forwarded to all the slaves.

tests/test_flows.py
```diff
@@ -488,6 +488,18 @@ def test_client_association_dummy(self):
         self.check_log(env.agents[0].radios[0],
                        r"Got client disallow request for {}".format(sta.mac))
 
+    def test_client_association_link_metrics(self):
```
vitalybu, line 548: Please this change to a separate commit

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2080,13 +2080,17 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     }
 }
 
-bool backhaul_manager::get_sta_bssid(const sMacAddr &sta_mac, sMacAddr &bssid)
+bool backhaul_manager::get_sta_bssid_and_socket(const sMacAddr &sta_mac, sMacAddr &bssid,
+                                                Socket **socket)
```
vitalybu: It better to use a reference to a pointer (https://stackoverflow.com/a/3834103)
```suggestion
                                                Socket *&socket)
```

But why do you need to return both the `bssid` and the slave socket?
You can return the `sRadioInfo` to which the STA is connected and then extract both the `bssid` and the slave socket from there.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2080,13 +2080,17 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
...
 {
     for (const auto &slave : slaves_sockets) {
         auto associated_clients_map = slave->associated_clients_map;
         for (const auto &vap : associated_clients_map) {
             if (vap.second.find(sta_mac) != vap.second.end()) {
                 bssid = vap.first;
+                if (socket) {
+                    *socket = slave->slave;
+                }
```
vitalybu: Assuming a reference to a pointer:
```suggestion
                socket = slave->slave;
```

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1411,6 +1411,29 @@ bool slave_thread::handle_cmdu_backhaul_manager_message(
 
         break;
     }
+    case beerocks_message::ACTION_BACKHAUL_HOSTAP_STATS_MEASUREMENT_REQUEST: {
+        LOG(DEBUG) << "ACTION_BACKHAUL_HOSTAP_STATS_MEASUREMENT_REQUEST";
+        if (monitor_socket) {
```
vitalybu: According to our coding standards:
```suggestion
        if (!monitor_socket) {
            break;
        }
        ...        
```

tests/test_flows.py
```diff
@@ -499,6 +499,19 @@ def test_client_association_link_metrics(self):
                                      tlv(0x95, 0x0006, '{sta_mac}'.format(sta_mac=sta_mac)))
         self.check_log(env.agents[0],
                        "client with mac address {sta_mac} not found".format(sta_mac=sta_mac))
+        time.sleep(1)
```
vitalybu, line 559: Again, tests should be updated in a separate commit

vitalybu (dismissed):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2073,6 +2073,25 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     }
 }
 
+bool backhaul_manager::get_sta_bssid(
+    const std::unordered_map<sMacAddr, associated_clients_t> &clients_map, const sMacAddr &sta_mac,
+    sMacAddr &bssid)
+{
+    return std::any_of(clients_map.begin(), clients_map.end(),
```
morantr: Can't it be done with ranged for-loops? They are more readable and have less indentation.
It is mention as preferred on https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md#loops.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1411,6 +1411,29 @@ bool slave_thread::handle_cmdu_backhaul_manager_message(
...
+            auto request_in = beerocks_header->addClass<
+                beerocks_message::cACTION_BACKHAUL_HOSTAP_STATS_MEASUREMENT_REQUEST>();
+            if (request_in == nullptr) {
```
morantr: With a smart pointer, prefer doing `if (!smart_ptr)` over `if (smart_ptr == nullptr)`

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1411,6 +1411,29 @@ bool slave_thread::handle_cmdu_backhaul_manager_message(
...
+
+            auto request_out = message_com::create_vs_message<
+                beerocks_message::cACTION_MONITOR_HOSTAP_STATS_MEASUREMENT_REQUEST>(
+                cmdu_tx, beerocks_header->id());
+            if (request_out == nullptr) {
+                LOG(ERROR) << "Failed building message!";
+                return false;
+            }
+            request_out->sync() = request_in->sync();
+            LOG(DEBUG) << "Will now send ACTION_MONITOR_HOSTAP_STATS_MEASUREMENT_REQUEST";
```
morantr: the "will now" is redundant.
`LOG(DEBUG) << "send ACTION_MONITOR_HOSTAP_STATS_MEASUREMENT_REQUEST";`

morantr (approved):
Nothing major so approving.
I really don't like the use of std::any instead of ranged loops.
also, lasts commit "cppcheck: existing issues: add unfixable issues for associated sta li……nk metrics" headline is too long.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -179,11 +181,53 @@ void monitor_stats::process()
...
+            bss_info.earliest_measurement_delta = sta_stats.delta_ms;
+            // TODO: MAC data rate and Phy rate are not necessarily the same
```
tomereli: TODO should be accompanied with a github issue URL

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -179,11 +181,53 @@ void monitor_stats::process()
...
         beerocks_header->actionhdr()->id() = requests_list.front();
         requests_list.pop_front();
         message_com::send_cmdu(slave_socket, cmdu_tx);
+
+        LOG(DEBUG) << "Preparing ACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE. "
+                   << "Number of entries: " << bss_infos.size();
+
+        auto sta_metrics = message_com::create_vs_message<
+            beerocks_message::cACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE>(cmdu_tx);
+        if (sta_metrics == nullptr) {
```
tomereli:
```suggestion
        if (!sta_metrics) {
```

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -179,11 +181,53 @@ void monitor_stats::process()
...
+            LOG(ERROR) << "Failed building "
+                          "cACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE message!";
+            return;
+        }
+
+        if (!sta_metrics->alloc_bssid_info_list(bss_infos.size())) {
+            LOG(ERROR) << "Failed allocate_bssid_info_list";
```
tomereli:
```suggestion
            LOG(ERROR) << "Failed allocate_bssid_info_list";
            return;
```

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -179,11 +181,53 @@ void monitor_stats::process()
...
+        }
+
+        // TODO: set sta_metrics->sta_mac
```
tomereli: TODO is both unclear and missing link to a github issue.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: You are right Moran, but `cppcheck` thinks otherwise :/

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2073,6 +2073,25 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     }
 }
 
+bool backhaul_manager::get_sta_bssid(
```
tomereli: Another option is to return sMacAddr instead of having it as a parameter - we can return an `network_utils::ZERO_MAC` if not found (nitpick).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2091,11 +2110,10 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx
     bool associated_client_found = false;
     for (const auto &slave : slaves_sockets) {
         auto associated_clients_map = slave->associated_clients_map;
-        for (const auto &vap : associated_clients_map) {
-            if (vap.second.find(client_info_tlv_r->client_mac()) != vap.second.end()) {
-                associated_client_found = true;
-                break;
-            }
+        sMacAddr bssid;
```
tomereli: So this will change to:

```
auto bssid = get_sta_bssid(associated_clients_map, client_info_tlv_r->client_mac());
if (bssid != network_utils::ZERO_MAC) {
  associated_client_found = true;
  break;
}
```
(nitpick continued :) )

tomereli (commented):
Not finished.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
vitalii-komisarenko: shall `ZERO_MAC` convert implicitly to `false` in boolean context? :)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: >> Can't it be done with ranged for-loops? They are more readable and have less indentation.

That they are more readable is a matter of taste.

>> It is mention as preferred on https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md#loops.

That is only compared to "normal" for-loops, not compared to std::algorithms.


> You are right Moran, but `cppcheck` thinks otherwise :/

That's not good enough of a reason. If a tool doesn't do what we want, we have to change the tool, not what we want.

That said, I guess the main argument against using any_of is that it's too verbose? There are actually two ways that we can reduce the verbosity: by using C++14 auto, and by defining wrappers for the algorithms that take a container rather than an iterator.

```
    return any_of(clients_map, [&](const auto &vap) {
        return any_of(vap.second, [&](const auto &mac) {
            if (mac.first == sta_mac) {
                bssid = vap.first;
                return true;
            }
            return false;
        });
    });
```
where
```
template <class T, class UnaryPredicate>
bool any_of(const T& container, UnaryPredicate p)
{
    return any_of(std::begin(container), std::end(container), p);
}
```

Without that helper function, it's still not that bad if you use auto:

```
    return std::any_of(clients_map.begin(), clients_map.end(), [&](const auto &vap) {
        return std::any_of(vap.second.begin(), vap.second.end(), [&](const auto &mac) {
            if (mac.first == sta_mac) {
                bssid = vap.first;
                return true;
            }
            return false;
        });
    });
```

but I agree that it becomes bad when the lambda declaration no longer fits on the same line.

So I think the indentation argument is not really valid. Thus, the question is: which of the two is more readable?

I think we'll need to take a vote on #democracy on this.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -976,3 +976,16 @@ sChannelScanResults:
   channel_utilization:
     _comment: Indicates the fraction of the time AP senses that the channel is in use by the neighboring AP for transmissions.
     _type: uint32_t
+
+sBssidInfo:
```
arnout, line 974: (nitpick) I don't think the name is very good. It's not info about a BSS, it's the STA link metrics for one particular BSS... Maybe `sSingleSTALinkMetric`?

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_monitor.yaml
```diff
@@ -90,6 +90,18 @@ cACTION_MONITOR_HOSTAP_ACTIVITY_NOTIFICATION:
   _type: class
   params: sApActivityNotificationParams 
 
+cACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE:
```
arnout, line 98: In #1198 we decided to use the IEEE1905.1 message type directly rather than defining a new TLV that copies everything from the EasyMesh message. You could do the same here...

That said, it's probably a lot of work to rework this big PR to do that now. So I'm not going to say you have to do it.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -220,11 +222,65 @@ void monitor_stats::process()
...
+        LOG(DEBUG) << "Preparing ACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE. "
+                   << "Number of entries: " << bss_infos.size();
+
+        std::for_each(
+            bss_infos.begin(), bss_infos.end(),
+            [&](std::pair<sMacAddr, std::vector<beerocks_message::sBssidInfo>> bss_info) {
+                auto sta_metrics = message_com::create_vs_message<
+                    beerocks_message::cACTION_MONITOR_CLIENT_ASSOCIATED_STA_LINK_METRIC_RESPONSE>(
+                    cmdu_tx, message_id);
+                if (sta_metrics == nullptr) {
```
arnout:
```suggestion
                if (!sta_metrics) {
```
https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md#klocwork

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2146,15 +2156,7 @@ bool backhaul_manager::handle_client_capability_query(ieee1905_1::CmduMessageRx
...
-    bool associated_client_found = false;
-    for (const auto &slave : slaves_sockets) {
-        auto associated_clients_map = slave->associated_clients_map;
-        if (get_sta_bssid(associated_clients_map, client_info_tlv_r->client_mac()) !=
-                network_utils::ZERO_MAC) {
-            associated_client_found = true;
-            break;
-        }
-    }
+    bool associated_client_found = (bool)get_sta_radio(client_info_tlv_r->client_mac());
```
arnout: Please no C-style casts. I believe this works:
```suggestion
    bool associated_client_found { get_sta_radio(client_info_tlv_r->client_mac()) };
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2141,6 +2148,61 @@ backhaul_manager::get_sta_radio(const sMacAddr &sta_mac)
     return radio != slaves_sockets.end() ? *radio : nullptr;
 }
 
+bool backhaul_manager::handle_associated_sta_link_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                                const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received ASSOCIATED_STA_LINK_METRICS_QUERY_MESSAGE , mid=" << std::dec
```
arnout, line 2130: I believe the std::dec shouldn't be needed any more because we reset the formatters in the LOG destructor, no?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2141,6 +2148,61 @@ backhaul_manager::get_sta_radio(const sMacAddr &sta_mac)
...
+               << int(mid);
+
+    if (!cmdu_tx.create(mid,
+                        ieee1905_1::eMessageType::ASSOCIATED_STA_LINK_METRICS_RESPONSE_MESSAGE)) {
+        LOG(ERROR)
+            << "cmdu creation of type ASSOCIATED_STA_LINK_METRICS_RESPONSE_MESSAGE, has failed";
+        return false;
+    }
+
+    auto mac = cmdu_rx.getClass<wfa_map::tlvStaMacAddressType>();
```
arnout, line 2140: (nitpick) mac can be anything, sta_mac would be better.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2141,6 +2148,61 @@ backhaul_manager::get_sta_radio(const sMacAddr &sta_mac)
...
+    if (!mac) {
```
arnout, line 2141: (nitpick) do all checks before the cmdu_tx::create.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2141,6 +2148,61 @@ backhaul_manager::get_sta_radio(const sMacAddr &sta_mac)
...
+    if (!assoc_link_metrics) {
+        LOG(ERROR) << "Failed to create tlvAssociatedStaLinkMetrics tlv";
+        return false;
+    }
+
+    // Check if it is an error scenario - if the STA specified in the STA link Query message is not associated
+    // with any of the BSS operated by the Multi-AP Agent
+    std::shared_ptr<sRadioInfo> radio = get_sta_radio(mac->sta_mac());
+    if (!radio) {
+        LOG(ERROR) << "client with mac address " << mac->sta_mac() << " not found";
```
arnout, line 2156: This is not really an error, rather info or warning. It is expected to happen, occasionally, because the controller may send the request just as the station is disassociating.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2141,6 +2148,61 @@ backhaul_manager::get_sta_radio(const sMacAddr &sta_mac)
...
+        return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
+    } else {
+        sMacAddr bssid = get_sta_bssid(radio->associated_clients_map, mac->sta_mac());
+        if (bssid == network_utils::ZERO_MAC) {
+            LOG(ERROR) << "Cannot find sta bssid";
+            return false;
+        }
+        LOG(DEBUG) << "client with mac address " << mac->sta_mac() << " connected to " << bssid;
+        //TODO: non-error flow for Associated STA Link Metrics Request
+        return false;
```
arnout: (nitpick) since you already created the response message, I'd still send it here but without the required TLV. But since you fix that later in the same PR it's not important.

tests/test_flows.py
```diff
@@ -531,6 +531,18 @@ def test_client_association_dummy(self):
         self.check_log(env.agents[0].radios[0],
                        r"Got client disallow request for {}".format(sta.mac))
 
+    def test_client_association_link_metrics(self):
+        ''' This test verifies that a MAUT with an associated STA responds to
+        an Associated STA Link Metrics Query message with an Associated STA Link Metrics
+        Response message containing an Associated STA Link Metrics TLV for the associated STA.'''
+
+        sta_mac = "11:11:33:44:55:66"
```
arnout, line 553:
```suggestion
        sta = env.Station.create()
```

tests/test_flows.py
```diff
@@ -531,6 +531,18 @@ def test_client_association_dummy(self):
...
+        debug("Send link metrics query for unconnected STA")
+        env.controller.dev_send_1905(env.agents[0].mac, 0x800D,
+                                     tlv(0x95, 0x0006, '{sta_mac}'.format(sta_mac=sta_mac)))
```
arnout, line 556:
```suggestion
                                     tlv(0x95, 0x0006, '{}'.format(sta.mac)))
```
The `{var}` form is very useful in f-strings, but without them it's just too verbose IMO.

tests/test_flows.py
```diff
@@ -531,6 +531,18 @@ def test_client_association_dummy(self):
...
+        self.check_log(env.agents[0],
```
arnout, line 557: We should actually check the error TLV in the reply, but that requires #1142 of course...

tests/test_flows.py
```diff
@@ -542,6 +542,19 @@ def test_client_association_link_metrics(self):
                                      tlv(0x95, 0x0006, '{sta_mac}'.format(sta_mac=sta_mac)))
         self.check_log(env.agents[0],
                        "client with mac address {sta_mac} not found".format(sta_mac=sta_mac))
+        time.sleep(1)
+
+        sta = env.Station.create()
```
arnout, line 561: To distinguish from the previous sta, use sta1 and sta2.

ci/cppcheck/cppcheck_existing_issues.txt
```diff
@@ -678,3 +678,7 @@
 '/builds/prpl-foundation/prplMesh/controller/src/beerocks/master/son_management.cpp: style: Variable 'op_error_code' is assigned a value that is never used. [unreadVariable]        auto op_error_code = eChannelScanOpErrCode::CHANNEL_SCAN_OP_SUCCESS;
 '/builds/prpl-foundation/prplMesh/controller/src/beerocks/master/son_management.cpp: style: Variable 'op_error_code' is assigned a value that is never used. [unreadVariable]            op_error_code = eChannelScanOpErrCode::CHANNEL_SCAN_OP_SCAN_IN_PROGRESS;
 '/builds/prpl-foundation/prplMesh/controller/src/beerocks/master/son_management.cpp: style: Variable 'result_status' is assigned a value that is never used. [unreadVariable]            result_status = last_scan_success;
+'/builds/prpl-foundation/prplMesh/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h: warning: Virtual function 'on_thread_stop' is called from destructor '~backhaul_manager()' at line 332. Dynamic binding is not used. [virtualCallInConstructor]    virtual void on_thread_stop() override;
```
arnout: This one was already in there, it shouldn't be duplicated.

To regenerate this file, just run ci/cppcheck/cppcheck-new-issues.sh manually and copy cppcheck_results.txt to this file. Or better yet let CI do the job and take it from the cppcheck-diff artefacts.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -220,11 +222,65 @@ void monitor_stats::process()
             sta_stats_msg.rx_rssi           = sta_stats.rx_rssi_curr;
 
             sta_count++;
+
+            beerocks_message::sBssidInfo bss_info;
```
arnout: You should also get bssid here. It's a bit hard to get, because the sta_node only has a vap_id:
```
    bss_info.bssid = network_utils::mac_from_string(mon_db->vap_get_by_id(sta_node->get_vap_id())->get_mac());
```
and you need to handle the possible nullpointer from vap_get_by_id.

But setting bssid this way removes the need for the backhaul manager to be involved in the response path.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -51,7 +51,7 @@ void monitor_stats::add_request(uint16_t id, uint8_t sync)
         }
         next_poll_id = mon_db->get_poll_id() + 1;
     }
-    requests_list.push_back(id);
+    requests_list.push_back(sMeasurementsRequest(id, sta_mac));
```
arnout, line 54:
```suggestion
    requests_list.push_back({ id, sta_mac });
```
With that, the constructor is not needed.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -54,6 +54,159 @@ void monitor_stats::add_request(uint16_t id, uint8_t sync, const sMacAddr &sta_m
...
+    beerocks_header->actionhdr()->id() = request.message_id;
+    message_com::send_cmdu(slave_socket, cmdu_tx);
+}
+
+void monitor_stats::send_associated_sta_link_metrics(const sMeasurementsRequest &request)
+{
+    for (auto it = mon_db->sta_begin(); it != mon_db->sta_end(); ++it) {
+        auto sta_mac  = it->first;
+        auto sta_node = it->second;
+        if (sta_node == nullptr) {
```
arnout:
```suggestion
        if (!sta_node) {
```

arnout (approved):
Although I have a lot of nitpick comments, there is only one that is important enough to be absolutely required to change: don't use C-style cast.

Feel free to ignore all the other comments.

I propose to merge this PR even though it doesn't pass yet in certification.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2142,6 +2142,16 @@ sMacAddr backhaul_manager::get_sta_bssid(
     return bssid;
 }
 
+std::shared_ptr<backhaul_manager::sRadioInfo>
+backhaul_manager::get_sta_radio(const sMacAddr &sta_mac)
+{
+    auto radio = std::find_if(
+        slaves_sockets.begin(), slaves_sockets.end(), [&sta_mac](std::shared_ptr<sRadioInfo> r) {
```
mariomaz, line 2120: I think that parameter to predicate function should be const reference
```
const std::shared_ptr<sRadioInfo> &radio_info
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Most of the times the mid value is formatted as hex with `std::hex` while logging

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
rmelotte: Good point. 
Also, since it does not share the ownership of the sRadioInfo, I don't see why it couldn't be a `sRadioInfo &` instead of a shared_ptr.

rmelotte (approved):
:+1: 
There are still some nitpick comments remaining, but as this PR is needed for https://github.com/prplfoundation/prplMesh/issues/394 and is a giant leap forward for 4.7.7, let's merge it!

PR #1151, opened 2020-04-16
controller: dcs-task: rename single-scan data member

adam1985d:
added a comment explaining the need for single-scan data members and the
need to differentiate between single-scan-requests and continuous-scans.

changed `m_is_single_scan` to `m_single_scan_request`.
changed `m_is_single_scan_pending` to `m_single_scan_request_pending`.

### DIFF
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
@@ -93,8 +93,13 @@ private:
         beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
     eEvent m_dcs_waiting_for_event = eEvent::INVALID_EVENT;
 
-    bool m_is_single_scan_pending = false;
-    bool m_is_single_scan         = false;
+    // The single scan request scan-parameters and scan-results are stored separately from the
+    // scan-parameters and scan-results of the continuous scans. This enables the support of both
+    // continuous-scans mode and single-scan-request simultaneously.
+    // The following data-members are used to read the scan parameters and save the scan results of
+    // the scan that is currently handled.
+    bool m_single_scan_request_pending = false;
+    bool m_single_scan                 = false;
 
     static int m_scanning_task_id; // Currently scanning task id. -1 means no scan in progress.
 };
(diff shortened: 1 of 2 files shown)


kantera800 (approved):

tomereli (commented):
Its good to have the comment but I don't get why you change the name from `m_single_scan` to `m_single_scan_request` - if it is true, doesn't it mean that we are currently running a single scan? If so, adding `_request` suffix is confusing.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
```diff
@@ -92,8 +92,13 @@ class dynamic_channel_selection_task : public task, public std::enable_shared_fr
         beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
     eEvent m_dcs_waiting_for_event = eEvent::INVALID_EVENT;
 
-    bool m_is_single_scan_pending = false;
-    bool m_is_single_scan         = false;
+    // The single scan request scan-parameters and scan-results are stored separately than the
```
vitalybu:
```suggestion
    // The single scan request scan-parameters and scan-results are stored separately from the
```

vitalybu: Are there really cases where both scans are triggered simultaneously?
Why don't we handle it by creating two instances of the DCS task?

vitalybu (approved):
Have some logic related questions, but the renaming looks fine to me, so approving.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.h
adam1985d: fixed the comment.

an instance is per interface and not per run
the task is created as part of agent-join and "lives" until agent gets disconnected

adam1985d:
> See, here I think it will be more clear if we name it as follows:
> 
> ```
> if (m_single_scan_request_pending) {
>   m_single_scan = true;
>   m_single_scan_request_pending = false;
> }
> ```

@kantera800 I agree with @tomereli on this one - what do you think?

kantera800:
> > See, here I think it will be more clear if we name it as follows:
> > ```
> > if (m_single_scan_request_pending) {
> >   m_single_scan = true;
> >   m_single_scan_request_pending = false;
> > }
> > ```
> 
> @kantera800 I agree with @tomereli on this one - what do you think?

yes its better

tomereli (approved):

Issue #1154, opened 2020-04-16
[BUG] bwl: nl80211_client::get_radio_info missing channels on RDK-B


itayx:
**Describe the bug** -
as described in #1046, channels with the flag NO-IR still get filtered in the get_radio_info function.

arnout:
@itayx Could you include the output of `read_supported_channels` and `iw phy0 info` like you did in #1046?

It looks like #1053 didn't completely fix #1046. Now the channels are mostly correct, but DFS channels are still missing.

The reason is that RDK-B uses a different version of the driver, and there the radar channels are masked as NO_IR. This means it is not allowed to configure an AP on that channel, but it is allowed to configure a STA. Since the supported_channels_list corresponds to the channels that can be used as an AP, the current behaviour appears to be correct at first sight.

However, it turns out that hostapd has an exception for this situation: if a channel is marked as a radar channel (`NL80211_FREQUENCY_ATTR_RADAR`) and DFS is enabled in hostapd, then the NO_IR bit is not considered. So in prplMesh we need to emulate the same behaviour.

After one-on-one discussion with Itay, we came to the following solution:
- nl80211_client_impl should keep track of the NO_IR attribute explicitly instead of simply ignoring the channel;
- the bwl implementation should take into account the `no_ir` and `is_dfs` fields of the channel list when constructing the supported_channel_list;
- for now, we assume that hostapd and driver *do* support DFS. If we ever run into a situation where this isn't the case, we'll handle it at that time. It's easy to notice it because hostapd will give an error when we try to select a DFS channel.

kantera800:
>  DFS is enabled in hostapd

this flag **only** can be read from  hosatapd (patched msg) GET_HW_FEATURES that we are replacing ...

> * for now, we assume that hostapd and driver _do_ support DFS. If we ever run into a situation where this isn't the case, we'll handle it at that time. It's easy to notice it because hostapd will give an error when we try to select a DFS channel.

since we should do this assumption on every platform that we support i suggest: 
-> add method to bwl   `is_radar_detection_supported()` that returns true.
-> implement this method in DWPAL bwl  by reading from GET_HW_FEATURES 

@arnout @vitalybu  what do you think ?

vitalybu:
> since we should do this assumption on every platform that we support i suggest:
> -> add method to bwl `is_radar_detection_supported()` that returns true.
> -> implement this method in DWPAL bwl by reading from GET_HW_FEATURES

[...]

PR #1156, opened 2020-04-16
bwl: update the get_radio_info to include additional flags

itayx:
Currently all channels that are not supported get filtered out.
The get_radio_info filters flags that indicate the channel cannot be
used as an AP. the flag NO_IR is filtered as part of this process, which
is incorrect seeing as radios with radar detection can use those
channels.

Lessen the filtering check in get_radio_info for channels.
Add a "no_ir" flag to the channel_info struct

Signed-off-by: itay elenzweig <itayx.elenzweig@intel.com>

### DIFF
--- a/common/beerocks/bwl/shared/nl80211_client_impl.cpp
+++ b/common/beerocks/bwl/shared/nl80211_client_impl.cpp
@@ -317,9 +317,9 @@ bool nl80211_client_impl::get_radio_info(const std::string &interface_name, radi
                         return;
                     }
 
+                    // Ignore "channels" that are disabled or with no frequency.
                     if (!tb_freq[NL80211_FREQUENCY_ATTR_FREQ] ||
-                        tb_freq[NL80211_FREQUENCY_ATTR_DISABLED] ||
-                        tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
+                        tb_freq[NL80211_FREQUENCY_ATTR_DISABLED]) {
                         continue;
                     }
 
@@ -366,10 +366,25 @@ bool nl80211_client_impl::get_radio_info(const std::string &interface_name, radi
                             beerocks::eWiFiBandwidth::BANDWIDTH_160);
                     }
 
+                    if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR] ||
+                        tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
+                        /**
+                         * NL80211_FREQUENCY_ATTR_NO_IR superseded NL80211_FREQUENCY_ATTR_NO_IBSS
+                         * for legacy kernels we still need to check
+                         * __NL80211_FREQUENCY_ATTR_NO_IBSS, for both attributes we use the no_ir.
+                         */
+                        channel.no_ir = true;
+                    }
+
                     if (tb_freq[NL80211_FREQUENCY_ATTR_DFS_STATE] &&
                         nla_get_u32(tb_freq[NL80211_FREQUENCY_ATTR_DFS_STATE]) ==
                             NL80211_DFS_UNAVAILABLE) {
                         channel.radar_affected = true;
+                    }
+
+                    //Filter out chanels that are not DFS but still have the NO_IR flag
+                    if (!channel.is_dfs && channel.no_ir) {
+                        continue;
                     }
 
                     band.supported_channels[channel.number] = channel;
--- a/common/beerocks/bwl/include/bwl/nl80211_client.h
+++ b/common/beerocks/bwl/include/bwl/nl80211_client.h
@@ -141,6 +141,13 @@ public:
         bool radar_affected = false;
 
         /**
+         * This channel does not initiate radiation, meaning it is supported only when radar
+         * detection is supported as well.
+         * Set to true if attribute NL80211_FREQUENCY_ATTR_NO_IR is present
+         */
+        bool no_ir = false;
+
+        /**
          * @brief Gets the maximum supported bandwidth by the channel.
          *
          * @return Maximum supported bandwidth


tomereli (approved):
Looking good. @mariomaz should review this as well.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -317,9 +317,9 @@ bool nl80211_client_impl::get_radio_info(const std::string &interface_name, radi
                         return;
                     }
 
+                    // Ignore "channels" that are disabled or with no frequency.
                     if (!tb_freq[NL80211_FREQUENCY_ATTR_FREQ] ||
-                        tb_freq[NL80211_FREQUENCY_ATTR_DISABLED] ||
-                        tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
```
kantera800, line 322: if removing  `__NL80211_FREQUENCY_ATTR_NO_IBSS`   is part of 

> Lessen the filtering

then you should mention why in the commit message.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -335,6 +335,10 @@ bool nl80211_client_impl::get_radio_info(const std::string &interface_name, radi
                         channel.is_dfs = true;
                     }
 
+                    if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR]) {
+                        channel.no_ir = true;
+                    }
```
kantera800: In the future we would want to check if radar detection is supported and if not these NO_IR (DFS) channels should not be included in the AP capabilities.

This brings me to the next point, 
I suggest that we  add `is_radar_detection_supported`  flag to the AP class , IMO should be read driver but maybe for now it can be read from prplmesh_db (need to add it there also) @arnout  what do you think
Then add this  NO_IR channels check :

```
                    // Ignore DFS channels in case AP doesn't support radar detection.
                    if (channel.no_ir && is_radar_detection_supported) {
                        continue;
                    }
```

kantera800 (changes requested):

common/beerocks/bwl/shared/nl80211_client_impl.cpp
itayx: the __NL80211_FREQUENCY_ATTR_NO_IBSS is deprecated by NL80211_FREQUENCY_ATTR_NO_IR
and we need the NO_IR flag, meaning we cannot filter it out

Updated the commit message to better reflect that

common/beerocks/bwl/shared/nl80211_client_impl.cpp
itayx: I believe a better check is to check the `is_dfs` tag as well as the `no_ir` flag

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -335,6 +335,10 @@ bool nl80211_client_impl::get_radio_info(const std::string &interface_name, radi
                         channel.is_dfs = true;
                     }
 
+                    if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR]) {
```
mariomaz: At some point in time NL80211_FREQUENCY_ATTR_NO_IBSS and NL80211_FREQUENCY_ATTR_PASSIVE_SCAN were replaced by NL80211_FREQUENCY_ATTR_NO_IR. 

Also NL80211_FREQUENCY_ATTR_NO_IBSS was renamed to __NL80211_FREQUENCY_ATTR_NO_IBSS.

From then on, both NL80211_FREQUENCY_ATTR_NO_IR and __NL80211_FREQUENCY_ATTR_NO_IBSS
are set if NO_IR. 

However, in old kernels __NL80211_FREQUENCY_ATTR_NO_IBSS will be present but NL80211_FREQUENCY_ATTR_NO_IR will not.

You should probably do something similar to what the `iw` command line tool is doing:
```c
	/* If both flags are set assume an new kernel */
	if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR] && tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
		print_flag("no IR", &open);
	} else if (tb_freq[NL80211_FREQUENCY_ATTR_PASSIVE_SCAN]) {
		print_flag("passive scan", &open);
	} else if (tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]){
		print_flag("no ibss", &open);
	}
```

mariomaz (changes requested):

common/beerocks/bwl/shared/nl80211_client_impl.cpp
itayx: Changed the if statement and added a comment to describe it

mariomaz: In old kernels __NL80211_FREQUENCY_ATTR_NO_IBSS will be present and NL80211_FREQUENCY_ATTR_NO_IR will not. 

We should decide on how these kernels should be treated. Shall we ignore the channel? Shall we set no_ir to true? Shall we add a no_ibss flag?

mariomaz (changes requested):
This condition is not being considered (maybe because we don't have to deal with old kernels)
```c
if (!tb_freq[NL80211_FREQUENCY_ATTR_NO_IR] && tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
    // TODO
}

File: common/beerocks/bwl/shared/nl80211_client_impl.cpp (hunk 3 above)
Comment (arnout) [reply common/beerocks/bwl/shared/nl80211_client_impl.cpp]: I guess we should then make it a 4-state enum like the iw tool does: NO_IR, PASSIVE_SCAN, NO_IBSS or nothing at all.

Or, we could make a very simplifying assumption, and say:
- if RADAR is set, we mark it as a DFS channel;
- else if NO_IBSS is set, we ignore the channel.
That more or less corresponds to what hostapd does.

By the way, please note:
```
#define NL80211_FREQUENCY_ATTR_PASSIVE_SCAN	NL80211_FREQUENCY_ATTR_NO_IR
#define NL80211_FREQUENCY_ATTR_NO_IBSS		NL80211_FREQUENCY_ATTR_NO_IR
```
so actually the passive scan case is NO_IR without NO_IBSS. So I assume the split between NO_IR and NO_IBSS was made to support the passive scan case.

**Comment by arnout** [comment]
Note that (at least in its current state) this PR doesn't fully solve #1154. It does add the `no_ir` marker, but then that value is completely ignored. We should still remove the `no_ir` channels that are not radar channels.

File: common/beerocks/bwl/shared/nl80211_client_impl.cpp [hunk 4]
```diff
@@ -335,6 +335,13 @@ bool nl80211_client_impl::get_radio_info(const std::string &interface_name, radi
                         channel.is_dfs = true;
                     }
 
+                    // NL80211_FREQUENCY_ATTR_NO_IBSS and NL80211_FREQUENCY_ATTR_NO_IR are set to
+                    // NO_IR. for legacy, check __NL80211_FREQUENCY_ATTR_NO_IBSS
```
kantera800: If its legacy then  why not 
```
                    if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR] ||
                        tb_freq[__NL80211_FREQUENCY_ATTR_NO_IBSS]) {
```

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: > if NO_IBSS is set, we ignore the channel.

This is what we had before this PR. The problem is that we have some devices that set the NO_IBSS attribute and others that don't when AFAIK they should behave similarly.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: That's fine as long as you don't want to distinguish older kernels and want to treat older NO_IBSS meaning as the new NO_IR.
This solution however does not consider NL80211_FREQUENCY_ATTR_PASSIVE_SCAN. I don't mean we should do it. I don't even know what all these flags mean.

kantera800:
> Note that (at least in its current state) this PR doesn't fully solve #1154. It does add the `no_ir` marker, but then that value is completely ignored. We should still remove the `no_ir` channels that are not radar channels.

If by radar channels you mean dfs than I agree (pretty sure they are all dfs thought) still you cannot  use no_ir data without assuming AP has radar detection.  my suggestion is to add a flag to the ap class

itayx:
The `NL80211_FREQUENCY_ATTR_NO_IR` attribute is a deprecation of `NL80211_FREQUENCY_ATTR_PASSIVE_SCAN` & `NL80211_FREQUENCY_ATTR_NO_IBSS` as seen here:
```
#define NL80211_FREQUENCY_ATTR_PASSIVE_SCAN	NL80211_FREQUENCY_ATTR_NO_IR
#define NL80211_FREQUENCY_ATTR_NO_IBSS		NL80211_FREQUENCY_ATTR_NO_IR
```
So if we want to handle the different flags with an enum, that will support both the new and old version of kernel as suggested by @arnout & @mariomaz, then I suggest the following:
```
if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IR]) {
    if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IBSS]) {
        channel.flag = state_enum::no_ir;
    } else {
        channel.flag = state_enum::passive_scan;
    }
} else {
    if (tb_freq[NL80211_FREQUENCY_ATTR_NO_IBSS]) {
        channel.flag = state_enum::no_ibss;
    }
}
channel.flag = state_enum::NA;
```

I've also noticed an attribution flag `NL80211_FREQUENCY_ATTR_RADAR`
Its description in `linux/nl80211.h` from what I saw is
> Radar detection is mandatory on this channel in current regulatory domain.

We store this value as 
```
if (tb_freq[NL80211_FREQUENCY_ATTR_RADAR]) {
    channel.is_dfs = true;
}
```
So maybe a validation check can be something along the lines of:
```
if (channel.flag == state_enum::no_ir && !channel.is_dfs) {
    // do stuff
}
```

vitalybu (commented):
Why do we need the `passive_mode` flag anyway?
Looking at the latest `iw` [source code](https://git.kernel.org/pub/scm/linux/kernel/git/jberg/iw.git/tree/event.c#n35), it looks like the only "interesting" flags are `no_ir` and `no_ibss`.

>  NL80211_FREQUENCY_ATTR_NO_IR: no mechanisms that initiate radiation
> are permitted on this channel, this includes sending probe
> requests, or modes of operation that require beaconing.
> NL80211_FREQUENCY_ATTR_NO_IBSS: IBSS networks are not permitted
> on this channel in current regulatory domain.

So if as @arnout stated in #1154 -
> for now, we assume that hostapd and driver do support DFS. If we ever run into a situation where this isn't the case, we'll handle it at that time. It's easy to notice it because hostapd will give an error when we try to select a DFS channel.

Shouldn't we simply ignore the `no_ir` flag (or include any `no_ir` channels in the list of supported channels)?

kantera800:
> 
> Shouldn't we simply ignore the `no_ir` flag (or include any `no_ir` channels in the list of supported channels)?

IMO we can't ignore these flags and just add all channels to the supported list.
DCS task works with the supported channel list so we could end up scanning a frequency that is not allowed.

mariomaz (approved):
Approving as I cannot add any more value to this discussion.

kantera800 (approved):
looks good.
Just fix the clang issue

common/beerocks/bwl/shared/nl80211_client_impl.cpp
itayx: Added a filtering check before the end of the iteration
if the channel is not a DFS channel (denoted by the `is_dfs` flag) but still has the NO_IR or NO_IBSS attributes (the `no_ir` is set to `true`) then the channel will be ignored

PR #1165, opened 2020-04-17
tests: add a list with all the MAUT tests we can currently pass

rmelotte:
Previously this list was contained in variables on Gitlab.com.

However, Gitlab is not the only place where we want to run the MAUT
tests we can currently pass.

Also, having this list in the repository makes it easier for people to
add new passing tests when working on a flow.

The list of passing controller tests is not added because we can
already pass them all, so all_controller_tests.txt can be used for the
same purpose.


tomereli (approved):

arnout (approved):
The list may already be out of date, but we can fix that later :-)

arnout:
I guess as soon as this is merged (i.e., now), the gitlab variable will be updated?

Also, did you document this anywhere?

rmelotte:
I updated [the wiki page](https://github.com/prplfoundation/prplMesh/wiki/Testing-in-CI) and made a hotfix to mention it in CONTRIBUTING.md: https://github.com/prplfoundation/prplMesh/pull/1171 .

I updated the variables on Gitlab.com already since this should get merged really soon by mergify.

Issue #1105, opened 2020-04-07
[BUG] [client association] starvation due to max client connection 


kantera800:
**Describe the bug** -
max number of Clients that can be connected to a bssid is 255.
agent handles messages from hostapd by polling in 250ms intervals.
when clients start connecting every 100ms , association requests that are sent from hostapd not handled  right away. 
As a result system experience starvation witch cause major delays .

## Steps to reproduce

1. Configure the DUT AP in 5G  ANACAX 160MHz BW , Channel 36., Security - Open.
2. Configure the maxassoc=255
3. Try connecting 255 clients to DUT AP in 100ms intervals
4. Verify in the sniffer capture that AP  Probe Responses delay increases up to max 40 sec.

## Expected behavior

1. All 255 Clients Should connect to AP
2. prplmesh still in operational mode.

## DUT
 UGW AXEPOINT6000

## Analysis 
The bug root cause is that on the monitor `update_sta_stats()` function we sample all station statistics on every 250 msec. It is observed that when clients are connecting, pulling station statistics from dwpal takes about ~3-4 msec per station, which results in 750 msec (on the worst-case scenario) for 255 stations, and causing to the starvation which was observed in the bug description.

## Proposed solution
1. Add a timer which will start right after `select` call on socket thread, and finish before the select call.
Counting the time that the thread operation took, and if it took more than the `select timeout` print warning message. This will help us to understand if any thread exceeds is allowed time window.
2. Stop `update_sta_stats()` operation if it exceeds the thread time window, and continue collecting skipped station statistics on the next thread iteration.
3. Find out wav bwl parser solution is much more efficient than dwpal parser.

PR #1166, opened 2020-04-19
Bugfix/fix starvation due to max clients connected

morantr:
Fix to #1105 

Change the save the select wakeup time on socket thread and charged set_select_timeout, to be able to stop a thread work in order to let it do other stuff such as message handling so it wouldn't be blocked for a long time and change temporarily the select timeout so the thread could finish its work quickly and not wait for a whole select timeout cycle.

Change also the way we send HOSTAP_STATS_MEASUREMENT_RESPONSE message since it causes to a buffer overflow and could not be sent on the bus since it is one big message that is bigger than MTU size restriction.

### DIFF
--- a/agent/src/beerocks/monitor/monitor_db.h
+++ b/agent/src/beerocks/monitor/monitor_db.h
@@ -312,6 +312,8 @@ public:
         return sta_nodes.end();
     }
 
+    size_t get_sta_count() const { return sta_nodes.size(); }
+
     // Monitor parameters //
     void set_arp_burst_delay(int arp_burst_delay_) { arp_burst_delay = arp_burst_delay_; }
     uint8_t get_arp_burst_delay() { return arp_burst_delay; }
--- a/agent/src/beerocks/monitor/monitor_stats.cpp
+++ b/agent/src/beerocks/monitor/monitor_stats.cpp
@@ -134,7 +134,17 @@ void monitor_stats::process()
             return;
         }
 
-        auto &ap_stats_msg                  = response->ap_stats();
+        if (!response->alloc_ap_stats()) {
+            LOG(ERROR) << "tx_buffer overflow!";
+        }
+
+        auto ap_stats_tuple = response->ap_stats(0);
+        if (!std::get<0>(ap_stats_tuple)) {
+            LOG(ERROR) << "Failed to get element";
+            return;
+        }
+        auto &ap_stats_msg = std::get<1>(ap_stats_tuple);
+
         ap_stats_msg.stats_delta_ms         = radio_stats.delta_ms;
         ap_stats_msg.rx_packets             = radio_stats.hal_stats.rx_packets;
         ap_stats_msg.tx_packets             = radio_stats.hal_stats.tx_packets;
@@ -150,6 +160,27 @@ void monitor_stats::process()
         ap_stats_msg.client_tx_load_percent = radio_stats.client_tx_load_tot_curr;
         ap_stats_msg.client_rx_load_percent = radio_stats.client_rx_load_tot_curr;
 
+        auto total_sta_count = mon_db->get_sta_count();
+        size_t elements_to_allocate;
+        uint32_t sta_count;
+
+        auto allocate_sta_stats_elements = [&]() {
+            elements_to_allocate =
+                cmdu_tx.elements_in_message(sizeof(beerocks_message::sStaStatsParams));
+
+            elements_to_allocate = std::min(total_sta_count, elements_to_allocate);
+
+            if (!response->alloc_sta_stats(elements_to_allocate)) {
+                LOG(ERROR) << "tx_buffer overflow! elements_to_allocate="
+                           << int(elements_to_allocate);
+                return;
+            }
+
+            sta_count = 0;
+        };
+
+        allocate_sta_stats_elements();
+
         for (auto it = mon_db->sta_begin(); it != mon_db->sta_end(); ++it) {
             auto sta_mac  = it->first;
             auto sta_node = it->second;
@@ -159,13 +190,21 @@ void monitor_stats::process()
 
             auto &sta_stats = sta_node->get_stats();
 
-            if (!response->alloc_sta_stats()) {
-                LOG(ERROR) << "tx_buffer overflow!";
-                return;
+            if (elements_to_allocate == sta_count) {
+                message_com::send_cmdu(slave_socket, cmdu_tx);
+                response = message_com::create_vs_message<
+                    beerocks_message::cACTION_MONITOR_HOSTAP_STATS_MEASUREMENT_RESPONSE>(cmdu_tx);
+                if (!response) {
+                    LOG(ERROR) << "Failed building "
+                                  "cACTION_MONITOR_HOSTAP_STATS_MEASUREMENT_RESPONSE message!";
+                    return;
+                }
+
+                total_sta_count -= sta_count;
+                allocate_sta_stats_elements();
             }
 
-            uint8_t num_of_stas = response->sta_stats_size();
-            auto &sta_stats_msg = std::get<1>(response->sta_stats(num_of_stas - 1));
+            auto &sta_stats_msg = std::get<1>(response->sta_stats(sta_count));
 
             sta_stats_msg.mac               = network_utils::mac_from_string(sta_mac);
             sta_stats_msg.rx_packets        = sta_stats.hal_stats.rx_packets;
@@ -179,6 +218,8 @@ void monitor_stats::process()
             sta_stats_msg.rx_load_percent   = sta_stats.rx_load_percent_curr;
             sta_stats_msg.stats_delta_ms    = sta_stats.delta_ms;
             sta_stats_msg.rx_rssi           = sta_stats.rx_rssi_curr;
+
+            sta_count++;
         }
 
         beerocks_header->actionhdr()->id() = requests_list.front();
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -489,6 +489,11 @@ bool monitor_thread::update_sta_stats()
     auto poll_cnt  = mon_db.get_poll_cnt();
     auto poll_last = mon_db.is_last_poll();
 
+    if (m_sta_stats_polling_completed) {
+        m_sta_stats_polling_start_timestamp = std::chrono::steady_clock::now();
+        m_sta_stats_polling_completed       = false;
+    }
+
     for (auto it = mon_db.sta_begin(); it != mon_db.sta_end(); ++it) {
 
         auto sta_mac  = it->first;
@@ -501,6 +506,19 @@ bool monitor_thread::update_sta_stats()
 
         auto vap_node   = mon_db.vap_get_by_id(sta_node->get_vap_id());
         auto &sta_stats = sta_node->get_stats();
+
+        // Skip stations that were already updated in the current cycle
+        if (sta_stats.last_update_time > m_sta_stats_polling_start_timestamp) {
+            continue;
+        }
+
+        if (std::chrono::steady_clock::now() > awake_timeout()) {
+            // If we haven't finish to iterate on all stations, skip one time on the select timeout
+            // so the select will not be stuck on full select timeout and the thread will be able to
+            // finish this operation quickly.
+            skip_next_select_timeout();
+            return true;
+        }
 
         // Update the stats
         if (!mon_wlan_hal->update_stations_stats(vap_node->get_iface(), sta_mac,
@@ -576,6 +594,8 @@ bool monitor_thread::update_sta_stats()
         sta_stats.delta_ms         = float(time_span.count());
         sta_stats.last_update_time = now;
     }
+
+    m_sta_stats_polling_completed = true;
 
     return true;
 }
--- a/agent/src/beerocks/slave/ap_manager_thread.cpp
+++ b/agent/src/beerocks/slave/ap_manager_thread.cpp
@@ -22,7 +22,7 @@ using namespace beerocks::net;
 ////////////////////////// Local Module Definitions //////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-#define SELECT_TIMEOUT_MSC 200
+#define SELECT_TIMEOUT_MSC 1000
 #define ACS_READ_SLEEP_USC 1000
 #define READ_ACS_ATTEMPT_MAX 5
 #define DISABLE_BACKHAUL_VAP_TIMEOUT_SEC 30
@@ -341,7 +341,9 @@ void ap_manager_thread::after_select(bool timeout)
             return;
         } else {
             LOG(INFO) << "waiting to attach to " << ap_wlan_hal->get_radio_info().iface_name;
-            UTILS_SLEEP_MSEC(200);
+            // Set the sleep to little less than the select timeout so we won't get awake timeout
+            // print to the log on the socket thread.
+            UTILS_SLEEP_MSEC(SELECT_TIMEOUT_MSC * 0.9);
         }
 
         // AP is attached
--- a/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
@@ -55,6 +55,8 @@ protected:
     inline void clear_ready(Socket *s) { select.clearReady(s); }
     virtual bool read_ready(Socket *s) { return select.readReady(s); }
 
+    void skip_next_select_timeout();
+
     ieee1905_1::CmduMessageTx cmdu_tx;
     ieee1905_1::CmduMessageTx cert_cmdu_tx;
     const std::string unix_socket_path;
@@ -80,6 +82,9 @@ private:
 
     int server_max_connections;
     SocketSelect select;
+
+    uint32_t m_select_timeout_msec  = 0;
+    bool m_skip_next_select_timeout = false;
 };
 
 } // namespace beerocks
--- a/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
@@ -16,6 +16,8 @@
 #include <tlvf/CmduMessageRx.h>
 #include <tlvf/CmduMessageTx.h>
 
+#include <chrono>
+
 #define DEFAULT_SELECT_TIMEOUT_MS 500
 
 namespace beerocks {
@@ -57,6 +59,17 @@ protected:
 
     void skip_next_select_timeout();
 
+    // This function needs to be used on code parts in before/after select functions, where we
+    // suspect that its operation might take a long time (> select_timeout) and will not finish the
+    // thread operation before the select timeout should have been expired if the thread had ideal
+    // operation time of zero.
+    std::chrono::steady_clock::time_point awake_timeout()
+    {
+        // Set the awake timeout to 50% of the select timeout to create a buffer of 50%
+        // to let the thread finish its work until getting to Select again.
+        return m_select_wake_up_time + std::chrono::milliseconds(int(0.50 * m_select_timeout_msec));
+    }
+
     ieee1905_1::CmduMessageTx cmdu_tx;
     ieee1905_1::CmduMessageTx cert_cmdu_tx;
     const std::string unix_socket_path;
@@ -83,8 +96,9 @@ private:
     int server_max_connections;
     SocketSelect select;
 
-    uint32_t m_select_timeout_msec  = 0;
-    bool m_skip_next_select_timeout = false;
+    uint32_t m_select_timeout_msec                              = 0;
+    bool m_skip_next_select_timeout                             = false;
+    std::chrono::steady_clock::time_point m_select_wake_up_time = std::chrono::steady_clock::now();
 };
 
 } // namespace beerocks
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -278,6 +278,7 @@ bool socket_thread::work()
         m_skip_next_select_timeout = false;
         set_select_timeout(m_select_timeout_msec);
     }
+    m_select_wake_up_time = std::chrono::steady_clock::now();
 
     after_select(bool(sel_ret == 0));
 
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -92,9 +92,21 @@ bool socket_thread::init()
 
 void socket_thread::set_select_timeout(unsigned msec)
 {
+    m_select_timeout_msec = msec;
+
     struct timeval tv;
     tv.tv_sec  = (msec / 1000);
     tv.tv_usec = (1000 * (msec % 1000));
+    select.setTimeout(&tv);
+}
+
+void socket_thread::skip_next_select_timeout()
+{
+    m_skip_next_select_timeout = true;
+
+    struct timeval tv;
+    tv.tv_sec  = 0;
+    tv.tv_usec = 0;
     select.setTimeout(&tv);
 }
 
@@ -259,6 +271,12 @@ bool socket_thread::work()
 
         THREAD_LOG(ERROR) << "select error: " << strerror(errno);
         return false;
+    }
+
+    // If select was skipped, rest the select timeout to default value
+    if (m_skip_next_select_timeout) {
+        m_skip_next_select_timeout = false;
+        set_select_timeout(m_select_timeout_msec);
     }
 
     after_select(bool(sel_ret == 0));
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -261,6 +261,15 @@ bool socket_thread::work()
 {
     before_select();
 
+    auto awake_time_msec = std::chrono::duration_cast<std::chrono::milliseconds>(
+                               std::chrono::steady_clock::now() - m_select_wake_up_time)
+                               .count();
+
+    if (m_select_timeout_msec > 0 && awake_time_msec > m_select_timeout_msec) {
+        THREAD_LOG(WARNING) << "Thread awake time is exceptionally long: " << int(awake_time_msec)
+                            << " [ms]";
+    }
+
     int sel_ret = select.selectSocket();
     if (sel_ret < 0) {
         // Do not fail for the following "errors"
--- a/controller/src/beerocks/cli/beerocks_cli_main.cpp
+++ b/controller/src/beerocks/cli/beerocks_cli_main.cpp
@@ -267,10 +267,8 @@ static void cli_interactive(std::string path, std::string tmp_path, std::string 
               << BEEROCKS_BUILD_DATE << std::endl;
     std::cout << "Enter 'help' for help, 'exit' to exit" << std::endl;
 
-    beerocks::cli_socket cli_soc(tmp_path, ip);
     beerocks::cli_bml cli_bml(path);
-    beerocks::cli *cli_ptr = &cli_soc;
-    int is_onboarding      = -1;
+    int is_onboarding = -1;
 
     if (!cli_bml.connect()) {
         std::cout << "cli_bml: Can't connect to BML, exit..." << std::endl;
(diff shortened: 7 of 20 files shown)


common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
```diff
@@ -29,7 +31,7 @@ class socket_thread : public thread_base {
     socket_thread(const std::string &unix_socket_path_ = std::string());
     virtual ~socket_thread();
     void set_server_max_connections(int connections);
-    virtual void set_select_timeout(unsigned msec);
+    virtual void set_select_timeout(unsigned msec, bool sticky = true);
```
kantera800: if you are changing the signature then at least add doxygen that explains the new parameter

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -575,8 +585,20 @@ bool monitor_thread::update_sta_stats()
...
         sta_stats.delta_ms         = float(time_span.count());
         sta_stats.last_update_time = now;
+
+        if (std::chrono::steady_clock::now() > awake_timeout()) {
+            // If we haven't finish to iterate on all stations, set the select timeout to minimal
+            // value (1 msec) temporarily so the select will not be stuck on full select timeout
+            // and the thread will be able to finish this operation quickly.
+            set_select_timeout(1, false);
+            return true;
+        }
```
kantera800: If I'm not mistaken basically here you are lowering the  after_select cycle to 75% .
All other task will be polled more frequently , CPU will increase and I'm not sure we want that.

Why not just continue to other tasks when `awake_timeout() `is reaching 75% and keep the cycle as is.
You could also move the update_sta_stats() to the end of the after_select that way you will make sure 
after select will be occupied with tasks only   75% of the time.

*if you want to lower the cycle in such cases just do `set_select_timeout(0.75*TO)`  and restore it after a while .

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -575,8 +585,20 @@ bool monitor_thread::update_sta_stats()
...
     }
 
+    // Reset select timeout value to init value
+    set_select_timeout(0);
```
kantera800: a bit confusing , if you want reset the TO just add `reset_select_timeout`.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
```diff
@@ -29,7 +29,7 @@ class socket_thread : public thread_base {
...
```
vitalybu: **I don't like this.**
The idea is to allow skipping the `select()` timeout (while still checking for incoming message) if the thread has additional work to do.
A better solution would be to add a totally separate boolean flag with a protected setter (e.g. `skip_next_select_timeout(bool skip=true)`). This flag can be checked by the `work()` method of the `beerocks_socket_thread` class and if set, skip the timeout.

The whole sticky/non-sticky is confusing.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
```diff
@@ -55,6 +57,17 @@ class socket_thread : public thread_base {
     inline void clear_ready(Socket *s) { select.clearReady(s); }
     virtual bool read_ready(Socket *s) { return select.readReady(s); }
 
+    // This function needs to be used on code parts in before/after select functions, where we
+    // suspect that its operation might take a long time (> select_timeout) and will not finish the
+    // thread operation before the select timeout should have been expired if the thread had ideal
+    // operation time of zero.
```
vitalybu, line 65: Instead of explaining where this function needs to be used, you should explain what it does.
Maybe a more intuitive version would be a function that returns the number of remaining milliseconds the thread is "allowed" to work.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -575,8 +585,20 @@ bool monitor_thread::update_sta_stats()
             std::chrono::duration_cast<std::chrono::milliseconds>(now - sta_stats.last_update_time);
         sta_stats.delta_ms         = float(time_span.count());
         sta_stats.last_update_time = now;
+
+        if (std::chrono::steady_clock::now() > awake_timeout()) {
+            // If we haven't finish to iterate on all stations, set the select timeout to minimal
+            // value (1 msec) temporarily so the select will not be stuck on full select timeout
+            // and the thread will be able to finish this operation quickly.
+            set_select_timeout(1, false);
```
vitalybu: See my comment on your first commit, but I don't see any reason to have a `1ms` timeout if there's more work that we need to complete.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -256,6 +256,15 @@ bool socket_thread::work()
 {
     before_select();
 
+    auto awake_time_msec = std::chrono::duration_cast<std::chrono::milliseconds>(
+                               std::chrono::steady_clock::now() - m_select_wake_up_time)
+                               .count();
+
+    if (m_select_timeout_msec > 0 && awake_time_msec > m_select_timeout_msec) {
+        THREAD_LOG(WARNING) << "Thread awake time is exceptionally long: " << int(awake_time_msec)
```
vitalybu, line 269: You should probably print the actual timeout value, otherwise it's a bit hard to understand why did you decide that the awake time is exceptionally long.

Another interesting feature we can add is an average processing time of the thread.
For example, store processing times for 10sec (or 30sec), calculate the average value and print it.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -78,4 +78,34 @@ bool CmduMessageTx::finalize()
     if (!addClass<tlvEndOfMessage>())
         return false;
     return msg.finalize();
-};
+}
+
+/**
+ * @brief Return how much element of given size can be allocated without create a MTU restriction
```
vitalybu:
```suggestion
 * @brief Return how much elements of the given size can be allocated within MTU sized buffer.
```

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -78,4 +78,34 @@ bool CmduMessageTx::finalize()
...
+ * overflow.
```
vitalybu:
```suggestion
```

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -78,4 +78,34 @@ bool CmduMessageTx::finalize()
...
+        ptr = class_list->prevClass();
+    }
+    // auto ptr = msg.prevClass()->getInnerClassList()->prevClass();
+    if (!valid_ptr) {
+        return 0;
+    }
+
+    current_msg_length = valid_ptr->getBuffPtr() - getMessageBuff();
+    auto size_left     = MTU_SIZE - current_msg_length - tlvEndOfMessage::get_initial_size();
+    return size_left / size;
```
vitalybu, line 116: If `size == 0` you divide by 0.

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -78,4 +78,34 @@ bool CmduMessageTx::finalize()
...
+    while (ptr) {
+        valid_ptr       = ptr;
+        auto class_list = ptr->getInnerClassList();
+        if (!class_list) {
+            break;
+        }
+        ptr = class_list->prevClass();
+    }
+    // auto ptr = msg.prevClass()->getInnerClassList()->prevClass();
+    if (!valid_ptr) {
```
vitalybu, line 107:
```suggestion
    if (!valid_ptr || !size) {
```

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -78,4 +78,34 @@ bool CmduMessageTx::finalize()
...
+        return 0;
+    }
+
+    current_msg_length = valid_ptr->getBuffPtr() - getMessageBuff();
+    auto size_left     = MTU_SIZE - current_msg_length - tlvEndOfMessage::get_initial_size();
```
vitalybu: Can't this have a negative value?

framework/tlvf/src/src/CmduMessageTx.cpp
```diff
@@ -78,4 +78,34 @@ bool CmduMessageTx::finalize()
...
+}
+
+/**
+ * @brief Return how much element of given size can be allocated without create a MTU restriction
+ * overflow.
+ * 
+ * @param[in] size size of one element in bytes.
+ * @return size_t number of elements of 'size' that can be inserted.
+ */
+size_t CmduMessageTx::mtu_size_restriction_overflow_check(size_t size)
```
vitalybu: Not the best method name.
Consider using something like `elements_in_message`

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -159,13 +183,31 @@ void monitor_stats::process()
...
+                total_sta_count -= sta_count;
+                elements_to_allocate = cmdu_tx.mtu_size_restriction_overflow_check(
+                    sizeof(beerocks_message::sStaStatsParams));
+
+                elements_to_allocate = std::min(total_sta_count, elements_to_allocate);
+
+                if (!response->alloc_sta_stats(elements_to_allocate)) {
+                    LOG(ERROR) << "tx_buffer overflow!";
+                    return;
+                }
```
vitalybu: All this code is a duplication.
Consider wrapping it with a local lambda function and just calling it twice.

controller/src/beerocks/cli/beerocks_cli_main.cpp
```diff
@@ -279,6 +277,8 @@ static void cli_interactive(std::string path, std::string tmp_path, std::string
         is_onboarding = cli_bml.get_onboarding_status();
     }
 
+    beerocks::cli_socket cli_soc(tmp_path, ip);
```
vitalybu, line 280: What? Why?

vitalybu (changes requested):

kantera800:
These changes indeed fix  #1105
The only thing that bothers me is the "select timeout to minimal" in such cases ( above ~80 clients connected) 
` set_select_timeout(1, false);` 

The monitor thread  will work at 99%  constantly (less than 1ms left for sleep) until clients will be disconnected (that could take a while).
This also applies to any other threads/tasks  that we will use this solution in the future.

IMO it should be a configurable `min_select_timeout`, and it should represent  % of time remaining until after_select timeout.
I would not set it below 20% leaving the thread at least at 80%.

**Pros**:
thread will be at  80% maximum.
**Cons**:
RSSI measurement delta polling (for optimal path triggering) will be delayed a bit  since currently we are waiting for 4 measurement average on each client )

Regarding the 4 rssi samples average mechanism , 
IMO we we can remove it and use only 1 sample  since each sample from sta_statistics is already an average  (ShortTermRSSIAverage)- not necessarily related to this  PR

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
morantr: I revert this change and reimplement it as @vitalybu has proposed.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: We talked about this on the phone. Resolving for now.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: I changed it so it would be done automatically by the socket thread.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
morantr: Done.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
morantr: What it does is explained in the function body, also it is kind of obvious, no? I mean, the line explains itself ...
I don't about returning number of remaining msec, since it will cause to add more code on the caller to decide what to do with the returned value. I prefer keeping it simple for now. It fit our needs.  Maybe in the future.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: Done.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
morantr: I'm don't think that it is interesting what is the select timeout and how we decide what is "exceptionally long".
IMO the printing of how much time we are awake is the only thing matters.
Regarding the average value calculation. I don't think that printing the average will help us...

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: Done.

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: Done.

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: change it to "element_in_message_mtu_limit" as a compromise. :)

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: Done, but added as a separate condition at the beginning of the function since there is no point doing anything if the size is 0.

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: Add protection at the function beginning.

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: Added protection.

agent/src/beerocks/monitor/monitor_stats.cpp
morantr: ok

common/beerocks/bcl/source/beerocks_socket_thread.cpp
vitalybu: The whole purpose of log messages is to inform the reader about progress and/or issues.
In this case, printing the actual timeout will help the user to quickly understand the problem.

Also, please don't resolve conversations that you decide to ignore :)

framework/tlvf/src/src/CmduMessageTx.cpp
vitalybu: What if tomorrow we'll use another transport that is not limited to MTU size?
Don't see any reason to mention MTU in the method name...

controller/src/beerocks/cli/beerocks_cli_main.cpp
morantr: I wrote why in the commit message.
I guess the `cli_bml.connect()` is taking too long so it prints the warning message I added to the socket_thread.

framework/tlvf/src/src/CmduMessageTx.cpp
morantr: It is not the transport that limits us. It is IEEE1950.1.

morantr:
> These changes indeed fix #1105
> The only thing that bothers me is the "select timeout to minimal" in such cases ( above ~80 clients connected)
> ` set_select_timeout(1, false);`
> 
> The monitor thread will work at 99% constantly (less than 1ms left for sleep) until clients will be disconnected (that could take a while).
> This also applies to any other threads/tasks that we will use this solution in the future.
> 
> IMO it should be a configurable `min_select_timeout`, and it should represent % of time remaining until after_select timeout.
> I would not set it below 20% leaving the thread at least at 80%.
> 
> **Pros**:
> thread will be at 80% maximum.
> **Cons**:
> RSSI measurement delta polling (for optimal path triggering) will be delayed a bit since currently we are waiting for 4 measurement average on each client )
> 
> Regarding the 4 rssi samples average mechanism ,
> IMO we we can remove it and use only 1 sample since each sample from sta_statistics is already an average (ShortTermRSSIAverage)- not necessarily related to this PR

I changed the implementation, so you might want to update what you have written to fit new implementation.

framework/tlvf/src/src/CmduMessageTx.cpp
vitalybu: Whatever, there's still no reason to have "MTU" as part of the method name.
The users (developers that will call this method) shouldn't care about the underlying limitations, all they want is to know how many elements they can write into a message...

morantr: I don't agree. We are bounded to 1905.1 restrictions. It is not something that will change,
and that is the limit the function uses to calculate how many elements we can add, so the developer actually does care about it. The reason that it is important is that the real limitation is the buffer size limitation, but for this specific message, we want to make sure we haven't crossed the MTU limitation.
For any other message, we wouldn't care about it because the transport will fragment the message,

kantera800:
>
> I changed the implementation, so you might want to update what you have written to fit new implementation.

please elaborate because i don't see any significant changes.
monitor thread will be still at 100% but again maybe i missed something.

framework/tlvf/src/src/CmduMessageTx.cpp
vitalybu: You're correct that we do not plan replacing 1905.1, but the MTU size is actually defined by the data-link layer (Ethernet in our case):
image

I would accept your argument if it was an internal mechanism that would automatically split messages based on the actual MTU size, but you're forcing the user to make this calculation.
And since it's a user facing method, it should have a simple and user-friendly name.

Again, the user shouldn't care why there's a limitation on the number of elements (MTU or any other reason) in the message, but just that there is.

By the way, assuming that MTU is always 1500 bytes is also wrong, the hardware could support Jambo frames and then the MTU size is 9000 bytes...

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -92,12 +92,25 @@ bool socket_thread::init()
 
 void socket_thread::set_select_timeout(unsigned msec)
 {
+    m_select_timeout_msec      = msec;
+    m_skip_next_select_timeout = false;
```
vitalybu: I think it better to set this flag to `false` inside the `work()` method.
It'll make the code a little bit more readable, since the flow will look like:
```cpp
if (m_skip_next_select_timeout) {
    select.setTimeout(0);
}

select.selectSocket();

if (m_skip_next_select_timeout) {
    m_skip_next_select_timeout = false;
    select.setTimeout(m_select_timeout_msec);
}
```

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -261,6 +274,11 @@ bool socket_thread::work()
         return false;
     }
 
+    // If select was skipped, rest the select timeout to default value
+    if (m_skip_next_select_timeout) {
+        set_select_timeout(m_select_timeout_msec);
+    }
+
```
vitalybu: This is a bit confusing. At least update the flag here:
```suggestion
    // If select was skipped, rest the select timeout to default value
    if (m_skip_next_select_timeout) {
        m_skip_next_select_timeout = false;
        set_select_timeout(m_select_timeout_msec);
    }
```

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
```diff
@@ -56,6 +58,16 @@ class socket_thread : public thread_base {
     virtual bool read_ready(Socket *s) { return select.readReady(s); }
 
     void skip_next_select_timeout();
+    // This function needs to be used on code parts in before/after select functions, where we
```
vitalybu, line 62:
```suggestion

    // This function needs to be used on code parts in before/after select functions, where we
```

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -502,6 +507,11 @@ bool monitor_thread::update_sta_stats()
         auto vap_node   = mon_db.vap_get_by_id(sta_node->get_vap_id());
         auto &sta_stats = sta_node->get_stats();
 
+        // Skip station if not finished previous cycle
```
vitalybu:
```suggestion
        // Skip stations that were already updated in the current cycle
```

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -575,8 +585,18 @@ bool monitor_thread::update_sta_stats()
...
         sta_stats.delta_ms         = float(time_span.count());
         sta_stats.last_update_time = now;
+
+        if (std::chrono::steady_clock::now() > awake_timeout()) {
+            // If we haven't finish to iterate on all stations, skip one time on the select timeout
+            // so the select will not be stuck on full select timeout and the thread will be able to
+            // finish this operation quickly.
+            skip_next_select_timeout();
+            return true;
+        }
```
vitalybu: I would move this condition above the call to `update_stations_stats`.
Otherwise it could trigger the `skip_next_select_timeout` even if we just processed the last station in this cycle.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -262,6 +262,15 @@ bool socket_thread::work()
 {
     before_select();
 
+    auto awake_time_msec = std::chrono::duration_cast<std::chrono::milliseconds>(
+                               std::chrono::steady_clock::now() - m_select_wake_up_time)
+                               .count();
+
+    if (m_select_timeout_msec > 0 && awake_time_msec > m_select_timeout_msec) {
+        THREAD_LOG(WARNING) << "Thread awake time is exceptionally long: " << int(awake_time_msec)
+                            << " [ms]";
```
vitalybu, line 270:
```suggestion
        THREAD_LOG(WARNING) << "Thread awake time is exceptionally long: " << int(awake_time_msec)
                            << "/" << m_select_timeout_msec << " [ms]";
```

framework/tlvf/src/include/tlvf/CmduMessageTx.h
```diff
@@ -12,6 +12,8 @@
 #include <tlvf/CmduMessage.h>
 #include <tlvf/ieee_1905_1/tlvVendorSpecific.h>
 
+#define MTU_SIZE (size_t)1500
```
vitalybu, line 15: Should be dynamically read from the interface

framework/tlvf/src/include/tlvf/CmduMessageTx.h
```diff
@@ -27,6 +29,7 @@ class CmduMessageTx : public CmduMessage {
     template <class T> std::shared_ptr<T> addClass() { return msg.addClass<T>(); }
     void reset() { msg.reset(false); }
     bool finalize();
+    size_t element_in_message_mtu_limit(size_t size);
```
vitalybu:
```suggestion
    size_t max_elements_in_message(size_t size);
```

agent/src/beerocks/monitor/monitor_db.h
```diff
@@ -312,6 +312,8 @@ class monitor_db {
         return sta_nodes.end();
     }
 
+    size_t get_sta_count() { return sta_nodes.size(); }
```
vitalybu:
```suggestion
    size_t get_sta_count() const { return sta_nodes.size(); }
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -341,7 +341,9 @@ void ap_manager_thread::after_select(bool timeout)
             return;
         } else {
             LOG(INFO) << "waiting to attach to " << ap_wlan_hal->get_radio_info().iface_name;
-            UTILS_SLEEP_MSEC(200);
+            // Set the sleep to little less than the select timeout so we won't get awake timeout
+            // print to the log on the socket thread.
+            UTILS_SLEEP_MSEC(SELECT_TIMEOUT_MSC * 0.9);
```
vitalybu, line 346: This change has nothing to do with this PR.
Also, it's fine for the threads to work "harder" while trying to attach, so I don't think it's necessary at all.

vitalybu (changes requested):

common/beerocks/bcl/source/beerocks_socket_thread.cpp
morantr: I moved the `m_skip_next_select_timeout = false;` statement outside as you requested. But I left the skipping of the timeout to the skip function. It is not that critical IMO, and with my implementation, I don't need to change the `setTimeout` function on the select class to get msec value instead of proprietary struct.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
morantr: done.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: Done.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: moved.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
morantr: I prefer not. The select timeout is not interesting IMO. The user will understand what's wrong.

agent/src/beerocks/monitor/monitor_db.h
morantr: done

agent/src/beerocks/slave/ap_manager_thread.cpp
morantr: Yes but it is not working hard, it sleeps, and it adds the annoying warning messages to the log while attaching.

framework/tlvf/src/include/tlvf/CmduMessageTx.h
morantr: This number is defined hardcoded on the transport. I think it is better to leave it that way.

framework/tlvf/src/include/tlvf/CmduMessageTx.h
morantr: maybe `bus_limit_elements_in_meassage`? It is more accurate...

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -502,6 +507,19 @@ bool monitor_thread::update_sta_stats()
...
 
+        // Skip stations that were already updated in the current cycle
+        if (sta_stats.last_update_time > m_sta_stats_polling_start_timestamp) {
+            continue;
+        }
+
+        if (std::chrono::steady_clock::now() > awake_timeout()) {
+            // If we haven't finish to iterate on all stations, skip one time on the select timeout
+            // so the select will not be stuck on full select timeout and the thread will be able to
+            // finish this operation quickly.
```
vitalybu, line 518: IMO it's better to move this comment above the `if` condition, but maybe it's just my personal preference :)

common/beerocks/bcl/source/beerocks_socket_thread.cpp
vitalybu: It's **very** interesting because you're using it to calculate the "allowed" awake timeout.

framework/tlvf/src/include/tlvf/CmduMessageTx.h
vitalybu: It's definitely not for this PR, but should probably be handled sometime in the future

agent/src/beerocks/slave/ap_manager_thread.cpp
vitalybu: Why does it show the warning if it's sleeping for 200ms?

.gitignore
```diff
@@ -16,3 +16,4 @@ __pycache__/
 .quilt_checked
 .source_dir
 ipkg-*/
+settings.json
```
vitalybu: We don't add IDE related files to the `.gitignore`.
You can ignore this file in your global `.gitignore` file.

vitalybu (changes requested):

.gitignore
morantr: done

agent/src/beerocks/slave/ap_manager_thread.cpp
morantr: It used to show because the select timeout was 200ms too. Anyway raising it is a good idea since there will be fewer attach prints, and also I wanted to set it as a number which is less than than the select timeout to avoid the warning messages.

framework/tlvf/src/include/tlvf/CmduMessageTx.h
vitalybu: It doesn't matter what the underlying layer is... Just what is the maximal number of logical elements you can add to the message.

vitalybu (approved):

kantera800 (approved):

PR #1168, opened 2020-04-20
Dev/bap wps onboarding 3

### LINKED ISSUE #347: [TASK] Pass flow 4.2.3 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.2.3 in the certification testbed.

# Exit criteria

### LINKED ISSUE #864: [TASK] Add bAP (backhaul AP) support to prplmesh
## Description

Implement the design that was discussed for #861 and refined in #862 

## Exit criteria
- [ ] Backhaul AP support is present in prplMesh
- [ ] The implementation conforms to the documented design (see #862). This may require fine tuning of the design
- [ ] The tests defined in #863 related to backhaul AP support are passing

## Detailed description

To be filled in by the implementer

### LINKED ISSUE #1070: [TASK] Use Multi-AP information elements in beacons and WPS exchange
For onboarding and wireless backhaul AP support, hostapd needs some additional information elements in association request/response messages and WPS credentials. This requires:

- making sure that hostapd has the relevant commits (currently not the case in prplWrt);
- configure hostapd.conf appropriately.

Relevant commits from hostapd:
- 9c06f0f6aed26c1628acaa74df0232dd7b345e9a
- 5abc7823bd01f69b8afbe1fd19f65fff86137c44
- 7488e0ade6dffb6df4c1fb6526a9f3ede0eb18ef
- 0dab477335d9ea8d71bdb631aab7483542ed1bd0
- bfcdac1c8bb7189310d0797b9bdf19d92d0414f3
- cb3c156e7bb057d84f1f62264134892262f25e5d
- 56a2d788f9dea36c32eebb3c54081ecea29766d4
- 83ebf558654615e02beb3be7011e63307bf328b7
- 66819b07b5fc95d0ffe6defafdb9526427c1955b
- 8682f384cee6a8499c7c04793d1998dd3f82e418
- b1daf498a14934c5b0f85d5373998e5c9e7bbdb4
- 1415d4b82d5eab4c8edf2c2b32200137ae0178c3
- 1d9cff86bd04c52fb448a4a83ee18a094613abff

(not all of these are strictly required, e.g. tests and some of the bugfixes).

Required additions to hostapd.conf:
```
# Enable Multi-AP functionality
# 0 = disabled (default)
# 1 = AP support backhaul BSS
# 2 = AP support fronthaul BSS
# 3 = AP supports both backhaul BSS and fronthaul BSS
#multi_ap=0

# Multi-AP backhaul BSS config
# Used in WPS when multi_ap=2 or 3. Defines "backhaul BSS" credentials.
# These are passed in WPS M8 instead of the normal (fronthaul) credentials
# if the Enrollee has the Multi-AP subelement set. Backhaul SSID is formatted
# like ssid2. The key is set like wpa_psk or wpa_passphrase.
#multi_ap_backhaul_ssid="backhaul"
#multi_ap_backhaul_wpa_psk=0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
#multi_ap_backhaul_wpa_passphrase=secret passphrase
```

tomereli:
Add support bAPs for allowing other agents to onboard the prplMesh agent using wireless backhaul:
- Use Intel hostapd mesh mode instead of multi_ap for VAPs configuration
- Add WPS PBC implementation using hostapd upstream support
- Remove backhaul vaps unused code

PR in prplwrt iwlwav - https://git.prpl.dev/prplmesh/iwlwav/-/merge_requests/12

Fixes #1070 
Fixes #347


agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -87,22 +87,23 @@ static void copy_vaps_info(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
 
         // If the VAP ID exists
         if (radio_vaps.find(vap_id) != radio_vaps.end()) {
-            const auto &curr_vap = radio_vaps.at(vap_id);
+            const auto &curr_vap  = radio_vaps.at(vap_id);
+            sMacAddr curr_vap_mac = network_utils::mac_from_string(curr_vap.mac);
```
arnout: This change is unrelated to the rest, and it has limited merit (it's not necessary for this commit). If you want to keep it, it should be a separate commit.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -258,20 +259,6 @@ void ap_manager_thread::after_select(bool timeout)
...
-            } else if (low_filter) {
-                current_type = bwl::eBackhaulVapType::BH_VAP_TYPE_5G_SECONDARY;
-            } else {
-                current_type = bwl::eBackhaulVapType::BH_VAP_TYPE_5G;
-            }
-
-            backhaul_vaps_list.remove_if([&](const bwl::backhaul_vap_list_element_t &element) {
-                return (element.type != current_type);
-            });
-
```
arnout: This change is not explained in the commit message, and should *definitely* be a separate commit.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,6 +934,96 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
     return true;
 }
 
+static bool set_vap_multiap_mode(std::vector<std::string> &vap_hostapd_config,
```
arnout: This should have been done in 3 separate commits:
1. Refactor set_vap_multiap_mode into a separate function without changing 
2. Add mesh_mode
3. Add set_multiap_wps

However, that's probably difficult to do after the fact, and it's mostly useful for reviewing, not for longer-term history. So OK.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,6 +934,96 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    std::string bssid, ifname;
+    if (!hostapd_config_get_value(vap_hostapd_config, "bssid", bssid)) {
+        LOG(ERROR) << "Failed to get VAP BSSID";
+        return false;
+    }
+    if (!hostapd_config_get_value(vap_hostapd_config, "bss", ifname)) {
+        LOG(ERROR) << "Failed to get VAP ifname (bss)";
+        return false;
+    }
+    backhaul_vaps_list.remove_if([&](const bwl::backhaul_vap_list_element_t &element) {
```
arnout: The same should also be done for the VAPs that are disabled later. So maybe it would be better to just clear backhaul_vaps_list completely and add to them again?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,6 +934,96 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+        return (element.bssid == beerocks::net::network_utils::mac_from_string(bssid));
+    });
+    bool backhaul  = !!(bss_type & WSC::eWscVendorExtSubelementBssType::BACKHAUL_BSS);
+    bool fronthaul = !!(bss_type & WSC::eWscVendorExtSubelementBssType::FRONTHAUL_BSS);
+    bool teardown  = (bss_type == WSC::eWscVendorExtSubelementBssType::TEARDOWN);
```
arnout: If this is true, it's an error. Teardown is handled differently in the bwl interface (by giving an empty bssinfo list).

Maybe it would be better to split up the fronthaul and backhaul as separate booleans in the bssinfo structure, then we don't create confusion about the teardown and bSTA bits, and we don't have to "parse" the bss_type in every bwl implementation. But that idea got shot down when I originally proposed it, so I guess it will still get shot down now :-)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,6 +934,96 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    // Not supporting hybrid mode for in mesh mode (TODO - move to hybrid mode after
+    // https://github.com/prplfoundation/prplMesh/issues/889)
+    if (fronthaul && backhaul) {
+        LOG(ERROR) << "Not supporting hybrid VAP";
+        return false;
+    }
+    hostapd_config_set_value(vap_hostapd_config, "mesh_mode", backhaul ? "bAP" : "fAP");
+    hostapd_config_set_value(vap_hostapd_config, "sFourAddrMode", backhaul ? "1" : "");
+    hostapd_config_set_value(vap_hostapd_config, "max_num_sta", backhaul ? "1" : "");
+    hostapd_config_set_value(vap_hostapd_config, "start_disabled", teardown ? "1" : "0");
```
arnout: So this shouldn't be done - the teardown elements are just missing from bssinfos.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,6 +934,96 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    LOG(DEBUG) << "Configuring VAP " << ifname << ": bssid=" << bssid
+               << ", backhaul=" << beerocks::string_utils::bool_str(backhaul)
+               << ", teardown=" << beerocks::string_utils::bool_str(teardown);
+    // BSS type (backhaul, fronthaul or both)
+    // Use Intel Mesh-Mode (upstream Multi-AP functionality not supported by Intel):
+    // Not supporting hybrid mode for in mesh mode (TODO - move to hybrid mode after
+    // https://github.com/prplfoundation/prplMesh/issues/889)
+    if (fronthaul && backhaul) {
+        LOG(ERROR) << "Not supporting hybrid VAP";
+        return false;
```
arnout, line 960: Since the return value is ignored by the caller, we should at least still set the config to something. Probably disable the VAP is best.

Alternatively, we should handle the return value in the caller.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,6 +934,96 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+        backhaul_vaps_list.push_back(std::move(element));
+    }
+    return true;
+}
+
+bool ap_wlan_hal_dwpal::set_multiap_wps(
+    std::map<std::string, std::vector<std::string>> &hostapd_config_vaps)
+{
+    // Find first fAP VAP
+    auto fap = std::find_if(hostapd_config_vaps.begin(), hostapd_config_vaps.end(),
```
arnout: I think it's easier to understand if this search is done while we're anyway iterating over bss_conf_list in update_vap_credentials(). Add
```
   auto first_fronthaul_vap = hostapd_config_vaps.end();
```
and set it in the loop.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,6 +934,96 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+                                return (std::find(vap.second.begin(), vap.second.end(),
+                                                  "mesh_mode=fAP") != vap.second.end());
+                            });
+    if (fap == hostapd_config_vaps.end()) {
+        LOG(DEBUG) << "No fronthaul configured, wps will not be set.";
+        return false;
+    }
+
+    // Find first bAP VAP
+    auto bap = std::find_if(hostapd_config_vaps.begin(), hostapd_config_vaps.end(),
```
arnout: Can't we use backhaul_vaps_list for this?

If not, same comment as above.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -164,16 +134,6 @@ void ap_manager_thread::ap_manager_config(ap_manager_conf_t &conf)
...
     low_filter   = conf.iface_filter_low;
 
-    // initialize backhaul_vaps_list
-    int num_of_elements = sizeof(beerocks_message::sPlatformSettings::backhaul_vaps_bssid) /
-                          sizeof(beerocks_message::sPlatformSettings::backhaul_vaps_bssid[0]);
-    for (int i = 0; i < num_of_elements; i++) {
-        bwl::backhaul_vap_list_element_t element = {};
-        element.bssid = network_utils::mac_to_string(conf.backhaul_vaps_bssid[i]);
-        element.type  = bwl::eBackhaulVapType(i % 3);
-        backhaul_vaps_list.push_back(std::move(element));
```
arnout, line 159: I wonder if backhaul_vaps_list couldn't be removed entirely if we instead replace VAPElement with sBssInfoConf everywhere. Then the front/backhaul info is in there directly.

But no need to do that in this PR.

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
```diff
@@ -164,12 +164,14 @@ class ap_wlan_hal : public virtual base_wlan_hal {
...
      * @param [in] List of wifi credentials.
+     * @param [in/out] List of backhaul vaps to update
      *
      * @return true on success or false on error.
      */
 
     virtual bool
-    update_vap_credentials(std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list) = 0;
+    update_vap_credentials(std::list<son::wireless_utils::sBssInfoConf> &bss_info_conf_list,
+                           std::list<bwl::backhaul_vap_list_element_t> &backhaul_vaps_list) = 0;
```
arnout: I don't like this. The BWL should be an abstraction of the interface to wireless. It's not right that we rely on a bwl function to update an internal structure of ap_manager_thread. So unless the bwl actually does something with that backhaul vaps list that the ap_manager_thread can't do itself, I don't think it's right.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,8 +934,99 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    // multiap wps in the fronthaul on which WPS PBC will run.
+    std::string backhaul_ssid, backhaul_passphrase;
+    if (!hostapd_config_get_value(bap->second, "ssid", backhaul_ssid)) {
+        LOG(ERROR) << "Failed to get " << bap->first << " ssid";
+        return false;
+    }
+    if (!hostapd_config_get_value(bap->second, "wpa_passphrase", backhaul_passphrase)) {
+        LOG(ERROR) << "Failed to get " << bap->first << "wpa_passphrase";
+        return false;
+    }
```
arnout: If we get the appropriate sBssInfo structures while iterating from them, we don't need this hack of parsing the config again, and this function can never fail. Would be much better.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -562,6 +564,22 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
         reply_ucc(eWfaCaStatus::COMPLETE, params["parameter"] + "," + value);
         break;
     }
+    case eWfaCaCommand::START_WPS_REGISTRATION: {
+        std::unordered_map<std::string, std::string> params{};
```
arnout: I believe the band and wpsconfigmethod arguments are mandatory, so we can just as well add them here.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -562,6 +564,22 @@ void beerocks_ucc_listener::handle_wfa_ca_command(const std::string &command)
...
+        if (!parse_params(cmd_tokens_vec, params, err_string)) {
+            LOG(ERROR) << err_string;
+            reply_ucc(eWfaCaStatus::INVALID, err_string);
+            break;
+        }
+        if (!handle_start_wps_registration(params, err_string)) {
```
arnout: ... and pass them explicitly as arguments here instead of the generic "params".

Note that I would also do the string->enum conversion here.

(nitpick)

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -389,6 +389,7 @@ enum eFreqType {
     FREQ_5G     = 1,
     FREQ_58G    = 2,
     FREQ_24G_5G = 3,
+    FREQ_60G    = 4,
```
arnout: This PR doesn't end up using it, so why add it?

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -201,6 +201,10 @@ cACTION_APMANAGER_WIFI_CREDENTIALS_UPDATE_REQUEST:
     _type: WSC::cConfigData
     _length: [ wifi_credentials_size ]
 
+cACTION_APMANAGER_START_WPS_PBC_REQUEST:
+  _type: class
+  bssid: sMacAddr
```
arnout: I believe we just agreed to use interface names everywhere, so not ifindex, mac address, radio_uid or vap_id. In my view, "everywhere" includes the internal messages.

In other words, I think we should use hostap_iface here instead.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -47,6 +47,22 @@ class backhaul_manager : public btl::transport_socket_thread {
 
     // For agent_ucc_listener
     const std::string &get_controller_bridge_mac() { return controller_bridge_mac; }
+    /**
+     * @brief get radio mac (ruid) of registered slave based on frequency type
+     * 
+     * @param freq radio frequency to search
+     * @return radio mac of the found slave if found, otherwise empty string
+     */
+    const std::string freq_to_radio_mac(enum eFreqType freq) const;
```
arnout: I really don't like this approach that is taken in master::db of using a MAC address as an identifier. We're doing C++ here, not tcl, so not everything is a string - we have actual *objects* to manipulate. In other words, you can just return a `std::shared_ptr<sRadioInfo>`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -47,6 +47,22 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+
+    /**
+     * @brief start WPS PBC
+     * 
+     * @param radio_mac radio mac of the radio on which to start WPS
+     * @param bssid bssid on which to start WPS
+     * @return true on success, false on failure
+     */
+    bool start_wps_pbc(const sMacAddr &radio_mac, const sMacAddr &bssid);
```
arnout: (nitpick) it makes more sense to me to put the logic of which VAP to start WPS on *inside* this function.

And in keeping with my previous comment, the radio should be passed as a `const sRadioInfo&`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3592,4 +3592,45 @@ bool backhaul_manager::add_ap_he_capabilities(const sRadioInfo &radio_info)
     return true;
 }
 
+const std::string backhaul_manager::freq_to_radio_mac(enum eFreqType freq) const
```
arnout: (nitpick) although the type is called eFreqType, I think it makes more sense to call the variable `band` instead of `freq`.

Also, AFAIK, our coding style prescribes to drop the `enum`.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -186,8 +199,50 @@ bool agent_ucc_listener::handle_start_wps_registration(
         err_string = "invalid WpsConfigMethod, supporting only PBC";
         return false;
     }
-    // TODO trigger WPS PBC on the selected band
-    return true;
+
+    auto freq      = band_to_freq(band->second);
+    auto radio_mac = m_backhaul_manager_ctx.freq_to_radio_mac(freq);
+    if (radio_mac.empty()) {
+        err_string = "Failed to get radio mac for " + band->second;
```
arnout: (nitpick) the CAPI errors should be relevant to the outside world. In this case:
```suggestion
        err_string = "Failed to get radio for " + band->second;
```

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -186,8 +199,50 @@ bool agent_ucc_listener::handle_start_wps_registration(
...
+        return false;
+    }
+
+    auto it = vaps_map.find(radio_mac);
+    if (it == vaps_map.end()) {
+        err_string = "radio_mac " + radio_mac + " not found";
```
arnout: ... and here the same message, but a LOG(ERROR) with the internal message.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -186,8 +199,50 @@ bool agent_ucc_listener::handle_start_wps_registration(
...
+        }
+    }
+    if (fronthaul_bssid == network_utils::ZERO_MAC) {
+        err_string = "No fAP found for radio_mac " + radio_mac;
+        return false;
+    }
+    if (backhaul_bssid == network_utils::ZERO_MAC) {
+        err_string = "No bAP found for radio_mac " + radio_mac;
+        return false;
+    }
```
arnout: We shouldn't do this. If PBC is triggered, it is *simultaneously* for fronthaul and backhaul. Therefore, we should trigger it even if no backhaul AP is configured.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -186,8 +199,50 @@ bool agent_ucc_listener::handle_start_wps_registration(
...
+        if (fronthaul_bssid == network_utils::ZERO_MAC && !vap.backhaul_vap) {
+            fronthaul_bssid = vap.mac;
+        }
+        if (backhaul_bssid == network_utils::ZERO_MAC && vap.backhaul_vap) {
+            backhaul_bssid = vap.mac;
+        }
+    }
+    if (fronthaul_bssid == network_utils::ZERO_MAC) {
+        err_string = "No fAP found for radio_mac " + radio_mac;
+        return false;
```
arnout: This is a tricky one... It is possible that you have fronthaul only configured on one radio and backhaul only on another. Therefore, it is certainly not an error if no fronthaul is configured on this radio. Clearly we should not start WPS on this radio in that case, so we still have to return, but without error.

That said, it *is* an error if no fronthaul is configured on *any* radio. So to be completely correct, we should loop over all radios (of that band) and trigger an error if there is no fronthaul BSS on any of them.

But that is going a bit far, so let's just say that we don't consider it an error if no fronthaul is configured.

arnout (changes requested):
There is an inconsistency now: the logic to select the VAP for WPS is duplicated. When configuring it, the logic is inside the bwl implementation, but when WPS is triggered, the logic is in backhaul_manager (well, actually in agent_ucc_listener at the moment but I don't think that that's right). It would make more sense if the logic was completely in backhaul_manager. But that is not possible at the moment, because the configuration is handled by the son_slave while the WPS handling is done by backhaul_manager... So, it would actually make more sense if all the logic in either son_slave or ap_manager (depends on whether ap_manager is supposed to be nothing more than a bridge to bwl, or if it really should contain logic - I believe it should be just a bridge).

In a similar vein, the backhaul credentials should really be set on *all* radios which have WPS enabled, not only on the radios that have fronthaul (and therefore WPS) configured. So that would mean that when backhaul credentials are set, they have to be distributed to all slaves. And it probably means we need a separate bwl function for setting backhaul credentials for wps. That, however, is way out of scope for this PR. So let's create an issue for it.

Since I'll want to think a bit more about these two main issues, I'm marking it as changes requested.

The rest of my comments are smaller coding style issues.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: obsolete since we removed the backhaul vaps list

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: removed backhaul vaps list completely.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: Removed entirely

common/beerocks/bwl/include/bwl/ap_wlan_hal.h
tomereli: Done

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Done

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
tomereli: Not passing params anymore, I'm passing the band

common/beerocks/bcl/include/bcl/beerocks_defines.h
tomereli: removed

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
tomereli: removed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
tomereli: I think its easier to use the radio_mac instead of the sRadioInfo.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
tomereli: Refactored.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Done, but as a new commit.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: You are correct, fixed.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Removed

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Correct, fixed to check the return value in the caller.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Completely different now, no backhaul vaps list.

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
tomereli: Correct. Changed.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -55,9 +55,17 @@ struct WiFiChannel {
     bool radar_affected;
 };
 
+enum eVapType : uint8_t {
```
arnout: (nitpick) why not enum class?

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -55,9 +55,17 @@ struct WiFiChannel {
...
+    VAP_TYPE_FRONTHAUL = 0,
+    VAP_TYPE_BACKHAUL  = 1,
+    VAP_TYPE_HYBRID    = 3,
+    VAP_TYPE_INVALID   = 4,
```
arnout: (nitpick) R2 will add more VAP types (R2-only and R1-only backhaul). So maybe we should use a bigger number already.

Why do we have values here anyway?

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -825,6 +825,49 @@ bool base_wlan_hal_dwpal::refresh_radio_info()
...
+
+    for (uint8_t i = 0; i < (sizeof(numOfValidArgs) / sizeof(size_t)); i++) {
+        if (numOfValidArgs[i] == 0) {
+            LOG(ERROR) << "Failed reading parsed parameter " << (int)i << " ==> Abort";
+            return eVapType::VAP_TYPE_INVALID;
+        }
+    }
+
+    std::string mesh_mode = std::string(mode);
+    if (mesh_mode.find("bAP") != std::string::npos) {
```
arnout, line 872: Shouldn't this be *exactly* bAP? I.e., `mesh_mode=foobAPbar` should not match, right?

In other words:
```suggestion
    if (mesh_mode == "bAP") {
```

Note: I know that the rest of this file uses this find pattern, but I don't think that's a reason to continue doing this weird and incorrect thing.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -90,58 +88,17 @@ copy_vaps_info(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
...
             LOG(DEBUG) << "vap_id = " << int(vap_id) << ", mac = " << curr_vap.mac
-                       << ", ssid = " << curr_vap.ssid
-                       << ", backhaul_vap = " << beerocks::string_utils::bool_str(bVAP);
+                       << ", ssid = " << curr_vap.ssid << ", type = " << curr_vap.type;
 
             // Copy the VAP MAC and SSID
             vaps[i].mac = network_utils::mac_from_string(curr_vap.mac);
             mapf::utils::copy_string(vaps[i].ssid, curr_vap.ssid.c_str(),
                                      beerocks::message::WIFI_SSID_MAX_LENGTH);
-            vaps[i].backhaul_vap = bVAP;
```
arnout, line 93: Shouldn't there be something here like
```
     vaps[i].backhaul_vap = (curr_vap.type == BACKHAUL || curr_vap.type == HYBRID);
```
?

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -90,58 +88,17 @@ copy_vaps_info(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
...
         }
     }
 }
 
-static bool enable_backhaul_vap(
```
arnout, line 98: Your original commit message had a very good explanation of why this could be removed. Please keep that explanation.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -1681,7 +1577,7 @@ bool ap_manager_thread::handle_ap_enabled(int vap_id)
     notification->vap_info().mac = network_utils::mac_from_string(vap_info.mac);
     mapf::utils::copy_string(notification->vap_info().ssid, vap_info.ssid.c_str(),
                              beerocks::message::WIFI_SSID_MAX_LENGTH);
-    notification->vap_info().backhaul_vap = bVAP;
+    notification->vap_info().backhaul_vap = (vap_info.type == bwl::eVapType::VAP_TYPE_BACKHAUL);
```
arnout:
```suggestion
    notification->vap_info().backhaul_vap = (vap_info.type == bwl::eVapType::VAP_TYPE_BACKHAUL || vap_info.type == bwl::eVapType::VAP_TYPE_HYBRID);
```

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -55,9 +55,17 @@ struct WiFiChannel {
     bool radar_affected;
 };
 
+enum eVapType : uint8_t {
+    VAP_TYPE_FRONTHAUL = 0,
```
arnout: This will always be fully scoped, so the VAP_TYPE_ part isn't needed.

Also, doesn't our coding style say this should be lower_snake_case? This file actually uses a mix of UPPER_SNAKE_CASE and UpperCamelCase, so I don't know what is the best now...

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -95,10 +95,6 @@ sPlatformSettings:
   backhaul_max_vaps: uint8_t 
   backhaul_network_enabled: uint8_t 
   backhaul_preferred_radio_band: uint8_t 
-  backhaul_vaps_bssid:
```
arnout, line 98: Please explain in the commit message why this can be removed as well.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -863,7 +868,8 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
             bss_info_conf_list.push_back(bss_info_conf);
         }
 
-        ap_wlan_hal->update_vap_credentials(bss_info_conf_list);
+        ap_wlan_hal->update_vap_credentials(bss_info_conf_list, backhaul_wps_ssid,
+                                            backhaul_wps_passphrase);
```
arnout, line 869: (nitpick) I would have done this change (adding backhaul credentials to update_vap_credentials()) in a separate commit from the mesh_mode change, but you've already done so much rebasing that I don't want to ask you to split this off too :-)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,8 +934,53 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+{
+    std::string bssid, ifname;
+    if (!hostapd_config_get_value(vap_hostapd_config, "bssid", bssid)) {
+        LOG(ERROR) << "Failed to get VAP BSSID";
+        return false;
+    }
+    if (!hostapd_config_get_value(vap_hostapd_config, "bss", ifname)) {
+        LOG(ERROR) << "Failed to get VAP ifname (bss)";
+        return false;
+    }
```
arnout, line 949: This is only used for a debug message, I don't think it's worth keeping it just for that.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -934,8 +934,53 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
+    if (fronthaul && backhaul) {
+        LOG(ERROR) << "Not supporting hybrid VAP";
+        return false;
+    }
+    hostapd_config_set_value(vap_hostapd_config, "mesh_mode", backhaul ? "bAP" : "fAP");
+    hostapd_config_set_value(vap_hostapd_config, "sFourAddrMode", backhaul ? "1" : "");
+    hostapd_config_set_value(vap_hostapd_config, "max_num_sta", backhaul ? "1" : "");
+    hostapd_config_set_value(vap_hostapd_config, "start_disabled", teardown ? "1" : "0");
+    if (fronthaul) {
+        // Oddly enough, multi_ap_backhaul_wpa_passphrase has to be quoted, while wpa_passphrase does not...
```
arnout, line 968: I wonder, shouldn't we also set wps_mode=2 in the fronthaul case and wps_mode=0 in the backhaul case? I.e., make sure WPS is only enabled on fronthaul.

I know that for upstream hostapd this wlll be needed, but I'm not sure for Intel hostapd.

agent/src/beerocks/slave/agent_ucc_listener.cpp
```diff
@@ -169,6 +169,41 @@ bool agent_ucc_listener::send_cmdu_to_destination(ieee1905_1::CmduMessageTx &cmd
...
+                                                       std::string &err_string)
+{
+    auto freq      = band_to_freq(band);
+    auto radio_mac = m_backhaul_manager_ctx.freq_to_radio_mac(freq);
+    if (radio_mac.empty()) {
+        err_string = "Failed to get radio for " + band;
+        return false;
+    }
+
+    auto it = vaps_map.find(radio_mac);
```
arnout, line 195: Ah, this is why radio_mac is easier than sRadioInfo... It's the vaps_map that is broken :-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3606,4 +3606,42 @@ bool backhaul_manager::add_ap_he_capabilities(const sRadioInfo &radio_info)
...
+        return false;
+    }
+
+    auto it = std::find_if(
+        slaves_sockets.begin(), slaves_sockets.end(),
+        [&](std::shared_ptr<sRadioInfo> slave) { return slave->radio_mac == radio_mac; });
+    if (it == slaves_sockets.end()) {
+        LOG(ERROR) << "couldn't find slave for radio mac " << radio_mac;
+        return false;
+    }
```
arnout, line 3883: Just to be clear: this is why I think using the radio_mac as the identifier is wrong: it requires looking up the sRadioInfo again, while we already did that. Also, if we ever want to start pbc on all radios, then we will not be able to reuse this function.

That said, I'm OK with merging it as is now. I guess the private struct didn't work out after all.

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -875,7 +875,13 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
     }
     case beerocks_message::ACTION_APMANAGER_START_WPS_PBC_REQUEST: {
         LOG(DEBUG) << "Got ACTION_APMANAGER_START_WPS_PBC_REQUEST";
-        //TODO
+        auto request =
+            beerocks_header->addClass<beerocks_message::cACTION_APMANAGER_START_WPS_PBC_REQUEST>();
+        if (!request) {
+            LOG(ERROR) << "addClass cACTION_APMANAGER_START_WPS_PBC_REQUEST failed";
+            return false;
+        }
```
arnout: It's a bit silly that we actually include this empty class, but OK.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1675,6 +1675,34 @@ bool ap_wlan_hal_dwpal::generate_connected_clients_events()
     return ret;
 }
 
+bool ap_wlan_hal_dwpal::start_wps_pbc()
+{
+    refresh_vaps_info(beerocks::IFACE_RADIO_ID);
```
arnout, line 1661: Why is it needed to refresh vaps info? Shouldn't it be up to date at this point?

Please add a comment to explain this.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1675,6 +1675,34 @@ bool ap_wlan_hal_dwpal::generate_connected_clients_events()
...
+    // start wps on the first fronthaul vap
+    auto it = std::find_if(m_radio_info.available_vaps.begin(), m_radio_info.available_vaps.end(),
+                           [&](const std::pair<int, VAPElement> &vap) -> bool {
+                               if (vap.second.ssid.empty() ||
+                                   vap.second.type != eVapType::VAP_TYPE_FRONTHAUL) {
+                                   return false;
+                               }
+                               return true;
```
arnout:
```suggestion
                               return (!vap.second.ssid.empty() &&
                                   vap.second.type == eVapType::VAP_TYPE_FRONTHAUL);
```

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -1675,6 +1675,34 @@ bool ap_wlan_hal_dwpal::generate_connected_clients_events()
     return ret;
 }
 
+bool ap_wlan_hal_dwpal::start_wps_pbc()
+{
+    refresh_vaps_info(beerocks::IFACE_RADIO_ID);
+    // start wps on the first fronthaul vap
```
arnout: I think it's worth to explain in the commit message that specifically dwpal doesn't support WPS on multiple interfaces, so we take the first fronthaul interface.

common/beerocks/bwl/nl80211/ap_wlan_hal_nl80211.cpp
```diff
@@ -501,6 +501,12 @@ bool ap_wlan_hal_nl80211::generate_connected_clients_events()
     return true;
 }
 
+bool ap_wlan_hal_nl80211::start_wps_pbc()
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED!";
+    return true;
```
arnout:
```suggestion
    LOG(DEBUG) << "Start WPS PBC on interface " << m_radio_info.iface_name;
    std::string cmd = "WPS_PBC";
    if (!wpa_ctrl_send_msg(cmd)) {
        LOG(ERROR) << "start_wps_pbc() failed!";
        return false;
    }
    return true;
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -659,12 +659,23 @@ bool master_thread::autoconfig_wsc_add_m2(WSC::m1 &m1,
...
-        cfg.bss_type    = bss_info_conf->bss_type;
+        // Since currently we do not support hybrid mode, the options are either fronthaul,
+        // backhaul, or teardown.
+        if (bss_info_conf->fronthaul) {
+            cfg.bss_type = WSC::eWscVendorExtSubelementBssType::FRONTHAUL_BSS;
+        } else if (bss_info_conf->backhaul) {
+            cfg.bss_type = WSC::eWscVendorExtSubelementBssType::BACKHAUL_BSS;
+        } else {
+            cfg.bss_type = WSC::eWscVendorExtSubelementBssType::TEARDOWN;
+        }
```
arnout: NACK this!
- We *never* should set TEARDOWN here, teardown is handled in the else branch below.
- We should simply set the hybrid mode if hybrid mode is requested. If we don't do this, controller certification *will* fail.

```suggestion
        cfg.bss_type = 0;
        if (bss_info_conf->fronthaul) {
            cfg.bss_type |= WSC::eWscVendorExtSubelementBssType::FRONTHAUL_BSS;
        }
        if (bss_info_conf->backhaul) {
            cfg.bss_type |= WSC::eWscVendorExtSubelementBssType::BACKHAUL_BSS;
        }
```

ci/certification/netgear-rax40.yml
```diff
@@ -98,6 +98,26 @@ MAP-4.4.1_ETH:netgear-rax40:
     - when: manual
       allow_failure: true
 
+MAP-4.2.3_ETH_FH5GL:netgear-rax40:
```
arnout: These are already there, we just merged a commit that adds all certification tests as jobs. So this commit should be dropped.

tests/capi.py
```diff
@@ -159,6 +159,18 @@ def cmd_reply(self, command: str, verbose: bool = False) -> Dict[str, str]:
             self.send_cmd(command)
             return self.get_reply(verbose)
 
+    def start_wps_registration(self, band: str):
```
arnout:
```suggestion
    def start_wps_registration(self, band: str) -> None:
```

arnout (approved):
One big NACK, one thing which will make gitlab CI fail, and a lot of nitpicks.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
arnout: Actually, since we now have bssinfo splitting fronthaul and backhaul as booleans, it would make sense to do the same in the VapElement.

tomereli:
gitlab failed due to `RECONF` failure in the rax40 - this sometimes happens, we need to restart the RAX40 and re-run the test - @rmelotte.
```

ERROR 14:17:38:926 <1988345080> ap_wlan_hal_dwpal.cpp[1117] --> Autoconfiguration: "RECONF wlan2.0" command to hostapd has failed!
DEBUG 14:17:38:930 <1988345080> base_wlan_hal_dwpal.cpp[501] --> 
cmd failed: RECONF wlan2.1
DEBUG 14:17:38:931 <1988345080> base_wlan_hal_dwpal.cpp[502] --> 
reply: FAIL

ERROR 14:17:38:932 <1988345080> ap_wlan_hal_dwpal.cpp[1117] --> Autoconfiguration: "RECONF wlan2.1" command to hostapd has failed!
DEBUG 14:17:38:936 <1988345080> base_wlan_hal_dwpal.cpp[501] --> 
cmd failed: RECONF wlan2.2
DEBUG 14:17:38:937 <1988345080> base_wlan_hal_dwpal.cpp[502] --> 
reply: FAIL

ERROR 14:17:38:938 <1988345080> ap_wlan_hal_dwpal.cpp[1117] --> Autoconfiguration: "RECONF wlan2.2" command to hostapd has failed!
DEBUG 14:17:38:942 <1988345080> base_wlan_hal_dwpal.cpp[501] --> 
cmd failed: RECONF wlan2.3
DEBUG 14:17:38:942 <1988345080> base_wlan_hal_dwpal.cpp[502] --> 
reply: FAIL
```

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
tomereli: No, actually the value returned from hostapd is `mesh_mode=bAP (1)`, this is why I used `std::find`.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
tomereli: removed altogether, added boolean values to VAPElement.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
tomereli: Done - removed the enum.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: Obsolete

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: Done. Thanks :)

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: Obsolete since we use backhaul and fronthaul boolean values.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
tomereli: Deleted.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
tomereli: I added the original explanation, all the backhaul vaps are not used anymore. We have the backhaul and fronthaul booleans in the VAPElement.

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: You know that when you ask nicely I can't say no, so done :)

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Without the bssid or the interface name there is no much use of the debug message at all, so I prefer to keep them.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: You are absolutely right, I've been meaning to fix this and forgot - thank you for paying attention!
Its called `wps_state` though.

agent/src/beerocks/slave/agent_ucc_listener.cpp
tomereli: Perhaps we can fix it then in a different PR :)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: We we should do is perhaps make the agent_ucc_listener friend class of the backhaul manager, then we can just use it as is even though it is private. What do you think (anyway its for a different PR) ?

agent/src/beerocks/slave/ap_manager_thread.cpp
tomereli: Its leftovers from back when we passed the bssid in the message. It can be removed now.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: I'm actually not sure if its up to date, I don't know when we start_wps_pbc() will be called so its better to be on the safe side. It doesn't happen too often.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
tomereli: Done, updated the commit message + comment.

controller/src/beerocks/master/son_master_thread.cpp
tomereli: You're right, fixed.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -825,6 +825,56 @@ bool base_wlan_hal_dwpal::refresh_radio_info()
...
+    size_t numOfValidArgs[1] = {0}, replyLen = 0;
+    char mode[16]                 = {0};
+    FieldsToParse fieldsToParse[] = {
+        {(void *)mode, &numOfValidArgs[0], DWPAL_STR_PARAM, "mesh_mode=", sizeof(mode)},
+        /* Must be at the end */
+        {NULL, NULL, DWPAL_NUM_OF_PARSING_TYPES, NULL, 0}};
+    std::string cmd = "GET_MESH_MODE " + ifname;
+
+    fronthaul = false;
+    backhaul  = false;
```
vitalybu: I think there's no need to initialize these variables to `false`, since if this method returns `false` the values should be ignored anyway (nitpick)

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -880,10 +930,16 @@ bool base_wlan_hal_dwpal::refresh_vap_info(int vap_id)
...
+        LOG(ERROR) << "Failed to get VAP type";
+        return false;
+    }
 
     // Store the VAP element
     LOG(DEBUG) << "Detected VAP ID (" << vap_id << ") - MAC: " << vapElement.mac
-               << ", SSID: " << vapElement.ssid;
+               << ", SSID: " << vapElement.ssid
+               << ", fronthaul: " << beerocks::string_utils::bool_str(vapElement.fronthaul)
+               << ", backhaul: " << beerocks::string_utils::bool_str(vapElement.backhaul);
```
vitalybu, line 952: Why not use [`std::boolalpha`](https://en.cppreference.com/w/cpp/io/manip/boolalpha) instead?
```suggestion
               << ", fronthaul: " << std::boolalpha << vapElement.fronthaul
               << ", backhaul: " << vapElement.backhaul;
```

agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -877,7 +877,7 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
     }
     case beerocks_message::ACTION_APMANAGER_START_WPS_PBC_REQUEST: {
         LOG(DEBUG) << "Got ACTION_APMANAGER_START_WPS_PBC_REQUEST";
-        //TODO
+        ap_wlan_hal->start_wps_pbc();
```
vitalybu: It can fail, so better check the return value

vitalybu (approved):

tomereli:
Added the don't merge tag since it looks like with this PR, although WPS works, CTT Agent 2 fails with autoconfig. 
Passing PR (gathering statistics):
https://github.com/prplfoundation/prplMesh/pull/1096

tomereli:
Merging manually since the CI is stuck. After analysis it appears that when 4.2.3 tests fail, any message from the controller is not received by the DUT at all. Will open a separate issue on this.

PR #1170, opened 2020-04-20
Feature/prplwrt single codebase

### LINKED ISSUE #763: [TASK] Single codebase for prplWrt
## Description

Currently, we need a different version of prplwrt (actually, the iwlwav feed) when we build for RAX40 than what is used for B1300 and Turris Omnia.

RAX40: https://git.prpl.dev/prplmesh/iwlwav.git^e9ee6b109764dbb91b6cb399ccc0f78e5b04e260
others: https://git.prpl.dev/prplmesh/iwlwav.git^06a0126d5fb53b1d65bad90757a5f9f5f77419ca

(these commits come from different branches).

The failure is because dwpal gets built even for non-Intel targets, and dwpal depends on packages in the intel feed which is not included for non-Intel targets.

## Exit criteria

Both prplMesh variants can be built using the same prplWrt version, and feed versions (except for the intel feed).

---

This issue was first reported on Trello, but didn't have a corresponding issue on Github yet: https://trello.com/c/4CvH9Ov9/61-single-code-base-for-prplwrt

rmelotte:
Use the same prplMesh package definition for both variants (nl80211 and dwpal).

- Move the prplMesh package from [iwlwav](https://git.prpl.dev/prplmesh/iwlwav/) to [feed-prpl](https://git.prpl.dev/prplmesh/feed-prpl/).
- For the nl80211 variant:
  - Update the prplMesh package to use wlan1 instead of wlan2 for the second radio
  - Add hostapd-utils as a dependency (required by prplMesh at boot).
- Make the docker image build log smaller.
- Exclude the toolchain and SDK for the rax40.

The final step before this PR can be merged is to update the commit hashes once the 2 following MR have been merged:
https://git.prpl.dev/prplmesh/iwlwav/-/merge_requests/13/diffs
https://git.prpl.dev/prplmesh/feed-prpl/-/merge_requests/1/diffs?commit_id=531c27d86a912a12d649320418817db281534eb0

Fixes #763

### DIFF
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -94,7 +94,7 @@ RUN cp feeds.conf.default feeds.conf \
     fi ;\
     make defconfig \
     && cat .config \
-    && make -j$(nproc) V=sc \
+    && make -j$(nproc) \
     && make package/feeds/prpl/prplmesh/clean
     # note that the result from diffconfig.sh with a minimal
     # configuration has the 3 CONFIG_TARGET items we set here, but NOT
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -37,6 +37,8 @@ ARG PRPLMESH_VARIANT
 ARG PRPL_FEED
 # optional: intel feed to use.
 ARG INTEL_FEED
+# optional: iwlwav feed to use.
+ARG IWLWAV_FEED
 
 # Target to build for (CONFIG_TARGET_ will be prepended).
 # Example: TARGET_SYSTEM=mvebu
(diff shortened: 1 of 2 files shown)


arnout (approved):
I've made some comments in the prpl-feed, otherwise LGTM.

Excellent commit message on the first commit BTW!

tools/docker/builder/openwrt/Dockerfile
```diff
@@ -90,7 +90,7 @@ RUN cp feeds.conf.default feeds.conf \
     && echo "CONFIG_TARGET_${TARGET_SYSTEM}_${SUBTARGET}_${TARGET_PROFILE}=y" >> .config \
     && echo "CONFIG_PACKAGE_prplmesh${PRPLMESH_VARIANT}=y" >> .config \
     && if [ "$TARGET_PROFILE" = "DEVICE_NETGEAR_RAX40" ] ; then \
-        cat feeds/intel/rax40.profile >> .config ;\
+        grep -v -E 'CONFIG_SDK|CONFIG_MAKE_TOOLCHAIN' feeds/intel/rax40.profile >> .config ;\
```
tomereli, line 93: Regarding this I think we should just remove these from the rax40.profile instead of hacking it away here. Better to sync with John why these are exported and if we can remove them before we workaround this in our Dockerfile.

tomereli (approved):
One major comment but nothing blocking so approving.

tools/docker/builder/openwrt/Dockerfile
rmelotte: I asked John about it, but he might not reply immediately so I'll merge this, we can still change it later if it turns out we can remove it from the profile file.

PR #1172, opened 2020-04-21
Feature/implement certification 4.7.9 ii

RanRegev:
This PR replaces PR #1114
This PR contains the entire agent flow to pass certification test 4.7.9 with ETH backhaul.
It added two tlvs (beacon query and response), it takes care for sending the request only to the correct radio, and it handles the response.
The result of the full flow is a beacon measurements response, 1905 format, received at the controller. The data from the message is NOT stored or processed in any way.


agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -503,11 +503,17 @@ bool monitor_thread::update_sta_stats()
...
         if (!mon_wlan_hal->update_stations_stats(vap_node->get_iface(), sta_mac,
                                                  sta_stats.hal_stats)) {
-            LOG(ERROR) << "Failed updating STA (" << sta_mac << ") statistics!";
+            if (stat_state_changed) {
+                LOG(ERROR) << "Failed updating STA (" << sta_mac << ") statistics!";
+                stat_state_changed = false;
+            }
             return false;
         }
+        stat_state_changed = true;
```
vitalybu: I think this is incorrect.
You prevent all error messages not just for the specific `sta_mac`.
Also, it's probably important to see in the logs that something doesn't work and we're unable to update the station statistics.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -100,6 +101,35 @@ namespace beerocks {
 #define FSM_IS_IN_STATE(eState) (m_eFSMState == EState::eState)
 #define FSM_CURR_STATE_STR s_arrStates[int(m_eFSMState)]
 
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Helper Functions /////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+template <class SlavesContainer, class MAC>
```
vitalybu, line 108: Any specific reason for templating this?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -100,6 +101,35 @@ namespace beerocks {
...
+            LOG(DEBUG) << "looking into radio info for (hostap_iface): " << slave->hostap_iface;
+
+            for (const auto &client : slave->associated_clients_map) {
+                LOG(DEBUG) << "client: " << client.first;
+
+                if (client.second.find(mac) != client.second.end()) {
+                    return slave;
+                }
+            }
+        }
```
vitalybu, line 127: Aligned to our coding standards:

```suggestion
        if (!slave) {
            continue;
        }
        
        LOG(DEBUG) << "looking into radio info for (hostap_iface): " << slave->hostap_iface;

        for (const auto &client : slave->associated_clients_map) {
            LOG(DEBUG) << "client: " << client.first;

            if (client.second.find(mac) != client.second.end()) {
                return slave;
            }
        }
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -339,24 +369,25 @@ bool backhaul_manager::init()
...
-            ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE,
-            ieee1905_1::eMessageType::LINK_METRIC_QUERY_MESSAGE,
             ieee1905_1::eMessageType::COMBINED_INFRASTRUCTURE_METRICS_MESSAGE,
-            ieee1905_1::eMessageType::ACK_MESSAGE,
+            ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE,
+            ieee1905_1::eMessageType::LINK_METRIC_QUERY_MESSAGE,
+            ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE,
+            ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE,
+            ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE,
+            ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE,
```
vitalybu, line 390: If you decide to sort the list alphabetically, please do it in a separate commit

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1650,7 +1681,8 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
         // Handle the CMDU message. If the message was processed locally
         // (by the Backhaul Manager), this function will return 'true'.
         // Otherwise, it should be forwarded to the slaves.
-        if (handle_1905_1_message(cmdu_rx, src_mac)) {
+        Socket *destinationAP = nullptr;
```
vitalybu, line 1698:
```suggestion
        Socket *target_slave = nullptr;
```

**BUT**, I think it's not required at all.
You can just forward the CMDU from your handler function to the correct slave and return `true` (so it won't be forwarded to all slaves here).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2890,6 +2935,74 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+    // extract the desired STA mac
+    auto tlvBeaconMetricsQuery = cmdu_rx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+    if (!tlvBeaconMetricsQuery) {
+        LOG(ERROR) << "handle_1905_beacon_metrics_query should handle only tlvBeaconMetrics, but "
+                      "got something else: 0x"
+                   << std::hex << (uint16_t)cmdu_rx.getMessageType();
+        return false;
+    }
+
+    const sMacAddr &requiredMac = tlvBeaconMetricsQuery->associated_sta_mac();
```
vitalybu, line 2967: We use Snake Case not Camel Case, so the name of the variable (as an example) should change to:
```suggestion
    const sMacAddr &required_mac = tlvBeaconMetricsQuery->associated_sta_mac();
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: The original thought was that we have mac as a struct and mac as a string, so by specializing the template with string it would be transparent to the user. This thought lead to another thought, since it is a helper function, loosely coupled to the backhaul (the return type is a backhaul inner type) why not free the container to be whatever it is, as long as it is forward-iterable.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: done.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: I don't like this either, for either template argument.
- MAC should always be sMacAddr. If there's a string anywhere, call mac_from_string or (better) refactor so that it's not a string.
- Using a template to work around private is super-ugly IMO. I see two better ways to handle this:
    1. Make it a private function of backhaul_manager. Yes, I agree that the class is already overloaded, but still...
    2. Make sRadioInfo a public class (which would imply renaming it to RadioInfo, and moving it to a separate file). This opens opportunities to add more members to RadioInfo, which is what you wanted I think. While you're at it, you'd probably want to add a class RadioList or something like that that abstracts the container.

I leave it up to you to choose which one is better.

RanRegev:
Replaced by #1274 and #1273

PR #1173, opened 2020-04-21
update iterator after erase

### LINKED ISSUE #1164: [BUG] run-tests (test_flows.py) occasionaly fails on master (client_capability_query and client_steering_dummy)
It was first noticed by @RanRegev  [on his branch](https://gitlab.com/prpl-foundation/prplMesh/-/jobs/513883964), and today it happened on master.
Logs are available here:
https://gitlab.com/prpl-foundation/prplMesh/-/jobs/515794163

We know it doesn't happen for every run. Here is a second (passing) run on the same commit as the one above: 
https://gitlab.com/prpl-foundation/prplMesh/-/jobs/515886112

arnout:
When an element is erased, any iterator pointing to it is invalidated. 
Thus, `foo.erase(it)` invalidates `it` and `it` should no longer be used.
Fortunately, erase() returns a new iterator that points to the next
element. Therefore, update `it` with the returned value.

In most places, this is not needed, since the iterator is no longer used 
after the erase. However, it never hurts to update the iterator, and it 
avoids problems when the code is changed afterwards. Also, it's easier to
not forget to update the iterator if the same pattern is used everywhere.

### DIFF
--- a/agent/src/beerocks/monitor/monitor_db.cpp
+++ b/agent/src/beerocks/monitor/monitor_db.cpp
@@ -334,7 +334,7 @@ void monitor_db::sta_erase(const std::string sta_mac)
     if (it != sta_nodes.end()) {
         auto vap_id = it->second->get_vap_id();
         delete it->second;
-        sta_nodes.erase(it);
+        it            = sta_nodes.erase(it);
         auto vap_node = vap_get_by_id(vap_id);
         if (vap_node) {
             vap_node->sta_count_dec();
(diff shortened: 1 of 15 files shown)


arnout:
Second commit is optional of course. Instead of replacing everything, I could have replaced only the 3 or 4 instances that really are wrong.

I'm not sure if this really fixes #1164, but since it's randomly occurring, the only way to find out is to try it.

rmelotte (approved):

vitalybu (approved):

PR #1178, opened 2020-04-21
Collect wireless link metrics and add them to Link Metrics Response

### LINKED ISSUE #792: [TASK] Link metric collection for WiFi links
## Description

The agent should collect link metrics for WiFi links.

## Exit criteria

- A mechanism exists to collect link metrics from the WiFi interfaces registered in the agent.
- LinkMetricsResponse TLVs are filled in with the correct data.
- Test flow with dummy link metrics.

## Detailed description

See [TASK] Link metric collection for Ethernet links #390

To obtain link metrics programmatically for wireless interfaces, NL80211_CMD_GET_STATION netlink command should be issued. This was initially planned to be done through DWPAL but issuing commands other than NL80211_CMD_VENDOR is not currently supported in DWPAL (Send standard NL80211 commands using DWPAL #782). Thus this feature has finally been implemented in all BWL flavors through NL80211 (Feature/add support for nl80211 standard commands to BWL #900)

mariomaz:
See #792


RanRegev (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -369,6 +404,24 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+     *
+     * @param[in] reporter_al_mac 1905.1 AL MAC address of the device that transmits the response message.
+     * @param[in] link_interface Connecting interface in this device.
+     * @param[in] link_neighbor Neighbor connected to interface.
+     * @param[in] link_metrics Metrics information associated to the link between the local interface and the neighbor's interface.
+     * @param[in] link_metrics_type The link metrics type requested: TX, RX or both.
+     *
+     * @return True on success and false otherwise.
+     */
+    bool add_link_metrics(const sMacAddr &reporter_al_mac, const sLinkInterface &link_interface,
```
vitalybu, line 434: Any reason to not make this method a static function in the cpp file?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3493,18 +3394,98 @@ bool backhaul_manager::get_media_type(const std::string &interface_name,
...
+    }
+
+    if (ieee1905_1::eMediaTypeGroup::IEEE_802_11 == media_type_group) {
+        return std::make_unique<ieee802_11_link_metrics_collector>();
+    }
+
+    LOG(ERROR) << "Unable to create link metrics collector for interface '"
+               << link_interface.iface_name << "' (unsupported media type " << std::hex
+               << (int)media_type << ")";
+    return nullptr;
```
vitalybu, line 3458:
```suggestion
    
    return nullptr;
```

vitalybu (approved):

common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -111,6 +111,17 @@ class network_utils {
...
+
+    /**
+     * @brief Gets the interface name of the network interface with given MAC address.
+     *
+     * @param[in] mac MAC address of the network interface.
+     * @param[out] iface On success, name of the network interface. On error, empty string.
+     *
+     * @return True on success and false otherwise.
+     */
+    static bool linux_iface_get_name(const std::string &mac, std::string &iface);
```
arnout: NACK for any change that adds a function or field using a string for a MAC address. Please use sMacAddr.

That said, I would really prefer that we keep a database of interfaces in the agent, where each interface has the name, MAC address, type and possibly other info, and where things like getting link metrics is done by subclassing per type. But that's probably a refactoring for M3.

agent/src/beerocks/slave/link_metrics/ieee802_11_link_metrics_collector.cpp
```diff
@@ -0,0 +1,43 @@
...
+    link_metrics.transmitter.mac_throughput_capacity_mbps = sta_info.tx_bitrate_100kbps / 10;
+    // Note: For simplicity, link availability is set to "100% of the time"
+    link_metrics.transmitter.link_availability = 100;
+    link_metrics.transmitter.phy_rate_mbps     = UINT16_MAX;
+    // Note: rx errors cannot be obtained with NL80211_CMD_GET_STATION command.
+    link_metrics.receiver.packet_errors    = 0;
+    link_metrics.receiver.packets_received = sta_info.rx_packets;
+    // Note: according to <linux/nl80211.h>, the value of NL80211_STA_INFO_SIGNAL
+    // attribute (i.e.: signal strength of last received PPDU) is expressed in dBm.
+    // However IEEE 1905.1 standard requires RSSI to be expressed in dB.
```
arnout, line 37: Actually, it says pretty explicitly that it's expressed in dB with an implementation-defined baseline. So mW is as good a baseline as any other.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -279,6 +279,34 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+     * @brief Neighbor 1905.1 device which connects to an interface in this device.
+     *
+     * A neighbor is defined by its 1905.1 AL MAC address and the MAC address of the interface in
+     * the neighbor that connects to this device.
+     */
+    struct sLinkNeighbor {
+        sMacAddr al_mac =
+            beerocks::net::network_utils::ZERO_MAC; /**< The MAC address of the 1905.1 AL. */
+        sMacAddr iface_mac =
+            beerocks::net::network_utils::ZERO_MAC; /**< The MAC address of the interface. */
```
arnout, line 322: I haven't looked yet at how this will be used, but intuitively it would make more sense to me to have a reference to a sLinkInterface here.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3507,6 +3508,27 @@ bool backhaul_manager::get_media_type(const std::string &interface_name,
     return result;
 }
 
+std::unique_ptr<link_metrics_collector>
+backhaul_manager::create_link_metrics_collector(const sLinkInterface &link_interface) const
+{
+    ieee1905_1::eMediaType media_type = link_interface.media_type;
+    ieee1905_1::eMediaTypeGroup media_type_group =
+        static_cast<ieee1905_1::eMediaTypeGroup>(media_type >> 8);
```
arnout, line 3444: I believe this should work as well, and is [what CppCoreGuidelines advises](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-construct):
```suggestion
    ieee1905_1::eMediaTypeGroup media_type_group {media_type >> 8};
```

For consistency, the other initializer should probably use `{}` instead of `=` as well, then.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -149,157 +149,68 @@ static const std::vector<std::tuple<eFreqType, eWiFiBandwidth, ieee1905_1::eMedi
...
+ *
+ * @param[in] iface_name Name of the network interface.
+ * @param[out] iface_mac MAC address of the network interface on success and ZERO_MAC on error.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_iface_mac(const std::string &iface_name, sMacAddr &iface_mac)
+{
+    std::string mac;
+    if (network_utils::linux_iface_get_mac(iface_name, mac)) {
```
arnout, line 184: I would prefer to add this function to network_utils, i.e. add an overload of linux_ifac_get_mac that takes sMacAddr as the second argument rather than a string. And probably that function should be called from the existing linux_iface_get_mac rather than the reverse.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -149,157 +149,68 @@ static const std::vector<std::tuple<eFreqType, eWiFiBandwidth, ieee1905_1::eMedi
...
-        return false;
+/**
+ * @brief Gets the name of the interface with given MAC address.
+ *
+ * @param[in] iface_mac MAC address of the network interface.
+ * @param[out] iface_name Name of the network interface on success and empty string on error.
+ *
+ * @return True on success and false otherwise.
+ */
+static bool get_iface_name(const sMacAddr &iface_mac, std::string &iface_name)
```
arnout, line 203: This should also be in network_utils.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2674,9 +2585,9 @@ bool backhaul_manager::handle_1905_link_metric_query(ieee1905_1::CmduMessageRx &
 
     /**
      * Get the list of neighbor links from the topology database.
-     * Neighbor links are grouped by the interface that connects to them.
+     * Neighbors are grouped by the interface that connects to them.
      */
-    std::map<std::string, std::vector<sMacAddr>> neighbor_links_map;
+    std::map<sLinkInterface, std::vector<sLinkNeighbor>> neighbor_links_map;
```
arnout, line 2607: Remind me: why is this not an unordered_map?

(the only reason you need an operator< in sLinkInterface is because of this ordered map)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
mariomaz: There are a few other class methods to add TLVs to response CMDU message already, so I tried to keep it coherent

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2704,25 +2618,32 @@ bool backhaul_manager::handle_1905_link_metric_query(ieee1905_1::CmduMessageRx &
...
-            std::string local_interface_name = entry.first;
-            std::vector<sMacAddr> neighbors  = entry.second;
-
-            ieee1905_1::eMediaTypeGroup media_type_group = ieee1905_1::eMediaTypeGroup::UNKNOWN;
-            if (!get_media_type_group(local_interface_name, media_type_group)) {
-                LOG(ERROR) << "Unable to compute media type group for interface "
-                           << local_interface_name;
-                return false;
-            }
+            sLinkInterface interface             = entry.first;
```
```suggestion
            auto interface             = entry.first;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2704,25 +2618,32 @@ bool backhaul_manager::handle_1905_link_metric_query(ieee1905_1::CmduMessageRx &
...
+            std::vector<sLinkNeighbor> neighbors = entry.second;
```
```suggestion
            const auto &neighbors = entry.second;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2704,25 +2618,32 @@ bool backhaul_manager::handle_1905_link_metric_query(ieee1905_1::CmduMessageRx &
...
+                continue;
             }
 
             for (const auto &neighbor : neighbors) {
-                if (!add_link_metrics(local_interface_name, media_type, reporter_al_mac, neighbor,
-                                      link_metrics_type, cmdu_tx)) {
+
+                LOG(TRACE) << "Getting link metrics for interface " << interface.iface_name
+                           << " (MediaType = " << std::hex << (int)interface.media_type
+                           << ") and neighbor " << network_utils::mac_to_string(neighbor.iface_mac);
```
```suggestion
                           << ") and neighbor " << neighbor.iface_mac;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2812,7 +2733,7 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
             return true;
         }
     } else {
-        LOG(ERROR) << "tlvSupportedRole missing - ignoring autconfig search message";
+        LOG(ERROR) << "tlvSupportedRole missing - ignoring autoconfig response message";
```
arnout: I don't really like unrelated changes in the same commit - they surprise me.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3531,16 +3442,71 @@ backhaul_manager::create_link_metrics_collector(const sLinkInterface &link_inter
...
+
+    // Also include a link for each associated client
+    for (const auto &slave : slaves_sockets) {
+
+        for (const auto &associated_clients_entry : slave->associated_clients_map) {
+            auto bssid              = associated_clients_entry.first;
+            auto associated_clients = associated_clients_entry.second;
+
+            sLinkInterface interface;
+            if (get_iface_name(bssid, interface.iface_name)) {
```
arnout: (nitpick) The usual pattern is
```suggestion
            if (!get_iface_name(bssid, interface.iface_name)) {
                return false;
            }
            ... // with less indentation
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3531,16 +3442,71 @@ backhaul_manager::create_link_metrics_collector(const sLinkInterface &link_inter
...
+                // discovery message, which will give us the neighbor interface and AL MAC addresses.
+                neighbor.iface_mac = associated_client.first;
+                neighbor.al_mac    = neighbor.iface_mac;
+
+                if ((neighbor_mac_filter == network_utils::ZERO_MAC) ||
+                    (neighbor_mac_filter == neighbor.al_mac)) {
+                    neighbor_links_map[interface].push_back(neighbor);
+                }
+            }
+        }
```
arnout, line 3531: Since we're building up interface per interface, we don't actually need a map for neighbor_links_map. It could just as well be a vector-of-vectors. Of course, to make this a little bit manageable, that should be factored into structures or even better, classes.

But let's not do that for this PR.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3628,4 +3594,88 @@ bool backhaul_manager::add_ap_he_capabilities(const sRadioInfo &radio_info)
...
+            LOG(ERROR) << "Failed accessing interface_pair_info";
+            return false;
+        }
+        auto interfacePairInfo                      = std::get<1>(interface_pair_info);
+        interfacePairInfo.rc_interface_mac          = link_interface.iface_mac;
+        interfacePairInfo.neighbor_interface_mac    = link_neighbor.iface_mac;
+        interfacePairInfo.link_metric_info.intfType = link_interface.media_type;
+        // TODO
+        //Indicates whether or not the 1905.1 link includes one or more IEEE 802.1 bridges
+        //eIEEE802_1BridgeFlag IEEE802_1BridgeFlag;
```
arnout: We do have to set that to something!

In fact, since we currently consider the bridge as the only interface on the local device for the wired backhaul, we should set the bridge flag to true for the wired link.

arnout (approved):
I have a big comment about the use of string instead of sMacAddr, but that is trivial to solve, and the rest are nitpicks or suggestions for future refactorings. So approving.

I did remove the ready for merge tag though.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

common/beerocks/bcl/include/bcl/network/network_utils.h
mariomaz: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

agent/src/beerocks/slave/link_metrics/ieee802_11_link_metrics_collector.cpp
mariomaz: I can't see that clarification in my copy of the IEEE 1905.1 standard nor in Amendment 1. Which documentation are you referring to?
Shall I remove my comment?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
mariomaz: We need both the AL MAC address and the MAC address of the interface in the neighbor

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Sorry, I cannot do that since I get a compilation error:
```
../agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp:3401:61: error: invalid conversion from 'int' to 'ieee1905_1::eMediaTypeGroup' [-fpermissive]
     ieee1905_1::eMediaTypeGroup media_type_group{media_type >> 8};
                                                  ~~~~~~~~~~~^~~~
```

Since a cast is required anyway, I'll leave it as it was

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: To create an unordered_map with a custom class key we need a hash function and the class must implement operator == for handling collisions. To create a map with a custom class key we must implement operator < to provide a sorting rule. Then I think it's simpler to create a map than an unordered_map.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Reverted

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: I thought we had to implement IEEE 802.1 bridge discovery to be able to set this field appropriately and then set it to true if "the 1905.1 neighbor device is connected on this
particular interface through one or more IEEE 802.1 bridges". Are you sure this flag must be set to true for the wired link just because the wired backahul link is the bridge interface?

agent/src/beerocks/slave/link_metrics/ieee802_11_link_metrics_collector.cpp
arnout: Grmbl I'm sure I read that somewhere but now I can't find it. Just leave things as is.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
arnout: Oh, I misinterpreted the meaning of iface_mac. Never mind then.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: OK.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: I don't think it matters much either way, because I don't think anyone will use this. Probably leaving the TODO is the best option anyway. 

My idea was that this flag could be used to "know" that the topology is inaccurate (because there's a bridge in-between that doesn't participate in 1905.1, we don't have an accurate picture of the exact links). Since that is exactly the situation in our case (e.g. the controller may not be a direct neighbour, but still we treat it as one), I feel it is "safer" to set the bridge flag.

PR #1188, opened 2020-04-23
[hotfix] docker runner fixes

arnout:
Various small fixes to make the runner image work on my laptop.

Note that the entrypoint update requires an update of the docker images on dockerhub.
However, since nobody else seems to be bothered by this, there's no urgency for doing the update.

### DIFF
--- a/tools/docker/runner/entrypoint.sh
+++ b/tools/docker/runner/entrypoint.sh
@@ -13,7 +13,7 @@ run() {
 
 # Use the ip address that was allocated by the daemon to this
 # container (from eth0)
-bridge_ip="$(ip addr show dev eth0 | grep inet | awk '{print $2}')"
+bridge_ip="$(ip addr show dev eth0 | awk '/^ *inet / {print $2}')"
 
 run ip link add          br-lan   type bridge
 run ip link add          wlan0    type dummy
--- a/tools/docker/runner/entrypoint.sh
+++ b/tools/docker/runner/entrypoint.sh
@@ -11,6 +11,7 @@ run() {
     "$@" || exit $?
 }
 
+
 # Use the ip address that was allocated by the daemon to this
 # container (from eth0)
 bridge_ip="$(ip addr show dev eth0 | awk '/^ *inet / {print $2}')"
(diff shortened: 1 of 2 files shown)


tools/docker/runner/entrypoint.sh
```diff
@@ -24,6 +25,7 @@ run ip link set      dev wlan2    master br-lan
 run ip address flush dev eth0
 run ip link set      dev wlan0    up
 run ip link set      dev wlan2    up
+run ip link set      dev br-lan   addr "$bridge_mac"
```
rmelotte, line 34: For the Turris Omnia, we have this in the wiki:
```
The backhaul interface must have a different MAC address than the bridge interface. 
```

Does it not apply to the dummy variant?

rmelotte (approved):
LGTM.
Why is it ok in this case to use the same MAC address for both the bridge and eth0 though?

tools/docker/runner/entrypoint.sh
arnout: That is an extremely good observation! I missed the fact that originally we had a fixed MAC address that was different from eth0.

After you removed the explicit setting of the addresses, br-lan receives the same address as eth0 in CI:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: br-lan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 02:42:ac:13:00:03 brd ff:ff:ff:ff:ff:ff
3: wlan0: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-lan state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether d6:82:42:41:62:97 brd ff:ff:ff:ff:ff:ff
4: wlan2: <BROADCAST,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-lan state UNKNOWN mode DEFAULT group default qlen 1000
    link/ether c2:39:82:4b:44:b1 brd ff:ff:ff:ff:ff:ff
188015: eth0@if188016: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master br-lan state UP mode DEFAULT group default 
    link/ether 02:42:ac:13:00:03 brd ff:ff:ff:ff:ff:ff link-netnsid 0
```

Apparently, it now works by accident... I think it might have something to do with the way that docker generates MAC addresses: they are sequential, so repeater1's address is the gateway address + 1. This happens to be the same as the fake address that is generated for gateway's ETHERNET node, and apparently the two problems compensate for each other...

But this might very well explain why test_flows still doesn't work for me. I'll look further into it.

tools/docker/runner/entrypoint.sh
```diff
@@ -11,13 +11,24 @@ run() {
     "$@" || exit $?
```
rmelotte, line 11: The commit message says "we could use any of the techniques
in [1]" but there is no "[1]".
Did you forget to add it?

rmelotte (approved):
Weird, when I add a bridge interface interactively on my laptop I get different MAC addresses...

VladyslavTupikin (approved):

abelog (approved):
LGTM

arnout:
> Weird, when I add a bridge interface interactively on my laptop I get different MAC addresses...

Indeed, when I do it inside the container, `ip link set      dev eth0     master br-lan` makes br-lan get the MAC address of eth0. But when I do it directly on my laptop, this doesn't happen...

However, this makes me realize that br-lan already has a random address, so I don't need to create the dummy interface! Updating coming in...

PR #1189, opened 2020-04-23
tools: deploy_ipk.sh: restart network before deploy

tomereli:
The hostapd configuration files are created during initialization by the
netifd in /var/run/hostapd-phyX.conf, before prplMesh is
started.
During autoconfig, prplmesh overrides these files, and runs RECONF on
the hostapd ctrl interface for changes to take effect.

Since the target device is used to test pull requests, it may be that
these files get corrupted after a faulty PR has run. The PR might fail,
but the contents of the files will remain and will cause the next run to
also fail, thus failing a potentially valid PR, or even worse, a full
nightly certification run.

In order to prevent this from happening, restart netifd when deploying
prplmesh.

### DIFF
--- a/tools/deploy_ipk.sh
+++ b/tools/deploy_ipk.sh
@@ -25,6 +25,14 @@ deploy() {
     IPK_FILENAME="$(basename "$2")"
     DEST_FOLDER=/tmp/prplmesh_ipks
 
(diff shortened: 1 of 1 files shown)


tomereli:
After syncing with John and Raphael, the decision is to add this instead to the preinst script in the package Makefile, so closing.

PR #1190, opened 2020-04-24
ci: owncloud_definitions.sh: add rootdir

rmelotte:
Since commit "633a4032d50141ded8cde64a79c6f4f848feda0d", $rootdir must
be defined to be able to source tools/functions.sh.

Since it was not, every upload to owncloud on the master branch was
failing with:
ci/owncloud/../../tools/functions.sh: line 10: rootdir: parameter null or not set

owncloud_definitions.sh doesn't make any use of rootdir, we only
define it because we need the functions from functions.sh


arnout (approved):

arnout:
I'm force-merging since it's trivial and annoying for other people.

PR #1192, opened 2020-04-25
Reduce the number of cppcheck-related issues

vitalii-komisarenko:
The list of existing cppcheck issues is almost 700 entries long.
It is hard to find the important ones because there are too many of them.

The purpose of this pull request is to reduce their number to make this list more manageable.
And it is nice to have code that follows code style.


common/beerocks/bcl/source/beerocks_string_utils.cpp
```diff
@@ -114,7 +114,6 @@ std::string string_utils::int_to_hex_string(const unsigned int integer,
     // for example: int_to_hex_string(255, 4) -> "00ff"
     //              int_to_hex_string(255, 1) -> "ff"
 
-    std::string return_string;
```
arnout, line 129: I really liked your first few commits: they are really clear and though they are large, it's trivial to check that they do the right thing.

Here, however, it's a lot more difficult, because there are two different kind of changes in the same commit, and the commit message doesn't describe them. So I'd like this to be split into two commits: the first one that does the string replacements, and the second one that fixes this issue (which I assume is an unused variable).

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -139,7 +139,7 @@ static std::string get_sta_iface_from_hostap_iface(const std::string &hostap_ifa
     return sta_iface;
 }
 
-static void fill_son_slave_config(beerocks::config_file::sConfigSlave &beerocks_slave_conf,
+static void fill_son_slave_config(const beerocks::config_file::sConfigSlave &beerocks_slave_conf,
```
arnout, line 142: This is also a const-ref fix so please reuse the same commit message as in the first two.

agent/src/beerocks/slave/beerocks_slave_main.cpp
```diff
@@ -206,11 +206,10 @@ static void son_slave_watchdog(const std::string &beerocks_slave_temp_path,
     }
 }
 
-static int system_hang_test(beerocks::config_file::sConfigSlave &beerocks_slave_conf, int argc,
-                            char *argv[])
+static int system_hang_test(const beerocks::config_file::sConfigSlave &beerocks_slave_conf,
+                            int argc, char *argv[])
 {
     std::string name = std::string("system_hang_test");
-    std::ofstream versionfile;
```
arnout, line 213: This one again is separate.

You could combine a few unused variable fixes in a single commit, if you want to reduce the number of commits.

common/beerocks/bcl/source/beerocks_utils.cpp
```diff
@@ -215,7 +215,7 @@ std::string utils::convert_channel_ext_above_to_string(bool channel_ext_above_se
 void utils::merge_list(std::vector<uint8_t> &ret, std::vector<uint8_t> &list)
 {
     std::vector<uint8_t>::iterator it;
-    for (it = list.begin(); it != list.end(); it++) {
+    for (it = list.begin(); it != list.end(); ++it) {
```
arnout: For this one, a better fix is to remove the function entirely since it's not used.

Or failing that, use ranged loop:
```
    for (element : list) {
```

arnout (approved):
Excellent PR! I like it when things are improved!

I made a few comments about commit splitting, but it's not worth spending the time to actually do that - it's more a suggestion for the future.

The only thing I would like to change is the merge_list function.

common/beerocks/bcl/source/beerocks_utils.cpp
RanRegev: For my understanding this function might not do what you want:
- it does not copies zeros from the source (list), but it does not verify there are no zeros in the destination (ret). So you might end with zeros at ret.
if order does not matter, the schema of the function should be:
```
ret.insert(list.begin(), list.end());
std::sort(ret);
std::remove(ret.begin(),ret.end(),0);
std::unique(ret.begin(),ret.end());
```
Using merely algorithms from std. No raw loops.
example reference: https://youtu.be/W2tWOdzgXHA

agent/src/beerocks/monitor/monitor_db.h
```diff
@@ -23,7 +23,10 @@ namespace son {
 ////////////////////////////////////////////
 class monitor_sta_node {
 public:
-    monitor_sta_node(const int8_t vap_id_, const std::string mac_) { vap_id = vap_id_, mac = mac_; }
+    monitor_sta_node(const int8_t vap_id_, const std::string &mac_)
+        : vap_id(vap_id_), mac(mac_), m_sta_stats()
```
RanRegev, line 27: No need to call the default constructor explicitly. I am referring to the explicit call to `m_sta_stats()`
references -
standard: http://eel.is/c++draft/class.base.init#9
godbolt: https://godbolt.org/z/4YEwWS
Also apply for more default initialized members along the code.

vitalii-komisarenko: cppcheck actually complains about uninitialized members in the constructor:
`'/builds/prpl-foundation/prplMesh/agent/src/beerocks/monitor/monitor_db.h: warning: Member variable 'monitor_sta_node::m_sta_stats' is not initialized in the constructor. [uninitMemberVar]    monitor_sta_node(const int8_t vap_id_, const std::string mac_) { vap_id = vap_id_, mac = mac_; }`

common/beerocks/bcl/source/beerocks_utils.cpp
vitalii-komisarenko: @RanRegev , I prefer @arnout 's approach, not only because it is easier, but also because this function is too bad to keep:
* it is unused
* use case too niche to happen often
* code style does not correspond to current guidelines
* the name is misleading (should be something like `merge_non_zero_values_from_list`)

agent/src/beerocks/monitor/monitor_db.h
rmelotte: Actually, I think cppcheck is right here.
`m_sta_stats` is of type SStaStats, which comes from `common/beerocks/bwl/include/bwl/mon_wlan_hal_types.h` (and there is an other, different definition in `agent/src/beerocks/monitor/monitor_db.h`).

The struct is defined as follow:
```
struct SStaStats {
    float rx_rssi_watt;
    uint8_t rx_rssi_watt_samples_cnt;
    float rx_snr_watt;
    uint8_t rx_snr_watt_samples_cnt;
    // int8_t   rx_rssi_prev=beerocks::RSSI_INVALID;
    // int8_t   rx_rssi_curr=beerocks::RSSI_INVALID;
    uint16_t tx_phy_rate_100kb;
    // uint16_t tx_phy_rate_100kb_avg;
    // uint16_t tx_phy_rate_100kb_min;
    // uint16_t tx_phy_rate_100kb_acc;
    uint16_t rx_phy_rate_100kb;
    // uint16_t rx_phy_rate_100kb_avg;
    // uint16_t rx_phy_rate_100kb_min;
    // uint16_t rx_phy_rate_100kb_acc;
    uint64_t tx_bytes_cnt;
    uint64_t rx_bytes_cnt;
    uint64_t tx_packets_cnt;
    uint64_t rx_packets_cnt;
    uint32_t tx_packets;
    uint32_t tx_bytes;
    uint32_t rx_packets;
    uint32_t rx_bytes;
    uint32_t retrans_count;
    // uint8_t  tx_load_percent_curr=0;
    // uint8_t  tx_load_percent_prev=0;
    // uint8_t  rx_load_percent_curr=0;
    // uint8_t  rx_load_percent_prev=0;
};
```

If you don't initialize m_sta_stats explicitly, all of these will be default-initialized, so they will have a random value.

I adapted your example on godbolt: https://godbolt.org/z/LWmccS

rmelotte (approved):
It solves so many cppcheck issues that Github does not load the diff by default.
Nice! :-)

agent/src/beerocks/monitor/monitor_db.h
RanRegev: yes, I you are correct.
No reason to default initialize this struct unless specifically requested.
My mistake.
standard: http://eel.is/c++draft/dcl.init#7.3

Issue #1158, opened 2020-04-16
[TASK] Add BML to test operational status


LiorAmram:
## Description

Currently, prplmesh_utils.sh status returns an operational status determined by parsing the logs.
This creates 2 problems:
1. Any bug in the logs (for example, truncated logs as a result of early roll) might deliver a wrong result.
2. Operational status might be updated in DB and not be reflected in the status check.

Proposed Solution:
beerocks_cli command to test operational status in real-time.
The command will trigger a topology query that will update the DB value, and then BML will check this value to get an updated result.

PR #1194

## Detailed description

![fPDHIyCm58NV-ol2dbhOP3xuO1nZtIXbsAspdGz3vP9beHJJLfCKzDzxku7DWZB4bzB2ETyzzwIPQSEKQIlfZQPaBuHg-Ai6BajwzaL6d3pNWtqbeH06cZUXc2cRcaaoOOMe3J095UFJWNTkSw7u0AHGwc00SvWMHKnAeNoc2Xty3tLaN6cZohe8yVLf_7A9sArdWvGeIcs40hxtkVaOzYD](https://user-images.githubusercontent.com/43096199/79459730-674ff180-7ffc-11ea-94e3-e1c288627e81.png)

tomereli:
Looking good to me. @arnout any comment before starting implementation?

arnout:
What is not clear from the UML: will every topology response trigger an ACTION_BML_TOPOLOGY_RESPONSE or only if it was triggered by BML to begin with? (I vote for the former)

Also not clear is which parameters the ACTION_BML_TOPOLOGY_RESPONSE will have.

arnout:
(Otherwise LGTM of course)

PR #1194, opened 2020-04-27
add bml to test operational status

kantera800:
This PR introduces new API to retrieve radio operational status from a device in real time.
Issue:
#1158

format:
beerocks_cli bml_get_device_operational_radios <al_mac>

sequence:
- register bml listener to topology events
- send topology query
- wait for response (blocking) using beerocks::promise (RESPONSE_TIMEOUT 5sec) 
- unregister bml listener from topology events
- copy topology response data from shared memory.
- parse returned data and print message to console in the same format as used in prplmesh_utils.sh status.

 Next phase would be to use this API in prplmesh_utils.sh status instead of parsing the logs.


controller/src/beerocks/master/tasks/bml_task.h
```diff
@@ -67,6 +67,12 @@ class bml_task : public task {
         uint8_t cac_completed;
     };
 
+    struct operation_status_update_event {
+        uint8_t backhaul_connection_status;
```
tomereli: If you get the event - than why do you need backhaul connection status? Its obviously working...

controller/src/beerocks/master/tasks/bml_task.h
```diff
@@ -83,6 +89,7 @@ class bml_task : public task {
         ACS_START_EVENT_AVAILABLE,
         CSA_NOTIFICATION_EVENT_AVAILABLE,
         CAC_STATUS_CHANGED_NOTIFICATION_EVENT_AVAILABLE,
+        OPERATIONAL_STATUS_UPDATE,
```
tomereli: IMO we should call TOPOLOGY_RESPONSE, the BML triggers topology query and waits for topology response, the name operational is just for the command.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_bml.yaml
```diff
@@ -20,6 +20,16 @@ cACTION_BML_PING_RESPONSE:
 
 cACTION_BML_NW_MAP_REQUEST:
   _type: class
+  
+cACTION_BML_OPERATIONAL_STATUS_RESPONSE:
+  _type: class
+  buffer_size:
+    _type: uint32_t
+    _length_var: True
+  buffer:
```
tomereli: Why buffer and not actual response parameters?

controller/src/beerocks/master/tasks/bml_task.h
kantera800: good point

controller/src/beerocks/master/tasks/bml_task.h
kantera800: done

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_bml.yaml
kantera800: done

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1852,6 +1852,45 @@ int bml_internal::register_nw_map_update_cb(BML_NW_MAP_QUERY_CB pCB)
...
 }
 
+
+int bml_internal::register_topology_discovery_query()
+{
+    auto request =
+        message_com::create_vs_message<beerocks_message::cACTION_BML_REGISTER_TOPOLOGY_QUERY>(
+            cmdu_tx);
+
+    if (request == nullptr) {
```
morantr:
```cpp
if (!request) {
    // code
}
```
Please apply it on the other function as well.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1852,6 +1852,53 @@ int bml_internal::register_nw_map_update_cb(BML_NW_MAP_QUERY_CB pCB)
...
+int bml_internal::device_oper_radios_query(BML_DEVICE_DATA *device_data)
+{
+    // Initialize the promise for receiving the response
+    beerocks::promise<bool> prmDeviceDataGet;
+    m_prmDeviceDataGet = &prmDeviceDataGet;
+    int iOpTimeout     = RESPONSE_TIMEOUT; // Default timeout
+
+    beerocks_message::sDeviceData DeviceData;
+    m_device_data         = &DeviceData;
+    m_device_data->al_mac = network_utils::mac_from_string(device_data->al_mac);
```
morantr: need to add nullptr check before accessing input argument `*device_data`.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1852,6 +1852,53 @@ int bml_internal::register_nw_map_update_cb(BML_NW_MAP_QUERY_CB pCB)
...
+    trigger_topology_discovery(device_data->al_mac);
+
+    int iRet = BML_RET_OK;
+
+    if (!prmDeviceDataGet.wait_for(iOpTimeout)) {
+        LOG(WARNING) << "Timeout while waiting for topology query response...";
+        iRet = -BML_RET_TIMEOUT;
+    }
+
+    unregister_topology_discovery_query();
```
morantr: Why did you implement a register/unregister mechanism for a simple request-response operation?
The event registration mechanism is relevant if you wish to receive events without any specific request. I didn't see that you added option to register to such event on the bml_cli.
So, the previous commit should be dropped IMO and all other code that related to this event registration.

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -905,6 +909,14 @@ int cli_bml::bml_connection_map_caller(int numOfArgs)
         return connection_map();
 }
 
+int cli_bml::bml_get_device_operational_radios_caller(int numOfArgs)
+{
+    if (numOfArgs != 1)
+        return -1;
+    else
+        return get_device_operational_radios(args.stringArgs[0]);
```
morantr:
```cpp
if (numOfArgs != 1) {
   return -1;
}
return get_device_operational_radios(args.stringArgs[0]);
```

controller/src/beerocks/cli/beerocks_cli_bml.cpp
```diff
@@ -1351,6 +1363,28 @@ int cli_bml::connection_map()
...
+{
+    BML_DEVICE_DATA device_data = {};
+    device_data.al_mac          = al_mac.c_str();
+
+    int ret = bml_device_oper_radios_query(ctx, &device_data);
+
+    // Main agent
+    std::cout << ((ret == BML_RET_OK) ? "OK" : "FAIL") << " Main radio agent operational"
+              << std::endl;
+    for (auto radio : device_data.radios) {
```
morantr: Please use const refernece.
```for (const auto &radio : device_data.radios)```

File: controller/src/beerocks/master/db/db.cpp [hunk 9]
```diff
@@ -2569,7 +2569,8 @@ bool db::is_bml_listener_exist()
 {
     bool listener_exist;
     for (auto it = bml_listeners_sockets.begin(); it < bml_listeners_sockets.end(); it++) {
-        listener_exist = (*it).map_updates || (*it).stats_updates || (*it).events_updates;
+        listener_exist = (*it).map_updates || (*it).stats_updates || (*it).events_updates ||
+                         (*it).topology_updates;
```
morantr: While you change it, can you please break the ugliness sequence and change it to like this:
```cpp
bool db::is_bml_listener_exist()
{
    for (const auto &soc : bml_listeners_sockets) {
        bool listener_exist =
            soc.map_updates || soc.stats_updates || soc.events_updates || soc.topology_updates;
        if (listener_exist) {
            return true;
        }
    }
    return false;
}
```

Although, as I have said it some previous comment, the registration for topology event if you only use it as request-response style.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2494,6 +2494,31 @@ bool db::get_bml_nw_map_update_enable(Socket *sd)
...
+{
+    if (sd) {
+        for (auto it = bml_listeners_sockets.begin(); it < bml_listeners_sockets.end(); it++) {
+            if (sd == (*it).sd) {
+                return (*it).topology_updates;
+            }
+        }
+    }
+    return false;
+}
```
morantr: Please always use this style:
```cpp
if (error) {
    // handle error
}
// rest of the code
```

Also, if you choose to loop with iterators, please use `it->something` and not `(*it).something`.
But I would prefer that you loop with ranged loops:
```cpp
for (auto &soc : bml_listeners_sockets) {
    // Your code here
}
```

controller/src/beerocks/master/tasks/bml_task.cpp
```diff
@@ -193,6 +193,32 @@ void bml_task::handle_event(int event_type, void *obj)
...
+        if (obj) {
+            auto event_obj = (listener_general_register_unregister_event *)obj;
+            TASK_LOG(DEBUG) << "REGISTER_TO_TOPOLOGY_UPDATES event was received";
+            database.add_bml_socket(event_obj->sd);
+            if (!database.set_bml_topology_update_enable(event_obj->sd, true)) {
+                TASK_LOG(DEBUG) << "fail in topology_updates registration";
+            }
+            state = LISTENING;
+        }
+        break;
```
morantr:
```cpp
if (!obj) {
    // Handle error
}

// Rest of the code

File: controller/src/beerocks/master/tasks/bml_task.cpp [hunk 12]
```diff
@@ -297,6 +297,57 @@ void bml_task::handle_event(int event_type, void *obj)
         }
         break;
     }
+    case TOPOLOGY_RESPONSE_UPDATE: {
+        if (obj) {
```
morantr:
```cpp
if (!obj) {
    // handle error
}
// rest of the code

File: controller/src/beerocks/master/tasks/bml_task.cpp [hunk 13]
```diff
@@ -297,6 +297,57 @@ void bml_task::handle_event(int event_type, void *obj)
...
+            auto event_obj = (topology_response_update_event *)obj;
```
morantr: please use `reinterpret_cast`.

controller/src/beerocks/master/tasks/bml_task.cpp
```diff
@@ -297,6 +297,57 @@ void bml_task::handle_event(int event_type, void *obj)
...
+                                    << " , added to the list";
+                }
+                idx++;
+            }
+
+            if (!topology_response_updates_listeners.empty()) {
+                auto response =
+                    message_com::create_vs_message<beerocks_message::cACTION_BML_TOPOLOGY_RESPONSE>(
+                        cmdu_tx);
+                if (response == nullptr) {
```
morantr:
```cpp
if (!response) {
```

controller/src/beerocks/master/tasks/bml_task.cpp
```diff
@@ -297,6 +297,57 @@ void bml_task::handle_event(int event_type, void *obj)
...
+            while ((sd = database.get_bml_socket_at(idx)) != nullptr) {
+                if (database.get_bml_topology_update_enable(sd)) {
+                    topology_response_updates_listeners.push_back(sd);
+                    TASK_LOG(DEBUG) << "TOPOLOGY_RESPONSE_UPDATE listener found with idx = " << idx
+                                    << " , added to the list";
+                }
+                idx++;
+            }
+
+            if (!topology_response_updates_listeners.empty()) {
```
morantr:
```cpp
if (topology_response_updates_listeners.empty()) {
    break;
}
// Rest of the code

File: controller/src/beerocks/master/tasks/bml_task.cpp [hunk 16]
```diff
@@ -297,6 +297,57 @@ void bml_task::handle_event(int event_type, void *obj)
...
+
+                if (event_obj->radio_interfaces.size() > beerocks::message::DEV_MAX_RADIOS) {
+                    LOG(ERROR) << "number of reported interfaces is higher than expected, "
+                                  "ACTION_BML_TOPOLOGY_RESPONSE will not be sent";
+                    return;
+                }
+
+                size_t i = 0;
+                for (auto iface : event_obj->radio_interfaces) {
+                    auto iface_mac     = iface.mac();
```
morantr: const reference

morantr (approved):
Most of it is styling so approving.

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -67,6 +67,7 @@ enum eStructsConsts {
     CHANNEL_SCAN_LIST_LENGTH             = 8,
     HT_MCS_SET_SIZE                      = 16,
     VHT_MCS_SET_SIZE                     = 32,
+    DEV_MAX_RADIOS                       = 3,
```
tomereli, line 70: Why do we need this? This should be part of the platform, so the correct location for this is not hard coded define but in UCI (nitpick, we can fix it later).

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
```diff
@@ -660,6 +660,23 @@ sDeviceInfo:
...
+    _length: [ "beerocks::message::IFACE_NAME_LENGTH" ]
+  iface_mac: sMacAddr
+  iface_status: beerocks::eNodeState
+
+sDeviceData:
+  _type: struct
+  al_mac: sMacAddr
+  radios:
+    _type: sIfaceInfo
+    _length: [ "beerocks::message::DEV_MAX_RADIOS" ]
```
tomereli, line 676: You should consider declaring this as tlvf class where variable length list is supported. Then there is no need for hard coding max radios.
So:
```
cDeviceData:
  _type: class
  al_mac: sMacAddr
  num_radios: uint8_t
  radios:
    _type: sIface_info
    _length: [ num_radios ]
    _length_var: True

File: controller/src/beerocks/master/tasks/bml_task.h [hunk 19]
```diff
@@ -83,6 +83,9 @@ class bml_task : public task {
         ACS_START_EVENT_AVAILABLE,
```
tomereli, line 97: I think this can be squashed to the previous commit, not much benefit from standing on its own.

controller/src/beerocks/bml/bml_defs.h
```diff
@@ -471,6 +471,25 @@ struct BML_EVENT_CAC_STATUS_CHANGED_NOTIFICATION {
 struct BML_EVENT_BEACON_MEASUREMENT {
     //TBD
 };
+
+/**
+ * Beerocks radio information
+ */
```
tomereli, line 474: Please add proper Doxygen comments explaining each member in this struct.

controller/src/beerocks/bml/internal/bml_internal.h
```diff
@@ -47,6 +47,12 @@ class bml_internal : public beerocks::socket_thread {
     // Query the beerocks master for the network map
     int nw_map_query();
 
+    // Register topology discovery query so bml listener could receive the responce event.
```
tomereli:
```suggestion
    // Register topology discovery query so bml listener could receive the response event.
```

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1852,6 +1852,53 @@ int bml_internal::register_nw_map_update_cb(BML_NW_MAP_QUERY_CB pCB)
     return (BML_RET_OK);
 }
 
+int bml_internal::device_oper_radios_query(BML_DEVICE_DATA *device_data)
+{
+    // Initialize the promise for receiving the response
+    beerocks::promise<bool> prmDeviceDataGet;
+    m_prmDeviceDataGet = &prmDeviceDataGet;
+    int iOpTimeout     = RESPONSE_TIMEOUT; // Default timeout
```
tomereli: Don't use cample case for simple variables, its against the coding style.
```suggestion
    int timeout = RESPONSE_TIMEOUT; // Default timeout
```
(nitpick) since it is already like that in this file..

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1852,6 +1852,53 @@ int bml_internal::register_nw_map_update_cb(BML_NW_MAP_QUERY_CB pCB)
...
+
+    beerocks_message::sDeviceData DeviceData;
+    m_device_data         = &DeviceData;
+    m_device_data->al_mac = network_utils::mac_from_string(device_data->al_mac);
+
+    register_topology_discovery_query();
```
tomereli: The function name is not OK - it should be `register_topology_discovery_response()`, you don't register to a "query".

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1852,6 +1852,53 @@ int bml_internal::register_nw_map_update_cb(BML_NW_MAP_QUERY_CB pCB)
...
+    // device radio status can be retrieved from topology discovery response
+    trigger_topology_discovery(device_data->al_mac);
```
tomereli: And this one should be `trigger_topology_discovery_query(device_data->al_mac)`

controller/src/beerocks/bml/internal/bml_internal.cpp
tomereli: I agree with Moran here. If its possible to avoid extra mechanism and just be notified on each topology response (doesn't really have to be sent from the BML, can be eny response)

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -1852,6 +1852,53 @@ int bml_internal::register_nw_map_update_cb(BML_NW_MAP_QUERY_CB pCB)
...
+    int iRet = BML_RET_OK;
+
+    if (!prmDeviceDataGet.wait_for(iOpTimeout)) {
+        LOG(WARNING) << "Timeout while waiting for topology query response...";
+        iRet = -BML_RET_TIMEOUT;
+    }
+
+    unregister_topology_discovery_query();
+
+    // Clear the credentials member
```
tomereli: This is not credentials member, copy paste bug?

controller/src/beerocks/bml/bml.h
```diff
@@ -109,6 +109,19 @@ int bml_nw_map_register_update_cb(BML_CTX ctx, BML_NW_MAP_QUERY_CB cb);
  */
 int bml_nw_map_query(BML_CTX ctx);
 
+/**
+ * Query the beerocks for the latest radio status on a device
```
tomereli: Don't use "beerocks" keyword. 
Consider changing to 
>Query prplmesh for radio status for a specific agent, blocking.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2494,6 +2494,31 @@ bool db::get_bml_nw_map_update_enable(Socket *sd)
     return false;
 }
 
+bool db::set_bml_topology_update_enable(Socket *sd, bool update_enable)
+{
+    if (sd) {
+        for (auto it = bml_listeners_sockets.begin(); it < bml_listeners_sockets.end(); it++) {
```
tomereli: please use range based loops
```suggestion
        for (const auto &soc : bml_listeners_sockets) {
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2494,6 +2494,31 @@ bool db::get_bml_nw_map_update_enable(Socket *sd)
...
+            }
+        }
+    }
+    return false;
+}
+
+bool db::get_bml_topology_update_enable(Socket *sd)
+{
+    if (sd) {
+        for (auto it = bml_listeners_sockets.begin(); it < bml_listeners_sockets.end(); it++) {
```
tomereli: same here, please use range based loop

controller/src/beerocks/master/tasks/bml_task.h
```diff
@@ -67,6 +68,11 @@ class bml_task : public task {
         uint8_t cac_completed;
     };
 
+    struct topology_response_update_event {
```
tomereli, line 79: This commit doesn't help on its own, just add the code when you use it.

tomereli, line 79: Please add Doxygen comments for every new struct.

controller/src/beerocks/bml/internal/bml_internal.cpp
```diff
@@ -555,6 +555,33 @@ int bml_internal::process_cmdu_header(std::shared_ptr<beerocks_header> beerocks_
...
+            auto response =
+                beerocks_header->addClass<beerocks_message::cACTION_BML_TOPOLOGY_RESPONSE>();
+            if (response == nullptr) {
+                LOG(ERROR) << "addClass cACTION_BML_TOPOLOGY_RESPONSE failed";
+                return BML_RET_OP_FAILED;
+            }
+
+            // Signal any waiting threads
+            if (m_prmDeviceDataGet) {
+                if (m_device_data != nullptr) {
```
tomereli:
```suggestion
                if (m_device_data) {
```

controller/src/beerocks/master/db/network_map.cpp
```diff
@@ -371,9 +371,7 @@ void network_map::send_bml_nodes_statistics_message_to_listeners(
 
             //LOG(DEBUG) << "sending message, last=0";
             // sending to all listeners
-            for (auto it = bml_listeners.begin(); it != bml_listeners.end(); ++it) {
-                message_com::send_cmdu(*it, cmdu_tx);
-            }
+            send_bml_event_to_listeners(cmdu_tx, bml_listeners);
```
tomereli, line 372: Nice!

controller/src/beerocks/master/db/db.cpp
```diff
@@ -964,7 +964,7 @@ std::set<std::string> db::get_node_siblings(const std::string &mac, int type)
 
     auto parent = get_node(n->parent_mac);
     if (!parent) {
-        LOG(ERROR) << "parent for node " << mac << " does not exist";
+        LOG(WARNING) << "parent for node " << mac << " does not exist";
```
tomereli, line 965: Please explain in the commit message why you changed to WARNING. I really don't get what's the difference.. this commit can be dropped I recon.

tomereli (approved):
I have many comments but most are simple enough to fix, so approving at first shot! Nice job!

tomereli:
You forgot the most important thing - update prplmesh_utils.sh to use this new BML to check the status instead of grepping the logs!

controller/src/beerocks/bml/internal/bml_internal.cpp
kantera800: I'm not familiar of any other way to signal bml_cli from son_master_thread (internal controller message)

Also couple of points for your suggestion, 
 1. It sounds wrong to go through the bus to send an internal controller message.
2. topology response sent periodically, do we want to put more load on the CPU  especially when this API will not be used in operational state.
4. currently we only use events to signal bml_task from son_master_thread (ACTION_BML directly not used)
5. we don't have eAction for controller->controller VS message (not much of a problem to add one but there is a reason to it)

common/beerocks/bcl/include/bcl/beerocks_defines.h
kantera800: actually now i see that we already have a define for this BML_NODE_MAX_RADIOS ....
i'll use it instead

controller/src/beerocks/bml/internal/bml_internal.cpp
kantera800: first i wrote it exactly like you propose but then i aligned to `register_nw_map_query_cb` naming.
I'll change it back.

controller/src/beerocks/master/db/db.cpp
kantera800: Its printed every time 1905_topology_response arrives (its a lot or error msg in the log)
The reason for this error is that GW's al_mac  fails to retrieve a parent  (because it doesn't have one).
Since we apply this method on every al_mac  on a neighbor list this specific case can be fixed but i didn't want to do it as part of my PR.
IMO this is a legit usage of the method  and definitely not an error (i can add this explanation to the PR if you want )

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_common.yaml
kantera800: I considered creating beerocks_message_common_classes.yaml and adding it there.
After receiving the message i would still need to use a struct with const array to copy the message (BML_DEVICE_DATA) so not sure there is added value doing it.
Also i'm using sDeviceData in other places.

Prefer to leave it a const array.

controller/src/beerocks/master/tasks/bml_task.h
kantera800: done

controller/src/beerocks/master/tasks/bml_task.cpp
kantera800: I prefer not to add breaks in the middle of the code

PR #1197, opened 2020-04-27
beerocks_ucc_listener: remove unnecessary check

### LINKED ISSUE #380: 4.6.3 Client capability query transmit
The prplMesh agent on Turris Omnia with OpenWrt needs to pass the certification test case 4.6.3 as defined by the Wi-Fi CERTIFIED EasyMesh (tm) Test Plan.

- [ ] #379 Pass flow 4.6.3 test in testbed

CoralMalachi:
### Description

currently, agent_ucc_listener overrides the `validate_destination_alid`
pure virtual method, in order to validate that the destination socket
will always be the controller socket on the agent side.

There are multiple tests which require the agent to send a 1905 message
to another agent, so this check is wrong.

### Changes

Remove validate_destination_alid method completely from the UCC listener
implementation.

### Testbed status

Those changes were checked on the testbed and test 4.6.3 passed.
In addition, tests MAP-4.10.3_ETH_* also pass with this change but require an additional fix in the easymesh_cert repo.

Fixes #380 #379 

Signed-off-by: Tomer Eliyahu<tomer.b.eliyahu@intel.com>


morantr:
> Test 4.6.3 is the first time our agent is requested by the UCC to send a
1905 message to another agent, and not the controller, so this
implementation is not correct anymore.

Does the `another Agent` is a known Agent which we have received Topology Discovery message from it? Because if it does, you can verify that the "dest_alid" is the controller or one of the known neighbors on `m_1905_neighbor_devices`.

tomereli:
Agree with Moran. Also, please follow the contributing.md and add a commit to update the passing agent tests file.
Finally, add <TEST NAME>:netgear-rax40 to the last commit message, this will make the CI add this as a mandatory test.

tomereli (changes requested):
1. As Moran suggested, the correct solution is to make sure the alid exists in one of the known neighbours updated via discovery protocol. So first change this.
2. Please align to the [definition of done](https://github.com/prplfoundation/prplMesh/blob/master/CONTRIBUTING.md#definition-of-done) defined in the CONTRIBUTING.md:
> For each new flow being added, tests/test_flows.py must be updated accordingly to test the new flow, and the corresponding certification test (if any) must be added to tests/certification/passing_agent_tests.txt.
3. If your change fixes a certification test, then the last commit in your PR must mention the passing test in the following format: <test name>:<device> - see [WiKi - Testing in CI](https://github.com/prplfoundation/prplMesh/wiki/Testing-in-CI#certification-tests). This will make sure that your change is added the relevant test as a mandatory step in the CI.

arnout:
> > Test 4.6.3 is the first time our agent is requested by the UCC to send a
> > 1905 message to another agent, and not the controller, so this
> > implementation is not correct anymore.
> 
> Does the `another Agent` is a known Agent which we have received Topology Discovery message from it? Because if it does, you can verify that the "dest_alid" is the controller or one of the known neighbors on `m_1905_neighbor_devices`.

To be honest, I'd remove the validation entirely. If the UCC says we have to send a CMDU to some destination, who are we to claim that that's a destination you're not allowed to send to? I can imagine that we would add tests of corner cases where we want to force sending a CMDU to an unknown agent, e.g. to verify that a security measure works.

That said, I don't care much - we can still remove the check entirely when we need to. But I wouldn't spend time on adding "security" measures to code that isn't even enabled in production.

tomereli:
> > > Test 4.6.3 is the first time our agent is requested by the UCC to send a
> > > 1905 message to another agent, and not the controller, so this
> > > implementation is not correct anymore.
> > 
> > 
> > Does the `another Agent` is a known Agent which we have received Topology Discovery message from it? Because if it does, you can verify that the "dest_alid" is the controller or one of the known neighbors on `m_1905_neighbor_devices`.
> 
> To be honest, I'd remove the validation entirely. If the UCC says we have to send a CMDU to some destination, who are we to claim that that's a destination you're not allowed to send to? I can imagine that we would add tests of corner cases where we want to force sending a CMDU to an unknown agent, e.g. to verify that a security measure works.
> 
> That said, I don't care much - we can still remove the check entirely when we need to. But I wouldn't spend time on adding "security" measures to code that isn't even enabled in production.

I suggest to remove the check entirely and just print a warning if the destination ALid is not one of the known neighbors.

arnout:
So, conclusion:
1. Check m_1905_neighbor_devices and print a warning if it's not in there.
2. Add a test to test_flows.py
3. Mention 4.6.3 in the last commit's commit message as per [WiKi - Testing in CI](https://github.com/prplfoundation/prplMesh/wiki/Testing-in-CI#certification-tests)

I'd like to get this merged ASAP so we can report one more passing test on Wednesday's architecture call...

tomereli (approved):
Please refer to https://github.com/prplfoundation/prplMesh/pull/1197#issuecomment-623577879
You need to add a comment in the last commit message which will tell CI to run your test and make it mandatory - please do this ASAP before you merge this

tomereli:
> Please refer to [#1197 (comment)](https://github.com/prplfoundation/prplMesh/pull/1197#issuecomment-623577879)
> You need to add a comment in the last commit message which will tell CI to run your test and make it mandatory - please do this ASAP before you merge this

Nevermind, I did that for you. Next time:)

tomereli:
> So, conclusion:
> 
> 1. Check m_1905_neighbor_devices and print a warning if it's not in there.
> 2. Add a test to test_flows.py
> 3. Mention 4.6.3 in the last commit's commit message as per [WiKi - Testing in CI](https://github.com/prplfoundation/prplMesh/wiki/Testing-in-CI#certification-tests)
> 
> I'd like to get this merged ASAP so we can report one more passing test on Wednesday's architecture call...

So apparently this is needed for many more tests. I reviewed it again and I think it is stupid to have this check in the first place, so I removed it completely. With this (and a change to the 4.10.3 tests controller), we pass both 4.6.4 and all 4.10.3 test variants, so this is one small change for prplmesh but a big step for EasyMesh agent certification :)

arnout (approved):
I love it when people follow my suggestion :-)

PR #1200, opened 2020-04-28
Feature/ci certification use run tests py

rmelotte:
Use run_test_file.py instead of run_test_file.sh.

Note:
Uploading to the google sheet is not enabled yet, because it requires https://git.prpl.dev/prplmesh/wfa-certification/easymesh_cert/-/issues/4 to be done first


tomereli (approved):

arnout (approved):

rmelotte:
There is something that makes the dummy tests hang in CI, so I added the don't merge tag for now until we can find the root cause.

RanRegev (approved):

rmelotte:
@mergifyio rebase

PR #1201, opened 2020-04-28
tests: initial_ap_config.py: remove utils.py class

abelog:
### Description 

This PR is reaction to discussion of #1136 which was closed too early.

To avoid unnecessary refactoring, we have to be as close as possible to
original `test_flows.py`. Thus,  suggestion to introduce base class for the prplMesh test and derive all other tests from it was raised.

### Changes

- Remove separate `utils.py` and embed its content in
`prplmesh_base_test.py`. 
- Renames in `InitialApConfig` test due to move
of code to `PrplMeshBaseTest` class.


arnout (approved):
I think runTest itself will eventually move to the base class as well. But that's refactoring for later.

rmelotte (approved):

PR #1202, opened 2020-04-28
DCS cleanups

### LINKED ISSUE #665: [TASK] DCS cleanups

itayx:
# Cleanups
> Fix all remaining review comments which were marked as such which can be cleaned up later.
### common/beerocks/bwl/CMakeLists.txt:62
- Fixed in #680 
### common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:311
- [x] ` if (!m_dwpal_nl_ctx) {`
> - The dwpal netlink context is created in the Init state.
> - On failure, we transition the FSM to the Detach state.
> - In what scenario do we get to the Attach state without a valid context? this statement is redundant.
> - RAX40 has a much older driver version at the moment so the attach will fail
> - We mustn't fail the FSM on that! Just log an warning and proceed without NL. Similar to what we do when we lack permissions to open raw socket - we simply don't open it.
> - Still, this should be removed
### Remove / Alter redundant logs
- [x] common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp:467
- [x] common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:594
- [x] common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp:284
### Small patches
- [x] common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:311 This should be removed
- [x] common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:593 `if (!vendor_data) { ... }`
- [x] common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:616 `if (!out_buffer) { ... }`
- [x] **[Redundant]**  common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:609 Function should return ssize_t (signed size_t). 
- [x] common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp:626 Either initialize data_size to -1, or simply return -1 here to indicate a failure.
### remove `dummy_return`
- [x] controller/src/beerocks/master/db/db.cpp:2355
- [x] controller/src/beerocks/master/db/db.cpp:2300
### controller/src/beerocks/master/db/db.h:400
- [x] channel_scan_enabled
> - bool channel_scan_enabled(const std::string &mac) const; //get
> - void channel_scan_enabled(const std::string &mac, bool enable); //set

To keep the DCS APIs with the pattern of the DB APIs I would recommend keeping the `get` & `set` pattern in the function names
### Enum refactoring
- [x] common/beerocks/bcl/include/bcl/beerocks_defines.h:450
- [x] common/beerocks/bcl/include/bcl/beerocks_defines.h:461
### `set_channel_scan_in_progress`, check if can read task fsm state instead
- [x] controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp: 81

The 'scan_in_progress' flag is access from outside the task, in the `son_management` therefore using a flag valiable in the DB is easier & cleaner then trying to access the FSM state.
### enum -> class enum
- [x] common/beerocks/bwl/include/bwl/mon_wlan_hal_types.h:351
> - `enum class eChannelScanResultMode : uint8_t {NA, AdHoc, Infra};`


morantr (approved):

adam1985d (commented):
spelling mistake in commit message "sone" instead of "some"

please re-review all commit messages

adam1985d (changes requested):
is this change done due to KW report?
if so - please also provide updated KW report files

adam1985d (commented):
commit 5ce7717f38bc0b42bf8db221af76abf77ca4580a:
commit message exeeds allowed commit message line length

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -628,7 +628,7 @@ size_t base_wlan_hal_dwpal::dwpal_nl_cmd_get(const std::string &ifname, unsigned
                                              unsigned char *out_buffer,
                                              const size_t max_buffer_size)
 {
-    size_t data_size = 0;
+    size_t data_size = -1;
 
     if (!out_buffer) {
         LOG(ERROR) << "out_buffer is invalid ==> Abort!";
```
adam1985d, line 634: `data_size` is -1 and never changed until the following while loop is used:
`while ((data_size == 0) && std::chrono::steady_clock::now() < timeout) {`
this condition will never be met.

did you test this flow?

anyways I prefer to return -1 and not data_size in case of failure - instead of changing data_size to -1 just for that

1. return `-1` instead of the `data_size on` failure
2. set data_size back to `0` and you can move the point of declaration to above the lambda (close to where it will be first used after the above-mentioned change)

adam1985d (commented):
your commit message includes grammar mistakes
please use Grammarly or tool of some sort to avoid this

adam1985d (changes requested):
your change here is renaming the already implemented empty container `dummy_return` to `empty`

your commit message should refer to the change performed

adam1985d (commented):
in commit message `respectably` should be `respectively`

adam1985d (commented):
spelling and grammar mistakes in commit message

kantera800 (approved):

adam1985d (commented):
`This is now depreciated by the LOG define and is no longer needed.`
- deprecated not depreciated
- it is not deprecated by the LOG define - it is encapsulated inside the LOG define and therefor can be deprecated from the LOG macro use

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
```diff
@@ -73,8 +73,8 @@ class base_wlan_hal_dwpal : public virtual base_wlan_hal,
      * @param max_buffer_size buffer size
      * @return size of returned data (success if > 0)
      */
-    size_t dwpal_nl_cmd_get(const std::string &ifname, unsigned int nl_cmd,
-                            unsigned char *out_buffer, const size_t max_buffer_size);
+    ssize_t dwpal_nl_cmd_get(const std::string &ifname, unsigned int nl_cmd,
```
adam1985d, line 76: it is good you changed the function signature to `ssize_t` but in `mon_wlan_hal_dwpal` which makes use of this function you didn't check the returned value.
also, did you trigger a negative test? to see the `mon_wlan_hal_dwpal` handles error correctly?

adam1985d (commented):
a short and clear commit message here would be:
`In the sChannelScanResults several enums are used to denote different parameters of a scan result.

Added class keyword to enums, for strongly typed enums.
Refactor all uses in mon_wlan_hal_dwpal

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
itayx: Changed the code in `mon_wlan_hal_dwpal` to match the new signature
I didn't test the error handling, but since the change was from  `size_t ` to `ssize_t` they should function identically

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
itayx: Currently the function returns `ssize_t` which accepts either a positive `size_t` value or `-1`
And for each of the error handling in the function I returned `-1` to signal an error

adam1985d (approved):

PR #1205, opened 2020-04-28
Hotfix/certification deploy dummy

rmelotte:
We have an issue in the certification testbed where the dummy prplMesh is not always stopped.

This PR deploys the dummy build to /opt/prplMesh, so that it's always accessible on subsequent runs, to fix the issue.

It also uses `ci/run_easymesh_cert.sh` in Gitlab CI, so that both the nightly tests on the intel testbed and the tests in CI use the same script.

It requires the following change in the easymesh_cert repository: https://git.prpl.dev/prplmesh/wfa-certification/easymesh_cert/-/merge_requests/12

### DIFF
--- a/ci/run_easymesh_cert.sh
+++ b/ci/run_easymesh_cert.sh
@@ -32,6 +32,7 @@ usage() {
     echo "      --owncloud-upload - whether or not to upload results to owncloud. (default: false)"
     echo "      --owncloud-path - relative path in the owncloud server to upload results to. (default: $OWNCLOUD_PATH)"
     echo "      --skip-upgrade - skip prplmesh upgrade (default: false)"
+    echo "      --skip-download - skip prplmesh download (default: false)"
     echo "'test' can either be a test name, or the path to a file containing"
     echo " test names (newline-separated)."
     echo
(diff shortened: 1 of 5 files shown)


ci/run_easymesh_cert.sh
```diff
@@ -54,29 +56,47 @@ is_prplmesh_device() {
...
+    if ! mv "$rootdir/build/install" /opt/prplMesh ; then
+        err "Failed to deploy prplMesh locally, aborting"
+        exit 1
+    fi
+}
+
 upgrade_prplmesh() {
+    if [ "$TARGET_DEVICE" = prplmesh ] ; then
+        deploy_dummy_prplmesh && return 0 || return 1
+    fi
```
arnout, line 74: I think it would be nicer to do:
```suggestion
    if [ "$TARGET_DEVICE" = prplmesh ] ; then
        deploy_dummy_prplmesh
    else    
        deploy_prplwrt_prplmesh
    fi
```
Note that the return is implied: the return value of the function is the one of the last statement. So if we have nothing after this, that is sufficient. But maybe a bit fragile.

ci/run_easymesh_cert.sh
```diff
@@ -54,29 +56,47 @@ is_prplmesh_device() {
     return 1
 }
 
+deploy_dummy_prplmesh() {
+    info "Deploy prplmesh locally to /opt/prplMesh"
+    rm -rf /opt/prplMesh
```
arnout, line 63: Is this script run as root?

ci/run_easymesh_cert.sh
```diff
@@ -54,29 +56,47 @@ is_prplmesh_device() {
...
+    mkdir -p /opt/prplMesh
+    if ! mv "$rootdir/build/install" /opt/prplMesh ; then
```
arnout, line 65: Err, won't this put the installation under /opt/prplMesh/install instead of under /opt/prplMesh?

More importantly, however: AFAIR, prplMesh isn't relocatable. When you move it somewhere else, some things won't work anymore. For example configuration file paths are hardcoded. So if we want to install it there, we should also build with CMAKE_INSTALL_PREFIX=/opt/prplMesh (and then install with DESTDIR=...). And that is a much bigger change I'm afraid...

ci/run_easymesh_cert.sh
rmelotte: ~~Yes, because run_test_file.sh needs to be run as root.~~
Actually no, this particular script isn't run as root, which means the runner will have to have write permissions on /opt

ci/run_easymesh_cert.sh
rmelotte: Indeed that's annoying...

Unless you see an issue with that, I'll update the CI configuration to always build with the prefix set to `/opt/prplMesh`.

arnout: > Unless you see an issue with that, I'll update the CI configuration to always build with the prefix set to `/opt/prplMesh`.

I see an issue with that! test_flows.py assumes that it's installed in rootdir/build/install... So, if we move it to /opt/prplMesh (which is a good thing IMO), we should also change test_flows.py to start with a deployment that puts it there, and change maptools.py to use DESTDIR during installation. So it's quite a big change...

rmelotte: Can't we do the deployment outside of test_flows.py and update test_flows.py to point to /opt/prplMesh?

rmelotte:
Building and deploying to /opt could be an annoying change for people working and testing on the same shared machine, so closing for now.

PR #1208, opened 2020-04-28
VSCode Development Container and Linux Debugging

vitalybu:
[Visual Studio Code](https://code.visualstudio.com/) is a highly popular, cross-platform, code editor. Since it's quite popular among prplMesh developers, I thought that it's a good idea to use some of the editor's features to simplify the development and debugging flows.

This pull requests adds support for [VSCode Remote - Containers](https://code.visualstudio.com/docs/remote/containers) to create an all-in-one environment using a single docker container.

The devcontainer automatically installs all the necessary tools for developing (Intellisense), linting (clang-format/flake8), running (podman) and debugging (attaching to processes running in containers). The only requirement is a Docker daemon on the host machine.

Everything was tested on a Windows 10 machine running Docker for Windows.
Would appreciate if someone could make sure everything works on additional platforms.

Known issues/limitations:
- Due to a podman bug/limitation, the devcontainer is started in privileged mode
- podman uses `vfs` as the `storage-driver` since `overlay` is not supported inside a docker container, and `fuse-overlayfs` is not available for Ubuntu 18.04. 
  > I did try upgrading everything (devcontainer and prplmesh-runner) to Ubuntu 20.04, but the performance gain wasn't dramatic, so decided to ignore it for now.
- [VSCode Remote - SSH](https://code.visualstudio.com/docs/remote/ssh) is not supported, as VSCode is unable to run in a container on remote machines. [SSH Tunneling](https://code.visualstudio.com/docs/containers/ssh) can be a workaround, but didn't have time to test it yet.
- Debugging is currently supported only for Linux (dummy).

Short demonstration of attaching and debugging one of the `beerocks_agent` processes:

Capturing traffic on the internal podman network bridge using Wireshark's [`sshdump`](https://www.wireshark.org/docs/man-pages/sshdump.html):


tools/docker/image-build.sh
```diff
@@ -34,10 +36,11 @@ main() {
 
     while true; do
         case "$1" in
-            -v | --verbose)         VERBOSE=true; shift ;;
             -h | --help)            usage; exit 0; shift ;;
             -b | --base-image)      IMAGE="$2"; shift ; shift ;;
             -t | --tag)             TAG=":$2"; shift ; shift ;;
+            --no-builder)           BUILDER=false; shift ;;
+            --no-runner)            RUNNER=false; shift ;;
```
rmelotte: Wouldn't it be nice to be able to specify the name of the image to build instead? (maybe the default could be to generate them all as it was before).
We have at least 2 other images (that are not yet built with this script), but I can't imagine having a `--no-<image>` option for all of them.

(nitpick)

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,72 @@
...
+COPY scripts/root/.inputrc /root/
+
+# Switch back to dialog for any ad-hoc use of apt-get
+ENV DEBIAN_FRONTEND=dialog
+
+# Default user is root
+ENV USER=root
+
+# SSH Server
+EXPOSE 2222
```
rmelotte, line 72: I don't really understand: why do you need an SSH server running inside the container?
Is the container running on a remote host?
In that case unless I'm missing something, you still need some shell access to setup the container on the remote host. Can't the same access be used to get a shell to the container directly with `docker exec`?

tools/docker/image-build.sh
vitalybu: You're right, it's a good idea, I'll make the change

.devcontainer/Dockerfile
vitalybu: To use with Wireshark's [`sshdump`](https://www.wireshark.org/docs/man-pages/sshdump.html).
On Linux you could probably just pipe the output using `docker exec` as well, but it's a bit more complicated to do so on Windows.
Also the option to use `sshdump` is conveniently available inside the Wireshark UI (second `gif` in the PR description).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,29 @@
...
+		2222
+	],
+	// IDs of extensions to install inside the container
+	"extensions": [
+		"ms-vscode.cpptools",
+		"twxs.cmake",
+		"eamodio.gitlens",
+		"github.vscode-pull-request-github",
+		"ms-python.python",
+		"redhat.vscode-yaml"
```
morantr: This is great! I would also recommend adding:
spellcheck (timonwong.shellcheck)
Markdown Preview Enhanced (shd101wyy.markdown-preview-enhanced)
Doxygen Documentation Generator (cschlosser.doxdocgen)
Code Spell Checker (streetsidesoftware.code-spell-checker) - Hopefully we will have less typo error with this plugin

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,29 @@
...
+	]
+}
```
morantr, line 31: On a lot of files, there isn't a newline at the end of the file.
I remember it was written somewhere that it is required, but I couldn't find it.

morantr (approved):

.devcontainer/Dockerfile
rmelotte: Ok, thanks for the explanations :-)

rmelotte (approved):
I don't see changes for the selective build although the conversation is marked as resolved, but there's nothing blocking anyway, so approving.

vitalybu:
> I don't see changes for the selective build although the conversation is marked as resolved, but there's nothing blocking anyway, so approving.

Look at this [commit](https://github.com/prplfoundation/prplMesh/pull/1208/commits/2e9504365d8f3aaab7eb219f5894d21f0061aea6). Pushed the rebased branch (not the fixups).

rmelotte:
> Look at this commit. Pushed the rebased branch (not the fixups).

Ok, then it's not exactly what I had in mind.
My suggestion was to have a single option to be able to choose what you want to build (and by default build them all).
For example: `--image runner` would build the runner only.
With the current implementation, you still have an option for each image which is a bit cumbersome IMO. It's also weird since by default nothing will be built.

It could even go further and allow multiple `--image` options to build multiple ones at once, but if the default is to build them all it's enough IMO. 

Also if you don't like to have an `--image` option for that, you could instead say that the image names are positional, so that the script would be invoked like this: `image-build.sh runner builder`.
From the usage text, either I'm reading it wrong, or it's how it was supposed to be (although <type> is missing from the first usage line):
`
"  mandatory:"
    echo "      type - image type <runner/builder>"
`

.devcontainer/devcontainer.json
vitalybu: JSON linter removes empty lines from end-of-files.
We have this in CPP files, but not sure it should apply to JSON settings files as well.

vitalybu:
> Ok, then it's not exactly what I had in mind.

Made another change to specify the images to build using an `--image` argument.

Galvien (approved):

.devcontainer/devcontainer.json
vitalybu: Added the Code Spell Checker and a Markdown All in One extensions

Issue #1056, opened 2020-03-31
[BUG] ARP monitor reply messages no received  back from associated client


kantera800:
**Describe the bug** -
The issue is spotted with clients with IP 0.0.0.0 (dhcp monitor fails to detect the IP)

ARP message is sent broadcatst dest_ip 255.255.255.255  (0.0.0.0)

but no valid response (or no response at all) messages received from the client.

update:
same behavior when sending unicast message to client specific IP 

PR #1209

## Steps to reproduce

1. Build + falsh on platform
2. Run controller & agent
3. connect a client to one of the radios
4. enable band steering from beerocks_cli
5. trigger optimal path from beerocks_cli
6. controller ARP monitor request  reaches timeout

see  controller logs:
_optimal_path_task.cpp[776] --> task  id 38: requesting rssi measurements for c4:0b:cb:ea:79:46 from 00:50:f1:80:e3:80
optimal_path_task.cpp[1279] --> task  id 38: response from 00:50:f1:80:e3:80 timed out_

see monitor logs:
_monitor_thread.cpp[940] --> received ACTION_MONITOR_CLIENT_RX_RSSI_MEASUREMENT_REQUEST
monitor_thread.cpp[984] --> RX_RSSI_MEASUREMENT REQUEST: state IDLE -> SEND_ARP, cross=0 sta_mac=c4:0b:cb:ea:79:46 id=38
monitor_rssi.cpp[331] --> state: SEND_ARP -> WAIT_REPLY, arp_iface = brlan0, arp_iface_ipv4 = 192.168.0.1, arp_iface_mac = 02:50:f1:20:e3:81, is_4addr_client = 0, sta_mac = c4:0b:cb:ea:79:46, dest_ip = 0.0.0.0, dst_mac = c4:0b:cb:ea:79:46
monitor_rssi.cpp[331] --> state: SEND_ARP -> WAIT_REPLY, arp_iface = brlan0, arp_iface_ipv4 = 192.168.0.1, arp_iface_mac = 02:50:f1:20:e3:81, is_4addr_client = 0, sta_mac = c4:0b:cb:ea:79:46, dest_ip = 0.0.0.0, dst_mac = c4:0b:cb:ea:79:46
[...]

PR #1209, opened 2020-04-30
Bugfix/fix arp response is not received

morantr:
It is observed that sometimes mon_rssi_want to send arp request, but it doesn't have the
ipv4 of the client. If we don't have the station ipv4, the arp request will not get a response, and sending it to wildcard IP doesn't work apparently.

This PR adds the following:
Fix bug in the updating of the station IP on the Agent.
Fix ipv4 that was unset on start monitoring message.

### DIFF
--- a/common/beerocks/bcl/source/network/network_utils.cpp
+++ b/common/beerocks/bcl/source/network/network_utils.cpp
@@ -1010,7 +1010,7 @@ bool network_utils::arp_send(const std::string &iface, const std::string &dst_ip
     // with broadcast IP, so all clients will receive it and answer, but since the request is
     // being sent to a specific mac address, then only the requested client will answer.
     uint32_t dst_ip_uint;
-    if (dst_ip.empty()) {
+    if (dst_ip.empty() || dst_ip == ZERO_IP_STRING) {
         dst_ip_uint = 0xFFFFFFFF; // "255.255.255.255" equivalent
     } else {
         dst_ip_uint = network_utils::uint_ipv4_from_string(dst_ip);
--- a/controller/src/beerocks/master/son_master_thread.cpp
+++ b/controller/src/beerocks/master/son_master_thread.cpp
@@ -3026,46 +3026,49 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
             database.update_node_last_seen(client_mac);
         }
 
-        if (database.get_node_type(client_mac) == beerocks::TYPE_CLIENT) {
-            auto db_ipv4 = database.get_node_ipv4(client_mac);
-            if (!database.set_node_ipv4(client_mac, ipv4)) {
-                LOG(ERROR) << "set node ipv4 failed";
+        if (database.get_node_type(client_mac) != beerocks::TYPE_CLIENT) {
+            LOG(INFO) << "Ignoring DHCP notification for mac " << client_mac
+                      << ", as it's not a client";
+            return true;
+        }
+
+        database.set_node_ipv4(client_mac, ipv4);
+        database.set_node_name(client_mac,
+                               std::string(notification_in->name(message::NODE_NAME_LENGTH)));
+
+        if (database.is_node_wireless(client_mac)) {
+            auto notification_out = message_com::create_vs_message<
+                beerocks_message::cACTION_CONTROL_CLIENT_NEW_IP_ADDRESS_NOTIFICATION>(cmdu_tx);
+
+            if (!notification_out) {
+                LOG(ERROR) << "Failed building message!";
+                return false;
             }
+            notification_out->mac()  = notification_in->mac();
+            notification_out->ipv4() = notification_in->ipv4();
 
-            if (!database.set_node_name(
-                    client_mac, std::string(notification_in->name(message::NODE_NAME_LENGTH)))) {
-                LOG(ERROR) << "set node name failed";
+            auto client_bssid = database.get_node_parent(client_mac);
+            if (client_bssid.empty()) {
+                LOG(WARNING) << "Client does not have a valid parent hostap on the database";
+                return true;
             }
+            auto client_radio = database.get_node_parent_radio(client_bssid);
+            LOG(WARNING) << "Client " << client_mac
+                         << " is connected wirelessly, Sending IP addr notification to radio="
+                         << client_radio;
 
-            if (database.is_node_wireless(client_mac)) {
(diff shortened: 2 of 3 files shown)


common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -101,8 +101,8 @@ class network_utils {
     static int get_iface_info(network_utils::iface_info &info, const std::string &iface_name);
     static bool get_raw_iface_info(const std::string &iface_name, raw_iface_info &info);
 
-    //temp
     static std::string get_mac_from_arp_table(const std::string &ipv4);
+    static std::string get_ipv4_from_arp_table(const std::string &mac);
```
arnout: Please, no new functions which take a mac-as-string as argument...

arnout (approved):
I'll approve it, but I wonder if this really works in practice? How is the MAC address going to end up in the ARP table - that only happens if either we sent an IP packet to the STA, or the STA sent an IP packet to us. That is only going to happen in practice if we're the DHCP server - but normally, only the gateway runs a DHCP server.

That said, if it works, I'll accept it.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -419,23 +419,24 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
- *
- * IP address       HW type     Flags       HW address            Mask     Device
- * 192.168.12.31    0x1         0x2         00:09:6b:00:02:03     *        eth0
- * 192.168.12.70    0x1         0x2         00:01:02:38:4c:85     *        eth0
- */
+    // /proc/net/arp looks like this:
+    // IP address       HW type     Flags       HW address            Mask     Device
+    // 192.168.12.31    0x1         0x2         00:09:6b:00:02:03     *        eth0
+    // 192.168.12.70    0x1         0x2         00:01:02:38:4c:85     *        eth0
     std::ifstream arp_table("/proc/net/arp");
```
vitalybu: This is definitely not the way to do this.
If we decide to go this way, should use netlink sockets instead.

vitalybu (changes requested):

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -1030,21 +1030,16 @@ bool network_utils::arp_send(const std::string &iface, const std::string &dst_ip
...
+    }
+
     int tx_len;
     arp_hdr arphdr;
     struct sockaddr_ll sock;
     uint8_t packet_buffer[128];
-
-    // If the destination IP is empty, there is no point sending the arp, therefore replace the
-    // with broadcast IP, so all clients will receive it and answer, but since the request is
-    // being sent to a specific mac address, then only the requested client will answer.
```
kantera800, line 1011: You need to have a better reason than "not working well"  to remove this code.
It used to work so clearly something broke , or maybe it does work on some clients (another reason to keep it) 
Also , this commit is not related to this PR and you cannot close  #1056 until this issue is resolved or at least debugged.

agent/src/beerocks/monitor/monitor_rssi.cpp
```diff
@@ -337,6 +337,12 @@ void monitor_rssi::process()
                            << ", sta_mac = " << sta_mac << ", dest_ip = " << sta_node->get_ipv4()
                            << ", dst_mac = " << arp_dst_mac;
 
+                if (sta_node->get_ipv4().empty() ||
+                    sta_node->get_ipv4() == network_utils::ZERO_IP_STRING) {
+                    auto ipv4 = network_utils::get_ipv4_from_arp_table(arp_dst_mac);
+                    sta_node->set_ipv4(ipv4);
+                }
```
kantera800: instead of adding this w/a  you should  fix the real problem.
i see that in the past you already added a w/a for this issue https://github.com/prplfoundation/prplMesh/commit/873bf9ed6a3fba6322e8eb001071e991bf108bf1 , but then the `Wait until ipv4 will be received on DHCP event` part was removed .
So now cACTION_CONTROL_CLIENT_START_MONITORING_REQUEST  always sent with  ip=0

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3026,46 +3026,53 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
             database.update_node_last_seen(client_mac);
         }
 
-        if (database.get_node_type(client_mac) == beerocks::TYPE_CLIENT) {
-            auto db_ipv4 = database.get_node_ipv4(client_mac);
-            if (!database.set_node_ipv4(client_mac, ipv4)) {
-                LOG(ERROR) << "set node ipv4 failed";
-            }
+        if (database.get_node_type(client_mac) != beerocks::TYPE_CLIENT) {
+            LOG(DEBUG) << "node " << client_mac << " is not of TYPE_CLIENT";
```
vitalybu:
```suggestion
            LOG(INFO) << "Ignoring DHCP notification for mac " << client_mac << ", as it's not a client";
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -3026,46 +3026,53 @@ bool master_thread::handle_cmdu_control_message(const std::string &src_mac,
...
+            return true;
+        }
 
-            if (!database.set_node_name(
-                    client_mac, std::string(notification_in->name(message::NODE_NAME_LENGTH)))) {
-                LOG(ERROR) << "set node name failed";
-            }
+        if (!database.set_node_ipv4(client_mac, ipv4)) {
+            LOG(ERROR) << "set node ipv4 failed";
```
vitalybu: Does it makes sense to continue in this case?
If you don't have an IP, what are you actually sending to the agent?

vitalybu (approved):
Few more comments, but approving as it should also work as is

controller/src/beerocks/master/son_master_thread.cpp
morantr: Actually it cannot be failed if the previous db setter success.
The set ipv4 failed if the client has not been found on the db.
I'll remove the failure handling.

controller/src/beerocks/master/son_master_thread.cpp
morantr: done.

kantera800 (approved):

PR #1210, opened 2020-04-30
test_gw_repeater.sh: correctly use array for _EXTRA_OPT

arnout:
Commit c6d6aa05cadc1dae51b678f79d34ce0497b16aff fixed the shellcheck
errors SC2086 about double quoting around the GW_EXTRA_OPT and
RP_EXTRA_OPT variables by using the bash array syntax. However, the
place where those variables are set was not updated to actually create
an array.

Do so now.

This issue was observed by @RanRegev.
Since I don't use WSL myself, I wasn't able to test it.


rmelotte (approved):
LGTM, although I couldn't test it either

vitalybu (approved):
Tested with WSL, works fine.

PR #1213, opened 2020-04-30
Dev/fix script paths

tomereli:
scriptdir and rootdir are defined globally in many of the prplmesh
scripts. Some scripts source other scripts, so these variables make the
scripts extremely fragile, since they may get overriden when sourcing
another script which defines them with the same name.

Update all utility scripts are bash, and use ${BASH_SOURCE[0] once
in the functions.sh script which is meant to be sourced, and export
ROOT_DIR only once so it can be safely used by all other scripts.
Since this is already done for tools/docker/run.sh with installdir,
remove this and update run.sh as the first script to use ROOT_DIR.
Followup commit will update the rest of the script to use ROOT_DIR
without defining global variables which may be overriden by other
sourced scripts.

Note that we use upper case for ROOT_DIR since shellcheck doesn't count
this as error/warning when using an upper case variable in a script if
it is not defined in the same script -
https://github.com/koalaman/shellcheck/wiki/SC2154#rationale.

### DIFF
--- a/tools/docker/build.sh
+++ b/tools/docker/build.sh
@@ -6,24 +6,21 @@
 # See LICENSE file for more details.
 ###############################################################
 
-scriptdir="$(cd "${0%/*}" || exit 1; pwd)"
-rootdir="${scriptdir%/*/*}"
-
 # shellcheck source=../../tools/functions.sh
-. "${rootdir}/tools/functions.sh"
+. "$(dirname "${BASH_SOURCE[0]}")/../../tools/functions.sh"
 
 main() {
     docker image inspect prplmesh-builder >/dev/null 2>&1 || {
         echo "Image prplmesh-build does not exist, creating..."
-        run "${scriptdir}"/image-build.sh
+        run "${ROOT_DIR}"/tools/docker/image-build.sh
     }
 
     # Default docker arguments
     docker_args=(
-        --workdir "${rootdir}"
+        --workdir "${ROOT_DIR}"
         --user "${SUDO_UID:-$(id -u)}:${SUDO_GID:-$(id -g)}"
         -e "USER=${SUDO_USER:-${USER}}"
-        -v "${rootdir}:${rootdir}"
+        -v "${ROOT_DIR}:${ROOT_DIR}"
         --entrypoint "./tools/maptools.py"
     )
 
--- a/tools/docker/builder/openwrt/build.sh
+++ b/tools/docker/builder/openwrt/build.sh
@@ -1,10 +1,7 @@
-#!/bin/sh -e
+#!/bin/bash
 
-scriptdir="$(cd "${0%/*}"; pwd)"
-rootdir="${scriptdir%/*/*/*/*}"
-
-# shellcheck source=tools/functions.sh
-. "${rootdir}/tools/functions.sh"
+# shellcheck source=../../../../tools/functions.sh
+. "$(dirname "${BASH_SOURCE[0]}")/../../../../tools/functions.sh"
 
 usage() {
     echo "usage: $(basename "$0") -d <target_device> [-hfiortv]"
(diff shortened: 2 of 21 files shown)


tools/functions.sh
```diff
@@ -1,4 +1,4 @@
-# shellcheck shell=sh
+#!/bin/bash
```
arnout, line 1: Not here. This script is not meant to be executed directly, so it shouldn't have a shebang line.
```suggestion
# shellcheck shell=bash
```

tools/docker/run.sh
```diff
@@ -6,11 +6,10 @@
 # See LICENSE file for more details.
 ###############################################################
 
-scriptdir=$(dirname "$(readlink -f "$0")")
-rootdir=$(realpath "$scriptdir/../..")
+# shellcheck source=../../tools/functions.sh
```
arnout, line 9: We're going in circles with this shellcheck source stuff...

Let's wait for #1204 to be merged, and not change it anymore after that.

tools/docker/run.sh
```diff
@@ -6,11 +6,10 @@
...
+. "$(dirname "${BASH_SOURCE[0]}")/../../tools/functions.sh"
```
arnout, line 10: Why do you remove scriptdir? There is no reason to stop using it IMO...
```
scriptdir=$(dirname "$(readlink -f "$0")")
# shellcheck source=tools/functions.sh
. "${scriptdir}/../../tools/functions.sh"
```

tools/functions.sh
```diff
@@ -6,8 +6,8 @@
 # See LICENSE file for more details.
 ###############################################################
 
-# shellcheck disable=SC2034
-installdir="${rootdir:?}/build/install"
```
arnout, line 10: Actually, this entire PR could be avoided by simply removing this line and moving it to the only script that uses it, run.sh.

tools/docker/build.sh
```diff
@@ -6,24 +6,21 @@
 # See LICENSE file for more details.
 ###############################################################
 
-scriptdir="$(cd "${0%/*}" || exit 1; pwd)"
```
arnout, line 9: To make this script source-able, it would be sufficient to change this to
```
scriptdir="$(dirname "${BASH_SOURCE[0]}")"
```
and change nothing else in the script.

It might still make sense to move the definition of ROOT_DIR into functions.sh, but I think it has only limited merit.

tools/docker/builder/openwrt/build.sh
```diff
@@ -1,10 +1,7 @@
-#!/bin/sh -e
```
arnout, line 1: This should have been in the first commit.

arnout (approved):
As noted, this is a very bit change with IMO very little merit.

That said, it doesn't hurt either, so the least effort path is probably to just merge it. Two changes are needed though:
- Adapt to the changes in #1204 (which probably means rebasing on that)
- Don't give a shebang to functions.sh.

tools/docker/run.sh
tomereli: OK.

tools/docker/run.sh
tomereli: I wanted to avoid using global variables as much as possible to reduce the risk of the same variable being used in a sourced script which brought us to this mess in the first place.

tools/functions.sh
tomereli: That's right. I will create a separate PR which does that.

tomereli: Not only this, but also remove scriptdir and rootdir from owncloud_definitions.sh. Any script which may be sourced should not define global variables which may be used by the callers.

tomereli: https://github.com/prplfoundation/prplMesh/pull/1218

tomereli:
Replaced with #1218

PR #1214, opened 2020-04-30
ci: use images from gitlab registry

arnout:
Whenever we change anything in a docker image, it is very hard to make sure that different PRs keep on working.
Therefore, the docker images themselves should also be built in CI.
To make this possible, the following changes are made:
- images are always specified with the registry component in the image name;
- an additional CI stage builds all images;
- git commit hashes are used as tags.

In addition this PR does some cleanups and random fixes.

For now, only the prplmesh-builder image is created as part of CI. For the runner, we will have to add options to the test_run.py script to specify the tag. Also the other docker containers we use (alpine, dind) will need to be built in the containers stage. Those things will be done in later PRs.


tomereli (approved):

tools/functions.sh
```diff
@@ -9,6 +9,9 @@
 # shellcheck disable=SC2034
 installdir="${rootdir:?}/build/install"
 
+# Docker registry which is used in all docker-related scripts
+DOCKER_REGISTRY=registry.gitlab.com/prpl-foundation/prplmesh/
```
vitalybu: `functions.sh` is a kinda generic script, not sure it's the right place for this global.
Maybe we should have a spearate, sourcable, script for docker specific scripts

tools/docker/build.sh
```diff
@@ -32,7 +27,7 @@ main() {
         docker_args+=(-v "${PRPLMESH_PLATFORM_BASE_DIR}:${PRPLMESH_PLATFORM_BASE_DIR}")
     fi
 
-    run docker container run "${docker_args[@]}" --rm prplmesh-builder build "$@"
+    run docker container run "${docker_args[@]}" --rm ${DOCKER_REGISTRY}prplmesh-builder build "$@"
```
vitalybu:
```suggestion
    run docker container run "${docker_args[@]}" --rm "${DOCKER_REGISTRY}"prplmesh-builder build "$@"
```

tools/docker/run.sh
```diff
@@ -108,7 +103,7 @@ main() {
 
     # Save the container name so that it can easily be stopped/removed later
     echo "$NAME" >> "${scriptdir}/.test_containers"
-    run docker container run "${DOCKEROPTS[@]}" "prplmesh-runner$TAG" "$@"
+    run docker container run "${DOCKEROPTS[@]}" "${image}" "$@"
```
vitalybu, line 108: Why not build a local image if the pull fails?

.gitlab-ci.yml
```diff
@@ -1,13 +1,24 @@
...
 
+images-build:
+  stage: containers
+  tags:
+    - shell
+    - docker-build
+  script:
+    - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
+    - ./tools/docker/image-build.sh -i builder -t $CI_PIPELINE_ID
+    - docker push $CI_REGISTRY_IMAGE/prplmesh-builder:$CI_PIPELINE_ID
```
vitalybu, line 17: If the build is not tagged as `latest` how will the clients know to pull the latest version?

tools/functions.sh
arnout: I don't know... The only non-docker scripts that use functions.sh are:
- cppcheck-new-issues.sh
- upload_to_owncloud.sh - but this will be converted to Python
- run_easymesh_cert.sh - wasn't it going to be converted to Python as well?
- download_ipk.sh

For these, having some variable defined that is not used in that script shouldn't be an issue, should it? The script also defines some other things that are maybe not defined anywhere else.

tools/docker/build.sh
arnout: Damn, I was sure I had run shellcheck on it...

.gitlab-ci.yml
arnout: For now, I'm assuming latest will be pushed manually by a developer when the image changes.

tools/docker/run.sh
arnout: As explained in the commit message:

    docker: remove auto-building of images
    
    Since images are now pulled automatically, it is not needed to build
    them automatically if they don't exist. Remove that feature.

The automatic building only happens the very first time - at that point, you could just as well pull instead of build, so it's kind of pointless.

It would make sense to build every time. But then I think we should have a script that does the entire build-image, build prplmesh, run tests flow. Building the images just takes 4s on my laptop when everything is cached, so it's OK-ish to do it every time.

rmelotte (approved):
I like it!
Shouldn't there also be a job to untag the image at the end? Or do you intend to do it from time to time using Gitlab's API?

I know Gitlab recently made changes to make it easier to do registry cleanup, but I don't know what it's current state is.

arnout:
> I like it!
> Shouldn't there also be a job to untag the image at the end? Or do you intend to do it from time to time using Gitlab's API?
> 
> I know Gitlab recently made changes to make it easier to do registry cleanup, but I don't know what it's current state is.

If with "untag", you mean to untag it on the runner: the problem is that the runner running the "untag" job may be a different one than the runner(s) that have the images. Or put differently: if there are 3 runners, it's very likely that each of them is running part of the pipeline, so in the end all images end up on all runners. So the "untag" step should execute on *all* runners, and AFAIK there's no way to do that in gitlab-ci.yml. Therefore, the only solution is to run a cron job that take care of untagging. I believe we anyway already have something like that for the prplwrt-builder images (or if we don't, we need it for them).

If it is for the registry: I was indeed counting on some registry cleanup tool that exists already.

PR #1215, opened 2020-05-01
Some cppcheks fixes

vitalii-komisarenko:
It is nice to have code that follows code style guidelines.

**To reviewers**: something may be wrong in this PR.
@RanRegev noted that using const& is not always the right thing and the tests are indeed failing from time to time.


common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -451,7 +451,7 @@ static bool get_scan_results_from_nl_msg(sChannelScanResults &results, struct nl
 /////////////////////////////// Implementation ///////////////////////////////
 //////////////////////////////////////////////////////////////////////////////
 
-mon_wlan_hal_dwpal::mon_wlan_hal_dwpal(std::string iface_name, hal_event_cb_t callback)
+mon_wlan_hal_dwpal::mon_wlan_hal_dwpal(const std::string &iface_name, hal_event_cb_t callback)
```
RanRegev, line 454: This constructor just as an example, this comment is for all changes that replaced a copy with a const reference:
are we sure the const reference we pass does not go out of scope leaving us with a dangling reference?
If we bind to a temporary, its lifetime is extended. But if we bind to a variable and this variable goes out of scope, then we have a problem.
reference: https://en.cppreference.com/w/cpp/language/reference#Dangling_references

RanRegev (approved):
1. I have a general comment that I attached to one of the constructors.
2. I think it would be better to be more precise in the commit, namely, state that std::string is replaced with const std::string& and not the general "pass by const reference" case.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
vitalii-komisarenko: I have to check it.
I hope there is no constructor/method that stores references to `std::string`s instead of copying them.

arnout: In fact, if it was a reference to a std::string, it wouldn't have worked to begin with, because the temporary that was passed originally goes out of scope much quicker and it *will* be overwritten with something else on the stack. So if a reference was stored in the object, we'd already have noticed.

arnout (approved):
Excellent PR with good commit messages. There are a few mistakes in it though...
- some of the commit message say "bcl" while it's actually bwl;
- typo agrs -> args;
- the commit messages are sometimes worded slightly differently. It's better to always use the same wording, that's easier to read when you browse the logs. Use `git commit -c HEAD` to repeat the same commit message.

arnout:
Oh BTW no reason to keep this in draft. It's better to merge things as quickly as possible. Less risk of conflicts that way. Better two separate PRs than a single one that takes days to get merged.

PR #1218, opened 2020-05-03
scripts: make owncloud_definitions.sh sourceable

tomereli:
When a shell script which may be sourced defines a global variable,
things can break easilly if the caller script defines a global variable
with the same name befoure the source directive.

Both functions.sh and owncloud_definitions.sh are sourcable scripts and
therefore this commit removes the global definitions since they are
anyway not needed:
1. functions.sh defines installdir which is only used by run.sh - so
move it there
2. owncloud_definitions.sh declares both scriptdir and rootdir which are
unused (we use BASH_SOURCE[0] to get the path to functions.sh). When
this script is sourced by run_easymesh_cert.sh, it breaks since the
rootdir path is overriden, so remove both.

### DIFF
--- a/ci/owncloud/owncloud_definitions.sh
+++ b/ci/owncloud/owncloud_definitions.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-scriptdir="$(cd "${0%/*}" || exit 1; pwd)"
-rootdir="${scriptdir%/*/*}"
-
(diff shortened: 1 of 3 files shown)


vitalybu (approved):

arnout (approved):
Could have been two separate commits, but OK.

PR #1219, opened 2020-05-03
.gitlab-ci.yml: use verbose mode when running cert tests

tomereli:
It is always useful to see the test's output in runtime, so add verbose
flag.

To make sure this looks OK - first run:
MAP-4.2.1:netgear-rax40

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -254,7 +254,7 @@ run-certification-tests:
             echo "Deploying to $DEVICE_UNDER_TEST"
             tools/deploy_ipk.sh --certification-mode $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/prplmesh.ipk
         fi
-      - /easymesh_cert/run_test_file.sh -o logs -d $DEVICE_UNDER_TEST $TESTS_TO_RUN
+      - /easymesh_cert/run_test_file.sh -v -o logs -d $DEVICE_UNDER_TEST $TESTS_TO_RUN
   artifacts:
     paths:
       - logs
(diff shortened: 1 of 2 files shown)


.gitlab-ci.yml
```diff
@@ -245,7 +245,7 @@ run-certification-tests:
             echo "Deploying to $DEVICE_UNDER_TEST"
             tools/deploy_ipk.sh --certification-mode $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/prplmesh.ipk
         fi
-      - /easymesh_cert/run_test_file.sh -o logs -d $DEVICE_UNDER_TEST $TESTS_TO_RUN
+      - /easymesh_cert/run_test_file.sh -v -o logs -d $DEVICE_UNDER_TEST $TESTS_TO_RUN
```
RanRegev, line 257: I don't know if run_test_file.sh script has a "long options" for parameters, but if it does I prefer using the long options in scripts and use the short ones when typing in shell. In this case I would guess it is --verbose.
This way maintenance is easier and simpler as future readers do not need to guess or to go and look for meaning in the script.

RanRegev (approved):

.gitlab-ci.yml
tomereli: I tend to agree, but `-v` is quite standard and since we're using short options now I prefer to keep it.

arnout (approved):

rmelotte (approved):
Please also add it to `ci/certification/generic.yml` before you merge it, to also enable it for the individual jobs.

PR #1223, opened 2020-05-04
DCS: Scan all available channels

### LINKED ISSUE #1216: Dynamic Channel Selection: Scan all available channels
The Dynamic Channel Selection feature scans a list of given channels (set by the user), and returns the scan results.
Currently the user must supply the requested channels, but we want to add the option to scan ALL available / supported channels  for a given radio.

To do this we need to add an option when setting the supported channels to also set the channel pool to all supported channels, then when a scan is request be it a single scan or a continuous scan.

After looking into the code, I found two options to add this feature:
1) In the DB function `set_channel_scan_pool`, add a condition that will check the given channel pool and in case the pool contains a single channel with a value of `0` all supported channels will be set as the current channel pool.
2) In the son_managment segments where we set the channel pool (during `SET_CONTINUOUS_PARAMS_REQUEST` & `START_SCAN_REQUEST`) check the requested channel pool and if the channel pool contains a single channel with a value of `0` set the new channel pool in the DB to all the radio's supported channels
3) in the DCS task, when checking the channel pool from the DB the task will check if the channel pool contains a single channel with a value of `0`, in that case the task will then send as part of the scan request all the supported channels for that radio.

Using options (1) and (3) will result in the least amount of code changed.
Using option (1) and (2) will result in the DB containing all supported channels in the DB, while in option (3) the channel pool stored in the DB will be the "All Channels" channel pool, but without the actual channels

I believe that using option (3) is the best solution.
When checking the channel pool it will show that the user intended that all supported channels are to be used in the upcoming scan.

itayx:
The Dynamic Channel Selection scans a list of given channels (set by the user), and returns the scan results.
Currently the user must supply the requested channels, but we want to add the option to scan ALL available / supported channels for a given radio.
In the DCS task, when checking the channel pool from the DB, need to check if the channel pool contains just one single channel with a the value of 0, This will indicate that the channel pool is set to "All Channels".
Therefore the scan request will contain all the supported channels for that radio.

Fixes #1216

### DIFF
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -52,41 +52,54 @@ beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_t
 {
     // When a scan is requested send the scan parameters Channel pool & Dwell time
 
+    auto radio_mac_str = beerocks::net::network_utils::mac_to_string(m_radio_mac);
+
     auto request = beerocks::message_com::create_vs_message<
         beerocks_message::cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST>(cmdu_tx);
     if (!request) {
         LOG(ERROR) << "Failed building message cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST";
         return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
     }
+    request->scan_params().radio_mac = m_radio_mac;
 
+    //Get dwell time, if dwell time invalid fail scan.
     int32_t dwell_time_msec = database.get_channel_scan_dwell_time_msec(m_radio_mac, m_single_scan);
     if (dwell_time_msec <= 0) {
-        LOG(ERROR) << "invalid dwell_time <= 0";
+        LOG(ERROR) << "invalid dwell_time=" << int(dwell_time_msec);
         return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
     }
+    request->scan_params().dwell_time_ms = dwell_time_msec;
 
-    //get current channel pool from DB
+    // Get channel pool;
+    // if pool is set to "all channel": use the supported channels instead.
+    // if the costum channel pool is empty or too big: fail scan and return error accordingly.
     auto &curr_channel_pool = database.get_channel_scan_pool(m_radio_mac, m_single_scan);
-    if (curr_channel_pool.empty()) {
-        LOG(ERROR) << "empty channel pool is not supported. please set channel pool for mac="
-                   << m_radio_mac;
-        return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
+    if (curr_channel_pool.size() == 1 && (*curr_channel_pool.begin()) == 0) {
+        LOG(DEBUG) << "Using all supported channels for channel scan on mac=" << m_radio_mac;
+        auto supported_channels = database.get_hostap_supported_channels(radio_mac_str);
+        request->scan_params().channel_pool_size = supported_channels.size();
+        std::transform(supported_channels.begin(), supported_channels.end(),
+                       request->scan_params().channel_pool,
+                       [](beerocks::message::sWifiChannel &c) -> uint8_t { return c.channel; });
+    } else {
+        LOG(DEBUG) << "Using custom set channels for channel scan on mac=" << m_radio_mac;
+        if (curr_channel_pool.empty()) {
+            LOG(ERROR) << "empty channel pool is not supported. please set channel pool for mac="
+                       << m_radio_mac;
+            return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
+        }
+        if (curr_channel_pool.size() > beerocks::message::SUPPORTED_CHANNELS_LENGTH) {
+            LOG(ERROR) << "channel_pool is too big [" << int(curr_channel_pool.size())
+                       << "] on mac=" << m_radio_mac;
+            return beerocks::eChannelScanErrCode::CHANNEL_SCAN_POOL_TOO_BIG;
+        }
+        request->scan_params().channel_pool_size = curr_channel_pool.size();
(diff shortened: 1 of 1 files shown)


controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -52,41 +52,54 @@ beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_t
...
+    //Get dwell time, if dwell time invalid fail scan.
     int32_t dwell_time_msec = database.get_channel_scan_dwell_time_msec(m_radio_mac, m_single_scan);
     if (dwell_time_msec <= 0) {
         LOG(ERROR) << "invalid dwell_time <= 0";
         return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
     }
+    request->scan_params().dwell_time_ms = dwell_time_msec;
 
-    //get current channel pool from DB
+    //Get channel pool, if channel pool empty fil scan, if pool is set to "all channel" use
```
arnout:
```suggestion
    // Get channel pool; if channel pool empty: fail scan; if pool is set to "all channel": use
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -52,41 +52,54 @@ beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_t
...
+        request->scan_params().channel_pool_size = supported_channels.size();
+        std::transform(supported_channels.begin(), supported_channels.end(),
+                       request->scan_params().channel_pool,
+                       [](beerocks::message::sWifiChannel &c) -> uint8_t { return c.channel; });
+    } else {
+        LOG(DEBUG) << "Using custom set channels for channel scan on mac=" << m_radio_mac;
+        if (curr_channel_pool.empty()) {
+            LOG(ERROR) << "empty channel pool is not supported. please set channel pool for mac="
+                       << m_radio_mac;
+            return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
```
arnout, line 89: I still don't understand why the empty channel pool can't be used as the marker... In this case, it's an error of the pool is empty. And the only other case where the channel pool is requested is to construct cACTION_BML_CHANNEL_SCAN_GET_CONTINUOUS_PARAMS_RESPONSE - whether we report an empty pool there or a pool with just zero shouldn't make a difference, should it?

Anyway I'll trust your judgement on this.

arnout (approved):

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: I've commented in the Linked Issue #1216
But basically the reason is that, currently on initialization the channel pool starts empty.
If an empty pool means "All Channels" then when we request a scan before setting the channel pool, the scan will trigger as if the user requested a scan on all channels since the initial pool is empty.
This is incorrect as we want the user to actively choose if they want to scan on all channels by setting it to a designated identifier, and fail to scan if no channel pool is set.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Fixed typos and updated comment to better describe what happens.

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -52,41 +52,55 @@ beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_t
 {
     // When a scan is requested send the scan parameters Channel pool & Dwell time
 
+    auto radio_mac_str = beerocks::net::network_utils::mac_to_string(m_radio_mac);
+
+    //Create request
```
tomereli: This comment is useless..

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -52,41 +52,55 @@ beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_t
...
     if (!request) {
         LOG(ERROR) << "Failed building message cACTION_CONTROL_CHANNEL_SCAN_TRIGGER_SCAN_REQUEST";
         return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INTERNAL_FAILURE;
     }
+    request->scan_params().radio_mac = m_radio_mac;
 
+    //Get dwell time, if dwell time invalid fail scan.
     int32_t dwell_time_msec = database.get_channel_scan_dwell_time_msec(m_radio_mac, m_single_scan);
     if (dwell_time_msec <= 0) {
         LOG(ERROR) << "invalid dwell_time <= 0";
```
tomereli:
```suggestion
        LOG(ERROR) << "invalid dwell_time=" << dwell_time_msec;
```

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -52,41 +52,55 @@ beerocks::eChannelScanErrCode dynamic_channel_selection_task::dcs_request_scan_t
...
         return beerocks::eChannelScanErrCode::CHANNEL_SCAN_INVALID_PARAMS;
     }
+    request->scan_params().dwell_time_ms = dwell_time_msec;
 
-    //get current channel pool from DB
+    // Get channel pool;
+    // if pool is set to "all channel": use the supported channels instead.
+    // if the costume channel pool is empty or too big: fail scan and return error accordingly;
```
tomereli:
```suggestion
    // if the custom channel pool is empty or too big: fail scan and return error accordingly;
```

tomereli (approved):
LGTM, but please fix the typos before merge

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Fixed

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Removed

Issue #1222, opened 2020-05-04
[BUG] CI: the rax40 firmware is missing kernel modules


rmelotte:
The full firmware built in CI is missing some kernel modules. For example, the module for wired interfaces is missing: `ltq_eth_drv_xrx500.ko`.

It can be seen from the `.config` file: `CONFIG_PACKAGE_kmod-intel_eth_drv_xrx500=y` is not available at all.

PR #1230, opened 2020-05-04
Hotfix/rax40 include kernel modules

rmelotte:
- Fix for https://github.com/prplfoundation/prplMesh/issues/1222 (remove the `tmp` folder before generating the full defconfig
- Save the full .config file in the artifacts

### DIFF
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -96,7 +96,6 @@ RUN cp feeds.conf.default feeds.conf \
     : "force regeneration by removing it"; \
     rm -rf tmp \
     && make defconfig \
-    && cat .config \
     && make -j$(nproc) \
     && make package/feeds/prpl/prplmesh/clean
     # note that the result from diffconfig.sh with a minimal
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -92,7 +92,10 @@ RUN cp feeds.conf.default feeds.conf \
     && if [ "$TARGET_PROFILE" = "DEVICE_NETGEAR_RAX40" ] ; then \
         grep -v -E 'CONFIG_SDK|CONFIG_MAKE_TOOLCHAIN' feeds/intel/rax40.profile >> .config ;\
     fi ;\
-    make defconfig \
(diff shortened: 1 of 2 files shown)


tools/docker/builder/openwrt/Dockerfile
```diff
@@ -92,7 +92,8 @@ RUN cp feeds.conf.default feeds.conf \
     && if [ "$TARGET_PROFILE" = "DEVICE_NETGEAR_RAX40" ] ; then \
         grep -v -E 'CONFIG_SDK|CONFIG_MAKE_TOOLCHAIN' feeds/intel/rax40.profile >> .config ;\
     fi ;\
-    make defconfig \
+    rm -rf tmp \
```
arnout, line 97: Adding an explanatory comment here would be nice. A little trick to add a comment:
```suggestion
    : "Installing intel feed doesn't correctly regenerate kernel .package-info"; \
    : "force regeneration by removing it"; \
    rm -rf tmp \
```

tools/docker/builder/openwrt/scripts/build.sh
```diff
@@ -16,3 +16,4 @@ PRPLMESH_VERSION=${PRPLMESH_VERSION}
 EOT
 find bin -name 'prplmesh*.ipk' -exec cp -v {} "artifacts/prplmesh.ipk" \;
 find bin/targets/"$TARGET_SYSTEM"/"$SUBTARGET"/ -type f -maxdepth 1 -exec cp -v {} "artifacts/" \;
+cp .config artifacts/
```
arnout: Does gitlab-ci render "hidden" files nicely? If not, we may have to rename it, e.g. to `openwrt.config`.

arnout (approved):
Nitpicks only.

tomereli (approved):

tools/docker/builder/openwrt/Dockerfile
rmelotte: Nice! There were multiple other occasions where I wanted to put comments in the middle like here, but I couldn't think of any way to do it.

I'll do this one here and remember it for the next time.

Thanks!

tools/docker/builder/openwrt/scripts/build.sh
rmelotte: I don't know how Gitlab renders them, but I applied your suggestion anyway.

PR #1231, opened 2020-05-04
agent: backhaul_manager: Implement AP Metrics Query/Response message

VladyslavTupikin:
Need to implement handlers for transmiting AP Metrics Query/Response messages
according to the UML diagram.

Add implementation for handlers for transmiting AP Metrics Query/Response messages.

### DIFF
--- a/agent/src/beerocks/monitor/monitor_stats.cpp
+++ b/agent/src/beerocks/monitor/monitor_stats.cpp
@@ -14,6 +14,7 @@
 
 #include <beerocks/tlvf/beerocks_message.h>
 #include <beerocks/tlvf/beerocks_message_monitor.h>
+#include <tlvf/wfa_map/tlvApMetrics.h>
 
 using namespace beerocks;
 using namespace net;
@@ -464,6 +465,31 @@ void monitor_stats::process()
             // containing one AP Metrics TLV for each of the BSSs on this radio.
         }
     }
+}
+
+bool monitor_stats::add_ap_metrics(ieee1905_1::CmduMessageTx &cmdu_tx, const sMacAddr &bssid)
+{
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetrics>();
+
+    if (!ap_metrics_response_tlv) {
+        LOG(ERROR) << "Couldn't addClass tlvApMetrics";
+        return false;
+    }
+
+    //TODO: fill ap_metrics_response_tlv with valid data (now valid just bssid_response)
+    ap_metrics_response_tlv->bssid()                                      = bssid;
+    ap_metrics_response_tlv->channel_utilization()                        = 10;
+    ap_metrics_response_tlv->number_of_stas_currently_associated()        = 2;
+    ap_metrics_response_tlv->estimated_service_parameters().include_ac_be = 1;
+
+    if (!ap_metrics_response_tlv->alloc_estimated_service_info_field(3)) {
+        LOG(ERROR)
+            << "Couldn't allocate ap_metrics_response_tlv->alloc_estimated_service_info_field";
+        return false;
+    }
+    std::fill_n(ap_metrics_response_tlv->estimated_service_info_field(), 3, 0);
+
+    return true;
 }
 
 void monitor_stats::calculate_client_load(monitor_sta_node *sta_node,
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -14,6 +14,8 @@
 
 #include <beerocks/tlvf/beerocks_message.h>
 
+#include <tlvf/wfa_map/tlvApMetricQuery.h>
+
 #include <cmath>
 #include <vector>
 
@@ -1336,6 +1338,8 @@ bool monitor_thread::handle_cmdu_ieee1905_1_message(Socket &sd, ieee1905_1::Cmdu
     auto cmdu_message_type = cmdu_rx.getMessageType();
 
     switch (cmdu_message_type) {
+    case ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE:
+        return handle_ap_metrics_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE:
         return handle_multi_ap_policy_config_request(sd, cmdu_rx);
     default:
@@ -1401,6 +1405,48 @@ bool monitor_thread::handle_multi_ap_policy_config_request(Socket &sd,
     }
 
     return true;
+}
+
+bool monitor_thread::handle_ap_metrics_query(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid           = cmdu_rx.getMessageId();
+    auto ap_metric_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+    if (!ap_metric_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
+        return false;
+    }
+
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building AP_METRICS_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    for (size_t bssid_idx = 0; bssid_idx < ap_metric_query_tlv->bssid_list_length(); bssid_idx++) {
+        auto bssid_tuple = ap_metric_query_tlv->bssid_list(bssid_idx);
+        if (!std::get<0>(bssid_tuple)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            return false;
+        }
+        const auto &bssid = std::get<1>(bssid_tuple);
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << bssid;
+        if (!mon_stats.add_ap_metrics(cmdu_tx, bssid)) {
+            return false;
+        }
+
+        //TODO: find the way how to check policy config was enabled or not
+        //if (policy config) {
+        //    auto tlv1 = mdu_tx.addClass<wfa_map::tlv>(); //STA tlvAssociatedStaLinkMetrics
+        //
+        //    //TODO add tlvTrafficStats
+        //}
+    }
+
+    LOG(DEBUG) << "Sending AP_METRICS_RESPONSE_MESSAGE to slave_socket, mid=" << std::hex
+               << int(mid);
+
+    return message_com::send_cmdu(slave_socket, cmdu_tx);
 }
 
 bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event_ptr)
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -669,13 +669,13 @@ bool monitor_thread::update_ap_stats()
 bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     if (cmdu_rx.getMessageType() == ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE) {
-        return handle_cmdu_vs_message(sd, cmdu_rx);
+        return handle_cmdu_vs_message(*sd, cmdu_rx);
     }
 
-    return handle_cmdu_ieee1905_1_message(sd, cmdu_rx);
+    return handle_cmdu_ieee1905_1_message(*sd, cmdu_rx);
 }
 
-bool monitor_thread::handle_cmdu_vs_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+bool monitor_thread::handle_cmdu_vs_message(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     auto beerocks_header = message_com::parse_intel_vs_message(cmdu_rx);
     if (beerocks_header == nullptr) {
@@ -686,7 +686,7 @@ bool monitor_thread::handle_cmdu_vs_message(Socket *sd, ieee1905_1::CmduMessageR
     if (beerocks_header->action() != beerocks_message::ACTION_MONITOR) {
         LOG(ERROR) << "Unsupported action: " << int(beerocks_header->action());
         return false;
-    } else if (slave_socket != sd) {
+    } else if (slave_socket != &sd) {
         LOG(ERROR) << "slave_socket != sd";
         return false;
     }
@@ -1331,7 +1331,7 @@ bool monitor_thread::handle_cmdu_vs_message(Socket *sd, ieee1905_1::CmduMessageR
     return true;
 }
 
-bool monitor_thread::handle_cmdu_ieee1905_1_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+bool monitor_thread::handle_cmdu_ieee1905_1_message(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     auto cmdu_message_type = cmdu_rx.getMessageType();
 
@@ -1345,7 +1345,7 @@ bool monitor_thread::handle_cmdu_ieee1905_1_message(Socket *sd, ieee1905_1::Cmdu
     return false;
 }
 
-bool monitor_thread::handle_multi_ap_policy_config_request(Socket *sd,
+bool monitor_thread::handle_multi_ap_policy_config_request(Socket &sd,
                                                            ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     /**
--- a/agent/src/beerocks/monitor/monitor_thread.h
+++ b/agent/src/beerocks/monitor/monitor_thread.h
@@ -125,6 +125,7 @@ private:
     bool handle_cmdu_ieee1905_1_message(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
 
     bool handle_multi_ap_policy_config_request(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
+    bool handle_ap_metrics_query(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
 };
 } // namespace son
 
--- a/agent/src/beerocks/monitor/monitor_thread.h
+++ b/agent/src/beerocks/monitor/monitor_thread.h
@@ -121,10 +121,10 @@ private:
     std::chrono::steady_clock::time_point m_sta_stats_polling_start_timestamp;
     bool m_sta_stats_polling_completed = true;
 
-    bool handle_cmdu_vs_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
-    bool handle_cmdu_ieee1905_1_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
+    bool handle_cmdu_vs_message(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
+    bool handle_cmdu_ieee1905_1_message(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
 
-    bool handle_multi_ap_policy_config_request(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
+    bool handle_multi_ap_policy_config_request(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
 };
 } // namespace son
 
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -65,6 +65,8 @@
 #include <tlvf/wfa_map/tlvApCapability.h>
 #include <tlvf/wfa_map/tlvApHeCapabilities.h>
 #include <tlvf/wfa_map/tlvApHtCapabilities.h>
+#include <tlvf/wfa_map/tlvApMetricQuery.h>
+#include <tlvf/wfa_map/tlvApMetrics.h>
 #include <tlvf/wfa_map/tlvApOperationalBSS.h>
 #include <tlvf/wfa_map/tlvApRadioBasicCapabilities.h>
 #include <tlvf/wfa_map/tlvApVhtCapabilities.h>
@@ -1633,7 +1635,11 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
             }
 
             if (soc && soc->slave == sd) {
-                return handle_slave_backhaul_message(soc, cmdu_rx);
+                if (cmdu_rx.getMessageType() == ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE) {
+                    return handle_slave_backhaul_message(soc, cmdu_rx);
+                } else {
+                    return handle_slave_1905_1_message(cmdu_rx, src_mac);
+                }
             } else {
                 LOG(ERROR) << "ACTION_BACKHAUL from none slave socket!";
                 return false;
@@ -2117,10 +2123,26 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     }
     case ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE: {
         return handle_multi_ap_policy_config_request(cmdu_rx, src_mac);
+    case ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE: {
+        return handle_ap_metrics_query(cmdu_rx, src_mac);
     }
     default: {
         // TODO add a warning once all vendor specific flows are replaced with EasyMesh
         // flows, since we won't expect a 1905 message not handled in this function
+        return false;
+    }
+    }
+    }
+}
+bool backhaul_manager::handle_slave_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                   const std::string &src_mac)
+{
+    switch (cmdu_rx.getMessageType()) {
+    case ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE: {
+        return handle_slave_ap_metrics_response(cmdu_rx, src_mac);
+    }
+    default: {
+        LOG(DEBUG) << "Unexpected 1905 message " << int(cmdu_rx.getMessageType());
         return false;
     }
     }
@@ -2410,6 +2432,162 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
 
     LOG(DEBUG) << "Sending AP_CAPABILITY_REPORT_MESSAGE , mid: " << std::hex << (int)mid;
     return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
+}
+
+bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                               const std::string &src_mac)
+{
+    std::vector<sMacAddr> bssid;
+    const auto mid           = cmdu_rx.getMessageId();
+    auto ap_metric_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+    if (!ap_metric_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
+        return false;
+    }
+    for (size_t bssid_idx = 0; bssid_idx < ap_metric_query_tlv->bssid_list_length(); bssid_idx++) {
+        auto bssid_tuple = ap_metric_query_tlv->bssid_list(bssid_idx);
+        if (!std::get<0>(bssid_tuple)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            return false;
+        }
+        bssid.push_back(std::get<1>(bssid_tuple));
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << std::get<1>(bssid_tuple);
+    }
+
+    auto ret = false;
+    for (auto socket : slaves_sockets) {
+        for (const auto &mac : bssid) {
+            int i = 0;
+            if (mac == socket->vaps_list.vaps[i].mac) {
+                LOG(DEBUG) << "Forwarding AP_METRICS_QUERY_MESSAGE message to son_slave, bssid: "
+                           << std::hex << network_utils::mac_to_string(mac);
+
+                auto forward =
+                    cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
+                if (!forward) {
+                    LOG(ERROR) << "Failed to create AP_METRICS_QUERY_MESSAGE";
+                    return false;
+                }
+
+                auto query = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
+                if (!query) {
+                    LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetricQuery>";
+                    return false;
+                }
+
+                if (!query->alloc_bssid_list(1)) {
+                    LOG(ERROR) << "Failed allocate memory for bssid_list";
+                    return false;
+                }
+
+                auto list         = query->bssid_list(0);
+                std::get<0>(list) = true;
+                std::get<1>(list) = mac;
+
+                if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
+                    LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
+                    ret = false;
+                    continue;
+                } else {
+                    ret = true;
+                    // Fill a query vector
+                    m_ap_metric_query.push_back({socket->slave, mac});
+                }
+            }
+            i++;
+        }
+    }
+    return ret;
+}
+
+bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
+
+    auto ap_metrics_tlv = cmdu_rx.getClass<wfa_map::tlvApMetrics>();
+    if (!ap_metrics_tlv) {
+        LOG(ERROR) << "Failed cmdu_rx.getClass<wfa_map::tlvApMetrics>(), mid=" << std::hex
+                   << int(mid);
+        return false;
+    }
+
+    auto bssid_tlv = ap_metrics_tlv->bssid();
+    auto mac       = std::find_if(
+        m_ap_metric_query.begin(), m_ap_metric_query.end(),
+        [&bssid_tlv](sApMetricsQuery const &query) { return query.bssid == bssid_tlv; });
+
+    if (mac == m_ap_metric_query.end()) {
+        LOG(ERROR) << "Failed search in ap_metric_query for bssid: " << bssid_tlv
+                   << " from mid=" << std::hex << int(mid);
+        return false;
+    }
+
+    sApMetrics metric;
+    // Copy data to the response vector
+    metric.bssid               = ap_metrics_tlv->bssid();
+    metric.channel_utilization = ap_metrics_tlv->channel_utilization();
+    metric.number_of_stas_currently_associated =
+        ap_metrics_tlv->number_of_stas_currently_associated();
+    metric.estimated_service_parameters = ap_metrics_tlv->estimated_service_parameters();
+    auto info                           = ap_metrics_tlv->estimated_service_info_field();
+    if (ap_metrics_tlv->estimated_service_info_field_length()) {
+        for (size_t i = 0; i < ap_metrics_tlv->estimated_service_info_field_length(); i++) {
+            metric.estimated_service_info_field.push_back(info[i]);
+        }
+    }
+
+    // Fill a response vector
+    m_ap_metric_response.push_back({metric});
+
+    // Remove an entry from the processed query
+    m_ap_metric_query.erase(
+        std::remove_if(m_ap_metric_query.begin(), m_ap_metric_query.end(),
+                       [&](sApMetricsQuery const &query) { return mac->bssid == query.bssid; }),
+        m_ap_metric_query.end());
+
+    if (!m_ap_metric_query.empty()) {
+        return true;
+    }
+
+    // We received all responses - prepare and send response message to the controller
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    // Prepare tlvApMetrics for each processed query
+    for (const auto &response : m_ap_metric_response) {
+        auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetrics>();
+        if (!ap_metrics_response_tlv) {
+            LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetrics>";
+            return false;
+        }
+
+        ap_metrics_response_tlv->bssid()               = response.metric.bssid;
+        ap_metrics_response_tlv->channel_utilization() = response.metric.channel_utilization;
+        ap_metrics_response_tlv->number_of_stas_currently_associated() =
+            response.metric.number_of_stas_currently_associated;
+        ap_metrics_response_tlv->estimated_service_parameters() =
+            response.metric.estimated_service_parameters;
+        if (!ap_metrics_response_tlv->alloc_estimated_service_info_field(
+                response.metric.estimated_service_info_field.size())) {
+            LOG(ERROR) << "Couldn't allocate "
+                          "ap_metrics_response_tlv->alloc_estimated_service_info_field";
+            return false;
+        }
+        std::copy_n(response.metric.estimated_service_info_field.begin(),
+                    response.metric.estimated_service_info_field.size(),
+                    ap_metrics_response_tlv->estimated_service_info_field());
+    }
+
+    LOG(DEBUG) << "Sending AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
+
+    return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 }
 
 /**
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -24,6 +24,8 @@
 #include <tlvf/ieee_1905_1/eLinkMetricsType.h>
 #include <tlvf/ieee_1905_1/eMediaType.h>
 
+#include <tlvf/wfa_map/tlvApMetrics.h>
+
 #include "../agent_ucc_listener.h"
 #include "../link_metrics/link_metrics.h"
 
@@ -94,7 +96,8 @@ private:
     bool handle_slave_backhaul_message(std::shared_ptr<sRadioInfo> soc,
                                        ieee1905_1::CmduMessageRx &cmdu_rx);
     bool handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx, const std::string &src_mac);
-
+    bool handle_slave_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                     const std::string &src_mac);
     // 1905 messages handlers
     bool handle_1905_topology_discovery(const std::string &src_mac,
                                         ieee1905_1::CmduMessageRx &cmdu_rx);
@@ -114,6 +117,9 @@ private:
                                                   const std::string &src_mac);
     bool handle_multi_ap_policy_config_request(ieee1905_1::CmduMessageRx &cmdu_rx,
                                                const std::string &src_mac);
+    bool handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx, const std::string &src_mac);
+    bool handle_slave_ap_metrics_response(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                          const std::string &src_mac);
     //bool sta_handle_event(const std::string &iface,const std::string& event_name, void* event_obj);
     bool hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event_ptr, std::string iface);
 
@@ -502,6 +508,26 @@ private:
     bool add_link_metrics(const sMacAddr &reporter_al_mac, const sLinkInterface &link_interface,
                           const sLinkNeighbor &link_neighbor, const sLinkMetrics &link_metrics,
                           ieee1905_1::eLinkMetricsType link_metrics_type);
+
+    struct sApMetricsQuery {
+        Socket *soc;
+        sMacAddr bssid;
+    };
+
+    struct sApMetrics {
+        sMacAddr bssid;
+        uint8_t channel_utilization;
+        uint16_t number_of_stas_currently_associated;
+        wfa_map::tlvApMetrics::sEstimatedService estimated_service_parameters;
+        std::vector<uint8_t> estimated_service_info_field;
+    };
+
+    struct sApMetricsResponse {
+        sApMetrics metric;
+    };
+
+    std::vector<sApMetricsQuery> m_ap_metric_query;
+    std::vector<sApMetricsResponse> m_ap_metric_response;
 
     /*
  * State Machines
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -27,6 +27,7 @@
 #include <tlvf/WSC/AttrList.h>
 #include <tlvf/ieee_1905_1/tlvAlMacAddressType.h>
 #include <tlvf/ieee_1905_1/tlvLinkMetricQuery.h>
+#include <tlvf/ieee_1905_1/tlvMacAddress.h>
 #include <tlvf/ieee_1905_1/tlvSupportedFreqBand.h>
 #include <tlvf/ieee_1905_1/tlvSupportedRole.h>
 #include <tlvf/wfa_map/tlvApMetricQuery.h>
@@ -42,7 +43,6 @@
 #include <tlvf/wfa_map/tlvSteeringBTMReport.h>
 #include <tlvf/wfa_map/tlvSteeringRequest.h>
 #include <tlvf/wfa_map/tlvTransmitPowerLimit.h>
-
 // BPL Error Codes
 #include <bpl/bpl_cfg.h>
 #include <bpl/bpl_err.h>
@@ -337,6 +337,8 @@ bool slave_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
             LOG(ERROR) << "Unknown message, action: " << int(beerocks_header->action());
         }
         }
+    } else if (sd == monitor_socket) {
+        handle_cmdu_monitor_ieee1905_1_message(*sd, cmdu_rx);
     } else { // IEEE 1905.1 message
         return handle_cmdu_control_ieee1905_1_message(sd, cmdu_rx);
     }
@@ -356,8 +358,7 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
         LOG(WARNING) << "master_socket == nullptr";
         return true;
     } else if (master_socket != sd) {
-        LOG(WARNING) << "Unknown socket, cmdu message type: " << int(cmdu_message_type);
-        return true;
+        LOG(DEBUG) << "Unknown socket, cmdu message type: " << int(cmdu_message_type); //TODO:
     }
 
     if (slave_state == STATE_STOPPED) {
@@ -376,7 +377,7 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
     case ieee1905_1::eMessageType::CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE:
         return handle_client_association_request(sd, cmdu_rx);
     case ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE:
-        return handle_ap_metrics_query(sd, cmdu_rx);
+        return handle_ap_metrics_query(*sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_PREFERENCE_QUERY_MESSAGE:
         return handle_channel_preference_query(sd, cmdu_rx);
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE:
@@ -393,6 +394,19 @@ bool slave_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
     }
 
     return true;
+}
+
+bool slave_thread::handle_cmdu_monitor_ieee1905_1_message(Socket &sd,
+                                                          ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto cmdu_message_type = cmdu_rx.getMessageType();
+    switch (cmdu_message_type) {
+    case ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE:
+        return handle_monitor_ap_metrics_response(sd, cmdu_rx);
+    default:
+        LOG(ERROR) << "Unknown CMDU message type: " << std::hex << int(cmdu_message_type);
+        return false;
+    }
 }
 
 bool slave_thread::handle_cmdu_control_message(Socket *sd,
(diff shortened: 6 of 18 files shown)


agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -653,12 +656,62 @@ bool monitor_thread::update_ap_stats()
...
+
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetric>();
+    LOG(DEBUG) << "Sending AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
+    return message_com::send_cmdu(slave_socket, cmdu_tx);
+}
+
+bool monitor_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
+                                                            ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto cmdu_message_type = cmdu_rx.getMessageType();
```
RanRegev: better put it after the next if statement, because the value is not used if the if is true.
The if statement is "escape" if.

agent/src/beerocks/monitor/monitor_thread.h
```diff
@@ -54,6 +54,9 @@ class monitor_thread : public beerocks::socket_thread {
     virtual std::string print_cmdu_types(const beerocks::message::sUdsHeader *cmdu_header) override;
 
 private:
+    bool handle_cmdu_control_ieee1905_1_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
```
RanRegev: I noticed you escape this function if sd is nullptr. In this case, if you don't know how to work with a nullptr Socket, please _require_ it in your interface. The best would be:
```suggestion
    bool handle_cmdu_control_ieee1905_1_message(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2313,6 +2319,66 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
 }
 
+bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                               const std::string &src_mac)
+{
+    std::vector<sMacAddr> bssid;
+    const auto mid            = cmdu_rx.getMessageId();
+    auto ap_metrics_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+    if (!ap_metrics_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
```
RanRegev, line 2444: Good! the mid is a hint you added for later investigation.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2313,6 +2319,66 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+                   << "  bssid " << std::get<1>(bssid_tupl);
+    }
+
+    auto cmdu_header = cmdu_tx.create(0, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_QUERY_MESSAGE";
+        return false;
+    }
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
```
RanRegev: These two lines are extremely weird. I guess they are OK because addCLass returns a pointer to internal member and then you set the value of this member. But please note that the structure of these two lines is like writing:
```cpp
A a;
auto x = a.x;
x = a.y;
```
which is strange.

RanRegev: Also I don't understand why you named it "response".
This is not a response, you _forward_ the message.
```suggestion
    auto ap_metrics_query_tlv = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2313,6 +2319,66 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+        if (!std::get<0>(bssid_tupl)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            return false;
+        }
+        bssid.push_back(std::get<1>(bssid_tupl));
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << std::get<1>(bssid_tupl);
+    }
+
+    auto cmdu_header = cmdu_tx.create(0, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
```
RanRegev: Why don't you use the same mid as received line 2326?
I mean like:
```suggestion
    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2313,6 +2319,66 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
     return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
```
RanRegev, line 2437: Unless I'm missing something crucial - this function and the slave_thread function with the same name (6f9a662#diff-f3dcf05057f5ecc09b7bc39daa3fe7f2R4528) are doing the same:

- verification on the rx
- build tx from the data inthe rx.

Maybe it would be better to separate this functionality into one function that both threads would use?

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -653,12 +656,62 @@ bool monitor_thread::update_ap_stats()
...
+}
+
 bool monitor_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
 {
     auto beerocks_header = message_com::parse_intel_vs_message(cmdu_rx);
     if (beerocks_header == nullptr) {
-        LOG(ERROR) << "Not a vendor specific message";
-        return false;
+        LOG(DEBUG) << "Received IEEE 1905.1 message";
+        handle_cmdu_control_ieee1905_1_message(sd, cmdu_rx);
```
mariomaz: I've also had to implement this sort of differentiating and dispatching vendor-specific messages and IEEE 1905.1 messages. However, I've done it in a slightly different way, by checking the type of message. At this point we'll have a merge conflict and we should agree on which mechanism should be used.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -653,12 +656,62 @@ bool monitor_thread::update_ap_stats()
...
+bool monitor_thread::handle_ap_metrics_query(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid            = cmdu_rx.getMessageId();
+    auto ap_metrics_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+
+    if (!ap_metrics_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
+        return false;
+    }
+    for (int bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
```
mariomaz: Type of bssid_idx should be `size_t`

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -653,12 +656,62 @@ bool monitor_thread::update_ap_stats()
...
+        auto bssid = ap_metrics_query_tlv->bssid_list(bssid_idx);
+        if (!std::get<0>(bssid)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            //return false;
```
mariomaz: This line should not be commented out.
You cannot use the second element of the tuple if the first one is false.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2313,6 +2319,66 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+                                               const std::string &src_mac)
+{
+    std::vector<sMacAddr> bssid;
+    const auto mid            = cmdu_rx.getMessageId();
+    auto ap_metrics_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+    if (!ap_metrics_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
+        return false;
+    }
+    for (int bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
```
mariomaz: Use `size_t`

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2313,6 +2319,66 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+
+    auto cmdu_header = cmdu_tx.create(0, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_QUERY_MESSAGE";
+        return false;
+    }
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
+    ap_metrics_response_tlv      = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+    for (auto socket : slaves_sockets) {
```
mariomaz, line 2459: I've also had to forward received message to slaves, but I did it in a different way using `message_com::forward_cmdu_to_uds`. which I think is more convenient than creating a new copy of the message.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2313,6 +2319,66 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetric>();
+    ap_metrics_response_tlv      = ap_metrics_query_tlv;
+    LOG(DEBUG) << "Sending AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
+
+    return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
```
mariomaz, line 2590: I think that backhaul manager should wait until responses from all the slaves had been received. At that moment, backhaul manager should create and send a new response message with metrics received from all requested radios.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4534,14 +4537,41 @@ bool slave_thread::handle_ap_metrics_query(Socket *sd, ieee1905_1::CmduMessageRx
...
     }
+    LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex
+               << int(mid); //TODO: remove this string after solving problem with query tlv.
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_QUERY_MESSAGE";
+        return false;
+    }
+    message_com::send_cmdu(monitor_socket, cmdu_tx);
```
mariomaz: I've also used `message_com::forward_cmdu_to_uds` here to forward message to monitor thread.

mariomaz (changes requested):
We have used different approaches to implement similar things (see #1233). We should agree on which is the most appropriate way before continuing.

agent/src/beerocks/monitor/monitor_thread.cpp
arnout: Well, bssid_list_length() is a uint8_t, so if anything, bssid_idx should be uint8_t...

arnout: Never mind, size_t is better.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -653,12 +656,62 @@ bool monitor_thread::update_ap_stats()
...
+bool monitor_thread::handle_cmdu_control_ieee1905_1_message(Socket *sd,
+                                                            ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto cmdu_message_type = cmdu_rx.getMessageType();
+
+    if (!sd) {
+        LOG(ERROR) << "Socket error";
+        return false;
+    }
+
```
arnout: (nitpick) We don't do such a check anywhere else in this file. Actually, it should really be passed by reference instead of as a pointer.

agent/src/beerocks/monitor/monitor_thread.cpp
arnout: @mariomaz If you think your mechanism is better, please make a quick PR with only that commit, and Vlad can rebase on it.

agent/src/beerocks/monitor/monitor_thread.h
arnout: I agree, but note that this is a shortcoming in this entire file. That said, for new functions we can definitely do the right thing.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2092,6 +2094,10 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     case ieee1905_1::eMessageType::ASSOCIATED_STA_LINK_METRICS_QUERY_MESSAGE: {
         return handle_associated_sta_link_metrics_query(cmdu_rx, src_mac);
     }
+    case ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE:
+        return handle_ap_metrics_query(cmdu_rx, src_mac);
+    case ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE:
+        return handle_ap_metrics_response(cmdu_rx, src_mac);
```
arnout: I don't think this is needed. This function is only called when the message comes from the bus, not when it comes from a slave. Instead, you have to add the handling to handle_slave_backhaul_message() (which at the moment is also only supporting Intel VS messages).

arnout (commented):
Not a real review, just pointing out where the bug is.

agent/src/beerocks/monitor/monitor_thread.cpp
VladyslavTupikin: We do it in monitor_thread::socket_disconnected

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: Done, see #1236.
New mechanism is also implemented in #1233

VladyslavTupikin: So we do not have conflicts, @mariomaz please check updated changes

agent/src/beerocks/monitor/monitor_thread.cpp
VladyslavTupikin: outdated

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: applied when forward required

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Implemented this mechanism in the last update

framework/tlvf/src/src/CmduMessageRx.cpp
```diff
@@ -219,6 +224,7 @@ std::shared_ptr<BaseClass> CmduMessageRx::parseNextTlv()
         return msg.addClass<wfa_map::tlvApCapability>();
     }
     default: {
+        LOG(DEBUG) << "tlvUnknown message type: " << tlv_type;
```
RanRegev: great.

tests/test_flows.py
```diff
@@ -493,51 +493,33 @@ def test_link_metric_query(self):
     def test_combined_infra_metrics(self):
         debug("Send AP Metrics query message to agent 1")
         env.controller.dev_send_1905(env.agents[0].mac, 0x800B,
-                                     tlv(0x93, 0x0007, "0x01 {%s}" % (env.agents[0].radios[0].mac)))
+                                     tlv(0x93, 0x0007, "0x01 {%s}" % (env.agents[0].radios[0].vaps[0].bssid)))  # (env.agents[0].radios[0].vaps[0].bssid)))
```
RanRegev: The comment at the end of the line should be removed.

tests/test_flows.py
```diff
@@ -493,51 +493,33 @@ def test_link_metric_query(self):
...
-        debug("Send 1905 Link metric query to agent 1 (neighbor gateway)")
-        env.controller.dev_send_1905(env.agents[0].mac, 0x0005,
-                                     tlv(0x08, 0x0008, "0x01 {%s} 0x02" % env.controller.mac))
-        self.check_log(env.agents[0], "Received LINK_METRIC_QUERY_MESSAGE")
-        self.check_log(env.controller, "Received LINK_METRIC_RESPONSE_MESSAGE")
-        self.check_log(env.controller, "Received TLV_TRANSMITTER_LINK_METRIC")
-        self.check_log(env.controller, "Received TLV_RECEIVER_LINK_METRIC")
+        self.check_log(env.controller, "Received AP_METRICS_RESPONSE_MESSAGE")
 
+        #debug("Send 1905 Link metric query to agent 1 (neighbor gateway)")
```
RanRegev: The final file seems to be ok: you send a request and looks for request and response in the log.
also you mentioned that the code is _temporary_ closed, which I guess means later the code would be removed.
this is fine. 
the only thing that I would do is add a comment above the commented line explaining that these lines should be removed in the future after \<this date\> / \<this event\> / \<before merge\> / etc.  Just for the case they will wrongly be left there.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -666,6 +670,87 @@ bool monitor_thread::update_ap_stats()
     return true;
 }
 
+bool monitor_thread::handle_ap_metrics_query(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid            = cmdu_rx.getMessageId();
+    auto ap_metrics_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+    sMacAddr bssid_response;
```
RanRegev: Move after the escape `if` and before the `for` it is being used

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -666,6 +670,87 @@ bool monitor_thread::update_ap_stats()
...
+    if (!ap_metrics_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
+        return false;
+    }
+    for (size_t bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
```
RanRegev: This seems like validation and log `for` loop. If so please comment above it accordingly.
Also the last line of the loop seems to be useless, the value in `bssid_response` after the loop would be of the _last_ in ap_metrics_query_tlv list. If this is the *desired* behavior it must be well documented as this is not natural.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -666,6 +670,87 @@ bool monitor_thread::update_ap_stats()
...
+        auto bssid = ap_metrics_query_tlv->bssid_list(bssid_idx);
+        if (!std::get<0>(bssid)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            return false;
+        }
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << std::get<1>(bssid);
+        bssid_response = std::get<1>(bssid);
+    }
+    //TODO: add
```
RanRegev: confusing. not sure what is there is TODO. what should be added.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -666,6 +670,87 @@ bool monitor_thread::update_ap_stats()
...
+    if (!ap_metrics_response_tlv) {
+        LOG(ERROR) << "Couldn't addClass tlvApMetric";
+    }
+
+    //TODO: fill ap_metrics_response_tlv with valid data (now valid just bssid_response)
+    uint8_t buf[10];
+    wfa_map::tlvApMetric::sEstimatedService estimated_service_parameters;
+    estimated_service_parameters.include_ac_be = 1;
+
+    ap_metrics_response_tlv->bssid()                               = bssid_response;
```
RanRegev: see comment above, bssid_response in the last in the list.
If this is ok, please comment the code correctly.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1633,11 +1635,12 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
                 }
             }
 
-            if (soc && soc->slave == sd) {
+            if (soc && soc->slave == sd &&
```
RanRegev: I still feel uncomfortable with this `if` statement, I don't have a suggestion, I know what I would like:
I want to be able to clearly understand the conditions that led me to a point in the code.
In the way it is written now, we get to `handle_slave_1905_1_message` if either is true:
* !soc
* !soc->slave
*  cmdu_rx.getMessageType() is not a VENDOR_SPECIFIC_MESSAGE

I don't have a suggestion.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2118,13 +2121,29 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
...
+}
+bool backhaul_manager::handle_slave_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                   const std::string &src_mac)
+{
+    switch (cmdu_rx.getMessageType()) {
+    case ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE: {
+        return handle_ap_metrics_response(cmdu_rx, src_mac);
+    }
+    default: {
+        LOG(DEBUG) << "Unknown 1905 message";
```
RanRegev: also would be great to add the message type for the error log.
```suggestion
        LOG(DEBUG) << "Unknown 1905 message:" << cmdu_rx.getMessageType()
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    if (!ap_metric_tlv) {
+        LOG(ERROR) << "Failed cmdu_rx.getClass<wfa_map::tlvApMetric>(), mid=" << std::hex
+                   << int(mid);
+        return false;
+    }
+
+    auto bssid_tlv = ap_metric_tlv->bssid();
+    auto mac       = std::find_if(
+        ap_metric_query.begin(), ap_metric_query.end(),
+        [&](sApMetricsQuery const &ap_metric_query) { return ap_metric_query.bssid == bssid_tlv; });
```
RanRegev: It is better to be specific, just for clarity -
```suggestion
        [&bssid_tlv](sApMetricsQuery const &ap_metric_query) { return ap_metric_query.bssid == bssid_tlv; });
```
(However, I don't have strong opinion about this issue)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+                                             return mac->bssid == ap_metric_query.bssid;
+                                         }),
+                          ap_metric_query.end());
+
+    processed_queries++;
+    if (!ap_metric_query.empty()) {
+        return true;
+    }
+
+    // If all queries processed - prepare and send response message to the controller
```
RanRegev:
```suggestion
    // at this point all queries processed - prepare and send response message to the controller
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
```
RanRegev, line 2556: If I am not missing something, `mid` is the mid of the _last_ received response. Is this correct?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: ... However, In case of error, when not all responses were received, `ap_metric_query`  is left in an unstable state - it contains the metrics queries that did not receive any response.
This situation actually stops any farther queries from succeeding as the case to send a response is the emptiness of this vector (see above, line 2548)

RanRegev (approved):
I have few considerations that need to be answered before merging

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -133,7 +133,7 @@ class slave_thread : public beerocks::socket_thread {
     bool handle_cmdu_monitor_message(Socket *sd,
                                      std::shared_ptr<beerocks::beerocks_header> beerocks_header);
     bool handle_cmdu_control_ieee1905_1_message(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
-
+    bool handle_cmdu_monitor_ieee1905_1_message(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
```
arnout, line 136: Excellent naming!

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -238,7 +238,8 @@ class slave_thread : public beerocks::socket_thread {
     bool handle_autoconfiguration_renew(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
     bool autoconfig_wsc_add_m1();
     bool send_operating_channel_report();
-    bool handle_ap_metrics_query(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_rx);
+    bool handle_ap_metrics_query(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
+    bool handle_ap_metrics_response(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx);
```
arnout: Might be useful to include the fact that it comes from monitor in the name here...

tests/test_flows.py
```diff
@@ -493,51 +493,33 @@ def test_link_metric_query(self):
     def test_combined_infra_metrics(self):
         debug("Send AP Metrics query message to agent 1")
         env.controller.dev_send_1905(env.agents[0].mac, 0x800B,
-                                     tlv(0x93, 0x0007, "0x01 {%s}" % (env.agents[0].radios[0].mac)))
+                                     tlv(0x93, 0x0007, "0x01 {%s}" % (env.agents[0].radios[0].vaps[0].bssid)))  # (env.agents[0].radios[0].vaps[0].bssid)))
         self.check_log(env.agents[0].radios[0], "Received AP_METRICS_QUERY_MESSAGE")
         # TODO agent should send response autonomously, with same MID.
```
arnout: TODO should be removed.

Conversely, we should use the sniffer to verify that the agent sends the expected AP Metrics Response.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1633,11 +1635,12 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
...
 
-            if (soc && soc->slave == sd) {
+            if (soc && soc->slave == sd &&
+                cmdu_rx.getMessageType() == ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE) {
+
                 return handle_slave_backhaul_message(soc, cmdu_rx);
             } else {
-                LOG(ERROR) << "ACTION_BACKHAUL from none slave socket!";
-                return false;
+                return handle_slave_1905_1_message(cmdu_rx, src_mac);
```
arnout: This should also be inside the `if (soc && soc->slave == sd)` condition. The second branch of that condition is actually redundant, because the loop above sets `soc` to non-nullptr only if `soc->slave == sd`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
     return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                               const std::string &src_mac)
+{
+    std::vector<sMacAddr> bssid;
+    auto ret                  = false;
```
arnout: In the context of the "multiple points of exit" principle, something like this is suspicious. I'd rather expect patterns like:
```
if (!func()) {
   LOG(ERROR) << ...;
   return false;
}
```
So no `ret` is needed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    const auto mid            = cmdu_rx.getMessageId();
+    auto ap_metrics_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
+    if (!ap_metrics_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
+        return ret;
```
arnout: So here 
```suggestion
        return false;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    }
+    for (size_t bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
+        auto bssid_tupl = ap_metrics_query_tlv->bssid_list(bssid_idx);
```
arnout: (nitpick) 
```suggestion
        auto bssid_tuple = ap_metrics_query_tlv->bssid_list(bssid_idx);
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -503,6 +508,28 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+        std::vector<uint8_t> estimated_service_info_field;
+    } sApMetric;
+    typedef struct sApMetricsResponse {
+        Socket *soc;
+        sMacAddr bssid;
+        sApMetric metric;
+    } sApMetricsResponse;
+
+    size_t processed_queries = 0;
+    std::vector<sApMetricsQuery> ap_metric_query;
```
arnout:
```suggestion
    std::vector<sApMetricsQuery> m_ap_metric_query;
```
(same for the others)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+        }
+        bssid.push_back(std::get<1>(bssid_tupl));
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << std::get<1>(bssid_tupl);
+    }
+
+    for (auto socket : slaves_sockets) {
+        for (auto mac : bssid) {
+            int i = 0;
+            if (mac == socket->vaps_list.vaps[i].mac) {
```
arnout: It would actually make more sense to have a structure like this:
```
for (size_t bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
    const auto &bssid = ...
    auto slave = find_slave_for_bssid(bssid);
    if (!slave) {
       // Error, send ACK with error code
    }
    if (slave not in m_ap_metric_query) {
        forward_cmdu
        add slave to m_ap_metric_query
    }
}
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+                LOG(DEBUG) << "Forwarding AP_METRICS_QUERY_MESSAGE message to son_slave, bssid: "
+                           << std::hex << network_utils::mac_to_string(mac);
```
arnout, line 2464: std::hex is useless here. So is the mac_to_string.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+
+                if (!query->alloc_bssid_list(1)) {
+                    LOG(ERROR) << "Failed allocate memory for bssid_list";
+                }
+
+                auto list         = query->bssid_list(0);
+                std::get<0>(list) = true;
+                std::get<1>(list) = mac;
+
+                if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
```
arnout: Instead of forwarding a separate query with a single bssid, I think we should forward it as is.
```suggestion
                if (!message_com::forward_cmdu_to_slave(socket->slave, cmdu_rx, length)) {
```

tests/test_flows.py
mariomaz: This TODO comment should be removed

tests/test_flows.py
mariomaz: You'd better create an issue describing the work left to be done and copy the issue number in a TODO comment here

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: My PRs have already been merged onto master. You should rebase your changes

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -666,6 +670,87 @@ bool monitor_thread::update_ap_stats()
...
+    }
+    //TODO: add
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetric>();
```
mariomaz: I think you should move this code out to a method. This way we'll be able to reuse it when channel utilization threshold is crossed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Why aren't you using `message_com::forward_cmdu_to_uds`?
I think code will be much simpler

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -503,6 +508,28 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+    std::vector<sApMetricsResponse> ap_metric_response;
```
mariomaz: According to coding style rules, class data members should have an m_ prefix: m_processed_queries, m_ap_metric_query, m_ap_metric_response

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+                    LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
+                    ret = false;
+                    continue;
```
mariomaz, line 2491: If received query message cannot be forwarded to one of the slaves, then this method should return false here.
And before that, probably rollback by cleaning ap_metric_query vector. This way, responses from slaves to which received query message has already been forwarded to will be ignored.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+                } else {
+                    ret = true;
+                    // Fill a query vector
+                    ap_metric_query.push_back({socket->slave, mac});
```
mariomaz: You are not considering the case in which several AP_METRICS_QUERY_MESSAGE messages are received in a row, before their responses from slaves had been processed.
Maybe you should store the mid of the AP_METRICS_QUERY_MESSAGE to discard responses from slaves which belong to a previous request.

mariomaz (changes requested):
Request/response mechanism should be more robust

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    // If all queries processed - prepare and send response message to the controller
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    // Prepare tlvApMetric for each processed query
+    for (size_t i = 0; i < processed_queries; i++) {
```
arnout: We should just iterate over ap_metric_response here.
```suggestion
    for (response : ap_metric_response) {
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+        auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetric>();
+        if (!ap_metrics_response_tlv) {
+            LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetric>";
+            return false;
+        }
+        LOG(DEBUG) << "VT_back in tlv loop ";
```
arnout: I'm not sure what this is supposed to mean.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -503,6 +508,28 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+
+    typedef struct sApMetric {
+        sMacAddr bssid;
+        uint8_t channel_utilization;
+        uint16_t number_of_stas_currently_associated;
+        wfa_map::tlvApMetric::sEstimatedService estimated_service_parameters;
+        std::vector<uint8_t> estimated_service_info_field;
+    } sApMetric;
+    typedef struct sApMetricsResponse {
+        Socket *soc;
```
arnout: The soc field is not used for anything, so it can be removed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -503,6 +508,28 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+        sMacAddr bssid;
```
arnout: The bssid field is not used for anything, so it can be removed.

Bottom line: we don't need this structure, we can use metric directly.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2413,6 +2432,163 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+            it->metric.number_of_stas_currently_associated;
+        ap_metrics_response_tlv->estimated_service_parameters() =
+            it->metric.estimated_service_parameters;
+        if (!ap_metrics_response_tlv->alloc_estimated_service_info_field(
+                it->metric.estimated_service_info_field.size())) {
+            LOG(ERROR) << "Couldn't allocate "
+                          "ap_metrics_response_tlv->alloc_estimated_service_info_field";
+            return false;
+        }
+        auto tmp = ap_metrics_response_tlv->estimated_service_info_field();
```
arnout: IMO this tmp variable is not helpful.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: So I think a better approach would be to have a separate structure that contains:
- the list of slaves from which we expect a response;
- for each slave, the bssids for which we expect a response;
- the mid of the expected response (to distinguish from duplicate queries);
- the list of received response information.

When we receive a query, this structure is reset.

When we receive a response that is not expected, it is forwarded as-is to the controller.

This still leaves a weird situation when there are two requests back to back: since the "expected" of the first request is discarded, any responses that we receive are forwarded as is to the controller. But this is a situation that shouldn't happen, so any behaviour is fine as long as we don't crash and at least one of the requests is handled correctly.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4553,6 +4568,38 @@ bool slave_thread::handle_ap_metrics_query(Socket *sd, ieee1905_1::CmduMessageRx
...
+bool slave_thread::handle_ap_metrics_response(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
+
+    auto ap_metrics_response_tlv = cmdu_rx.getClass<wfa_map::tlvApMetric>();
+    if (!ap_metrics_response_tlv) {
+        LOG(ERROR) << "Failed cmdu_rx.getClass<wfa_map::tlvApMetric>()" << std::hex << int(mid);
+        return false;
+    }
```
arnout: Since we don't do anything with this, it's not needed to check it. Just forward it.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -14,6 +14,10 @@
 
 #include <beerocks/tlvf/beerocks_message.h>
 
+#include <tlvf/ieee_1905_1/tlvMacAddress.h>
+#include <tlvf/wfa_map/tlvApMetric.h>
```
arnout: Hum, apparently the yaml file is not entirely accurate: it is called "AP Metrics TLV" so this should be `tlvApMetrics`.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -14,6 +14,10 @@
...
+#include <tlvf/wfa_map/tlvApMetricQuery.h>
```
arnout, line 17: Note that in the query it *is* singular. Confusing spec...

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -666,6 +670,87 @@ bool monitor_thread::update_ap_stats()
...
+    }
+    for (size_t bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
+        auto bssid = ap_metrics_query_tlv->bssid_list(bssid_idx);
+        if (!std::get<0>(bssid)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            return false;
+        }
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << std::get<1>(bssid);
+        bssid_response = std::get<1>(bssid);
```
arnout: Here we should also check that the bssid is indeed one of our vaps. And all the rest below should be done inside the loop.

agent/src/beerocks/monitor/monitor_thread.cpp
arnout: Since this will be used in multiple places, it would be better to create a new function for adding the tlvApMetrics.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -666,6 +670,87 @@ bool monitor_thread::update_ap_stats()
...
+    ap_metrics_response_tlv->bssid()                               = bssid_response;
+    ap_metrics_response_tlv->channel_utilization()                 = 10;
+    ap_metrics_response_tlv->number_of_stas_currently_associated() = 2;
+    ap_metrics_response_tlv->estimated_service_parameters()        = estimated_service_parameters;
+
+    if (!ap_metrics_response_tlv->alloc_estimated_service_info_field(sizeof(buf)))
+        LOG(ERROR)
+            << "Couldn't allocate ap_metrics_response_tlv->alloc_estimated_service_info_field";
+    auto tmp = ap_metrics_response_tlv->estimated_service_info_field();
+    std::copy_n(buf, sizeof(buf), tmp);
```
arnout: Rather than copying from an uninitialized buffer, it would be better to use `fill_n` to set it to 0.

agent/src/beerocks/monitor/monitor_thread.cpp
arnout: That's different: that one may actually be called with a nullptr.

arnout (changes requested):
Lots of comments from various reviewers but as promised I'll take over this PR.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: m_ap_metric_query should be cleared before entering the loop, just in case there is data from a previous query message that has not been completely processed yet

arnout:
First update done, but not all comments have been handled yet by far.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
VladyslavTupikin: In the backhaul manager, we are searching socket for specific BSSID and we should send a query with concrete BSSID instead of forwarding all payload because in the monitor_thread we should fill up BSSID in the response message and we can't find the correct BSSID from the full payload as we did in the backhaul manager.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -466,6 +467,30 @@ void monitor_stats::process()
     }
 }
 
+bool monitor_stats::addApMetrics(ieee1905_1::CmduMessageTx &cmdu_tx, const sMacAddr& bssid)
+{
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetrics>();
+
+    if (!ap_metrics_response_tlv) {
+        LOG(ERROR) << "Couldn't addClass tlvApMetrics";
```
MuteSpirit, line 475: Please add 
```
return false;
```
after this line

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2431,164 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    }
+    for (size_t bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
+        auto bssid_tupl = ap_metrics_query_tlv->bssid_list(bssid_idx);
+        if (!std::get<0>(bssid_tupl)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            return ret;
+        }
+        bssid.push_back(std::get<1>(bssid_tupl));
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << std::get<1>(bssid_tupl);
```
MuteSpirit: (nitpicks) Double call of `std::get<1>(bssid_tupl)`
```suggestion
        const auto &bssid = std::get<1>(bssid_tuple);
        bssid.push_back(bssid);
        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
                   << "  bssid " << bssid;
```

MuteSpirit (dismissed):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: They (should) have all the same MID. But in fact, this should be checked.

arnout:
Most comments are taken into account. Remaining will be done in separate PRs:
- test_flows -> #1288
- collecting responses (refactoring + make more robust): TODO

I think this one can and should be merged as is now.

arnout (approved):

MuteSpirit (approved):

framework/tlvf/src/src/CmduMessageRx.cpp
mariomaz: (nitpick) There is a "typo" in the log message: should be "Unknown" instead of "tlvUnknown".
Better off, "Unknown TLV type"

controller/src/beerocks/master/db/node.cpp
```diff
@@ -343,7 +343,7 @@ bool node::link_metrics_data::add_receiver_link_metric(
     return true;
 }
 
-bool node::ap_metrics_data::add_ap_metric_data(std::shared_ptr<wfa_map::tlvApMetric> ApMetricData)
+bool node::ap_metrics_data::add_ap_metric_data(std::shared_ptr<wfa_map::tlvApMetrics> ApMetricData)
```
mariomaz, line 346: (nitpick)
Maybe the name of the parameter should be fixed as well

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1403,6 +1407,48 @@ bool monitor_thread::handle_multi_ap_policy_config_request(Socket &sd,
     return true;
 }
 
+bool monitor_thread::handle_ap_metrics_query(Socket &sd, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    const auto mid            = cmdu_rx.getMessageId();
+    auto ap_metrics_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
```
mariomaz: Should be named ap_metric_query_tlv

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1403,6 +1407,48 @@ bool monitor_thread::handle_multi_ap_policy_config_request(Socket &sd,
...
+    if (!ap_metrics_query_tlv) {
+        LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
+        return false;
+    }
+
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_RESPONSE_MESSAGE";
```
mariomaz: (nitpick) Message is not IEEE1905

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1403,6 +1407,48 @@ bool monitor_thread::handle_multi_ap_policy_config_request(Socket &sd,
...
+            return false;
+        }
+        const auto &bssid = std::get<1>(bssid_tuple);
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << bssid;
+        if (!mon_stats.add_ap_metrics(cmdu_tx, bssid)) {
+            return false;
+        }
+
+        //TODO: find the way how to check policy config was enabled or not
```
mariomaz, line 1438: Policy config is stored in monitor_db, inside a member variable of type sApMetricsReportingInfo

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -466,6 +467,31 @@ void monitor_stats::process()
     }
 }
 
+bool monitor_stats::add_ap_metrics(ieee1905_1::CmduMessageTx &cmdu_tx, const sMacAddr &bssid)
+{
+    auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetrics>();
```
mariomaz, line 472: Name should be ap_metrics_tlv

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1403,6 +1407,48 @@ bool monitor_thread::handle_multi_ap_policy_config_request(Socket &sd,
...
+    for (size_t bssid_idx = 0; bssid_idx < ap_metrics_query_tlv->bssid_list_length(); bssid_idx++) {
+        auto bssid_tuple = ap_metrics_query_tlv->bssid_list(bssid_idx);
+        if (!std::get<0>(bssid_tuple)) {
+            LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
+            return false;
+        }
+        const auto &bssid = std::get<1>(bssid_tuple);
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << bssid;
+        if (!mon_stats.add_ap_metrics(cmdu_tx, bssid)) {
```
mariomaz, line 1434: Maybe it should check if BSSID is valid before adding TLV

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2434,159 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
     return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                               const std::string &src_mac)
+{
+    std::vector<sMacAddr> bssid;
+    const auto mid            = cmdu_rx.getMessageId();
+    auto ap_metrics_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
```
mariomaz: Name should be ap_metric_query_tlv

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2434,159 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+            return false;
+        }
+        bssid.push_back(std::get<1>(bssid_tuple));
+        LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
+                   << "  bssid " << std::get<1>(bssid_tuple);
+    }
+
+    auto ret = false;
+    for (auto socket : slaves_sockets) {
+        for (auto mac : bssid) {
```
mariomaz: const auto &

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2434,159 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+                    LOG(ERROR) << "Failed to create AP_METRICS_QUERY_MESSAGE";
+                }
+
+                auto query = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
+                if (!query) {
+                    LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetricQuery>";
+                }
+
+                if (!query->alloc_bssid_list(1)) {
+                    LOG(ERROR) << "Failed allocate memory for bssid_list";
```
mariomaz, line 2480: return false if error

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2434,159 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+
+                auto forward =
+                    cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
+                if (!forward) {
+                    LOG(ERROR) << "Failed to create AP_METRICS_QUERY_MESSAGE";
+                }
+
+                auto query = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
+                if (!query) {
+                    LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetricQuery>";
```
mariomaz, line 2475: return false if error

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2434,159 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+        for (auto mac : bssid) {
+            int i = 0;
+            if (mac == socket->vaps_list.vaps[i].mac) {
+                LOG(DEBUG) << "Forwarding AP_METRICS_QUERY_MESSAGE message to son_slave, bssid: "
+                           << std::hex << network_utils::mac_to_string(mac);
+
+                auto forward =
+                    cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
+                if (!forward) {
+                    LOG(ERROR) << "Failed to create AP_METRICS_QUERY_MESSAGE";
```
mariomaz, line 2469: return false if error

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2434,159 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    return ret;
+}
+
+bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
+
+    auto ap_metric_tlv = cmdu_rx.getClass<wfa_map::tlvApMetrics>();
```
mariomaz: Name should be ap_metrics_tlv

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2412,6 +2434,159 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
...
+    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    // Prepare tlvApMetrics for each processed query
+    for (const auto &response : m_ap_metric_response) {
+        auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetrics>();
```
mariomaz, line 2565: Name should be ap_metrics_tlv

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: > When we receive a response that is not expected, it is forwarded as-is to the controller.

I'd ignore this one

mariomaz: We also have to consider the case of periodically reporting AP metrics (without a query, no mid)

mariomaz (approved):
A few nitpick comments and some others that I don't know if you want to address in this PR or in a coming PR, so I've removed the "ready for merge" label and approve to let you decide.

controller/src/beerocks/master/db/node.cpp
arnout: That would be a separate commit. Since this PR has already taken too long, I'm going to leave it for the database refactoring.

agent/src/beerocks/monitor/monitor_thread.cpp
arnout: This is implemented in #1285; changing the TODO will just create a merge conflict, so I'm keeping it as is.

agent/src/beerocks/monitor/monitor_thread.cpp
arnout: Yes, that's part of the remaining refactoring.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: I disagree, I think it's good to distinguish from the ap_metrics_tlv variable that is used above.

PR #1232, opened 2020-05-04
Hotfix/remove annoying prints

morantr:
This PR fixes some print issues discovered on RDKB where using UDS_BUS, and print issue which appears also on Linux.

### DIFF
--- a/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
@@ -42,6 +42,7 @@ protected:
     {
         return false;
     }
+    virtual bool skip_filtered_message_type(ieee1905_1::eMessageType msg_type) { return false; }
     virtual std::string print_cmdu_types(const message::sUdsHeader *cmdu_header)
     {
         return std::string();
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -184,6 +184,11 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
         return false;
     }
 
+    // On UDS bus, unsubscribed messages needs to be filtered manually.
+    if (skip_filtered_message_type(cmdu_rx.getMessageType())) {
+        return true;
+    }
+
     if (!handle_cmdu(sd, cmdu_rx)) {
         return false;
     }
--- a/common/beerocks/btl/CMakeLists.txt
+++ b/common/beerocks/btl/CMakeLists.txt
@@ -21,6 +21,10 @@ set_target_properties(${PROJECT_NAME} PROPERTIES VERSION ${prplmesh_VERSION} SOV
 target_link_libraries(${PROJECT_NAME} PUBLIC bcl tlvf elpp btlvf common)
 if (NOT MSGLIB STREQUAL "None")
      target_link_libraries(${PROJECT_NAME} PRIVATE ieee1905_transport_messages)
+else()
+    # Export the UDS_BUS definition publicly to any target that
+    # may include the <btl.h> header
+    target_compile_definitions(${PROJECT_NAME} PUBLIC UDS_BUS)
 endif()
 
 # Include paths
--- a/common/beerocks/btl/btl_uds.cpp
+++ b/common/beerocks/btl/btl_uds.cpp
@@ -55,6 +55,7 @@ bool transport_socket_thread::read_ready(Socket *s) { return socket_thread::read
 
 bool transport_socket_thread::bus_subscribe(const std::vector<ieee1905_1::eMessageType> &msg_types)
 {
+    m_subscribed_messages.insert(msg_types.begin(), msg_types.end());
     return true;
 }
 
@@ -123,6 +124,14 @@ bool transport_socket_thread::bus_send(ieee1905_1::CmduMessage &cmdu, const std:
 }
 
 bool transport_socket_thread::from_bus(Socket *sd) { return sd == bus; }
+
+bool transport_socket_thread::skip_filtered_message_type(ieee1905_1::eMessageType msg_type)
+{
+    if (m_subscribed_messages.find(msg_type) == m_subscribed_messages.end()) {
+        return true;
+    }
+    return false;
+}
 
 bool transport_socket_thread::work()
 {
--- a/common/beerocks/btl/include/btl/btl.h
+++ b/common/beerocks/btl/include/btl/btl.h
@@ -15,6 +15,8 @@
 #include <beerocks/tlvf/beerocks_message.h>
 #include <tlvf/ieee_1905_1/eMessageType.h>
 
+#include <unordered_set>
+
 // Forward Declarations
 namespace mapf {
 class Poller;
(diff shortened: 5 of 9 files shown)


vitalybu:
Do not merge this PR.
I want to examine the issue further.

controller/src/beerocks/master/tasks/task.cpp
```diff
@@ -46,9 +46,7 @@ void task::response_received(std::string mac,
 void task::event_received(int event_type, void *obj)
 {
     auto it = pending_events.find(event_type);
-    if (it == pending_events.end()) {
-        TASK_LOG(DEBUG) << "received non-pending event " << event_type;
-    } else {
+    if (it != pending_events.end()) {
```
tomereli, line 49: I don't get this. Each task is registering for events, right? And in the controller we pass events only to tasks which should receive these events. If this is not the case, it is a bug, and this commit is just hiding it instead of fixing it.
Please correct me if i'm wrong..

common/beerocks/btl/include/btl/btl.h
```diff
@@ -60,12 +62,19 @@ class transport_socket_thread : public socket_thread {
...
     std::unique_ptr<SocketServer> bus_server_socket;
 #else
     std::shared_ptr<mapf::LocalBusInterface> bus = nullptr;
     std::shared_ptr<mapf::Poller> poller         = nullptr;
 #endif
+
+    // This member supposes to be defined inside "#ifdef UDS_BUS" block, but for some reason, when
+    // defined there, there is a crash when trying to insert more than one element to the container.
+    // For now, define it here where it doesn't cause a crash.
+    std::unordered_set<ieee1905_1::eMessageType> m_subscribed_messages;
```
tomereli: Not acceptable. If you have a bug you solve it, not work around it to get the change merged.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -184,6 +184,11 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
         return false;
     }
 
+    // On UDS bus, unsubscribed messages needs to be filtered manually.
```
tomereli, line 187: This completely breaks the abstraction! Why didn't you just add this to `btl_uds.cpp` before calling `handle_cmdu`? Why add it to the socket thread class? This whole commit is wrong.

common/beerocks/btl/btl_uds.cpp
```diff
@@ -124,6 +125,14 @@ bool transport_socket_thread::bus_send(ieee1905_1::CmduMessage &cmdu, const std:
 
 bool transport_socket_thread::from_bus(Socket *sd) { return sd == bus; }
 
+bool transport_socket_thread::skip_filtered_message_type(ieee1905_1::eMessageType msg_type)
```
tomereli, line 128: This function should be removed completely and replaced with a static function in the btl_uds.cpp implementation only.

common/beerocks/btl/CMakeLists.txt
```diff
@@ -21,6 +21,10 @@ set_target_properties(${PROJECT_NAME} PROPERTIES VERSION ${prplmesh_VERSION} SOV
 target_link_libraries(${PROJECT_NAME} PUBLIC bcl tlvf elpp btlvf common)
 if (NOT MSGLIB STREQUAL "None")
```
tomereli, line 22: Good - so this should be a fixup commit.

tomereli (changes requested):

controller/src/beerocks/master/tasks/task.cpp
morantr: > Each task is registering for events, right?

No.
A task doesn't register for anything. A task can decide that it is pending for an event to block its work cycle until the event is being received.
But, it is normal behavior that events would get to a task even if it is not pending for it.

common/beerocks/btl/include/btl/btl.h
morantr: Vitaly fixed it. I had difficulties to find out the root cause of the problem so I committed it so the change will be committed, knowing it will be fixed later.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
morantr: Why do you say it is breaking the abstraction? This function is implemented on socket_thread, and can be overridden if needed.
The socket thread is base class for btl classes, and this function is needed for abstraction.
Anyway, it cannot be added to btl classes since I had to call it from a function implemented only on the socket thread, which I didn't want to override.

common/beerocks/btl/btl_uds.cpp
morantr: That is not possible since it is being called on socket_thread.

common/beerocks/btl/CMakeLists.txt
morantr: Sure, I will squash it before I'll mark it as ready for merge.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
```diff
@@ -42,6 +42,7 @@ class socket_thread : public thread_base {
     {
         return false;
     }
+    virtual bool skip_filtered_message_type(ieee1905_1::eMessageType msg_type) { return false; }
```
vitalybu, line 45: I don't like this name... Maybe change it to `should_handle_ieee1905_msg()`?

tomereli: OK I understand why it is needed, but please change the name to @vitalybu 's suggestion before merge

common/beerocks/btl/include/btl/btl.h
tomereli: You marked the PR "ready for merge". If you push a commit like that and plan to fix it, it is perfectly fine, but you need to mention that in the commit message that this commit should be dropped before merge, like @rmelotte does often.

controller/src/beerocks/master/tasks/task.cpp
tomereli: Well I think its not an ideal design choice, why is it implemented like that? Why can't a task "subscribe" to events? we can have the base class drop events for which the task didn't subscribe to. @vitalybu ?

common/beerocks/bcl/source/beerocks_socket_thread.cpp
tomereli: I got it all wrong, forget about it. Just change the name and don't mention UDS - perhaps on other bus types the filtering will also be implemented manually so just in the local bus case we skip this function.

tomereli (approved):

controller/src/beerocks/master/tasks/task.cpp
morantr: Why do you think that a task subscribing to an event is an ideal design? IMO it is not.
What is the benefit of such registration? Nothing.
It is not like when you push event, its reach to all the tasks and they have to filter it if they are subscribing to it or not. When we send an event we send it to a specific task ID, so only one instance of a specific task class would get it.
In the current design, each task has it is own event types, and on subscribing event design it would have to be global shared events for all the tasks, like TLVF, but that is just not necessary.

itayx (approved):

PR #1236, opened 2020-05-05
agent: monitor: handle both vendor-specific and ieee1905 messages

mariomaz:
Modify handle_cmdu() method to dispatch received CMDU messages to
either handle_cmdu_vs_message() or handle_cmdu_ieee1905_1_message(),
depending on if message is a vendor-specific message or not
respectively.


morantr (approved):

tomereli (approved):

RanRegev (approved):
I wish we could dispatch messages in a more type safe way.
With much less "query what we hold" along the way.
But with current code structure it is fine and clean.
It actually says: if it is not a vendor specific message, go ahead and try to extract 1905 out of it. if that fails, then we don't know what is it.

PR #1237, opened 2020-05-05
Support multi-user environment in VSCode tasks

vitalybu:


morantr (approved):

.vscode/launch.json
```diff
@@ -64,7 +64,7 @@
                 "pipeArgs": [
                     "exec",
                     "-i",
-                    "gateway-root",
+                    "gateway-${env:USER}",
```
tomereli, line 67: Note that this will not work if we run vscode as sudo, and then it needs to be `"gateway-${env:SUDO_USER}"`. But I don't know how to specify this in the launch.json file, so OK.

.vscode/tasks.json
```diff
@@ -32,7 +32,7 @@
         "command": "rm",
         "args": [
           "-rf",
-          "build"
+          "build/*"
```
tomereli, line 35: Why not delete the full build folder than? Should be a separate commit (nitpick)

tomereli (approved):

.vscode/tasks.json
vitalybu: In case you want to link the `build` directory to a different location (e.g. ramdisk to speed up the build).

.vscode/launch.json
vitalybu: Why would you run vscode as sudo?

PR #1244, opened 2020-05-06
hotfix: tlvf: don't use swapped length field in class_swap

arnout:
The class_swap() function for variable length fields generated something like this:

```
       tlvf_swap(16, reinterpret_cast<uint8_t*>(m_foo_length));
      for (size_t i = 0; i < *(m_foo_length); i++){
          std::get<1>(foo(i)).class_swap();
      }
```

In other words, for variable length fields where the length is 16-bit or 
longer, the swap function would iterate over the swapped length. For small
values of length this has no effect except for logging errors like:

```
      --> TLVF: Requested index is greater than the number of available entries
```

However, for larger lengths, it would have the effect of only swapping part
of the list.

Fix this by using the index instead of the length field directly. Since 
variable length fields are only allowed in classes, there is always an 
index field, so we can just as well use it.

This didn't use to be possible because originally the swapping was done 
before parsing. Now, however, swapping is done afterwards and swapping of
the length field is done separately. So also for variable length lists, we
can rely on an up-to-date index field when class_swap is called.


tomereli (approved):

arnout:
Depends on #1249 to resolve CI issues.

rmelotte (approved):

PR #1246, opened 2020-05-07
Bugfix/various cppcheck improvements

rmelotte:
A couple of improvements to the cppcheck-diff job that I worked on in the background.

- output issues with relative paths, this makes it possible to generate and use the list of issues locally. 
- add a shell wrapper for cppcheck-new-issues.sh to run the cppcheck-diff job locally.
- report the solved issues, and update the list of issues automatically. This way, people just have to run the script and solved issues are removed from the list automatically (or it can also be copied from gitlab).
- fix some issues in tlvf.py
- ignore the tlvf folder, since we can't fix them all at the moment. This avoids the cppcheck-diff job from bugging people with issues in tlvf.py we can't solve easily.


framework/tlvf/AutoGenerated/src/tlvf/test/tlvVarList.cpp
```diff
@@ -564,7 +564,7 @@ bool tlvTestVarList::init()
         // swap back since complex_list will be swapped as part of the whole class swap
         complex_list->class_swap();
     }
-    m_var1 = (cInner*)m_buff_ptr__;
+    m_var1 = reinterpret_cast<cInner*>(m_buff_ptr__);
```
arnout, line 567: This ended up in the wrong commit.

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -25,8 +25,19 @@ do
...
+        "$i"
     fi
 done
 
+# Generate a new list of existing issues by removing the ones that are now solved
+comm -12 <(sort "$rootdir"/cppcheck_results.txt) <(sort "$rootdir"/ci/cppcheck/cppcheck_existing_issues.txt)  > "$rootdir"/cppcheck_existing_issues.txt
+
+mv "$rootdir"/cppcheck_existing_issues.txt "$rootdir"/ci/cppcheck/cppcheck_existing_issues.txt
+
+if ! git diff -U0 --exit-code -- "$rootdir"/ci/cppcheck/cppcheck_existing_issues.txt ; then
```
arnout, line 37: I think this should only be done if status was still 0. Otherwise, you'll get both the report of the new issue above, and again in the diff here.

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -25,8 +25,19 @@ do
     if ! grep -q -F "$i" "$rootdir"/ci/cppcheck/cppcheck_existing_issues.txt ; then
         status=1
         err "New issue:"
-        grep -F "$i" "$rootdir/cppcheck_results.txt"
+        "$i"
     fi
 done
 
+# Generate a new list of existing issues by removing the ones that are now solved
+comm -12 <(sort "$rootdir"/cppcheck_results.txt) <(sort "$rootdir"/ci/cppcheck/cppcheck_existing_issues.txt)  > "$rootdir"/cppcheck_existing_issues.txt
```
arnout, line 33: This relies very heavily on the order of the cppcheck_results.txt file to be stable. We run cppcheck with just the rootdir argument. I don't know how it iterates, but I assume it uses directory order without sorting. In that case, the order is not stable, so we may get different results on different runs.

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -25,8 +25,19 @@ do
...
+
+mv "$rootdir"/cppcheck_existing_issues.txt "$rootdir"/ci/cppcheck/cppcheck_existing_issues.txt
```
arnout, line 35: Why move it instead of directly generating it in the right location?

Actually, why move it on top of the existing one and use git diff? You could just as well leave it in the root directory and use normal diff. And if you want to avoid needing to install normal diff, you can use git diff with two files as arguments even if one of them is not tracked.

ci/cppcheck/cppcheck_existing_issues.txt
```diff
@@ -1,399 +1,762 @@
...
-'/builds/prpl-foundation/prplMesh/framework/tlvf/AutoGenerated/src/tlvf/wfa_map/tlvMetricReportingPolicy.cpp: style: The if condition is the same as the previous if condition [duplicateCondition]    if (m_parse__) {
-'/builds/prpl-foundation/prplMesh/framework/tlvf/AutoGenerated/src/tlvf/wfa_map/tlvMetricReportingPolicy.cpp: style: C-style pointer casting [cstyleCast]    m_metrics_reporting_conf_list = (sMetricsReportingConf*)m_buff_ptr__;
-'/builds/prpl-foundation/prplMesh/framework/tlvf/AutoGenerated/src/tlvf/wfa_map/tlvSteeringPolicy.cpp: style: The if condition is the same as the previous if condition [duplicateCondition]    if (m_parse__) {
-'/builds/prpl-foundation/prplMesh/framework/tlvf/AutoGenerated/src/tlvf/wfa_map/tlvSteeringPolicy.cpp: style: C-style pointer casting [cstyleCast]    m_radio_ap_control_policy_list = (sRadioApControlPolicy*)m_buff_ptr__;
+'common/beerocks/bcl/include/bcl/network/socket.h: style: Statements following return, break, continue, goto or throw will never be executed. [unreachableCode]        m_error.clear();
+        ^'
```
arnout: There seems to be some commit reordering going on here, because the carets are re-introduced here...

tools/docker/cppcheck-diff.sh
```diff
@@ -0,0 +1,12 @@
...
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2019-2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+scriptdir="$(cd "${0%/*}" && pwd)"
+rootdir="$(realpath "$scriptdir/../..")"
+
+docker run --rm -it -w "$rootdir" -v "$rootdir:$rootdir" --user="${SUDO_UID:-$UID}" prplfoundationinc/prplmesh-builder:alpine-3.11.3 "$rootdir"/ci/cppcheck/cppcheck-new-issues.sh .
```
arnout: (nitpick) I'd prefer this as an option to cppcheck.sh instead, to avoid a proliferation of small wrapper scripts...

tools/docker/cppcheck-diff.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2019-2020 the prplMesh contributors (see AUTHORS.md)
```
arnout: This file 2020 only (but becomes irrelevant with my comment below)

framework/tlvf/tlvf.py
```diff
@@ -779,7 +779,7 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
                     lines_h.append( "}" )
                 else:
                     # add default value to init func
-                    lines_cpp.append("m_%s = (%s*)m_%s__;" % ( param_name, param_type, self.MEMBER_BUFF_PTR))
+                    lines_cpp.append("m_%s = reinterpret_cast<%s*>(m_%s__);" % ( param_name, param_type, self.MEMBER_BUFF_PTR))
```
arnout, line 782: Please squash this in the other commit doing the same thing somewhere else.

framework/tlvf/AutoGenerated/include/tlvf/test/tlvVarList.h
```diff
@@ -48,7 +48,7 @@ class tlvTestVarList : public BaseClass
         std::tuple<bool, cInner&> complex_list(size_t idx);
         std::shared_ptr<cInner> create_complex_list();
         bool add_complex_list(std::shared_ptr<cInner> ptr);
-        bool isPostInitSucceeded();
+        bool isPostInitSucceeded() override;
```
arnout, line 51: Again, this ended up in the wrong commit.

ci/cppcheck/cppcheck-new-issues.sh
```diff
@@ -13,6 +13,10 @@ rootdir="${scriptdir%/*/*}"
 # Remove the caret lines, we don't need them:
 sed -i '/^\s*\^\s*$/d' "$rootdir/cppcheck_results.txt"
 
+# Sort the list of issues to have a stable list, easier to compare:
+sort "$rootdir/cppcheck_results.txt" > "$rootdir/cppcheck_results_sorted.txt"
+mv "$rootdir/cppcheck_results_sorted.txt" "$rootdir/cppcheck_results.txt"
+
```
arnout: It would actually make more sense if cppcheck.sh would not create any file but simply output to stdout, and that it was the CI job that creates the output file. I find it pretty annoying that running a tool like that creates a file (which BTW isn't even in gitignore).

With that, you could use a nice pipeline here: `cppcheck.sh | sed | sort > results`.

.gitlab-ci.yml
```diff
@@ -66,7 +66,7 @@ flake8-diff:
   image: pipelinecomponents/flake8:f087c4c
   script:
     - apk add --update --no-cache git
-    - git diff -U0 origin/master | flake8 --verbose --diff
+    -  git diff -U0 origin/master -- . ':(exclude)framework/tlvf/tlvf.py' | flake8 --verbose --diff
```
arnout: #1249 removes flake8-diff entirely instead, which IMO is better :-)

arnout (approved):
Lots of small comments, as usual. Nothing really stands out, so I'm fine with merging as is (except the last commit has to be dropped because it conflicts with #1249). We shouldn't spend too much time on this stuff.

MuteSpirit:
Looks like there are similar issues with C-style type casting in next lines of `framework/tlvf/tlvf.py`
* line 778
```
lines_h.append( "%sreturn (%s)(%s);" % (self.getIndentation(1), param_type, ret_const_val) )
```
* line 834
```
                    lines_cpp.append( "%sreturn (%s%s&)(*m_%s);" % (self.getIndentation(1), const, param_type, param_name) )  
```
* line 864
```
            lines_cpp.append("m_%s = (%s*)m_%s__;" % (param_name, param_type, self.MEMBER_BUFF_PTR) )
```
* line 1025
```
                lines_cpp.append( "%sreturn ((%s*)m_%s);" % (self.getIndentation(1), param_type, param_name) )
```

Were they missed or will be fixed another time?

framework/tlvf/tlvf.py
```diff
@@ -745,7 +745,7 @@ def addClassParam(self, obj_meta, param_name, param_type, param_type_info, param
                 lines_h.append("std::shared_ptr<%s> %s() { return m_%s_ptr; }" % (param_type, param_name, param_name))
 
                 # Add param handling to init function
-                lines_cpp.append("m_%s = (%s*)m_%s__;" % ( param_name, param_type, self.MEMBER_BUFF_PTR))
+                lines_cpp.append("m_%s = reinterpret_cast<%s*>(m_%s__);" % ( param_name, param_type, self.MEMBER_BUFF_PTR))
```
RanRegev, line 748: C style cast and reinterpret cast when dealing with object pointers have no much different in the "high level" point of view: effectively you _guess_ what the bytes you point to means, and you tell the compiler to think the same as you.
I didn't follow the code but how can we be sure that `self.MEMBER_BUFF_PTR` is really of `param_type`. And if we are sure it is, why didn't we declare it to be such from the first place?

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -21,10 +21,11 @@ run_cppcheck() {
              --force \
              --error-exitcode=1 \
              --enable=warning,style,information \
+             -rp="$rootdir" \
              -j"$(nproc)" \
              -q \
              "${INCLUDES[@]}" \
-             "$@" 2>&1 | tee "$OUTPUT_FILE"
+             "$@" 2>&1 | sed "s|$rootdir/||g" | tee "$OUTPUT_FILE"
```
RanRegev, line 28: +1 for replacing the `/` command with another one because of having forward slash in the search.
I would use `@` but `|` is good too.

tools/docker/cppcheck.sh
```diff
@@ -10,4 +10,4 @@
 scriptdir="$(cd "${0%/*}" && pwd)"
 rootdir="$(realpath "$scriptdir/../..")"
 
-docker run --rm -it -w "$rootdir" -v "$rootdir:$rootdir"  prplfoundationinc/prplmesh-builder:alpine-3.11.3 "$rootdir"/tools/docker/static-analysis/cppcheck.sh -j"$(nproc)" .
+docker run --rm -it -w "$rootdir" -v "$rootdir:$rootdir" --user="${SUDO_UID:-$UID}"  prplfoundationinc/prplmesh-builder:alpine-3.11.3 "$rootdir"/tools/docker/static-analysis/cppcheck.sh -j"$(nproc)" .
```
RanRegev, line 13: I feel uncomfortable with this change.
I didn't dive into the details, so it might be that everything is OK. However I would like the author (@rmelotte) to reference in the final decision the following:
* https://echorand.me/posts/docker-user-namespacing-remap-system-user/
* https://github.com/docker/docker-ce/blob/531930f3294c31db414f17f80fa8650d4ae66644/components/engine/api/types/container/host_config.go#L169
* and also other work that might have been done in this area by others in prplMesh

framework/tlvf/AutoGenerated/include/tlvf/test/tlvVarList.h
RanRegev: also autogenerated tlvf are now ignored.

ci/cppcheck/cppcheck-new-issues.sh
RanRegev: I agree with @arnout - scripts that generate files are less compose-able and are therefore harder to use in other contents. A mid solution is to add an _option_ to output to a file and use stdout as the default for this option.

RanRegev (approved):
A. I miss documentation regarding the way we use cppcheck: CI, local, results, sort, diff, etc. I found [only few lines](https://github.com/prplfoundation/prplMesh/wiki/Testing-in-CI#cpp-check) in the wiki that explains how to run cppcheck locally. Maybe there is more documentation somewhere?
B. There are few commits in this PR that change things in tlvf.py that in turn change the AutoGenerated result. However, the entire AutoGenerated directory is ignored by another commit in this PR. So from the cppcheck point of view the changes are not checked, so maybe it is better to introduce these changes into another PR, unrelated to cppcheck improvements.

MuteSpirit (approved):

ci/cppcheck/cppcheck-new-issues.sh
rmelotte: The list of issues is always sorted first, so it should always produce the same result. Or am I misunderstanding something?

tools/docker/cppcheck-diff.sh
rmelotte: Done

framework/tlvf/tlvf.py
rmelotte: Done

framework/tlvf/AutoGenerated/include/tlvf/test/tlvVarList.h
rmelotte: Since the change to the autogenerated files was so small, I intentionally kept it in the same commit as the change to tlvf.py.

.gitlab-ci.yml
rmelotte: Done

rmelotte:
I started solving some of the issues in tlvf.py, but then there were too many to solve in one go.
The reason TLVs are later ignored is to save other developers the time that they would have to spend dealing with those every time they add/modify a TLV.

rmelotte:
I addressed some of the comments, there are still some others valid and left open.
I would like to still merge it as is because:
- working on them is not a priority at moment.
- this PR already allows running the cppcheck-diff job locally and also makes it easier to add new tlvs (since issues coming from tlvf.py are ignored)

PR #1249, opened 2020-05-07
Remove flake8-diff from CI

arnout:
flake8-diff is causing trouble when editing tlvf.py (cfr. https://gitlab.com/prpl-foundation/prplMesh/-/jobs/542918185).
Rather than trying to fix the issue, remove the job entirely.
Instead, run flake8 on everything except the two files that are not OK.


rmelotte (approved):
FYI I also already ignored tlvf.py in https://github.com/prplfoundation/prplMesh/pull/1246, so we'll have to rebase one of them manually

tomereli (approved):

abelog (approved):
LGFM

PR #1251, opened 2020-05-08
boardfarm: expose 8002 port in spawned dockerized devices

abelog:
### Description

During boardfarm tests, port 8002 is used to communicate with dockerized
devices. Currently, it's not exposed. So, test is failing as it's unable
to reach this port.

### Changes

Add `--expose 8002` to call of `run.sh` to allow boardfarm tests to reach this port.

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
@@ -43,7 +43,7 @@ class PrplMeshDocker(PrplMeshBase):
 
         docker_cmd = os.path.join(rootdir, "tools", "docker", "run.sh")
         docker_args = ["--verbose", "--detach", "--force", "--name", self.name,
-                       "--network", self.docker_network]
+                       "--network", self.docker_network, "--expose", "8002"]
 
         if self.role == "controller":
             # Spawn dockerized controller


tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, *args, **kwargs):
 
         docker_cmd = os.path.join(rootdir, "tools", "docker", "run.sh")
         docker_args = ["--verbose", "--detach", "--force", "--name", self.name,
-                       "--network", self.docker_network]
+                       "--network", self.docker_network, "--expose", "8002"]
```
arnout, line 46: We should use the same approach as in test_gw_receiver.sh:
```
    # Read GW & Repeater UCC ports
    local GW_UCC_PORT
    GW_UCC_PORT=$(grep ucc_listener_port \
    "${rootdir}"/build/install/config/beerocks_controller.conf | \
    awk -F'[= ]' '{ print $2 }')

    local RP_UCC_PORT
    RP_UCC_PORT=$(grep ucc_listener_port \
    "${rootdir}"/build/install/config/beerocks_agent.conf | \
    awk -F'[= ]' '{ print $2 }')
```
and expose the correct one depending on the role.

Also, to improve it even more, for the controller, if the agent and controller UCC port are configured differently, we should expose both of them. That way we could do full tests with only a single device.

arnout (approved):
Although my single comment is relevant, it doesn't necessarily have to be done in this PR, since having this flexibility is not necessary at the moment. Maybe adding a todo would be nice but even that is not really needed - we can do it whenever there's a real need for it.

VladyslavTupikin (approved):

PR #1252, opened 2020-05-08
bwl: add get_survey_info method

mariomaz:
Add a new method to nl80211_client to obtain survey information
(channel occupation and noise level).

This information will later be used to periodically compute channel
utilization and report AP metrics when the difference between computed
value and previous value crosses a configured threshold.

### DIFF
--- a/common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
+++ b/common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
@@ -110,6 +110,24 @@ bool nl80211_client_dummy::get_sta_info(const std::string &interface_name,
     return true;
 }
 
+bool nl80211_client_dummy::get_survey_info(const std::string &interface_name,
+                                           std::vector<sSurveyInfo> &survey_info_list)
+{
+    for (size_t i = 0; i < 8; i++) {
+        sSurveyInfo survey_info;
+
+        survey_info.in_use        = (0 == i);
+        survey_info.frequency_mhz = 2412 + i * 5;
+        survey_info.noise_dbm     = i;
+        survey_info.time_on_ms    = i * 2000;
+        survey_info.time_busy_ms  = i * 1000;
+
+        survey_info_list.push_back(survey_info);
+    }
+
+    return true;
+}
+
 bool nl80211_client_dummy::set_tx_power_limit(const std::string &interface_name, uint32_t limit)
 {
     return true;
--- a/common/beerocks/bwl/dummy/nl80211_client_dummy.h
+++ b/common/beerocks/bwl/dummy/nl80211_client_dummy.h
@@ -61,6 +61,16 @@ public:
                       sta_info &sta_info) override;
 
     /**
+     * @brief Gets dummy survey information.
+     *
+     * @see nl80211_client::get_survey_info
+     *
+     * This implementation returns fixed survey info for the first 8 2.4GHz channels.
+     */
+    bool get_survey_info(const std::string &interface_name,
+                         std::vector<sSurveyInfo> &survey_info_list) override;
+
+    /**
      * @brief Set the tx power limit
      *
      * Set tx power limit for a radio
--- a/common/beerocks/bwl/include/bwl/nl80211_client.h
+++ b/common/beerocks/bwl/include/bwl/nl80211_client.h
@@ -298,6 +298,60 @@ public:
     };
 
     /**
+     * @brief Survey information
+     *
+     * Information obtained with NL80211_CMD_GET_SURVEY command through a NL80211 socket.
+     * See NL80211_SURVEY_INFO* in <linux/nl80211.h> for a description of each field.
+     */
+    struct sSurveyInfo {
+        /**
+         * Center frequency of channel.
+         */
+        uint32_t frequency_mhz = 0;
+
+        /**
+         * Channel is currently being used.
+         */
+        bool in_use = false;
+
+        /**
+         * Noise level of channel (u8, dBm).
+         */
+        int8_t noise_dbm = 0;
+
+        /**
+         * Amount of time (in ms) that the radio was turned on (on channel or globally)
+         */
+        uint64_t time_on_ms = 0;
+
+        /**
+         * Amount of the time (in ms) the primary channel was sensed busy (either due to activity
+         * or energy detect).
+         */
+        uint64_t time_busy_ms = 0;
+
+        /**
+         * Amount of time (in ms) the extension channel was sensed busy.
+         */
+        uint64_t time_ext_busy_ms = 0;
+
+        /**
+         * Amount of time (in ms) the radio spent receiving data (on channel or globally)
+         */
+        uint64_t time_rx_ms = 0;
+
+        /**
+         * Amount of time (in ms) the radio spent transmitting data (on channel or globally).
+         */
+        uint64_t time_tx_ms = 0;
+
+        /**
+         * Time (in ms) the radio spent for scan (on this channel or globally).
+         */
+        uint64_t time_scan_ms = 0;
+    };
+
+    /**
      * @brief Class destructor.
      */
     virtual ~nl80211_client() = default;
@@ -342,6 +396,20 @@ public:
      */
     virtual bool get_sta_info(const std::string &interface_name, const sMacAddr &sta_mac_address,
                               sta_info &sta_info) = 0;
+
+    /**
+     * @brief Gets survey information.
+     *
+     * Survey information includes channel occupation and noise level.
+     *
+     * @param[in] interface_name Interface name, either radio or Virtual AP (VAP).
+     * @param[out] survey_info_list List of survey information structures, one for each channel,
+     * as returned by the NL80211_CMD_GET_SURVEY command.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_survey_info(const std::string &interface_name,
+                                 std::vector<sSurveyInfo> &survey_info_list) = 0;
 
     /**
      * @brief Set the tx power limit
--- a/common/beerocks/bwl/shared/nl80211_client_impl.cpp
+++ b/common/beerocks/bwl/shared/nl80211_client_impl.cpp
@@ -549,6 +549,107 @@ bool nl80211_client_impl::get_sta_info(const std::string &interface_name,
         });
 }
 
+bool nl80211_client_impl::get_survey_info(const std::string &interface_name,
+                                          std::vector<sSurveyInfo> &survey_info_list)
+{
+    if (!m_socket) {
+        LOG(ERROR) << "Socket is NULL!";
+        return false;
+    }
+
+    // Get the interface index for given interface name
+    int iface_index = if_nametoindex(interface_name.c_str());
+    if (0 == iface_index) {
+        LOG(ERROR) << "Failed to read the index of interface " << interface_name << ": "
+                   << strerror(errno);
+
+        return false;
+    }
+
+    return m_socket.get()->send_receive_msg(
+        NL80211_CMD_GET_SURVEY, NLM_F_DUMP,
+        [&](struct nl_msg *msg) -> bool {
+            nla_put_u32(msg, NL80211_ATTR_IFINDEX, iface_index);
+
+            return true;
+        },
+        [&](struct nl_msg *msg) {
+            struct nlattr *tb[NL80211_ATTR_MAX + 1];
+            struct genlmsghdr *gnlh = static_cast<genlmsghdr *>(nlmsg_data(nlmsg_hdr(msg)));
+            struct nlattr *sinfo[NL80211_SURVEY_INFO_MAX + 1];
+            sSurveyInfo survey_info;
+
+            // Parse the netlink message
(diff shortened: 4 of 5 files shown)


common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -297,6 +297,24 @@ class nl80211_client {
         uint16_t rx_bitrate_100kbps = 0;
     };
 
+    /**
+     * @brief Survey information
+     *
+     * Information obtained with NL80211_CMD_GET_SURVEY command through a NL80211 socket.
+     * See NL80211_SURVEY_INFO* in <linux/nl80211.h> for a description of each field.
```
arnout, line 304: Meh, can't you copy that here?

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -343,6 +361,22 @@ class nl80211_client {
     virtual bool get_sta_info(const std::string &interface_name, const sMacAddr &sta_mac_address,
                               sta_info &sta_info) = 0;
 
+    /**
+     * @brief Gets survey information.
+     *
+     * Survey information includes channel occupation and noise level.
+     * It is intended to periodically compute channel utilization and report AP metrics when the
+     * difference between computed value and previous value crosses a configured threshold.
```
arnout: That's not its only use, and I don't think this remarks belongs here anyway.

In fact, the main use of survey info is for ACS/DCS.

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -297,6 +297,24 @@ class nl80211_client {
         uint16_t rx_bitrate_100kbps = 0;
     };
 
+    /**
+     * @brief Survey information
+     *
+     * Information obtained with NL80211_CMD_GET_SURVEY command through a NL80211 socket.
+     * See NL80211_SURVEY_INFO* in <linux/nl80211.h> for a description of each field.
+     */
+    struct survey_info {
```
arnout: Coding style says `sSurveyInfo`

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -297,6 +297,24 @@ class nl80211_client {
...
+        bool in_use               = false;
```
arnout: (nitpick) putting the in_use after the frequency reduces the size by 4 bytes, and it's more logical anyway.

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -343,6 +361,22 @@ class nl80211_client {
...
+     * difference between computed value and previous value crosses a configured threshold.
+     *
+     * @param[in] interface_name Interface name, either radio or Virtual AP (VAP).
+     * @param[out] survey_info_list List of survey information structures, one for each channel,
+     * as returned by the NL80211_CMD_GET_SURVEY command.
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool get_survey_info(const std::string &interface_name,
+                                 std::vector<survey_info> &survey_info_list) = 0;
```
arnout: Is there any particular reason why it's better to use an output parameter here rather than a return value (taking into account that std::vector has a move constructor)?

(Just to be clear: this is not a review comment, but a genuine question.)

Maybe it's to be able to signal errors. But to that I say:
- exceptions :-) ;
- what is the practical difference between failure and an empty survey result? i can't imagine any use case where you'd want to treat them differently.

common/beerocks/bwl/dummy/nl80211_client_dummy.h
```diff
@@ -60,6 +60,22 @@ class nl80211_client_dummy : public nl80211_client {
...
+     * Survey information includes channel occupation and noise level.
+     * It is intended to periodically compute channel utilization and report AP metrics when the
+     * difference between computed value and previous value crosses a configured threshold.
+     *
+     * @param[in] interface_name Interface name, either radio or Virtual AP (VAP).
+     * @param[out] survey_info_list List of survey information structures, one for each channel,
+     * as returned by the NL80211_CMD_GET_SURVEY command.
+     *
+     * @return True on success and false otherwise.
+     */
```
arnout: I wonder if it wouldn't be better to remove the doxygen comments from overridden virtual functions. The comment should really be in the base class. It would still be useful to add something, if necessary, like:
```
/** @brief Gets dummy survey information
 *
 * @see nl80211_client::get_survey_info
 *
 * This implementation returns fixed survey info for the first 8 2.4GHz channels.
 */
```

(this remarks has a much bigger scope than this PR, obviously.

common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
```diff
@@ -110,6 +110,24 @@ bool nl80211_client_dummy::get_sta_info(const std::string &interface_name,
...
 }
 
+bool nl80211_client_dummy::get_survey_info(const std::string &interface_name,
+                                           std::vector<survey_info> &survey_info_list)
+{
+    for (size_t i = 0; i < 8; i++) {
+        survey_info survey_info;
+
+        survey_info.in_use        = (0 == i);
+        survey_info.frequency_mhz = 2400 + i * 10;
```
arnout: I think we should at least use valid frequencies...
```suggestion
        survey_info.frequency_mhz = 2412 + i * 5;
```

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -549,6 +549,103 @@ bool nl80211_client_impl::get_sta_info(const std::string &interface_name,
         });
 }
 
+bool nl80211_client_impl::get_survey_info(const std::string &interface_name,
+                                          std::vector<survey_info> &survey_info_list)
+{
+    static struct nla_policy survey_policy[NL80211_SURVEY_INFO_MAX + 1];
+    survey_policy[NL80211_SURVEY_INFO_FREQUENCY] = {NLA_U32, 0, 0};
+    survey_policy[NL80211_SURVEY_INFO_NOISE]     = {NLA_U8, 0, 0};
```
arnout: (nitpick) IMO it doesn't make sense to declare a static member which you then write every time. Unfortunately, C++ doesn't have designated initialisers so there's no other option.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -549,6 +549,103 @@ bool nl80211_client_impl::get_sta_info(const std::string &interface_name,
         });
 }
 
+bool nl80211_client_impl::get_survey_info(const std::string &interface_name,
+                                          std::vector<survey_info> &survey_info_list)
+{
+    static struct nla_policy survey_policy[NL80211_SURVEY_INFO_MAX + 1];
```
arnout: (nitpick) Declare a variable just before you use it ([ES.21](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-introduce))

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -549,6 +549,103 @@ bool nl80211_client_impl::get_sta_info(const std::string &interface_name,
...
+            struct genlmsghdr *gnlh = static_cast<genlmsghdr *>(nlmsg_data(nlmsg_hdr(msg)));
+            struct nlattr *sinfo[NL80211_SURVEY_INFO_MAX + 1];
+            survey_info survey_info;
+
+            // Parse the netlink message
+            if (nla_parse(tb, NL80211_ATTR_MAX, genlmsg_attrdata(gnlh, 0), genlmsg_attrlen(gnlh, 0),
+                          NULL)) {
+                LOG(ERROR) << "Failed to parse netlink message!";
+                return;
+            }
```
arnout: (nitpick, maybe not for this PR)
Maybe it would make sense now to add a helper function around send_receive_msg that:

- resolves ifname and adds ifindex;
- still allows additional request attributes;
- takes policy and calls nla_parse
- still has an additional lambda to allow nested parsing

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -549,6 +549,103 @@ bool nl80211_client_impl::get_sta_info(const std::string &interface_name,
...
+                LOG(ERROR) << "Failed to parse nested survey attributes!";
+                return;
+            }
+
+            survey_info.in_use = sinfo[NL80211_SURVEY_INFO_IN_USE];
+
+            if (sinfo[NL80211_SURVEY_INFO_FREQUENCY]) {
+                survey_info.frequency_mhz = nla_get_u32(sinfo[NL80211_SURVEY_INFO_FREQUENCY]);
+            } else {
+                LOG(DEBUG) << "NL80211_SURVEY_INFO_FREQUENCY attribute is missing";
```
arnout: I think this should be a WARNING at least - if frequency is missing, it's pretty much useless. And I would skip the survey entirely in that case.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -549,6 +549,103 @@ bool nl80211_client_impl::get_sta_info(const std::string &interface_name,
...
+            }
+
+            if (sinfo[NL80211_SURVEY_INFO_NOISE]) {
+                survey_info.noise_dbm = (int8_t)nla_get_u8(sinfo[NL80211_SURVEY_INFO_NOISE]);
+            } else {
+                LOG(DEBUG) << "NL80211_SURVEY_INFO_NOISE attribute is missing";
```
arnout: If this is missing, it will always be missing. So the log would get very noisy. So I think we should:
- use LOG_N_TIMES or maybe LOG_EVERY_N;
- (maybe) use TRACE instead of DEBUG.

common/beerocks/bwl/shared/nl80211_client_impl.cpp
```diff
@@ -549,6 +549,103 @@ bool nl80211_client_impl::get_sta_info(const std::string &interface_name,
...
+
+            if (sinfo[NL80211_SURVEY_INFO_CHANNEL_TIME_TX]) {
+                survey_info.time_tx_ms = nla_get_u64(sinfo[NL80211_SURVEY_INFO_CHANNEL_TIME_TX]);
+            }
+
+            if (sinfo[NL80211_SURVEY_INFO_TIME_SCAN]) {
+                survey_info.time_scan_ms = nla_get_u64(sinfo[NL80211_SURVEY_INFO_TIME_SCAN]);
+            }
+
+            survey_info_list.push_back(survey_info);
```
arnout: (nitpick) I think we should take the habit of using emplace_back instead of push_back. In this case it doesn't make a difference because none of the fields have move semantics, but if there would be e.g. a string in there, it would be more efficient.

arnout (approved):
Looks goot to me. One small nitpick about the commit message: it should mention nl80211_client in the summary line.

tomereli (approved):

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: Done

common/beerocks/bwl/include/bwl/nl80211_client.h

common/beerocks/bwl/include/bwl/nl80211_client.h

common/beerocks/bwl/include/bwl/nl80211_client.h

common/beerocks/bwl/include/bwl/nl80211_client.h
mariomaz: All functions in nl80211_client API return a bool to signal error and use output parameters for actual results. Newly added function just sticks to this policy to keep coherence.

I'd rather use exceptions to signal errors, but since they are not allowed in this project, I think it's better to use a bool than an empty result so the user of the API can decide whether to log an error and exit or retry after a while, for example. We cannot anticipate all the possible usages of the API, so it's better not to add limitations by design.

common/beerocks/bwl/dummy/nl80211_client_dummy.h
mariomaz: I agree with you. Changed for this case only

common/beerocks/bwl/dummy/nl80211_client_dummy.cpp
mariomaz: Done

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Done (I used TRACE)

common/beerocks/bwl/shared/nl80211_client_impl.cpp

common/beerocks/bwl/shared/nl80211_client_impl.cpp
mariomaz: Fixed

PR #1256, opened 2020-05-08
framework: mapfcommon: change library name from 'common' to 'mapfcommon'

MuteSpirit:
framework: mapfcommon: change library name from "mapf" to "mapfcommon"
    
    After "common" library build the output files are
        lib/mapfcommon.so.1
        lib/mapfcommon.so.1.4.0
        lib/mapfcommon.so
    Thats name is not common for library so file without "lib" prefix.
    Because of that for example Yocto cannot understand that it's library
    file and split correctly between dev and non-dev packages.
    
    Change library target from "common" to "mapfcommon"
    Fix name of library in all CMakeLists.txt files.


tomereli (approved):

framework/common/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ endif()
 
 # Create a single common library
 add_library(common ${sources})
-set_target_properties(common PROPERTIES PREFIX "mapf" VERSION ${prplmesh_VERSION} SOVERSION ${prplmesh_VERSION_MAJOR})
+set_target_properties(common PROPERTIES PREFIX "libmapf" VERSION ${prplmesh_VERSION} SOVERSION ${prplmesh_VERSION_MAJOR})
```
RanRegev: There are [more options](https://stackoverflow.com/questions/31038963/how-do-you-rename-a-library-filename-in-cmake) to get the desired result.
Also note that since [PREFIX](https://cmake.org/cmake/help/latest/prop_tgt/PREFIX.html) property is used, the automated 'lib' prefix is removed, that is the purpose of setting the PREFIX property.

RanRegev (approved):
This change seems correct to me.

framework/common/CMakeLists.txt
tomereli:
```suggestion
set_target_properties(common VERSION ${prplmesh_VERSION} SOVERSION ${prplmesh_VERSION_MAJOR})
```

arnout: In other words, your proposal is to just remove the PREFIX, right? I like that...

arnout: I think the PROPERTIES still need to remain.

RanRegev: CMake is _designed_ to set the prefix based on the target, so if we don't have special case we can leave it to CMake.
However, here it seems that from some reason the name `libcommon` was not enough, the required name is `libmapfcommon`. I don't know why and what were the considerations, but just removing the prefix will leave us with `libcommon`.

vitalybu: Should probably change the library name to `mapfcommon` and remove the custom prefix.

vitalybu (dismissed):

framework/common/CMakeLists.txt
MuteSpirit: Done

MuteSpirit: Change library name from 'common' to 'mapfcommon'

MuteSpirit: change library name from 'common' to 'mapfcommon'

PR #1258, opened 2020-05-10
logs: added informative logs

RanRegev:
during the work on 4.7.9 certification added
logs that may serve other users as well


common/beerocks/bcl/source/beerocks_os_utils.cpp
```diff
@@ -81,7 +81,7 @@ std::string os_utils::system_call(std::string cmd, int log_lvl, bool detached)
     // }
 
     if (system(cmd.c_str()) != 0) {
-        LOG(ERROR) << "System returned non-zero";
+        LOG(ERROR) << "System returned non-zero for the command:" << cmd.c_str();
```
tomereli: If you're already changing this, why not also include the return code?
```
int ret = system(cmd.c_str());
LOG_IF(!ret, ERROR) << cmd << " failed with return code " << ret;
```

framework/transport/ieee1905_transport/ieee1905_transport_local_bus.cpp
```diff
@@ -83,7 +83,9 @@ void Ieee1905Transport::handle_local_bus_cmdu_tx_message(CmduTxMessage &msg)
         }
 
         if (msg.metadata()->msg_type != ntohs(ch->messageType)) {
-            MAPF_WARN("CmduTxMessage messageType mismatch.");
```
tomereli, line 86: OK, but this should be a separate commit.

tomereli (approved):
The change itself is good, but the split to commit is not. Each change should be a separate commit, please split accordingly before merging.

common/beerocks/bcl/source/beerocks_os_utils.cpp
```diff
@@ -81,7 +81,7 @@ std::string os_utils::system_call(std::string cmd, int log_lvl, bool detached)
...
     }
 
 //if (log_lvl == 2) LOG(INFO) << "system_call ret:\n" << ret_str << "\n";
```
vitalybu: Also a good oportunity to remove this comment

vitalybu (approved):
@tomereli has some valid comments, but since they are minor, approving anyway.
@ranregev please implement the changes before merging.

rmelotte (approved):

PR #1267, opened 2020-05-11
build: removed userns when running on WSL

RanRegev:
WSL does not recognize --userns for docker run command
added a special check to not use this option on WSL


rmelotte:
@RanRegev Which docker version are you using?
It's also not supported for some older version of docker on Linux.

vitalybu:
Try updating "Docker for Windows". Mine (v2.3.0.2) uses Docker Engine v19.03.8 which does support this option.

arnout:
Superseded by #1277

PR #1268, opened 2020-05-11
bwl: nl80211_client: init attr. validation policy statically

mariomaz:
Currently we are using static variables of type nla_policy[] to
declare the attribute validation policy to use when calling
nla_parse*() functions. However, it doesn't make sense to declare a
static member which you then write every time and it is not reentrant.

To avoid initializing the array more thant once, a static lambda
function that returns an already initialized static array is used.

Dynamic initialization of function-scope static variables is thread-
safe (per standard C++11 behavior).

At the same time, move variable declarations just before they are used.

### DIFF
--- a/common/beerocks/bwl/shared/nl80211_client_impl.cpp
+++ b/common/beerocks/bwl/shared/nl80211_client_impl.cpp
@@ -299,20 +299,27 @@ bool nl80211_client_impl::get_radio_info(const std::string &interface_name, radi
 
                 struct nlattr *nl_freq;
                 int rem_freq;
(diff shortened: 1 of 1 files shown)


vitalybu (approved):

arnout (approved):
I think a cleaner alternative would be something like:
```
class policy {
   policy(const std::vector<std::pair<unsigned, struct nla_policy>>& policies) {
       for (const auto &p : policies) {
         m_policy[p.first] = p.second;
     }
   }

private:
   const std::array<nla_policy, NL80211_FREQUENCY_ATTR_MAX + 1> m_policy;
};

```

(Still needs a function to convert to `nla_policy*`, but you see where this is going).

That said, maybe I'm overengineering. So OK.

Issue #1082, opened 2020-04-02
[TASK] [Refactor] Unified Agent


morantr:
## Description

The agent architecture is complicated and does not fit to MultiAP standard. 
It has the backhaul manager as an internal switch that forwarding the messages to multiple "son_slaves" which does the same things, but for different radios.
In MultiAp standard there are TLVs that should contain information on both radio, but since we have two separate entities for the radios, we need to send these messages twice.

### Benefits:

1. Reducing processes number:
The current number of processes:
Main Agent, Son_slave0, Son_slave2, Monitor0, Monitor2 - Total 6 Processes
The new architecture number of processes:
Agent, Fronthaul_manager0, Fronthaul_manager2 - Total 3 Processes
**Reduction in 3 processes.**

2. Reducing threads number:
The current number of threads:
Platform_manager, Backhaul_manager, Son_slave0, Son_slave2, Ap_manager0, Ap_manager2, Monitor0, Monitor2 - Total 8 threads
The new architecture number of threads:
Platform_manager, Backhaul_manager, Agent, Ap_manager0, Ap_manager2, Monitor0, Monitor2 - Total 7 threads
**Reduction in 1 thread.**

3. Backhaul manager simplicity.
Currently, the backhaul manager has responsibilities that it should not have, such as:
   - Messages forwarding to son_slaves.
   - Managing as topology manager
   - Searching for the controller with Autoconfig-search message.

   After Unified Agent implementation, the backhaul manager will only be responsible for a backhaul link establishment, and it will not have access to the bus.
  **The key benefit of it, is that there will no more message forwarding from the backhaul manager to two son slaves, and total fewer messages transmitted in the system**

4. One entity controlling the Agent.
Currently, there are two son_slaves that do the same things and send the same messages as duplicates for each radio entity.
On the new architecture, there will be only one Agent thread which will send messages containing information on both radios as the standard allows.
   - Fewer total messages will be sent.
   - Lesser prints to the logs.
   - Simpler handling in the controller side.

5. Lower memory consumption.
Ther are some memory fields that are duplicate for both son_slaves, which unified Agent will save.

### The downside of the new architecture:
Unless we unify the monitor and the ap_manager logs, we would have to create two more log files for the ap_manager threads.

![image](https://user-images.githubusercontent.com/10611984/78244237-dfdb8c00-74ed-11ea-9ccd-cce3e7f734b9.png)

### Phase 1 - Unify ap_manager and monitor thread under the same process.
[...]

PR #1271, opened 2020-05-11
Dev/fronthaul manager

morantr:
This PR creates a new process on the Agent - "Fronthaul_manager", and moves the ap_manager and the monitor threads to it.

See phase 1 on #1082

I am sorry it is a BIG pull request with a lot of commits, but all of them are needed to be in single PR since without all of them the code is broken or not passing the GitHub checks.


framework/external/easylogging/easylogging++.cc
```diff
@@ -2605,7 +2605,12 @@ Writer& Writer::construct(int count, const char* loggerIds, ...) {
 
 void Writer::initializeLogger(const std::string& loggerId, bool lookup, bool needLock) {
   if (lookup) {
-    m_logger = ELPP->registeredLoggers()->get(loggerId, ELPP->hasFlag(LoggingFlag::CreateLoggerAutomatically));
+    auto thread_logger_id = ELPP->getThreadLoggerId();
+    if (thread_logger_id.empty()) {
+      m_logger = ELPP->registeredLoggers()->get(loggerId, ELPP->hasFlag(LoggingFlag::CreateLoggerAutomatically));
+    } else {
+      m_logger = ELPP->registeredLoggers()->get(thread_logger_id, ELPP->hasFlag(LoggingFlag::CreateLoggerAutomatically));
+    }
```
vitalybu:
```suggestion
    auto logger_id = ELPP->getThreadLoggerId().empty() ? loggerId : ELPP->getThreadLoggerId();
    m_logger = ELPP->registeredLoggers()->get(logger_id, ELPP->hasFlag(LoggingFlag::CreateLoggerAutomatically));
```

framework/external/easylogging/easylogging++.h
```diff
@@ -2694,6 +2694,22 @@ class Storage : base::NoCopy, public base::threading::ThreadSafe {
...
+  }
+
+  /// @brief Get attached logger ID of the current thread if exists. Requires std::thread.
+  inline std::string getThreadLoggerId() {
+    base::threading::ScopedLock scopedLock(m_threadIdLoggerIdLock);
+    auto found = m_threadIdLoggerId.find(base::threading::getCurrentThreadId());
+    if (found == m_threadIdLoggerId.end()) {
+      return std::string();
+    }
+    return m_threadIdLoggerId.at(base::threading::getCurrentThreadId());
```
vitalybu:
```suggestion
    return (found == m_threadIdLoggerId.end()) ? std::string() : found.second;
```

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
```diff
@@ -213,6 +213,11 @@ bool monitor_thread::init()
 
 void monitor_thread::after_select(bool timeout)
 {
+    if (!logger_thread_name_set) {
```
vitalybu: `logger_thread_name_set` - name?
Also should probably be moved to `before_select()`

framework/external/easylogging/easylogging++.cc
vitalybu: But, I don't think this change should be here.
You should pass the correct `loggerId` in the first place (from whoever initializes the `Writer` [constructor?]).
When is the `Writer` initialized anyway?

framework/external/easylogging/easylogging++.h
```diff
@@ -3688,6 +3706,14 @@ class Helpers : base::StaticClass {
   static inline void setArgs(int argc, const char** argv) {
     ELPP->setApplicationArguments(argc, const_cast<char**>(argv));
   }
+
+  /// @brief Attach the thread ID to a logger ID . Requires std::thread.
+  static inline void attachLoggerIdThreadId(const std::string& logger_id) {
+    ELPP->attachLoggerIdThreadId(logger_id);
+  }
+  static inline const std::string getThreadLoggerId() {
```
vitalybu:
```suggestion
  static inline const std::string getCurrentThreadLoggerId() {
```

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -17,6 +17,8 @@ namespace beerocks {
 
 #define BEEROCKS_CONTROLLER "beerocks_controller"
 #define BEEROCKS_AGENT "beerocks_agent"
+#define BEEROCKS_FRONTHAUL "beerocks_fronthaul"
+#define BEEROCKS_AP_MANAGER "beerocks_ap_manager"
```
vitalybu, line 21: This will be removed in the future?

morantr: No. The BEEROCKS_FRONTHAUL is for pid file touch, and BEEROCKS_AP_MANAGER/MONITOR is for the logger.

framework/external/easylogging/easylogging++.cc
morantr: I don't like your suggestion since defining a new variable with the same name as one of the function's argument is confusing. Also, the cppcheck will complain about it that it is a variable shadowing.

A `Writer` is an object that is constructed when calling to `LOG(LEVEL)` or `CLOG(LEVEL, ID)`.

> You should pass the correct loggerId in the first place (from whoever initializes the Writer [constructor?]).

Yes, but we don't have it on external libs, which is the whole point of this commit.

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
morantr: Oops! While I wrote it I set the thread on the logger but then I removed it and forgot to change it. I will rename this flag. I also forgot "m_" prefix.
Anyway, let me check it with syslog first, maybe I would have to set the thread name again for it.

The monitor thread does not override `before_select()`. It is redundant to add it just for that.

common/beerocks/bcl/include/bcl/beerocks_logging.h
```diff
@@ -73,12 +69,12 @@ class logging {
 public:
     typedef std::map<std::string, std::string> settings_t;
 
-    logging(const std::string &config_path = std::string(),
-            const std::string &module_name = BEEROCKS_LOGGING_MODULE_NAME);
-    explicit logging(const settings_t &settings, bool cache_settings = false,
-                     const std::string &module_name = BEEROCKS_LOGGING_MODULE_NAME);
+    logging(const std::string &module_name, const std::string &config_path = std::string(),
+            const std::string &logger_id = std::string());
+    explicit logging(const std::string &module_name, const settings_t &settings,
```
arnout: There's no need for an explicit here any more, since it has two mandatory arguments.

However, now we need an explicit in the other constructor, since it has only a single argument so it would be called as an implicit conversion from string.

common/beerocks/bcl/include/bcl/beerocks_logging.h
```diff
@@ -129,6 +127,8 @@ class logging {
     bool m_syslog_enabled = false;
 
     settings_t m_settings_map;
+
+    std::string m_logger_id = "default";
```
arnout, line 133: (nitpick) since there is no default constructor and all constructors have a logger_id argument, I don't think there should be a default here.

common/beerocks/bcl/source/beerocks_logging.cpp
```diff
@@ -301,13 +302,17 @@ logging::logging(const std::string &config_path, const std::string &module_name)
     if (!found_settings) {
         found_settings = load_settings(conf_path);
     }
+    if (!logger_id.empty()) {
+        m_logger_id = logger_id;
+    }
```
arnout, line 307: (nitpick) I think it would be better to put this in the initializer list, like `m_logger_id(logger_id.empty() ? "default" : logger_id)`

(same for the other constructors).

common/beerocks/bcl/include/bcl/beerocks_logging.h
```diff
@@ -73,12 +69,12 @@ class logging {
...
 
-    logging(const std::string &config_path = std::string(),
-            const std::string &module_name = BEEROCKS_LOGGING_MODULE_NAME);
-    explicit logging(const settings_t &settings, bool cache_settings = false,
-                     const std::string &module_name = BEEROCKS_LOGGING_MODULE_NAME);
+    logging(const std::string &module_name, const std::string &config_path = std::string(),
+            const std::string &logger_id = std::string());
+    explicit logging(const std::string &module_name, const settings_t &settings,
+                     const std::string &logger_id = std::string(), bool cache_settings = false);
     logging(const beerocks::config_file::SConfigLog &settings, const std::string &module_name,
```
arnout: Since module_name is the first argument in the other constructors, it would be better to do the same here.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
```diff
@@ -48,6 +48,7 @@ static void handle_signal()
 
         s_pLogger->apply_settings();
         LOG(INFO) << "--- Start of file after roll ---";
+
```
arnout, line 54: (nitpick) unrelated and doesn't seem useful.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -91,12 +91,13 @@ static void copy_vaps_info(std::shared_ptr<bwl::ap_wlan_hal> &ap_wlan_hal,
...
-ap_manager_thread::ap_manager_thread(std::string slave_uds_)
-    : socket_thread(), bss_steer_valid_int(BSS_STEER_VALID_INT),
-      bss_steer_imminent_valid_int(BSS_STEER_IMMINENT_VALID_INT)
+ap_manager_thread::ap_manager_thread(const std::string &slave_uds_, const std::string &iface,
+                                     beerocks::logging &logger)
+    : socket_thread(), m_logger(logger)
 {
     thread_name = "ap_manager";
     slave_uds   = slave_uds_;
+    m_iface     = iface;
```
arnout: Why is this done with assignment instead of with initializer? ([C.49](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#c49-prefer-initialization-to-assignment-in-constructors))

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
```diff
@@ -22,7 +24,8 @@ namespace son {
 class ap_manager_thread : public beerocks::socket_thread {
 
 public:
-    explicit ap_manager_thread(std::string slave_uds_);
+    explicit ap_manager_thread(const std::string &slave_uds_, const std::string &iface,
```
arnout: Explicit is no longer needed.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
```diff
@@ -77,12 +80,19 @@ class ap_manager_thread : public beerocks::socket_thread {
     void remove_client_from_disallowed_list(const sMacAddr &mac, const sMacAddr &bssid);
     void allow_expired_clients();
 
+    // Class constants
+    static constexpr uint8_t BSS_STEER_IMMINENT_VALID_INT = 100;
```
arnout: Genuine question: why constexpr and not just const?

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -92,8 +92,6 @@ enum eGlobals {
     PHY_RATE_100KB_MIN                      = 72,
     PHY_RATE_100KB_INVALID                  = 0,
     BSS_STEER_DISASSOC_TIMER_MS             = 200, // ~200ms
-    BSS_STEER_IMMINENT_VALID_INT            = 100,
-    BSS_STEER_VALID_INT                     = 50,
```
arnout: I like this!

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
tomereli: If you add a member with m_, you're breaking the style in this file. I suggest to add a preparative commit which adds the _m to all members in this file first.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
tomereli: Why did you move these from beerocks_defines.h?

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -167,36 +167,6 @@ void ap_manager_thread::after_select(bool timeout)
 
     auto now = std::chrono::steady_clock::now();
```
tomereli, line 172: Great commit and commit message!

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
```diff
@@ -35,6 +35,8 @@ class ap_manager_thread : public beerocks::socket_thread {
         bool iface_filter_low;
     };
 
+    enum class eApManagerState { INIT, WAIT_FOR_CONFIGURATION, ATTACHING, ATTACHED, OPERATIONAL };
```
tomereli, line 33: Please add a doxygen comment explaining each state.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -202,10 +202,12 @@ eActionOp_APMANAGER:
   _type: enum
   _enum_storage: uint8_t
   ACTION_APMANAGER_INIT_DONE_NOTIFICATION: 0
-  ACTION_APMANAGER_JOINED_NOTIFICATION: 1
-  ACTION_APMANAGER_ENABLE_APS_REQUEST: 2
-  ACTION_APMANAGER_ENABLE_APS_RESPONSE: 3
-  ACTION_APMANAGER_ACK: 4
+  ACTION_APMANAGER_UP_NOTIFICATION: 1
```
tomereli:
```suggestion
  ACTION_APMANAGER_IFACE_UP_NOTIFICATION: 1
```

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -202,10 +202,12 @@ eActionOp_APMANAGER:
...
+  ACTION_APMANAGER_CONFIGURE: 2
```
tomereli: This one definitely needs a better name, I just don't know which one since I have no clue what configure means in this regard.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
```diff
@@ -16,6 +16,21 @@ _multi_class_auto_insert:
 #################################################
 #################################################
  
+cACTION_APMANAGER_UP_NOTIFICATION:
+  _type: class
+  iface_name_length:
+    _type: uint8_t
+    _length_var: True
```
tomereli, line 23: I think that `_length_var` is used only for TLV classes, tlvf is using it to know if it needs to update the length var.
Therefore:
```suggestion
    iface_name_length: uint8_t
```

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -132,30 +132,26 @@ bool ap_manager_thread::init()
...
     }
 
-    //connect to slave //
+    return true;
+}
+
+void ap_manager_thread::connect_to_agent()
+{
+    if (slave_socket) {
+        LOG(WARNING) << "Agent socket is already open";
```
tomereli, line 139: IMO if slave socket is not nullptr, we should add the log but still restart the connection.

tomereli: I now see you only call this when slave_socket is `nullptr`. In this case, this would be more appropriate:
`LOG_IF(slave_socket, FATAL) << "Agent socket is already open!"`

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -158,9 +158,34 @@ void ap_manager_thread::ap_manager_fsm()
 {
     switch (m_state) {
     case eApManagerState::INIT: {
+        if (!slave_socket) {
+            return;
```
tomereli, line 160: Shouldn't there be a warning log here?

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -158,9 +158,34 @@ void ap_manager_thread::ap_manager_fsm()
...
+        m_state = eApManagerState::WAIT_FOR_CONFIGURATION;
+
+        // Set the timout to next select cycle
+        m_state_timeout =
+            std::chrono::steady_clock::now() + std::chrono::milliseconds(SELECT_TIMEOUT_MSC);
         break;
     }
     case eApManagerState::WAIT_FOR_CONFIGURATION: {
+        // On ACTION_APMANAGER_CONFIGURE handler, the ap_hal will be created, and the state will
+        // change to ATTACHING state.
```
tomereli, line 182: Why do we need the "attaching" state? Can't we wait till attached or timeout? 
The name of the state is also weird - "WAIT_FOR_CONFIGURATION" - what configuration are we waiting here?
Perhaps a better name can be "WAIT_FOR_BWL_ATTACH" - we are waiting for the BWL to complete the attach FSM. Once it is done, we exit this state. If there is an error, we will be notified, or a timeout can occur. So I think we can save a couple of states here. But perhaps I'm missing something...

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -143,8 +142,8 @@ class slave_thread : public beerocks::socket_thread {
     void platform_manager_stop();
     void hostap_services_off();
     bool hostap_services_on();
-    bool ap_manager_start();
-    void ap_manager_stop();
+    bool ap_manager_start() { return true; };
```
tomereli: It looks a bit weird that these are nops.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
```diff
@@ -238,21 +238,18 @@ eActionOp_APMANAGER:
   ACTION_APMANAGER_CLIENT_BSS_STEER_REQUEST: 48
   ACTION_APMANAGER_CLIENT_BSS_STEER_RESPONSE: 49
   ACTION_APMANAGER_CLIENT_RX_RSSI_MEASUREMENT_CMD_RESPONSE: 50
```
tomereli, line 240: I guess this should be part of the commit which removed the 4addr connection flow (nitpick)

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
```diff
@@ -203,34 +204,52 @@ int main(int argc, char *argv[])
...
+            touch_time_stamp_timeout = std::chrono::steady_clock::now() +
+                                       std::chrono::seconds(beerocks::TOUCH_PID_TIMEOUT_SECONDS);
+        }
 
-            if (!monitor.work()) {
+        if (!ap_manager.work()) {
+            break;
+        }
+
+        // The ap_manager is the main process thread. After the ap_manager is finish the attach
```
tomereli:
```suggestion
        // The ap_manager is the main process thread. After the ap_manager finishes the attach
```

agent/src/beerocks/slave/son_slave_thread.h
```diff
@@ -86,18 +86,14 @@ class slave_thread : public beerocks::socket_thread {
...
-        STATE_START_MONITOR,
         STATE_WAIT_FOR_MONITOR_JOINED,
-        STATE_MONITOR_JOINED,
-        STATE_WAIT_BEFORE_JOIN_MASTER,
         STATE_BACKHAUL_ENABLE,
         STATE_SEND_BACKHAUL_MANAGER_ENABLE,
         STATE_WAIT_FOR_BACKHAUL_MANAGER_CONNECTED_NOTIFICATION,
         STATE_BACKHAUL_MANAGER_CONNECTED,
         STATE_WAIT_BACKHAUL_MANAGER_BUSY,
+        STATE_WAIT_BEFORE_JOIN_MASTER,
```
tomereli, line 94: Shouldn't this be explained in a bit more details (and perhaps as a separate commit)?

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
```diff
@@ -97,9 +97,10 @@ class socket_thread : public thread_base {
     int server_max_connections;
     SocketSelect select;
 
-    uint32_t m_select_timeout_msec                              = 0;
-    bool m_skip_next_select_timeout                             = false;
-    std::chrono::steady_clock::time_point m_select_wake_up_time = std::chrono::steady_clock::now();
+    uint32_t m_select_timeout_msec  = 0;
+    bool m_skip_next_select_timeout = false;
+    std::chrono::steady_clock::time_point m_select_wake_up_time =
+        std::chrono::steady_clock::time_point::min();
```
tomereli: I'm not sure I understand the problem and fix, but this feels like a bit of a hack

tools/docker/static-analysis/cppcheck.sh
```diff
@@ -21,6 +21,7 @@ run_cppcheck() {
              --force \
              --error-exitcode=1 \
              --enable=warning,style,information \
+             --suppressions-list="$rootdir"/tools/docker/static-analysis/suppressions.txt \
```
tomereli, line 27: I don't understand this commit. Please add explanation.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -132,30 +132,26 @@ bool ap_manager_thread::init()
         return false;
     }
 
-    //connect to slave //
+    return true;
```
arnout, line 127: I don't understand from the commit message why it is better to do the connect to agent from after_select instead of directly from init.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -357,6 +354,32 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+
+        auto config = beerocks_header->addClass<beerocks_message::cACTION_APMANAGER_CONFIGURE>();
+        if (!config) {
+            LOG(ERROR) << "addClass cACTION_APMANAGER_CONFIGURE failed";
+            return false;
+        }
+
+        low_filter   = config->iface_filter_low();
+        m_iface_type = config->iface_type();
+        wifi_channel = config->channel();
```
arnout: AFAICS, this is not used at all, so maybe remove it entirely?

Actually, originally it was used to set acs_enabled, but I don't see that any more? And also I think acs_enabled could be handled better than through the configuration. AFAICS it is only used in two places:
- in dwpal's attach function, but only for error messages;
- in handle_hostap_attached(), to choose between read_acs_report or read_preferred_channels.

Note that *doesn't* seem to be used to actually configure bwl into ACS, so I guess there's a paralle configuration knob in the wireless UCI config to enable ACS.

Therefore, it would make a whole lot more sense to me to handle the ACS option completely within bwl. Indeed, whether ACS is enabled or not, we should simply be able to read the channel preferences from hostapd - the difference is that hostapd will return different preference values.

Obviously, that's too big to include in this PR though...

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -357,6 +354,32 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+        LOG(TRACE) << "ACTION_APMANAGER_CONFIGURE";
+
+        auto config = beerocks_header->addClass<beerocks_message::cACTION_APMANAGER_CONFIGURE>();
+        if (!config) {
+            LOG(ERROR) << "addClass cACTION_APMANAGER_CONFIGURE failed";
+            return false;
+        }
+
+        low_filter   = config->iface_filter_low();
+        m_iface_type = config->iface_type();
```
arnout: This doesn't seem to be used anywhere, maybe remove it?

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -339,6 +358,32 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
...
+
+        LOG(TRACE) << "ACTION_APMANAGER_CONFIGURE";
+
+        auto config = beerocks_header->addClass<beerocks_message::cACTION_APMANAGER_CONFIGURE>();
+        if (!config) {
+            LOG(ERROR) << "addClass cACTION_APMANAGER_CONFIGURE failed";
+            return false;
+        }
+
+        low_filter   = config->iface_filter_low();
```
arnout: This doesn't seem to be used anywhere, maybe remove it?

arnout (commented):
Reviewed up to "agent: Make the Agent compile successfully".

I think the first few commits could be squashed together - but don't bother doing that.

This PR is a bit big, so maybe it's possible to split off a few commits that *don't* break stuff into a separate PR? E.g. remove old 4addr should be non-breaking, and the logger changes as well.

common/beerocks/bcl/include/bcl/beerocks_logging.h
morantr: The logger ID is an optional argument. That's why it is needed.

framework/external/easylogging/easylogging++.h
morantr: added fixup

framework/external/easylogging/easylogging++.h
morantr: added fixup

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: I followed our coding style document. I can refactor this if you want.
But anyway it will be a separate commit at the end.
Please reply if you want me to add a commit that refactors this.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
morantr: Because it is only used in the ap_manager. There isn't a good reason for that to be on beerocks_defines.h.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: Thanks.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
morantr: I think that the state name explains itself, and that trying to explain it will look ridiculous since I would have explain it with the state name. What's done in each state can be understood by looking at what is being done on the fsm function.

All that said, I think that documenting an enum/struct is weird and we should not do it.
We need to document what is not obvious like functions and logic. Documenting any other thing is redundant.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
morantr: This message doesn't notify about an interface that went up, it notifies about the ap_manager thread is up and running.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
morantr: It configures the following:
  iface_type
  channel
  iface_filter_low

These are all needed to configure the ap_hal.
The name seems ok to me. One can always look at which data is being sent on this message on the class implementation.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_apmanager.yaml
morantr: It is a class, so it is ok :)

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: If I remember correctly, it was because I planned to not call the init function from the fronthaul main file.
The only thing left there is  ap_hal command that calls to a shell command "iwpriv" which I'm not sure is still working.

Anyway, I didn't check it and left the init function for now.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: FATAL??? It is a little bit exaggerated don't you think?
FATAL is for serious system errors. This is just a wrong use of the function and therefore prints warning and returns with doing nothing.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: Not critical. If there is an error when opening the socket, it will be printed on the connect function.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: > Why do we need the "attaching" state? Can't we wait till attached or timeout?

The attaching state is needed because the HAL fsm needs to from the ap_manager to call its "attach()" function until it succeeds to attach. This is how it worked before, the only difference it is now on fsm.

>The name of the state is also weird - "WAIT_FOR_CONFIGURATION" - what configuration are we waiting here?
Perhaps a better name can be "WAIT_FOR_BWL_ATTACH" - we are waiting for the BWL to complete the attach FSM. Once it is done, we exit this state. If there is an error, we will be notified, or a timeout can occur. So I think we can save a couple of states here. But perhaps I'm missing something...

Configuration message. I send it on few commits ahead. I mentioned it on the comment in the state.
The "WAIT_FOR_BWL_ATTACH" is not fit since it is just not how the fsm of the hal is working.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: Answered on: https://github.com/prplfoundation/prplMesh/pull/1271#discussion_r426739414

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: Answered on: https://github.com/prplfoundation/prplMesh/pull/1271#discussion_r426739414

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: You are right, I didn't notice that they are not being used anymore.
I will remove `low_filter` and `wifi_channel`.
`m_iface_type` is actually do needed, and it should be set on base_wlan_hal() constructor. For some reason, it uses everywhere `IfaceType::Intel` although it could be another platform.
On the other hand, why we need to set it from outside so we could use get_iface_type() and send it back to ap_manager when son_slave already has this information.

I will remove the above from the config message and from the class. But I'll leave the `acs_enable`. Also, I've just noticed that I forgot to set this member.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
arnout: Although I absolutely applaud the concept of doing small refactorings like this as part of a feature-oriented PR, it is not a good idea to do that if the PR already has 38 commits... So no, please don't do that as part of this PR.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
morantr: Edit: I removed `iface_type` and `iface_filter_low` after @arnout comment.
https://github.com/prplfoundation/prplMesh/pull/1271#discussion_r426739414

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: I just hate to initialize things on the initializer list, and only do it for things I have to.
IMO it is ugly, and it very annoying that the compiler force you to initialize the members in the order they are in the class!
I would have done it if it is a struct that is often being created. But for a class that is created once, I prefer doing it that way.

agent/src/beerocks/slave/son_slave_thread.h
morantr: This is temporary so everything will be compiled successfully. It is removed few on a commit ahead.

common/beerocks/tlvf/yaml/beerocks/tlvf/beerocks_message_action.yaml
morantr: I preferred to split commit that is related to tlvf to make the review more convenient.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
morantr: actually it should be "is finished".

agent/src/beerocks/slave/son_slave_thread.h
morantr: Not critical, I've just moved it to the right location on the enum. Nothing new has been added.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
morantr: It is not a hack, it is a more correct initialization value. I'll try to explain.
There is code on the work() that check how much time it took the thread to do a single select cycle and print a warning message if it exceeds the select timeout value.
In the fronthaul, we create the monitor thread sooner, but we don't start the thread right away. So on the first cycle of the work function it will print a warning message.
This happens since I have chosen to use `now()`, and we get delta time which exceeds the select_timeout.
Choosing `min()` cause to the delta to be negative and thus the condition for select_timeout result in "false" and the error is not printed.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
morantr: Ok.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
morantr: constexpr is set on compile time. It is more like #define macro.
Honestly, I think we should forbid defining constants with `#define` and replace it with `constexpr`.

common/beerocks/bcl/include/bcl/beerocks_logging.h
morantr: Done.

common/beerocks/bcl/include/bcl/beerocks_logging.h
morantr: Done.

common/beerocks/bcl/source/beerocks_logging.cpp
morantr: But it is already "default" by default on the class definition. :)
Yes it will be one line, but it will increase the code size unnecessarily. So no.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
morantr: Why not? It helps to understand when rolling has occurred.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
morantr: Done.

tools/docker/static-analysis/cppcheck.sh
morantr: Add detailed commit message.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
arnout: It seems that [the rest of the developers don't share your opinion](https://www.strawpoll.me/20160132).

That said, no need to change this right now. We have this pattern all over already at the moment. But keep it in mind for future changes.

morantr: But "Preferer assignment in constructors" has the majority of votes.

arnout: At the time I made the comment, only 9 people had voted - I thought that was about everybody but apparently not. Now 7 more have voted and they all chose the "prefer assignment in constructors" option.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1645,10 +1639,31 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
...
+        if (ap_manager_socket) {
+            LOG(ERROR) << "AP manager opened new socket altough the exist socket is already open!";
+            remove_socket(ap_manager_socket);
+            delete ap_manager_socket;
+            ap_manager_socket = nullptr;
+        }
+
         ap_manager_socket = sd;
-        slave_state       = STATE_WAIT_FOR_AP_MANAGER_JOINED;
+        add_socket(ap_manager_socket);
```
arnout, line 1625: TBH, I think it's silly to go through the listen-accept dance for intra-process communication. It would make a lot more sense to simply create a socketpair() and pass one of them to the child. In fact, the same can be used even if it's a separate process, because the file descriptors are inherited by forked children.

But OK, that's a completely unrelated refactoring.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1697,12 +1697,13 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
             supported_channels.begin(), &std::get<1>(tuple_supported_channels),
             &std::get<1>(tuple_supported_channels) + notification->supported_channels_size());
 
+        // If Monitor has already joined the stats has been incremented on the Monitor join handler,
```
arnout: I don't understand this... Do you mean
```suggestion
        // If Monitor has already joined the state has been incremented on the Monitor join handler,
```
?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2424,7 +2425,10 @@ bool slave_thread::handle_cmdu_monitor_message(Socket *sd,
 
     switch (beerocks_header->action_op()) {
     case beerocks_message::ACTION_MONITOR_JOINED_NOTIFICATION: {
-        if (slave_state == STATE_WAIT_FOR_MONITOR_JOINED) {
+        // Since the ap_manager and the Monitor are brought up together, the monitor join could be
+        // received before the ap_manager join message.
+        if (slave_state == STATE_WAIT_FOR_MONITOR_JOINED ||
+            slave_state == STATE_WAIT_FOR_AP_MANAGER_JOINED) {
```
arnout: So now, when we receive the monitor join but no apmanager join, we directly go to backhaul enable state... That doesn't sound right...

I think the only way to start them in parallel is to have a single state "WAIT_FOR_THREADS_JOINED" and a separate boolean for each of them. You don't even need a boolean, just check monitor_socket and ap_manager_socket.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
arnout: I think this would be better:
```suggestion
        std::chrono::steady_clock::time_point::max();
```
Indeed, what you want to avoid is for this condition to trigger:
```
    auto awake_time_msec = std::chrono::duration_cast<std::chrono::milliseconds>(
                               std::chrono::steady_clock::now() - m_select_wake_up_time)
                               .count();

    if (m_select_timeout_msec > 0 && awake_time_msec > m_select_timeout_msec)
```
When you initialize it with min(), awake_time_msec should actually be very large (and positive, since now is definitely later than min) so larger than m_select_timeout_msec. The only reason that it happens to work is that it is so large that it overflows and thus becomes negative. But you can hardly rely on that... With max, it simply be negative and the condition won't trigger.

(My analysis could be wrong though.)

common/beerocks/bcl/include/bcl/beerocks_logging_custom.h
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * @brief This file helps to define and use a custom logger id which is required for using
+ * easylogging with more than one log files on a single thread.
+ * 
+ * To use custom logger id:
+ * 1. Create beerock::logging object with custom logging id string: "logging_id_str"
+ * 2. Define BEEROCKS_CUSTOM_LOGGER_ID macro as logging_id_str":
+ *    #define BEEROCKS_CUSTOM_LOGGER_ID "logging_id_str"
+ * 3. Include this file where there is a need to write to the non-default logger id.
+ */
```
arnout, line 18: I could be wrong, but isn't this exactly the same as defining `ELPP_DEFAULT_LOGGER` before including easylogging++.h?

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
```diff
@@ -10,7 +10,9 @@
 
 #include <bcl/beerocks_utils.h>
 #include <bcl/network/network_utils.h>
-#include <easylogging++.h>
+
+#define BEEROCKS_CUSTOM_LOGGER_ID BEEROCKS_MONITOR
+#include <bcl/beerocks_logging_custom.h>
```
arnout, line 15: As noted elsewhere:
```suggestion

#define ELPP_DEFAULT_LOGGER BEEROCKS_MONITOR
#include <easylogging++.h>
```

Also, just to be sure that easylogging++.h is not included earlier in any of the already-included header files, move the definition before *all* includes.

agent/src/beerocks/fronthaul_manager/monitor/monitor_db.cpp
```diff
@@ -10,7 +10,6 @@
 
 #include <bcl/network/network_utils.h>
 #include <bcl/son/son_wireless_utils.h>
-#include <easylogging++.h>
```
arnout: This *happens* to work because network_utils.h includes easylogging++.h, but it's not good to rely on that. So please don't remove this.

framework/external/easylogging/easylogging++.cc
arnout: > I don't like your suggestion since defining a new variable with the same name as one of the function's argument is confusing. Also, the cppcheck will complain about it that it is a variable shadowing.

I @vitalybu's point was not about the variable name, but rather about limiting the condition to the part which is different (i.e. not repeat the `ELPP->registeredLoggers()->get(foo, ELPP->hasFlag(LoggingFlag::CreateLoggerAutomatically))` part). Of course, by all means, choose a better variable name.

vitalybu: So change the name of the variable.. It's not the point here.
So do you mean we don't have it on external libs? What external libs?

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
vitalybu: It's perfectly fine to add it just for that.
You want to set apply this configuration as soon as possible.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
vitalybu: Maybe in this case the cases are self-explanatory, but in other places this may not be the case.
Documentation is not meant for the developer writing it, but to others that may read the code (or even the generated doxygen file, and in this case they don't even have the source code to look at).

framework/external/easylogging/easylogging++.cc
```diff
@@ -2605,7 +2605,12 @@ Writer& Writer::construct(int count, const char* loggerIds, ...) {
 
 void Writer::initializeLogger(const std::string& loggerId, bool lookup, bool needLock) {
   if (lookup) {
-    m_logger = ELPP->registeredLoggers()->get(loggerId, ELPP->hasFlag(LoggingFlag::CreateLoggerAutomatically));
+    auto thread_logger_id = ELPP->getThreadLoggerId();
```
arnout: I think it would be more appropriate to use the thread logger ID as the equivalent of a default - now the thread logger is taking precedence over an explicitly specified one, which I don't think is appropriate. In other words, define ELPP_CURR_FILE_LOGGER_ID as ELPP->getThreadLoggerId(). However, that would have far-reaching effects, and at least for us it's not necessary, so let's keep it as is.

tools/docker/static-analysis/suppressions.txt
```diff
@@ -0,0 +1,5 @@
+// Suppress warning: Member variable 'X' is not initialized in the constructor.
+uninitMemberVar
```
arnout, line 2: This warning should not be given for a member that is initialized in the struct... For example, in ap_manager_thread, the warning is given for acs_enabled (which is not initialized in the class definition nor in the constructor) but not for ap_hal_ext_events (which is initialized in the class definition but not in the constructor).

So IMO this suppression should not be used... Do you have a concrete example where it goes wrong?

tools/docker/static-analysis/suppressions.txt
```diff
@@ -0,0 +1,5 @@
...
+
+// Suppress performance: Variable 'X' is assigned in constructor body. Consider performing initialization in initialization list.
+useInitializationList
```
arnout, line 5: This one makes sense - since we diverge from the recommendation, it's clear that we need to suppress the warnings for it.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
```diff
@@ -178,15 +178,19 @@ int main(int argc, char *argv[])
         }
     }
 
+    auto syslog = (beerocks_slave_conf.sLog.syslog_enabled == "true");
+
     // Init logger ap_manager
-    std::string base_ap_manager_name = std::string(BEEROCKS_AP_MANAGER) + "_" + fronthaul_iface;
+    std::string base_ap_manager_name =
+        std::string(syslog ? BEEROCKS_FRONTHAUL : BEEROCKS_AP_MANAGER) + "_" + fronthaul_iface;
```
arnout: This smells like a serious break of abstraction... Could this condition somehow be moved into init_logger itself?

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
```diff
@@ -87,10 +87,7 @@ class ap_manager_thread : public beerocks::socket_thread {
     std::string m_iface;
     beerocks::logging &m_logger;
     bool m_ap_manager_configured = false;
-    uint8_t wifi_channel;
-    uint8_t m_iface_type;
     bool acs_enabled;
-    bool low_filter;
```
arnout, line 83: This shouldn't be a fixup commit but a real separate commit.

arnout (approved):
I don't want to look at this any more so I'm approving even though I have so many comments :-)

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
```diff
@@ -213,24 +215,19 @@ bool monitor_thread::init()
...
-    message_com::send_cmdu(slave_socket, cmdu_tx);
-
     add_socket(slave_socket);
     LOG(DEBUG) << "init() end";
     return true;
 }
 
 void monitor_thread::after_select(bool timeout)
 {
+    if (!m_logger_configured) {
```
vitalybu, line 232: Please move this to `before_select`

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
```diff
@@ -147,60 +178,82 @@ int main(int argc, char *argv[])
...
+    // Get Agent UDS file
+    std::string agent_uds =
+        beerocks_slave_conf.temp_path + std::string(BEEROCKS_SLAVE_UDS) + "_" + fronthaul_iface;
+
+    // Create ap_manager
+    son::ap_manager_thread ap_manager(agent_uds, fronthaul_iface, *g_logger_ap_mananger);
+
+    if (!ap_manager.init()) {
+        CLOG(ERROR, g_logger_ap_mananger->get_logger_id()) << "ap manager init() has failed!";
+        return 0;
```
vitalybu: That's an error, so you should probably return a non-zero value.
Same goes for every other error returned by this module.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
```diff
@@ -147,60 +178,82 @@ int main(int argc, char *argv[])
...
+        // The ap_manager is the main process thread. After the ap_manager is finished the attach
+        // process, start the monitor thread. There is no point to start it before.
+        auto ap_manager_state = ap_manager.get_state();
+        if (ap_manager_state < son::ap_manager_thread::eApManagerState::ATTACHED) {
+            continue;
+        } else if (ap_manager_state == son::ap_manager_thread::eApManagerState::ATTACHED) {
+            if (monitor.is_running()) {
+                continue;
+            } else if (!monitor.start()) {
+                CLOG(ERROR, g_logger_monitor->get_logger_id()) << "monitor.init() has failed";
```
vitalybu:
```suggestion
                CLOG(ERROR, g_logger_monitor->get_logger_id()) << "monitor.start() has failed";
```

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
```diff
@@ -147,60 +178,82 @@ int main(int argc, char *argv[])
...
+        }
+
+        if (!ap_manager.work()) {
+            break;
+        }
+
+        // The ap_manager is the main process thread. After the ap_manager is finished the attach
+        // process, start the monitor thread. There is no point to start it before.
+        auto ap_manager_state = ap_manager.get_state();
+        if (ap_manager_state < son::ap_manager_thread::eApManagerState::ATTACHED) {
```
vitalybu: This condition is not really needed.
It's enough to if the AP Manager state is `>= ATTACHED` and the monitor is not running...

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1669,10 +1637,29 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
     }
 
     switch (beerocks_header->action_op()) {
-    case beerocks_message::ACTION_APMANAGER_INIT_DONE_NOTIFICATION: {
-        LOG(INFO) << "received ACTION_APMANAGER_INIT_DONE_NOTIFICATION from sd=" << intptr_t(sd);
+    case beerocks_message::ACTION_APMANAGER_UP_NOTIFICATION: {
+        LOG(INFO) << "received ACTION_APMANAGER_UP_NOTIFICATION from sd=" << intptr_t(sd);
+        if (ap_manager_socket) {
+            LOG(ERROR) << "AP manager opened new socket altough the exist socket is already open!";
```
vitalybu:
```suggestion
            LOG(ERROR) << "AP manager opened new socket altough the existing socket is still open!";
```

vitalybu: Also, not sure it's an ERROR (maybe a WARNING), as we simply close the old socket and continue...

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3760,34 +3687,39 @@ bool slave_thread::hostap_services_on()
...
+    if (ap_manager_socket) {
+        remove_socket(ap_manager_socket);
+        delete ap_manager_socket;
+        ap_manager_socket = nullptr;
+    }
 
-    if (did_stop)
-        LOG(DEBUG) << "monitor_stop() - done";
+    // kill monitor slave pid
+    os_utils::kill_pid(config.temp_path,
```
vitalybu, line 3698: Isn't the monitor is just a thread in our process now? What are we trying to kill here?

common/beerocks/bcl/source/beerocks_logging.cpp
```diff
@@ -506,7 +529,13 @@ void logging::apply_settings()
     syslogConf.set(el::Level::Trace, el::ConfigurationType::Enabled,
                    string_utils::bool_str(m_syslog_levels.trace_enabled()));
 
-    el::Loggers::reconfigureLogger("default", defaultConf);
+    auto logger = el::Loggers::getLogger(m_logger_id);
+    if (!logger) {
+        LOG(ERROR) << "invalid logger!";
+        return;
```
vitalybu, line 537: If this function can now fail, it should return a (boolean?) value...

framework/external/easylogging/easylogging++.h
```diff
@@ -3688,6 +3703,14 @@ class Helpers : base::StaticClass {
   static inline void setArgs(int argc, const char** argv) {
     ELPP->setApplicationArguments(argc, const_cast<char**>(argv));
   }
+
```
vitalybu: Remove extra newline

framework/external/easylogging/easylogging++.cc
morantr: External lib is not accurate. Just "lib".
Rember we talked about that If we want to write to specific logger ID we have to use `CLOG(LEVEL, ID)`.
So we did macro hack for that that could work on files of a specific process, but we had a problem with a print in a lib.
For example if there is a function on network_utils that print something. So in order to make it print to the right file, we would have to provide the longer ID to each function and change every LOG(LEVEL) to CLOG(LEVEL, ID).
So this is the solution that allows us not to provide this, and instead identifying it automatically.

Anyway, I narrowed the condition as you wanted.

framework/external/easylogging/easylogging++.h
morantr: done.

framework/external/easylogging/easylogging++.cc
morantr: Yes, I prefer doing the minimal possible change on easylogging.

common/beerocks/bcl/source/beerocks_logging.cpp
morantr: It could have failed before, I've just moved it.
Anyway, I'm not really sure what to do with that and I don't want to do an additional refactoring as part of this PR.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
morantr: Done, but changed only this return value.
This PR is big enough and I will need to do a separate commit for this.
So probably it will be better to change it here and on any other process as part of a different PR.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
morantr: Yes, you right. Remove it and leave only the condition on the "else".

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
morantr: fixed.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: The thing is, that is should not occur at all, that is why it is an error.
Changed the print.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: It kills the fronthaul process. The comments are wrong. Fix them.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: yes, fixed.

agent/src/beerocks/slave/son_slave_thread.cpp
morantr: Done.

common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
morantr: Apparently both give the same value: https://repl.it/repls/BriefPoorKeygen
But no problem, I'll change it to max.

common/beerocks/bcl/include/bcl/beerocks_logging_custom.h
morantr: Now that you say that and I look at that, it seems that it will give us the same result.
I didn't know about this option...
I will test it, and it will work, I'll remove this commit, and fix the follow-up commit.
I'm just not sure if it will affect other files on another thread.

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
morantr: Why? Currently, we don't override "before_select" on this thread.

tools/docker/static-analysis/suppressions.txt
morantr: If I remember correctly, I say this warning when a class member was initialized on class definition instead of on the initialized list.
I will recheck it.

agent/src/beerocks/fronthaul_manager/fronthaul_manager_main.cpp
morantr: No, I wish...
init_logger is just a static function I created in this file. I still would have to provide the two options so it will be kind of the same.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.h
morantr: You will forgive me for not adding more commits to this PR, right? :)

common/beerocks/bcl/include/bcl/beerocks_logging_custom.h
arnout: ELPP_DEFAULT_LOGGER is only used to set kDefaultLoggerId which is a module-local variable. So it doesn't help for functions from libraries. But it can be used for the "main" functions.

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
arnout: > Why?

Because it makes sure that any logs that are done during the first select are also done correctly.

Ideally, it should be done in a thread initialization function, but beerocks_thread_base doesn't have that (init() is called from the context of the parent, not of the thread itself).

> Currently, we don't override "before_select" on this thread.

That's not a good reason :-)

kantera800 (commented):
This is a big change so please don't merge this until RDKB/UGW platforms will pass sanity,
Issues as #1404 should be resolved first

common/beerocks/bcl/include/bcl/beerocks_logging_custom.h
morantr: @arnout 
I tested it with:
```cpp
#define BEEROCKS_CUSTOM_LOGGER_ID BEEROCKS_MONITOR
#include <bcl/beerocks_logging_custom.h>
```
It doesn't work. I see the monitor constructor print on the ap_manager.
I guess it is because the ap_manager is the default logger.
Anyway, I leave it as I wrote it.

agent/src/beerocks/fronthaul_manager/monitor/monitor_thread.cpp
morantr: Ok added "before_select" and move it to there.

agent/src/beerocks/fronthaul_manager/monitor/monitor_db.cpp
morantr: returned it.

vitalybu (approved):

rmelotte:
I'm also to blame because I didn't review it, but this should never have been merged without at the very least running /some/ of the agent certification tests (using the magic string to also make it run after mergify's rebases).
@morantr  please take it into account for the next parts of https://github.com/prplfoundation/prplMesh/issues/1082

morantr:
> I'm also to blame because I didn't review it, but this should never have been merged without at the very least running /some/ of the agent certification tests (using the magic string to also make it run after mergify's rebases).
> @morantr please take it into account for the next parts of #1082

Since I didn't change anything related to certification flows, it was not required.
All the things I changed has been tested.
The future PRs will actually change things related to the certification flows so it will be tested.
Anyway, as we debug the issue it seems that the root cause is not my change.
@rmelotte

Issue #1272, opened 2020-05-11
[BUG] test_flows.py : beacon metrics query


RanRegev:
**Bug Descrition**
There *might* be inconsistency between the specification and the way beacon metrics query is implemented.

Here we have two tlvs for 0x8011:
* according to the SPEC.
* as implemented

The one that is according to the spec does not work, it fails somewhere along the way.
The incorrect one works.

Investigation is needed.

The probably correct tlv according to the spec:
                tlv(0x99, 0x0016, "{11:22:33:44:55:66 0x73 0xFF 0XFFFFFFFFFFFF 0x02 0x00 0x01 0x02 0x73 0x24 0x30 0x00}"))

This is what is sent (please verify against test_flow.py code):
                tlv(0x99, 0x0014,
                    associated: "{11:22:33:44:55:66}
                    op class: 0x73
                    channel:  0xff
                    bssid: 0xffffffffffff
                    reporting detail: 0x02
                    ssid length: 0x00
                    ap-channel-report-list-len: 0x01
                    inner-channel-report-list-len: 0x02
                    operating class: 0x73
                    channels: 0x24 0x30
                    number of element id: 0"))

arnout:
The problem was that `0XFF` was used in the string, and ucc_listener only supports `0x`.

Issue #325, opened 2019-09-26
[TASK] agent: client steering: support multiple STA


arnout:
## Description

The Client Steering Request message may contain multiple STA MAC addresses. If it does, there may be either one target BSSID per MAC address, or a single target BSSID to be used by all. Support these two scenarios.

## Exit criteria

- Agent client steering supports multiple STA MAC addresses with a single target BSSID.
- Agent client steering supports multiple STA MAC addresses with separate BSSIDs.
- Agent client steering returns an error message in the ACK if either of the above is not satisfied.
- A test is created to validate all of the above scenarios.

PR #1273, opened 2020-05-11
Feature/certification 4.7.9 iii request

RanRegev:
This pull request contains the *query* flow from the controller to the STA for "Beacon Metrics Query".
A.K.A 4.7.9 certification test.
The commits are divided into the different parts of the system, see inside for details.
This is a complementary PR to #1274  PR.
#1274 is built on top of this.
Only together the two PRs implement 4.7.9 certification.


tools/docker/build.sh
```diff
@@ -13,11 +13,22 @@ rootdir="${scriptdir%/*/*}"
 . "${rootdir}/tools/functions.sh"
 
 main() {
+
```
vitalybu: This commit should be dropped

framework/tlvf/yaml/tlvf/wfa_map/tlvBeaconMetricsQuery.yaml
```diff
@@ -33,12 +33,12 @@ tlvBeaconMetricsQuery:
     _type: uint8_t
     _length: [ elemnt_id_list_length ]
 
-sApChannelReports:
-  _type: struct
+cApChannelReports:
+  _type: class
```
vitalybu, line 37: Why did you change the type to `class`?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -105,6 +106,36 @@ namespace beerocks {
...
+template <class SlavesContainer, class MAC>
+std::shared_ptr<backhaul_manager::sRadioInfo> findRadioInfo(const SlavesContainer &slave_container,
+                                                            const MAC &mac)
+{
+    // for each radio info in the container
+    //      for each associated client
+    //          find if there is an entry with the given mac within this associated client
+
+    for (const auto &slave : slave_container) {
+        if (slave) {
```
```suggestion
        if (!slave) {
            continue;
        }
        ...
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -105,6 +106,36 @@ namespace beerocks {
 #define FSM_IS_IN_STATE(eState) (m_eFSMState == EState::eState)
 #define FSM_CURR_STATE_STR s_arrStates[int(m_eFSMState)]
 
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Helper Functions /////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+template <class SlavesContainer, class MAC>
+std::shared_ptr<backhaul_manager::sRadioInfo> findRadioInfo(const SlavesContainer &slave_container,
+                                                            const MAC &mac)
```
vitalybu: Why a template?
I don't think you'll have different types here

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -253,25 +284,26 @@ bool backhaul_manager::init()
...
-            ieee1905_1::eMessageType::LINK_METRIC_QUERY_MESSAGE,
-            ieee1905_1::eMessageType::ASSOCIATED_STA_LINK_METRICS_QUERY_MESSAGE,
             ieee1905_1::eMessageType::COMBINED_INFRASTRUCTURE_METRICS_MESSAGE,
-            ieee1905_1::eMessageType::ACK_MESSAGE,
+            ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE,
+            ieee1905_1::eMessageType::LINK_METRIC_QUERY_MESSAGE,
+            ieee1905_1::eMessageType::MULTI_AP_POLICY_CONFIG_REQUEST_MESSAGE,
+            ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE,
+            ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE,
+            ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE,
```
vitalybu, line 278: This should be a separate commit (similar to the one you did in tlvf)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1610,16 +1643,25 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
...
         // Message from controller (bus) to agent (uds)
         // Send the data (uds_header + cmdu) how it is on UDS, without changing it
+        //
 
-        // Forward cmdu to all slaves how it is on UDS, without changing it
         cmdu_rx.swap(); // swap back before forwarding
-        for (auto soc_iter : slaves_sockets) {
-            if (!message_com::forward_cmdu_to_uds(soc_iter->slave, cmdu_rx, length)) {
+
+        if (!destinationAP) {
```
vitalybu: Why so negative? ;)
Unless checking for errors, I think it's more readable to make positive conditions:
```c++
if (true) {
    ...
} else {
    ...
}
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1601,7 +1633,8 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
         // Handle the CMDU message. If the message was processed locally
         // (by the Backhaul Manager), this function will return 'true'.
         // Otherwise, it should be forwarded to the slaves.
-        if (handle_1905_1_message(cmdu_rx, src_mac)) {
+        Socket *destinationAP = nullptr;
```
vitalybu: We're not camelCasing:
```suggestion
        Socket *dest_slave = nullptr;
```

Also, maybe it's a good idea to update the comment explaining the new variable.
And extract this entire change into a separate commit...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1610,16 +1643,25 @@ bool backhaul_manager::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_r
 
         // Message from controller (bus) to agent (uds)
         // Send the data (uds_header + cmdu) how it is on UDS, without changing it
+        //
```
```suggestion
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
     return true;
 }
 
+bool backhaul_manager::handle_1905_beacon_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac,
+                                                        Socket *&forward_to)
+{
+    LOG(INFO) << "now going to handle BEACON METRICS QUERY";
```
vitalybu: That's a DEBUG message

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+
+    // extract the desired STA mac
+    auto tlvBeaconMetricsQuery = cmdu_rx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+    if (!tlvBeaconMetricsQuery) {
+        LOG(ERROR) << "handle_1905_beacon_metrics_query should handle only tlvBeaconMetrics, but "
+                      "got something else: 0x"
+                   << std::hex << (uint16_t)cmdu_rx.getMessageType();
```
vitalybu, line 3301: I don't understand this error. The message type is correct, otherwise the handler wouldn't have been called.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+        return false;
+    }
+
+    const sMacAddr &requiredMac = tlvBeaconMetricsQuery->associated_sta_mac();
+
```
```suggestion
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+    LOG(DEBUG) << "the requested STA mac is: " << requiredMac;
```
vitalybu: Requested or required?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+    // extract the desired STA mac
+    auto tlvBeaconMetricsQuery = cmdu_rx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+    if (!tlvBeaconMetricsQuery) {
+        LOG(ERROR) << "handle_1905_beacon_metrics_query should handle only tlvBeaconMetrics, but "
+                      "got something else: 0x"
+                   << std::hex << (uint16_t)cmdu_rx.getMessageType();
+        return false;
+    }
+
+    const sMacAddr &requiredMac = tlvBeaconMetricsQuery->associated_sta_mac();
```
```suggestion
    const sMacAddr &sta_mac = tlvBeaconMetricsQuery->associated_sta_mac();
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -105,6 +106,36 @@ namespace beerocks {
 #define FSM_IS_IN_STATE(eState) (m_eFSMState == EState::eState)
 #define FSM_CURR_STATE_STR s_arrStates[int(m_eFSMState)]
 
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Helper Functions /////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+template <class SlavesContainer, class MAC>
+std::shared_ptr<backhaul_manager::sRadioInfo> findRadioInfo(const SlavesContainer &slave_container,
```
```suggestion
std::shared_ptr<backhaul_manager::sRadioInfo> find_radio_info(const SlavesContainer &slave_container,
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+                      "got something else: 0x"
+                   << std::hex << (uint16_t)cmdu_rx.getMessageType();
+        return false;
+    }
+
+    const sMacAddr &requiredMac = tlvBeaconMetricsQuery->associated_sta_mac();
+
+    LOG(DEBUG) << "the requested STA mac is: " << requiredMac;
+
+    auto radio = findRadioInfo(slaves_sockets, requiredMac);
```
vitalybu: Why not a const?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+    // build ACK message CMDU
+    const auto mid      = cmdu_rx.getMessageId();
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+
+    if (!radio) {
+        LOG(WARNING) << "couldn't find any agent for the requested mac: " << requiredMac;
```
```suggestion
        LOG(DEBUG) << "STA with MAC [" << sta_mac << "] is not associated with any BSS operated by the agent";
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+
+        // debug
+        std::stringstream errorSS;
+        auto error_tlv = cmdu_tx.getClass<wfa_map::tlvErrorCode>();
+        if (error_tlv) {
+            errorSS << "0x" << error_tlv->reason_code();
+        } else {
+            errorSS << "no error";
+        }
+        // end debug
```
vitalybu: Remove

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+    }
+
+    const sMacAddr &requiredMac = tlvBeaconMetricsQuery->associated_sta_mac();
+
+    LOG(DEBUG) << "the requested STA mac is: " << requiredMac;
+
+    auto radio = findRadioInfo(slaves_sockets, requiredMac);
+
+    // build ACK message CMDU
+    const auto mid      = cmdu_rx.getMessageId();
```
vitalybu, line 3311: Perhaps the entire ACK + Error flow should be extracted into a local static function.
I guess that almost every flow may have errors and will have to send this response.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3090,6 +3135,79 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+    LOG(DEBUG) << "found the radio that has the sation. radio: " << radio->radio_mac
+               << "; station: " << requiredMac;
+
+    LOG(DEBUG) << "BEACON METRICS QUERY: sending ACK message to the originator mid: "
+               << int(mid); // USED IN TESTS
+
+    send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
+
+    // continue processing
+    return false;
```
vitalybu, line 3361:
```suggestion
    return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
```

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -995,8 +996,20 @@ bool master_thread::handle_cmdu_1905_ack_message(const std::string &src_mac,
...
+    std::stringstream errorSS;
+    auto error_tlv = cmdu_rx.getClass<wfa_map::tlvErrorCode>();
+    if (error_tlv) {
+        errorSS << "0x" << error_tlv->reason_code();
+    } else {
+        errorSS << "no error";
+    }
+
+    LOG(DEBUG) << "Received ACK_MESSAGE, mid=" << std::hex << int(mid)
+               << " tlv error code: " << errorSS.str();
```
vitalybu, line 1009:
```suggestion
    LOG(DEBUG) << "Received ACK_MESSAGE, mid=" << std::hex << int(mid)
               << " tlv error code: " << (error_tlv) ? error_tlv->reason_code() : "no error";
```

agent/src/beerocks/slave/1905_beacon_query_to_vs.cpp
```diff
@@ -0,0 +1,113 @@
+
+#include "1905_beacon_query_to_vs.h"
+#include <bcl/beerocks_defines.h>
+#include <mapf/common/utils.h>
+
+namespace gate {
```
vitalybu: gate?

agent/src/beerocks/slave/1905_beacon_query_to_vs.h
```diff
@@ -0,0 +1,26 @@
...
+#define _1905_BEACONB_QUERY_TO_VS_H_
+
+#include <beerocks/tlvf/beerocks_message_monitor.h>
+#include <tlvf/CmduMessageRx.h>
+#include <tlvf/wfa_map/tlvBeaconMetricsQuery.h>
+
+namespace gate {
+
+// loads lhs with rhs. as if it was written: lhs = rhs; or lhs << rhs;
+// however load() was chosen over operator=() or operator<<() as it permits returning success/fail with the bool return value,
```
vitalybu: Why not simply return `nullptr` on failure?
Using operators is more intuitive IMO

agent/src/beerocks/slave/1905_beacon_query_to_vs.cpp
```diff
@@ -0,0 +1,113 @@
+
```
vitalybu: Adding a separate file for every transformation feels like an overkill.
If we're going this way, and I'm still not sure it's the right way, we should have all the functions in one file.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -739,6 +739,28 @@ bool mon_wlan_hal_dwpal::sta_beacon_11k_request(const SBeaconRequest11k &req, in
...
+        std::stringstream ch_report_ss;
+        uint8_t current_channel = 0;
+        for (; current_channel < req.use_optional_ap_ch_report - 1; ++current_channel) {
+            ch_report_ss << +req.ap_ch_report[current_channel] << ",";
+        }
+        ch_report_ss << +req.ap_ch_report[current_channel];
+
+        cmd += " ap_ch_report=" + ch_report_ss.str();
+    } else {
+        LOG(WARNING) << "optinal channel report was not set (might be also ok)";
```
vitalybu: > might be also ok

Isn't the definition of optional that it may be missing?

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -739,6 +739,28 @@ bool mon_wlan_hal_dwpal::sta_beacon_11k_request(const SBeaconRequest11k &req, in
...
+    }
+
+    // report detail: rep_detail=
+    // this value should be set by default to 2 if not stated otherwise.
+    // however hostap does not set it to 2. So we add it hardcoded here
+    // note: currently there is no field in SBeaconRequest11k that holds this optional field
+    cmd += " rep_detail=2";
+
+    LOG(DEBUG) << __func__ << " the command: " << cmd;
```
vitalybu: Remove

vitalybu (dismissed):

agent/src/beerocks/slave/1905_beacon_query_to_vs.cpp
RanRegev: No. Separate files are the correct direction if we go this way.
Tiny pieces are always better in terms of decoupling and modularity.
Once we place together pieces that have no relation between them we couple unnecessarily.
This in turn makes it harder and harder to do things differently. We loose flexibility.
In addition, the benefits of storing the entire code only in few files is questionable - what are the benefits?

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: I felt uncomfortable with this warning the moment I wrote it and ever since when I see it.
The point is that `ap_ch_report` is an _optional_ field. It means that it might wasn't set deliberately - which is ok( hence `might be also ok`). on the other hand, you might _wanted_ to set it but something went wrong (hence the warning).
I didn't find a good way to express this in a short message.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: Remove what?
Should I remove the only debug log message along the way that shows me what was actually sent? If it was me to decide I would make it an `INFO` message. This line tells a story in the log. I think that when you read the log, you should understand what was happening not based only on errors (the way it was up till this line), you should have beacons that signal to you "I am here, everything is fine". 
Of course that a balance is needed: if the line I added creates 10th or 100th of these signals along the log, this message shouldn't be. I think that this specific line is correct as it prints only beacon query messages.

framework/tlvf/yaml/tlvf/wfa_map/tlvBeaconMetricsQuery.yaml
RanRegev: the yaml struct does not support variable length lists. I'll amend the commit to describe the change better.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: Done.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: I agree. However, I spent SO much time on this from few reasons. I had to do, redo and re-re-do the change, that I really prefer not to go into this specific area again.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: Me also don't think I'll ever have different types here.
However, this is a free function, not part of the backhaul manager. Potentially it has nothing to do with the exact types that are kept in the backhaul manager. Potentially the backhaul manager may decide to keep the data differently.
I keep saying  `potentially` because the amount of the constrains that are applied on the parameters and return value of this function make it unlikely to ever be reused.
Bottom line: I think I'll use the backhaul types for this free function.

agent/src/beerocks/slave/1905_beacon_query_to_vs.cpp
vitalybu: Ok, in this case just put these files in a subfolder.

agent/src/beerocks/slave/1905_beacon_query_to_vs.h
RanRegev: 1. from the code: `and also it opens the possibilities for adding additional parameters in the future if needed`
2. operator in this case is non natural:
user code: 
```cpp
auto vsPtr = message_com::create_vs_message</*....*/>( /*....*/);
vsPtr = rx; // operator=() may change lhs to point to nullptr, this is strange. normally you don't expect this. Especially you don't expect that operator=() will take care for life cycle of the given pointer. In this case it is might be deleting it in case the operation fails.
if (vsPtr) // ??? maybe, I don't know.
```
```cpp
vsPtr << rx // even less natural. you wouldn't imagine that vsPtr will point to something else at the end of this operation.
```
Moreover, I really wanted to write the load function to have `beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST &` as its input parameter, I don't know why I finally decided not to.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
vitalybu: You don't issue warnings for missing optional fields. Simply remove it.

agent/src/beerocks/slave/1905_beacon_query_to_vs.cpp
RanRegev: :-)
There was a naming discussion before the decision was taken.
The bottom line of the discussion was "whatever name is selected, a comment will be"
So here is the logic, and the options, and personally I have no strong feeling about any of the options:
The functionality of this namespace is to convert data type from one domain to its counterpart in another domain. In this example 1905 beacon request to intel's counterpart (vendor specific)
Normally I would call this namespace `bridge` because it bridges between two worlds, two domains. In the large view of this concept it is not just translates from one domain to another, it also might take aggregation operations, filtering, etc (imagine that you need 2 1905 messages before you can construct a VS specific message. The `bridge`, in between the two domains, is the place to make it. )
This is the dependency structure: 
* 1905 world - knows noting about vs
* bridge/gate - knows both worlds: 1905, vs
* vs specific - knows nothing about 1905

The name should reflect this nature.
The use of `bridge`, in my opinion, is less affordable in the networking world as it has a well known and very precise meaning. therefore `gate` was chosen. However `gate` might also be too close to `gateway`.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: I'll remove but note that this is a valid situation:

>  you might wanted to set it but something went wrong

In this specific case I can check another field (`channel` I think) and if its value is 255 then it indicates that ap_ch_report _must_ exists.

framework/tlvf/tlvf_conf.yaml
```diff
@@ -10,35 +10,34 @@ include_yaml_path: {
   # "wfa_map/",
   "tlvf/wfa_map/eTlvTypeMap.yaml",
   "tlvf/wfa_map/tlvApCapability.yaml",
-  "tlvf/wfa_map/tlvChannelPreference.yaml",
-  "tlvf/wfa_map/tlvSupportedService.yaml",
+  "tlvf/wfa_map/tlvApHeCapabilities.yaml",
```
rmelotte, line 13: There are some extra whitespaces at the end of some of the lines (this one for example).

(nitpick)

framework/tlvf/yaml/tlvf/wfa_map/tlvBeaconMetricsQuery.yaml
```diff
@@ -33,12 +33,12 @@ tlvBeaconMetricsQuery:
...
   ap_channel_report_list_length:
     _type: uint8_t
     _length_var: True
-  operating_class_in_ap_channel_report: uint8_t
   ap_channel_report_list:
     _type: uint8_t
     _length: [ ap_channel_report_list_length ]
+    _comment: The first element of this list is the operating class.
```
rmelotte: There are 2 trailing whitespaces at the end (nitpick).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -253,25 +286,26 @@ bool backhaul_manager::init()
...
+            ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE,
+            ieee1905_1::eMessageType::ASSOCIATED_STA_LINK_METRICS_QUERY_MESSAGE,
+            ieee1905_1::eMessageType::BEACON_METRICS_QUERY_MESSAGE,
             ieee1905_1::eMessageType::CHANNEL_PREFERENCE_QUERY_MESSAGE,
             ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE,
-            ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE,
-            ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE,
             ieee1905_1::eMessageType::CLIENT_ASSOCIATION_CONTROL_REQUEST_MESSAGE,
-            ieee1905_1::eMessageType::HIGHER_LAYER_DATA_MESSAGE,
+            ieee1905_1::eMessageType::CLIENT_CAPABILITY_QUERY_MESSAGE,
```
rmelotte, line 270: It looks like you sorted the list, and also added a new element.
Ideally the sorting should be done in a separate commit because it makes the diff harder to understand. 
At the very least, please mention it in the commit message.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4536,6 +4542,31 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
     }
 }
 
+bool slave_thread::handle_beacon_metrics_query_request(Socket *sd,
```
rmelotte: Why is it called query_request? Is it a typo?

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
```diff
@@ -0,0 +1,113 @@
+
+#include "1905_beacon_query_to_vs.h"
+#include <bcl/beerocks_defines.h>
+#include <mapf/common/utils.h>
+
+namespace gate {
```
rmelotte, line 7: For my own curiosity: why did you choose "gate" as a namespace? What does it refer to?

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4536,6 +4542,31 @@ bool slave_thread::handle_client_steering_request(Socket *sd, ieee1905_1::CmduMe
...
+                                                       ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received BEACON_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid);
+
+    // create vs message
+    auto request_out =
+        message_com::create_vs_message<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST>(
+            cmdu_tx, mid);
+    if (request_out == nullptr) {
```
rmelotte, line 4566:
```suggestion
    if (!request_out) {
```

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include <beerocks/tlvf/beerocks_message_monitor.h>
+#include <tlvf/CmduMessageRx.h>
+#include <tlvf/wfa_map/tlvBeaconMetricsQuery.h>
+
+namespace gate {
+
+// loads lhs with rhs. as if it was written: lhs = rhs; or lhs << rhs;
+// however load() was chosen over operator=() or operator<<() as it permits returning success/fail with the bool return value,
+// and also it opens the possibilites for adding additional parameters in the future if needed
```
rmelotte: Please use a doxygen format for the documentation.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -739,6 +739,30 @@ bool mon_wlan_hal_dwpal::sta_beacon_11k_request(const SBeaconRequest11k &req, in
         cmd += " ssid=" + req_ssid;
     }
 
+    // use ap_ch_report in the request.
+    if (req.use_optional_ap_ch_report) {
+        std::stringstream ch_report_ss;
+        uint8_t current_channel = 0;
+        for (; current_channel < req.use_optional_ap_ch_report - 1; ++current_channel) {
+            ch_report_ss << +req.ap_ch_report[current_channel] << ",";
```
rmelotte, line 775: What is this "+"? it looks like a typo

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -739,6 +739,30 @@ bool mon_wlan_hal_dwpal::sta_beacon_11k_request(const SBeaconRequest11k &req, in
...
+        }
+    }
+
+    // report detail: rep_detail=
+    // this value should be set by default to 2 if not stated otherwise.
+    // however hostap does not set it to 2. So we add it hardcoded here
+    // note: currently there is no field in SBeaconRequest11k that holds this optional field
+    cmd += " rep_detail=2";
+
+    LOG(DEBUG) << __func__ << " the command: " << cmd;
```
rmelotte: Please drop the `__func__`, AFAIK it shouldn't be needed.

tests/test_flows.py
```diff
@@ -851,6 +851,28 @@ def test_topology(self):
...
+        env.agents[0].radios[0].vaps[0].associate(sta)
+
+        # send beacon query request (please take a look at https://github.com/prplfoundation/prplMesh/issues/1272)
+        debug("Sending beacon report query to repeater:")
+        request = '{mac} 0x73 0xFFFFFFFFFFFF 0x02 0x00 0x01 0x03 0x73 0x24 0x30'.format(mac=sta.mac)
+        debug(request)
+        mid = env.controller.dev_send_1905(env.agents[0].mac, 0x8011,
+                                     tlv(0x99, 0x0014, "{" + request + "}"))
+
+        self.check_log(env.agents[0], r"BEACON METRICS QUERY: sending ACK message to the originator mid: {:d}".format(mid))
```
rmelotte: Now that we have packet captures integrated into test_flows.py, we should use it whenever it's possible.
Here you should be able to check that the "ACK" message is sent by using `check_cmdu_type`

tests/test_flows.py
```diff
@@ -861,17 +861,21 @@ def test_beacon_report_query(self):
...
         request = '{mac} 0x73 0xFFFFFFFFFFFF 0x02 0x00 0x01 0x03 0x73 0x24 0x30'.format(mac=sta.mac)
         debug(request)
         mid = env.controller.dev_send_1905(env.agents[0].mac, 0x8011,
-                                     tlv(0x99, 0x0014, "{" + request + "}"))
+                                           tlv(0x99, 0x0014, "{" + request + "}"))
 
-        self.check_log(env.agents[0], r"BEACON METRICS QUERY: sending ACK message to the originator mid: {:d}".format(mid))
+        self.check_log(env.agents[0],
+                       r"BEACON METRICS QUERY: "
+                       r"sending ACK message to the originator mid: {:d}".format(mid))
```
rmelotte, line 1062: I guess this is a fixup commit, but it's not marked as such. Before merging it should be squashed with the one that adds the changes.

rmelotte (approved):
Many small comments, but I don't see anything blocking, so approving.

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.h
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#ifndef _1905_BEACONB_QUERY_TO_VS_H_
+#define _1905_BEACONB_QUERY_TO_VS_H_
+
+#include <beerocks/tlvf/beerocks_message_monitor.h>
+#include <tlvf/CmduMessageRx.h>
+#include <tlvf/wfa_map/tlvBeaconMetricsQuery.h>
+
+namespace gate {
```
arnout: Currently everything is in namespace beerocks (or mapf), so please also put this in namespace beerocks::gate.

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+bool load(std::shared_ptr<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST> lhs,
+          const ieee1905_1::CmduMessageRx &rhs)
+{
+    // escape I
```
arnout, line 12: (nitpick) IMHO this comments doesn't contribute much.

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
```diff
@@ -0,0 +1,113 @@
...
+                             beerocks::message::WIFI_SSID_MAX_LENGTH);
+
+    // how many elements in the list of channels, we support only 1 at the moment
+    auto ap_ch_report_length = beacon_metrics_query->ap_channel_reports_list_length();
+
+    // escape III - data validation
+    // ap_ch_report_length != 0 ==> channel_number == 255
+    // we verify with the already extracted values (we could do it withthe original as well)
+    if (ap_ch_report_length != 0 && lhs_params.channel != 255) {
+        LOG(ERROR) << "inconsistency between channel report length and channel number. please take "
```
arnout, line 62: (nitpick) The message comes from the outside, so referring to the spec is of limited value.

arnout (commented):
Partial review only.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -739,6 +739,30 @@ bool mon_wlan_hal_dwpal::sta_beacon_11k_request(const SBeaconRequest11k &req, in
         cmd += " ssid=" + req_ssid;
     }
 
+    // use ap_ch_report in the request.
+    if (req.use_optional_ap_ch_report) {
+        std::stringstream ch_report_ss;
+        uint8_t current_channel = 0;
+        for (; current_channel < req.use_optional_ap_ch_report - 1; ++current_channel) {
```
arnout, line 774: According to the yaml file, use_optional_ap_ch_report is a boolean. So I think this should be 
```suggestion
        for (; current_channel < sizeof(ap_ch_report) - 1; ++current_channel) {
            if (req.ap_ch_report[current_channel] == 0)
                break;
```

Or if you want to use it as a length field, rename it in the yaml.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
arnout, line 775: (nitpick) the rest of this file uses the pattern
```suggestion
            cmd += std::to_string(req.ap_ch_report[current_channel]) + ",";
```
I think it's better to be consistent.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -739,6 +739,30 @@ bool mon_wlan_hal_dwpal::sta_beacon_11k_request(const SBeaconRequest11k &req, in
...
+            ch_report_ss << +req.ap_ch_report[current_channel] << ",";
+        }
+        ch_report_ss << +req.ap_ch_report[current_channel];
```
arnout, line 777: (nitpick) I guess it's just a matter of habit, but it's easier on my brain to see something like:
```
foo += bar[0];
for (int i = 1; i < length; i++) {
   foo += "," + bar[i];
}
```

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -739,6 +739,30 @@ bool mon_wlan_hal_dwpal::sta_beacon_11k_request(const SBeaconRequest11k &req, in
...
+
+        cmd += " ap_ch_report=" + ch_report_ss.str();
+    } else {
+        if (255 == req.channel) {
```
arnout, line 781: Maybe there should be a similar check that channel must be 255 if the ap_ch_report list is given?

tests/test_flows.py
```diff
@@ -851,6 +851,28 @@ def test_topology(self):
...
+
+        # for testing non existing STA, when want to test the error flow
+        # sta1 = env.Station.create()
+
+        debug("Connect dummy STA (" + sta.mac + ") to wlan0")
+        env.agents[0].radios[0].vaps[0].associate(sta)
+
+        # send beacon query request (please take a look at https://github.com/prplfoundation/prplMesh/issues/1272)
+        debug("Sending beacon report query to repeater:")
+        request = '{mac} 0x73 0xFFFFFFFFFFFF 0x02 0x00 0x01 0x03 0x73 0x24 0x30'.format(mac=sta.mac)
```
arnout: This request can't even parse: the channel field is missing and the number of element IDs field as well. This should parse as:
- bssid: FF:FF:FF:FF:FF:02
- reporting detail: 0
- ssid: "\x03"
- number of channel reports: 0x73
- first channel report: 
    - length: 0x24
    - operating class: 0x30 (invalid operating class)
    - the rest is cut off

And indeed, [the log file shows](https://prpl-foundation.gitlab.io/-/prplMesh/-/jobs/551918886/artifacts/logs/repeater1-551918886/beerocks_agent_wlan0.20200514_080859.log) that the conversion fails:
```
DEBUG 08:09:39:796 <140122159392000> son_slave_thread.cpp[4549] --> Received BEACON_METRICS_QUERY_MESSAGE, mid=0x8004
ERROR 08:09:39:796 <140122159392000> 1905_beacon_query_to_vs.cpp[71] --> too many channles requested (the maximum is 237, current support is only for 1): s
```

So if the point of this test was to show that the conversion function fails if you send an invalid packet, I guess it's an OK test :-)


I replaced this with the correct TLV as specified in #1272, and then the following appears in the monitor log:
```
TRACE 09:54:59:722 <139820967271296> monitor_thread.cpp[1051] --> received ACTION_MONITOR_CLIENT_BEACON_11K_REQUEST
TRACE 09:54:59:722 <139820967271296> mon_wlan_hal_dummy.cpp[116] --> sta_beacon_11k_request
DEBUG 09:54:59:722 <139820967271296> mon_wlan_hal_dummy.cpp[121] --> Beacon 11k request to sta 51:a1:10:20:00:00 on bssid e6:9d:a0:33:43:5c channel 1
WARNING 09:54:59:722 <139820967271296> mon_wlan_hal_dummy.cpp[136] --> No RRM beacon response for sta 51:a1:10:20:00:00
DEBUG 09:54:59:723 <139820967271296> monitor_thread.cpp[1102] --> inserting 1 RRM_EVENT_BEACON_REP_RXED event(s) to the pending list
```
... which looks correct to me. In addition, the certification test (which passes) gives the correct TLV as well. So I really don't understand what #1272 is about...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -105,6 +106,38 @@ namespace beerocks {
 #define FSM_IS_IN_STATE(eState) (m_eFSMState == EState::eState)
 #define FSM_CURR_STATE_STR s_arrStates[int(m_eFSMState)]
 
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Helper Functions /////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+template <class SlavesContainer, class MAC>
+std::shared_ptr<backhaul_manager::sRadioInfo> findRadioInfo(const SlavesContainer &slave_container,
```
arnout: In addition to the naming issue as pointed out by Vitaly:
- IMO it would be better as a private member of backhaul_manager_thread (and without the slave_container argument);
- IMO there's absolutely no reason for it to be a template. The container *has* to be a container with std::shared_ptr<sRadioInfo> as the value type (although admittedly it could be a vector, list or set), and there is absolutely no reason for MAC to be anything other than sMacAddr.
- It should be possible to search the radio by radio_uid, VAP bssid, or connected STA MAC address, and probably in other ways as well. So the name should specify how you search for it: `find_radio_info_by_sta_mac`.
- The function should have a doxygen explanation.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -105,6 +106,38 @@ namespace beerocks {
...
+    // for each radio info in the container
+    //      for each associated client
+    //          find if there is an entry with the given mac within this associated client
+
+    for (const auto &slave : slave_container) {
+        if (!slave) {
+            continue;
+        }
+
+        LOG(DEBUG) << "looking into info for hostap iface: " << slave->hostap_iface;
```
arnout: General comment/discussion: if we keep this kind of super-detailed debug output for everything, the logs will just explode and the debug output becomes fairly useless. Since it is quite easy to add this specific log message while debugging this specific function, you can also remove it again when you're satisfied the function works. IMO the only logs that should remain is stuff that makes sense at the system level. E.g. probably the caller should log that it forwards the request to slave->hostap_iface.

arnout (changes requested):
I will want to re-review.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: In one commit I sorted the list _and_ added my code, then I tried to split it. Failed. I know.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: > otherwise the handler wouldn't have been called

This assumptions comes from the knowledge of the developer, there is nothing in the code that supports it. We, as developers, know that other wise it wouldn't have been called, but reality shows that these are _exactly_ the places the code breaks. It happens for example when code starts to move around, reconstructed - then, these hidden assumptions about the state of the inputs are broken. 
In robust code each function is a _standalone_ unit, does not have any assumptions about the flow that led to it being called, it is self contained and validates its inputs before working with them.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: The only reason I had in mind is to keep the "most likely" first.
I changed it to a positive statement.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: [done](https://github.com/prplfoundation/prplMesh/issues/1306)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: I fill uncomfortable with this mechanism: when the function returns true the caller interprets it as "fine, no more processing is needed, don't forward the message" - which is _not_ what we want. 
I think this mechanism should be revisited as part of "handling error flows" correctly task.
See also: 
[1306](https://github.com/prplfoundation/prplMesh/issues/1306), 
[1085](https://github.com/prplfoundation/prplMesh/issues/1085),
[326](https://github.com/prplfoundation/prplMesh/issues/326), 
[325 - exit criteria 3](https://github.com/prplfoundation/prplMesh/issues/325)

controller/src/beerocks/master/son_master_thread.cpp
RanRegev: I prefer not to. The reason is this:
It is most likely that in some future, when error flows are handled, we would like to do _more_ than just printing the error. e.g. adding some counters, taking recovery actions, etc.
Current code construction supports better these future changes. It would be easier for the developer to construct a code that behaves differently upon success and failure.
(there is always the option that the entire section will be rewritten. Nevertheless, current schema better reveals the options).

agent/src/beerocks/slave/son_slave_thread.cpp
RanRegev: yes. indeed.

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
RanRegev: [see here](https://github.com/prplfoundation/prplMesh/pull/1273#discussion_r423278282)

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: [take a look](https://github.com/prplfoundation/prplMesh/pull/1275#discussion_r425008369)

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: It is quite funny: the [LOG(TRACE)](https://github.com/prplfoundation/prplMesh/blob/master/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp#L712) should have a `__func__`.
But [LOG(DEBUG)](https://github.com/prplfoundation/prplMesh/blob/master/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp#L131) does not.
So I guess I copied from `LOG(TRACE)` and change to `DEBUG`.
Anyway, I fixed.

tests/test_flows.py
RanRegev: ...yes, this is an [issue](https://github.com/prplfoundation/prplMesh/issues/1308) :-)

agent/src/beerocks/slave/son_slave_thread.cpp
RanRegev: This is reoccurring pattern ([1](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/son_slave_thread.cpp#L490), [2](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/son_slave_thread.cpp#L513), [3](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/son_slave_thread.cpp#L544), ...)  in this file.
So I prefer either fix them all or none (decided none).

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.h
RanRegev: did it, [However...](https://github.com/prplfoundation/prplMesh/issues/1309)

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
RanRegev: I don't understand the comment.
The reason I am referring to the specification is because this is the _sole_ reason the code is written the way it is. the idea is to save the burden from a future developer that will try to figure out why this code was written this way, and what does this error log mean.

tests/test_flows.py
RanRegev: You did the "investigation is needed" part of #1272 
When I work I see "things" along the way, things that I know a second look / investigation is needed for but are not in the main focus at the time they were spotted. It happens all the time and it is inevitable. I try not to drop them. I describe them the best I can, spending less time then it would take to deeply understand or solve the issue and continue.
This is one of those cases.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: Actually, [get_sta_radio](https://github.com/prplfoundation/prplMesh/blob/master/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h#L369)  is a private member function that seems to do the same thing. I removed mine completely and going to test with this function.
Also I [really think](https://github.com/prplfoundation/prplMesh/issues/1305) we should continue a step farther for simplifying the work with this data structure.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: In this specific case, yes, it seems that the log may be removed.
As for the general case: [I happened](https://github.com/CppCon/CppCon2019/tree/master/Posters/yet_another_fast_log) to deal with logs in the past. So I _have_ a strong opinion on the matter ;-).
If I need to summarize it in few sentences it would be:
Like any other aspect of the supporting environment for the code, logs must be useful. For a log to be useful it needs to tell a story, it needs to guide you through the events happened in the system. You should have a background in the way the system is built but you shouldn't have a deep knowledge in the code itself. For example, [tech support](https://www.bmc.com/blogs/support-levels-level-1-level-2-level-3/) should be able to monitor the system using the logs after a short while working with it, yet without being able to read the code itself.
It is not easy, and it is a continues mission to keep the log file useful, clean, without repeated messages, searchable - useful. Too often, however, projects neglect the log though it sometimes the only tool for monitoring and debugging.
I have never seen a real discussion (except for those I initiated), that sets customers, targets, guidelines, etc. regarding the logs. Although, in each project, everyone has an opinion and everyone uses the logs.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
arnout: After discussion in zoom: let's just keep the structure as it is and use it as a length, and only change the comment in the yaml file so it says length instead of boolean.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: I am aware to the `"," +`  option, I don't like it, doesn't come natural to me.
As you said - habit.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: The mechanism here is a bit different:
I am building the string to be placed _after_ `ap_ch_report=` and then I apply this value to `cmd` together with the `ap_ch_report=` prefix.
The structure is: `cmd += "ap_ch_report="` + `<built list>`
So the only question left is if to build the list using `stringstream operator<<` or `string operator +`. I find the former easier.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: > Potentially it has nothing to do [...]

I guess you're too young to have been washed with the Extreme Programming principles of [Do The Simplest Thing That Could Possibly Work](http://wiki.c2.com/?DoTheSimplestThingThatCouldPossiblyWork) :-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: > This assumptions comes from the knowledge of the developer, there is nothing in the code that supports it.

However, this check is making assumptions that don't exist. This function doesn't care what the type of the CMDU is that gets passed to it. It only cares about the TLVs that are in the CMDU. The caller has decided that this particular handler is OK for the CMDU. By adding the check here, you're not checking an invariant, but you're rather forcing the caller to be one specific caller. If you really want to force that, don't make a function of it...

To make it concrete: maybe a future update will define a different CMDU type that can be handled in the same way. Or maybe we will define a vendor-specific message that can be handled the same way.

Bottom line: I agree with Vitaly, this check is not only redundant, it is wrong.

That said, I don't think it's important enough to bother with removing it.

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
arnout: Hm, I didn't word that very well... What I mean is: if we get invalid values here, it's not a programming error, but it's invalid input. Referring to the specification in a comment makes complete sense, but not so much in the log. The log is going to be looked at by a system administrator, not by a prplmesh programmer. Hm... the more I'm arguing about this, the more it looks like I'm wrong - it *is* actually useful to refer to the spec so the sysadmin can debug the device that was sending this invalid message...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: Ok, I think I understand the confusion:
I _check_ for the correct tlv because I _expect_ it and _use_ it later in the code.
when the expected tlv is not found, I do the _best I can_ to give the user a hint about what I found, which is the message's type.

In general about the structure of this code:
We create concrete types of messages based on IDs (kind of a factory: in:ID, out:object)
These IDs _may_ (and will) conflict with the concrete types in certain scenarios.
The root of the complexity is the factory itself - there is no need to create concrete types based on IDs - and the logic is as simple as: at development time, if you know the ID - you know the concrete type (otherwise the factory wouldn't know what to construct). Therefore the factory itself is just a redundant, complex, unnecessary phase in the way.

Our code dresses and strips tlv object all the time out of their IDs.

tests/test_flows.py
RanRegev: Problem solved, see #1272 for details.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
RanRegev: Done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3266,6 +3287,80 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+        std::stringstream errorSS;
+        auto error_tlv = cmdu_tx.getClass<wfa_map::tlvErrorCode>();
+        if (error_tlv) {
+            errorSS << "0x" << error_tlv->reason_code();
+        } else {
+            errorSS << "note: error constructing the error itself";
+        }
+
+        LOG(DEBUG) << "sending ACK message to the originator with an error, mid: " << std::hex
+                   << int(mid) << " tlv error code: " << errorSS.str();
```
arnout, line 3344: Why do you insist on making simple things complicated?
```suggestion
        LOG(DEBUG) << "sending ACK message to the originator with an error, mid: " << std::hex
                   << int(mid) << " tlv error code: STA_NOT_ASSOCIATED_WITH_ANY_BSS_OPERATED_BY_THE_AGENT";
```

But even that is redundant - you could just add to the log message at the beginning of this condition that you're going to send an ack with error code. Or do you really want a log message for every line of code?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3266,6 +3287,80 @@ bool backhaul_manager::handle_1905_autoconfiguration_response(ieee1905_1::CmduMe
...
+
+        // send the error
+        return send_cmdu_to_bus(cmdu_tx, src_mac, bridge_info.mac);
+    }
+
+    forward_to = radio->slave;
```
arnout, line 3350: I would find it a lot easier to understand what happens if you simply do the forward_cmdu_to_bus here like we do in a few other places. However, all this code is going away with the unified agent, so don't bother.

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <mapf/common/utils.h>
+
+namespace beerocks {
+namespace gate {
+
+bool load(std::shared_ptr<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST> lhs,
+          const ieee1905_1::CmduMessageRx &rhs)
+{
+    // escape I
+    if (!lhs) {
```
arnout, line 13: If it must not be null, why is it a pointer to begin with?

agent/src/beerocks/slave/gate/1905_beacon_query_to_vs.cpp
```diff
@@ -0,0 +1,115 @@
...
+    if (ap_ch_report_length != 0 && lhs_params.channel != 255) {
+        LOG(ERROR) << "inconsistency between channel report length and channel number. please take "
+                      "look at the specification. v1 17.2.27";
+        return false;
+    }
+
+    // copy m_ap_channel_reports_list from the 1905 beacon_metrics_query
+    // into ap_ch_report char array of the vs message (lhs_params)
+    if (ap_ch_report_length > 1) {
+        // the reason we support only 1 channel list is that the hostapd does not support more
```
arnout, line 70: Actually it's the VS message that doesn't support more. And Intel's hacked-up hostapd (AKA dwpal). Upstream hostapd supports it (because you actually have to give it the complete beacon request frame).

tests/test_flows.py
```diff
@@ -958,6 +958,34 @@ def test_topology(self):
...
+        sta = env.Station.create()
+
+        # for testing non existing STA, when want to test the error flow
+        # sta1 = env.Station.create()
+
+        debug("Connect dummy STA (" + sta.mac + ") to wlan0")
+        env.agents[0].radios[0].vaps[0].associate(sta)
+
+        # send beacon query request
+        # (please take a look at https://github.com/prplfoundation/prplMesh/issues/1272)
```
arnout, line 1051: This is no longer relevant, right?

tests/test_flows.py
```diff
@@ -958,6 +958,34 @@ def test_topology(self):
...
+        request = '{mac} '.format(mac=sta.mac)
+        request += '0x73 0xFF 0xFFFFFFFFFFFF 0x02 0x00 0x01 0x02 0x73 0x24 0x30 0x00'
+
+        debug(request)
+        mid = env.controller.dev_send_1905(env.agents[0].mac, 0x8011,
+                                           tlv(0x99, 0x0016, "{" + request + "}"))
+
+        self.check_log(env.agents[0],
+                       r"BEACON METRICS QUERY: "
+                       r"sending ACK message to the originator mid: {:d}".format(mid))
```
arnout, line 1062: Please don't add this kind of log checks any more. Instead, check for the ACK in the sniffer.

arnout (approved):
There are still a lot of things that can be improved in commit ordering, splitting and commit messages. But since this PR has been lying around for more than a month already, let's just get it merged.

For the future, whenever you do some small improvement or refactoring, do it on a fresh branch from master and make a PR of it, then rebase on top of that.

arnout:
@mergifyio rebase

Issue #1261, opened 2020-05-10
Complete Beacon Response


RanRegev:
## Description

The 1905 Beacon Metric Query is answered by Beacon Metrics Response.
The details of the response returned from the station should be filled by the agent and propagate to the controller.

## Exit criteria

All data from the STA is stored correctly at the controller internal data (database).

## Detailed description

[tlvBeaconMetricsResponse.yaml](https://github.com/prplfoundation/prplMesh/blob/master/framework/tlvf/yaml/tlvf/wfa_map/tlvBeaconMetricsResponse.yaml) contains only a sketch of the response - a vector of uint8_t

**Problem statement**
The entire flow is useless in its current state, as the data does not reach the controloler.

**Possible solutions**
Implement according to WFA specification

PR #1274, opened 2020-05-11
Feature/certification 4.7.9 iii response

RanRegev:
This commit adds the handling of the beacon metrics response.
A.K.A certification 4.7.9
It is a complementary PR to #1273 PR.
And it is built on top of it.

### DIFF
--- a/agent/src/beerocks/slave/gate/vs_beacon_response_to_1905.cpp
+++ b/agent/src/beerocks/slave/gate/vs_beacon_response_to_1905.cpp
@@ -1,0 +1,64 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "vs_beacon_response_to_1905.h"
+#include <tlvf/wfa_map/tlvBeaconMetricsResponse.h>
+
+namespace beerocks {
+namespace gate {
+
+/*
+typedef struct sBeaconResponse11k {
+    uint8_t channel;
+    uint8_t op_class;
+    uint8_t dialog_token;
+    uint8_t measurement_token;
+    uint8_t rep_mode;
+    uint8_t phy_type;
+    uint8_t frame_type;
+    int8_t rcpi;
+    uint8_t rsni;
+    uint8_t ant_id;
+    uint16_t duration;
+    uint32_t parent_tsf;
+    uint64_t start_time;
+    sMacAddr sta_mac;
+    sMacAddr bssid;
+    uint32_t new_ch_width;
+    uint32_t new_ch_center_freq_seg_0;
+    uint32_t new_ch_center_freq_seg_1;
+    uint8_t use_optional_wide_band_ch_switch;
+    }
+*/
+
+bool load(ieee1905_1::CmduMessageTx &lhs,
+          std::shared_ptr<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_RESPONSE> rhs)
+{
+    // escape I
+    if (!rhs) {
+        LOG(ERROR) << "nullptr vs pointer. exiting";
+        return false;
+    }
+
+    // escape II
+    auto tlv = lhs.getClass<wfa_map::tlvBeaconMetricsResponse>();
+    if (!tlv) {
+        LOG(ERROR)
+            << "could not find wfa_map::tlvBeaconMetricsResponse within the input tx. exiting";
+        return false;
+    }
+
+    const auto &beacon_response_11k = rhs->params();
+
+    tlv->associated_sta_mac() = beacon_response_11k.sta_mac;
+
+    return true;
+}
+
+} // namespace gate
+} // namespace beerocks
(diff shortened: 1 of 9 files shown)


tomereli (commented):
Its a 35 commits PR where most are fixups, please organize this PR before asking for review

arnout:
Note that the first 15 commits are from #1273, so only the last 5 "count".

RanRegev:
> Note that the first 15 commits are from #1273, so only the last 5 "count".

I used to work in a place where each commit was preceded with `issue number # <commit message>`  e.g. `prplCommon-476 # <message>`
It had few advantages:
* It was extremely visible later on for which issue this change was made for
* It _forces_ each change to be backed up by a valid, existing, well document issue.
* It created a nice grep-able history
* It made us all name our branches after the name of the issue.

Now that we are about to switch to Jira, maybe we should consider doing it here too.
For me and my team it worked great.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -2686,14 +2688,19 @@ bool slave_thread::handle_cmdu_monitor_message(Socket *sd,
...
+        // flow: extract data from response_in (vendor specific response) and build
+        // with the extracted data 1905 reponse_out message
         auto response_in =
             beerocks_header
                 ->addClass<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_RESPONSE>();
         if (response_in == nullptr) {
             LOG(ERROR) << "addClass ACTION_MONITOR_CLIENT_BEACON_11K_RESPONSE failed";
             break;
         }
+        /* old vs response:
```
arnout: This is not OK. As long as the controller is not updated to handle the new response, and the response of course contains all the required information, we have to keep the VS response.

But it's OK to send both messages from this function.

agent/src/beerocks/slave/gate/vs_beacon_response_to_1905.cpp
```diff
@@ -0,0 +1,64 @@
...
+        return false;
+    }
+
+    // escape II
+    auto tlv = lhs.getClass<wfa_map::tlvBeaconMetricsResponse>();
+    if (!tlv) {
+        LOG(ERROR)
+            << "could not find wfa_map::tlvBeaconMetricsResponse within the input tx. exiting";
+        return false;
+    }
```
arnout, line 54: IMO it makes no sense to create the TLV in the caller and then check here that the cmdu_tx already contains it. Simple use addClass here and remove it from the caller.

Actually, getClass should be removed entirely from CmduMessageTx.

agent/src/beerocks/slave/gate/vs_beacon_response_to_1905.cpp
```diff
@@ -0,0 +1,64 @@
...
+    sMacAddr bssid;
+    uint32_t new_ch_width;
+    uint32_t new_ch_center_freq_seg_0;
+    uint32_t new_ch_center_freq_seg_1;
+    uint8_t use_optional_wide_band_ch_switch;
+    }
+*/
+
+bool load(ieee1905_1::CmduMessageTx &lhs,
+          std::shared_ptr<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_RESPONSE> rhs)
```
arnout, line 40: Since this must be non-nullptr, it should be a reference instead of a pointer.

Also, it should be const.

arnout (approved):
Please make sure the last commit in the series has something like this in the commit message:

```
With this, certification test 4.7.9 should pass:
MAP-4.7.9_ETH:netgear-rax40
```

Also this is affecting the existing functionality, so someone in @kantera800's team should validate client steering with 11k on an actual platform.

agent/src/beerocks/slave/son_slave_thread.cpp
RanRegev: [1261](https://github.com/prplfoundation/prplMesh/issues/1261) already stated the problem.
Should we "solve" it meanwhile by sending two responses - one to pass certification and the second with the usable data?

agent/src/beerocks/slave/gate/vs_beacon_response_to_1905.cpp
RanRegev: As I see it, this function checks its inputs, verify that it got everything it needs and then executes. It is careless about the caller. Just input verification.
The way to remove verification is to remove the _need_ for it. e.g. require a reference instead of a not-null pointer, or requesting a concrete object and not some wrapper object that might contain the desired type inside it. In this specific example: `wfa_map::tlvBeaconMetricsResponse` and not a `CmduMessageTx` that holds (or not) the concrete `tlvBeaconMetricsResponse` inside it.

agent/src/beerocks/slave/gate/vs_beacon_response_to_1905.cpp
RanRegev: Agree (and you stole my comment :-))
I don't know why eventually I ended up with this interface, it should have been a reference.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: That's exactly what we have to do (not should, *have to*) if you want to get this PR merged. Breaking existing features is not acceptable.

VladyslavTupikin (approved):
Do we have a UML diagram and .md file for 4.7.9 feature?

kantera800 (approved):
when over teh code briefly.
tested on UGW  and it works fine so approving.

DEBUG 12:13:40:516 <2012233024> son_management.cpp[493] --> CLI beacon request for c4:0b:cb:ea:79:46 to 00:e0:92:00:01:47
DEBUG 12:13:40:839 <2012233024> son_master_thread.cpp[3315] --> beacon response , ID: 0
sta_mac: c4:0b:cb:ea:79:46
measurement_rep_mode: 0
op_class: 5
channel: 1
duration: 49
rcpi: -13
rsni: 0
bssid: 00:50:f1:85:dc:b5

arnout:
@mergifyio rebase

arnout:
@mergifyio rebase

PR #1275, opened 2020-05-11
tests: added two unit test applications

RanRegev:
This PR adds two test targets that check
the conversions between vs and 1905
for beacon metrics query and response.
it is used during development and can also
be used during regualr builds as a test.
currently the DUT, the conversion cpp function
implementation is compiled twice:
* as part of the son-slave
* as part of the test application
The best would be to create a library for conversion.

### DIFF
--- a/agent/src/beerocks/slave/CMakeLists.txt
+++ b/agent/src/beerocks/slave/CMakeLists.txt
@@ -35,3 +35,34 @@ target_link_libraries(${PROJECT_NAME} rt dl bcl btl tlvf elpp btlvf bwl bpl ${LI
 
 # Install
 install(TARGETS ${PROJECT_NAME} DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+# Tests
+if (BUILD_TESTS)
+    set(TEST_PROJECT_NAME ${PROJECT_NAME}_unit_tests)
+    set(unit_tests_sources
+        gate/unit_tests/gate_test.cpp
+        ${MODULE_PATH}/gate/1905_beacon_query_to_vs.cpp 
+        ${MODULE_PATH}/gate/vs_beacon_response_to_1905.cpp
+    )
+    add_executable(${TEST_PROJECT_NAME}
+        ${unit_tests_sources}
+    )
+    if (COVERAGE)
+        set_target_properties(${TEST_PROJECT_NAME} PROPERTIES COMPILE_FLAGS "--coverage -fPIC -O0")
+        set_target_properties(${TEST_PROJECT_NAME} PROPERTIES LINK_FLAGS "--coverage")
+    endif()
+    target_include_directories(${TEST_PROJECT_NAME}
+        PRIVATE
+            ${PLATFORM_INCLUDE_DIR}
+            ${CMAKE_CURRENT_SOURCE_DIR}/gate
+        PUBLIC
+            $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    )
+    target_link_libraries(${TEST_PROJECT_NAME} btlvf tlvf mapfcommon)
+    target_link_libraries(${TEST_PROJECT_NAME} gtest_main)
+    install(TARGETS ${TEST_PROJECT_NAME} DESTINATION bin/tests)
+    add_test(NAME ${TEST_PROJECT_NAME} COMMAND $<TARGET_FILE:${TEST_PROJECT_NAME}>)
+endif()
+
+
+
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -198,6 +198,9 @@ bwl_dummy_unit_tests:
 bcl_unit_tests:
   extends: .run-test-in-docker
 
+beerocks_agent_unit_tests:
+  extends: .run-test-in-docker
+
 run-tests:
   stage: test
   image: $CI_REGISTRY_IMAGE/prplmesh-tests-runner:$CI_PIPELINE_ID
--- a/agent/src/beerocks/slave/gate/unit_tests/gate_test.cpp
+++ b/agent/src/beerocks/slave/gate/unit_tests/gate_test.cpp
@@ -1,0 +1,191 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "1905_beacon_query_to_vs.h"
+#include "vs_beacon_response_to_1905.h"
+#include <bcl/network/network_utils.h>
+#include <tlvf/CmduMessageTx.h>
+#include <tlvf/wfa_map/tlvBeaconMetricsResponse.h>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+/////// 1905 --> vs   ////////////////////////////////////////////////////////////////////////////////////
+
+/**
+* const values to be used for both assignment and test
+*/
+
+const auto associated_sta_mac_1 = sMacAddr{0x01, 0x02, 0x03, 0x04, 0x05, 0x06};
+const uint8_t operating_class   = 0x73;
+const uint8_t channel_number    = 255;
+const uint8_t channel_1         = 46;
+const uint8_t channel_2         = 31;
+
+/**
+* helper function to fill 1905 values
+*/
+static bool set_1905_values(ieee1905_1::CmduMessageTx &cmdu1905)
+{
+    auto beacon_metrics_query = cmdu1905.getClass<wfa_map::tlvBeaconMetricsQuery>();
+    if (!beacon_metrics_query) {
+        return false;
+    }
+
+    // fill some values
+    beacon_metrics_query->associated_sta_mac() = associated_sta_mac_1;
+    beacon_metrics_query->operating_class()    = operating_class;
+    // for ap channel report list, this must be set to 255
+    beacon_metrics_query->channel_number() = channel_number;
+
+    auto ap_channel_list = beacon_metrics_query->create_ap_channel_reports_list();
+    if (!ap_channel_list) {
+        return false;
+    }
+
+    ap_channel_list->alloc_ap_channel_report_list(3);
+    // the first is operating class, which is currently ignored system wide
+    *ap_channel_list->ap_channel_report_list(0) = operating_class;
+    *ap_channel_list->ap_channel_report_list(1) = channel_1;
+    *ap_channel_list->ap_channel_report_list(2) = channel_2;
+
+    beacon_metrics_query->add_ap_channel_reports_list(ap_channel_list);
+
+    return true;
+}
+
+TEST(gate_beacon_query_test, loading_vs_from_1905_beacon_query)
+{
+    // preperation for beerocks
+
+    // class list
+
+    constexpr uint16_t buf_size_beerocks = 10000; // just an arbitrary size
+    uint8_t buff_beerocks[buf_size_beerocks];
+    ClassList cl(buff_beerocks, buf_size_beerocks);
+
+    // beerocks
+    auto added_class_vs =
+        cl.addClass<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST>();
+
+    EXPECT_NE(added_class_vs, nullptr);
+
+    auto beerocks = cl.getClass<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST>();
+
+    EXPECT_NE(beerocks, nullptr);
+
+    // build 1905 message whthin a Tx message
+
+    // construct a Tx message with beacon metrics query
+    constexpr uint16_t buf_size_1905_tx = 10000; // just an arbitrary size
+    uint8_t buff_1905_tx[buf_size_1905_tx];
+    ieee1905_1::CmduMessageTx cmdu_1905(buff_1905_tx, buf_size_1905_tx);
+
+    auto metrics_query =
+        cmdu_1905.create(0x1, ieee1905_1::eMessageType::BEACON_METRICS_QUERY_MESSAGE);
+    EXPECT_NE(metrics_query, nullptr);
+
+    auto added_class_1905 = cmdu_1905.addClass<wfa_map::tlvBeaconMetricsQuery>();
+    EXPECT_NE(added_class_1905, nullptr);
+
+    // fill 1905 message with values
+    bool filled = set_1905_values(cmdu_1905);
+    EXPECT_TRUE(filled);
+
+    // Rx - built from the tx message above
+    ieee1905_1::CmduMessageRx rx(cmdu_1905.getMessageBuff(), cmdu_1905.getMessageBuffLength());
+
+    bool parse_ok = rx.parse();
+    EXPECT_EQ(parse_ok, true);
+
+    auto tlv_becon_metrics_query = rx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+    EXPECT_NE(tlv_becon_metrics_query, nullptr);
+
+    // DUT -
+    bool load_ok = beerocks::gate::load(/*dst*/ beerocks, /*src*/ rx);
+    EXPECT_EQ(load_ok, true);
+
+    auto &beerocks_params = beerocks->params();
+
+    // variables that were set in the test
+    EXPECT_EQ(beerocks_params.channel, channel_number);
+    EXPECT_EQ(beerocks_params.sta_mac, associated_sta_mac_1);
+    EXPECT_EQ(beerocks_params.op_class, operating_class);
+
+    // default values
+    EXPECT_EQ(beerocks_params.bssid, beerocks::net::network_utils::ZERO_MAC);
+    EXPECT_EQ(beerocks_params.measurement_mode, beerocks::MEASURE_MODE_ACTIVE);
+    EXPECT_EQ(beerocks_params.duration, beerocks::BEACON_MEASURE_DEFAULT_ACTIVE_DURATION);
+    EXPECT_EQ(beerocks_params.expected_reports_count, 1);
+    EXPECT_EQ(beerocks_params.rand_ival, beerocks::BEACON_MEASURE_DEFAULT_RANDOMIZATION_INTERVAL);
+    EXPECT_EQ(beerocks_params.report, 0);
+    EXPECT_EQ(beerocks_params.repeats, 0);
+    EXPECT_EQ(beerocks_params.parallel, 0);
+    EXPECT_EQ(beerocks_params.enable, 0);
+    EXPECT_EQ(beerocks_params.request, 1);
+    EXPECT_EQ(beerocks_params.mandatory_duration, 0);
+    EXPECT_EQ(beerocks_params.use_optional_ssid, 0);
+}
+
+/////// vs --> 1905 ////////////////////////////////////////////////////////////////////////////////////
+
+/**
+* const values to be used for both assignment and test
+*/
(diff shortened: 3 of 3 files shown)


arnout:
You should also add it to CI. Should be sufficient to add the following to .gitlab-ci.yml:
```
1905_beacon_query_to_vs_test:
    extends: .run-test-in-docker
```

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
+
+#include "../1905_beacon_query_to_vs.h"
+#include <bcl/network/network_utils.h>
+#include <tlvf/CmduMessageTx.h>
+
+constexpr const int RET_OK          = 0;
```
mariomaz: `constexpr` implies `const`, so no need to use both. `constexpr` is enough and preferred.
According to naming convention, the name of the variable should use lower_snake_case (UPPER_SNAKE_CASE is for #define)

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+constexpr const int RET_NO_BEEROCKS = -1;
+constexpr const int RET_NO_1905     = -2;
+
+std::ostream &operator<<(std::ostream &o, const beerocks_message::sBeaconRequest11k &request)
+{
+    // NOTE: not all fields are printed
+    // TODO: complete all fields
```
mariomaz: (nitpick) 
When I add a TODO, there's always someone who asks me in the PR comments to create an issue and add the issue number here.

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+      << "enable:                   " << +request.enable << "\n"
+      << "request:                  " << +request.request << "\n"
+      << "report:                   " << +request.report << "\n"
+      << "mandatory_duration:       " << +request.mandatory_duration << "\n"
+      << "expected_reports_count:   " << +request.expected_reports_count << "\n"
+      << "use_optional_ssid:        " << +request.use_optional_ssid << "\n"
+      << "ssid:                     " << request.ssid << "\n"
+      << "use_optional_ap_ch_report:" << +request.use_optional_ap_ch_report << "\n";
+
+    for (int idx = 0; idx < request.use_optional_ap_ch_report; ++idx) {
```
mariomaz: (nitpick) 
Use size_t instead of int

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+        o << "channel[" << idx << "]:    " << std::dec << +request.ap_ch_report[idx] << "\n";
+    }
+
+    return o;
```
mariomaz: I'd add a std::flush to ostream before returning

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+constexpr const int RET_NO_BEEROCKS = -1;
+constexpr const int RET_NO_1905     = -2;
+
+std::ostream &operator<<(std::ostream &o, const beerocks_message::sBeaconRequest11k &request)
+{
+    // NOTE: not all fields are printed
+    // TODO: complete all fields
+
+    o << "\n"
+      << "measurement_mode:         " << +request.measurement_mode << "\n"
```
mariomaz: I think the '+' sign before the variable is a typo.
Be aware it also appears in some other lines below

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+    return o;
+}
+
+std::ostream &operator<<(std::ostream &o,
+                         beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST &msg)
+{
+    // currently we don't print anything except for the params field
+    // we might want to print / verify the action op (currently hard coded in the class)
+    o << "ACTION_MONITOR_CLIENT_BEACON_11K_REQUEST:"
+      << "\n-----------------------------------------\n"
```
mariomaz: Some static analyzers would complain because you are using the same literal string more than once.

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+      << msg.params() << "\n-----------------------------------------\n";
+
+    return o;
+}
+
+void fillTxValues(ieee1905_1::CmduMessageTx &tx)
```
mariomaz: Class members and functions shall use lower_snake_case.

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+    // escape I
+    auto tlv = tx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+    if (!tlv) {
+        LOG(ERROR) << "this Tx message does not contain tlv beacon metrics query";
```
mariomaz: You'd probably want to return in case of error

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    // fill some values
+    tlv->associated_sta_mac() = sMacAddr{0x01, 0x02, 0x03, 0x04, 0x05, 0x06};
+    tlv->operating_class()    = 0x73;
```
mariomaz: Some static analyzers would complain because you are using magic numbers here

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    tlv->add_ap_channel_reports_list(ap_channel_list);
+}
+
+int main()
+{
+    // preperation for beerocks
+
+    // class list
+    constexpr const uint16_t BUF_SIZE_BEEROCKS = 10000;
```
mariomaz: constexpr implies const.
Use lower_snake_case for the variable name

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+        return RET_NO_BEEROCKS;
+    }
+
+    // build Rx message from Tx message
+    //
+    // construct a Tx message with beacon metrics query
+    constexpr const uint16_t BUF_SIZE_1905_TX = 10000;
+    uint8_t buff_1905_tx[BUF_SIZE_1905_TX];
+    ieee1905_1::CmduMessageTx tx(buff_1905_tx, BUF_SIZE_1905_TX);
+    tx.create(0x1, ieee1905_1::eMessageType::BEACON_METRICS_QUERY_MESSAGE);
```
mariomaz: Check returned value

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+    rx.parse();
+    auto tlvRx = rx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+
+    if (!tlvRx) {
+        LOG(ERROR) << "no 1905 rx";
+        return RET_NO_1905;
+    }
+
+    // DUT -
+    gate::load(/*dst*/ beerocks, /*src*/ rx);
```
mariomaz: I think `gate` and `load` are not meaningful names as they don't reflect what the method is really doing

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    // verification (hunman readable output_
+    beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST &msg = *beerocks;
+    std::cout << msg;
```
mariomaz: Unit tests should be designed to run without interaction, no human should be required to decide if tests PASS or FAIL by inspecting console output.
A better approach is to use assertions which automatically indicate test results.

mariomaz (changes requested):
Use assertions instead of print statements to verify the code executed correctly.

Regarding dual compilation, it's ok if code under test has to be compiled twice. You cannot even avoid it if you create a library. Most of the times is simply because your target platform is not the same as your testing platform. But even if they were, reason is compilation flags must be different for production binary than for unit tests. For example, you might want to use some optimization flags to create production binary that you have to avoid if you want to generate an accurate code coverage report. Also you need special compilation flags to generate code coverage information that you should skip in production binary.

RanRegev:
> Use assertions instead of print statements to verify the code executed correctly.
> 
> Regarding dual compilation, it's ok if code under test has to be compiled twice. You cannot even avoid it if you create a library. Most of the times is simply because your target platform is not the same as your testing platform. But even if they were, reason is compilation flags must be different for production binary than for unit tests. For example, you might want to use some optimization flags to create production binary that you have to avoid if you want to generate an accurate code coverage report. Also you need special compilation flags to generate code coverage information that you should skip in production binary.

I think that tests that are changing the DUT to their need are only partial tests.
If you are compiling _differently_ for production and for testing then something will eventually slip away. As for different targets, yes, you should compile differently for each target - this is inherent in the word "target". Library _does_ solve the problem as you compile the library (for each target) once and link against it with your _code_ and with your _test_.

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
RanRegev: Will fix.
And yes, you are correct: http://eel.is/c++draft/dcl.spec#dcl.constexpr-10

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
RanRegev: good advice

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
RanRegev: ok, will do. I'll use std::endl instead of "\n":
https://en.cppreference.com/w/cpp/io/manip/endl
http://eel.is/c++draft/ostream.manip#2

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
RanRegev: Take a look at ["Method D"](https://bytefreaks.net/programming-2/c/c-how-to-print-an-unsigned-character-unsigned-byte-uint8_t-using-cout)
and also at [unary _ operator](http://eel.is/c++draft/expr.unary.op#7), [conv.prom](http://eel.is/c++draft/conv.prom#1), [conv.rank](http://eel.is/c++draft/conv.rank#1.3) and finally on [cout printing arithmetics](http://eel.is/c++draft/ostream.inserters.arithmetic).

In words: the type of `unary + operator` on `unsigned char` is `int`. This is because the `unsigned char` was promoted to `int`. The promotion to int is because its `integer conversion rank` is less than of `int`.

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
RanRegev: [take a look](https://github.com/prplfoundation/prplMesh/pull/1273#discussion_r424333678)

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
RanRegev: ok

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
mariomaz: TIL

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
mariomaz: I see I'm not the only one who thinks it's not a good name, then it's a smell you should think it twice.
Some people say that comments in code should not be necessary if code was clear enough.
What name will you choose when creating a gate between other two worlds?

agent/src/beerocks/slave/test/1905_beacon_query_to_vs_test.cpp
arnout: Nice! I should have read this yesterday, then I had a learning to report on the sprint meeting!

agent/src/beerocks/slave/CMakeLists.txt
```diff
@@ -34,3 +34,33 @@ target_link_libraries(${PROJECT_NAME} rt dl bcl btl tlvf elpp btlvf bwl bpl ${LI
 
 # Install
 install(TARGETS ${PROJECT_NAME} DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+# Tests
+if (BUILD_TESTS)
+    set(TEST_PROJECT_NAME ${PROJECT_NAME}_unit_tests)
+    set(unit_tests_sources
+        gate/unit_test/gate_test.cpp
```
mariomaz: Directory name for unit tests in other projects has been named `unit_tests`

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+    return true;
+}
+
+TEST(gate_beacon_query_test, loading_vs_specific_from_1905_beacon_query)
+{
+    // preperation for beerocks
+
+    // class list
+
+    constexpr const uint16_t BUF_SIZE_BEEROCKS = 10000; // just an arbitrary size
```
mariomaz: No need for `const` after `constexpr`
UPPER_CASE is for #define's

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    ASSERT_NE(beerocks, nullptr);
+
+    // build 1905 message whthin a Tx message
+
+    // construct a Tx message with beacon metrics query
+    constexpr const uint16_t BUF_SIZE_1905_TX = 10000; // just an arbitrary size
+    uint8_t buff_1905_tx[BUF_SIZE_1905_TX];
+    ieee1905_1::CmduMessageTx cmdu_1905(buff_1905_tx, BUF_SIZE_1905_TX);
+    cmdu_1905.create(0x1, ieee1905_1::eMessageType::BEACON_METRICS_QUERY_MESSAGE);
```
mariomaz: Please check return value

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+    cmdu_1905.addClass<wfa_map::tlvBeaconMetricsQuery>();
+
+    // fill 1905 message with values
+    bool filled = set_1905_values(cmdu_1905);
+    ASSERT_TRUE(filled);
+
+    // Rx - built from the tx message above
+    ieee1905_1::CmduMessageRx rx(cmdu_1905.getMessageBuff(), cmdu_1905.getMessageBuffLength());
+    rx.parse();
```
mariomaz: Please check return value

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+    auto tlv_becon_metrics_query = rx.getClass<wfa_map::tlvBeaconMetricsQuery>();
+
+    ASSERT_NE(tlv_becon_metrics_query, nullptr);
+
+    // DUT -
+    beerocks::gate::load(/*dst*/ beerocks, /*src*/ rx);
```
mariomaz: Have I ever mentioned that `gate` and `load` are not appropriate names ;-)
You had to add comments in the parameters to remember what is being loaded into what

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+    auto &beerocks_params = beerocks->params();
+
+    // set
+    ASSERT_EQ(beerocks_params.channel, channel_number);
```
mariomaz: Please consider using EXPECT_EQ instead of ASSERT_EQ
See [Assertions](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#assertions)

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
```diff
@@ -0,0 +1,121 @@
...
+    // preperation for beerocks
+
+    // class list
+
+    constexpr const uint16_t BUF_SIZE_BEEROCKS = 10000; // just an arbitrary size
+    uint8_t buff_beerocks[BUF_SIZE_BEEROCKS];
+    ClassList cl(buff_beerocks, BUF_SIZE_BEEROCKS);
+
+    // beerocks
+    cl.addClass<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_REQUEST>();
```
mariomaz: Please check return value

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
```diff
@@ -118,4 +122,53 @@ TEST(gate_beacon_query_test, loading_vs_specific_from_1905_beacon_query)
...
+{
+    // preperation for beerocks
+
+    // class list
+    constexpr const uint16_t BUF_SIZE_BEEROCKS = 10000;
+    uint8_t buff_beerocks[BUF_SIZE_BEEROCKS];
+    ClassList cl(buff_beerocks, BUF_SIZE_BEEROCKS);
+
+    // beerocks
+    cl.addClass<beerocks_message::cACTION_MONITOR_CLIENT_BEACON_11K_RESPONSE>();
```
mariomaz: Same comments here than in previous test about checking returned values and constexpr + const

mariomaz (approved):
Some very simple to fix comments, so approving

agent/src/beerocks/slave/CMakeLists.txt
RanRegev: I checked other directories... missed that `s`.
Fixed.

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
RanRegev: I prefer `beerocks << rx;` or `rx >> beerocks;` but it leaves no space for future parameters if needed (like some context to do the translation): `beerocks::gate::load(beerocks, rx, context_to_support_translation)`

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
RanRegev: done.

agent/src/beerocks/slave/gate/unit_test/gate_test.cpp
RanRegev: done.

arnout (approved):
The unit test isn't executed in CI, so it's quite useless. Look at the bcl_unit_test in .gitlab-ci.yml for how it's done.

PR #1277, opened 2020-05-12
build.sh: set valid value of docker run option --userns

MuteSpirit:
Latest available Docker version  (19.03.8) on Ubuntu 18.04 from official
Docker APT repo does not support value "keep-id" for docker run option "--userns"

Set "--userns" option value to format acceptable by docker client:
    "keep-id" -> "host:USER_ID"

### DIFF
--- a/tools/docker/build.sh
+++ b/tools/docker/build.sh
@@ -17,7 +17,6 @@ main() {
     docker_args=(
         --workdir "${rootdir}"
         --user "${SUDO_UID:-$(id -u)}:${SUDO_GID:-$(id -g)}"
-        --userns keep-id
         -e "USER=${SUDO_USER:-${USER}}"
         -v "${rootdir}:${rootdir}"
         --entrypoint "./tools/maptools.py"


tomereli (approved):

tools/docker/build.sh
```diff
@@ -17,7 +17,7 @@ main() {
     docker_args=(
         --workdir "${rootdir}"
         --user "${SUDO_UID:-$(id -u)}:${SUDO_GID:-$(id -g)}"
-        --userns keep-id
+        --userns host:${SUDO_UID:-$(id -u)}
```
rmelotte:
```suggestion
        --userns "host:${SUDO_UID:-$(id -u)}"
```

rmelotte (approved):

arnout (changes requested):
It doesn't work.

Referring to the source code https://github.com/docker/docker-ce/blob/531930f3294c31db414f17f80fa8650d4ae66644/components/engine/api/types/container/host_config.go#L169 it seems that docker in that version only looks at the part before the colon anyway. And the meaning of "host" is exactly the way that it behaved before, which doesn't allow to do the build completely unprivileged.

So if it doesn't work, it's better to revert my commit completely and just remove the userns option.

MuteSpirit:
I'm agree with @arnout suggestion

tools/docker/build.sh
MuteSpirit: "--userns" option usage has been removed

arnout (approved):

rmelotte (approved):

PR #1278, opened 2020-05-12
tests: capi: Fix timeout issue on prplMesh server [HotFix]

CoralMalachi:
### Description

Currently, each one of the tests in test_flows.py causing a timeout error (in prplMesh server only)
This PR is aiming to resolve this issue (maybe temporarily)

### Changes

After exploring online, I found out that others faced the same issue and the solution was to set the timeout value with the `settimeout` method.
For that reason, I did the same and that solved the issue.

### DIFF
--- a/tests/capi.py
+++ b/tests/capi.py
@@ -90,7 +90,8 @@ class UCCSocket:
         self.timeout = timeout
 
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


arnout:
I think this is superseded by #1279, no? Please reopen if I'm wrong.

PR #1279, opened 2020-05-12
tests: capi.py: UCCSocket use longer timeout

tomereli:
On prplmesh-s1 (Intel) server, UCC connection now time outs when trying
to reach the controller/agents over CAPI.
One possible reason is that for some reason it now takes longer for the
controller / agents UCC listener to come up.
Fix it by increasing the default timeout in the UCCSocket class.

### DIFF
--- a/tests/capi.py
+++ b/tests/capi.py
@@ -72,7 +72,7 @@ class UCCSocket:
     CAPI commands from it.
     """
 
-    def __init__(self, host: str, port: int, timeout: int = 10):
+    def __init__(self, host: str, port: int, timeout: int = 30):
         """Constructor for UCCSocket
 
         Parameters


arnout (approved):
The "it now takes longer for the controller / agents UCC listener to come up." is not correct, because it also happens with --skip-init (i.e. when the agent has been running already for a long time).

That said, it doesn't hurt to increase the timeout, if that helps.

tomereli:
> The "it now takes longer for the controller / agents UCC listener to come up." is not correct, because it also happens with --skip-init (i.e. when the agent has been running already for a long time).
> 
> That said, it doesn't hurt to increase the timeout, if that helps.

I couldn't reproduce this with --skip-init when netcat works, meaning the ucc listenr is up and running.

vitalybu (approved):

PR #1282, opened 2020-05-12
Deploy the prplwrt firmwares in CI

### LINKED ISSUE #946: [TASK] Deploy prplWrt from CI
## Description

Now that the full firmwares are built in CI, we can use them to automatically update and deploy the new firmware when it's changed in this repository.

## Exit criteria

- [ ] Whenever the firmware is updated in this repository, it gets built again and deployed to every relevant device.

- [ ] If the device already has the same firmware, it's not updated to avoid reflashing every device all the time. This will require a way to check if the things we track in the Dockerfile have changed from what the device has.

- [ ] The settings we previously applied manually when setting up the device are now applied automatically when deploying the firmware

- [ ] The previously passing certification tests still pass after the device was updated automatically.

Note: the first attempt to use `sysupgrade` on the rax40 was not successful, so we might need to have a different procedure for some devices (using tftp for example) (or fix sysupgrade on the rax40).

(I'm only assigning myself to track it, feel free to take this issue)

rmelotte:
- Store the hashes we use in the Dockerfile, in the firmware itself.
- Add a python script to check if a device needs to be upgraded by comparing hashes, and do the upgrade with sysupgrade or u-boot depending on the target.
- Use the python script in CI.

I tested that the upgrade works, and that re-trying the upgrade with the same firmware version doesn't do anything (for both the "test-on-target" jobs, and the certification testbed).

Fixes https://github.com/prplfoundation/prplMesh/issues/946

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -223,6 +223,7 @@ run-tests:
     # kill prplmesh on each target to make sure they don't interfere with the test
     - for i in $ALL_TARGETS ; do ssh "$i" 'pgrep -f beerocks | xargs kill -9 2>/dev/null' || true ; done
   script:
+    - tools/deploy_firmware.py --device "$TARGET_DEVICE" --target-name "$TARGET_DEVICE_NAME" --image "$OPENWRT_IMAGE"
     - tools/deploy_ipk.sh $TARGET_DEVICE_NAME "build/$TARGET_DEVICE/"prplmesh.ipk
     - tests/openwrt/test_status.sh $TARGET_DEVICE_NAME
   artifacts:
@@ -232,7 +233,7 @@ run-tests:
     when: always
   tags:
     - targets
-  timeout: 5m
+  timeout: 20m
 
 build-for-turris-omnia:
   extends: .build-for-openwrt
@@ -254,6 +255,7 @@ test-on-turris-omnia:
   variables:
     TARGET_DEVICE: turris-omnia
     TARGET_DEVICE_NAME: turris-omnia-1
+    OPENWRT_IMAGE: openwrt-mvebu-cortexa9-cznic_turris-omnia-sysupgrade.img.gz
   needs: ["build-for-turris-omnia"]
 
 test-on-glinet-b1300:
@@ -261,6 +263,7 @@ test-on-glinet-b1300:
   variables:
     TARGET_DEVICE: glinet-b1300
     TARGET_DEVICE_NAME: glinet-b1300-1
+    OPENWRT_IMAGE: openwrt-ipq40xx-generic-glinet_gl-b1300-squashfs-sysupgrade.bin
   needs: ["build-for-glinet-b1300"]
 
 test-on-netgear-rax40:
@@ -268,6 +271,7 @@ test-on-netgear-rax40:
   variables:
     TARGET_DEVICE: netgear-rax40
     TARGET_DEVICE_NAME: netgear-rax40-1
+    OPENWRT_IMAGE: NETGEAR_RAX40-squashfs-fullimage.img
   needs: ["build-for-netgear-rax40"]
 
 run-certification-tests:
@@ -279,7 +283,8 @@ run-certification-tests:
     DEVICE_UNDER_TEST: prplmesh
   script:
     - |
-      if [ "$DEVICE_UNDER_TEST" != prplmesh ] ; then
+      if [ "$DEVICE_UNDER_TEST" = netgear-rax40 ] ; then
+          tools/deploy_firmware.py --device netgear-rax40 --target-name "$DEVICE_UNDER_TEST" --image NETGEAR_RAX40-squashfs-fullimage.img
           echo "Deploying to $DEVICE_UNDER_TEST"
           tools/deploy_ipk.sh --certification-mode $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/prplmesh.ipk
       fi
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,3 +1,6 @@
+variables:
+  PYTHONUNBUFFERED: "1"
+
 stages:
   - containers
   - build
--- a/tools/deploy_firmware.py
+++ b/tools/deploy_firmware.py
@@ -1,0 +1,288 @@
+#!/usr/bin/env python3
+
+import time
+import sys
+import os
+import subprocess
+import difflib
+import argparse
+import shutil
+from typing import List
+
+import pexpect
+import pexpect.fdpexpect
+import pexpect.pxssh
+import serial
+
+
+class PrplwrtDevice:
+    """Represents a prplWrt device.
+
+    Offers methods to check if a device needs to be upgraded and to do the actual upgrade.
+
+    It needs to have access to the artifacts of a build job to determine when an upgrade
+    is needed (see `artifacts_dir`).
+    """
+
+    serial_prompt = "_pexpect_prompt_ "
+    """For serial connections we will set this prompt to make it easier to "expect" it."""
+
+    baudrate = 115200
+    """The baudrate of the serial connection to the device."""
+
+    def __init__(self, device: str, name: str, image: str, username: str = "root"):
+        """
+
+        Parameters
+        -----------
+        device: str
+            The name of the platform (example: netgear-rax40).
+        name: str
+            The name of the device (it should ne reachable through ssh without a password).
+        image: str
+            The name of the image that can be used to upgrade the device.
+        username: str, optional
+            The username to use when connecting to the device over SSH.
+        """
+        self.device = device
+        self.name = name
+        self.image = image
+        self.username = username
+
+        self.rootdir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..")
+        self.artifacts_dir = os.path.join(self.rootdir, "build/{}".format(self.device))
+        """The directory where artifacts are stored. It's expected to contain
+        prplwrt-version, and the image file."""
+
+    def set_prompt(self, pexp: pexpect):
+        """Set the prompt to `serial_prompt` on a pexpect object."""
+        pexp.sendline("export PS1='{}'".format(self.serial_prompt))
+
+    def read_remote_version(self) -> List[str]:
+        """Read prplwrt-version on a remote device
+
+        Returns
+        -------
+        List[str]
+            The content of the prplwrt-version file on the device as a list of strings.
+        """
+        version = None
+        with pexpect.pxssh.pxssh(logfile=sys.stdout.buffer) as shell:
+            shell.login(self.name, self.username)
+            shell.sendline("cat /etc/prplwrt-version")
+            shell.expect("cat /etc/prplwrt-version")
+            shell.prompt()
+            # remove the first (empty) line:
+            version = shell.before.decode().splitlines()[1:]
+        return version
+
+    def read_artifacts_dir_version(self) -> List[str]:
+        """Reads the local prplwrt-version.
+
+        The version file is read from the artifacts (see `artifacts_dir`).
+
+        Returns
+        -------
+        List[str]
+            The content of the local prplwrt-version as a list of strings.
+        """
+        with open(os.path.join(self.artifacts_dir, "prplwrt-version")) as version_file:
+            return version_file.read().splitlines()
+
+    def needs_upgrade(self):
+        """Check if a device needs to be updated
+
+        The check is done by comparing prplwrt-version on the target
+        with a local one (see `read_artifacts_dir_version).
+        """
+        artifacts_dir_version = self.read_artifacts_dir_version()
+        remote_version = self.read_remote_version()
+        diff = difflib.unified_diff(artifacts_dir_version, remote_version,
+                                    fromfile='artifacts', tofile='device')
+        diff_str = '\n'.join(diff)
+        if diff_str:
+            print(diff_str)
+        return bool(diff_str)
+
+    def reach(self, attempts: int = 5, wait: int = 5):
+        """Check if the device is reachable via SSH (and optionally try multiple times).
+
+        Parameters
+        ----------
+        attempts: int, optional
+            How many times we try before concluding that the device is unreachable.
+        wait: int, optional
+            The number of seconds to wait between attempts.
+
+        Returns
+        -------
+        bool
+            True if the device is reachable via SSH, false otherwise.
+        """
+        for _ in range(attempts):
+            try:
+                with pexpect.pxssh.pxssh() as shell:
+                    shell.login(self.name, self.username, login_timeout=5)
+                    return True
+            except pexpect.pxssh.ExceptionPxssh:
+                print("Waiting for the device to be reachable")
+                time.sleep(wait)
+        return False
+
+    def sysupgrade(self):
+        """Upgrade the device using sysupgrade."""
+        raise NotImplementedError
+
+    def upgrade_uboot(self):
+        """Upgrade the device through uboot."""
+        raise NotImplementedError
+
+
+class NetgearRax40(PrplwrtDevice):
+    """A Netgear rax40.
+
+    At the moment, the rax40 can only be updated through uboot using a tftp server.
+    """
+
+    tftp_dir = "/srv/tftp"
+    """The root directory of the tftp server. The image will be copied there"""
+
+    uboot_prompt = "GRX500 #"
+    """The u-boot prompt on the target."""
+
+    def upgrade_uboot(self):
+        """Upgrade the device through u-boot.
+
+        It requires a running tftp server listenning on the IP the
+        target device will expect (see `ipaddr` in the uboot environment).
+        """
+        serial_path = "/dev/{}".format(self.name)
+        if not os.path.exists(serial_path):
+            raise ValueError("The serial device {} does not exist!\n".format(serial_path)
+                             + "Please make sure you have an appropriate udev rule for it.")
+        print("Copying image '{}' to '{}'".format(self.image, self.tftp_dir))
+        shutil.copy(os.path.join(self.artifacts_dir, self.image), self.tftp_dir)
+        print("Image copied to {}.".format(self.tftp_dir))
+        with serial.Serial(serial_path, self.baudrate) as ser:
+            print("Connecting to serial")
+            shell = pexpect.fdpexpect.fdspawn(ser, logfile=sys.stdout.buffer)
+            if not shell.isalive():
+                raise ValueError("Unable to connect to the serial device!")
+            print("Connected")
+            # The console might not be active yet:
+            shell.sendline("")
+            # make the shell prompt appear:
+            self.set_prompt(shell)
+            shell.expect(self.serial_prompt)
+            # reboot:
+            shell.sendline("reboot")
+            shell.expect(["Hit any key to stop autoboot:",
+                          pexpect.EOF, pexpect.TIMEOUT], timeout=120)
+            # stop autoboot:
+            shell.sendline("")
+            shell.expect(self.uboot_prompt)
+            # do the actual upgrade:
+            shell.sendline("run update_fullimage")
+            shell.expect(self.uboot_prompt, timeout=600)
+            # reboot:
+            shell.sendline("reset")
+            shell.expect("Please press Enter to activate this console", timeout=180)
+            # activate the console:
+            shell.sendline("")
+            self.set_prompt(shell)
+            shell.expect(self.serial_prompt)
+            shell.sendline("exit")
+        if not self.reach(attempts=10):
+            raise ValueError("The device was not reachable after the upgrade!")
+
+
+class Generic(PrplwrtDevice):
+    """A generic PrplwrtDevice.
+
+    Since upgrading through uboot is generally target-specific, it
+    only offers the sysupgrade option.
+    """
+
+    def sysupgrade(self):
+        serial_path = "/dev/{}".format(self.name)
+        if not os.path.exists(serial_path):
+            raise ValueError("The serial device {} does not exist!\n".format(serial_path)
+                             + "Please make sure you have an appropriate udev rule for it.")
+        print("Copying image '{}' to the target".format(self.image))
+        try:
+            subprocess.check_output(["scp",
+                                     "{}/{}".format(self.artifacts_dir, self.image),
+                                     "{}:/tmp/{}".format(self.name, self.image)])
+        except subprocess.CalledProcessError as exc:
+            print("Failed to copy the image to the target:\n{}".format(exc.output))
+            raise exc
+        with serial.Serial(serial_path, self.baudrate) as ser:
+            print("Connecting to serial")
+            shell = pexpect.fdpexpect.fdspawn(ser, logfile=sys.stdout.buffer)
+            if not shell.isalive():
+                raise ValueError("Unable to connect to the serial device!")
+            print("Connected")
+            # The console might not be active yet:
+            shell.sendline("")
+            # make the shell prompt appear:
+            self.set_prompt(shell)
+            shell.expect(self.serial_prompt)
+            # Do the upgrade
+            shell.sendline("sysupgrade -v /tmp/{}".format(self.image))
+            # first give it 30 seconds, and fail early if the upgrade didn't start:
+            shell.expect(r"Performing system upgrade\.\.\.", timeout=30)
+            # now give it more time to apply the upgrade and reboot:
+            shell.expect(r"Rebooting system\.\.\.", timeout=300)
+            shell.expect("Please press Enter to activate this console", timeout=180)
+            # activate the console:
+            shell.sendline("")
+            self.set_prompt(shell)
+            shell.expect(self.serial_prompt)
+            shell.sendline("exit")
+        if not self.reach(attempts=10):
+            raise ValueError("The device was not reachable after the upgrade!")
+
+
+def main():
+    parser = argparse.ArgumentParser(prog=sys.argv[0],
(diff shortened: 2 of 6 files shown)


tools/docker/builder/openwrt/Dockerfile
```diff
@@ -95,7 +95,15 @@ RUN cp feeds.conf.default feeds.conf \
...
-    && make defconfig \
+    && make defconfig; \
+    : "We need to keep the hashes in the firmware, to later know whether an upgrade is needed:"; \
+    mkdir -p files/etc \
+    && printf '%s=%s\n' "OPENWRT_REPOSITORY" "$OPENWRT_REPOSITORY" >> files/etc/prplwrt-version \
+    && printf '%s=%s\n' "OPENWRT_VERSION" "$OPENWRT_VERSION" >> files/etc/prplwrt-version \
+    && printf '%s=%s\n' "PRPLMESH_VARIANT" "$PRPLMESH_VARIANT" >> files/etc/prplwrt-version \
+    && printf '%s=%s\n' "PRPL_FEED" "$PRPL_FEED" >> files/etc/prplwrt-version \
+    && printf '%s=%s\n' "INTEL_FEED" "$INTEL_FEED" >> files/etc/prplwrt-version \
+    && printf '%s=%s\n' "IWLWAV_FEED" "$IWLWAV_FEED" >> files/etc/prplwrt-version \
```
arnout, line 106: Most of these are already available in the standard /etc/os-release. We will need to customize them, e.g.:
```
CONFIG_VERSIONOPT=y
CONFIG_VERSION_NUMBER="$OPERNWRT_VERSION"
CONFIG_VERSION_REPO="$OPENWRT_REPOSITORY"
CONFIG_VERSION_MANUFACTURER_URL="$PRPL_FEED"
CONFIG_VERSION_HOME_URL="$INTEL_FEED"
CONFIG_VERSION_BUG_URL="$IWLWAV_FEED" # Notice what I did here? :-)
```

As to PRPLMESH_VARIANT: why do we need that?

(Consider this entire comment a nitpick)

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+
+import pexpect
+import pexpect.fdpexpect
+import serial
+
+
+class PrplwrtDevice:
+    """Represents a prplWrt device.
+
+        Offers methods to check if a device needs to be upgraded and to do the actual upgrade.
```
arnout: I believe this is indented one space too much.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+        username: str, optional
+            The username to use when connecting to the device over SSH.
+        """
+        self.device = device
+        self.name = name
+        self.image = image
+        self.username = username
+
+        self.scriptdir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..")
+        self.rootdir = os.path.realpath(os.path.join(self.scriptdir))
```
arnout: Something fishy here...
```suggestion
        self.rootdir = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..")
```

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+        self.artifacts_dir = os.path.join(self.rootdir, "build/{}".format(self.device))
+        """The directory where artifacts are stored. It's expected to contain
+        prplwrt-version, and the image file."""
+
+    def set_prompt(self, pexp: pexpect):
+        """Set the prompt to `shell_prompt` on a pexpect object."""
+        pexp.sendline("export PS1='{}'".format(self.shell_prompt))
+        pexp.sendline("")
```
arnout: Why the additional sendline?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def read_remote_version(self) -> List[str]:
+        """Read prplwrt-version on a remote device
+
+        Returns
+        -------
+        List[str]
+            The content of the prplwrt-version file on the device as a list of strings.
+        """
+        with pexpect.spawn("ssh {}@{}".format(self.username, self.name)) as shell:
```
arnout: Why not use [pxssh](https://pexpect.readthedocs.io/en/stable/api/pxssh.html)?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+            shell.expect("{}@".format(self.username))
+            shell.sendline("cat /etc/prplwrt-version")
+            shell.expect("cat /etc/prplwrt-version")
+            shell.expect("{}@".format(self.username))
```
arnout: With pxssh, this wouldn't be needed.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+
+        The version file is read from the artifacts (see `artifacts_dir`).
+
+        Returns
+        -------
+        List[str]
+            The content of the local prplwrt-version as a list of strings.
+        """
+        with open(os.path.join(self.artifacts_dir, "prplwrt-version")) as version_file:
+            return version_file.read().splitlines()
```
arnout, line 90: Isn't this exactly the same as
```suggestion
            return version_file.readlines()
```
?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+        -------
+        List[str]
+            The content of the prplwrt-version file on the device as a list of strings.
+        """
+        with pexpect.spawn("ssh {}@{}".format(self.username, self.name)) as shell:
+            shell.expect("{}@".format(self.username))
+            shell.sendline("cat /etc/prplwrt-version")
+            shell.expect("cat /etc/prplwrt-version")
+            shell.expect("{}@".format(self.username))
+            return shell.before.decode().split()
```
arnout: Theoretically some of the strings may contain spaces, so splitlines would be better.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def read_local_version(self) -> List[str]:
```
arnout: (nitpick) 'local' wasn't very clear to me (I thought it meant "locally on the target device" :-)). Maybe `read_artifacts_dir_version`?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+
+        The check is done by comparing prplwrt-version on the target
+        with a local one (see `read_local_version).
+        """
+        local_version = self.read_local_version()
+        remote_version = self.read_remote_version()
+        diff = difflib.unified_diff(local_version, remote_version,
+                                    fromfile='file1', tofile='file2', lineterm='')
+        diff_str = '\n'.join(diff)
+        print(diff_str)
```
arnout:
```suggestion
        if diff_str:
            print(diff_str)
```
Otherwise you get an empty line.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+            The number of seconds to wait between attempts.
+
+        Returns
+        -------
+        bool
+            True if the device is reachable via SSH, false otherwise.
+        """
+        for _ in range(attempts):
+            try:
+                subprocess.check_output(["ssh", "{}@{}"
```
arnout: This should be done with a reasonably small timeout (e.g. if the device still listens to the TCP connection but is unable to spawn a new ssh process to service it) - maybe 3 or 5 seconds? pxssh has a timeout argument can probably be used for this.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+        serial_path = "/dev/{}".format(self.name)
+        if not os.path.exists(serial_path):
+            raise ValueError("The serial device {} does not exist!\n".format(serial_path)
+                             + "Please make sure you have an appropriate udev rule for it.")
+        print("Copying image '{}' to '{}'".format(self.image, self.tftp_dir))
+        shutil.copy(os.path.join(self.artifacts_dir, self.image), self.tftp_dir)
+        print("Image copied to {}.".format(self.tftp_dir))
+        with serial.Serial(serial_path, self.baudrate) as ser:
+            print("Connecting to serial")
+            shell = pexpect.fdpexpect.fdspawn(ser, logfile=sys.stdout.buffer)
```
arnout, line 168: Maybe pexpect.replwrap would be more suited here?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+            if not shell.isalive():
+                raise ValueError("Unable to connect to the serial device!")
+            print("Connected")
+            # The console might not be active yet:
+            shell.sendline("")
+            # make the shell prompt appear:
+            self.set_prompt(shell)
+            shell.expect(self.shell_prompt)
+            # reboot:
```
arnout: It would be nice to refactor all this with the powercycle script at some point.

Also, we will have to think about how to refactor this with boardfarm. Maybe we should even use boardfarm classes for this stuff?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+            shell.sendline("reboot")
+            shell.expect(["Hit any key to stop autoboot:",
+                          pexpect.EOF, pexpect.TIMEOUT], timeout=120)
+            # stop autoboot:
+            shell.sendline("")
+            shell.expect(self.uboot_prompt)
+            # do the actual upgrade:
+            shell.sendline("run update_fullimage")
+            shell.expect(self.uboot_prompt, timeout=600)
```
arnout, line 186: 10 minutes? Does it really take that long?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+    """A generic PrplwrtDevice.
+
+    Since upgrading through uboot is generally target-specific, it
+    only offers the sysupgrade option.
+    """
+
+    def sysupgrade(self):
+        serial_path = "/dev/{}".format(self.name)
+        if not os.path.exists(serial_path):
+            raise ValueError("The serial device {} does not exist!\n".format(serial_path)
```
arnout, line 209: (nitpick) This and the opening of the serial could be refactored into a function in the base class. Something like `open_serial_shell` that opens the serial and returns a pexpect.replwrap.REPLWrapper object.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+        if not self.reach(attempts=10):
+            raise ValueError("The device was not reachable after the upgrade!")
+
+
+def main():
+    parser = argparse.ArgumentParser(prog=sys.argv[0],
+                                     description="""Update a prplWrt
+                                     device, either through u-boot or
+                                     using sysupgrade, depending on
+                                     the target device.""")
```
arnout: Wrapping a bit too enthusiastically here...

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+    parser.add_argument('-d', '--device',
+                        help="""Device to upgrade. Currently supported targets are:
+                        netgear-rax40
+                        glinet-b1300
+                        turris-omnia
+                        """, required=True)
+    parser.add_argument(
+        '-t',
+        '--target-name',
+        help="Name of the target to upgrade (make sure it's reachable through ssh).", required=True)
```
arnout: In my mind, 'device' would be the device instance (e.g. netgear-rax40-1) while 'target' would be the device type (e.g. netgear-rax40)... But apparently we already use this terminology in various places, so I'll have to adapt :-)

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+
+    parser.add_argument(
+        '-i',
+        '--image',
+        help="Name of the image to use for the upgrade (should be from the artifacts folder).",
```
arnout:
```suggestion
        help="Name of the image to use for the upgrade (should exist in the artifacts folder).",
```

tools/deploy_firmware.py
```diff
@@ -0,0 +1,288 @@
...
+        required=True)
+
+    args = parser.parse_args()
+
+    if args.device == "netgear-rax40":
+        dev = NetgearRax40(args.device, args.target_name, args.image)
+    elif args.device in ["glinet-b1300", "turris-omnia"]:
```
arnout: (nitpick) not really needed, since it's *generic*, remember... This could be used with anything as long as build/DEVICE/IMAGE exists.

.gitlab-ci.yml
```diff
@@ -228,7 +229,7 @@ run-tests:
     when: always
   tags:
     - targets
-  timeout: 5m
+  timeout: 10m
```
arnout: If the overall timeout is only 10 minutes, then the timeout of each step should definitely be smaller (see my other comment about the 10 minute timeout).

However, this is a step that we really *don't* want to interrupt, so the timeout here should probably be even larger.

.gitlab-ci.yml
```diff
@@ -275,7 +279,8 @@ run-certification-tests:
     DEVICE_UNDER_TEST: prplmesh
   script:
       - |
-        if [ "$DEVICE_UNDER_TEST" != prplmesh ] ; then
+        if [ "$DEVICE_UNDER_TEST" = netgear-rax40 ] ; then
+            tools/deploy_firmware.py --device netgear-rax40 --target-name "$DEVICE_UNDER_TEST" --image NETGEAR_RAX40-squashfs-fullimage.img
```
arnout: Now I see this, I'm thinking that it would be nicer to have a .deploy-to-target template with a before_script that:
- kills all targets;
- deploys firmware;
- deploys ipk.

Because I think the certification setup would also benefit from killing the other targets.

(This and my other comments to this file are probably for a follow-up PR).

.gitlab-ci.yml
```diff
@@ -275,7 +279,8 @@ run-certification-tests:
     DEVICE_UNDER_TEST: prplmesh
   script:
       - |
-        if [ "$DEVICE_UNDER_TEST" != prplmesh ] ; then
+        if [ "$DEVICE_UNDER_TEST" = netgear-rax40 ] ; then
```
arnout: Maybe it's better to just check that OPENWRT_IMAGE is set (and use OPENWRT_IMAGE below)?

arnout (approved):
Lots of new python code so obviously lots of comments. I don't think there are many that are critical though - feel free to ignore them.

tools/docker/builder/openwrt/Dockerfile
rmelotte: I like the suggestion, but the upcoming prplwrt release will use around 9 feeds for the rax40, so we will need to store them somewhere else anyway (unless we store them all in BUG_URL? :eyes: ).

So for now and also because I need this for the current task I'm working on, I'll keep them in  the prplwrt-release file.

tools/deploy_firmware.py
rmelotte: In fact if I look at [numpy's docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) (which is mostly what we've been following up to now) and also what is in `test_flows.py`, it's rather 3 spaces too many. I'll update it.

tools/deploy_firmware.py
rmelotte: Indeed, thanks.

tools/deploy_firmware.py
rmelotte: I wasn't sure it worked without it the first time, it's probably not needed.

tools/deploy_firmware.py
rmelotte: Because I initially thought I had to "manually" configure the location of the config file, as well as the ssh keys to use.
Apparently it's not needed, so I switched to pxssh.

tools/deploy_firmware.py
rmelotte: The only difference I can see is that readlines() doesn't strip the "\n" at the end of each lines, while read().splitlines() does (and we need them to be stripped).

tools/deploy_firmware.py
rmelotte: Done.

tools/deploy_firmware.py
rmelotte: Done

tools/deploy_firmware.py
rmelotte: I also expect some refactoring to be required once we start using Boardfarm, that's why kept to what I strictly need, and all in a single file.

tools/deploy_firmware.py
rmelotte: It doesn't take that long, but since in most cases it will anyway not wait for the timeout, I went with a "big enough" number so that we don't have to change it later.
600 seconds is probably a bit exaggerated though, I'll change it to 300.

.gitlab-ci.yml
rmelotte: Very good point, some of the devices we have take more time than others, and we don't know about the devices we don't support yet.

Ideally this timeout should never be reached, so I'll set it to 20 minutes for now.

.gitlab-ci.yml
rmelotte: I agree on both comments. My secret plan was to get rid of `run-certification-tests` and to also use a single job per test for the nightly tests.

It requires some more work though, so as you said I'll keep it for another PR.

tools/deploy_firmware.py
rmelotte: I agree. 
I only added the serial connection for the "generic" targets later on and missed this.
I'll also keep this for later however, because I really need this PR to test the new prplWrt release candidate.

tools/deploy_firmware.py
rmelotte: In some other places we also use `TARGET_DEVICE` and `TARGET_DEVICE_NAME` :upside_down_face: 
I find "device" and "target" alone to also be ambiguous. Maybe we should use "device hostname" and "device type" or something like that.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+    print("Checking if the device is reachable over ssh")
+    if not dev.reach():
+        raise ValueError("The device {} is not reachable over ssh! check your ssh configuration."
+                         .format(dev.name))
+    print("Checking if the device needs to be upgraded")
+    if dev.needs_upgrade():
+        print("The device {} will be upgraded".format(dev.name))
+        try:
+            dev.sysupgrade()
+        except NotImplementedError:
```
ghost, line 280: You might want to also try to fall back to `upgrade_uboot` in other cases as well (i.e. when sysupgrade fails). So it might make sense to catch `Exception` here

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+import shutil
+from typing import List
+
+import pexpect
+import pexpect.fdpexpect
+import pexpect.pxssh
+import serial
+
+
+class PrplwrtDevice:
```
ghost, line 18: nitpick: I'd call the base class `Device` and the derived class `prplWrtDevice`, personally.

"prplWrt device" is more specific than "device", so this class hierarchy feels backwards

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+        device: str
+            The name of the platform (example: netgear-rax40).
+        name: str
+            The name of the device (it should ne reachable through ssh without a password).
+        image: str
+            The name of the image that can be used to upgrade the device.
+        username: str, optional
+            The username to use when connecting to the device over SSH.
+        """
+        self.device = device
```
ghost, line 47: nitpick: Consider not making the device name an argument of the constructor but rather a property of the device class, overwritten by each subclass (the base class implementation can raise `NotImplementedError`)

tools/docker/builder/openwrt/Dockerfile
ghost: (nitpick): Any reason to prefer printf here over echo?
`echo "OPENWRT_REPOSITORY=$OPENWRT_REPOSITORY" >> files/etc/prplwrt-version` seems simpler to me :-)

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+
+        The version file is read from the artifacts (see `artifacts_dir`).
+
+        Returns
+        -------
+        List[str]
+            The content of the local prplwrt-version as a list of strings.
+        """
+        with open(os.path.join(self.artifacts_dir, "prplwrt-version")) as version_file:
+            return version_file.read().splitlines()
```
ghost, line 90: Might be nice in the future to store or return it as a map, e.g.
```python
return dict([(k,v) for k, v in (l.split('=') for l in version_file.read().splitlines())])
```

ghost (approved):
No major issues, so approving

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+        -------
+        bool
+            True if the device is reachable via SSH, false otherwise.
+        """
+        for _ in range(attempts):
+            try:
+                with pexpect.pxssh.pxssh() as shell:
+                    shell.login(self.name, self.username, login_timeout=5)
+                    return True
+            except pexpect.pxssh.ExceptionPxssh:
```
LiorAmram, line 127: I'm not sure retrying would always be effective. If the exception is (for example) "permission denied" or "password refused" there is no point in retrying and spending waiting time.
The exception can be parsed and only retry when we have an error that can be solved by retrying.

The source details all the possible exceptions on login:
https://pexpect.readthedocs.io/en/stable/_modules/pexpect/pxssh.html

tools/deploy_firmware.py
rmelotte: That's what I initially wanted to do, but the content of the file might change a lot very soon (with the new prplwrt release candidate), and it could really be any format.
Since we really only need to know if there is any difference and not what the difference is, I just went with a list of lines instead.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+
+    parser.add_argument(
+        '-i',
+        '--image',
+        help="Name of the image to use for the upgrade (should exist in the artifacts folder).",
+        required=True)
+
+    args = parser.parse_args()
+
+    if args.device == "netgear-rax40":
```
LiorAmram, line 266: perhaps compare to `args.device.lower()` to avoid capital letter issues in comparison.

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+    print("Checking if the device needs to be upgraded")
+    if dev.needs_upgrade():
+        print("The device {} will be upgraded".format(dev.name))
+        try:
+            dev.sysupgrade()
+        except NotImplementedError:
+            dev.upgrade_uboot()
+        print("Done")
+    else:
+        print("The device is already using the same version, no upgrade will be done.")
```
LiorAmram, line 284: I wonder if `force` option should be added as an optional argument, for cases that we might want to flash regardless of the current version.
What do you think?

tools/deploy_firmware.py
```diff
@@ -0,0 +1,287 @@
...
+    At the moment, the rax40 can only be updated through uboot using a tftp server.
+    """
+
+    tftp_dir = "/srv/tftp"
+    """The root directory of the tftp server. The image will be copied there"""
+
+    uboot_prompt = "GRX500 #"
+    """The u-boot prompt on the target."""
+
+    def upgrade_uboot(self):
```
tomereli, line 153: @LiorAmram will soon work on adding support for this script to upgrade UGW and RDKB.
Both UGW and prplWrt RAX40 are using the same uboot so upgrading a version is practially the same. UGW is however not a PrplwrtDevice so we won't be able to reuse this like this.

RAX40 is basically prplWrt with UGW BSP. The SoC is called GRX (we have GRX350 and GRX500 but the differences are negligible) - so its better to move the upgrade_uboot and sysupgrade to a different class. Perhaps `class device_operations` which will have abstract methods for `sysupgrade` and `uboot_upgrade`, and add it as a member of the prplwrt_device class?

tools/deploy_firmware.py
tomereli: Not every device will support `sysupgrade` since not all devices are OpenWRT devices - so I think it makes more sense to abstract this here and just call `dev.upgrade()` which internally will try sysupgrade if it exists for this device and fallback to uboot_upgrade.
For example, CGR device doesn't have either - there's a BIOS from which we run tftp commands directly to upgrade firmware  - here is how, in case you want to add (untested) support for Intel RDKB as part of this PR:
image

tomereli (approved):
Mainly comments about future support for other platforms, but nothing that needs to be addressed right away, so approving.

tools/deploy_firmware.py
arnout: Since it is anyway work to rewrite the code to satisfy those future constraints, it is better to do that rewrite when we're actually going to do it, rather than trying to predict now what it should look like (in other words: merge as is, refactor later).

arnout: Agh sorry that's exactly what you said.

tools/deploy_firmware.py
rmelotte: True, that would be an improvement (for later, it won't have any big impact in CI).

pxssh raises the same exception for all the cases though, so it won't be possible to have different except closes for the different cases. Is it what you were suggesting?
It would also be possible to distinguish the cases based on the exception message, but it's ugly and prone to fail between different versions of pxssh.

tools/deploy_firmware.py
rmelotte: We use "netgear-rax40" everywhere as the name of the device/target, I don't see much advantage in making it case-insensitive. What use-case do you have in mind?

tools/deploy_firmware.py
rmelotte: It could be useful if you want to try experimental builds I guess, but I'll keep that for later also.

PR #1283, opened 2020-05-12
Implement AP Metrics reporting with channel utilization

mariomaz:
If a Multi-AP Agent receives a Metric Reporting Policy TLV with AP Metrics Channel Utilization Reporting Threshold field set to a non-zero value for a given radio, it shall measure the channel utilization on that radio in each consecutive implementation-specific measurement period and, if the most recently measured channel utilization has crossed the reporting threshold in either direction (with respect to the previous measurement), it shall send an AP Metrics Response message     to the Multi-AP Controller containing one AP Metrics TLV for each of the BSSs on that radio.
    
This PR also adds a placeholder for AP Metrics Response message creation in a method that will also be called when processing AP Metrics Query message.

### DIFF
--- a/agent/src/beerocks/monitor/monitor_thread.cpp
+++ b/agent/src/beerocks/monitor/monitor_thread.cpp
@@ -27,6 +27,13 @@ using namespace son;
 #define OPERATION_SUCCESS 0
 #define OPERATION_FAIL -1
 
+/**
+ * Implementation-specific measurement period of channel utilization.
+ * Currently we use this constant value but a more elaborate solution should read it from
+ * configuration.
+ */
+static constexpr uint8_t ap_metrics_channel_utilization_measurement_period_s = 10;
+
 monitor_thread::monitor_thread(const std::string &slave_uds_, const std::string &monitor_iface_,
                                beerocks::config_file::sConfigSlave &beerocks_slave_conf_,
                                beerocks::logging &logger_)
@@ -496,7 +503,86 @@ void monitor_thread::after_select(bool timeout)
 #ifdef BEEROCKS_RDKB
         mon_rdkb_hal.process();
 #endif
+
+        /**
+         * If a Multi-AP Agent receives a Metric Reporting Policy TLV with AP Metrics Channel
+         * Utilization Reporting Threshold field set to a non-zero value for a given radio, it
+         * shall measure the channel utilization on that radio in each consecutive implementation-
+         * specific measurement period and, if the most recently measured channel utilization has
+         * crossed the reporting threshold in either direction (with respect to the previous
+         * measurement), it shall send an AP Metrics Response message to the Multi-AP Controller
+         * containing one AP Metrics TLV for each of the BSSs on that radio.
+         */
+        auto radio_node = mon_db.get_radio_node();
+        auto &info      = radio_node->ap_metrics_reporting_info();
+        if (0 != info.ap_channel_utilization_reporting_threshold) {
+            int elapsed_time_s =
+                std::chrono::duration_cast<std::chrono::seconds>(
+                    now - info.ap_metrics_channel_utilization_last_reporting_time_point)
+                    .count();
+
+            if (elapsed_time_s >= ap_metrics_channel_utilization_measurement_period_s) {
+                info.ap_metrics_channel_utilization_last_reporting_time_point = now;
+
+                on_channel_utilization_measurement_period_elapsed();
+            }
+        }
     }
+}
+
+void monitor_thread::on_channel_utilization_measurement_period_elapsed()
+{
+    /**
+     * Measure current channel utilization on the radio.
+     */
+    uint8_t channel_utilization;
+    if (!mon_wlan_hal->get_channel_utilization(channel_utilization)) {
+        LOG(ERROR) << "Unable to get channel utilization";
+        return;
+    }
+
+    /**
+     * If previous channel utilization was lower than the threshold and now it is higher than the
+     * threshold, report it.
+     * Or if previous channel utilization was higher than the threshold and now it's lower than
+     * the threshold, report it.
+     */
+    auto radio_node        = mon_db.get_radio_node();
+    auto &info             = radio_node->ap_metrics_reporting_info();
+    bool threshold_crossed = false;
+    if (channel_utilization > info.ap_channel_utilization_reporting_threshold) {
+        if (info.ap_metrics_channel_utilization_reporting_value <=
+            info.ap_channel_utilization_reporting_threshold) {
+            threshold_crossed = true;
+        }
+    } else if (info.ap_metrics_channel_utilization_reporting_value >
+               info.ap_channel_utilization_reporting_threshold) {
+        threshold_crossed = true;
+    }
+
+    if (threshold_crossed) {
+        if (!create_ap_metrics_response()) {
+            LOG(ERROR) << "Unable to create AP Metrics Response message";
+            return;
+        }
+
+        message_com::send_cmdu(slave_socket, cmdu_tx);
+    }
+
+    info.ap_metrics_channel_utilization_reporting_value = channel_utilization;
+}
+
+bool monitor_thread::create_ap_metrics_response()
+{
+    auto cmdu_tx_header = cmdu_tx.create(0, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "Failed creating AP_METRICS_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    // TODO: add TLVs to message
+
+    return true;
 }
 
 bool monitor_thread::update_sta_stats()
--- a/agent/src/beerocks/monitor/monitor_thread.h
+++ b/agent/src/beerocks/monitor/monitor_thread.h
@@ -59,6 +59,23 @@ private:
     void stop_monitor_thread();
     bool hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event_ptr);
 
+    /**
+     * @brief Executes when channel utilization measurement period has elapsed.
+     *
+     * This method is periodically invoked if AP Metrics Channel Utilization Reporting Threshold
+     * has been set to a non-zero value in last received Metric Reporting Policy TLV.
+     * Measurement period is set to an implementation-specific value.
+     * On invocation, method shall measure current channel utilization on the radio. If difference
+     * with respect to the previous measurement has crossed the reporting threshold, it shall send
+     * an AP Metrics Response message to the controller.
+     */
+    void on_channel_utilization_measurement_period_elapsed();
+
+    /**
+     * @brief Creates AP Metrics Response message
+     */
+    bool create_ap_metrics_response();
+
     bool update_ap_stats();
     bool update_sta_stats();
 
--- a/common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
+++ b/common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
@@ -51,6 +51,19 @@ public:
     virtual bool process_nl_events() override;
     virtual std::string get_radio_mac() override;
 
+    /**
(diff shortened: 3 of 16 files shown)


rmelotte (approved):
Can you think of any way to test the dummy implementation from test_flows.py at the moment?

common/beerocks/bwl/shared/nl80211_client.cpp
```diff
@@ -0,0 +1,37 @@
...
+        return false;
+    }
+
+    auto survey_info = *it;
+
+    if (0 == survey_info.time_on_ms) {
+        return false;
+    }
+
+    channel_utilization = survey_info.time_busy_ms / survey_info.time_on_ms;
```
arnout: I guess you're missing a unit test :-)

This will always be zero, because you're doing integer division and time_on_ms >= time_busy_ms. You'll have to multiply time_busy_ms by 255 before the division.

Also, there should probably be some handling of the boundary conditions (time_busy_ms > time_on_ms, which shouldn't happen, so return false).

common/beerocks/bwl/shared/nl80211_client.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace bwl {
+
+bool nl80211_client::SurveyInfo::get_channel_utilization(uint8_t &channel_utilization) const
+{
+    channel_utilization = UINT8_MAX;
+
+    auto it =
+        std::find_if(data.begin(), data.end(), [](const sChannelSurveyInfo &channel_survey_info) {
+            return channel_survey_info.in_use;
```
arnout, line 30: What if there are multiple channels in use? This might actually happen in the 80+80 case. Or maybe even in the 80MHz case because it covers multiple channels.

(note: can be a TODO)

common/beerocks/bwl/include/bwl/nl80211_client.h
```diff
@@ -351,6 +351,28 @@ class nl80211_client {
...
+         * NL80211_CMD_GET_SURVEY command.
+         */
+        std::vector<sChannelSurveyInfo> data;
+
+        /**
+         * @brief Gets channel utilization.
+         *
+         * Gets channel utilization as time busy divided by time on for the channel in use.
+         *
+         * @param[out] channel_utilization Channel utilization value or UINT8_MAX if it cannot be
```
arnout: This should specify the unit. Well, since it's a fraction no unit *should* be needed, but representing a fraction as a single 8-bit integer is not going to give you much joy :-) I propose to use a unit of 1/255 (as per 802.11 spec section 9.4.2.28:  "percentage of time, linearly scaled with 255 representing 100%").

common/beerocks/bwl/include/bwl/base_wlan_hal.h
```diff
@@ -110,6 +110,15 @@ class base_wlan_hal {
      */
     virtual bool process_int_events();
 
+    /**
+     * @brief Gets channel utilization.
+     *
+     * @param[out] channel_utilization Channel utilization value.
```
arnout, line 121: Again, we need to specify the unit here.

agent/src/beerocks/monitor/monitor_db.h
```diff
@@ -296,7 +296,7 @@ class monitor_radio_node {
          * Last value reported for AP Metrics Channel Utilization Reporting.
          * Must be compared with threshold value to decide if current value has to be reported.
          */
-        uint8_t ap_metrics_channel_utilization_reporting_value = 0;
+        uint8_t ap_metrics_channel_utilization_reporting_value = UINT8_MAX;
```
arnout: Unfortunately, UINT8_MAX is *not* invalid, it represents 100%. If we need to support a sentinel value, we'll have to use a bigger size.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -25,6 +25,13 @@ using namespace son;
 #define OPERATION_SUCCESS 0
 #define OPERATION_FAIL -1
 
+/**
+ * Implementation-specific measurement period of channel utilization.
+ * Currently we use this constant value but a more elaborate solution should read it from
+ * configuration.
```
arnout, line 33: :+1:

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -25,6 +25,13 @@ using namespace son;
...
+ */
+static constexpr uint8_t ap_metrics_channel_utilization_measurement_period_s = 10;
```
arnout, line 35: (for my info) why constexpr and not const?

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -494,7 +501,75 @@ void monitor_thread::after_select(bool timeout)
...
+         * measurement), it shall send an AP Metrics Response message to the Multi-AP Controller
+         * containing one AP Metrics TLV for each of the BSSs on that radio.
+         */
+        auto radio_node = mon_db.get_radio_node();
+        auto &info      = radio_node->ap_metrics_reporting_info();
+        if (0 != info.ap_channel_utilization_reporting_threshold) {
+            int elapsed_time_s =
+                std::chrono::duration_cast<std::chrono::seconds>(
+                    now - info.ap_metrics_channel_utilization_last_reporting_time_point)
+                    .count();
```
arnout, line 522: We need a better chrono abstraction to make this kind of stuff easier (and mockable)...

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -494,7 +501,75 @@ void monitor_thread::after_select(bool timeout)
...
+        return;
     }
+
+    /**
+     * Compare with previous measurement and if difference has crossed the reporting threshold,
+     * then send an AP Metrics Response message to the controller.
+     */
+    auto radio_node = mon_db.get_radio_node();
+    auto &info      = radio_node->ap_metrics_reporting_info();
+    if ((UINT8_MAX == info.ap_metrics_channel_utilization_reporting_value) ||
```
arnout: IIUC this is *only* used for the initial report. The spec doesn't say anything about what the initial one should be. So... see below.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -494,7 +501,75 @@ void monitor_thread::after_select(bool timeout)
...
+        (abs(channel_utilization - info.ap_metrics_channel_utilization_reporting_value) >
+         info.ap_channel_utilization_reporting_threshold)) {
```
arnout: This is not correct IMO. The threshold is not a difference like it is for RCPI - it's when you go over the threshold. "if the most recently measured channel utilization has crossed the
reporting threshold in either direction (with respect to the previous measurement)". So:
- the ap_metrics_channel_utilization_reporting_value should be stored every time, not only when the threshold is exceeded;
- we should check if current and previous are on the same side of the threshold.

The second can be done easily be checking the sign of (a - threshold) * (b - threshold).

So the code would look something like this:
```
    auto current_diff = channel_utilization - info.ap_channel_utilization_reporting_threshold;
    auto prev_diff = info.ap_metrics_channel_utilization_reporting_value - info.ap_channel_utilization_reporting_threshold;
    if (cur_diff * prev_diff < 0)
        ...
```

Please double-check the above w.r.t. types and overflows. It may need to but put in an int32_t to be sure. 

As for the initial case: since it's not specified, it doesn't matter on which side of the threshold we start. So I think initialising it to 0 is fine.

agent/src/beerocks/monitor/monitor_thread.h
```diff
@@ -59,6 +59,23 @@ class monitor_thread : public beerocks::socket_thread {
...
+     * On invocation, method shall measure current channel utilization on the radio. If difference
+     * with respect to the previous measurement has crossed the reporting threshold, it shall send
+     * an AP Metrics Response message to the controller.
+     */
+    void on_channel_utilization_measurement_period_elapsed();
+
+    /**
+     * @brief Creates AP Metrics Response message
+     */
+    bool create_ap_metrics_response();
```
arnout, line 77: I put this function in monitor_stats. It would read its information from the database anyway, so putting it in monitor_stats is fine. So on_channel_utilization_measurement_period_elapsed would update the database before calling create_ap_metrics_response(). And I think ti's pretty easy: ap_metrics_channel_utilization_reporting_value and _time should move to sRadioStats. Note that that object already has channel_load field which AFAIU are more or less equivalent (but apparently never got implemented because channel_load_tot_curr is never set and all the rest depends on it).

That said, I'm perfectly fine with moving create_ap_metrics_response to monitor_thread. Like you, I don't see a clear separation of concerns between the monitor classes (except for the db, of course).

Note that the response needs a list of bssids that have to be included in the response.

arnout (changes requested):
Given that there are a few fundamental problems with this PR, I'm marking it as changes requested.

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: In this case we could use either. 
AFAIK `constexpr` is required to ensure we have a compile-time constant but here a `const` would work too. `const` is required when initialization must be done with a computation done at run time. 

I hadn't used `constexpr` before I started working for the prplMesh project. In one of my PRs I used a `#define` and @rmelotte suggested me to use `constexpr` instead. I'm using `constexpr` since then.

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: I've been thinking about using the [Reactor Design Pattern](https://www.dre.vanderbilt.edu/~schmidt/PDF/reactor-rules.pdf) here to handle both socket I/O and timer events

common/beerocks/bwl/shared/nl80211_client.cpp
mariomaz: Fixed

common/beerocks/bwl/include/bwl/nl80211_client.h

common/beerocks/bwl/shared/nl80211_client.cpp
mariomaz: Created #1312

common/beerocks/bwl/include/bwl/base_wlan_hal.h
mariomaz: Done

agent/src/beerocks/monitor/monitor_db.h
mariomaz: Reverted

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: Fixed

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: Fixed

agent/src/beerocks/monitor/monitor_thread.h
mariomaz: My create_ap_metrics_response() method is just a hint of how/where it could be done. 
This is something left to be done when merging your branch and mine

common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
```diff
@@ -51,6 +51,19 @@ class base_wlan_hal_dummy : public virtual base_wlan_hal,
     virtual bool process_nl_events() override;
     virtual std::string get_radio_mac() override;
 
+    /**
+     * @brief Gets channel utilization.
+     *
+     * @see base_wlan_hal::get_channel_utilization
```
arnout, line 57: :+1: 

As @RanRegev noted in #1309 we should have documentation of how to write doxygen, and this should be part of it.

arnout (approved):

arnout:
@mergifyio rebase

PR #1286, opened 2020-05-13
Feature/tlvf types

tomereli:
Add tlvftypes.h, which will be used as a container header for all the
tlvf defined types, and in addition include all the operator overloads,
thus allowing a single header include by users of the tlvf library for
using a certain type.

Move all the operator overloads of sMacAddr which are now defined in
BCL to tlvftypes.h. This will now allow an external application to link
to TLVF alone for using sMacAddr effectively, the first "customer" being
the transport.


common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -84,59 +84,27 @@ const std::string network_utils::MULTICAST_1905_MAC_ADDR("01:80:c2:00:00:13");
...
-                      string_utils::int_to_hex_string((uint32_t)mac_address[5], 2);
-
-    return mac_addr_string;
+    return tlvf::mac_to_string(mac_address);
 }
 
 std::string network_utils::mac_to_string(const sMacAddr &mac)
 {
-    return mac_to_string((const uint8_t *)mac.oct);
+    return tlvf::mac_to_string((const uint8_t *)mac.oct);
```
morantr: I think it is better to remove these functions from network utils, and add one more commit that replaces all the calling for these functions to have the `tlvf::` prefix instead of `network_utils::`.

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -541,18 +541,18 @@ std::ostream &Ieee1905Transport::Packet::print(std::ostream &os) const
...
+       << std::hex << std::setfill('0') << std::setw(2) << (unsigned)dst.oct[2] << ":" << std::hex
+       << std::setfill('0') << std::setw(2) << (unsigned)dst.oct[3] << ":" << std::hex
+       << std::setfill('0') << std::setw(2) << (unsigned)dst.oct[4] << ":" << std::hex
+       << std::setfill('0') << std::setw(2) << (unsigned)dst.oct[5] << std::endl;
+    ss << "  src        : " << std::hex << std::setfill('0') << std::setw(2) << (unsigned)src.oct[0]
+       << ":" << std::hex << std::setfill('0') << std::setw(2) << (unsigned)src.oct[1] << ":"
+       << std::hex << std::setfill('0') << std::setw(2) << (unsigned)src.oct[2] << ":" << std::hex
+       << std::setfill('0') << std::setw(2) << (unsigned)src.oct[3] << ":" << std::hex
+       << std::setfill('0') << std::setw(2) << (unsigned)src.oct[4] << ":" << std::hex
+       << std::setfill('0') << std::setw(2) << (unsigned)src.oct[5] << std::endl;
```
morantr: You don't need to do that. We already have operator overload for this. If you do ` << dst` it will be converted to a string.

morantr (approved):
It makes me wonder. A lot of code on BCL may be fit more to the framework and can be used there.
string utils for example, too bad that we are duplicating code for that.

tomereli:
> It makes me wonder. A lot of code on BCL may be fit more to the framework and can be used there.
> string utils for example, too bad that we are duplicating code for that.

You are right, but I think it is OK to move things in steps when they are needed. I needed sMacAddr so I used this opportunity to move it now.

framework/tlvf/src/src/tlvftypes.cpp
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+#include <climits>
+#include <linux/if_ether.h>
+#include <tlvf/tlvftypes.h>
+
+namespace tlvf {
+std::string int_to_hex_string(const unsigned int integer, const uint8_t number_of_digits)
+{
+    // 'number_of_digits' represent how much digits the number should have, so the function will
```
mariomaz: These comments should be moved to header file, inside a properly doxygen-formatted documentation

framework/tlvf/src/src/tlvftypes.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    return ss_hex_string.str();
+};
+
+// Converts a mac address in a human-readable format
+std::string mac_to_string(const uint8_t *mac_address)
+{
+    std::string mac_addr_string;
+
+    mac_addr_string = int_to_hex_string((uint32_t)mac_address[0], 2) + ":" +
```
mariomaz: Define a constexpr "separator" character instead of repeating ":" all along this file

framework/tlvf/src/src/tlvftypes.cpp
```diff
@@ -0,0 +1,81 @@
...
+                      int_to_hex_string((uint32_t)mac_address[1], 2) + ":" +
+                      int_to_hex_string((uint32_t)mac_address[2], 2) + ":" +
+                      int_to_hex_string((uint32_t)mac_address[3], 2) + ":" +
+                      int_to_hex_string((uint32_t)mac_address[4], 2) + ":" +
+                      int_to_hex_string((uint32_t)mac_address[5], 2);
```
mariomaz, line 36: Write a loop instead of repeating 6 times the same line

framework/tlvf/src/src/tlvftypes.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+std::string mac_to_string(const sMacAddr &mac) { return mac_to_string((const uint8_t *)mac.oct); }
+
+// Converts uint64_t mac address to string format
+std::string mac_to_string(const uint64_t mac)
+{
+    uint8_t mac_address[ETH_ALEN];
+    int8_t i;
+    uint8_t *p = mac_address;
+    for (i = 5; i >= 0; i--) {
```
mariomaz, line 49: Use ETH_ALEN instead of magic number

framework/tlvf/src/src/tlvftypes.cpp
```diff
@@ -0,0 +1,81 @@
...
+        *p++ = mac >> (CHAR_BIT * i);
+    }
+    return mac_to_string(mac_address);
+}
+
+void mac_from_string(uint8_t *buf, const std::string &mac)
+{
+    if (mac.empty()) {
+        memset(buf, 0, ETH_ALEN);
```
mariomaz: "buf" might be null

common/beerocks/bcl/source/network/network_utils.cpp
tomereli: You just love these huge changes hu? :)
OK Done, but I really hope it will be merged soon otherwise it will be a nightmare on rebases.

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
```diff
@@ -143,8 +144,8 @@ class Ieee1905Transport {
         unsigned int dst_if_index = 0;
         uint8_t src_if_type       = CmduRxMessage::IF_TYPE_NONE;
         unsigned int src_if_index = 0;
-        uint8_t dst[ETH_ALEN]     = {0}; // destination mac address
-        uint8_t src[ETH_ALEN]     = {0}; // source mac address
+        sMacAddr dst              = {.oct = {0}}; // destination mac address
```
mariomaz, line 147: Maybe you should also move ZERO_MAC from network_utils to TLVF

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
```diff
@@ -220,7 +221,7 @@ class Ieee1905Transport {
         bool operator()(const DeFragmentationKey &lhs, const DeFragmentationKey &rhs) const
         {
             // compare based on src address first then messageId
-            int srccmp = memcmp(lhs.src, rhs.src, ETH_ALEN);
+            int srccmp = memcmp(lhs.src.oct, rhs.src.oct, ETH_ALEN);
```
mariomaz, line 224: Now you can use operator ==

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
tomereli: right, fixed.

mariomaz (approved):
A few comments but approving, you decide

framework/tlvf/src/src/tlvftypes.cpp
tomereli: Done.

framework/tlvf/src/src/tlvftypes.cpp
tomereli: I don't really see the benefit for doing this cause it is used only in two functions, once as char * and once as a char, but OK.

framework/tlvf/src/src/tlvftypes.cpp
tomereli: Its not the same line, the last one doesn't include the `":"` separator.

framework/tlvf/src/src/tlvftypes.cpp
tomereli: ETH_ALEN is 6.

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
tomereli: True, but there if I do this I can also move a lot more. I prefer to keep this for a separate PR.

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
tomereli: I prefer to do this as well in a separate PR since it is changing the logic.

framework/tlvf/src/src/tlvftypes.cpp
mariomaz: I meant this:
```c
for (size_t i = 0; i < ETHA_LEN; i++) {
    if (i > 0) {
        mac_addr_string += mac_bytes_separator;
    }
    mac_addr_string += int_to_hex_string(mac_address[i], 2);
}

framework/tlvf/src/src/tlvftypes.cpp
mariomaz: I meant this:
```c
for (i = ETH_ALEN - 1; i >= 0; i--) {

PR #1287, opened 2020-05-13
Bugfix / Fix issues with "scan all channels"

itayx:
As part of the new "Scan on all channels" the client can requests a scan
with a specific pool what will indicate they wish to run a scan on all
the radio's supported channels.
A channel pool with only a single channel with a value of '0' will
indicate that the client wants to use the "Scan on all channels".

### DIFF
--- a/common/beerocks/bcl/include/bcl/beerocks_defines.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_defines.h
@@ -471,6 +471,7 @@ enum class eChannelScanOperationCode : uint8_t {
 };
 
 #define CHANNEL_SCAN_INVALID_PARAM -1
+#define SCAN_ALL_CHANNELS 0
 
 } // namespace beerocks
 
--- a/controller/src/beerocks/master/son_management.cpp
+++ b/controller/src/beerocks/master/son_management.cpp
@@ -30,6 +30,48 @@ using namespace beerocks;
 using namespace net;
 using namespace son;
 
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////// Local Module Functions ///////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+/**
+ * @brief Check if the given channel pool fulfils the "all-channel" scan requirement.
+ * @param channel_pool_set: Unordered set containing the current channel pool.
+ * @return true if pool matches the "all-channel" scan requirement
+ * @return false if not.
+ */
+bool is_scan_all_channels_request(std::unordered_set<uint8_t> &channel_pool_set)
+{
+    return channel_pool_set.size() == 1 && *channel_pool_set.begin() == SCAN_ALL_CHANNELS;
+}
+
+/**
+ * @brief Get the channel pool containing all the supported channels.
+ * @param channel_pool_set:  Set containing the current channel pool.
+ * @param channel_pool_size: Size of the given set.
+ * @param database:          Reference to the DB.
+ * @param radio_mac:         MAC address of radio.
+ */
+static bool get_pool_of_all_supported_channels(std::unordered_set<uint8_t> &channel_pool_set,
+                                               db &database, const sMacAddr &radio_mac)
+{
+    LOG(DEBUG) << "Setting channel pool to all channels";
+    channel_pool_set.clear();
+    auto all_channels = database.get_hostap_supported_channels(tlvf::mac_to_string(radio_mac));
+    if (all_channels.empty()) {
+        LOG(ERROR) << "Supported channel list is empty, failed to set channel pool!";
+        return false;
+    }
+    std::transform(all_channels.begin(), all_channels.end(),
+                   std::inserter(channel_pool_set, channel_pool_set.end()),
+                   [](beerocks::message::sWifiChannel &c) -> uint8_t { return c.channel; });
+    return true;
+}
+
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Implementation ///////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
 void son_management::handle_cli_message(Socket *sd,
                                         std::shared_ptr<beerocks_header> beerocks_header,
                                         ieee1905_1::CmduMessageTx &cmdu_tx, db &database,
@@ -1952,9 +1994,19 @@ void son_management::handle_bml_message(Socket *sd,
             op_error_code == eChannelScanOperationCode::SUCCESS) {
             auto channel_pool_set =
                 std::unordered_set<uint8_t>(channel_pool, channel_pool + channel_pool_size);
-            op_error_code = database.set_channel_scan_pool(radio_mac, channel_pool_set, false)
-                                ? op_error_code
-                                : eChannelScanOperationCode::INVALID_PARAMS_CHANNELPOOL;
+            // Check if "all-channel" scan is requested
+            if (is_scan_all_channels_request(channel_pool_set)) {
+                op_error_code =
+                    get_pool_of_all_supported_channels(channel_pool_set, database, radio_mac)
+                        ? op_error_code
+                        : eChannelScanOperationCode::INVALID_PARAMS_CHANNELPOOL;
+            }
+            // Set the channel pool
+            if (op_error_code == eChannelScanOperationCode::SUCCESS) {
+                op_error_code = database.set_channel_scan_pool(radio_mac, channel_pool_set, false)
+                                    ? op_error_code
+                                    : eChannelScanOperationCode::INVALID_PARAMS_CHANNELPOOL;
+            }
         }
         response->op_error_code() = uint8_t(op_error_code);
         //send response to bml
@@ -2218,7 +2270,14 @@ void son_management::handle_bml_message(Socket *sd,
             break;
         }
 
-        LOG(DEBUG) << "set_channel_scan_pool " << channel_pool_size;
+        if (is_scan_all_channels_request(channel_pool_set))
+            if (!get_pool_of_all_supported_channels(channel_pool_set, database, radio_mac)) {
+                LOG(ERROR) << "set_channel_scan_pool failed";
+                response->op_error_code() =
+                    uint8_t(eChannelScanOperationCode::INVALID_PARAMS_CHANNELPOOL);
+                message_com::send_cmdu(sd, cmdu_tx);
+                break;
+            }
         if (!database.set_channel_scan_pool(radio_mac, channel_pool_set, true)) {
             LOG(ERROR) << "set_channel_scan_pool failed";
             response->op_error_code() =
--- a/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
+++ b/controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
@@ -44,6 +44,10 @@ beerocks::eChannelScanStatusCode dynamic_channel_selection_task::dcs_request_sca
     // get the parent node to send the CMDU to the agent
     auto radio_mac_str = tlvf::mac_to_string(m_radio_mac);
     auto ire           = database.get_node_parent_ire(radio_mac_str);
+    if (ire.empty()) {
+        LOG(ERROR) << "Failed to get node_parent_ire!";
+        return beerocks::eChannelScanStatusCode::INTERNAL_FAILURE;
+    }
     son_actions::send_cmdu_to_agent(ire, cmdu_tx, database, radio_mac_str);
 
     return beerocks::eChannelScanStatusCode::SUCCESS;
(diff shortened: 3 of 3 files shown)


controller/src/beerocks/master/db/db.cpp
```diff
@@ -2298,6 +2298,11 @@ int db::get_channel_scan_dwell_time_msec(const sMacAddr &mac, bool single_scan)
 bool db::is_channel_scan_pool_supported(const sMacAddr &mac,
                                         const std::unordered_set<uint8_t> &channel_pool)
 {
+    if (channel_pool.size() == 1 && channel_pool.find(0) != channel_pool.end()) {
```
kantera800: you should use a define 

```suggestion
    if (channel_pool.size() == 1 && *channel_pool.begin() == SCANALLCHAN) {
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2359,6 +2364,12 @@ bool db::is_channel_in_pool(const sMacAddr &mac, uint8_t channel, bool single_sc
 
     auto &pool =
         (single_scan ? hostap->single_scan_config : hostap->continuous_scan_config).channel_pool;
+
+    if (pool.size() == 1 && pool.find(0) != pool.end()) {
```
kantera800: same here

kantera800 (approved):
Approved , 
just fix the condition

controller/src/beerocks/master/db/db.cpp

controller/src/beerocks/master/db/db.cpp

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2298,6 +2298,11 @@ int db::get_channel_scan_dwell_time_msec(const sMacAddr &mac, bool single_scan)
 bool db::is_channel_scan_pool_supported(const sMacAddr &mac,
                                         const std::unordered_set<uint8_t> &channel_pool)
 {
+    if (channel_pool.size() == 1 && *channel_pool.begin() == SCAN_ALL_CHANNELS) {
```
tomereli: I don't like this at all. If the user wants to scan all channels, then add an API to scan on all channels. APIs should be simple - this function based on its name should only check if the given pool is supported.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2359,6 +2364,12 @@ bool db::is_channel_in_pool(const sMacAddr &mac, uint8_t channel, bool single_sc
 
     auto &pool =
         (single_scan ? hostap->single_scan_config : hostap->continuous_scan_config).channel_pool;
+
+    if (pool.size() == 1 && *pool.begin() == SCAN_ALL_CHANNELS) {
```
tomereli: scan all channels should set the pool to empty and set a member bool scan_all_channels=true instead of this hack.

tomereli (changes requested):

itayx:
@tomereli 
I don't think that creating a new API just for the "scan on all channels" is the correct course of action,
Instead I reworked the PR to do the following:
In the `set_channel_scan_pool` I expanded the signature to accept a "scan all channels" option (set by default to `false`) which, if set to `true` will fill the DCS channel pool with the radio's supported channels.
That way the DCS task will be unaware if the client uses the "scan all channels" feature, thus containing the changes only to the DB's `set_channel_scan_pool` function and its uses in the `son_management`.
I believe this is the better course of action since it allows the DB to the the sole manager of the DCS's channel pool, and we won't need to use "hacks" to fix the issues that rise from it

controller/src/beerocks/master/db/db.cpp
itayx: See my comment below

controller/src/beerocks/master/db/db.cpp
itayx: See my comment below

adam1985d (commented):
I think you meant to write `Preparative Commit` instead of `Preemptive commit`

adam1985d:
> @tomereli
> I don't think that creating a new API just for the "scan on all channels" is the correct course of action,
> Instead I reworked the PR to do the following:
> In the `set_channel_scan_pool` I expanded the signature to accept a "scan all channels" option (set by default to `false`) which, if set to `true` will fill the DCS channel pool with the radio's supported channels.
> That way the DCS task will be unaware if the client uses the "scan all channels" feature, thus containing the changes only to the DB's `set_channel_scan_pool` function and its uses in the `son_management`.
> I believe this is the better course of action since it allows the DB to the the sole manager of the DCS's channel pool, and we won't need to use "hacks" to fix the issues that rise from it

I have to say I disagree with the approach taken.
I think the son_management (not the task and not the DB) should handle the operation of `get supported channels` and then call the `set_channel_scan_pool` function with the channel_pool comprised of those channels.
This way you don't change the `set_channel_scan_pool` function at all and the `son_management` is handling the user request as it should.

itayx:
> > @tomereli
> > I don't think that creating a new API just for the "scan on all channels" is the correct course of action,
> > Instead I reworked the PR to do the following:
> > In the `set_channel_scan_pool` I expanded the signature to accept a "scan all channels" option (set by default to `false`) which, if set to `true` will fill the DCS channel pool with the radio's supported channels.
> > That way the DCS task will be unaware if the client uses the "scan all channels" feature, thus containing the changes only to the DB's `set_channel_scan_pool` function and its uses in the `son_management`.
> > I believe this is the better course of action since it allows the DB to the the sole manager of the DCS's channel pool, and we won't need to use "hacks" to fix the issues that rise from it
> 
> I have to say I disagree with the approach taken.
> I think the son_management (not the task and not the DB) should handle the operation of `get supported channels` and then call the `set_channel_scan_pool` function with the channel_pool comprised of those channels.
> This way you don't change the `set_channel_scan_pool` function at all and the `son_management` is handling the user request as it should.

I thought about it, but it will result in duplicated code for the single scan and the continuous scan.

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -470,6 +470,7 @@ enum class eChannelScanOpErrCode : uint8_t {
 };
 
 #define CHANNEL_SCAN_INVALID_PARAM -1
+#define SCAN_ALL_CHANNELS 0
```
tomereli, line 474: I really don't like this (did I say this already?). I still think that splitting APIs is better, but OK, we can do this in a separate PR I guess.

controller/src/beerocks/master/son_management.cpp
```diff
@@ -1917,7 +1917,10 @@ void son_management::handle_bml_message(Socket *sd,
             op_error_code == eChannelScanOpErrCode::CHANNEL_SCAN_OP_SUCCESS) {
             auto channel_pool_set =
                 std::unordered_set<uint8_t>(channel_pool, channel_pool + channel_pool_size);
-            op_error_code = database.set_channel_scan_pool(radio_mac, channel_pool_set, false)
+            auto scan_all_channels =
```
tomereli: So ugly...

tomereli (approved):
Approving cause its a bug fix, but I think this is a very bad change and we should consider splitting the API for better readability.

controller/src/beerocks/master/son_management.cpp
itayx: Updated the code, and moved the logic to a helper function in the son_management

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
```diff
@@ -52,54 +52,42 @@ beerocks::eChannelScanStatusCode dynamic_channel_selection_task::dcs_request_sca
...
+    request->scan_params().radio_mac         = m_radio_mac;
+    request->scan_params().dwell_time_ms     = dwell_time_msec;
+    request->scan_params().channel_pool_size = curr_channel_pool.size();
+    std::copy(curr_channel_pool.begin(), curr_channel_pool.end(),
+              request->scan_params().channel_pool);
+
     // get the parent node to send the CMDU to the agent
-    auto ire = database.get_node_parent_ire(radio_mac_str);
+    auto radio_mac_str = tlvf::mac_to_string(m_radio_mac);
+    auto ire           = database.get_node_parent_ire(radio_mac_str);
```
adam1985d, line 94: need to add a check here:
if we have for some reason `get_node_parent_ire` return an empty string - we don't check this and perform `send_cmdu_to_agent_ with empty std::string

controller/src/beerocks/master/son_management.cpp
```diff
@@ -30,6 +30,34 @@ using namespace beerocks;
...
+//////////////////////////////////////////////////////////////////////////////
+
+/**
+ * @brief Set the channel pool to contain all supported channels.
+ * @param channel_pool_set:  Set containing the current channel pool.
+ * @param channel_pool_size: Size of the given set.
+ * @param database:          Reference to the DB.
+ * @param radio_mac:         MAC address of radio.
+ */
+static void set_pool_to_all_supported_channels(std::unordered_set<uint8_t> &channel_pool_set,
```
adam1985d: instead of `set_pool_to_all_supported_channels` I would suggest a `get_pool_of_all_supported_channels` which is what you do.

controller/src/beerocks/master/son_management.cpp
```diff
@@ -30,6 +30,34 @@ using namespace beerocks;
...
+                                               uint8_t &channel_pool_size, db &database,
+                                               const sMacAddr &radio_mac)
+{
+    LOG(DEBUG) << "Setting channel pool to all channels";
+    channel_pool_set.clear();
+    auto all_channels = database.get_hostap_supported_channels(tlvf::mac_to_string(radio_mac));
```
adam1985d, line 60: `get_hostap_supported_channels` can fail and return an empty string `std::vector<beerocks::message::sWifiChannel>()`
please add a check here.

this means the function should return `bool` and not `void`

controller/src/beerocks/master/son_management.cpp
```diff
@@ -1950,8 +1978,12 @@ void son_management::handle_bml_message(Socket *sd,
         }
         if (channel_pool != nullptr && channel_pool_size != CHANNEL_SCAN_INVALID_PARAM &&
             op_error_code == eChannelScanOperationCode::SUCCESS) {
+            auto u_channel_pool_size = uint8_t(channel_pool_size);
             auto channel_pool_set =
                 std::unordered_set<uint8_t>(channel_pool, channel_pool + channel_pool_size);
+            if (channel_pool_size == 1 && *channel_pool_set.begin() == SCAN_ALL_CHANNELS)
```
adam1985d: I understand why @tomereli  doesn't like the `*channel_pool_set.begin() == SCAN_ALL_CHANNELS)`
if we are going to "learn to live with it" - at least wrap it with a meaningful naming function:
```
bool is_scan_all_channels_request(channel_pool_set_type &channel_pool_set){
    return ((channel_pool_set == 1) &&*channel_pool_set.begin() == SCAN_ALL_CHANNELS));
}
```
or something similar

adam1985d (approved):
looks much better and cleaner now that you moved it to the son_managemet.
please fix my comments - approving

controller/src/beerocks/master/son_management.cpp
itayx: Agree, much better name,
Changed

controller/src/beerocks/master/son_management.cpp
itayx: The helper function actually makes the code look much better.
Changed

controller/src/beerocks/master/son_management.cpp
itayx: Even if `get_hostap_supported_channels` fails and we get an empty `std::vector<beerocks::message::sWifiChannel>()` it won't matter since later on in the DCS task we validate the pool with
```
if (curr_channel_pool.empty()) {
        LOG(ERROR) << "empty channel pool is not supported. please set channel pool for mac="
                   << m_radio_mac;
        return beerocks::eChannelScanStatusCode::INVALID_PARAMS;
}
```
So even if an empty pool is set, the scan will fail with an `INVALID_PARAMS` in `m_last_scan_error_code`

controller/src/beerocks/master/tasks/dynamic_channel_selection_task.cpp
itayx: Added check, in case `ire` returns as an empty string the DCS will reach the `Abort` state with an `INTERNAL_FAILURE` status

controller/src/beerocks/master/son_management.cpp
adam1985d: we shouldn't care - the DCS task should not depend on the way we decide to implement the son_management and vise versa - or to be more precise - the dependency on the implementation should be minimal - each one should do its own part without depending on the other to verify for it.

son_management should verify it doesn't set and invalid parameter and then responds to the user "all params are ok" and the task should not send a scan request with invalid params.

adam1985d (changes requested):
please see my comment on the last commit

controller/src/beerocks/master/son_management.cpp
itayx: You're right,
Updated

controller/src/beerocks/master/son_management.cpp
```diff
@@ -1952,6 +1994,11 @@ void son_management::handle_bml_message(Socket *sd,
             op_error_code == eChannelScanOperationCode::SUCCESS) {
             auto channel_pool_set =
                 std::unordered_set<uint8_t>(channel_pool, channel_pool + channel_pool_size);
+            if (is_scan_all_channels_request(channel_pool_set))
```
adam1985d: 1. you didn't add `{}` which is against our coding convention
2. in case of error you are still setting the DB! this part should be fixed similar to what you did in the `beerocks_message::ACTION_BML_CHANNEL_SCAN_START_SCAN_REQUEST` handling

adam1985d (changes requested):

adam1985d (approved):

PR #1285, opened 2020-05-12
agent: backhaul_manager: Implement AP Metrics Query/Response message part 2

VladyslavTupikin:
The second part should contain implementation for methods for creation tlvAssociatedStaLinkMetrics and tlvAssociatedStaTrafficStats payloads and update for ap_metrics_query_message in the monitor_thread for using 2 new methods in Policy Config enabled and the same implementation in the ap_metrics_response message for the backhaul manager.

PR #1288, opened 2020-05-13
tests: extend combined_infra_metrics test to cover ap metrics query

arnout:
Use the check_cmdu functionality to validate that the ap metrics responses are correct.

While we're at it, check the combined infra metrics as well.
Turns out they are wrong - an issue will be created for it.

Lots of supporting changes to the sniffer infrastructure.

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -179,6 +179,28 @@ class TestFlows:
             return None
         return tlvs[0]
 
+    def check_cmdu_has_tlvs_exact(
+        self, packet: Union[sniffer.Packet, None], tlvs: [sniffer.Tlv]
+    ) -> None:
+        '''Check that the CMDU has exactly the TLVs given.'''
+        if not packet:
+            return False
+        if not packet.ieee1905:
+            self.fail("Packet is not IEEE1905: {}".format(packet))
+            return False
+        packet_tlvs = list(packet.ieee1905_tlvs)
+        result = True
+        for t in tlvs:
+            if t in packet_tlvs:
+                packet_tlvs.remove(t)
+            else:
+                self.fail("Packet misses tlv:\n {}".format(str(t)))
+                result = False
+        if packet_tlvs:
+            self.fail("Packet has unexpected tlvs:\n {}".format("\n ".join(map(str, packet_tlvs))))
+            result = False
+        return result
+
     def run_tests(self, tests):
         '''Run all tests as specified on the command line.'''
         total_errors = 0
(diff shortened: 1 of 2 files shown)


VladyslavTupikin (approved):

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <beerocks/tlvf/beerocks_message.h>
 #include <beerocks/tlvf/beerocks_message_monitor.h>
+#include <tlvf/wfa_map/tlvApMetrics.h>
```
rmelotte: It looks like this commit comes from https://github.com/prplfoundation/prplMesh/pull/1285
Is this PR going to be rebased on 1285 after 1285 is merged?

tests/test_flows.py
```diff
@@ -133,6 +133,52 @@ def check_cmdu_type_single(
...
+        '''Like check_cmdu_has_tlvs, but also check that only one TLV of that type is found.'''
+        tlvs = self.check_cmdu_has_tlvs(packet, tlv_type)
+        if not tlvs:
+            return None
+        if len(tlvs) > 1:
+            self.fail("More than one ({}) TLVs of type 0x{:02x} found".format(len(tlvs), tlv_type))
+            debug("  {}".format(packet))
+            return None
+        return tlvs[0]
+
```
rmelotte, line 181: To me these 2 functions clearly look like something the Packet class should be doing (`Packet.has_tlvs` and `Packet.has_tlv_single` or something like that).
Is there a specific reason to have them here?

tests/test_flows.py
```diff
@@ -553,32 +553,99 @@ def test_link_metric_query(self):
...
+                check_tlv_equal(combined_2, ap_metrics_2)
+            elif combined_2 == ap_metrics_1:
+                check_tlv_equal(combined_1, ap_metrics_2)
+            else:
+                self.fail("Combined metrics not identical to individual ones")
+                debug("Combined: {}".format(ap_metrics_combined))
+                debug("AP metrics 1: {}".format(ap_metrics_1))
+                debug("AP metrics 2: {}".format(ap_metrics_2))
+
+        def check_equal_tlv(tlv_type, expected, msg):
```
rmelotte: The combined naming of `check_equal_tlv` and `check_tlv_equal` is really confusing.
What about something like `check_has_tlv` for `check_equal_tlv`?

rmelotte (approved):
I didn't review the commit "agent: backhaul_manager: Implement AP Metrics Query/Response message" (4f77d2757bc53bcc6b28657862c1880539e63a73) , I will review it in the other PR (https://github.com/prplfoundation/prplMesh/pull/1285/commits)
Edit: scratch that, it looks like this PR has this one included:  https://github.com/prplfoundation/prplMesh/pull/1231/
So it just needs to be rebased on master AFAICS.

tests/test_flows.py
arnout: The idea is that the "check" style functions are part of the test_flows class (which should really be a base class that other classes derive from, but that's for later). If we strip down the has_tlvs function to something that could be part of the Packet class, it would just be `return [tlv for tlv in packet.ieee1905_tlvs if tlv.tlv_type == tlv_type]` - and we'd still need a check function in test_flows.py to do the actual check (i.e. to indicate failure if the result is empty).

tests/test_flows.py
arnout: I have instead added a function check_cmdu_has_tlvs_exact that does the job in a much more concise and readable way.

PR #1289, opened 2020-05-13
[4.7.7] starting measurements automatically - Monitor (patch 2/3)

vitalii-komisarenko:
Because of its size PR #1139 needs to be split.
Description: https://github.com/prplfoundation/prplMesh/pull/1139#pullrequestreview-398271772

This PR reflects changes in monitor subsystem.

This PR depends on https://github.com/prplfoundation/prplMesh/pull/1290 in order to work properly.


agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -53,6 +53,8 @@ monitor_thread::monitor_thread(const std::string &slave_uds_, const std::string
 
     using namespace std::placeholders; // for `_1`
 
+    network_utils::linux_iface_get_mac(bridge_iface, al_mac);
```
tomereli: You can mention in the commit message why you need the al_mac in the monitor and that this is a preparative commit.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -816,7 +814,7 @@ bool monitor_thread::handle_cmdu_vs_message(Socket *sd, ieee1905_1::CmduMessageR
 
         sta_node = mon_db.sta_add(sta_mac, vap_id);
         sta_node->set_ipv4(sta_ipv4);
-        sta_node->set_bridge_4addr_mac(set_bridge_4addr_mac);
+        sta_node->set_bridge_4addr_mac(al_mac);
```
tomereli: Please also remove `bridge_4addr_mac` from `sClientMonitoringParams` and all its use throught the code.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -668,6 +668,37 @@ bool monitor_thread::update_ap_stats()
...
+bool monitor_thread::start_monitoring_sta(const std::string &sta_mac, const std::string &sta_ipv4,
+                                          const int vap_id)
+{
+    auto sta_node = mon_db.sta_find(sta_mac);
+    if (sta_node) {
+        mon_db.sta_erase(sta_mac);
+    }
+
+    auto vap_node = mon_db.vap_get_by_id(vap_id);
+    if (vap_node == nullptr) {
```
tomereli:
```suggestion
    if (!vap_node) {
```

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1789,6 +1789,24 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
     } break;
+    case Event::STA_Connected: {
+        LOG(TRACE) << "Received STA_Connected event";
+        const auto msg = static_cast<bwl::sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION *>(data);
+
```
tomereli, line 1796: empty line break

tomereli (approved):

agent/src/beerocks/monitor/monitor_thread.h
```diff
@@ -95,6 +95,11 @@ class monitor_thread : public beerocks::socket_thread {
     Socket *mon_hal_nl_events  = nullptr;
     beerocks::logging &logger;
 
+    /**
+     * @brief bridge mac address
+     */
+    std::string al_mac;
```
morantr: Use "m_" prefix for new class members.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -830,13 +840,21 @@ bool monitor_thread::handle_cmdu_vs_message(Socket *sd, ieee1905_1::CmduMessageR
             return false;
         }
 
-        bool success = start_monitoring_sta(sta_mac, sta_ipv4, set_bridge_4addr_mac, vap_id);
+        bool success = start_monitoring_sta(sta_mac, vap_id);
+        if (success) {
+            success = monitored_sta_set_ipv4(sta_mac, sta_ipv4);
```
morantr: This commit is not needed. There is already a handler for IP update on the monitor:
`ACTION_MONITOR_CLIENT_NEW_IP_ADDRESS_NOTIFICATION`.

No need to create a function for that. It is already being done on the handler.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -1582,6 +1582,19 @@ bool slave_thread::handle_cmdu_platform_manager_message(
...
+            if (!monitor_notification) {
+                LOG(ERROR) << "Failed building message!";
+                return false;
+            }
+
+            monitor_notification->mac()  = notification->mac();
+            monitor_notification->ipv4() = notification->ipv4();
+
+            message_com::send_cmdu(monitor_socket, cmdu_tx);
+
```
morantr: It will not work. And now I remember why we did it through the controller.
The DHCP event will be received only on the GW which is where the DHCP server runs on.

So in order to make it work it is required to make an arp table query and get the IP from there.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -828,7 +828,7 @@ bool monitor_thread::handle_cmdu_vs_message(Socket *sd, ieee1905_1::CmduMessageR
         std::string sta_ipv4 = network_utils::ipv4_to_string(request->params().ipv4);
         int vap_id           = int(request->params().vap_id);
         LOG(INFO) << "ACTION_MONITOR_CLIENT_START_MONITORING_REQUEST=" << sta_mac
-                  << " ip=" << sta_ipv4 << " vap_id=" << vap_id;
+                  << " ip=" << sta_ipv4 << " vap_id=" << std::dec << static_cast<int>(vap_id);
```
morantr: This commit is not needed:
1. vap_id is already an `int` - "int vap_id = int(request->params().vap_id);" (two lines before).
2. Not need to add `std::dec`, easylogging set it before or after (not really important) print.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1789,6 +1789,24 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
     } break;
+    case Event::STA_Connected: {
+        LOG(TRACE) << "Received STA_Connected event";
+        const auto msg = static_cast<bwl::sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION *>(data);
+
+        const auto mac    = network_utils::mac_to_string(msg->mac);
+        const auto vap_id = msg->vap_id;
```
morantr: Declaring it as const is not needed. Please remove `const`.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1789,6 +1789,24 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
 
         message_com::send_cmdu(slave_socket, cmdu_tx);
     } break;
+    case Event::STA_Connected: {
+        LOG(TRACE) << "Received STA_Connected event";
+        const auto msg = static_cast<bwl::sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION *>(data);
```
morantr: Declaring it as const is not needed. Please remove `const`.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: I go over the other commits, and I didn't notice it serves any purpose changing it here and on the other follow-up commits. Also, it doesn't right to call it al_mac here, it is needed for arp_monitor where the bridge interface required.
al_mac does not logically have a single bridge mac.
So all of the commits related to changing bridge to al_mac should be dropped.

morantr (changes requested):

agent/src/beerocks/monitor/monitor_thread.cpp
arnout: Indeed, the "bridge_4addr_mac" is a brittle stack of undocumented hacks that shouldn't be changed...

Bottom line is that we still have to set it with a VS message, together with the ipv4 address. They are used together.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Indeed. So instead of start_monitoring, it should be a message that sets a station's IP address and bridge MAC address.

arnout: Just to be clear: the ACTION_DHCP_LEASE_ADDED_NOTIFICATION will only "happen" on the agent of the gateway, because that has the DHCP server. That agent sends it to the controller, then the controller updates the IP address in the database and sends it to all agents (not only the local one).

morantr: The controller sends it to the agent that the client is connected to.
But maybe the Agent of the GW could send it in multicast. That should work.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1806,6 +1806,33 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
...
     case Event::STA_Connected: {
         LOG(TRACE) << "Received STA_Connected event";
+        auto msg = static_cast<bwl::sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION *>(data);
+
+        auto sta_mac = tlvf::mac_to_string(msg->mac);
+        auto vap_id  = msg->vap_id;
+
+        LOG(INFO) << "STA_Connected: mac=" << sta_mac << " vap_id=" << vap_id;
+
+        mon_db.sta_erase(sta_mac);
```
arnout, line 1808: There's a bit a complicated race condition here...

The STAs IPv4 etc. will come from the controller via the "old flow". That old flow is triggered by the STA-CONNECTED event on the ap_manager. So, it is possible that the flow triggered by the ap_manager's STA-CONNECTED completes first (i.e. the STA was already added to the monitor through the VS message), and afterwards this function is triggered and removes the STA again. This way, we'll use the other information (IPv4 etc.) that is set by the VS flow.

Bottom line: I think we first have to do mon_db.find_sta to get the existing node, and copy all the info from it to the new node.

Or maybe it is easier to solve this issue by removing the STA-CONNECTED entirely from ap_manager and move it into monitor.

This can be done later, however, just create an issue for it.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1806,6 +1806,33 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
...
+
+        auto vap_node = mon_db.vap_get_by_id(vap_id);
+        if (!vap_node) {
+            LOG(ERROR) << "vap_id " << vap_id << " doesn't not exists";
```
arnout:
```suggestion
            LOG(ERROR) << "vap_id " << vap_id << " doesn't exist";
```
(nitpick)

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -801,19 +801,13 @@ bool monitor_thread::handle_cmdu_vs_message(Socket &sd, ieee1905_1::CmduMessageR
...
 
         auto sta_node = mon_db.sta_find(sta_mac);
-        if (sta_node) {
-            mon_db.sta_erase(sta_mac);
-        }
-
-        auto vap_node = mon_db.vap_get_by_id(vap_id);
-        if (vap_node == nullptr) {
-            LOG(ERROR) << "vap_id " << vap_id << " doesn't not exists";
+        if (!sta_node) {
```
arnout, line 804: So, due to the race condition I mentioned, the sta_node may actually not exists yet.

Or, it is possible that the STA has already disconnected by the time this message arrives. In the latter case, responding with error is actually OK... Meh, race conditions are hard...

arnout (approved):

agent/src/beerocks/monitor/monitor_thread.cpp
tomereli: Why isn't const needed here? I get why simple types like `vap_id` from above should not need const, but here its a pointer...

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: Yes, this could be a problem.
But the solution is simple. Just get the ipv4 of the station before erasing the node, and set it after creating it.
I thought about using the sta_find() on mon_db, to just update the vap_id, but I'm not sure about it. A lot of parameters are getting default values when creating the node from scratch.
So I guess saving the IP before removing the node, and then updating it on new node is a better idea.

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -801,19 +801,13 @@ bool monitor_thread::handle_cmdu_vs_message(Socket &sd, ieee1905_1::CmduMessageR
...
+            LOG(ERROR) << "sta " << sta_mac << " hasn't been found on mon_db";
```
morantr: I don't understand why do we need that message?
ipv4 is updated externally from the controller now, and `bridge_4addr_mac` from what I see on association handling task is filled only when the station is an extender and it is filled with the bridge mac of the station (a platform which is a station).

It s not critical, but it could be removed. Wasn't it the purpose of this PR and its friends?

One more thing.
The commit message headline is too long "agent: monitor: supply bridge_4addr_mac and ipv4 by START_MONITORING_……REQUEST",
and there is a typo on the commit message: 
"It does not trigger measurements per _**se**_, it supplies missing data"

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1806,10 +1800,44 @@ bool monitor_thread::hal_event_handler(bwl::base_wlan_hal::hal_event_ptr_t event
...
     case Event::STA_Connected: {
         LOG(TRACE) << "Received STA_Connected event";
+        auto msg = static_cast<bwl::sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION *>(data);
+
+        auto sta_mac = tlvf::mac_to_string(msg->mac);
+        auto vap_id  = msg->vap_id;
+
+        LOG(INFO) << "STA_Connected: mac=" << sta_mac << " vap_id=" << vap_id;
+
+        mon_db.sta_erase(sta_mac);
```
vitalybu, line 1808: Why do we need to erase the STA anyway?
If it's already in the DB, why not simply move it to the correct vap?
This logic can be added to `sta_add` method.

morantr (approved):
I have small comments that could be fixed easily.
Although I still think this PR has more work to be done (removing start_monitoring message), since it should work, I approve it.

agent/src/beerocks/monitor/monitor_thread.cpp
morantr: When we erase and add it again, all node class members are set to default values.
This is the only thing I could think of.

arnout: Yeah, I think it's better to keep that behaviour.

PR #1290, opened 2020-05-13
 [4.7.7] starting measurements automatically - BWL (patch 1/3)

vitalii-komisarenko:
Because of its size PR #1139 needs to be split.
Description: https://github.com/prplfoundation/prplMesh/pull/1139#pullrequestreview-398271772

This PR reflects changes to the BWL subsystem.

### DIFF
--- a/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
+++ b/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
@@ -217,6 +217,11 @@ typedef struct {
 
 typedef struct {
     sMacAddr mac;
+    int8_t vap_id;
+} sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION;
+
+typedef struct {
+    sMacAddr mac;
     sMacAddr source_bssid;
     sMacAddr target_bssid;
     uint8_t status_code;
--- a/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
+++ b/common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
@@ -222,6 +222,10 @@ typedef struct {
 
 typedef struct {
     sMacAddr mac;
+} sACTION_MONITOR_CLIENT_DISCONNECTED_NOTIFICATION;
+
+typedef struct {
+    sMacAddr mac;
     sMacAddr source_bssid;
     sMacAddr target_bssid;
     uint8_t status_code;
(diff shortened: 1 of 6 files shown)


common/beerocks/bwl/include/bwl/mon_wlan_hal.h
```diff
@@ -26,6 +26,8 @@ class mon_wlan_hal : public virtual base_wlan_hal {
     enum class Event {
         Invalid = 0,
 
+        STA_Connected,
```
tomereli, line 29: Extra line break

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -215,6 +215,11 @@ typedef struct {
     sClientAssociationParams params;
 } sACTION_APMANAGER_CLIENT_ASSOCIATED_NOTIFICATION;
 
+typedef struct {
```
tomereli, line 218: Please include a Doxygen comment for each new class/struct.

tomereli (approved):
Very nice split to commits!

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
morantr: We really should ban the use of "typdef". On cpp it is not needed.

morantr (commented):
Too many commit IMO.
I would squash commits called "add connect/disconnect events to bwl monitor":
beerocks: bwl: add STA_Connected event to mon_wlan_hal
beerocks: bwl: add STA_Disonnected event to mon_wlan_hal 
beerocks: bwl: add sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION
beerocks: bwl: add sACTION_MONITOR_CLIENT_DISCONNECTED_NOTIFICATION

and squash all other commits as "add handlers for connect/disconnect events on all bwl_mon type".

morantr (approved):

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
arnout: Both comments are correct; however:
- the whole file currently has no doxygen comments;
- it's pretty obvious what it does/means, so a doxygen comment doesn't help much;
- the whole file currently has typedefs;
- this will most likely be refactored for M3.

Therefore, we're not going to bother with fixing this.

arnout:
Issues are fixed and it should now pass in CI, so marking non-draft and ready for merge.

PR #1292, opened 2020-05-13
Hotfix/add correct operational class to steering request

kantera800:
up until now BSS_TM_REQ was sent with  operational class = 0
 This is wrong and some clients reject steering requests because of that.

Add oper_class parameter to function signature and use in BSS_TM_REQ message .


agent/src/beerocks/slave/ap_manager_thread.cpp
```diff
@@ -808,7 +808,8 @@ bool ap_manager_thread::handle_cmdu(Socket *sd, ieee1905_1::CmduMessageRx &cmdu_
         LOG(DEBUG) << "CLIENT_BSS_STEER (802.11v) for sta_mac = " << sta_mac
                    << " to bssid = " << target_bssid
                    << " channel = " << int(request->params().target.channel);
-        ap_wlan_hal->sta_bss_steer(sta_mac, target_bssid, request->params().target.channel,
+        ap_wlan_hal->sta_bss_steer(sta_mac, target_bssid, request->params().target.operating_class,
```
tomereli, line 811: Why can't this be squashed to the previous commit?

tomereli (approved):

agent/src/beerocks/slave/ap_manager_thread.cpp
kantera800: sure , just after the 20 commits in my last pr i felt i had to split something :)

arnout (approved):
The merge commit shouldn't be there, of course.

It's funny that certification test 4.8.1 passes even without this change. I guess the CTT STA doesn't even look at the operating class...

PR #1295, opened 2020-05-14
.gitlab-ci.yml: add ubuntu xenial to the registry

tomereli:
Commit 1635545c11b2b4ada215b51341aedab590967e41 added support for
building ubuntu xenial (16.04) as part of CI, but mistakingly used
docker.io registry.
Fix this now, by adding it to the images build jobs in .gitlab-ci.yml and
adding dependency on it for the build-in-docker-ubuntu-xenial so it will
be automatically built and pushed to the the gitlab registry as part of
CI.

While we're at it, move the Dockerfile to ubuntu/xenial and add a
Dockerfile for ubuntu bionic (18.04). 
Use this one instead of the top level ./tools/docker/builder/Dockerfile - it was
originally designed to support multiple base images, but  it is not
realistic to support all variants in a single Dockerfile since packages
may differ just like with completely different distributions.

Now that we don't use a Dockerfile with build argument to select the base image, remove this option from `image-build.sh`.


tools/docker/builder/ubuntu/bionic/Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM ubuntu:18.04
```
vitalybu, line 1: Why not use the same Dockerfile with a build argument controlling the version?

rmelotte: Although some package names might be the same between the 2 versions, others aren't.
So IMO it's really easier and clearer to have 2 different Dockerfile for the 2 different versions from the start.

tomereli: Yes, this is the point for this change.

arnout (commented):
It would have been better to add the bionic Dockerfile in the second commit which actually uses it.

Also, it would have been a good opportunity to align the bionic and xenial Dockerfiles (now there are a few pointless differences between them).

More importantly, though: you effectively remove the prplmesh-builder image, but there are still two scripts that use it: build.sh and clang-format.sh (in tools/docker). Oh, and now I notice that I made the same mistake... cppcheck.sh still refers to prplmesh-builder:alpine-3.11.3 from dockerhub instead of prplmesh-builder-alpine from gitlab. Could you fix the latter while you're at it?

Since this last comment is a somewhat bigger one, but someone else can do the final review as well as I can, I'm not approving nor requesting changes.

tools/docker/image-build.sh
```diff
@@ -59,16 +59,18 @@ main() {
...
-        run docker image build \
-            --tag "$full_image" \
-            "${scriptdir}/${image}" || exit $?
-        info "Generated $full_image"
+        for tag in "${tags[@]}"; do
+            full_image="${DOCKER_REGISTRY}prplmesh-$image_fixed$postfix$tag"
+            info "Generating $image docker image ($full_image)"
+            run docker image build \
+                --tag "$full_image" \
+                "${scriptdir}/${image}" || exit $?
```
arnout, line 67: It's a bit weird to do the full docker build again when there are multiple tags, but since it's going to use the cache anyway, I guess it's OK.

.gitlab-ci.yml
```diff
@@ -13,7 +13,7 @@ stages:
     - docker-build
   script:
     - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
-    - ./tools/docker/image-build.sh --image "${CI_JOB_NAME#image-build-}" --tag $CI_PIPELINE_ID --push
+    - ./tools/docker/image-build.sh --image "${CI_JOB_NAME#image-build-}" --tag $CI_PIPELINE_ID --tag latest --push
```
arnout: This is not good. We *don't* want to push the latest tag for every pipeline run for any random non-reviewed branch. So this should be done only for master. Since we don't do cppcheck on this, we can code it as:
```suggestion
    - if [ "$CI_COMMIT_BRANCH" = "master" ]; then latest_tag='--tag latest'; fi
    - ./tools/docker/image-build.sh --image "${CI_JOB_NAME#image-build-}" --tag $CI_PIPELINE_ID $latest_tag --push
```

arnout (approved):
Thing about latest tag is important but easy to fix.

arnout:
I needed to add an extra commit to add gcovr and valgrind to the bionic image (they were added to the builder image, but the builder image is removed by this PR).

#1368 updates the xenial image, so this PR really needs to be merged.

vitalybu (approved):

vitalybu:
@mergifyio rebase

PR #1296, opened 2020-05-14
Dev/transport reliable multicast

tomereli:
- Add reliable multicast support in the transport
- move topology notification on station connect disconnect generation from son slave to backhaul
- Enhance steering dummy test to cover the topology notification and reliable multicast correctness
- Add 4.11.1_ETH to the passing tests


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1986,6 +1988,41 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<sRadioInfo>
...
+            // Add vendor specific tlv
+            auto vs_tlv =
+                message_com::add_vs_tlv<beerocks_message::tlvVsClientAssociationEvent>(cmdu_tx);
+
+            if (!vs_tlv) {
+                LOG(ERROR) << "add_vs_tlv tlvVsClientAssociationEvent failed";
+                return false;
+            }
+
+            vs_tlv->mac()          = msg->client_mac();
```
vitalybu: `tlvClientAssociationEvent` already contains the `client_mac` and the `bssid`, why do we pass it again in the VS TLV?

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -21,6 +22,44 @@ namespace mapf {
...
+    if (packet.ether_type != ETH_P_1905_1) {
+        MAPF_DBG("Ignoring non-1905 packet");
+        return;
+    }
+    if (packet.src_if_type != CmduRxMessage::IF_TYPE_NET) {
+        MAPF_DBG("Ignoring packet from src_if_type " << int(packet.src_if_type));
+        return;
+    }
+
+    // Delete aged neighbors
```
vitalybu, line 36: This shouldn't happen for every received packet.
Can be performed every 30/35 seconds (half of the aging duration).
Otherwise, large number of neighbours may have negative impact on packet processing time.

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -21,6 +22,44 @@ namespace mapf {
...
+        Ieee1905Transport::ieee1905_neighbor neigh = {0};
+        neigh.al_mac                               = packet.src;
+        neigh.if_index                             = packet.src_if_index;
+        MAPF_INFO("Adding new neighbour with almac " << packet.src);
+        neighbors_map_[packet.src] = neigh;
+    }
+
+    // Update last seen for new / updated neighbours
+    MAPF_DBG("Updating last seen for neighbour with almac " << packet.src);
+    neighbors_map_[packet.src].last_seen = std::chrono::steady_clock::now();
```
vitalybu:
```suggestion
    // Add new neighbours
    auto neighbor_iter = neighbors_map_.find(packet.src);
    if (neighbor_iter == neighbors_map_.end()) {
        Ieee1905Transport::ieee1905_neighbor neigh = {0};
        neigh.al_mac                               = packet.src;
        neigh.if_index                             = packet.src_if_index;
        neigh.last_seen                           = std::chrono::steady_clock::now();
        MAPF_INFO("Adding new neighbour with almac " << packet.src);
        neighbors_map_[packet.src] = neigh;
    } else {
        // Update last seen for new / updated neighbours
        MAPF_DBG("Updating last seen for neighbour with almac " << packet.src);
        neighbor_iter.second.last_seen = std::chrono::steady_clock::now();
    }
```

vitalybu (changes requested):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: The VS TLV is there to extend the standard TLV. Since there can be multiple tlvClientAssociationEvent TLVS in the same CMDU, we have to include the client mac and bssid also in the VS TLV so we can have a 1 to 1 mapping from VS TLV to non-VS tlv.

tomereli: This made me see that we will fail topology handling for non-prplMesh agents --> added a commit wihch fixes this - https://github.com/prplfoundation/prplMesh/pull/1296/commits/ea4c559192efeaf5cc276feaae473fc70d69e0e3

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
tomereli: I disagree. We are doing the same worst case `O(n)` search 2 lines below in order to update the `last_seen` or add a new neighbor, so if we end up not deleting neighbors, its a linear complexity.

Most importantly - this is an EasyMesh home network. The number of neighbors is actually the number of other repeaters connected directly to this multiap agent, and this should be a very small number - mininum one neighbor, which is the backhaul, normal cascaded will be 2.
Much more than that doesn't make any sense. 

Bottom line - I don't think packet processing time will be impacted at all, and doing this as it is now is much simpler and easier to maintain.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
vitalybu: But each `tlvClientAssociationEvent` will always be followed by `tlvVsClientAssociationEvent` (which should contain only the extended information).
Still don't understand why we need these fields in both TLVs.

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
vitalybu: There's a different between a search with a worst case of `O(n)` and iterating over the entire list which is a constant `O(n)`.
I guess the number of neighbors depends on the architecture of the mesh deployment.
Nothing prevents one from deploying a network in a star topology where all (or most) repeaters are connected to the gateway over a wired backhaul. Since we don't enforce any limitations, theoretically we can end up with thousands of neighbors.
But, since I'm not sure if that's a real life scenario, I won't insist on this one :)

vitalybu (approved):

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -21,6 +21,45 @@ namespace mapf {
...
+        auto age = std::chrono::duration_cast<std::chrono::seconds>(
+            std::chrono::steady_clock::now() - neigh.second.last_seen);
+        if (age > kMaximumNeighbourAge) {
+            MAPF_INFO("Deleting aged out neighbour with almac " << neigh.first);
+            neighbors_map_.erase(neigh.first);
+        }
+    }
+
+    // Add new neighbours
+    if (neighbor_iter == neighbors_map_.end()) {
```
adam1985d, line 51: what is `neighbot_iter`? where is it defined?

adam1985d: I see it is added in the next commit - maybe a stub should have been created in prev. commit
nitpick

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -577,6 +579,64 @@ bool Ieee1905Transport::forward_packet(Packet &packet)
...
+
+    // Reliable multicast - send to all known neighbours.
+    // Avoid loops by skipping the neighbor which sent the message
+    // Note - we are reusing the same packet, so this part has to come
+    // after the original packet (relayed multicast) has been sent.
+    // Change the packet destination each time to the neighbour AL MAC
+    Ieee1905CmduHeader *ch = (Ieee1905CmduHeader *)packet.payload.iov_base;
+    ch->SetRelayIndicator(false);
+    for (const auto &neigh : neighbors_map_) {
+        if (neigh.second.al_mac == packet.src || packet.src_if_index == neigh.second.if_index) {
```
adam1985d: would be more readable if on both conditions the `neigh` was left-side and `packet` right-side (or the other way around)
nitpick

morantr:
Why do you move messages right now between son_slave and the backhaul manager?
I'm going to change soon anyway as part of the unified agent, and this create more work for me.

adam1985d (commented):
spelling mistake in commit message 
https://github.com/prplfoundation/prplMesh/pull/1296/commits/2381bc33aefca0d1106dd5421b100c7
`disassocuate` should be disassociate

adam1985d (approved):
small nitpick and commit message comments, but approved

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -536,6 +579,64 @@ bool Ieee1905Transport::forward_packet(Packet &packet)
...
+        MAPF_DBG("Failed to forward packet " << packet);
+        return false;
+    }
+
+    // Check if this packet should also go through reliable multicast:
+    // 1. Destination is multicast
+    // 2. Message type is topology notification, which is currently the only
+    //    message which should be sent using reliable multicast. See Multi-AP
+    //    specification Table 5
+    auto should_send_reliable_multicast = [&](const Packet &packet) -> bool {
```
kantera800: why create a lamda when you use it only once?

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
```diff
@@ -154,6 +154,21 @@ class Ieee1905Transport {
...
     friend std::ostream &operator<<(std::ostream &os, const Packet &m);
 
+    // reliable multicast neighbours map
+    struct ieee1905_neighbor {
+        sMacAddr al_mac;
+        unsigned int if_index;
+        std::chrono::steady_clock::time_point last_seen;
+    };
+
+    std::unordered_map<sMacAddr, ieee1905_neighbor> neighbors_map_;
```
arnout, line 164: Considering how this structure is used, a priority_queue would be more appropriate. You'd need to implement operator< on ieee1905_neighbor, of course.

Expiry would then just be calling `top` and `pop` in a loop.

And for insert/update, you'd use `find_if`. Note that unordered_map is not a hash table, so complexity of find_if on std::vector is the same as find on unordered_map.

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
tomereli: For better readability

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
```diff
@@ -154,6 +154,21 @@ class Ieee1905Transport {
...
+
+    // According to the ieee1905.1 specification section 8.2.1.1, a 1905.1 management entity shall
+    // transmit a topology discovery message *at least* once every 60 seconds. Since any message from
+    // the neighbor will update it's last_seen, in practice it should be safe to consider it dead and
+    // delete it if 70 seconds have passed.
```
arnout, line 169: Except that this is not actually a *neighbour* map, it's a map of known devices, neighbour or multi-hop. In fact, it's mainly the controller that should be in here, and if it's behind another repeater, it's quite likely that we don't hear from it every minute. A prplMesh controller will indeed send a topology query every minute (IIRC), but you can't rely on that.

In practice it's very unlikely that 1905.1 devices appear or disappear, so the timeout can be quite large.

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -577,6 +579,64 @@ bool Ieee1905Transport::forward_packet(Packet &packet)
     return true;
 }
 
+/**
```
arnout, line 582: (nitpick) doxygen comment should be in the header.

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
```diff
@@ -577,6 +579,64 @@ bool Ieee1905Transport::forward_packet(Packet &packet)
...
+    //    specification Table 5
+    auto should_send_reliable_multicast = [&](const Packet &packet) -> bool {
+        Ieee1905CmduHeader *ch = (Ieee1905CmduHeader *)packet.payload.iov_base;
+        return ETHER_IS_MULTICAST(packet.dst.oct) && ch->GetRelayIndicator() &&
+               ntohs(ch->messageType) ==
+                   static_cast<uint16_t>(ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE);
+    };
+    if (!should_send_reliable_multicast(packet)) {
+        return true;
+    }
```
arnout: Same comment as @kantera800 made, but please explain why this is more readable.
```suggestion
    // Check if this packet should also go through reliable multicast, return if not:
    // 1. Destination is multicast
    if (!ETHER_IS_MULTICAST(packet.dst.oct)) {
        return true;
    }
    // 2. Message is relayed multicast
    Ieee1905CmduHeader *ch = (Ieee1905CmduHeader *)packet.payload.iov_base;
    if (!ch->GetRelayIndicator()) {
        return true;
    }
    // 3. Message type is topology notification, which is currently the only
    //    message which should be sent using reliable multicast. See Multi-AP
    //    specification Table 5
    if (ntohs(ch->messageType) != static_cast<uint16_t>(ieee1905_1::eMessageType::TOPOLOGY_NOTIFICATION_MESSAGE)) {
        return true;
    }
```

Also, you forgot one condition: we should only do this if the message was generated locally, not if we're relaying it ourselves. The spec says to duplicate the message as part of the transmission procedure (section 7.3 of 1905.1) while relaying is part of the reception procedure (section 7.6 of 1905.1).

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
arnout: Since this should only be done for locally generated packets, this condition is redundant.

tests/test_flows.py
```diff
@@ -327,8 +391,9 @@ def test_ap_config_bss_tear_down_cli(self):
         # Same test as the previous one but using CLI instead of dev_send_1905
 
         env.beerocks_cli_command('bml_clear_wifi_credentials {}'.format(env.agents[0].mac))
-        env.beerocks_cli_command('bml_set_wifi_credentials {} {} {} {} {}'.format(env.agents[0].mac,
-                                 "Multi-AP-24G-3-cli", "maprocks1", "24g", "fronthaul"))
+        env.beerocks_cli_command('bml_set_wifi_credentials {} {} {} {} {}'
+                                 .format(env.agents[0].mac,
+                                         "Multi-AP-24G-3-cli", "maprocks1", "24g", "fronthaul"))
```
arnout, line 388: What changed here?

tests/test_flows.py
```diff
@@ -201,6 +201,70 @@ def check_cmdu_has_tlvs_exact(
...
+            station event - CONNECTED / DISCONNECTED
+
+        bssid: str
+            bssid Multi-AP Agent BSSID
+
+        Returns:
+        bool
+            True for valid topology notification, False otherwise
+        """
+        multicasts = self.check_cmdu_type("topology notification", 0x1, eth_src)
```
arnout: Since you checkpoint, there should be only a single topology notification, so you can use check_cmdu_type_single instead.

tests/test_flows.py
```diff
@@ -201,6 +201,70 @@ def check_cmdu_has_tlvs_exact(
...
+
+            mid = multicast.ieee1905_mid
+            for eth_dst in neighbors:
+                unicasts = self.check_cmdu_type("topology notification", 0x1, eth_src, eth_dst)
+                matches = [unicast for unicast in unicasts if unicast.ieee1905_mid == mid]
+                if len(matches) != 1:
+                    self.fail("Recieved {} topology notifications to {} with mid {} (should be one)"
+                              .format(len(matches), eth_dst, mid))
+                    return False
+                unicast = matches[0]
```
arnout:
```suggestion
                unicast = self.check_cmdu_type_single("topology notification", 0x1, eth_src, eth_dst, mid)
```

tests/test_flows.py
```diff
@@ -201,6 +201,70 @@ def check_cmdu_has_tlvs_exact(
...
+                if unicast.ieee1905_relay_indicator:
```
arnout: We could actually add that check to check_cmdu_type: if it's not multicast, it should never have the relay indicator.

tests/test_flows.py
```diff
@@ -201,6 +201,70 @@ def check_cmdu_has_tlvs_exact(
...
+                    self.fail("Unicast topology notification should not be relayed")
+                    return False
+
+        # search for requested event
+        debug("Search for event: sta mac={}, bssid={}, event={}".format(sta.mac, bssid, event))
+        matches = [m for m in multicasts if m.ieee1905_tlvs[0].assoc_event_client_mac == sta.mac
+                   and m.ieee1905_tlvs[0].assoc_event_agent_bssid == bssid
+                   and int(m.ieee1905_tlvs[0].assoc_event_flags, 16) == event.value]
```
arnout: Please use check_cmdu_has_tlv_single instead of assuming it's the first one.

tests/environment.py
```diff
@@ -106,9 +109,24 @@ def create():
...
 
+    def connect(self, bssid):
+        '''Simulate station connection'''
+        self.bssid = bssid
+
+    def disconnect(self):
+        '''Simulate station disconnection'''
+        self.bssid = None
+
+    def connected(self, bssid=None) -> bool:
```
arnout: I was going to comment on how I don't like this connect and disconnect methods (they should be private), but then I wasn't able to find how it is used to begin with...

arnout (approved):
As usual, lots of comments, some of them are essential, but nothing that can't be fixed easily.

arnout:
Ah I forgot to mention: the refactoring of the VS stuff is completely unrelated and should really be a separate PR. I haven't actually reviewed that bit.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
tomereli: If we rely on the TLVs order, then we can remove these from the VS TLV. However, parsing will be much harder and less readable. I think that having the (small) duplication here is fine.

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
tomereli: I think it makes more sense using one of the simpler containers such as unordered_map in this case. We don't expect the number of neighbors to be more than 10, and that's also exaggerating. Hence I prefer to keep the current implementation unless I'm mistaken and there could be hundreds or millions of neighbors...

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
tomereli: So do you think I should use a different name? how about ieee1905_known_devices?
Which timeout do you suggest?

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
tomereli: I agree, but since currently the transport doesn't follow this guideline I prefer to align to what is used in the transport. The whole thing is not doxygen compliant, we should fix this --> https://github.com/prplfoundation/prplMesh/issues/1322

framework/transport/ieee1905_transport/ieee1905_transport_packet_processing.cpp
tomereli: Its a matter of taste I guess, but I actually like your suggestion so I'll change it.

tests/test_flows.py
tomereli: Vitaly set something in the VSCODE configuration which runs formatting automatically whenever we save a file, so it decided to format it like this. Since I changed the file in the first place, I can't avoid it.

tests/test_flows.py
tomereli: Correct, but `check_topology_notification` can't rely that the caller will checkpoint nor should it - we can easily support the same without checkpointing, so why not do that?

tomereli: OK

tests/test_flows.py
tomereli: Nice!

tests/test_flows.py
tomereli: Is this what you mean?
```
if eth_dst is None:
            eth_dst = "01:80:c2:00:00:13"
        return self.check_cmdu(msg, lambda packet:
                               packet.eth_src == eth_src and
                               packet.eth_dst == eth_dst and
                               packet.ieee1905_message_type == msg_type and
                               (mid is None or packet.ieee1905_mid == mid) and
                               (eth_dst != "01:80:c2:00:00:13" or
                                packet.ieee1905_relay_indicator is False))
```

tests/environment.py
tomereli: Right, I meant to use them and eventually didn't, removed.

tests/test_flows.py
tomereli: Cause it doesn't work :/

framework/transport/ieee1905_transport/include/mapf/transport/ieee1905_transport.h
arnout: My point was: priority_queue is the simpler container for this use case. I doubt that it would be more efficient because (like most containers) it's implemented on top of std::vector.

And what I haven't made clear in my comment: there's no need to change the code now, I'm just pointing out that in the future you might consider a priority_queue for this kind of use case.

tests/test_flows.py
arnout: I was thinking more in check_cmdu itself:
```
for packet in result:
    if packet.ieee1905:
        if packet.eth_dst != "01:80:c2:00:00:13" and packet.ieee1905_relay_indicator:
            self.fail("Non-multicast CMDU with relay indicator set: {}".format(packet))
```

Could even be done on the full capture instead of the result, but then you'd get the same error several times most likely.

PR #1299, opened 2020-05-14
Basic integration of googletest

arnout:
This PR adds the basic integration of googletest in our build system:
- in CMakeLists.txt;
- in the builder docker image.

After this, it *should* (untested) be sufficient to follow https://github.com/google/googletest/blob/master/googletest/docs/primer.md to set up a test.
To link it, use:
```
if (GTest_FOUND)
    add_executable(testapp testapp.cpp)
    target_link_libraries(testapp GTest::gtest_main)
    install(TARGETS testapp DESTINATION bin/tests)
    add_test(testapp testapp)
endif()
```

We currently don't have a test executor yet, so the test has to be called manually.
Add the following to .gitlab-ci.yml:
```
testapp:
  extends: .run-test-in-docker
```

Note: the above should be added to README.md once it has been validated that it actually works.

Note that gmock requires more investigation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,21 @@ elseif (TARGET_PLATFORM STREQUAL "linux")
     set(BEEROCKS_BH_WIRE_IFACE "eth0")
 endif()
 
+# Test support
+option(BUILD_TESTS "build multiap unit tests" OFF)
+if (BUILD_TESTS)
+    enable_testing()
+    message(STATUS "Tests are enabled")
+    # GTest is optional. gtest-based test have to check GTest_FOUND (which implies BUILD_TESTS)
+    # and link with GTest::gtest GTest::gtest_main GTest::gmock GTest::gmock_main as appropriate
+    # CMake-bundled FindGTest doesn't support gmock, so force the one installed by GTest itself.
+    find_package(GTest CONFIG)
+    if (NOT GTest_FOUND)
+        message(STATUS "GTest not found, most tests are disabled")
+    endif()
+endif()
+
+
 ## Components
 
 add_subdirectory(framework)
--- a/tools/docker/builder/Dockerfile
+++ b/tools/docker/builder/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y \
     curl \
     gcc \
     git \
+    googletest \
     libjson-c-dev \
     libncurses-dev \
     libnl-genl-3-dev \
--- a/framework/CMakeLists.txt
+++ b/framework/CMakeLists.txt
@@ -1,9 +1,6 @@
 project(multiap_framework C CXX)
 
-enable_testing()
-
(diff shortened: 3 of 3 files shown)


tomereli (approved):

tools/docker/builder/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y \
     curl \
     gcc \
     git \
+    googletest \
```
mariomaz, line 21: Installed version is 1.8 whilst latest available version is 1.10
Version 1.10 includes a new way of creating mocks which is easier

tools/docker/builder/Dockerfile
```diff
@@ -18,6 +18,7 @@ RUN apt-get update && apt-get install -y \
...
     libjson-c-dev \
```
mariomaz, line 22: We should install also valgrind and gcovr packages

CMakeLists.txt
```diff
@@ -146,6 +146,21 @@ elseif (TARGET_PLATFORM STREQUAL "linux")
...
 
+# Test support
+option(BUILD_TESTS "build multiap unit tests" OFF)
+if (BUILD_TESTS)
+    enable_testing()
+    message(STATUS "Tests are enabled")
+    # GTest is optional. gtest-based test have to check GTest_FOUND (which implies BUILD_TESTS)
+    # and link with GTest::gtest GTest::gtest_main GTest::gmock GTest::gmock_main as appropriate
+    # CMake-bundled FindGTest doesn't support gmock, so force the one installed by GTest itself.
+    find_package(GTest CONFIG)
```
mariomaz, line 157: Maybe it's just a typo, but even if googletest has been installed in docker image, this results in "Could NOT find GTest (missing: GTest_DIR)"

googlemock is installed together with googletest in /usr/src/googletest so I think that if GTest is found, so will be GMock. GoogleTest and GoogleMock used to be released as separate projects, but they are so closely related that maintainers decided to merge them into the same repository.

mariomaz (changes requested):
GTest_FOUND is not set.
If you want, please reassign this issue to me and I'll try to fix it.

tools/docker/builder/Dockerfile
arnout: That's not needed for "basic integration". IMO it doesn't make sense to add them until we actually use them in CI.

CMakeLists.txt
arnout: > Maybe it's just a typo, but even if googletest has been installed in docker image, this results in "Could NOT find GTest (missing: GTest_DIR)"

That's because it actually isn't found - the Ubuntu package only contains the source, not the ("compiled") CMake support files. I apparently only tested on my Fedora.

arnout:
Since the integration will have to be done differently (cfr. Mario's comments), this PR doesn't make much sense.

Issue #1072, opened 2020-04-01
Port test_initial_ap_config to boardfarm for prplWRT device


abelog:
### Description
As part of migration to boardfarm testing, we need to port `test_initial_ap_config` of `test_flows.py` to boardfarm for Netgear RAX40, which is treated as a generic prplWrt device.

### Exit criteria

- [ ] boardfarm plugin has a prplWrt device
- [ ] prplWrt devices exposes all interfaces from environment.py
- [ ] AlEntity for agent is populated with the radios that are present on the device
- [ ] wait_for_log interface is implemented
- [ ] all other interfaces are stubs that cause the test to be skipped when called
- [ ] as part of prplWrt device startup, prplMesh ipk is installed on device
- [ ] bft runs from gitlab-ci - this replaces the test-on-netgear-rax40 job.
- [ ] if necessary, documentation/boardfarm-testing.md is updated to reflect the new reality.

PR #1300, opened 2020-05-15
tests: enviroment.py: add prplWRT entities

abelog:
## Description
This PR introduces basic prpMesh-specific entities according to [UML](https://raw.githubusercontent.com/prplfoundation/prplMesh/master/documentation/images/plantuml/boardfarm_prplMesh_classes.png) and is a part of #1072 task.
As yet there is no config, and introduced classes are not yet used - code is totally untested.
Fixes are expected in separate PR, after example config and documentation about it creation is merged (it also will be done in separate PR).
Goal of this PR is to bring prplWRT entities with enough functionality to pass `InitialApConfig`

## Changes
Add:
- Class `RadioHostapd` 
- Class `ALEntityPrplWrt`
-  Stub class `VirtualAPHostapd` (more functionality isn't required for `InitialApConfig` test)

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
@@ -1,0 +1,69 @@
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+import boardfarm
+from boardfarm.devices import OpenWrtRouter
+from environment import ALEntityPrplWrt, _get_bridge_interface
+from .prplmesh_base import PrplMeshBase
+from sniffer import Sniffer
+
+
+class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
+    """prplWRT burned device with prplMesh installed."""
+
+    model = "prplWRT"
+    prompt = ['root\\@OpenWrt:/#', '/#', '@OpenWrt:/#']
+    wan_iface = "eth1"
+    uboot_eth = "eth0_1"
+    agent_entity = None
+    controller_entity = None
+
+    def __init__(self, *args, **kwargs):
+        """Initialize device."""
+        self.args = args
+        self.kwargs = kwargs
+        config = kwargs.get("config", kwargs)
+        self.docker_network = config.get("docker_network",
+                                         "prplMesh-net-{}".format(self.unique_id))
+        self.role = config.get("role", "agent")
+        if self.role == "controller":
+            self.controller_entity = ALEntityPrplWrt(self.name, is_controller=True)
+        else:
+            self.agent_entity = ALEntityPrplWrt(self.name, is_controller=False)
+
+        self.wired_sniffer = Sniffer(_get_bridge_interface(self.docker_network),
+                                     boardfarm.config.output_dir)
+        self.check_status()
+
+    def check_status(self) -> bool:
+        """Check status of device, return bool to indicate state.
+
+        It is used by boardfarm to indicate that spawned device instance is ready for test
+        and also after test - to insure that device still operational.
+        """
+        self.sendline("/etc/init.d/prplmesh status")
+        match = self.expect(
+                ["OPERATIONAL", self.device.pexpect.EOF, self.device.pexpect.TIMEOUT],
+                timeout=10)
+        if match == 1:
+            return True
+        else:
+            return False
+
+    def isalive(self):
+        """Check if device is alive.
+
+        States that device is operational and its consoles are accessible.
+        """
+        return self.check_status()
+
+    def touch(self):
+        """Keep consoles alive.
+
+        Purpose is to keep consoles active, so they don't disconnect for long running activities.
+        """
+        pass
--- a/tests/environment.py
+++ b/tests/environment.py
@@ -212,6 +212,19 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
         return (False, start_line, None)
 
 
+def _device_wait_for_log(device: None, log_path: str, regex: str,
+                         start_line: int, timeout: int) -> bool:
+    """Waits for log mathing regex expression to show up."""
+    device.sendline("tail -n10 -f {}".format(log_path))
+    match = device.expect(
+            pattern=[regex, device.pexpect.EOF, device.pexpect.TIMEOUT],
+            timeout=timeout)
+    if match == 1:
+        return True
+    else:
+        return False
+
+
 class ALEntityDocker(ALEntity):
     '''Docker implementation of ALEntity.
 
@@ -373,3 +386,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str 
     debug('agent2: {}'.format(agents[1].mac))
     debug('agent2 wlan0: {}'.format(agents[1].radios[0].mac))
     debug('agent2 wlan2: {}'.format(agents[1].radios[1].mac))
+
+
+class ALEntityPrplWrt(ALEntity):
+    """Abstraction of ALEntity in real device."""
+
+    def __init__(self, name: str, device: None, is_controller: bool = False):
+        self.device = device
+        self.name = name
+        self.bridge_name = 'br-lan'
+
+        if is_controller:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_controller.conf'
+        else:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_agent.conf'
+
+        ucc_port = self.command(("grep \"ucc_listener_port\" {} "
+                                "| cut -d'=' -f2 | cut -d\" \" -f 1").format(self.config_file_name))
+
+        device_ip_output = self.command(
+                "ip -f inet addr show {} | head -n 2".format(self.bridge_name))
+        device_ip = re.search(
+            r'inet (?P<ip>[0-9.]+)', device_ip_output.decode('utf-8')).group('ip')
+        self.log_folder = self.command(
+            "grep log_files_path {} | cut -d=\'=\' -f2".format(self.config_file_name))
+
+        ucc_socket = UCCSocket(device_ip, ucc_port)
+        mac = ucc_socket.dev_get_parameter('ALid')
+
+        super().__init__(mac, ucc_socket, installdir, is_controller)
+
+        # We always have two radios, wlan0 and wlan2
+        RadioHostapd(self, "wlan0", device=self)
+        RadioHostapd(self, "wlan2", device=self)
+
+    def command(self, *command: str) -> bytes:
+        """Execute `command` in device and return its output."""
+        self.device.sendline(command)
+        return self.device.read()
+
+    def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
+        """Poll the entity's logfile until it contains "regex" or times out."""
+        program = "controller" if self.is_controller else "agent"
+        return _device_wait_for_log(self.device,
+                                    "{}/beerocks_{}.log".format(self.log_folder, program),
+                                    regex, start_line, timeout)
+
+
+class RadioHostapd(Radio):
+    """Abstraction of real Radio in prplWRT device."""
+
+    def __init__(self, agent: ALEntityPrplWrt, bssid: str, device: None):
+        self.iface_name = bssid
+        ip_output = agent.command("ip -o link list dev {}".format(self.iface_name))
+        mac = re.search(r"link/ether (([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})", ip_output).group(1)
+        self.log_folder = self.command(
+            "grep log_files_path {} | cut -d=\'=\' -f2".format(self.config_file_name))
+        super().__init__(agent, mac)
+
+        VirtualAPHostapd(self, mac)
+        VirtualAPHostapd(self, mac)
+
+    def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
+        ''' Poll the Radio's logfile until it match regular expression '''
+        return _device_wait_for_log(self.device, "{}/beerocks_agent_{}.log".format(
+            self.log_folder, self.iface_name), regex, start_line, timeout)
+
+
+class VirtualAPHostapd(VirtualAP):
+    """Docker implementation of a VAP."""
+
+    def __init__(self, radio: RadioHostapd, bssid: str):
+        super().__init__(radio, bssid)
+
+    def associate(self, sta: Station) -> bool:
+        ''' Associate "sta" with this VAP '''
+        # TODO: complete this stub
+
+    def disassociate(self, sta: Station) -> bool:
+        ''' Disassociate "sta" from this VAP.'''
+        # TODO: complete this stub


tests/environment.py
```diff
@@ -188,6 +188,19 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
         return (False, start_line, None)
 
 
+def _device_wait_for_log(device: None, log_path: str, regex: str,
+                         start_line: int, timeout: int) -> bool:
+    """Waits for log mathing regex expression to show up."""
+    device.sendline("tail -n10 -f {}".format(log_path))
```
arnout, line 218: This assumes that the log line we're looking for appears in the last 10 lines. I think we should get the entire file instead. And we can actually make use of start_line here.
```suggestion
    device.sendline("tail -n +{:d} -f {}".format(start_line + 1, log_path))
```
(note that line numbering is 1-based, not 0-based, hence the +1)

tests/environment.py
```diff
@@ -188,6 +188,19 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
         return (False, start_line, None)
 
 
+def _device_wait_for_log(device: None, log_path: str, regex: str,
```
arnout, line 215: `None` can't be right, it's not a type. If you really expect device to be `None`, it should be `NoneType`. But that can't be right :-) It should actually be PrplMeshPrplWRT, but then you create a circular dependency... So, with nothing better, just drop the type specification.

However, I think it would be better to make this function a member of PrplMeshPrplWRT itself. Also, I think everything defined in this commit should move to prplmesh_prplwrt.py (which, BTW, nicely skates around the circular dependency issue).

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
+        program = "controller" if self.is_controller else "agent"
+        return _device_wait_for_log(self.device,
+                                    "{}/beerocks_{}.log".format(self.log_folder, program),
+                                    regex, start_line, timeout)
+
+
+class RadioHostapd(Radio):
+    """Abstraction of real Radio in prplWRT device."""
+
+    def __init__(self, agent: ALEntityPrplWrt, bssid: str, device: None):
```
arnout, line 439: No need for a `device` parameter here - you can use `self.agent.device`.

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
     debug('agent2 wlan2: {}'.format(agents[1].radios[1].mac))
+
+
+class ALEntityPrplWrt(ALEntity):
+    """Abstraction of ALEntity in real device."""
+
+    def __init__(self, name: str, device: None, is_controller: bool = False):
+        self.device = device
+        self.name = name
+        self.bridge_name = 'br-lan'
```
arnout, line 397: I believe we can get this out of the config file as well. Hardcoding it is bad.

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
+
+        if is_controller:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_controller.conf'
+        else:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_agent.conf'
+
+        ucc_port = self.command(("grep \"ucc_listener_port\" {} "
+                                "| cut -d'=' -f2 | cut -d\" \" -f 1").format(self.config_file_name))
```
arnout, line 405: Personally I'd prefer to do the parsing in Python instead of shell, but that's just me I guess.
```suggestion
        ucc_port_string = self.command('grep "ucc_listener_port" {}".format(self.config_file_name))
        ucc_port, _ = ucc_port_string.split(' = ', 1)
```

tests/environment.py
```diff
@@ -188,6 +188,19 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
         return (False, start_line, None)
 
 
+def _device_wait_for_log(device: None, log_path: str, regex: str,
+                         start_line: int, timeout: int) -> bool:
+    """Waits for log mathing regex expression to show up."""
```
RanRegev, line 217: nitpick 
```suggestion
    """Waits for log matching regex expression to show up."""
```

tests/environment.py
RanRegev, line 218: use long format in scripts. It increases readability and requires less efforts from a newcomer
```suggestion
    device.sendline("tail --lines=10 --follow {}".format(log_path))
```

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
     debug('agent2: {}'.format(agents[1].mac))
     debug('agent2 wlan0: {}'.format(agents[1].radios[0].mac))
     debug('agent2 wlan2: {}'.format(agents[1].radios[1].mac))
+
+
+class ALEntityPrplWrt(ALEntity):
+    """Abstraction of ALEntity in real device."""
+
+    def __init__(self, name: str, device: None, is_controller: bool = False):
```
RanRegev, line 394: Creating an object with an embedded invariant is a bad idea. The source of the problem is in the base class `ALEntity` that behaves _as if_ it is a general entity but in reality has to know if it a controller or agent.
However, we don't have to follow this guideline, nor we have to fix it in the base.
All that is required is to create two classes: one for `controller` and the second for `agent`.
See the comment above.

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
     debug('agent2: {}'.format(agents[1].mac))
     debug('agent2 wlan0: {}'.format(agents[1].radios[0].mac))
     debug('agent2 wlan2: {}'.format(agents[1].radios[1].mac))
+
+
+class ALEntityPrplWrt(ALEntity):
```
RanRegev, line 391:
```suggestion
class ALEntityPrplWrtController(ALEntity):
# and later: class ALEntityPrplWrtAgent(ALEntity):
```
See detailed comment below.

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
+        if is_controller:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_controller.conf'
+        else:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_agent.conf'
+
+        ucc_port = self.command(("grep \"ucc_listener_port\" {} "
+                                "| cut -d'=' -f2 | cut -d\" \" -f 1").format(self.config_file_name))
+
+        device_ip_output = self.command(
+                "ip -f inet addr show {} | head -n 2".format(self.bridge_name))
```
arnout, line 408: `head` is redundant since you do regex search below.

tests/environment.py
RanRegev, line 405:
```suggestion
                                "| cut --delimiter='=' --fields=2 | cut --delimiter\" \" -fields=1").format(self.config_file_name))
```

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
+        return _device_wait_for_log(self.device,
+                                    "{}/beerocks_{}.log".format(self.log_folder, program),
+                                    regex, start_line, timeout)
+
+
+class RadioHostapd(Radio):
+    """Abstraction of real Radio in prplWRT device."""
+
+    def __init__(self, agent: ALEntityPrplWrt, bssid: str, device: None):
+        self.iface_name = bssid
```
arnout, line 440: If it's the interface *name* call it iface_name, not bssid.

If it is the radio MAC address, call is radio_mac or radio_uid, not bssid.

bssid is definitely incorrect here.

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
+        self.device = device
+        self.name = name
+        self.bridge_name = 'br-lan'
+
+        if is_controller:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_controller.conf'
+        else:
+            self.config_file_name = '/opt/prplmesh/config/beerocks_agent.conf'
+
+        ucc_port = self.command(("grep \"ucc_listener_port\" {} "
```
RanRegev, line 404: These kind of commands are _extremely_ hard to maintain.
One important helper is a comment above this command with the expected `grep` output.
After that, to be even more helpful, adding an explanation of how the ucc_port is extracted out of the expected result using the piped commands - would be even better.
Below there is more code that this comment should be applied to. I **didn't** add it to each line.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -0,0 +1,69 @@
...
+        and also after test - to insure that device still operational.
+        """
+        self.sendline("/etc/init.d/prplmesh status")
+        match = self.expect(
+                ["OPERATIONAL", self.device.pexpect.EOF, self.device.pexpect.TIMEOUT],
+                timeout=10)
+        if match == 1:
+            return True
+        else:
+            return False
```
arnout, line 55: (nitpick)
```suggestion
        return match == 1
```

However, I thought the returned index was 0-based, so shouldn't this be
```suggestion
        return match == 0
```
?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, *args, **kwargs):
+        """Initialize device."""
+        self.args = args
+        self.kwargs = kwargs
+        config = kwargs.get("config", kwargs)
+        self.docker_network = config.get("docker_network",
+                                         "prplMesh-net-{}".format(self.unique_id))
+        self.role = config.get("role", "agent")
+        if self.role == "controller":
+            self.controller_entity = ALEntityPrplWrt(self.name, is_controller=True)
```
arnout, line 34: This can't be right, the device argument is missing... I guess this should be
```suggestion
            self.controller_entity = ALEntityPrplWrt(self.name, self, is_controller=True)
```
In that case, there is no need to pass self.name, right? The __init__ function can just get it from the device argument.

arnout (approved):
As usual, lots of small comments. It's architecturally sound so can be merged after fixing up the review comments.

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
+                                "| cut -d'=' -f2 | cut -d\" \" -f 1").format(self.config_file_name))
+
+        device_ip_output = self.command(
+                "ip -f inet addr show {} | head -n 2".format(self.bridge_name))
+        device_ip = re.search(
+            r'inet (?P<ip>[0-9.]+)', device_ip_output.decode('utf-8')).group('ip')
```
RanRegev, line 410: Regular expressions are better be explained explicitly. They are not part of the `common` or `expected` knowledge from the average maintainer. Moreover, there are few dialects of regexp so even if one is familiar with regexp she would still need to look for the specific one (python in this case).

tests/environment.py
```diff
@@ -336,3 +349,83 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
+        # We always have two radios, wlan0 and wlan2
+        RadioHostapd(self, "wlan0", device=self)
+        RadioHostapd(self, "wlan2", device=self)
+
+    def command(self, *command: str) -> bytes:
+        """Execute `command` in device and return its output."""
+        self.device.sendline(command)
+        return self.device.read()
+
+    def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
```
RanRegev, line 428: I searched for the means that I have, as a user of this function, to supply a value for `start_line` - and i couldn't find.
I don't think that there is a way for the user to get a value for it. I also looked at [test_flow.py](https://github.com/prplfoundation/prplMesh/blob/feature/n1072-prplwrt-device/tests/test_flows.py) and couldn't find a use of it.
Normally the interface exposes a way to know where to start the next search, e.g. by returning the last position it was found, or by explicit function call `get_last_searched_line()`

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -0,0 +1,69 @@
...
+        States that device is operational and its consoles are accessible.
+        """
+        return self.check_status()
+
+    def touch(self):
+        """Keep consoles alive.
+
+        Purpose is to keep consoles active, so they don't disconnect for long running activities.
+        """
+        pass
```
RanRegev, line 69: Can you elaborate in the comments how `pass` prevents the disconnection?

RanRegev (approved):

tests/environment.py
arnout: Unfortunately, [long options are non-POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/tail.html#tag_20_125_04) so we can't be sure that the `tail` implementation on target supports them. I believe in the busybox implementation, the long option support is optional.

tests/environment.py
arnout: > All that is required is to create two classes: one for controller and the second for agent.

The problem with that is that we would need to create 6 additional classes, and we would have to use multiple inheritance or duplicate code.

tests/environment.py
arnout: Again, [not POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cut.html)

tests/environment.py
RanRegev: * Duplicated code is a bad sign, so we'll avoid this.
* 6 additional classes does not imply wrong.
* multiple inheritance is a design decision

We _have_ variants, otherwise `is_controller` wouldn't be introduced.
The question is how to _express_ them.
I didn't look deeply into _other implementation options_ just indicated that `is_controller` is a red flag that complicates the code. I think we should implement this variant in a different way.

tests/environment.py
arnout: > Normally the interface exposes a way to know where to start the next search, e.g. by returning the last position it was found

[That's exactly what it does.](https://github.com/prplfoundation/prplMesh/blob/master/tests/environment.py#L187) However, this was never documented because I consider it a horrible interface, and it should be converted to use a [checkpoint mechanism](https://github.com/prplfoundation/prplMesh/blob/master/tests/environment.py#L155).

tests/environment.py
abelog: Fixed in #1358. `start_line` now counted in.

tests/environment.py
abelog: Fixed in #1358. Bridge is `grep`-ed from config file.

tests/environment.py
abelog: Parsing moved to Python primary. Fixed in #1358.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Yes, you are right. Fixed in #1358.

tests/environment.py
abelog: Sadly, but `tail` of `busybox` used in RAX40 images doesn't support `--lines` option. But, i was checking pre-RC1 image of prplWRT on RAX40.

tests/environment.py
abelog: Yes. It's totally redundant.

tests/environment.py
abelog: Fixed in #1358.

tests/environment.py
abelog: `head` is just to get last of IPs, since `ip` may report multiple IPs for same interface. E.g :
```
root@OpenWrt:/# ip -family inet addr show br-lan
13: br-lan: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP qlen 1000
    inet 192.168.1.1/24 brd 192.168.1.255 scope global br-lan
       valid_lft forever preferred_lft forever
    inet 192.168.250.170/24 brd 192.168.250.255 scope global br-lan
       valid_lft forever preferred_lft forever
    inet 172.21.0.245/16 scope global br-lan
       valid_lft forever preferred_lft forever
```

tests/environment.py
abelog: Fixed in #1358.

tests/environment.py
abelog: Resolving, as there is no way for us to use long options, as they are not supported on RAX40.

tests/environment.py
abelog: Resolving, as those options are not supported on RAX40.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: We currently are using UART, which do not have auto-hangup. Hence, no need in implementation of this method.

tests/environment.py
abelog: Resolved in #1358

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Than better to drop `name` argument at all. Fixed in #1358.

tests/environment.py
arnout: I'll repeat my statement: since you do a regex search below, head is redundant. re.search will return the first match, so whether you do the `head` or not, you get the same result.

arnout:
@mergifyio rebase

PR #1301, opened 2020-05-15
tests: sniffer: use dumpcap instead of tcpdump

arnout:
Apparently tcpdump sometimes doesn't capture everything, so switch to dumpcap.

### DIFF
--- a/tests/sniffer.py
+++ b/tests/sniffer.py
@@ -112,15 +112,16 @@ class Sniffer:
         os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
         self.current_outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
         self.checkpoint_frame_number = 0
-        command = ["tcpdump", "-i", self.interface, '-U', '-w', self.current_outputfile,
+        # '-q' avoids the output, which we don't need.
+        command = ["dumpcap", "-i", self.interface, '-q', '-w', self.current_outputfile, "-f",
                    "ether proto 0x88CC or ether proto 0x893A"]
         self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
-        # tcpdump takes a while to start up. Wait for the appropriate output before continuing.
-        # poll() so we exit the loop if tcpdump terminates for any reason.
+        # dumpcap takes a while to start up. Wait for the appropriate output before continuing.
+        # poll() so we exit the loop if dumpcap terminates for any reason.
         while not self.tcpdump_proc.poll():
             line = self.tcpdump_proc.stderr.readline()
             debug(line.decode()[:-1])  # strip off newline
-            if line.startswith(b"tcpdump: listening on " + self.interface.encode()):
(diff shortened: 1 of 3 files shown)


arnout:
Bloody great... Apparently the message "Starting capture" is printed *before* the capture is started, so the first packet may still be lost...

I'll look into using dumpcap instead.

tools/docker/README.md
```diff
@@ -35,9 +35,9 @@ The docker runner image can be used to run prplMesh inside containers using `run
 
 The runner docker allows running multiple containers, for example one with a controller+agent, the other with an agent only:
 The 2 containers are connected using a docker network (bridge), so can
-communicate + sniffed by running `wireshark` / `tcpdump` on the bridge from the host to see 1905 packets.
+communicate + sniffed by running `wireshark` / `tshark` / `tcpdump` on the bridge from the host to see 1905 packets.
 This is what is done by the test scripts in the tests/ directory.
-Therefore they require tcpdump and tshark to be installed in order to work.
+Therefore they require dumpcap and tshark to be installed in order to work.
```
tomereli, line 40: dumpcap is part of the `wireshark-common` package, so perhaps it should be mentioned here.

tools/docker/tests-runner/Dockerfile
```diff
@@ -6,7 +6,6 @@ RUN apk add --update --no-cache \
     python3 \
     grep \
     coreutils \
-    tcpdump \
```
tomereli, line 9: But why do we keep `tshark` and not replace it completely with `wireshark-common`? This also makes me notice that the PR name should be changed - "use dumpcap insteda of tcpdump".

tomereli (approved):
minor comments so approving.

tools/docker/README.md
arnout: tshark package is sufficient because it depends on wireshark-common.

tools/docker/tests-runner/Dockerfile
arnout: Because we still use tshark to parse the pcap file.

VladyslavTupikin (approved):

tests/sniffer.py
```diff
@@ -107,15 +107,16 @@ def start(self, outputfile_basename):
         os.makedirs(os.path.join(self.tcpdump_log_dir, 'logs'), exist_ok=True)
         self.current_outputfile = os.path.join(self.tcpdump_log_dir, outputfile_basename) + ".pcap"
         self.checkpoint_frame_number = 0
-        command = ["tcpdump", "-i", self.interface, '-U', '-w', self.current_outputfile,
+        # '-q' avoids the output, which we don't need.
+        command = ["dumpcap", "-i", self.interface, '-q', '-w', self.current_outputfile, "-f",
                    "ether proto 0x88CC or ether proto 0x893A"]
         self.tcpdump_proc = subprocess.Popen(command, stderr=subprocess.PIPE)
```
RanRegev, line 118: I would consider changing the variable name to `capture_proc` or alike.

RanRegev (approved):

PR #1302, opened 2020-05-16
CI: Port upload_to_owncloud.sh to python

### LINKED ISSUE #1185: [TASK] Port upload_to_owncloud.sh to python
## Description

There is a shell script right now to upload all of the logs generated by the certification tests to owncloud automatically.
As part of our efforts to clean up our scripts and workflow for certification tests, we need this to be refactored into a python script.

## Exit criteria

- [ ] A python script exists that uploads logs to owncloud
- [ ] The functionality of the script mirrors that of the existing shell script
- [ ] The script has been tested, and is verified to work

## Detailed description

Please use https://pypi.org/project/webdavclient3/ to do it.

Suggested approach:
1. Install the python module:
`pip install webdavclient3`
2. Make sure you have the correct credentials (bug Frederik, Tomer or Arnout for them)
3. Make a script that connects to our owncloud and uploads a sample file to a new directory.
The shell script reads the credentials from ~/.netrc, you will need to use a different approach:
```python
from webdav3.client import Client
options = {
 'webdav_hostname': "https://ftp.essensium.com/owncloud/remote.php/dav/files",
 'webdav_login':    "login",
 'webdav_password': "password"
}
client = Client(options)
```
Make sure to test that it works by running everything on your own computer. Keep testing whenever you make a change.
4. Now that you know how everything works, restructure the code so that it's nice and readable, with a good division between functions (documented with docstrings), and prepare a PR. See the CONTRIBUTING.md file for how to do that.
It's also a good idea to move configuration (URLs and pathnames) into its own file,so it's easy to edit.
4b. run `flake8` on your code and fix anything that flake8 complains about `flake8 --config <prplmesh_repo>/.flake8 myscript.py`
5. Submit the PR, and we'll do our best to review it :-)
Don't forget to link the PR to this issue

Galvien:
Quoting from #1185:
 > As part of our efforts to clean up our scripts and workflow for certification tests, we need this to be refactored into a python script.

Therefore, port the `upload_to_owncloud.sh` script to Python3, resulting in a cleaner and more maintainable script.

Closes #1185

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -152,11 +152,10 @@ upload-artifacts:
     - sed --follow-symlinks -i 's/^/machine ftp.essensium.com login prplmesh-robot-ci password /' ~/.netrc
     - |
       echo "Updating the 'latest' folder"
-      # Copying takes a lot of time, and we want the "latest" folder
-      # to be updated "as fast as possible", so we rely on the script
-      # to upload to a temporary directory first, and move it to
-      # "latest" afterwards:
-      ci/owncloud/upload_to_owncloud.sh -v "artifacts/latest" ./build
+      python3 ci/owncloud/upload_to_owncloud.py -v "artifacts/$CI_COMMIT_SHA-$CI_JOB_ID" ./build
+      # Copying takes a lot of time, and we want the "latest" folder to be updated
+      # "as fast as possible", so we make a copy and move it:
+      ci/owncloud/owncloud_definitions.sh move "artifacts/$CI_COMMIT_SHA-$CI_JOB_ID" "artifacts/latest"
   rules:
     - if: '$CI_COMMIT_BRANCH == "master"'
       when: on_success
--- a/ci/owncloud/opts.py
+++ b/ci/owncloud/opts.py
@@ -1,0 +1,1 @@
+../../tests/opts.py
--- a/ci/owncloud/owncloud_options.json
+++ b/ci/owncloud/owncloud_options.json
@@ -1,0 +1,5 @@
+{
+    "webdav_login": "Owncloud username",
+    "webdav_password": "Owncloud Password",
+    "webdav_hostname": "https://ftp.essensium.com/owncloud/remote.php/dav/files/"
+}
--- a/ci/owncloud/upload_to_owncloud.py
+++ b/ci/owncloud/upload_to_owncloud.py
@@ -21,9 +21,9 @@ def get_credentials(options_path=None, force_netrc=True):
     options_path : str, optional
                 path to the options file
 
-    force_netrc : bool
-                flag indicating whether to skip parsing the options
-                file and skip straight to the contents of ~/.netrc
+    force_netrc : bool, optional
+                flag indicating whether to skip parsing the options file and instead directly
+                load the credentials from ~/.netrc
     Returns
     -------
     dictionary with 3 key/value pairs,
--- a/ci/owncloud/upload_to_owncloud.py
+++ b/ci/owncloud/upload_to_owncloud.py
@@ -1,0 +1,3 @@
+from webdav3.client import Client
+from opts import debug, message, opts
+
--- a/ci/owncloud/upload_to_owncloud.py
+++ b/ci/owncloud/upload_to_owncloud.py
@@ -1,3 +1,47 @@
+from os import path
+from getpass import getuser
 from webdav3.client import Client
 from opts import debug, message, opts
+
+def get_path_components(dirname):
+    """
+    get_path_components(dirname)
+
+    Gets a list of all subpaths needed to build a directory, including itself.
+
+    Parameters
+    ----------
+    dirname : str
+        A given directory path
+    Returns
+    -------
+    A python list of path strings
+    For an input 'a/b/c' returns ['a/b/c', 'a/b', 'a']
+    """
+    path_list = [dirname, ]
+    tmp = path.split(dirname)
+    while tmp[0] != "" and not path.ismount(tmp[0]):
+        path_list.append(tmp[0])
+        tmp = path.split(tmp[0])
+    path_list.reverse()
+    return path_list
+
+
+def mkdir_recursively(webdav_client, remote_path):
+    """
+    mkdir_recursively(remote_path)
+
+    Recursively creates all directories needed to create the input path.
+
+    Parameters
+    ----------
+    remote_path: str
+    """
+    pathlist = get_path_components(remote_path)
+    for p in pathlist:
+        try:
+            webdav_client.mkdir(p)
+        except BaseException as e:
+            debug(e)
+
 
--- a/ci/owncloud/upload_to_owncloud.py
+++ b/ci/owncloud/upload_to_owncloud.py
@@ -1,7 +1,45 @@
 from os import path
 from getpass import getuser
+import netrc
 from webdav3.client import Client
+import json
 from opts import debug, message, opts
+
+default_hostname = "https://ftp.essensium.com/owncloud/remote.php/dav/files/"
+
+
+def get_credentials(options_path=None, force_netrc=True):
+    """
+    get_credentials(options_path=None, force_netrc=True)
+
+    Gets the user's default credentials from either an external options file or ~/.netrc
+
+    Parameters
+    ----------
+    options_path : str, optional
+                path to the options file
+
+    force_netrc : bool
+                flag indicating whether to skip parsing the options
+                file and skip straight to the contents of ~/.netrc
+    Returns
+    -------
+    dictionary with 3 key/value pairs,
+    keys are: {webdav_login", "webdav_password","webdav_hostname")
+    """
+    if not force_netrc and path.exists(options_path):
+        with open(options_path) as options_file:
+            options = json.load(options_file)
+            return options
+
+    netrc_file = netrc.netrc()
+    credentials = netrc_file.authenticators("ftp.essensium.com")
+    return {
+        "webdav_login": credentials[0],
+        "webdav_password": credentials[2],
+        "webdav_hostname": default_hostname + credentials[0]
+    }
+
 
 def get_path_components(dirname):
     """
--- a/ci/owncloud/upload_to_owncloud.py
+++ b/ci/owncloud/upload_to_owncloud.py
@@ -1,3 +1,4 @@
+import argparse
 from os import path
 from getpass import getuser
 import netrc
@@ -129,3 +130,24 @@ def upload(webdav_client, is_direct, remote_path, local_path, unique_id=None):
     message("Success")
 
 
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument("--verbose", "-v", action='store_true', default=False,
+                        help="Enable verbosity.")
+    parser.add_argument("--direct", "-d", action='store_true', default=False,
+                        help="Direct upload without first uploading to a temporary directory")
+    parser.add_argument("--url", "-u", action='store', nargs=1,
+                        default="https://ftp.essensium.com/owncloud/remote.php/dav/files/{0}"
+                        .format(getuser()),
+                        help="webDAV URL")
+    parser.add_argument("local-path", nargs=1, type=str, action='store',
+                        help="Local file or folder to upload")
+    parser.add_argument("remote-path", type=str, nargs=1, action='store',
+                        help="Path in the cloud, relative to the user home to upload")
+    args = parser.parse_args()
+    args.local_path = getattr(args, "local-path")[0]
+    args.remote_path = getattr(args, "remote-path")[0]
+    opts.verbose = args.verbose
+    webdav_options = get_credentials()
+    client = Client(webdav_options)
+    upload(client, args.direct, args.remote_path, args.local_path)
--- a/ci/owncloud/upload_to_owncloud.py
+++ b/ci/owncloud/upload_to_owncloud.py
@@ -2,6 +2,7 @@ from os import path
 from getpass import getuser
 import netrc
 from webdav3.client import Client
+from uuid import uuid4
 import json
 from opts import debug, message, opts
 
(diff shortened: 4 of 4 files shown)


ci/owncloud/opts.py
```diff
@@ -0,0 +1 @@
+../../tests/opts.py
```
tomereli, line 1: I'm not sure I like this - why do you need symlink? Why can't you just import where you need opts.py with relative path?

ci/owncloud/owncloud_options.json
```diff
@@ -0,0 +1,5 @@
+{
```
tomereli, line 1: We definitly don't need more files for getting the credentials. Please allow only `~/.netrc`.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,3 @@
+from webdav3.client import Client
+from opts import debug, message, opts
```
tomereli, line 8: https://stackoverflow.com/questions/7505988/importing-from-a-relative-path-in-python
No need for a soft link.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -1,3 +1,47 @@
...
+    Parameters
+    ----------
+    remote_path: str
+    """
+    pathlist = get_path_components(remote_path)
+    for p in pathlist:
+        try:
+            webdav_client.mkdir(p)
+        except BaseException as e:
+            debug(e)
```
tomereli, line 85: Shouldn't we also raise the exception instead of hiding it?

ci/owncloud/upload_to_owncloud.py
```diff
@@ -1,8 +1,46 @@
...
 from getpass import getuser
+import netrc
 from webdav3.client import Client
+import json
 from opts import debug, message, opts
 
+default_hostname = "https://ftp.essensium.com/owncloud/remote.php/dav/files/"
+
+
+def get_credentials(options_path=None, force_netrc=True):
```
tomereli, line 13: As I said, this is redundant. There is absolutely no gain from having 2 different files for fetching credentials.
What you can do is add an optional param `netrc_path` which will default to `~/.netrc`.

tomereli (approved):
have some small comments, but looking good so far assuming it works.

ci/owncloud/upload_to_owncloud.py
Galvien: Not exactly, since I'm pretty sure it might throw a 405 status code if the directory already exists, I got a few of those when running tests, so I did this just in case.
Looking at the source code though, It looks like `mkdir` doesn't raise an exception in the case of a 405, I didn't try running the code without the `try` block.

ci/owncloud/upload_to_owncloud.py
Galvien: Not exactly redundant as the `.netrc` file's hostanme field contains only a partial path, we could store only the `"owncloud/remote.php/dav/files/"` part, but in the shell version it's stored in the same way it was implemented here.

ci/owncloud/owncloud_options.json
Galvien: In your comment on #1185 you mentioned having an ability to support getting credentials from an external file.
By default the script uses `.netrc` credentials, but this leaves the ability to easily import credentials without digging through configurations.

ci/owncloud/upload_to_owncloud.py
Galvien: From some reading I understand that implementing it the proper way would require adjustments to the filesystem (adding `__init__.py` files so that we have a package system, which is also kind of wrong, since what we're using is not package).

Changing the import to `from ...tests.opts import debug, message, opts` will result in an `ImportError` since there is no parent package.

There's a hacky way of doing it but it is highly discouraged by PEP.

A good solution IMO would be to implement a package of python utilities inside the prplMesh repository and accessing it with relative imports. But as it stands now, I would say using a symlink is the method that involves creating fewer files.

ci/owncloud/opts.py
Galvien: See https://github.com/prplfoundation/prplMesh/pull/1302/files#r426802483

tomereli (changes requested):
I completely forgot - this is not complete until `gitlab-ci.yml` is updated to use the new script, so please add a commit which does that.

ci/owncloud/upload_to_owncloud.py
ghost: Also, you should not catch `BaseException` in python generally speaking. Catch `Exception` instead :-)

.gitlab-ci.yml
```diff
@@ -152,11 +152,10 @@ upload-artifacts:
...
     - |
       echo "Updating the 'latest' folder"
-      # Copying takes a lot of time, and we want the "latest" folder
-      # to be updated "as fast as possible", so we rely on the script
-      # to upload to a temporary directory first, and move it to
-      # "latest" afterwards:
-      ci/owncloud/upload_to_owncloud.sh -v "artifacts/latest" ./build
+      python3 ci/owncloud/upload_to_owncloud.py -v "artifacts/$CI_COMMIT_SHA-$CI_JOB_ID" ./build
+      # Copying takes a lot of time, and we want the "latest" folder to be updated
+      # "as fast as possible", so we make a copy and move it:
```
arnout, line 157: This comment should still be before the call to upload_to_owncloud, because it's about exactly that call.

ci/owncloud/owncloud_options.json
arnout: The comment you refer to says:
- it should be yaml, not json;
- it should be called owncloud_options.default.yml;
- if owncloud_options.yml exists, it should be used rather than the default (and it should be in .gitignore);
- credentials should *always* be taken from netrc so we don't have another file where we need to worry about visibility and permissions.

Nowhere does the comment say anything about getting credentials from a different file. It literally says "I disagree about using anything other than ~/.netrc".

ci/owncloud/owncloud_options.json
```diff
@@ -0,0 +1,5 @@
...
+    "webdav_login": "Owncloud username",
+    "webdav_password": "Owncloud Password",
+    "webdav_hostname": "https://ftp.essensium.com/owncloud/remote.php/dav/files/"
+}
```
arnout, line 5: Please make sure the file ends with a single newline (no empty line either).

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+import netrc
+from webdav3.client import Client
+from uuid import uuid4
+import json
+from opts import debug, message, opts
+
+default_hostname = "https://ftp.essensium.com/owncloud/remote.php/dav/files/"
+
+
+def get_credentials(options_path=None, force_netrc=True):
```
arnout, line 13: Please don't implement features that are not used, unless there is a very clear plan to start using them in the near future. In this case, the two arguments are not used, so just remove them.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    get_credentials(options_path=None, force_netrc=True)
+
+    Gets the user's default credentials from either an external options file or ~/.netrc
```
arnout, line 17: Don't repeat the function signature, and put the summary line on the same line as the `"""`.
```suggestion
    """Gets the user's default credentials from either an external options file or ~/.netrc
```

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Parameters
+    ----------
+    options_path : str, optional
+                path to the options file
+
+    force_netrc : bool, optional
+                flag indicating whether to skip parsing the options file and instead directly
+                load the credentials from ~/.netrc
+    Returns
```
arnout, line 27:
```suggestion

    Returns
```

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    -------
+    dictionary with 3 key/value pairs,
+    keys are: {webdav_login", "webdav_password","webdav_hostname")
+    """
+    if not force_netrc and path.exists(options_path):
+        with open(options_path) as options_file:
+            options = json.load(options_file)
+            return options
```
arnout, line 35: Even when reading from netrc, the rest of the options should be taken from options_path.

Also, there should be a default for it.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    options_path : str, optional
+                path to the options file
+
+    force_netrc : bool, optional
+                flag indicating whether to skip parsing the options file and instead directly
+                load the credentials from ~/.netrc
+    Returns
+    -------
+    dictionary with 3 key/value pairs,
+    keys are: {webdav_login", "webdav_password","webdav_hostname")
```
arnout, line 30: The dictionary will contain everything that is in the options file.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    if not force_netrc and path.exists(options_path):
+        with open(options_path) as options_file:
+            options = json.load(options_file)
+            return options
+
+    netrc_file = netrc.netrc()
+    credentials = netrc_file.authenticators("ftp.essensium.com")
```
arnout, line 38: It should take the hostname from the options that were parsed above.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
+import argparse
+from os import path
```
arnout, line 2: This is unusual - normally, you do just `import os` and use `os.path.foo()` in the code. Doing it differently is going to confuse people.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+from getpass import getuser
```
arnout, line 3: Avoid `from x import y` if there's not a good reason for it. Here, getuser is used only once, so `getpass.getuser` is definitely not a problem.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+import netrc
+from webdav3.client import Client
```
arnout, line 5: Here again: just `Client` is not very clear, it could be any client. `webdav3.client.Client` is much clearer.

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+        "webdav_password": credentials[2],
+        "webdav_hostname": default_hostname + credentials[0]
+    }
+
+
+def get_path_components(dirname):
+    """
+    get_path_components(dirname)
+
+    Gets a list of all subpaths needed to build a directory, including itself.
```
arnout, line 50:
```suggestion
    """Gets a list of all subpaths needed to build a directory, including itself.
```

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    Parameters
+    ----------
+    dirname : str
+        A given directory path
+    Returns
+    -------
+    A python list of path strings
+    For an input 'a/b/c' returns ['a/b/c', 'a/b', 'a']
+    """
+    path_list = [dirname, ]
```
arnout, line 61:
```suggestion
    path_list = [dirname]
```

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    Gets a list of all subpaths needed to build a directory, including itself.
+
+    Parameters
+    ----------
+    dirname : str
+        A given directory path
+    Returns
+    -------
+    A python list of path strings
+    For an input 'a/b/c' returns ['a/b/c', 'a/b', 'a']
```
arnout, line 59:
```suggestion
    For an input 'a/b/c' returns ['a', 'a/b', 'a/b/c']
```
At least, as far as I understand the code...

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    get_path_components(dirname)
+
+    Gets a list of all subpaths needed to build a directory, including itself.
+
+    Parameters
+    ----------
+    dirname : str
+        A given directory path
+    Returns
```
arnout, line 56:
```suggestion
        A given directory path

    Returns
```

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    -------
+    A python list of path strings
+    For an input 'a/b/c' returns ['a/b/c', 'a/b', 'a']
+    """
+    path_list = [dirname, ]
+    tmp = path.split(dirname)
+    while tmp[0] != "" and not path.ismount(tmp[0]):
```
arnout, line 63: Since this is used for creating a path on the server, checking whether the local path is a mount point seems weird... Why do you do that?

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+        path_list.append(tmp[0])
+        tmp = path.split(tmp[0])
+    path_list.reverse()
+    return path_list
```
arnout, line 67: This seems overly complicated to me... How about:
```suggestion
    if not path.startswith('/'):
        raise ImplementationError("Only absolute paths are supported")
    components = dirname.split('/')
    pathlist = []
    previous = ''
    # First one is empty string - skip it
    for component in components[1:]:
        previous += '/' + component
        pathlist.append(previous)
    return pathlist
```

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+            webdav_client.mkdir(p)
+        except BaseException as e:
+            debug(e)
+
+
+def upload(webdav_client, is_direct, remote_path, local_path, unique_id=None):
+    """
+    upload(webdav_client, is_direct, remote_path, local_path, unique_id)
+
+    Uploads the content of a directory to a directory in the remote webdav server
```
arnout, line 92:
```suggestion
    """Upload the content of a directory to a directory in the remote webdav server.
```
Summary line should be imperative, not third person. And it should be a full sentence ending with .

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+
+    unique_id : str, optional
+        a uuid to act as the identifier of the uploaded files in the remote directory
+    """
+    if not unique_id:
+        unique_id = str(uuid4())
+    message("""Uploading files from local directory {0} to remote directory {1},\n \
+         username is {2}, remote directory id is: {3}""".format(
+        local_path, remote_path, webdav_client.webdav.options["login"], webdav_client, unique_id))
+    remote_full_path = path.join(remote_path, unique_id)
```
arnout, line 119: I believe that in the original script, the unique_id was only added to the temporary path, not to the final destination? @rmelotte ?

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    mkdir_recursively(remote_full_path)
+    if not path.isdir(local_path):
+        raise FileNotFoundError("Local path location is not a directory")
+    if is_direct:
+        webdav_client.upload_sync(remote_full_path, local_path)
+    else:
+        remote_temp_path = path.join("temp_uploads", unique_id)
+        mkdir_recursively(remote_temp_path)
+        webdav_client.upload_sync(remote_temp_path, local_path)
+        webdav_client.copy(remote_temp_path, remote_full_path)
```
arnout, line 129: Shouldn't this be move rather than copy?

Also, if the remote_full_path already exists, isn't it going to add a new subdirectory under remote_full_path, rather than copying/moving all the individual files and directories?

ci/owncloud/upload_to_owncloud.py
```diff
@@ -0,0 +1,153 @@
...
+    parser.add_argument("local-path", nargs=1, type=str, action='store',
+                        help="Local file or folder to upload")
+    parser.add_argument("remote-path", type=str, nargs=1, action='store',
+                        help="Path in the cloud, relative to the user home to upload")
+    args = parser.parse_args()
+    args.local_path = getattr(args, "local-path")[0]
+    args.remote_path = getattr(args, "remote-path")[0]
+    opts.verbose = args.verbose
+    webdav_options = get_credentials()
+    client = Client(webdav_options)
```
arnout, line 152: (nitpick) I would move the creation of the client inside the upload function, i.e. pass the dictionary as the first argument. That makes it easier to use this function by importing.

PR #1307, opened 2020-05-17
[bugfix] missing supported channels

### LINKED ISSUE #1298: [BUG] Regulatory domain channels are skipped by the supported channels
## Description
When generating the Basics Radio Capabilities, some of the channels that should be supported are not preset. The channels are retrieved from the NL correctly as they appear in the log print of the ap_manager_thread. which would suggest that during the construction of the operating class structure that is sent as part of the WSC autoconfig radio basic capabilities TLV.

As can be seen from the logs (in the comments bellow), even though some channels are supported (for example while in country code `US`, channels: 108 & 112) as they are reported by the `hostapd_cli get_hw_features` and are retrieved by the `ap_manager` before being sent as part of the AP_JOINED message.
Therefore the most likely cause is the generating process of the WSC autoconfig radio basic capabilities.
From our testing the issue is only relevant for countries that use a different operating-class list from the currently used global operating-class list.

To fix this bug, I would suggest to expand the `operating_classes_list` currently used in the `wireless_utils` for constructing the WSC autoconfig radio basic capabilities by nesting the list in a map of country code groups (US|EU|JP|CH) with the global list as the default.

itayx:
When generating the Basics Radio Capabilities, some of the channels that should be supported are not preset. The channels are retrieved from the NL correctly as they appear in the log print of the ap_manager_thread.

But before they are processed by the `add_ap_radio_basic_capabilities` into the WSC autoconfig radio basic capabilities TLV, they are cut short due to the size limiter `SUPPORTED_CHANNELS_LENGTH` which is set to 64.

To fix this issue, the value of `SUPPORTED_CHANNELS_LENGTH` need to be increased

### DIFF
--- a/agent/src/beerocks/slave/son_slave_thread.cpp
+++ b/agent/src/beerocks/slave/son_slave_thread.cpp
@@ -1686,6 +1686,16 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         hostap_params    = notification->params();
         hostap_cs_params = notification->cs_params();
 
+        auto tuple_preferred_channels = notification->preferred_channels(0);
+        if (!std::get<0>(tuple_preferred_channels)) {
+            LOG(ERROR) << "getting preferred channels has failed!";
+            return false;
+        }
+        preferred_channels.clear();
+        preferred_channels.insert(
+            preferred_channels.begin(), &std::get<1>(tuple_preferred_channels),
+            &std::get<1>(tuple_preferred_channels) + notification->preferred_channels_size());
+
         auto tuple_supported_channels = notification->supported_channels(0);
         if (!std::get<0>(tuple_supported_channels)) {
             LOG(ERROR) << "getting supported channels has failed!";
@@ -2321,8 +2331,10 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         }
 
         auto tuple_preferred_channels = response->preferred_channels(0);
-        std::copy_n(&std::get<1>(tuple_preferred_channels), message::SUPPORTED_CHANNELS_LENGTH,
-                    hostap_params.preferred_channels);
+        preferred_channels.clear();
+        preferred_channels.insert(
+            preferred_channels.begin(), &std::get<1>(tuple_preferred_channels),
+            &std::get<1>(tuple_preferred_channels) + response->preferred_channels_size());
 
         // build channel preference report
         auto cmdu_tx_header = cmdu_tx.create(
@@ -2334,7 +2346,7 @@ bool slave_thread::handle_cmdu_ap_manager_message(Socket *sd,
         }
 
         auto preferences =
-            wireless_utils::get_channel_preferences(hostap_params.preferred_channels);
+            wireless_utils::get_channel_preferences(&std::get<1>(tuple_preferred_channels));
 
         auto channel_preference_tlv = cmdu_tx.addClass<wfa_map::tlvChannelPreference>();
         if (!channel_preference_tlv) {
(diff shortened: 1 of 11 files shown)


common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -43,7 +43,7 @@ enum eStructsConsts {
     VERSION_LENGTH            = 16,
     NODE_NAME_LENGTH          = 32,
     IFACE_NAME_LENGTH         = 32 + 4, //need extra 1 byte for null termination + alignment
-    SUPPORTED_CHANNELS_LENGTH = 64,     //support upto # channels, every channel item is 32-bit
+    SUPPORTED_CHANNELS_LENGTH = 80,     //support upto # channels, every channel item is 32-bit
```
kantera800: didn't we agree on 128 ?

itayx: When I expanded the size to 128, it failed the `test_flows` the same way currently the 80 fails with the `test-on-netgear-rax40`

I believe the tests fail due to an issue with the AP_JOINED and M1 messages, as I see (in both cases) that the agent is stuck in a never-ending loop of `STATE_WAIT_FOR_JOINED_RESPONSE`

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -3507,8 +3508,8 @@ bool slave_thread::slave_fsm(bool &call_slave_select)
         }
 
         std::array<beerocks::message::sWifiChannel, beerocks::message::SUPPORTED_CHANNELS_LENGTH>
-            supported_channels_arr;
-        std::copy_n(supported_channels.begin(), supported_channels_arr.size(),
+            supported_channels_arr{{}};
+        std::copy_n(supported_channels.begin(), supported_channels.size(),
```
arnout, line 3531: This is an actual fix, so the "preparative commit" part of the commit message is not really needed. (nitpick, obviously)

common/beerocks/bcl/include/bcl/beerocks_defines.h
arnout: That's probably because sNodeHostap becomes too large. With 128 channels, the preferred channel list in it becomes 1K so it's very easy to reach the maximum TLV size.

The best solution would be to remove preferred channels from sNodeHostap completely and instead use the channel preference flow. But that's more work, I guess.

arnout (approved):
Good enough for me. However, does it really solve the issue? IIRC there are actually 100 supported channels...

common/beerocks/bcl/include/bcl/beerocks_defines.h
kantera800: yes we should find a w/a for now.
@itayx  please  check what can we remove from supported_channels for now that is not used.
also go over sNodeHostap parameters and see if there is a something we can remove.
If nothing can be removed i propose removing them from sNodeHostap and create new VS CMDU only from supported channels.

itayx:
@arnout @kantera800
I've fixed the issue with the message size and fixed an unnoticed bug with `ap_wlan_hal_nl80211::read_preferred_channels`.
I moved the preferred_channels from the `sNodeHostap` to the body of `ACTION_APMANAGER_JOINED_NOTIFICATION` in a dynamically allocated list, then stored the preferred_channels in the son_slave directly (Before it was stored as part of the `sNodeHostap` struct in son_slave).
I've also increased the SUPPORTED_CHANNEL_LENGTH to 128. to it could support twice as many channels as before **(If there is a maximum number of channels that we should use, we should set it to that value)**.
As for the unnoticed bug i mentioned above, the read_preferred_channels never really stored the preferred channels, but this has gone unnoticed because the old fixed-size list would simply use blank sWifiChannel structs.

kantera800 (approved):
Looks good , not sure why we even sent preffered_channels in the first place to the controller.

PR #1313, opened 2020-05-18
controller: optimal-path-task: fix vap bssid on rrm beacon

adam1985d:
changed the VAP used for RRM beaconing to be the VAP that has SSID that
matches current-hostapd-ssid.
added a check of returned value - if returned vap_mac is empty the AP is
skipped and the beacon is not sent.

This PR is a hotfix to an issue observed where on UGW it was seen that optimal_path_task sent beacon-request with VAP-mac of zeros - might be due to VAP-list being corrupted in DB.

we could not reproduce the issue added this fix to improve the robustness of the optimal_path_task.

### DIFF
--- a/controller/src/beerocks/master/tasks/optimal_path_task.cpp
+++ b/controller/src/beerocks/master/tasks/optimal_path_task.cpp
@@ -270,9 +270,18 @@ void optimal_path_task::work()
                             beerocks::BEACON_MEASURE_DEFAULT_ACTIVE_DURATION;
                     }
                     // ap_mac is a radio mac, but we need to request measurment on some vap since radio don't beacon
-                    const std::string vap_mac   = *database.get_hostap_vaps_bssids(ap_mac).begin();
-                    measurement_request.bssid   = tlvf::mac_from_string(vap_mac);
-                    measurement_request.channel = database.get_node_channel(ap_mac);
+                    const std::string vap_mac =
+                        database.get_hostap_vap_with_ssid(ap_mac, current_hostap_ssid);
+                    if (vap_mac.empty()) {
+                        LOG(ERROR) << "Failed to get vap for client beacon request, skipping "
+                                      "measurement for "
+                                   << ap_mac;
+                        iterator_element_counter++;
+                        ++potential_ap_iter;
+                        continue;
+                    }
+                    measurement_request.bssid                  = tlvf::mac_from_string(vap_mac);
+                    measurement_request.channel                = database.get_node_channel(ap_mac);
                     measurement_request.expected_reports_count = 1;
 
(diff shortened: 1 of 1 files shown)


controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -270,9 +270,18 @@ void optimal_path_task::work()
...
-                    measurement_request.bssid   = tlvf::mac_from_string(vap_mac);
-                    measurement_request.channel = database.get_node_channel(ap_mac);
+                    const std::string vap_mac =
+                        database.get_hostap_vap_with_ssid(ap_mac, current_hostap_ssid);
+                    if (vap_mac.empty()) {
+                        LOG(ERROR) << "Failed to get vap for client beacon request, skipping "
+                                      "measurement for "
+                                   << ap_mac;
+                        iterator_element_counter++;
+                        ++potential_ap_iter;
```
tomereli, line 280:
```suggestion
                        potential_ap_iter--;
```

tomereli (commented):
The commit message is stating the obvious but doesn't explain the bug itself. It just explains the solution.
For this reason, I am neither approving nor requesting changes yet.

controller/src/beerocks/master/tasks/optimal_path_task.cpp
adam1985d: the cpp check was the reason for this change - it fails the cpp check otherwise

adam1985d:
> The commit message is stating the obvious but doesn't explain the bug itself. It just explains the solution.
> For this reason, I am neither approving nor requesting changes yet.

I will edit the commit message to include the details from the PR description

tomereli (approved):
Typo in commit message - `request with VAP-mac if zeros` should be `request with VAP-mac of zeros`

adam1985d:
> Typo in commit message - `request with VAP-mac if zeros` should be `request with VAP-mac of zeros`

thanks. fixed

controller/src/beerocks/master/tasks/optimal_path_task.cpp
morantr, line 280: Why one is `x++`and the other one is `++x`?
I would do them both in the form of `x++`.

morantr (approved):

controller/src/beerocks/master/tasks/optimal_path_task.cpp
adam1985d: this is a change due to cpp-check - which failed when the `potential_ap_iter++` was used - since it is an iterator. the only reason it is not "screaming" on the rest of locations - is that cpp-check only runs on the changes you make.

kantera800 (approved):

PR #1317, opened 2020-05-19
Hotfix/fix missing client capabilities

morantr:
@amritak23 has discovered issues on weighted phyrate while testing the platform.
The bug root cause was on the parsing of the client capabilities which occurred the function that generates connect events to a client that are connected to the ap before beerocks has brought up.
Also, there was a bug on the bandwidth enum on the station capabilities report.

### DIFF
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -261,6 +261,8 @@ static void print_sta_capabilities(beerocks::message::sRadioCapabilities &sta_ca
                << std::endl
                << "vht_ss = " << ((int(sta_caps.vht_ss)) ? std::to_string(sta_caps.vht_ss) : "n/a")
                << std::endl
+               << "vht_mcs = "
+               << ((int(sta_caps.vht_mcs)) ? std::to_string(sta_caps.vht_mcs) : "n/a") << std::endl
                << "vht_bw = "
                << ((sta_caps.vht_bw != 0xFF)
                        ? std::to_string(beerocks::utils::convert_bandwidth_to_int(
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -372,14 +372,14 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
     for (uint i = 0; (i < 8) && (i < sizeof(ht_mcs)); i++) {
         ht_mcs_str[2] = ht_mcs[2 * i];
         ht_mcs_str[3] = ht_mcs[2 * i + 1];
-        HT_MCS[i]     = beerocks::string_utils::stoi(ht_mcs_str);
+        HT_MCS[i]     = std::stoul(ht_mcs_str, nullptr, 16);
     }
 
     vht_mcs_str[2] = vht_mcs[0];
     vht_mcs_str[3] = vht_mcs[1];
     vht_mcs_str[4] = vht_mcs[2];
     vht_mcs_str[5] = vht_mcs[3];
-    VHT_MCS[0]     = beerocks::string_utils::stoi(vht_mcs_str);
+    VHT_MCS[0]     = std::stoul(vht_mcs_str, nullptr, 16);
 
     LOG(DEBUG) << "client_mac      : " << client_mac;
     LOG(DEBUG) << "supported_rates : " << std::dec << supported_rates[0]
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -134,7 +134,7 @@ static void get_ht_mcs_capabilities(int *HT_MCS, std::string &ht_cap_str,
     sta_caps.ht_bw        = 0xFF;
 
     if (!ht_cap_str.empty() && (HT_MCS != nullptr)) {
-        uint16_t ht_cap = uint16_t(std::stoul(ht_cap_str, nullptr, 16));
+        uint16_t ht_cap = uint16_t(std::strtoul(ht_cap_str.c_str(), nullptr, 16));
         sta_caps.ht_bw  = (ht_cap & HT_CAP_INFO_SUPP_CHANNEL_WIDTH_SET) != 0; // 20 == 0 / 40 == 1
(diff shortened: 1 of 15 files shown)


controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -420,7 +420,9 @@ void optimal_path_task::work()
             auto hostap_is_5ghz = database.is_node_5ghz(hostap);
             if (!force_signal_strength_decision &&
                 !database.settings_client_optimal_path_roaming_prefer_signal_strength()) {
-                //get sta capabilities....
+                // Get sta capabilities
```
adam1985d, line 432: nitpick:
no need for this comment once you added the log print

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -922,7 +924,9 @@ void optimal_path_task::work()
                 continue;
             }
 
-            //get sta capabilities....
+            // Get STA capabilities
```
adam1985d, line 936: nitpick:
no need for this comment once you added the log print

itayx (approved):

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1064,26 +1064,14 @@ bool db::set_station_capabilities(const std::string &client_mac,
...
-        if (n->m_sta_5ghz_capabilities.cap_flag == true) {
-            //skip setting 5Ghz station parameters if it is already learned once
-            LOG(INFO) << "m_sta_5ghz_capabilities already set once";
-        } else {
-            n->m_sta_5ghz_capabilities          = sta_cap;
-            n->m_sta_5ghz_capabilities.cap_flag = true;
-        }
-
-        n->capabilities = n->m_sta_5ghz_capabilities;
+        n->m_sta_5ghz_capabilities          = sta_cap;
```
adam1985d: is there a reason the original implementation didn't update the capabilities?
is there a side-effect to the set that was meant to be prevented

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1064,26 +1064,14 @@ bool db::set_station_capabilities(const std::string &client_mac,
...
-        if (n->m_sta_24ghz_capabilities.cap_flag == true) {
-            //skip setting 2.4Ghz station parameters if it is already learned once
-            LOG(INFO) << "m_sta_24ghz_capabilities already set once";
-        } else {
-            n->m_sta_24ghz_capabilities          = sta_cap;
-            n->m_sta_24ghz_capabilities.cap_flag = true;
-        }
-
-        n->capabilities = n->m_sta_24ghz_capabilities;
+        n->m_sta_24ghz_capabilities          = sta_cap;
```
adam1985d: same question here

adam1985d (approved):

controller/src/beerocks/master/db/db.cpp
morantr: > is there a reason the original implementation didn't update the capabilities?

Yes, to prevent capabilities struct copy.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1064,26 +1064,14 @@ bool db::set_station_capabilities(const std::string &client_mac,
...
-            //skip setting 5Ghz station parameters if it is already learned once
-            LOG(INFO) << "m_sta_5ghz_capabilities already set once";
-        } else {
-            n->m_sta_5ghz_capabilities          = sta_cap;
-            n->m_sta_5ghz_capabilities.cap_flag = true;
-        }
-
-        n->capabilities = n->m_sta_5ghz_capabilities;
+        n->m_sta_5ghz_capabilities          = sta_cap;
+        n->m_sta_5ghz_capabilities.cap_flag = true;
```
kantera800: as this commit headline  implies you can go ahead and remove cap_flag

morantr: Actually no. When this flag is not set, we return nullptr on capability getter function.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -372,14 +372,14 @@ static std::shared_ptr<char> generate_client_assoc_event(const std::string &even
     for (uint i = 0; (i < 8) && (i < sizeof(ht_mcs)); i++) {
         ht_mcs_str[2] = ht_mcs[2 * i];
         ht_mcs_str[3] = ht_mcs[2 * i + 1];
-        HT_MCS[i]     = beerocks::string_utils::stoi(ht_mcs_str);
```
vitalybu, line 373: There's a reason `string_utils::stoi` was used here and not `std::stoi`.
The `std` versions throw an `invalid_argument` or `out_of_range` exceptions if the input is not valid.
Since we don't catch exceptions, it'll simply crash the process...
If you really need `stoul`, please implement a safe version.

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
```diff
@@ -127,9 +127,11 @@ typedef struct {
 //ap_wlan_hal
 
 enum eWiFiBandwidth : uint8_t {
```
vitalybu, line 129: Why don't we use the `eWiFiBandwidth` from `beerocks_defines`?

controller/src/beerocks/master/db/db.cpp
vitalybu: So it determined the validity of the data?
In that case, maybe should be changed to `valid`

vitalybu (changes requested):

common/beerocks/bwl/include/bwl/base_wlan_hal_types.h
morantr: Good question! And why don't we use the struct from tlvf? They are the same, and we have a duplication for every message.
The reason is that we didn't want to create dependency (or dependency of messages and types, not sure) from other libs such BCL.
This is really stupid IMO, since we already depended on BCL for some utilities.
@tomereli

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
morantr: add a commit that replaces all calls across the code to safe functions.

controller/src/beerocks/master/db/db.cpp
morantr: rename it on a separate commit.

vitalybu (approved):

PR #1323, opened 2020-05-20
Dev/channel pool validation

itayx:


kantera800:
Looks good , not sure why we even sent preffered_channels in the first place to the controller.

PR #1327, opened 2020-05-21
Hotfix/Fix disappearing messages bug

### LINKED ISSUE #1310: [BUG] controller: controller doesn't receive topology notification and cconnected client not seen in bml_conn_map
**Describe the bug** -

The client is connected and son_slave received the `ACTION_APMANAGER_CLIENT_ASSOCIATED_NOTIFICATION` and sends to BH manager the `ACTION_BACKHAUL_CLIENT_ASSOCIATED_NOTIFICATION`, although the son_slave send the `TOPOLOGY_NOTIFICATION_MESSAGE` the controller doesn't receive and doesn't perform the `handle_cmdu_1905_topology_notification` which adds the node of the connected client to the database.

(reproduced on UGW platform but the flow is mutual to all platforms)

## Steps to reproduce

1. configure SSID
2. connect a client
- verify client connected in agent_wlanX log
- on setups running a hostapd, verify the client is connected using the `hostapd_cli -i wlanX stat` (look for `num_sta[1]=1` which indicates number of connected clients to that AP)
3. run bml_conn_map using `/opt/prplmesh/bin/beerocks_cli bml_conn_map`

## Expected behavior
- [ ] controller received `TOPOLOGY_NOTIFICATION_MESSAGE` and sends the `CONNECTION_CHANGE` event to the bml_task
- [ ] bml-connection-map should show the client is connected

attached logs:
[prplmesh_logs.tar.gz](https://github.com/prplfoundation/prplMesh/files/4640428/prplmesh_logs.tar.gz)

morantr:
Commit 2a1c86a3 added a filter that filters out message types that the BTL derived
class is not registered to them on UDS_BUS.
This caused a bug that some messages from son_slave like TOPOLOGY_NOTIFICATION
did not reach their destination (Controller) since they were filtered out.
The reason for that is the fact that indeed TOPOLOGY_NOTIFICATION is not a message
that the Agent should get from the BUS and it is not registered to it. But it is normal
behavior to get such a message from an internal socket such as the son_slave.

To solve this add a socket argument to the filter function so that the function will filter
out only messages that come from the bus and not from internal sockets.

Fixes #1310

### DIFF
--- a/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
+++ b/common/beerocks/bcl/include/bcl/beerocks_socket_thread.h
@@ -42,7 +42,10 @@ protected:
     {
         return false;
     }
-    virtual bool skip_filtered_message_type(ieee1905_1::eMessageType msg_type) { return false; }
+    virtual bool skip_filtered_message_type(Socket *sd, ieee1905_1::eMessageType msg_type)
+    {
+        return false;
+    }
     virtual std::string print_cmdu_types(const message::sUdsHeader *cmdu_header)
     {
         return std::string();
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -185,7 +185,7 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     }
 
     // On UDS bus, unsubscribed messages needs to be filtered manually.
-    if (skip_filtered_message_type(cmdu_rx.getMessageType())) {
+    if (skip_filtered_message_type(sd, cmdu_rx.getMessageType())) {
         return true;
     }
 
--- a/common/beerocks/btl/btl_uds.cpp
+++ b/common/beerocks/btl/btl_uds.cpp
@@ -125,8 +125,14 @@ bool transport_socket_thread::bus_send(ieee1905_1::CmduMessage &cmdu, const std:
 
 bool transport_socket_thread::from_bus(Socket *sd) { return sd == bus; }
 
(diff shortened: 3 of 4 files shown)


kantera800 (approved):

adam1985d (approved):
Testes on UGW(Axepoint) and the issue is resolved.

RanRegev (commented):
Clearly I don't understand the problem.
It feels like we don't control the messages flying around the system.
Messages are getting surprisingly into unexpected socket and instead of checking closely for the reason we just filter them as not being registered.
This mechanism however has problems as this PR now constraints the filtered messages to filter only non expected messages from some sockets (!from_bus).
I don't feel comfortable with the entire mechanism, with the constraint added to it and with the fact that messages are reaching incorrect targets.
I might be missing something completely, so please help me.

morantr:
> This mechanism however has problems as this PR now constraints the filtered messages to filter only non expected messages from some sockets (!from_bus).

The opposite, we filter out only messages that are from the socket of the bus.
The thing is that on normal BUS mode, we have the transport layer that filters for us messages that the process is not registered for.
In the mode of MSGLIB=None, we don't use a real bus, instead, we use a UDS that is connected between the agent (backhaul manager) and the controller (master_thread). This means that we don't have message filtering and that is why we had to add it manualy.

I'll try to give you an example for why we need it.
The Agent should send to the BUS a topology discovery message. The controller is not registered for that message and not handle it. But on UDS_BUS mode, the "bus" is a UDS between the agent and the controller, so the controller receives it although it should not receive it, and print a lot of error messages about receiving a message that it could not handle.
But with the added filter it does not happen anymore.

tomereli (approved):

PR #1330, opened 2020-05-21
tlvf: CmduMessage: move getClass down to CmduMessageRx

arnout:
The getClass and getClassList functions should never be used on a
CmduMessageTx, only on CmduMessageRx. Therefore, move them down into
CmduMessageRx.

Ideally it should also be renamed to getTlv, but that's a bigger change.

This was triggered by my review of #1274 which incorrectly uses getClass on a CmduMessageTx.

### DIFF
--- a/framework/tlvf/src/include/tlvf/CmduMessage.h
+++ b/framework/tlvf/src/include/tlvf/CmduMessage.h
@@ -36,13 +36,6 @@ public:
     uint16_t getMessageId();
     void setMessageId(uint16_t mid);
 
-    // Forward wrapper functions
-    // TODO check which of them can be removed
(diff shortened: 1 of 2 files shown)


mariomaz (approved):
The description of the PR does not explain *why* getClass and getClassList functions should never be used on a CmduMessageTx. According to the names of the methods and the inheritance tree, it's difficult to guess. It might be a smell of a name change is required.

tomereli (approved):

arnout:
> The description of the PR does not explain _why_ getClass and getClassList functions should never be used on a CmduMessageTx

I've updated the commit message with a little bit of explanation.

PR #1332, opened 2020-05-21
docker: openwrt: pass build-arg variables explicitly

arnout:
When the --build-arg option is given without a value for the argument,
docker will use an environment variable with the same name. However,
podman (a "drop-in" replacement for docker with better non-privileged
support) doesn't behave like that. Therefore, give the arguments
explicitly.

### DIFF
--- a/tools/docker/builder/openwrt/build.sh
+++ b/tools/docker/builder/openwrt/build.sh
@@ -30,16 +30,16 @@ usage() {
 build_image() {
     # We first need to build the corresponding images
     docker build --tag "$image_tag" \
-           --build-arg OPENWRT_REPOSITORY \
-           --build-arg OPENWRT_VERSION \
-           --build-arg TARGET_SYSTEM \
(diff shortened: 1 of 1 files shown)


tomereli (approved):

tools/docker/builder/openwrt/build.sh
```diff
@@ -174,6 +174,7 @@ main() {
 
 }
 
+VERBOSE=true
```
rmelotte: This change is not mentioned in the commit message.
I just wanted to mention there was a --verbose option before that set this variable, it has been removed with the shellcheck fixes.
I added the option back in my upcoming PR, but we could also not add the option back and keep this change (as you prefer).

rmelotte (approved):

tools/docker/builder/openwrt/build.sh
arnout: That was actually a debugging leftover that I forgot to remove again...

PR #1333, opened 2020-05-24
[BUGFIX] bml: fix unknown type name during build

itayx:
When building for RDKB, the bml is imported by a C compiler, which
causes the struct `BML_DEVICE_DATA` be ne unrecognized.
The signature  of `bml_device_oper_radios_query` need to updated to be C compatible.


vitalybu (commented):
This doesn't make sense.
You've modified a CPP file, which is most definitely not compiled using a C compiler.

PR #1335, opened 2020-05-25
tools: deploy_firmware.py: wait at least for the CAC

rmelotte:
When deploying the full firmware on the rax40, the device is rebooted.

When the device boots, it takes a lot of time for the wireless
interfaces to become available. We need to make sure they are before
deploying prplMesh and starting tests.

When deploying the firmware on the rax40, wait at least for 60
seconds, which corresponds to the CAC timer.


tomereli (approved):

arnout (approved):
This kind of check really should be part of the prplmesh start script. That said, for now this is a practical workaround.

PR #1336, opened 2020-05-25
bwl: Fix parsing of DFS-CAC-COMPLETED dwpal event

morantr:
The cac completed event string has changed to (dwpal):
"<3>DFS-CAC-COMPLETED wlan2 cac_status=1 freq=5180 ht_enabled=0
chan_offset=0 chan_width=5 cf1=5250 cf2=0 timeout=60"

"success" field is no longer, so it was replaced with "cac_status".

### DIFF
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -2356,7 +2356,7 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         FieldsToParse fieldsToParse[] = {
             {NULL /*opCode*/, &numOfValidArgs[0], DWPAL_STR_PARAM, NULL, 0},
             {NULL, &numOfValidArgs[1], DWPAL_STR_PARAM, NULL, 0},
-            {(void *)&msg->params.success, &numOfValidArgs[2], DWPAL_CHAR_PARAM, "success=", 0},
+            {(void *)&msg->params.success, &numOfValidArgs[2], DWPAL_CHAR_PARAM, "cac_status=", 0},
             {(void *)&msg->params.frequency, &numOfValidArgs[3], DWPAL_INT_PARAM, "freq=", 0},
             {(void *)&msg->params.timeout, &numOfValidArgs[4], DWPAL_INT_PARAM, "timeout=", 0},
             {(void *)&chan_width, &numOfValidArgs[5], DWPAL_CHAR_PARAM, "chan_width=", 0},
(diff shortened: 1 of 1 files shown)


yaniv-yitzhak (approved):

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -2368,10 +2368,10 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
         FieldsToParse fieldsToParse[] = {
             {NULL /*opCode*/, &numOfValidArgs[0], DWPAL_STR_PARAM, NULL, 0},
             {NULL, &numOfValidArgs[1], DWPAL_STR_PARAM, NULL, 0},
-            {(void *)&msg->params.success, &numOfValidArgs[2], DWPAL_CHAR_PARAM, "success=", 0},
+            {(void *)&msg->params.success, &numOfValidArgs[2], DWPAL_INT_PARAM, "cac_status=", 0},
```
arnout: I haven't investigated how dwpal_string_to_struct_parse() is implemented, but I assume the parameter type is used to determine the size of the argument pointer. So if it is changed to int, the type of msg->params.success has to be changed to int (or unsigned) as well. That said, I don't see what was wrong with char.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -2368,10 +2368,10 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
...
             {(void *)&msg->params.frequency, &numOfValidArgs[3], DWPAL_INT_PARAM, "freq=", 0},
             {(void *)&msg->params.timeout, &numOfValidArgs[4], DWPAL_INT_PARAM, "timeout=", 0},
-            {(void *)&chan_width, &numOfValidArgs[5], DWPAL_CHAR_PARAM, "chan_width=", 0},
+            {(void *)&chan_width, &numOfValidArgs[5], DWPAL_INT_PARAM, "chan_width=", 0},
```
arnout: Same here: char was OK.

arnout (approved):
Although I think this PR introduces a bug, it can be easily fixed, so OK to merge after fixing it.

vitalybu:
I am also not sure why the type was changed from CHAR to INT...

morantr:
> I am also not sure why the type was changed from CHAR to INT...

The values on the reply are integers, not strings.

arnout:
> > I am also not sure why the type was changed from CHAR to INT...
> 
> The values on the reply are integers, not strings.

The parameter type is about the type of the return argument, not of what is parsed (although obviously there should be a correlation between the two). To get a string parameter, it would have to be STR, not CHAR.

It's true though that CHAR is not correct either, it should be UNSIGNED_CHAR. But in practice that obviously doesn't make a difference.

yaniv-yitzhak:
The parameter type is for the parser only (look in dwpal.c).
CHAR means byte size, not a string, it can be changed to DWPAL_UNSIGNED_CHAR_PARAM which means value between 0-255.

morantr:
Changed it back to "DWPAL_CHAR_PARAM".

yaniv-yitzhak (approved):

ghost (approved):

PR #1337, opened 2020-05-25
hotfix/Fix VSCODE prplMesh workspace file

morantr:
Some developers and I use VSCODE as our IDE and encountered a problem with the C/CPP extension.
For some reason "go to definition/declaration" did not work properly.
I have found out that if I open the project with the workspace file on "tools/" instead of
"Open Folder" the problem is fixed, but a new problem has shown - The search files (ctrl+P)
was not able to find files that were not recently open.

The solution that fixes it all was to create a new workspace file on the main project folder.
The previous workspace file has been deleted. There were some folder paths that have been
removed since they are no longer exist.


vitalybu:
Why do you need a workspace and not just open the repo as a folder?

morantr:
> Why do you need a workspace and not just open the repo as a folder?

I wrote it on the commit message.

vitalybu:
> > Why do you need a workspace and not just open the repo as a folder?
> 
> I wrote it on the commit message.

It doesn't make sense.
You didn't fix the problem, but found a workaround you think works, but can't explain why...

morantr:
> > > Why do you need a workspace and not just open the repo as a folder?
> > 
> > 
> > I wrote it on the commit message.
> 
> It doesn't make sense.
> You didn't fix the problem, but found a workaround you think works, but can't explain why...

Does it really matter? It allows us to work.
I tried to find why it happens, but I couldn't find why. I assume it is a bug in the extension.

vitalybu:
> Does it really matter? It allows us to work.
> I tried to find why it happens, but I couldn't find why. I assume it is a bug in the extension.

Of course it does. We don't do things we don't understand.
In any case, it works perfectly fine for me, so maybe it's a local configuration issue on your end.

tomereli (approved):

itayx (approved):

PR #1342, opened 2020-05-26
Hotfix/remove disable arp feature

morantr:
The "disable_arp" feature is causing the monitor_rssi to not send an active arp request.
The problem with this feature is that when it is enabled, Intel steering will not work,
neither client rx_rssi measurement on bml.
Remove "disable_arp" flag from monitor_rssi.


kantera800 (approved):
Originally this option was introduced for debug purposes but we don't actually use it.
ARP feature is used to obtain rssi measurment for steering so it should be enabled as long as bandsteering enabled.
Approving since I don't see any scenario that ARP feature should be disabled.

IMO if we want to disable it on some platform in the future we should do it dynamically.

vitalybu (approved):

PR #1343, opened 2020-05-26
hotfix! Fix data filling in add_ap_metrics

VladyslavTupikin:
Add correct values for number_of_stas_currently_associated and
channel_utilization.

### DIFF
--- a/agent/src/beerocks/monitor/monitor_stats.cpp
+++ b/agent/src/beerocks/monitor/monitor_stats.cpp
@@ -435,7 +435,9 @@ void monitor_stats::process()
     }
 }
 
-bool monitor_stats::add_ap_metrics(ieee1905_1::CmduMessageTx &cmdu_tx, const sMacAddr &bssid)
+bool monitor_stats::add_ap_metrics(ieee1905_1::CmduMessageTx &cmdu_tx,
+                                   const monitor_vap_node &vap_node,
+                                   const monitor_radio_node &radio_node) const
 {
     auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetrics>();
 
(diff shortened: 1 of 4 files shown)


abelog (approved):

vitalybu (approved):

VladyslavTupikin:
Added don't merge label. tried to pass 4.7.4 on this branch - failed, need to figure out why

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -432,7 +432,8 @@ void monitor_stats::process()
     }
 }
 
-bool monitor_stats::add_ap_metrics(ieee1905_1::CmduMessageTx &cmdu_tx, const sMacAddr &bssid)
+bool monitor_stats::add_ap_metrics(ieee1905_1::CmduMessageTx &cmdu_tx, const sMacAddr &bssid,
+                                   const size_t &sta_count, const uint8_t &channel_utilization)
```
arnout: These should not be passed by const-ref but by value (const-ref only for objects that are larger than word size).

However, instead of passing in sta_count and channel_utilization, I think it's better to pass in vap_node (for sta_count) and radio_node (for channel utlization).

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -1501,7 +1501,10 @@ bool monitor_thread::handle_ap_metrics_query(Socket &sd, ieee1905_1::CmduMessage
         const auto &bssid = std::get<1>(bssid_tuple);
         LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
                    << "  bssid " << bssid;
-        if (!mon_stats.add_ap_metrics(cmdu_tx, bssid)) {
+        if (!mon_stats.add_ap_metrics(cmdu_tx, bssid, mon_db.get_sta_count(),
```
arnout: mon_db.get_sta_count is incorrect, it should only be the stations associated with the VAP.

arnout (changes requested):
Changes request just to block merging.

agent/src/beerocks/monitor/monitor_stats.cpp
mariomaz: Fixed

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: Fixed

agent/src/beerocks/monitor/monitor_thread.cpp
```diff
@@ -596,18 +596,18 @@ bool monitor_thread::create_ap_metrics_response(uint16_t mid,
             continue;
         }
 
-        if (!mon_stats.add_ap_metrics(cmdu_tx, bssid)) {
+        auto vap_node   = *mon_db.vap_get_by_id(vap_id);
```
arnout: Shouldn't we check that this is not null?

arnout (approved):

agent/src/beerocks/monitor/monitor_thread.cpp
mariomaz: Fixed

PR #1347, opened 2020-05-27
Add unit test for get_operating_class_by_channel()

arnout:
... and fix the issues discovered with it.

This goes on top of #1321, of course.

### DIFF
--- a/common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
+++ b/common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
@@ -1,0 +1,132 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include <bcl/beerocks_utils.h>
+#include <bcl/son/son_wireless_utils.h>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+/** @brief Helper function for testing get_operating_class_by_channel
+ *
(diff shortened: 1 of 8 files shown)


rmelotte (approved):

tomereli (approved):

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -0,0 +1,132 @@
...
+    ch.channel           = channel;
+    ch.channel_bandwidth = beerocks::utils::convert_bandwidth_to_enum(bandwidth);
+    return son::wireless_utils::get_operating_class_by_channel(ch);
+}
+
+// See https://docs.google.com/spreadsheets/d/1J2IYuGjFX_OQQpgb4OgsyDx73klAUk6qSmeB4acJ5us
+// for the table we need to conform to.
+TEST(wireless_utils_test, get_operating_class_by_channel_should_return_valid)
+{
+    EXPECT_EQ(get_operating_class_by_channel_helper(1, 20), 81);
```
mariomaz, line 36: For data-driven testing, use [Value-Parameterized Tests](https://github.com/google/googletest/blob/master/googletest/docs/advanced.md#value-parameterized-tests)

mariomaz (approved):
Approving, but since these initial tests are going to be used by others as a template, I think we should set the foundations for further development by adopting best practices

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
arnout: Thank you! I was looking for exactly that. When I skimmed the documentation, I just read the first two bullets of the "This is useful for..." sentence and I thought "this is not what I need" :-).

I'm still going to merge it as is, and create a follow-up PR that changes the approach. I think the latter can be instructive as well to learn how to convert a table to value parameters.

ghost (approved):

PR #1355, opened 2020-05-28
bwl: fix build bwl "NL80211" for platform "linux"

MuteSpirit:
If run build with BWL_TYPE=NL80211 for TARGET_PLATFORM linux:
```
./tools/docker/build.sh -f BWL_TYPE=NL80211 TARGET_PLATFORM=linux
```
then next compilation errors happen:

```
/home/mutespirit/inango/prplmesh/prplmesh_root/prplMesh/common/beerocks/bwl/shared/netlink_socket.cpp:11:10: fatal error: netlink/msg.h: No such file or directory
     #include <netlink/msg.h>
               ^~~~~~~~~~~~~~~
               compilation terminated.

    /home/mutespirit/inango/prplmesh/prplmesh_root/prplMesh/common/beerocks/bwl/nl80211/base_wlan_hal_nl80211.cpp:20:10: fatal error: wpa_ctrl.h: No such file or directory
     #include <wpa_ctrl.h>
              ^~~~~~~~~~~~
```

Fix bwl CMake file:
* add search hostapd include folders for "linux" target platforms too
* add "nl-3" library search for "NL80211" bwl
* keep requirements for "nl-tiny" for OpenWrt which does not support
  full libnl

Change include <wpa_client.h> to "wpa_client.h" to allow compiler find
it in hostapd directory

**Note** i didn't test build for prplwrt. In progress

### DIFF
--- a/common/beerocks/bwl/CMakeLists.txt
+++ b/common/beerocks/bwl/CMakeLists.txt
@@ -9,8 +9,10 @@ file(GLOB_RECURSE bwl_common_sources ${MODULE_PATH}/common/*.c*)
 # Support more permissive c++ language
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
 
-# Default defaults
-set(BWL_TYPE "DUMMY")
+# Defaults
+if (NOT DEFINED BWL_TYPE)
+    set(BWL_TYPE "DUMMY")
+endif()
 add_definitions(-DBEEROCKS_TMP_PATH="${TMP_PATH}")
 
 if (TARGET_PLATFORM STREQUAL "rdkb")
@@ -27,12 +29,19 @@ elseif(TARGET_PLATFORM STREQUAL "openwrt")
         set(BWL_TYPE "NL80211")
     endif()
 
-    # hostapd directory
-    file(GLOB HOSTAPD_SEARCH_PATHS "${PLATFORM_BUILD_DIR}/hostapd*/hostapd-*")
+endif()
+
+if (NOT BWL_TYPE STREQUAL "DUMMY")
+
+    file(GLOB HOSTAPD_SEARCH_PATHS "${PLATFORM_BUILD_DIR}/hostapd*/hostapd-*" "${PLATFORM_BUILD_DIR}/wpa-*")
     find_path(HOSTAPD_INCLUDE_DIR NAMES "src/common/wpa_ctrl.h" PATHS ${HOSTAPD_SEARCH_PATHS} NO_CMAKE_FIND_ROOT_PATH)
     set(HOSTAPD_DIR "${HOSTAPD_INCLUDE_DIR}")
+    message(STATUS "[IVANE] ${HOSTAPD_DIR}")
 
 endif()
+
+# If it's needed to build prplMesh with BWL_TYPE=NL80211
+# you should pass this variable via cmake cmdline options
 
 set(BWL_TYPE ${BWL_TYPE} CACHE STRING "Which BWL backend to use")
 
@@ -41,7 +50,7 @@ set(BWL_TYPE ${BWL_TYPE} CACHE STRING "Which BWL backend to use")
 ##########################################################################
 
 if(BWL_TYPE STREQUAL "DWPAL")
-    
+
     include_directories(
         ${HOSTAPD_DIR}/src/drivers
         ${PLATFORM_BUILD_DIR}/iwlwav-iw-4.14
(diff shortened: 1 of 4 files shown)


common/beerocks/bwl/CMakeLists.txt
```diff
@@ -9,8 +9,10 @@ file(GLOB_RECURSE bwl_common_sources ${MODULE_PATH}/common/*.c*)
 # Support more permissive c++ language
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
 
-# Default defaults
-set(BWL_TYPE "DUMMY")
+# Defaults
+if (NOT DEFINED BWL_TYPE)
+    set(BWL_TYPE "DUMMY")
+endif()
```
arnout, line 15: This should be approached differently, like this (skeleton code):
```
set(DEFAULT_BWL_TYPE "DUMMY")
if (TARGET_PLATFORM STREQUAL "rdkb")
    set(DEFAULT_BWL_TYPE "DWPAL")
...
endif()
set(BWL_TYPE ${DEFAULT_BWL_TYPE} CACHE STRING "Which BWL backend to use")
```

Also, it should be done in a separate commit from the rest of the changes.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -27,21 +29,28 @@ elseif(TARGET_PLATFORM STREQUAL "openwrt")
         set(BWL_TYPE "NL80211")
     endif()
 
-    # hostapd directory
-    file(GLOB HOSTAPD_SEARCH_PATHS "${PLATFORM_BUILD_DIR}/hostapd*/hostapd-*")
+endif()
+
+if (NOT BWL_TYPE STREQUAL "DUMMY")
```
arnout, line 34: I don't like this - it just happens that the nl80211 and dwpal types both need this, but if another bwl type is added, it will most likely not need the wpa_ctrl stuff...

I think a nicer way to organise this would be to add a new variable NEED_HOSTAPD_DIR (defaults to OFF) and set it in the different BWL_TYPE branches, and then add move this code and the rest of the code that depends on it (adding to sources and includes) after the condition.

And while you're at it: there should be an error if HOSTAPD_INCLUDE_DIR is not set.

Even better would be to use the package approach. But that's more difficult to get right.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -27,21 +29,28 @@ elseif(TARGET_PLATFORM STREQUAL "openwrt")
...
+
+    file(GLOB HOSTAPD_SEARCH_PATHS "${PLATFORM_BUILD_DIR}/hostapd*/hostapd-*" "${PLATFORM_BUILD_DIR}/wpa-*")
     find_path(HOSTAPD_INCLUDE_DIR NAMES "src/common/wpa_ctrl.h" PATHS ${HOSTAPD_SEARCH_PATHS} NO_CMAKE_FIND_ROOT_PATH)
     set(HOSTAPD_DIR "${HOSTAPD_INCLUDE_DIR}")
+    message(STATUS "[IVANE] ${HOSTAPD_DIR}")
```
arnout, line 39: I suppose this should be removed...

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -27,21 +29,28 @@ elseif(TARGET_PLATFORM STREQUAL "openwrt")
...
 
 endif()
 
+# If it's needed to build prplMesh with BWL_TYPE=NL80211
+# you should pass this variable via cmake cmdline options
+
```
arnout, line 45: Err, I don't see how this comment is useful.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -91,19 +100,29 @@ elseif(BWL_TYPE STREQUAL "NL80211")
...
-    include_directories(
+    list(APPEND BWL_INC_DIRS
         ${HOSTAPD_DIR}/src/utils
         ${HOSTAPD_DIR}/src/common
         ${HOSTAPD_DIR}/src/drivers
-        ${PLATFORM_INCLUDE_DIR}/libnl-tiny
     )
+    # OpenWrt support only libnl-tiny
+    if (TARGET_PLATFORM STREQUAL "openwrt")
+        list(APPEND BWL_INC_DIRS ${PLATFORM_INCLUDE_DIR}/libnl-tiny)
```
arnout, line 110: You can simply change this into
```
        include_directories(${PLATFORM_INCLUDE_DIR}/libnl-tiny)
```
without adding BWL_INC_DIRS variable.

However, Moran is making full nl3 a required dependency independent of the target (cfr. #1368) so you can simply remove this libnl-tiny option. In a separate commit, obviously.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -91,19 +100,29 @@ elseif(BWL_TYPE STREQUAL "NL80211")
...
         ${PLATFORM_STAGING_DIR}/usr/lib
     )
 
-    list(APPEND BWL_LIBS nl-tiny)
+    if (TARGET_PLATFORM STREQUAL "openwrt")
+        list(APPEND BWL_LIBS nl-tiny wpa_client)
+    else()
+        find_package(nl-3 REQUIRED)
+        find_package(nl-genl-3 REQUIRED)
+        list(APPEND BWL_LIBS nl-3 nl-genl-3 wpa_client)
```
arnout, line 124: wpa_client should not be added here, but in the NEEDS_HOSTAPD_PATH condition that I mentioned above. You could first check if it exists, like this:
```
find_library(WPA_CLIENT_LIB libwpa_client.so)
if (${WPA_CLIENT_LIB} STREQUAL WPA_CLIENT_LIB-NOTFOUND)
    ... do the hostapd_dir stuff
else()
    list(APPEND BWL_LIBS wpa_client)
endif()
```

common/beerocks/bwl/nl80211/base_wlan_hal_nl80211.cpp
```diff
@@ -17,7 +17,7 @@
 #include <easylogging++.h>
 
 extern "C" {
-#include <wpa_ctrl.h>
+#include "wpa_ctrl.h"
```
arnout, line 20: wpa_ctrl.h doesn't exist in that directory, so why do you change this?

tools/docker/builder/Dockerfile
```diff
@@ -24,7 +24,6 @@ RUN apt-get update && apt-get install -y \
     libnl-genl-3-dev \
     libreadline-dev \
     libssl-dev \
-    libssl-dev \
```
arnout, line 27: Unrelated change, please do it in a separate commit.

arnout (changes requested):
I'll probably want to re-review.

common/beerocks/bwl/CMakeLists.txt
rmelotte: AFAIK instead of: 
`if (${WPA_CLIENT_LIB} STREQUAL WPA_CLIENT_LIB-NOTFOUND)`
you can even do:
`if (NOT WPA_CLIENT_LIB)`

VladyslavTupikin:
Created new PR for fixing this issue: #1530

PR #1356, opened 2020-05-28
prplmesh_utils.sh.in: use BEEROCKS vars instead of hard-coded wlan0/2

MuteSpirit:
Replace hard-coded wlan ifaces names in "prplmesh_utils.sh.in"

  wlan0 -> @BEEROCKS_WLAN1_IFACE@
  wlan2 -> @BEEROCKS_WLAN2_IFACE@

### DIFF
--- a/common/beerocks/scripts/prplmesh_utils.sh.in
+++ b/common/beerocks/scripts/prplmesh_utils.sh.in
@@ -41,11 +41,11 @@ platform_init() {
     control_ip=192.168.250.140
 
     ip link add @BEEROCKS_BRIDGE_IFACE@ address "${base_mac}:00:00" type bridge
-    ip link add wlan0 address "${base_mac}:00:10" type dummy
-    ip link add wlan2 address "${base_mac}:00:20" type dummy
-    ip link set dev wlan0 up
-    ip link set dev wlan2 up
-    for iface in wlan0 wlan2 $DATA_IFACE
+    ip link add @BEEROCKS_WLAN1_IFACE@ address "${base_mac}:00:10" type dummy
+    ip link add @BEEROCKS_WLAN2_IFACE@ address "${base_mac}:00:20" type dummy
+    ip link set dev @BEEROCKS_WLAN1_IFACE@ up
+    ip link set dev @BEEROCKS_WLAN2_IFACE@ up
+    for iface in @BEEROCKS_WLAN1_IFACE@ @BEEROCKS_WLAN2_IFACE@ $DATA_IFACE
     do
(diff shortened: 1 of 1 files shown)


ghost (approved):

arnout:
The platform_init and platform_deinit functions are in fact not used, because we now do this in the Docker entrypoint. It could still be useful when running it "natively". But in that case you probably want different interfaces anyway, and perhaps use an actual wifi interface rather than a dummy interface. So I don't think it's worth keeping it.

In other words, I think we should:
- remove the platform_init and platform_deinit functions;
- remove the -p option;
- remove BEEROCKS_PLATFORM_INIT from CMakeLists.txt.

MuteSpirit:
> In other words, I think we should:
> 
> * remove the platform_init and platform_deinit functions;
> * remove the -p option;
> * remove BEEROCKS_PLATFORM_INIT from CMakeLists.txt.

If you look at it from the other side:
* Functions `platform_init` and `platform_deinit` will be useful for run prplMesh in QEMU VM built by Yocto
* Docker entrypoint also may start use `prplmesh_util.sh`.
  So it'll use actual values of `BEEROCKS_*` variables for up dummy wlan ifaces and bridge.

morantr (approved):

rmelotte (approved):
It probably won't be useful on rdkb on the raspberry pi if real interfaces are available.
However, I think the "-p" option to prplmesh_utils.sh can still be useful when running natively when you don't have actual interfaces.
I'm not against the change itself so approving.

arnout:
> > In other words, I think we should:
> > 
> > * remove the platform_init and platform_deinit functions;
> > * remove the -p option;
> > * remove BEEROCKS_PLATFORM_INIT from CMakeLists.txt.
> 
> If you look at it from the other side:
> 
>     * Functions `platform_init` and `platform_deinit` will be useful for run prplMesh in QEMU VM built by Yocto

In QEMU, IMO it would be better to use nl80211 rather than dummy, and use mac80211_hwsim instead of dummy interfaces. I thought RDK-B QEMU was already set up that way (but I could be wrong).


>     * Docker entrypoint also may start use `prplmesh_util.sh`.
>       So it'll use actual values of `BEEROCKS_*` variables for up dummy wlan ifaces and bridge.

I don't remember why we moved it from prplmesh_utils to the docker entrypoint - history doesn't tell me unfortunately. However, the manipulations of eth0 are very specific to docker, so they would still need to be done. Also, if you use it outside of docker (e.g. in qemu), you probably want to do something slightly different. So really, I don't think there's much point keeping this stuff in prplmesh_utils.sh.

arnout:
Other people disagree with me, so let's merge this.

Issue #1253, opened 2020-05-08
[BUG] Installed prplMesh has non-FHS folders for "linux" platform


MuteSpirit:
By default installed prplMesh folder tree look like:
```
/opt/prplmesh/bin/...
/opt/prplmesh/lib/...
/opt/prplmesh/include/...
/opt/prplmesh/scripts/prplmesh_utils.sh
/opt/prplmesh/share/prplmesh_platform_db
/opt/prplmesh/share/logging.conf
/opt/prplmesh/config/version
/opt/prplmesh/config/*.conf
```

But for `linux` target platform situation is [different](https://github.com/prplfoundation/prplMesh/blob/66899d5bd57ce28c16e7b3fad69f05b6a523376f/CMakeLists.txt#L96)
```
set(INSTALL_PATH "/opt/prplmesh")
set(TMP_PATH "/tmp/beerocks")

if(TARGET_PLATFORM STREQUAL "linux")
    set(INSTALL_PATH "${CMAKE_INSTALL_PREFIX}")
```

And installed prplMesh folder tree look another:
```
/usr/bin/...
/usr/lib/...
/usr/include/...
/usr/scripts/prplmesh_utils.sh
/usr/share/prplmesh_platform_db
/usr/share/logging.conf
/usr/config/version
/usr/config/*.conf
/usr/host
```

As a results next folders are not FHS (File System Hierarchy Standard) compatible:
```
/usr/config
/usr/scripts
```

## Expected behavior

* `/usr/config` -> `/etc/prplmesh`
* `/usr/share/logging.conf` -> `/etc/prplmesh/logging.conf`
* `/usr/scripts/*` -> `/usr/bin/*`
* `/usr/share/prplmesh_platform_db` -> `/var/lib/prplmesh/prplmesh_platform_db`
* no `/usr/host` folder (there is `tlvf.py` script there only and it's needed only on build stage)
* Instead of `INSTALL_PATH` special CMake variables are used:
```
CMAKE_INSTALL_PREFIX
CMAKE_INSTALL_BINDIR
CMAKE_INSTALL_SBINDIR
CMAKE_INSTALL_LIBEXECDIR
CMAKE_INSTALL_SYSCONFDIR
CMAKE_INSTALL_SHAREDSTATEDIR
CMAKE_INSTALL_LOCALSTATEDIR
CMAKE_INSTALL_LIBDIR
CMAKE_INSTALL_INCLUDEDIR
[...]

PR #1357, opened 2020-05-28
cmake: allow redefinition BEEROCKS_* vars from cmdline

MuteSpirit:
If CMake variable defined in CMakeLists.txt file like
```
    set(VARIABLE VALUE)
```
Then it's impossible to redefine it from command line.
Event tricks like next does not work:
```
    cmake ... -UVARIABLE -DVARIABLE=NEW_VALUE
```
So Yocto/RDKB recipes and OpenWrt/prplWrt makefiles cannot
redefine variables without modification CMakeLists.txt files.

As a result conf files for controller/agent/etc. generated with values unusable for RDKB/etc.

Define variables interested for redefinition following pattern:
```
    if(NOT DEFINED VARIABLE)
        set(VARIABLE DEFAUL_VALUE)
    endif()
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,7 +7,7 @@ project(prplmesh
 include(cmake/multiap-helpers.cmake)
 
 if(NOT DEFINED TARGET_PLATFORM)
-set(TARGET_PLATFORM "linux")
+    set(TARGET_PLATFORM "linux")
 endif()
 
 option (BUILD_AGENT "Build EasyMesh agent" ON)
@@ -39,7 +39,7 @@ message(STATUS "CMAKE_PREFIX_PATH - ${CMAKE_PREFIX_PATH}")
 message(STATUS "CMAKE_MULTIAP_OUTPUT_DIRECTORY - ${CMAKE_MULTIAP_OUTPUT_DIRECTORY}")
 
 if(NOT CMAKE_BUILD_TYPE)
-set(CMAKE_BUILD_TYPE "Release")
+    set(CMAKE_BUILD_TYPE "Release")
 endif()
 
 set(CMAKE_CXX_STANDARD 11)
@@ -175,10 +175,10 @@ endif()
 add_subdirectory(framework)
 add_subdirectory(common)
 if (BUILD_AGENT)
-add_subdirectory(agent)
+    add_subdirectory(agent)
 endif()
 if (BUILD_CONTROLLER)
-add_subdirectory(controller)
+    add_subdirectory(controller)
 endif()
 
 ## Version
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,7 @@ if ("${isSystemDir}" STREQUAL "-1")
 endif ("${isSystemDir}" STREQUAL "-1")
 set (CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
-set(INSTALL_PATH "/opt/prplmesh")
+set(INSTALL_PATH "/opt/prplmesh" CACHE PATH "prplMesh installation directory")
 set(TMP_PATH "/tmp/beerocks")
 
 if(TARGET_PLATFORM STREQUAL "linux")
@@ -102,22 +102,20 @@ if(TARGET_PLATFORM STREQUAL "linux")
 endif()
 
 # Default values
-set(BEEROCKS_WLAN1_IFACE "wlan0")
-set(BEEROCKS_WLAN2_IFACE "wlan2")
-set(BEEROCKS_WLAN3_IFACE "wlan4")
-set(BEEROCKS_BRIDGE_IFACE "br-lan")
-if (NOT DEFINED BEEROCKS_BH_WIRE_IFACE)
-    set(BEEROCKS_BH_WIRE_IFACE "eth1")
-endif()
+set(BEEROCKS_WLAN1_IFACE   "wlan0" CACHE STRING "1st WLAN iface")
+set(BEEROCKS_WLAN2_IFACE   "wlan2" CACHE STRING "2nd WLAN iface")
+set(BEEROCKS_WLAN3_IFACE   "wlan4" CACHE STRING "3rd WLAN iface")
+set(BEEROCKS_BRIDGE_IFACE  "br-lan" CACHE STRING "LAN bridge")
+set(BEEROCKS_BH_WIRE_IFACE "eth1" CACHE STRING "Backhaul wire iface")
 set(BEEROCKS_REPEATER_MODE 0)
 
 # Logging configuration
-set(BEEROCKS_LOG_FILES_ENABLED      "true")
-set(BEEROCKS_LOG_FILES_PATH         "${TMP_PATH}/logs")
-set(BEEROCKS_LOG_FILES_SUFFIX       ".log")
-set(BEEROCKS_LOG_FILES_AUTO_ROLL    "true")
-set(BEEROCKS_LOG_STDOUT_ENABLED     "false")
-set(BEEROCKS_LOG_SYSLOG_ENABLED     "false")
+set(BEEROCKS_LOG_FILES_ENABLED   "true" CACHE STRING "Write prplMesh logs into files")
+set(BEEROCKS_LOG_FILES_PATH      "${TMP_PATH}/logs" CACHE PATH "prplMesh logs directory")
+set(BEEROCKS_LOG_FILES_SUFFIX    ".log" CACHE STRING "prplMesh logs filenames suffix")
+set(BEEROCKS_LOG_FILES_AUTO_ROLL "true" CACHE STRING "Auto rollback prplMesh logs")
+set(BEEROCKS_LOG_STDOUT_ENABLED  "false" CACHE STRING "Print logs to stdout")
+set(BEEROCKS_LOG_SYSLOG_ENABLED  "false" CACHE STRING "Send logs to syslog")
 
 # Platform specific flags
 if (TARGET_PLATFORM STREQUAL "openwrt")
--- a/common/CMakeLists.txt
+++ b/common/CMakeLists.txt
@@ -1,7 +1,7 @@
 message("${BoldWhite}Preparing ${BoldGreen}common${BoldWhite} for the ${BoldGreen}${TARGET_PLATFORM}${BoldWhite} platform${ColourReset}")
 
 # Default values
-set(BEEROCKS_PLATFORM_INIT "false")
+set(BEEROCKS_PLATFORM_INIT "false" CACHE STRING "Configure minimal set of dummy WLAN ifaces and bridge")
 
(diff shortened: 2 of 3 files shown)


ghost (approved):

CMakeLists.txt
```diff
@@ -89,8 +89,12 @@ if ("${isSystemDir}" STREQUAL "-1")
 endif ("${isSystemDir}" STREQUAL "-1")
 set (CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
-set(INSTALL_PATH "/opt/prplmesh")
-set(TMP_PATH "/tmp/beerocks")
+if(NOT DEFINED INSTALL_PATH)
+    set(INSTALL_PATH "/opt/prplmesh")
+endif()
```
arnout: The normal way of allowing setting from the outside would be to make it a cache variable:
```suggestion
set(INSTALL_PATH "/opt/prplmesh" CACHE PATH "prplMesh installation directory")
```
However, I would much prefer to remove INSTALL_PATH entirely and instead fully rely on CMAKE_INSTALL_PREFIX.

CMakeLists.txt
```diff
@@ -89,8 +89,12 @@ if ("${isSystemDir}" STREQUAL "-1")
...
+if(NOT DEFINED TMP_PATH)
+    set(TMP_PATH "/tmp/beerocks")
+endif()
```
arnout: Again, use a cache variable instead. However, why do you need to override this at all for RDKB?

CMakeLists.txt
```diff
@@ -102,22 +106,44 @@ if(TARGET_PLATFORM STREQUAL "linux")
...
+if(NOT DEFINED BEEROCKS_WLAN2_IFACE)
+    set(BEEROCKS_WLAN2_IFACE "wlan2")
+endif()
+if(NOT DEFINED BEEROCKS_WLAN3_IFACE)
+    set(BEEROCKS_WLAN3_IFACE "wlan4")
+endif()
+if(NOT DEFINED BEEROCKS_BRIDGE_IFACE)
+    set(BEEROCKS_BRIDGE_IFACE "br-lan")
+endif()
```
arnout: All this stuff really should be removed. Instead, we should parse it out of the configuration file, if appropriate. For prplmesh_utils.sh that's actually a little difficult... So OK for now.

CMakeLists.txt
```diff
@@ -102,22 +106,44 @@ if(TARGET_PLATFORM STREQUAL "linux")
...
 if (NOT DEFINED BEEROCKS_BH_WIRE_IFACE)
     set(BEEROCKS_BH_WIRE_IFACE "eth1")
 endif()
-set(BEEROCKS_REPEATER_MODE 0)
+if (NOT DEFINED BEEROCKS_REPEATER_MODE)
+    set(BEEROCKS_REPEATER_MODE 0)
+endif()
```
arnout: Do you really need to change this for the RDKB build? Please only make this modification for stuff that really needs to be changed from the outside...

common/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 message("${BoldWhite}Preparing ${BoldGreen}common${BoldWhite} for the ${BoldGreen}${TARGET_PLATFORM}${BoldWhite} platform${ColourReset}")
 
 # Default values
-set(BEEROCKS_PLATFORM_INIT "false")
+if (NOT DEFINED BEEROCKS_PLATFORM_INIT)
+    set(BEEROCKS_PLATFORM_INIT "false")
+endif()
```
arnout: This one really should be removed entirely, as I note in your other PR.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -10,7 +10,9 @@ file(GLOB_RECURSE bwl_common_sources ${MODULE_PATH}/common/*.c*)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
 
 # Default defaults
-set(BWL_TYPE "DUMMY")
+if (NOT DEFINED BWL_TYPE)
+    set(BWL_TYPE "DUMMY")
+endif()
```
arnout: Cache variable please.

framework/CMakeLists.txt
```diff
@@ -2,8 +2,10 @@ project(multiap_framework C CXX)
 
 option(BUILD_EXAMPLES "build examples" OFF)
 
-set(MSGLIB "zmq" CACHE STRING "Which messaging library backend to use")
-set_property(CACHE MSGLIB PROPERTY STRINGS "zmq" "nng" "None")
+if (NOT DEFINED MSGLIB)
+    set(MSGLIB "zmq" CACHE STRING "Which messaging library backend to use")
+    set_property(CACHE MSGLIB PROPERTY STRINGS "zmq" "nng" "None")
+endif()
```
arnout: Here you're converting the correct thing into something wrong...

CMakeLists.txt
```diff
@@ -102,22 +106,44 @@ if(TARGET_PLATFORM STREQUAL "linux")
...
-set(BEEROCKS_BRIDGE_IFACE "br-lan")
+if(NOT DEFINED BEEROCKS_WLAN1_IFACE)
+    set(BEEROCKS_WLAN1_IFACE "wlan0")
+endif()
+if(NOT DEFINED BEEROCKS_WLAN2_IFACE)
+    set(BEEROCKS_WLAN2_IFACE "wlan2")
+endif()
+if(NOT DEFINED BEEROCKS_WLAN3_IFACE)
+    set(BEEROCKS_WLAN3_IFACE "wlan4")
+endif()
```
arnout: This one is actually not used, please remove it instead.

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ project(prplmesh
 include(cmake/multiap-helpers.cmake)
 
 if(NOT DEFINED TARGET_PLATFORM)
-set(TARGET_PLATFORM "linux")
+    set(TARGET_PLATFORM "linux")
```
arnout, line 10: Please do the whitespace-only changes in a separate commit.

arnout (changes requested):
Overall, four types of changes requested:
- Do the whitespace-only changes in a separate commit.
- Remove variables that are not used (in a separate commit).
- Use cache variables instead of the "if not" pattern.
- Don't change it for variables that you don't need to override for RDKB.

For the variables that really should be removed entirely (all the configuration stuff), it's OK to keep the "if not" pattern, since it's only temporary anyway.

framework/CMakeLists.txt
MuteSpirit: Agree
Reverted my changes here

CMakeLists.txt
MuteSpirit: Should i also remove next code from `agent/config/beerocks_agent.conf.in`?
```

[agent2]
radio_identifier=00:00:00:00:00:02
hostap_iface_type=WIFI_INTEL
hostap_ant_gain=0
enable_repeater_mode=0
```

CMakeLists.txt
MuteSpirit: No
Reverted changes for it

CMakeLists.txt
MuteSpirit: I used cached variable for them too to have 4 lines instead of 12 as here

CMakeLists.txt
MuteSpirit: I rechecked that it's not needed to redefine this variable in recipe.
Reverted my changes for it.

CMakeLists.txt
MuteSpirit: > However, I would much prefer to remove INSTALL_PATH entirely and instead fully rely on CMAKE_INSTALL_PREFIX.

There is special task for that #1253
I think that it's out of this PR scope

common/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 message("${BoldWhite}Preparing ${BoldGreen}common${BoldWhite} for the ${BoldGreen}${TARGET_PLATFORM}${BoldWhite} platform${ColourReset}")
 
 # Default values
-set(BEEROCKS_PLATFORM_INIT "false")
+set(BEEROCKS_PLATFORM_INIT "false" CACHE STRING "Configure minimal set of dummy WLAN ifaces and bridge")
```
rmelotte, line 4: This variable is only used in the prplmesh_utils.sh script.
There, it can be controlled using the "--platform-init" option. Why would we additionally need to set the default to true?
Also, with this change and if BEEROCKS_PLATFORM_INIT is set to true, the "--platform-init" option from the script becomes unusable, which can be confusing.

rmelotte (approved):

CMakeLists.txt
arnout: No, we will add a 3-radio platform at some point so that will still be useful. However, we'll probably won't have a docker test for it so the CMake option shouldn't be needed.

common/CMakeLists.txt
arnout: I also think it should be removed entirely, but as Ivan wrote somewhere else: that's not the purpose of this PR. As it is now, the PR nicely converts all existing options that should be configurable to cache variables, so I'm happy with it.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -10,7 +10,8 @@ file(GLOB_RECURSE bwl_common_sources ${MODULE_PATH}/common/*.c*)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
 
 # Default defaults
-set(BWL_TYPE "DUMMY")
+set(BWL_TYPE "DUMMY" CACHE STRING "BWL type")
```
arnout, line 13: It is already defined as a cache variable below. So that one should be removed. 

However, I still don't think it's OK then. IIUC a plain variable overrides a cache variable, and both can be defined at the same time. So the `set` commands within the `TARGET_PLATFORM` condition below will override whatever was set in the cache variable. I can't say that CMake is entirely clear for me though, so I could be wrong about this.

Even so, I think it would be much clearer to do what I suggested before: change the variable name to `BWL_TYPE_DEFAULT` and use that in the definition of the cache variable below.

common/beerocks/bwl/CMakeLists.txt
```diff
@@ -10,7 +10,8 @@ file(GLOB_RECURSE bwl_common_sources ${MODULE_PATH}/common/*.c*)
...
+set_property(CACHE BWL_TYPE PROPERTY STRINGS "DUMMY" "NL80211" "DWPAL")
```
arnout, line 14: Nice!

arnout (approved):
If it works in RDKB with the current definition of BWL_TYPE, I'm OK to merge it as is.

Really nice commit messages by the way, thanks!

PR #1358, opened 2020-05-29
Port test_initial_ap_config to boardfarm for prplWRT device

### LINKED ISSUE #893: [TASK] Run test flows on a device
## Description

Within prplmesh, we have 2 main kinds of tests:
- certification tests
- flow tests

Certification tests test that the device conforms to the specification and use tests designed by the WiFi alliance (running in a WiFi alliance specified setup) to achieve that goal.
Flow tests are our own tests, to check our work. They verify that prplmesh works in a sensible way, which is different from "just barely passing certification tests".

Until now, these tests have just used docker to simulate different devices; however, we want to actually run the flow tests on physical devices

## Exit criteria

- [ ] Gitlab CI is automatically running flow tests on physical hardware on each check-in

## Detailed description

See #738. The architecture and design of the flow tests is being rewritten and should be more extensible from now on.
This task is to extend the new python flow test architecture to run the tests on a physical device.

Concretely:
- [x] Set up a test machine which you'll be using to run the tests. This should be a linux-based machine with a wifi chip. Could be a development board like a raspberry pi (so long as it has working wifi) or it could be a laptop. It could even be the laptop you're working on, if you're running linux natively.
- [x] You will need a Wireless AP as well. Right now, we're focusing on the RAX-40, so you should get that, and integrate it into your local setup. Basically, we want to standardize on linux laptop+RAX40 as a local and remote test platform (good for rapid verification of new code)
- [ ] Create a new back-end for the python tests, to run them on linux using standard linux commands, and to control the RAX40. For details, please ask on slack (`#prplmesh-dev`) (#1074)
- [ ] Make sure that every test which passes in docker, also passes with your new back-end
- [ ] Finally, integrate these tests into gitlab CI.

### LINKED ISSUE #1072: Port test_initial_ap_config to boardfarm for prplWRT device

abelog:
## Description
This PR partially resolves #1072 .
Main goal is to bring InitialApConfig to RAX40.

**Prerequisite**:
1.  Interface, which has RAX40 connected, is included in docker bridge.
1.  Name of this docker network is stated in `boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json` 

This network is used for dockerized controller -> RAX40's agent. In order to add iface to the bridge, `sudo` powers are required. Currently, we do not plan to run `bft` with `sudo` power. Thus, docker network preparation and manual include of interface is required.

## Changes
- `prplWRT` device class to handle RAX40
- prplMesh specific entities for `environment.py`
- `rax40-1`RAX40 board added to boardfarm configuration

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
@@ -100,7 +100,7 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
         """
         deadline = time.monotonic() + timeout
         while time.monotonic() < deadline:
-            if self.prplMesh_check_state():
+            if self.get_prplMesh_status():
                 break
             time.sleep(5)
         else:
@@ -115,7 +115,7 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
         """
         return True
 
-    def prplMesh_check_state(self) -> bool:
+    def get_prplMesh_status(self) -> bool:
         """ Check prplMesh status. Return True if operational."""
         self.sendline("/etc/init.d/prplmesh status")
         self.expect(
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
@@ -6,19 +6,28 @@
 ###############################################################
 
 import boardfarm
-from boardfarm.devices import OpenWrtRouter
-from environment import ALEntityPrplWrt, _get_bridge_interface
+import json
+import os
+import pexpect
+import subprocess
+import sys
+
 from .prplmesh_base import PrplMeshBase
+from boardfarm.devices import connection_decider
+from boardfarm.devices.openwrt_router import OpenWrtRouter
+from environment import ALEntityPrplWrt, _get_bridge_interface
+from ipaddress import IPv4Network, IPv4Address
 from sniffer import Sniffer
 
 
 class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
     """prplWRT burned device with prplMesh installed."""
 
-    model = "prplWRT"
+    model = ("prplWRT")
     prompt = ['root\\@OpenWrt:/#', '/#', '@OpenWrt:/#']
     wan_iface = "eth1"
     uboot_eth = "eth0_1"
+    linesep = "\r"
     agent_entity = None
     controller_entity = None
 
@@ -27,17 +36,41 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
         self.args = args
         self.kwargs = kwargs
         config = kwargs.get("config", kwargs)
+
+        self.unique_id = os.getenv("SUDO_USER", os.getenv("USER", ""))
         self.docker_network = config.get("docker_network",
                                          "prplMesh-net-{}".format(self.unique_id))
         self.role = config.get("role", "agent")
-        if self.role == "controller":
-            self.controller_entity = ALEntityPrplWrt(self.name, is_controller=True)
-        else:
-            self.agent_entity = ALEntityPrplWrt(self.name, is_controller=False)
+        self.connection_type = config.get("connection_type", None)
+        self.conn_cmd = config.get("conn_cmd", None)
 
+        self.connection = connection_decider.connection(device=self,
+                                                        conn_type=kwargs['connection_type'],
+                                                        **kwargs)
+        self.connection.connect()
+
+        self.name = "-".join((config.get("name", "rax40"), self.unique_id))
+        self.consoles = [self]
+        self.logfile_read = sys.stdout
+
+        self.wan_network = self.get_docker_subnet(self.docker_network)
+        self.set_iface_to_bridge(self.wan_iface)
+        self.set_iface_ip("br-lan", self.wan_network[+245], self.wan_network.prefixlen)
         self.wired_sniffer = Sniffer(_get_bridge_interface(self.docker_network),
                                      boardfarm.config.output_dir)
-        self.check_status()
+        if self.role == "controller":
+            self.controller_entity = ALEntityPrplWrt(self, is_controller=True)
+        else:
+            self.agent_entity = ALEntityPrplWrt(self, is_controller=False)
+            self.prplMesh_start_agent()
+
+    def _prplMesh_exec(self, mode: str):
+        """Send line to prplmesh initd script."""
+        self.sendline("/etc/init.d/prplmesh {}".format(mode))
+
+    def kill_console_at_exit(self):
+        """Kill connections on device termination."""
+        self.connection.close()
 
     def check_status(self) -> bool:
         """Check status of device, return bool to indicate state.
@@ -45,11 +78,15 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
         It is used by boardfarm to indicate that spawned device instance is ready for test
         and also after test - to insure that device still operational.
         """
+        return True
+
+    def prplMesh_check_state(self) -> bool:
+        """ Check prplMesh status. Return True if operational."""
         self.sendline("/etc/init.d/prplmesh status")
         match = self.expect(
-                ["OPERATIONAL", self.device.pexpect.EOF, self.device.pexpect.TIMEOUT],
+                ["operational", "FAIL", pexpect.EOF, pexpect.TIMEOUT],
                 timeout=10)
-        if match == 1:
+        if match == 0:
             return True
         else:
             return False
@@ -67,3 +104,36 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
         Purpose is to keep consoles active, so they don't disconnect for long running activities.
         """
         pass
+
+    def get_docker_subnet(self, docker_network: str) -> IPv4Network:
+        """Get subnet used by docker network."""
+        docker_network_inspect_cmd = ('docker', 'network', 'inspect', docker_network)
+        inspect_raw = subprocess.run(docker_network_inspect_cmd, stdout=subprocess.PIPE)
+        if inspect_raw.returncode != 0:
+            # Assume network doesn't exist yet. Create it.
+            # Raise an exception if it fails (check=True).
+            subprocess.run(('docker', 'network', 'create', docker_network), check=True,
+                           stdout=subprocess.DEVNULL)
+            # Inspect again, now raise if it fails (check=True).
+            inspect_raw = subprocess.run(docker_network_inspect_cmd, check=True,
+                                         stdout=subprocess.PIPE)
+        inspect_json = json.loads(inspect_raw.stdout)
+
+        return IPv4Network(inspect_json[0]["IPAM"]["Config"][0]["Subnet"])
+
+    def set_iface_to_bridge(self, iface: str) -> bool:
+        """Add specified interface to the specified bridge."""
+        ip_command = ("ip link set {} master br-lan".format(iface))
+        self.sendline(ip_command)
+        self.expect(self.prompt, timeout=10)
+
+    def set_iface_ip(self, iface: str, ip: IPv4Address, prefixlen :int) -> bool:
+        """Set interface IPv4 address."""
+        self.sendline("ip a add {}/{} dev {}".format(ip, prefixlen, iface))
+        self.expect(self.prompt, timeout=10)
+
+    def prplMesh_start_agent(self) -> bool:
+        """Start prplMesh in certification_mode agent. Return true if done."""
+        self._prplMesh_exec("certification_mode agent")
+        self.expect_exact("CAC timer expired", timeout=120)
+        self.expect_exact("device br-lan entered promiscuous mode", timeout=40)
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
@@ -9,8 +9,10 @@ import boardfarm
 import json
 import os
 import pexpect
+import signal
 import subprocess
 import sys
+import time
 
 from .prplmesh_base import PrplMeshBase
 from boardfarm.devices import connection_decider
@@ -23,8 +25,8 @@ from sniffer import Sniffer
 class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
     """prplWRT burned device with prplMesh installed."""
 
-    model = ("prplWRT")
-    prompt = ['root\\@OpenWrt:/#', '/#', '@OpenWrt:/#']
+    model = "prplWRT"
+    prompt = ['root\\@OpenWrt:/#', '/#', '@OpenWrt:/#', "@OpenWrt:~#"]
     wan_iface = "eth1"
     uboot_eth = "eth0_1"
     linesep = "\r"
@@ -43,34 +45,67 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
         self.role = config.get("role", "agent")
         self.connection_type = config.get("connection_type", None)
         self.conn_cmd = config.get("conn_cmd", None)
+        self.wan_ip = config.get("wan_ip", None)
+        self.username = config.get("username", "root")
 
-        self.connection = connection_decider.connection(device=self,
-                                                        conn_type=kwargs['connection_type'],
-                                                        **kwargs)
-        self.connection.connect()
+        self.name = "-".join((config.get("name", "netgear-rax40"), self.unique_id))
 
-        self.name = "-".join((config.get("name", "rax40"), self.unique_id))
-        self.consoles = [self]
-        self.logfile_read = sys.stdout
+        # If no WAN IP is set in config file retrieve IP from docker network set in config
+        # X.X.X.245 IP will be selected from docker network
+        if not self.wan_ip:
+            self.connection = connection_decider.connection(device=self,
+                                                            conn_type="local_serial",
+                                                            **kwargs)
+            self.connection.connect()
+            self.consoles = [self]
+            self.logfile_read = sys.stdout
+            self.wan_network = self.get_docker_subnet()
+            self.wan_ip = self.wan_network[+245]
+            self.set_iface_ip("br-lan", self.wan_ip, self.wan_network.prefixlen)
+            self.connection.close()
+            self.kill(signal.SIGTERM)
+            # Removal of PID is required by pexpect in order to spawn a new process
+            # serial connection should be terminated by 2 commands above
+            self.pid = None
 
-        self.wan_network = self.get_docker_subnet(self.docker_network)
-        self.set_iface_to_bridge(self.wan_iface)
-        self.set_iface_ip("br-lan", self.wan_network[+245], self.wan_network.prefixlen)
         self.wired_sniffer = Sniffer(_get_bridge_interface(self.docker_network),
                                      boardfarm.config.output_dir)
+        self.connection = connection_decider.connection(device=self,
+                                                        conn_type="ssh",
+                                                        conn_cmd="ssh {}@{}".format(
+                                                            self.username, self.wan_ip))
+        self.connection.connect()
+        # Append active connection to the general array for logging
+        self.consoles = [self]
+        # Point what to log as data read from child process of pexpect
+        # Result: boardfarm will log communication in separate file
+        self.logfile_read = sys.stdout
+        self.add_iface_to_bridge(self.wan_iface, "br-lan")
+
         if self.role == "controller":
             self.controller_entity = ALEntityPrplWrt(self, is_controller=True)
         else:
             self.agent_entity = ALEntityPrplWrt(self, is_controller=False)
-            self.prplMesh_start_agent()
+            self.prplmesh_start_agent()
 
     def _prplMesh_exec(self, mode: str):
         """Send line to prplmesh initd script."""
         self.sendline("/etc/init.d/prplmesh {}".format(mode))
 
-    def kill_console_at_exit(self):
-        """Kill connections on device termination."""
-        self.connection.close()
+    def _prplmesh_status_poll(self, timeout: int = 120) -> bool:
+        """Poll prplMesh status for timeout time.
+
+        Main agent and wlan0, wlan2 radios should be operational.
+        Return True if status is operational and timeout not reached.
+        """
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
+            if self.prplMesh_check_state():
+                break
+            time.sleep(5)
+        else:
+            return False
+        return True
 
     def check_status(self) -> bool:
         """Check status of device, return bool to indicate state.
@@ -83,10 +118,12 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
     def prplMesh_check_state(self) -> bool:
         """ Check prplMesh status. Return True if operational."""
         self.sendline("/etc/init.d/prplmesh status")
-        match = self.expect(
-                ["operational", "FAIL", pexpect.EOF, pexpect.TIMEOUT],
-                timeout=10)
-        if match == 0:
+        self.expect(
+                ["(?P<main_agent>OK) Main agent.+"
+                 "(?P<wlan0>OK) wlan0.+"
+                 "(?P<wlan2>OK) wlan2", pexpect.TIMEOUT],
+                timeout=5)
+        if self.match is not pexpect.TIMEOUT:
             return True
         else:
             return False
@@ -105,14 +142,14 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
         """
         pass
 
-    def get_docker_subnet(self, docker_network: str) -> IPv4Network:
+    def get_docker_subnet(self) -> IPv4Network:
         """Get subnet used by docker network."""
-        docker_network_inspect_cmd = ('docker', 'network', 'inspect', docker_network)
+        docker_network_inspect_cmd = ('docker', 'network', 'inspect', self.docker_network)
         inspect_raw = subprocess.run(docker_network_inspect_cmd, stdout=subprocess.PIPE)
         if inspect_raw.returncode != 0:
             # Assume network doesn't exist yet. Create it.
             # Raise an exception if it fails (check=True).
-            subprocess.run(('docker', 'network', 'create', docker_network), check=True,
+            subprocess.run(('docker', 'network', 'create', self.docker_network), check=True,
                            stdout=subprocess.DEVNULL)
             # Inspect again, now raise if it fails (check=True).
             inspect_raw = subprocess.run(docker_network_inspect_cmd, check=True,
@@ -121,19 +158,25 @@ class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
 
         return IPv4Network(inspect_json[0]["IPAM"]["Config"][0]["Subnet"])
 
-    def set_iface_to_bridge(self, iface: str) -> bool:
+    def add_iface_to_bridge(self, iface: str, bridge: str) -> bool:
         """Add specified interface to the specified bridge."""
-        ip_command = ("ip link set {} master br-lan".format(iface))
+        ip_command = ("ip link set {} master {}".format(iface, bridge))
         self.sendline(ip_command)
         self.expect(self.prompt, timeout=10)
 
-    def set_iface_ip(self, iface: str, ip: IPv4Address, prefixlen :int) -> bool:
+    def set_iface_ip(self, iface: str, ip: IPv4Address, prefixlen: int) -> bool:
         """Set interface IPv4 address."""
         self.sendline("ip a add {}/{} dev {}".format(ip, prefixlen, iface))
         self.expect(self.prompt, timeout=10)
 
-    def prplMesh_start_agent(self) -> bool:
+    def prplmesh_start_agent(self) -> bool:
         """Start prplMesh in certification_mode agent. Return true if done."""
         self._prplMesh_exec("certification_mode agent")
-        self.expect_exact("CAC timer expired", timeout=120)
-        self.expect_exact("device br-lan entered promiscuous mode", timeout=40)
+        self.expect(self.prompt)
+        return True
+
+    def prprlmesh_status_check(self) -> bool:
+        """Check prplMesh status by executing status command to initd service.
+        Return True if operational.
+        """
+        return self._prplmesh_status_poll()
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
@@ -16,5 +16,22 @@
                 "conn_cmd": ""
             }
         ]
+    },
+    "netgear-rax40-1": {
+        "name": "agent-rax40",
+        "board_type": "prplWRT",
+        "role": "agent",
+        "docker_network": "prplMesh-net-rax40-1",
+        "connection_type": "local_serial",
+        "conn_cmd": "cu -s 115200 -l /dev/ttyUSB0",
+        "devices": [
+            {
+            "name": "controller",
+            "type": "prplmesh_docker",
+            "role": "controller",
+            "docker_network": "prplMesh-net-rax40-1",
+            "conn_cmd": ""
+            }
+        ]
     }
 }
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
@@ -22,7 +22,6 @@
         "board_type": "prplWRT",
         "role": "agent",
         "docker_network": "prplMesh-net-rax40-1",
-        "connection_type": "local_serial",
         "conn_cmd": "cu -s 115200 -l /dev/ttyUSB0",
         "devices": [
             {
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
@@ -15,6 +15,10 @@ class InitialApConfig(PrplMeshBaseTest):
                 dev.wired_sniffer.start(self.__class__.__name__ + "-" + dev.name)
 
                 self.check_log(dev.agent_entity.radios[0],
+                               r"Controller configuration \(WSC M2 Encrypted Settings\)")
+                self.check_log(dev.agent_entity.radios[1],
+                               r"Controller configuration \(WSC M2 Encrypted Settings\)")
+                self.check_log(dev.agent_entity.radios[0],
                                r"WSC Global authentication success")
                 self.check_log(dev.agent_entity.radios[1],
                                r"WSC Global authentication success")
@@ -22,13 +26,13 @@ class InitialApConfig(PrplMeshBaseTest):
                                r"KWA \(Key Wrap Auth\) success")
                 self.check_log(dev.agent_entity.radios[1],
                                r"KWA \(Key Wrap Auth\) success")
-                self.check_log(dev.agent_entity.radios[0],
-                               r".* Controller configuration \(WSC M2 Encrypted Settings\)")
-                self.check_log(dev.agent_entity.radios[1],
-                               r".* Controller configuration \(WSC M2 Encrypted Settings\)")
-
-                dev.wired_sniffer.stop()
 
     @classmethod
     def teardown_class(cls):
         """Teardown method, optional for boardfarm tests."""
+        test = cls.test_obj
+        for dev in test.dev:
+            if dev.agent_entity:
+                print("Sniffer - stop")
+                dev.agent_entity.device.send('\003')
+                dev.wired_sniffer.stop()
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
@@ -14,7 +14,7 @@ class InitialApConfig(PrplMeshBaseTest):
             if dev.agent_entity:
                 dev.wired_sniffer.start(self.__class__.__name__ + "-" + dev.name)
 
-                self.prplmesh_status_check(dev.agent_entity.device)
+                self.prplmesh_status_check(dev.agent_entity)
                 self.check_log(dev.agent_entity.radios[0],
                                r"\(WSC M2 Encrypted Settings\)")
                 self.check_log(dev.agent_entity.radios[1],
@@ -35,5 +35,10 @@ class InitialApConfig(PrplMeshBaseTest):
         for dev in test.dev:
             if dev.agent_entity:
                 print("Sniffer - stop")
-                dev.agent_entity.device.send('\003')
+                # Send Ctrl+C to the device to terminate "tail -f"
+                # Which is used to read log from device. Required only for tests on HW
+                try:
+                    dev.agent_entity.device.send('\003')
+                except AttributeError:
+                    pass
                 dev.wired_sniffer.stop()
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
@@ -14,10 +14,11 @@ class InitialApConfig(PrplMeshBaseTest):
             if dev.agent_entity:
                 dev.wired_sniffer.start(self.__class__.__name__ + "-" + dev.name)
 
+                self.prplmesh_status_check(dev.agent_entity.device)
                 self.check_log(dev.agent_entity.radios[0],
-                               r"Controller configuration \(WSC M2 Encrypted Settings\)")
+                               r"\(WSC M2 Encrypted Settings\)")
                 self.check_log(dev.agent_entity.radios[1],
-                               r"Controller configuration \(WSC M2 Encrypted Settings\)")
+                               r"\(WSC M2 Encrypted Settings\)")
                 self.check_log(dev.agent_entity.radios[0],
                                r"WSC Global authentication success")
                 self.check_log(dev.agent_entity.radios[1],
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/prplmesh_base_test.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/prplmesh_base_test.py
@@ -16,7 +16,7 @@ class PrplMeshBaseTest(bft_base_test.BftBaseTest):
     """
 
     def check_log(self, entity_or_radio: Union[env.ALEntity, env.Radio], regex: str,
-                  start_line: int = 0, timeout: float = 0.3) -> bool:
+                  start_line: int = 0, timeout: float = 0.6) -> bool:
         result, line, match = entity_or_radio.wait_for_log(regex, start_line, timeout)
         if not result:
             raise Exception
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/prplmesh_base_test.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/prplmesh_base_test.py
@@ -21,3 +21,12 @@ class PrplMeshBaseTest(bft_base_test.BftBaseTest):
         if not result:
             raise Exception
         return result, line, match
+
+    def prplmesh_status_check(self, entity_or_radio: Union[env.ALEntity, env.Radio]) -> bool:
+        """Check prplMesh status by executing status command to initd service.
+        Return True if operational.
+        """
+        result = entity_or_radio.prprlmesh_status_check()
+        if not result:
+            raise Exception
+        return result
--- a/tests/environment.py
+++ b/tests/environment.py
@@ -249,10 +249,13 @@ class ALEntityDocker(ALEntity):
 
     The entity is defined from the name of the container, the rest is derived from that.
     '''
-
-    def __init__(self, name: str, is_controller: bool = False):
+    # NOTE: name arg can be also extracted from the device class itself, but test_flows.py
+    # don't have it. We can remove this arg as soon, as we drop test_flows.py
+    def __init__(self, name: str, device: None = None, is_controller: bool = False):
         self.name = name
         self.bridge_name = 'br-lan'
+        if device:
+            self.device = device
 
         # First, get the UCC port from the config file
         if is_controller:
@@ -294,10 +297,12 @@ class ALEntityDocker(ALEntity):
         program = "controller" if self.is_controller else "agent"
         return _docker_wait_for_log(self.name, [program], regex, start_line, timeout)
 
+    def prprlmesh_status_check(self):
+        return self.device.prprlmesh_status_check()
+
 
 class RadioDocker(Radio):
     '''Docker implementation of a radio.'''
-
     def __init__(self, agent: ALEntityDocker, iface_name: str):
         self.iface_name = iface_name
         ip_output = agent.command("ip", "-o",  "link", "list", "dev", self.iface_name).decode()
@@ -398,8 +403,8 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str 
             wired_sniffer.stop()
 
     global controller, agents
-    controller = ALEntityDocker(gateway, True)
-    agents = (ALEntityDocker(repeater1), ALEntityDocker(repeater2))
+    controller = ALEntityDocker(name=gateway, is_controller=True)
+    agents = (ALEntityDocker(name=repeater1), ALEntityDocker(name=repeater2))
 
     debug('controller: {}'.format(controller.mac))
     debug('agent1: {}'.format(agents[0].mac))
@@ -454,6 +459,9 @@ class ALEntityPrplWrt(ALEntity):
         return _device_wait_for_log(self.device,
                                     "{}/beerocks_{}.log".format(self.log_folder, program),
                                     regex, start_line, timeout*100)
+
+    def prprlmesh_status_check(self):
+        return self.device.prprlmesh_status_check()
 
 
 class RadioHostapd(Radio):
--- a/tests/environment.py
+++ b/tests/environment.py
@@ -225,16 +225,23 @@ def _docker_wait_for_log(container: str, programs: [str], regex: str, start_line
 
 
 def _device_wait_for_log(device: None, log_path: str, regex: str,
-                         start_line: int, timeout: int) -> bool:
-    """Waits for log mathing regex expression to show up."""
-    device.sendline("tail -n10 -f {}".format(log_path))
-    match = device.expect(
-        pattern=[regex, device.pexpect.EOF, device.pexpect.TIMEOUT],
-        timeout=timeout)
-    if match == 1:
-        return True
+                         timeout: int, start_line: int = 0):
+    """Waits for log matching regex expression to show up."""
+    device.sendline("tail -f -n +{:d} {}".format(start_line + 1, log_path))
+    device.expect(regex, timeout=timeout)
+    match = device.match.group(0)
+    # Send Ctrl-C to interrupt tail -f
+    device.send('\003')
+    device.expect(device.prompt)
+    if match:
+        device.sendline("tail -n +{:d} {} | grep -n \"{}\"".format(start_line, log_path, match))
+        # Typical output of grep -n from log: "line_num:severity"
+        # this regex has to capture just number of line in log
+        device.expect(r"(?P<line_number>[0-9]+):[A-Z]+\s[0-9]", timeout=timeout)
+        matched_line = int(device.match.group('line_number')) + start_line
+        return (True, matched_line, match)
     else:
-        return False
+        return (False, start_line, None)
 
 
 class ALEntityDocker(ALEntity):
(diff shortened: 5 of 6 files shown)


tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -6,19 +6,27 @@
...
-from boardfarm.devices import OpenWrtRouter
-from environment import ALEntityPrplWrt, _get_bridge_interface
+import json
+import os
+import pexpect
+import subprocess
+import sys
+
 from .prplmesh_base import PrplMeshBase
+from boardfarm.devices import openwrt_router, connection_decider
```
MuteSpirit: To use class name without module name, for example, `openwrt_router.OpenWrtRouter` try
`from boardfarm.devices.openwrt_router import OpenWrtRouter`

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +103,37 @@ def touch(self):
...
+        """Set interface IPv4 address."""
+        self.sendline("ip a add {}/{} dev {}".format(ip[+245], ip.prefixlen, iface))
+        match = self.expect(self.prompt, timeout=10)
+
+    def prplMesh_start_agent(self) -> bool:
+        """Start prplMesh in certification_mode agent. Return true if done."""
+        self._prplMesh_exec("certification_mode agent")
+        self.sendline("/etc/init.d/prplmesh certification_mode agent")
+        self.expect_exact("CAC timer expired", timeout=120)
+        self.expect_exact("device br-lan entered promiscuous mode", timeout=40)
```
MuteSpirit: Name of bridge, `br-lan` in this case, is depend on image settings
I think that it'll be better to take it from setup configuration

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +103,37 @@ def touch(self):
...
+
+    def set_iface_ip(self, iface: str, ip: IPv4Network) -> bool:
+        """Set interface IPv4 address."""
+        self.sendline("ip a add {}/{} dev {}".format(ip[+245], ip.prefixlen, iface))
+        match = self.expect(self.prompt, timeout=10)
+
+    def prplMesh_start_agent(self) -> bool:
+        """Start prplMesh in certification_mode agent. Return true if done."""
+        self._prplMesh_exec("certification_mode agent")
+        self.sendline("/etc/init.d/prplmesh certification_mode agent")
```
MuteSpirit: Both lines perfome the same shell command on board
Why is it needed?

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
```diff
@@ -16,5 +16,22 @@
...
             }
         ]
+    },
+    "rax40-1": {
+        "name": "agent-rax40",
+        "board_type": "prplWRT",
+        "role": "agent",
+        "docker_network": "prplMesh-net-rax40-1",
+        "connection_type": "local_serial",
+        "conn_cmd": "cu -s 115200 -l /dev/ttyUSB0",
```
MuteSpirit, line 25: Assume that two UARTs are connected to the same PC
* 1st one device is `/dev/ttyUSB0`
* 2nd one device is `/dev/ttyUSB1`
But it's not guaranteed that after PC reboot they will not change their devices names and become:
* 1st one device is /dev/ttyUSB**1**
* 2nd one device is /dev/ttyUSB**0**

But if you will use path to serial ports like `/dev/serial/....` and will not replug UARTs into another USB ports then paths to the serial devices will not be changed

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -22,13 +26,14 @@ def runTest(self):
...
-                dev.wired_sniffer.stop()
 
     @classmethod
     def teardown_class(cls):
         """Teardown method, optional for boardfarm tests."""
+        test = cls.test_obj
+        for dev in test.dev:
+            if dev.agent_entity:
+                print("Sniffer - stop")
+                dev.agent_entity.device.send('\003')
```
MuteSpirit: Maybe use special pexpect function for that `dev.agent_entity.device.sendcontrol('c')` ?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +103,37 @@ def touch(self):
...
+            subprocess.run(('docker', 'network', 'create', docker_network), check=True,
+                           stdout=subprocess.DEVNULL)
+            # Inspect again, now raise if it fails (check=True).
+            inspect_raw= subprocess.run(docker_network_inspect_cmd, check=True,
+                                        stdout=subprocess.PIPE)
+        inspect_json = json.loads(inspect_raw.stdout)
+
+        return IPv4Network(inspect_json[0]["IPAM"]["Config"][0]["Subnet"])
+
+    def set_iface_to_bridge(self, iface: str) -> bool:
```
MuteSpirit: By my opinion more readable func name is `add_iface_into_bridge`

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +103,37 @@ def touch(self):
...
+        """Add specified interface to the specified bridge."""
+        ip_command = ("ip link set {} master br-lan".format(iface))
+        self.sendline(ip_command)
+        self.expect(self.prompt, timeout=10)
+
+    def set_iface_ip(self, iface: str, ip: IPv4Network) -> bool:
+        """Set interface IPv4 address."""
+        self.sendline("ip a add {}/{} dev {}".format(ip[+245], ip.prefixlen, iface))
+        match = self.expect(self.prompt, timeout=10)
```
MuteSpirit: This function does not set any IP on interface
Its set IP from passed subnet with last octet `245`
If you want to have general function then it's needed to pass IP and subnet prefix via function arguments and not generate IP inside

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,27 +35,55 @@ def __init__(self, *args, **kwargs):
...
+        """Kill connections on device termination."""
+        self.connection.close()
 
     def check_status(self) -> bool:
         """Check status of device, return bool to indicate state.
 
         It is used by boardfarm to indicate that spawned device instance is ready for test
         and also after test - to insure that device still operational.
         """
+        return True
```
MuteSpirit, line 116: Please add a comment why did you make this function a stub

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,27 +35,55 @@ def __init__(self, *args, **kwargs):
...
+
+    def prplMesh_check_state(self) -> bool:
         self.sendline("/etc/init.d/prplmesh status")
         match = self.expect(
-                ["OPERATIONAL", self.device.pexpect.EOF, self.device.pexpect.TIMEOUT],
+                ["operational", "FAIL", pexpect.EOF, pexpect.TIMEOUT],
                 timeout=10)
         if match == 1:
```
MuteSpirit: It's not enough to found `operational` in output because it's possible that controller is operational but one/both agents - no
All of cases - "FAIL", pexpect.EOF, pexpect.TIMEOUT - mean `False` result

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +103,37 @@ def touch(self):
...
+        """Get subnet used by docker network."""
+        docker_network_inspect_cmd = ('docker', 'network', 'inspect', docker_network)
+        inspect_raw = subprocess.run(docker_network_inspect_cmd, stdout=subprocess.PIPE)
+        if inspect_raw.returncode != 0:
+            # Assume network doesn't exist yet. Create it.
+            # Raise an exception if it fails (check=True).
+            subprocess.run(('docker', 'network', 'create', docker_network), check=True,
+                           stdout=subprocess.DEVNULL)
+            # Inspect again, now raise if it fails (check=True).
+            inspect_raw= subprocess.run(docker_network_inspect_cmd, check=True,
```
MuteSpirit: Missed space before assignment

tests/environment.py
```diff
@@ -189,16 +189,23 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
...
-            timeout=timeout)
-    if match == 1:
-        return True
+                         timeout_set: int, start_line: int = 0):
+    """Waits for log matching regex expression to show up."""
+    device.sendline("tail -f -n +{:d} {}".format(int(start_line)+int(1), log_path))
+    device.expect(regex, timeout=timeout_set)
+    match = device.match.group(0)
+    # Send Ctrl-C to interrupt tail -f
+    device.send('\003')
```
MuteSpirit, line 234: Why not `device.sendcontrol('c')`? It's more readable

tests/environment.py
```diff
@@ -189,16 +189,23 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
...
+    device.expect(device.prompt)
+    if match:
+        device.sendline("grep -n \"{}\" {}".format(match, log_path))
+        # Typical output of grep -n from log: "line_num:severity"
+        # this regex has to capture just number of line in log
+        device.expect(r"(?P<line_number>[0-9]+):[A-Z]+\s[0-9]", timeout=timeout_set)
```
MuteSpirit: Will it find last or first match of regex passed in arguments?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,27 +35,55 @@ def __init__(self, *args, **kwargs):
...
+        self.connection = connection_decider.connection(device=self,
+                                                        conn_type=kwargs['connection_type'],
+                                                        **kwargs)
+        self.connection.connect()
 
+        self.name = "-".join((config.get("name", "rax40"), self.unique_id))
+        self.consoles = [self]
+        self.logfile_read = sys.stdout
+
+        self.wan_network = self.get_docker_subnet(self.docker_network)
```
MuteSpirit: Are you expecting that any boards will be connected with test PC only with WAN port and not with LAN port or both LAN and WAN ports?
I think it's not common case

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
abelog: This is a generic config. AFAIK, path in `/dev/serial/by-path/...` can be very different between workstations. That why `/dev/ttyUSB0` was selected.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: In general, boardfarm uses this method `ping` alike. It executes it in parallel to the test to make sure that device isn't dead. As we are using serial connection, input from status check was mixed up with input from other interacting with console methods.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Both 2 logs are actually specific to RAX40 and driver it uses. Till those logs, device isn't yet ready.
Only good solution i see, is to move that code to separate device class which inherits `prplMesh`.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Left after debug. Should be removed.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Fixed.

tests/environment.py
abelog: For unknown to reason, `sendcotrol('c')` isn't stable. As result,  `tail -f` is not stopped.
`device.send('\003')` works always.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Fixed.
Changed method args to accept IP and subnet prefix separately.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Fixed.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
abelog: For unknown to reason, `sendcotrol('c')` isn't stable. As result,  `tail -f` is not stopped.
`device.send('\003')` works always.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
```diff
@@ -16,5 +16,22 @@
                 "conn_cmd": ""
             }
         ]
+    },
+    "rax40-1": {
```
arnout: It would be easier to call the device "netgear-rax40-1" because that's its name in the test setup.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
rmelotte: What we did in the testbed we use in CI is to use /dev/<device_name>, with a udev rule to create the correct symlink.
Maybe we could also do the same here?

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
```diff
@@ -16,5 +16,22 @@
...
+        "name": "agent-rax40",
+        "board_type": "prplWRT",
+        "role": "agent",
+        "docker_network": "prplMesh-net-rax40-1",
```
rmelotte, line 24: If the container is started with `--network=host` (which a gitlab-runner can also use), the container will use the same network namespace as the host.
Could it be used when there is only one instance of prplMesh running on the host?
If it works, it would avoid the need to configure the docker network beforehand.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -6,19 +6,27 @@
...
+
 from .prplmesh_base import PrplMeshBase
+from boardfarm.devices import openwrt_router, connection_decider
+from environment import ALEntityPrplWrt, _get_bridge_interface
+from ipaddress import IPv4Network
 from sniffer import Sniffer
 
 
-class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
+class PrplMeshPrplWRT(openwrt_router.OpenWrtRouter, PrplMeshBase):
```
rmelotte: Why is this change needed?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -6,19 +6,27 @@
...
     """prplWRT burned device with prplMesh installed."""
 
-    model = "prplWRT"
+    model = ("prplWRT")
```
rmelotte: Why does the model need to be a tuple? I know `PrplMeshDocker` also define it as a tuple, but I couldn't find the information anywhere.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -6,19 +6,27 @@
...
     prompt = ['root\\@OpenWrt:/#', '/#', '@OpenWrt:/#']
     wan_iface = "eth1"
     uboot_eth = "eth0_1"
+    linesep = "\r"
```
rmelotte, line 32: This doesn't seem to be used anywhere. Is it?
Also why is it `\r` and not `\n`?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
rmelotte: >As we are using serial connection, input from status check was mixed up with input from other interacting with console methods

This look like a good reason to use the SSH connection instead?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,29 +35,57 @@ def __init__(self, *args, **kwargs):
...
         and also after test - to insure that device still operational.
         """
+        return True
+
+    def prplMesh_check_state(self) -> bool:
+        """ Check prplMesh status. Return True if operational."""
         self.sendline("/etc/init.d/prplmesh status")
         match = self.expect(
-                ["OPERATIONAL", self.device.pexpect.EOF, self.device.pexpect.TIMEOUT],
+                ["operational", "FAIL", pexpect.EOF, pexpect.TIMEOUT],
```
rmelotte: I think I saw it mentioned elsewhere: looking for the "operational" string is not enough.
It's possible that the main agent is operational but not the separate radio agents for example.
A more reliable way to check the state would be to check the status code, as the script is supposed to exit with something different than 0 if it was not operational.
With pexpect I think it would require to call close() every time though (which probably isn't really an issue).

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,29 +35,57 @@ def __init__(self, *args, **kwargs):
...
                 timeout=10)
-        if match == 1:
+        if match == 0:
             return True
         else:
             return False
```
rmelotte:
```suggestion
    return match == 0
```

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +103,37 @@ def touch(self):
...
+
+    def set_iface_to_bridge(self, iface: str) -> bool:
+        """Add specified interface to the specified bridge."""
+        ip_command = ("ip link set {} master br-lan".format(iface))
+        self.sendline(ip_command)
+        self.expect(self.prompt, timeout=10)
+
+    def set_iface_ip(self, iface: str, ip: IPv4Network) -> bool:
+        """Set interface IPv4 address."""
+        self.sendline("ip a add {}/{} dev {}".format(ip[+245], ip.prefixlen, iface))
```
rmelotte: I don't understand what's going on here.
`set_iface_ip` takes an `ip` argument, but of type IPv4Network. Is it supposed to be an ip or a subnet?
It also always adds 245 to the IP it receives. Is the devices forced to have an IP ending with 245?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,29 +35,57 @@ def __init__(self, *args, **kwargs):
...
+                                                        **kwargs)
+        self.connection.connect()
 
+        self.name = "-".join((config.get("name", "rax40"), self.unique_id))
+        self.consoles = [self]
+        self.logfile_read = sys.stdout
+
+        self.wan_network = self.get_docker_subnet(self.docker_network)
+        self.set_iface_to_bridge(self.wan_iface)
+        self.set_iface_ip("br-lan", self.wan_network)
```
rmelotte: Why would we always have to set a specific IP?
I guess it's because you want to use an IP in the same subnet as the controller running in a docker container?
This looks unnecessarily restrictive to me. It means we won't be able to run tests with a rax40 unless the controller is running in docker.

If we can run the docker controller with the host network namespace, we can also just use the IP the device already has, or if we needed to, put them in a configuration file.

tests/environment.py
```diff
@@ -197,16 +197,23 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
...
+        matched_line = device.match.group('line_number')
+        return (True, matched_line, match)
     else:
-        return False
+        return (False, start_line, None)
```
rmelotte, line 244: >format return as it's expected by callers.

There is something to be sorted out with _device_wait_for_log.
It returns a free-form tuple that is not mentioned anywhere in the documentation.
(this hasn't been introduced in this PR)

tests/environment.py
```diff
@@ -265,7 +273,8 @@ class RadioDocker(Radio):
     def __init__(self, agent: ALEntityDocker, iface_name: str):
         self.iface_name = iface_name
         ip_output = agent.command("ip", "-o",  "link", "list", "dev", self.iface_name).decode()
-        mac = re.search(r"link/ether (([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})", ip_output).group(1)
+        mac = re.search(r"link/ether (([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})",
+                        ip_output).group(1)
```
rmelotte, line 310: Are this one and the 2 changes above whitespace changes only? 
If they are they should be done in separate commits, otherwise it forces reviewers to play the [Spot the difference game](https://en.wikipedia.org/wiki/Spot_the_difference).

tests/environment.py
```diff
@@ -365,24 +374,32 @@ class ALEntityPrplWrt(ALEntity):
...
-
-        ucc_socket = UCCSocket(device_ip, ucc_port)
+        ucc_port_raw = self.command("grep \"ucc_listener_port\" {}".format(self.config_file_name))
+        ucc_port = int(re.search(r'ucc_listener_port=(?P<port>[0-9]+)',
+                                 ucc_port_raw).group('port'))
+        bridge_name_raw = self.command("grep \"bridge_iface\" {}".format(self.config_file_name))
+        self.bridge_name = re.search(r'bridge_iface=(?P<bridge>.+)\r\n',
+                                     bridge_name_raw).group('bridge')
+
+        # Multiple IPs may be set to same interface. We are interested in the last one, which is set
```
rmelotte: This will also no longer be true if we can use the host network for the Docker container.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -126,14 +126,13 @@ def set_iface_to_bridge(self, iface: str) -> bool:
         self.sendline(ip_command)
         self.expect(self.prompt, timeout=10)
 
-    def set_iface_ip(self, iface: str, ip: IPv4Network) -> bool:
+    def set_iface_ip(self, iface: str, ip: IPv4Address, prefixlen :int) -> bool:
```
rmelotte: Ok this fixup makes my previous comment about IPv4Network no longer relevant.
I still think we should use the existing IP if possible.

rmelotte (commented):
Lots of comment, I think some of them should be discussed.
Since Boardfarm is not integrated in CI yet anyway, there is no way to rush this so I'm not approving yet.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
abelog: We can. `dev/ttyUSB0` is used because it's generic. What is serial path of RAX40 on CI machine?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
ghost: You may *think* that that's a tuple, but it's not:
```
frederik@fredevuan-pc:~$ python3
Python 3.7.3 (default, Dec 20 2019, 18:57:59) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> type(3)
<class 'int'>
>>> type((3))
<class 'int'>
>>> type(("prpl"))
<class 'str'>
>>> x=("prpl")
>>> x
'prpl'
>>> type(x)
<class 'str'>
>>> x=("prpl",)
>>> x
('prpl',)
>>> type(x)
<class 'tuple'>
>>>
```

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Wrong "path" was used. To avoid the change, i have to change import in next way:
````
from boardfarm.devices.openwrt_router import OpenWrtRouter
````

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 17 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: It's not a must. Either `"prplWRT"` and `("prplWRT")` will work.

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 18 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py:32]: This variable is used by `pexpect` internally and is not documented.
I have used `\r` following [example](https://github.com/mattsm/boardfarm/blob/master/boardfarm/devices/cougarpark.py) of `cougarpark.py` device class of boardfarm.

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 19 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: Already fixed! Have changed regex expression to match whole report of `status` command. So, in order to pass the check, device should have main agent and both radios in "OK" state.

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 20 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: Fixed already/

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 21 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: Yes. This part was updated, but if no IP/hostname of device is set in boardfarm config - it will be forced to have IP ending with 245.

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 26 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: Already done! Resolving.

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 22 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: > I guess it's because you want to use an IP in the same subnet as the controller running in a docker container?

Exactly.

> This looks unnecessarily restrictive to me. It means we won't be able to run tests with a rax40 unless the controller is running in docker.

Initially, it was planned to have one spawned controller per DUT. Currently, boardfarm is unable to handle non-dockerized controller. We can create a separate boardfarm "device" for it.

File: tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json (hunk 15 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json:24]: Boardfarm can handle multiple setups at the same time. Flow is quite simple, let's imagine you have 2 RAX40 set up, with names `nighthawk-rax40-1` and` nighthawk-rax40-2`.
`nighthawk-rax40-1` is already busy with ongoing test and you wish to already start some test with` nighthawk-rax40-2`. 2 options are available:
1. Ask bft (executable name of boardfarm) to start test X explicitly on `nighthawk-rax40-2`
1. Ask bft to start test X on `prplWRT` board type

In the second case, boardfarm will start looking for not busy board described in boardfarm config file and will start the test with the first free board.

As boardfarm can handle few tests at the same time, sharing the same controller entity between multiple agents under test may bring some corner cases (i am not sure if it is possible). Usage of docker bridges brings isolation between setups and excludes the possibility of a cross set up interference issues at all.

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 16 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: Fixed.

**Comment by abelog** [comment]
Branch was squashed for second cycle of review.

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py (hunk 1 above)
Comment (abelog) [reply tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py]: Fixed!

File: tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json (hunk 15 above)
Comment (arnout) [reply tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json:24]: This should be explained in the commit message or in the README.

File: tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json (hunk 4 above)
Comment (arnout) [reply tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json:25]: Serial path of device netgear-rax40-1 is /dev/netgear-rax40-1. You can take a guess as to the path of device turris-omnia-1 :-)

File: tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py [hunk 27]
```diff
@@ -27,29 +36,57 @@ def __init__(self, *args, **kwargs):
...
-            self.agent_entity = ALEntityPrplWrt(self.name, is_controller=False)
+        self.connection_type = config.get("connection_type", None)
+        self.conn_cmd = config.get("conn_cmd", None)
+
+        self.connection = connection_decider.connection(device=self,
+                                                        conn_type=kwargs['connection_type'],
+                                                        **kwargs)
+        self.connection.connect()
 
+        self.name = "-".join((config.get("name", "rax40"), self.unique_id))
```
arnout: I believe "name" is required in the config, so there shouldn't be a need for a default value. If there is one, it should be netgear-rax40, not just rax40.

Why do you append the unique_id? There can be only one boardfarm instance connected to the device, right? So distinguishing them with a unique_id seems pointless.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,29 +36,57 @@ def __init__(self, *args, **kwargs):
...
+        self.consoles = [self]
+        self.logfile_read = sys.stdout
```
arnout, line 82: This could use some explanation (in a comment).

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,29 +36,57 @@ def __init__(self, *args, **kwargs):
...
+
+        self.wan_network = self.get_docker_subnet(self.docker_network)
+        self.set_iface_to_bridge(self.wan_iface)
+        self.set_iface_ip("br-lan", self.wan_network[+245], self.wan_network.prefixlen)
```
arnout: I don't think we should do this kind of thing from boardfarm. The device is assumed to be already configured when boardfarm runs.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -27,29 +36,57 @@ def __init__(self, *args, **kwargs):
...
         """Check status of device, return bool to indicate state.
 
         It is used by boardfarm to indicate that spawned device instance is ready for test
         and also after test - to insure that device still operational.
         """
+        return True
+
+    def prplMesh_check_state(self) -> bool:
+        """ Check prplMesh status. Return True if operational."""
         self.sendline("/etc/init.d/prplmesh status")
```
arnout, line 120: At some point, this will have to be replaced with a beerocks_cli command on the controller. But not for this PR.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +104,36 @@ def touch(self):
         Purpose is to keep consoles active, so they don't disconnect for long running activities.
         """
         pass
+
+    def get_docker_subnet(self, docker_network: str) -> IPv4Network:
```
arnout: Since this is a member function, it's not necessary to pass docker_network as an argument.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +104,36 @@ def touch(self):
...
+        inspect_raw = subprocess.run(docker_network_inspect_cmd, stdout=subprocess.PIPE)
+        if inspect_raw.returncode != 0:
+            # Assume network doesn't exist yet. Create it.
+            # Raise an exception if it fails (check=True).
+            subprocess.run(('docker', 'network', 'create', docker_network), check=True,
+                           stdout=subprocess.DEVNULL)
+            # Inspect again, now raise if it fails (check=True).
+            inspect_raw = subprocess.run(docker_network_inspect_cmd, check=True,
+                                         stdout=subprocess.PIPE)
+        inspect_json = json.loads(inspect_raw.stdout)
```
arnout: This should be refactored with environment._get_bridge_interface (in a separate PR)

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +104,36 @@ def touch(self):
...
+    def set_iface_to_bridge(self, iface: str) -> bool:
+        """Add specified interface to the specified bridge."""
+        ip_command = ("ip link set {} master br-lan".format(iface))
+        self.sendline(ip_command)
+        self.expect(self.prompt, timeout=10)
+
+    def set_iface_ip(self, iface: str, ip: IPv4Address, prefixlen :int) -> bool:
+        """Set interface IPv4 address."""
+        self.sendline("ip a add {}/{} dev {}".format(ip, prefixlen, iface))
+        self.expect(self.prompt, timeout=10)
```
arnout: As mentioned elsewhere: I don't think these two functions should be needed.

*edit*

After reading the next commit, I now understand this is needed to make sure that the device has an IP address that belongs to the docker network. Still, I think this is taking things too far. Let's just start with the assumption that the device has been set up properly, so if you insist on using a specific network, assume that the interface to the device has already been added to that bridge and that the IP address has been set up properly.

Although even so, I don't see why it would be needed to add the IP address. No IP connectivity is needed between the devices, so why bother?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -67,3 +104,36 @@ def touch(self):
...
+
+    def prplMesh_start_agent(self) -> bool:
+        """Start prplMesh in certification_mode agent. Return true if done."""
+        self._prplMesh_exec("certification_mode agent")
+        self.expect_exact("CAC timer expired", timeout=120)
+        self.expect_exact("device br-lan entered promiscuous mode", timeout=40)
```
arnout: I get neither of these logs when I execute the command on my rax40...

Why do something new and different here? In the existing tests, we just poll prplmesh_utils.sh status until it's operational or times out, what's wrong with that?

tests/environment.py
```diff
@@ -213,16 +213,23 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
...
 def _device_wait_for_log(device: None, log_path: str, regex: str,
-                         start_line: int, timeout: int) -> bool:
-    """Waits for log mathing regex expression to show up."""
-    device.sendline("tail -n10 -f {}".format(log_path))
-    match = device.expect(
-            pattern=[regex, device.pexpect.EOF, device.pexpect.TIMEOUT],
-            timeout=timeout)
-    if match == 1:
-        return True
+                         timeout_set: int, start_line: int = 0):
```
arnout: Why timeout_set instead of timeout?

The arguments are swapped: start_line comes before timeout.

tests/environment.py
```diff
@@ -213,16 +213,23 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
...
+    """Waits for log matching regex expression to show up."""
+    device.sendline("tail -f -n +{:d} {}".format(start_line, log_path))
+    device.expect(regex, timeout=timeout_set)
+    match = device.match.group(0)
```
arnout, line 232: This will raise an exception if it timed out, that's not really what we want...

tests/environment.py
```diff
@@ -213,16 +213,23 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
...
-    """Waits for log mathing regex expression to show up."""
-    device.sendline("tail -n10 -f {}".format(log_path))
-    match = device.expect(
-            pattern=[regex, device.pexpect.EOF, device.pexpect.TIMEOUT],
-            timeout=timeout)
-    if match == 1:
-        return True
+                         timeout_set: int, start_line: int = 0):
+    """Waits for log matching regex expression to show up."""
+    device.sendline("tail -f -n +{:d} {}".format(start_line, log_path))
```
arnout: Tail starts counting at 1, and `tail -n +0` gives the same as `tail -n +1`.
```suggestion
    device.sendline("tail -f -n +{:d} {}".format(start_line + 1, log_path))
```

tests/environment.py
```diff
@@ -213,16 +213,23 @@ def _docker_wait_for_log(container: str, program: str, regex: str, start_line: i
...
+    device.expect(regex, timeout=timeout_set)
+    match = device.match.group(0)
+    # Send Ctrl-C to interrupt tail -f
+    device.send('\003')
+    device.expect(device.prompt)
+    if match:
+        device.sendline("tail -n +{:d} {} | grep -n \"{}\"".format(start_line, log_path, match))
+        # Typical output of grep -n from log: "line_num:severity"
+        # this regex has to capture just number of line in log
+        device.expect(r"(?P<line_number>[0-9]+):[A-Z]+\s[0-9]", timeout=timeout_set)
```
arnout: This second one could be done right in the first command, like so:
```
device.sendline("tail -f -n +{:d} {} | awk  '{ print NR ":" $0; }'".format(start_line + 1, log_path))
device.expect(r"^(?P<line_number>[0-9]+):.*" + regex, timeout=timeout)
```
That doesn't give us the match itself, but that's a feature that is currently unused so I think we can simply remove the feature.

tests/environment.py
arnout: Well, the idea is in fact to remove it again...
The first (boolean) element is redundant with the third one. The third one is not actually used.
start_line should be replaced with something internal that is stored by checkpoint()

tests/environment.py
```diff
@@ -242,7 +249,8 @@ def __init__(self, name: str, is_controller: bool = False):
             config_file_name = 'beerocks_agent.conf'
         with open(os.path.join(installdir, 'config', config_file_name)) as config_file:
             ucc_port = \
-                re.search(r'ucc_listener_port=(?P<port>[0-9]+)', config_file.read()).group('port')
+                re.search(r'ucc_listener_port=(?P<port>[0-9]+)',
+                          config_file.read()).group('port')
```
arnout, line 268: This seems to be a spurious change... The line wasn't too long, so why split it?

tests/environment.py
```diff
@@ -253,8 +261,8 @@ def __init__(self, name: str, is_controller: bool = False):
         else:
             device_ip_output = self.command(
                 'ip', '-f', 'inet', 'addr', 'show', self.bridge_name)
-            device_ip = re.search(
-                r'inet (?P<ip>[0-9.]+)', device_ip_output.decode('utf-8')).group('ip')
+            device_ip = re.search(r'inet (?P<ip>[0-9.]+)',
+                                  device_ip_output.decode('utf-8')).group('ip')
```
arnout, line 280: In the future, please do whitespace-only changes in a separate commit. That makes review a lot easier than trying to find out what changed here and then seeing that actually nothing changed.

tests/environment.py
```diff
@@ -391,66 +400,78 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
-        self.log_folder = self.command(
-            "grep log_files_path {} | cut -d=\'=\' -f2".format(self.config_file_name))
-
-        ucc_socket = UCCSocket(device_ip, ucc_port)
+        ucc_port_raw = self.command("grep \"ucc_listener_port\" {}".format(self.config_file_name))
+        ucc_port = int(re.search(r'ucc_listener_port=(?P<port>[0-9]+)',
+                                 ucc_port_raw).group('port'))
+        bridge_name_raw = self.command("grep \"bridge_iface\" {}".format(self.config_file_name))
+        self.bridge_name = re.search(r'bridge_iface=(?P<bridge>.+)\r\n',
+                                     bridge_name_raw).group('bridge')
```
arnout, line 435: (nitpick) it would be better to do a simple `cat` of the config file once, and then do the re.search twice on the same raw output.

tests/environment.py
```diff
@@ -391,66 +400,78 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
         super().__init__(mac, ucc_socket, installdir, is_controller)
 
         # We always have two radios, wlan0 and wlan2
-        RadioHostapd(self, "wlan0", device=self)
-        RadioHostapd(self, "wlan2", device=self)
+        RadioHostapd(self, "wlan0")
+        RadioHostapd(self, "wlan2")
 
-    def command(self, *command: str) -> bytes:
+    def command(self, command: str) -> bytes:
```
arnout: This change is wrong, it has to be `*command`, otherwise prplmesh_command will no longer work.

tests/environment.py
```diff
@@ -391,66 +400,78 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
         """Execute `command` in device and return its output."""
         self.device.sendline(command)
```
arnout: So this should be
```suggestion
        self.device.sendline(" ".join(command))
```
Actually the command should be shell-quoted of course.

Also it would be better to use pexpect.replwrap.run_command, if that is available in boardfarm.

tests/environment.py
```diff
@@ -391,66 +400,78 @@ def launch_environment_docker(unique_id: str, skip_init: bool = False, tag: str
...
-        return self.device.read()
+        self.device.expect(self.device.prompt, timeout=10)
+        return self.device.before
 
     def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
         """Poll the entity's logfile until it contains "regex" or times out."""
         program = "controller" if self.is_controller else "agent"
+        # Multiply timeout by 100, as test sets it in float.
```
arnout, line 458: Err, both pexpect and the wait_for_log interface have the timeout in seconds, as a float. Where is this centiseconds coming from?

If this is needed, it should be done inside _device_wait_for_log itself.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/initial_ap_config.py
```diff
@@ -22,13 +26,13 @@ def runTest(self):
...
```
arnout: This is not explained in the commit message.

It should definitely be a separate commit, with an explanation why.

It does seem like a good idea to stop the sniffer in the teardown function rather than in the test itself. However:
* wired_sniffer should probably not be associated with a device, it should be more "global" (like we have env.wired_sniffer in test_flows). So in this case, it should be associated with the test class itself.
* if it is stopped in teardown, it should also be started from setup.
* the ctrl-C is not explained. I guess you just want to stop whatever is running on the shell, but there should be a comment explaining that.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/prplmesh_base_test.py
```diff
@@ -16,7 +16,7 @@ class PrplMeshBaseTest(bft_base_test.BftBaseTest):
     """
 
     def check_log(self, entity_or_radio: Union[env.ALEntity, env.Radio], regex: str,
-                  start_line: int = 0, timeout: float = 0.3) -> bool:
+                  start_line: int = 0, timeout: float = 0.6) -> bool:
```
arnout, line 19: Again, this should be in a separate commit.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -23,8 +25,8 @@
 class PrplMeshPrplWRT(OpenWrtRouter, PrplMeshBase):
     """prplWRT burned device with prplMesh installed."""
 
-    model = ("prplWRT")
-    prompt = ['root\\@OpenWrt:/#', '/#', '@OpenWrt:/#']
+    model = "prplWRT"
+    prompt = ['root\\@OpenWrt:/#', '/#', '@OpenWrt:/#', "@OpenWrt:~#"]
```
arnout: I usually say that commits should be split, but this time it should be squashed with the first commit that updates prplmesh_prplwrt.py.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
```diff
@@ -83,10 +115,12 @@ def check_status(self) -> bool:
...
         self.sendline("/etc/init.d/prplmesh status")
-        match = self.expect(
-                ["operational", "FAIL", pexpect.EOF, pexpect.TIMEOUT],
-                timeout=10)
-        if match == 0:
+        self.expect(
+                ["(?P<main_agent>OK) Main agent.+"
+                 "(?P<wlan0>OK) wlan0.+"
+                 "(?P<wlan2>OK) wlan2", pexpect.TIMEOUT],
+                timeout=5)
```
arnout, line 125: You can actually just do
```
self.sendline("/etc/init.d/prplmesh status && echo "FULLY OPERATIONAL")
self.expect(["FULLY OPERATIONAL", pexpect.TIMEOUT], timeout=5)
```

arnout (approved):
This PR is doing too many things, adding too many features, making the whole thing pretty hard to follow...

Given this situation, it's probably better to merge it more or less as-is, and fix things up later. I have a few comments already that you can take into account, but most likely I'll have more architectural comments in future PRs.

In the future, please make the changes in a single PR much smaller and easier to parse.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: This is what actually done.  This part of PR is outdated.

tests/environment.py
abelog: To keep formatting same across file.

ghost (approved):
Approving because I think that with these big PRs, it's generally faster to just get them merged and fix whatever is still wrong or missing in a followup PR.
Please do make sure the open issues get fixed at some point, though

tests/environment.py
arnout: OK. In the future, please mention that in the commit message.

arnout:
> Approving because I think that with these big PRs, it's generally faster to just get them merged and fix whatever is still wrong or missing in a followup PR.

I agree with that, but the reviews should still happen and if it's already merged, *nobody* is going to review it...

Or did you do a full review and had no comments other than what I mentioned already?

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Idea was to keep it generalized. But i don't think that there is a case when we need to get the subnet of some unknown docker network.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: > Although even so, I don't see why it would be needed to add the IP address. No IP connectivity is needed between the devices, so why bother?

Do we require IP connectivity between agent and controller, don't we?

tests/environment.py
abelog: Left during debug.

Thanks for noticing second one.

tests/environment.py
abelog: Applied!

tests/environment.py
abelog: AFAIK, pexpect docs [do not state type of timeout](https://pexpect.readthedocs.io/en/stable/api/popen_spawn.html#popenspawn-class). But from default value i guessed that it's int.

> Where is this centiseconds coming from?

They are coming from `prplmesh_base_test.py ` and `initial_ap_config.py` don't define timeout.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Just remembered, that i have seen that `/etc/init.d/prplmesh status` actually sets appropriate shell exit code. Since functionality is the same, readability same (at least to me) don't see much sense changing.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: SSH is now used as primary connection.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: > I believe "name" is required in the config, so there shouldn't be a need for a default value. If there is one, it should be netgear-rax40, not just rax40.

Fixed.

> Why do you append the unique_id? There can be only one boardfarm instance connected to the device, right? So distinguishing them with a unique_id seems pointless.

Inheritance from Docker device. Don't see a way we may need it.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Added

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Created a task: #1443

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
abelog: Fixed.

tests/environment.py
abelog: Created a task: #1444

tests/environment.py
abelog: Fixed!

tests/environment.py
abelog: Applied. Thank you.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_prplwrt.py
arnout: No we don't. MultiAP is purely on Ethernet level.

tests/environment.py
arnout: > AFAIK, pexpect docs [do not state type of timeout](https://pexpect.readthedocs.io/en/stable/api/popen_spawn.html#popenspawn-class). But from default value i guessed that it's int.

I tested with a timeout of 1.9, and it waited for about two seconds, so it doesn't round down, so it uses float.

> > Where is this centiseconds coming from?
> 
> They are coming from `prplmesh_base_test.py ` and `initial_ap_config.py` don't define timeout.

Both say that it's in seconds, and also pexpect states that it's in seconds. If you multiply a timeout of .9 with 100, you get a timeout of 90 seconds, which is not what we want...

PR #1359, opened 2020-05-29
Bugfix/dwpal wps pbc update

### LINKED ISSUE #413: [TASK] Pass flow 4.10.3 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.10.3 in the certification testbed.

# Exit criteria

rmelotte:
With the new dwpal version, "WPS_PBC <ifname>" is not supported
anymore.

Use WPS_PBC without any interface name instead, to start it on all
configured interfaces.

We also need to set wps_independent to 0 for the vaps we configure.

Fixes https://github.com/prplfoundation/prplMesh/issues/413


rmelotte:
Almost all the known-to-be-passing agent tests were run here, they all passed: https://gitlab.com/prpl-foundation/prplMesh/pipelines/150574459

ghost (approved):

arnout (approved):
I was surprised that the second commit is needed. But indeed, it turns out that in openwrt this option defaults to 1 instead of 0...

PR #1360, opened 2020-05-29
[4.5.2] Channel selection with invalid channel should error out

### LINKED ISSUE #368: 4.5.2 Channel Selection Request Message
The prplMesh agent on Turris Omnia with OpenWrt needs to pass the certification test case 4.5.2 as defined by the Wi-Fi CERTIFIED EasyMesh (tm) Test Plan.

- [x] #366 Channel selection for nl80211
- [x] #725 Perform channel selection
- [x] #369 Implement Operating Channel Report for nl80211
- [ ] #367 Pass flow 4.5.2 test in testbed

### LINKED ISSUE #1349: [BUG] agent: channel selection: error if invalid channel is given [4.5.2]
In certification test 4.5.2, one of the tests is to trigger the controller to send a channel selection request which contains invalid channels. Concretely, one of the values is: `{0x73 0x03 {0x95 0x99 0x9D} 0x00}` i.e. operating class 115, channels 149, 153, 157, preference value 0. Since channels 149, 153, 157 are in operating class 124, not 115, this is invalid.

However, we currently just ignore the invalid channels and only take into account the valid ones. Since no preference value is set for any channel in operating class 115, we consider all of them to be allowed, so we don't switch channels and report success.

The test requires an error (DECLINE_VIOLATES_MOST_RECENTLY_REPORTED_PREFERENCES) in this case.

Example of failing test: https://gitlab.com/prpl-foundation/prplMesh/-/jobs/570137322

arnout:
Test 4.5.2 currently fails because an invalid channel selection ignore is accepted (without switching channels) while it should return an error.

Add a test for this in test_flows.
To simplify this test, start with refactoring the existing channel_selection test to use status information instead of log search.

Fixes: #1349


tests/environment.py
```diff
@@ -84,6 +85,17 @@ def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
         '''Poll the radio's logfile until it contains "regex" or times out.'''
         raise NotImplementedError("wait_for_log is not implemented in abstract class Radio")
 
+    def get_current_channel(self) -> (int, int, int):
+        '''Get the current channel information.
+
+        Returns
+        -------
+        (channel: int, bw: int, center_channel: int)
+            4-tuple with the channel information: channel index, bandwidth as an integer, and
```
ghost: Seems like a 3-tuple to me ...

tests/environment.py
```diff
@@ -270,6 +282,15 @@ def send_bwl_event(self, event: str) -> None:
         command = "echo \"{}\" > /tmp/$USER/beerocks/{}/EVENT".format(event, self.iface_name)
         self.agent.command('sh', '-c', command)
 
+    def read_tmp_file(self, filename: str) -> bytes:
+        '''Read the file given by `filename` from the radio's status directory'''
+        path = "/tmp/$USER/beerocks/{}/{}".format(self.iface_name, filename)
+        return self.agent.command("sh", "-c", 'cat "{}"'.format(path))
```
ghost, line 295: Why use `sh -c "cat file"` rather than simply `cat file`?
```suggestion
        return self.agent.command("cat", '"{}"'.format(path))
```

tests/environment.py
```diff
@@ -84,6 +85,17 @@ def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
         '''Poll the radio's logfile until it contains "regex" or times out.'''
         raise NotImplementedError("wait_for_log is not implemented in abstract class Radio")
 
+    def get_current_channel(self) -> (int, int, int):
+        '''Get the current channel information.
+
+        Returns
+        -------
+        (channel: int, bw: int, center_channel: int)
```
ghost: Suggestion: You could use a named tuple, it's more self-documenting that way (nitpick)

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -280,9 +280,23 @@ bool base_wlan_hal_dummy::set(const std::string &param, const std::string &value
 
 bool base_wlan_hal_dummy::ping() { return true; }
 
-bool base_wlan_hal_dummy::dummy_send_cmd(const std::string &cmd) { return false; }
-
-bool base_wlan_hal_dummy::dummy_send_cmd(const std::string &cmd, char **reply) { return false; }
+bool base_wlan_hal_dummy::write_status_file(const std::string &filename, const std::string &value)
+{
+    auto full_path{get_status_dir() + "/" + filename};
+    auto full_path_tmp{full_path + ".tmp"};
```
ghost, line 299: Best practice (IIRC) is to use `mkstemp` (or similar) for temporary files. No issues with concurrency, security (e.g. people guessing and pre-opening/redirecting the file), file is guaranteed to be on a writable FS, etc

tests/environment.py
arnout: Added comment to explain it.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
arnout: mkstemp has an annoying interface...

Concurrency is not an issue here, if multiple agents are running, we have bigger problems.

Security is also not really an issue since it's only used for testing.

Guaranteed on a writable FS: actually no, because we have to create it in the same FS as where it goes in the end, otherwise the rename is not atomic.

So I choose to ignore this comment. If you insist, I can still do it. Actually maybe it's not that much work, but IMO not really a priority now.

arnout:
I forgot to add the magic certification message in the commit. I've triggered it manually now and fixed my local branch, so next time I push it it should be OK.

tests/environment.py
```diff
@@ -270,6 +285,18 @@ def send_bwl_event(self, event: str) -> None:
...
 
+    def read_tmp_file(self, filename: str) -> bytes:
+        '''Read the file given by `filename` from the radio's status directory'''
+        # The file is only available within the docker container so we need to use cat.
+        # Inside the container, $USER is set to the username that was used for starting it.
+        # We need to pass through shell to expand it.
+        path = "/tmp/$USER/beerocks/{}/{}".format(self.iface_name, filename)
+        return self.agent.command("sh", "-c", 'cat "{}"'.format(path))
+
+    def get_current_channel(self) -> (int, int, int):
```
ghost:
```suggestion
    def get_current_channel(self) -> ChannelInfo:
```
and on line 298:
`return ChannelInfo(channel_info["channel"], channel_info["bw"], channel_info["center_channel"])`

arnout: D'oh

arnout:
It is unfortunately not sufficient to pass 4.5.2, at least not reliably. The problem is that we send two Channel Select Response messages, one for each radio. However, both of them has the same MID, so the transport will drop one of the responses (it's considered a duplicate).

In the error scenario, one radio will have completely ignored the request because there was no TLV with a matching radio ID. Still, it will send a response with success status. The other radio will send an error response, but that response gets dropped by the transport.

The solution is to fix things right by collecting the responses from the radios and sending them all together.

It's very unfortunate we have to do this now, because the unified agent would actually fix it.

arnout:
Created #1364 for the remaining issue, so this PR can be merged.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -128,7 +136,20 @@ bool ap_wlan_hal_dummy::disable() { return true; }
...
-bool ap_wlan_hal_dummy::set_channel(int chan, int bw, int center_channel) { return true; }
+bool ap_wlan_hal_dummy::set_channel(int chan, int bw, int center_channel)
+{
+    m_radio_info.channel = chan;
+    m_radio_info.bandwidth =
+        beerocks::utils::convert_bandwidth_to_int((beerocks::eWiFiBandwidth)bw);
+    m_radio_info.vht_center_freq = center_channel;
+    m_radio_info.is_dfs_channel  = son::wireless_utils::is_dfs_channel(chan);
+    std::stringstream value;
+    value << "channel: " << chan << std::endl;
```
RanRegev, line 147: nitpick, because I don't know what `flush()` means in `std::stringstream` ([example reference](https://youtu.be/GMqQOEZYVJQ))
```suggestion
    value << "channel: " << chan << '\n';    
```

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
```diff
@@ -128,7 +136,20 @@ bool ap_wlan_hal_dummy::disable() { return true; }
...
+    value << "bw: " << m_radio_info.bandwidth << std::endl;
+    value << "center_channel: " << center_channel << std::endl;
+    ;
```
mariomaz:
```suggestion

```

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -280,6 +280,24 @@ bool base_wlan_hal_dummy::set(const std::string &param, const std::string &value
 
 bool base_wlan_hal_dummy::ping() { return true; }
 
+bool base_wlan_hal_dummy::write_status_file(const std::string &filename, const std::string &value)
+{
+    auto full_path{get_status_dir() + "/" + filename};
+    auto full_path_tmp{full_path + ".tmp"};
+    std::ofstream statusfile(full_path_tmp);
+    statusfile << value;
+    if (!statusfile) {
```
mariomaz, line 303: I'd check if file is open before writing to it

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -280,6 +280,24 @@ bool base_wlan_hal_dummy::set(const std::string &param, const std::string &value
...
+        LOG(ERROR) << "Failed writing to " << full_path;
+        return false;
+    }
+    statusfile.close();
+    if (rename(full_path_tmp.c_str(), full_path.c_str()) < 0) {
```
mariomaz, line 307: On failure, a nonzero value is returned.
```suggestion
    if (rename(full_path_tmp.c_str(), full_path.c_str()) != 0) {
```

common/beerocks/bwl/dummy/base_wlan_hal_dummy.h
```diff
@@ -85,6 +85,13 @@ class base_wlan_hal_dummy : public virtual base_wlan_hal,
     virtual bool set(const std::string &param, const std::string &value,
                      int vap_id = beerocks::IFACE_RADIO_ID) override;
 
+    /** @brief Write/update a dummy status file.
+     *
+     * Write the @a value to a status file with name @a filename. This is used by test scripts to
+     * verify if the proper bwl commands were given.
+     */
+    bool write_status_file(const std::string &filename, const std::string &value);
```
mariomaz:
```suggestion
    bool write_status_file(const std::string &filename, const std::string &value) const;
```

tests/test_flows.py
```diff
@@ -568,30 +578,23 @@ def test_channel_selection(self):
...
+                  " step {}".format(i))
 
             self.check_log(env.agents[0].radios[0], "tlvTransmitPowerLimit {}".format(tp20dBm))
-
-            self.check_log(env.agents[0].radios[0],
-                           "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
-
             self.check_log(env.agents[0].radios[1], "tlvTransmitPowerLimit {}".format(tp20dBm))
 
-            self.check_log(env.agents[0].radios[0],
```
mariomaz, line 588: I think this was a bug in the test because the wrong log file was being checked. Nevertheless, the test was passing.
For the sake of curiosity: i think we were going to implement some kind of a cursor in the log. Is it available yet?

tests/test_flows.py
```diff
@@ -507,6 +507,28 @@ def test_channel_selection(self):
...
+
+        cs_resp = self.check_cmdu_type("channel selection response", 0x8007, env.agents[0].mac,
+                                       env.controller.mac, cs_req_mid)
+
+        # TODO should be a single response (currently two are sent)
+        for response in cs_resp:
+            cs_resp_tlv = self.check_cmdu_has_tlv_single(response, 0x8e)
+            if cs_resp_tlv and \
+               cs_resp_tlv.channel_select_radio_id == env.agents[0].radios[0].mac:
+                if int(cs_resp_tlv.channel_select_response_code, 16) != 0x02:
```
mariomaz, line 525: Use another 'and' condition instead of another 'if'

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4841,19 +4850,30 @@ bool slave_thread::handle_channel_selection_request(Socket *sd, ieee1905_1::Cmdu
...
 
-    // According to design only Resticted channels should be included in channel selection request
-    auto switch_required = channel_selection_current_channel_restricted();
+    auto response_code = wfa_map::tlvChannelSelectionResponse::eResponseCode::ACCEPT;
     beerocks::message::sWifiChannel channel_to_switch;
-    if (switch_required) {
-        channel_to_switch = channel_selection_select_channel();
+    bool switch_required = false;
+    if (channel_selection_get_channel_preference(cmdu_rx)) {
+        // Only Resticted channels are be included in channel selection request.
```
mariomaz: typo 'restricted'

mariomaz (approved):
It's a pity we didn't have testable code and unit tests before. This kind of things could be much more easily tested.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
RanRegev, line 303: [this is](https://en.cppreference.com/w/cpp/io/basic_ios/operator_bool) new to me, I have never noticed.
Thanks.

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
```diff
@@ -280,6 +280,24 @@ bool base_wlan_hal_dummy::set(const std::string &param, const std::string &value
...
+{
+    auto full_path{get_status_dir() + "/" + filename};
+    auto full_path_tmp{full_path + ".tmp"};
+    std::ofstream statusfile(full_path_tmp);
+    statusfile << value;
+    if (!statusfile) {
+        LOG(ERROR) << "Failed writing to " << full_path;
+        return false;
+    }
+    statusfile.close();
```
RanRegev: [No need](https://en.cppreference.com/w/cpp/io/basic_ofstream/close) (see the Notes section)
Also, calling `close()` without first calling `open()` breaks symmetry.
I think that symmetry is a good practice in software (no farther details here).

tests/test_flows.py
```diff
@@ -423,29 +423,35 @@ def test_ap_config_bss_tear_down_cli(self):
...
         debug("Send empty channel selection request")
         cs_req_mid = env.controller.dev_send_1905(env.agents[0].mac,
                                                   0x8006, tlv(0x00, 0x0000, "{}"))
         time.sleep(1)
 
-        debug("Confirming empty channel selection request has been received on agent")
-        self.check_log(env.agents[0].radios[0], "CHANNEL_SELECTION_REQUEST_MESSAGE")
-        self.check_log(env.agents[0].radios[1], "CHANNEL_SELECTION_REQUEST_MESSAGE")
-
         # TODO should be a single response (currently two are sent)
```
RanRegev, line 444: Is this still a relevant comment?

tests/test_flows.py
```diff
@@ -568,30 +578,23 @@ def test_channel_selection(self):
...
-            self.check_log(env.agents[0].radios[0],
-                           "ACTION_APMANAGER_HOSTAP_CHANNEL_SWITCH_ACS_START")
-
             # TODO should be a single response (currently two are sent)
             self.check_cmdu_type("channel selection response", 0x8007, env.agents[0].mac,
                                  env.controller.mac, cs_req_mid)
 
+            (cur_chan_channel_0, _, _) = env.agents[0].radios[0].get_current_channel()
+            (cur_chan_channel_1, _, _) = env.agents[0].radios[1].get_current_channel()
+            if cur_chan_channel_0 != 6:
```
RanRegev, line 616: This is probably a nitpick, but it is not clear to me where this `6` came from.
there is only a [comment](https://github.com/prplfoundation/prplMesh/pull/1360/commits/d32039cacc2b61f12ee2670587d77df6c6f44ee7#diff-f19fb9f58c705ea594e8e5c22bea2d5aR510) referencing it.
I don't know if using some variable would help here, because by just looking at the code I can't tell where this `6` can be replaced with a variable in the original `payload_wlan0`.

agent/src/beerocks/slave/son_slave_thread.cpp
```diff
@@ -4795,6 +4797,13 @@ bool slave_thread::channel_selection_get_channel_preference(ieee1905_1::CmduMess
...
                 }
 
+                // Check if channel is valid for operating class
+                if (op_class_chan_set.find(*channel) == op_class_chan_set.end()) {
+                    LOG(ERROR) << "Channel " << +*channel << " invalid for operating class "
+                               << +operating_class;
+                    return false;
+                }
+
                 ss << int(*channel);
```
RanRegev, line 4807: [I guess (line 10)](https://cppinsights.io/s/2fe9eb81) the same trick will work here too, but need to verify:
```suggestion
                ss << +(*channel);
```

RanRegev (approved):
only nitpicks

VladyslavTupikin (approved):

tools/docker/tests-runner/Dockerfile
```diff
@@ -3,9 +3,10 @@ FROM docker.io/library/docker@sha256:a490c83561c1cef49b6fe12aba2c31f908391ec3efe
 
 RUN apk add --update --no-cache \
     bash \
-    python3 \
-    grep \
     coreutils \
+    grep \
+    py3-yaml \
```
rmelotte, line 8: The commit message mentions it's installed using pip while here it's not (nitpick).

rmelotte (approved):

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
ghost: `man 2 rename`

```
RETURN VALUE
       On success, zero is returned.  On error, -1 is returned, and errno is set appropriately.
```

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
arnout: Why check before writing? If the file wasn't opened successfully, the write will simply fail and the stream stays in error state. Does it really matter whether it was the open or the writing that failed? We do check the result in this condition, I don't see why we would complicate things by adding a second condition before.

tests/test_flows.py
arnout: > I think this was a bug in the test because the wrong log file was being checked. Nevertheless, the test was passing.

Yes, well, that's forever the problem when writing the test after the code: how do you know that the test is really testing what it should?

> For the sake of curiosity: i think we were going to implement some kind of a cursor in the log. Is it available yet?

It is available, but the interface is clunky (you have to store the line number in the check_log return value). The idea is to instead store the line number in the checkpoint() function. Note that the checkpoint() function is currently already invoked in this particular test. And note that that wouldn't have caught this issue, because we don't checkpoint between the two log checks (would be a bad idea to do that, since it forces you to rely on the order of logs).

tests/test_flows.py
arnout: I find it clearer to separate the actual assertion condition from the condition(s) that verify the context. But it's a thin line.

Anyway, this code will either go away or change in #1383 so I'm not going to bother fixing it.

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
arnout: Sorry, I can't be bother to go through 11 minutes of chatter to get a simple point. Could you summarize the argument he is trying to make?

(note: we're currently using endl everywhere so it's quite a big thing you're proposing)

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
mariomaz: Ok, I see your point and agree. It was just a question of habit.
However, you should probably check error state after closing file, once buffers have been flushed and all information actually written to disk.

tests/test_flows.py
mariomaz: “Program testing can be used to show the presence of bugs, but never to show their absence!”
― Edsger W. Dijkstra

common/beerocks/bwl/dummy/base_wlan_hal_dummy.cpp
arnout: We have to close the file before the rename. The whole point of using a temporary file is to make sure it's atomic, i.e. the file is either the old one or the new one, not something in-between. If I don't close the file, then it's possible that the rename happens before the data is actually flushed to the file, and that a reader in another process sees an empty file.

Note that to be fully atomic, we should actually also do an fsync on the directory before the rename, but that's just making sure that the metadata is updated and we're not that interested in the metadata.

Also, we should be doing an fsync on the file itself. But again in this particular case it's not needed because we don't need atomicity over reboots, only between processes.

It might actually be sufficient to do a flush rather than close, but I feel safer this way.

And finally: the same could also be achieved by putting the statusfile in an inner scope, so its lifetime ends at this point. However, I think that that is even more confusing and error prone than an explicit clone.

That said, a comment would really be a good idea here.

tests/test_flows.py
arnout: Yes, #1383 fixes this.

tests/test_flows.py
arnout: The problem is that converting a channel number to the payload is quite complicated, not something you'd want in test code...

Maybe an additional comment would help here.

agent/src/beerocks/slave/son_slave_thread.cpp
arnout: Yes it would. However, with #1392 that wouldn't even be needed.

Regardless, this is code that is not changed by this PR so I'm not touching it :-)

common/beerocks/bwl/dummy/ap_wlan_hal_dummy.cpp
RanRegev: yes, sure:
`std::endl` does more than just adding a `new-line` to the stream. It actually asks the stream to `flush` the already collected data into the underlying media.
So when it comes to disk it takes _much_ longer to `std::endl` instead of `\n`.
What I mentioned that I have no idea what does this extra work actually do when it comes to `std::stringstream` - it might end up with a simple nop (no operation).

arnout: Indeed for strstream it's a nop.

And the flush behaviour is usually what you want. We, at least, use it only in the context of logging (and in beerocks_cli for user interaction). Line-buffered is exactly what you want in those cases.

PR #1365, opened 2020-05-30
bcl: wireless_unit_test: use parameters for GetOperatingClassByChannelTest

arnout:
Instead of open-coding the test with a few boundary values, use a
parameterized test and cover the entire range.

This was requested in https://github.com/prplfoundation/prplMesh/pull/1347#discussion_r431866359

I'm not entirely sure if this parameterized test is really better than open-coding a loop over all possibilities.


arnout:
Again, I'm not happy with this pull request...
- I don't think the code is very readable with the parametrization.
- Due to the complicated template instantiation, it's pretty dependent on compiler version (cfr. build failure in https://gitlab.com/prpl-foundation/prplMesh/-/jobs/574242569).
- The parametrization doesn't help to be sure we cover all cases - in fact, I know there are two 160MHz cases that are still missing in the failure test. Note that I used a different approach for "generating" the success and the failure test cases, so we have an example of both and can decide which one is better.
- Although we have individual unit tests now, they have unwieldy names (e.g. "get_operating_class_by_channel_should_return_valid_channels_100_140/WirelessUtilsGetOperatingClassByChannelTest.should_return_expected/6" and errors are difficult to interpret, e.g.
```
Expected equality of these values:
  son::wireless_utils::get_operating_class_by_channel(ch)
    Which is: '\x80' (128)
  0
[  FAILED  ] get_operating_class_by_channel_invalid_channel_80mhz/WirelessUtilsGetOperatingClassByChannelInvalidTest.should_fail/98, where GetParam() = ('d' (100, 0x64), 3) (0 ms)
```
(note that that is partly due to the uint8_t, which isn't great to print...).

Since the whole point of this PR is to make a good example of value-parametrized unit tests, I think it's essential to improve it. But I don't know how.

Personally, I think the approach taken for the failures is easier to read, because you can follow how we cover all the channels. However, I don't see any advantage of using this compared to using EXPECT directly in the function. It *would* make sense if we could do some other tests (of other functions) using the same data, but I didn't readily see how to do that.

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -13,120 +13,265 @@
...
+};
+
+// clang-format off
+const std::vector<sOperatingClass>
+    table_E_4_global_operating_classes_802_11{
+	{ 81, beerocks::BANDWIDTH_20, {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13}},
+	{ 82, beerocks::BANDWIDTH_20, {14}},
+	{ 83, beerocks::BANDWIDTH_40, {1, 2, 3, 4, 5, 6, 7, 8, 9}},
+	/* Note: we have no way to distinguish 40- and 40+, so operating class 83 is used for all bandwidth=40MHz */
+    { 84, beerocks::BANDWIDTH_40, {/*5, 6, 7, 8, 9, */10, 11, 12, 13}},
```
arnout: Whitespace issue here?

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -38,8 +38,8 @@ static const std::map<uint8_t, sOperatingClass> operating_classes_list = {
     {119,       {{52, 60},                                                     beerocks::BANDWIDTH_40}},
     {120,       {{56, 64},                                                     beerocks::BANDWIDTH_40}},
     {121,       {{100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144}, beerocks::BANDWIDTH_20}},
-    {122,       {{100, 108, 116, 124, 132, 140},                               beerocks::BANDWIDTH_40}},
-    {123,       {{104, 112, 120, 128, 136, 144},                               beerocks::BANDWIDTH_40}},
+    {122,       {{100, 108, 116, 124, 132, 136, 140},                          beerocks::BANDWIDTH_40}},
+    {123,       {{104, 112, 120, 128, 134, 136, 138, 144},                     beerocks::BANDWIDTH_40}},
```
arnout, line 42: This fix should be in a separate commit.

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -13,120 +13,265 @@
...
+std::vector<tGetOperatingClassByChannelParam> get_operating_class_by_channel_invalid_parameters()
+{
+    // clang-format off
+    const beerocks::eWiFiBandwidth bandwidths[]{
+        beerocks::BANDWIDTH_UNKNOWN,
+		beerocks::BANDWIDTH_20,
+        beerocks::BANDWIDTH_40,
+		beerocks::BANDWIDTH_80,
+        beerocks::BANDWIDTH_80_80,
+		beerocks::BANDWIDTH_160
```
arnout: This indentation is also weird. I guess intentional, but I don't understand why...

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -13,120 +13,265 @@
...
 
-    EXPECT_EQ(get_operating_class_by_channel_helper(40, 40), 117);
-    EXPECT_EQ(get_operating_class_by_channel_helper(48, 40), 117);
+    /**
+     * Note: Google Test documentation says that using value-parameterized tests to test your code
+     * over various inputs (a.k.a. data-driven testing) is a feature easy to abuse of.
+     * This might well be a good example of such an abuse.
+     */
+    uint8_t channel = 0;
+    do {
```
arnout, line 152: Why do-while instead of for loop?

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -13,120 +13,265 @@
...
+    auto bandwidth = static_cast<beerocks::eWiFiBandwidth>(value);
 
-    EXPECT_EQ(get_operating_class_by_channel_helper(165, 20), 125);
-    EXPECT_EQ(get_operating_class_by_channel_helper(169, 20), 125);
+    const auto &it = bandwidths.find(bandwidth);
+    if (bandwidths.end() == it) {
+        result += "invalid_" + std::to_string(value);
+    } else {
+        result += it->second;
+    }
```
arnout: (nitpick)
I would have:
- used the full strings directly, rather than += (i.e. `{beerocks::BANDWIDTH_20, "bandwidth_20"},`);
- used a switch rather than a lookup table.

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -13,120 +13,265 @@
...
+ * @brief Converts a parameter of a get_operating_class_by_channel() value-parameterized test to
+ * its string representation.
+ *
+ * Helper function to create a custom test name suffix based on the test parameter value.
+ *
+ * @param param value-parameterized test parameter.
+ * @return Value-parameterized test parameter as a string.
+ */
+std::string operating_class_by_channel_param_to_string(
+    const testing::TestParamInfo<WirelessUtilsGetOperatingClassByChannelTest::ParamType> &info)
```
arnout: Isn't this just
```suggestion
    const testing::TestParamInfo<tGetOperatingClassByChannelParam> &info)
```
?

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -13,120 +13,265 @@
...
+{
+    const tGetOperatingClassByChannelParam &param = info.param;
```
arnout:
```suggestion
    const auto &param = info.param;
```

arnout (commented):
Looks *much* better than what I produced. I have a few nitpick comments, but nothing untowards. Of couse, I can't approve myself.

I'm a bit concerned that much of the logic is a repetition of the actual implementation, but since they're independently implemented I guess it's OK.

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
```diff
@@ -13,120 +13,265 @@
...
+    {121, beerocks::BANDWIDTH_20, {100, 104, 108, 112, 116, 120, 124, 128, 132, 136, 140, 144}},
+    {122, beerocks::BANDWIDTH_40, {100, 108, 116, 124, 132, 136, 140}},
+    {123, beerocks::BANDWIDTH_40, {104, 112, 120, 128, 134, 138, 144}},
+    {124, beerocks::BANDWIDTH_20, {149, 153, 157, 161}},
+	/* Note: we have no way to distinguish repeated channels in operating classes 124 and 125 used for bandwidth=20MHz */
+    {125, beerocks::BANDWIDTH_20, {/*149, 153, 157, 161,*/ 165, 169}},
+    {126, beerocks::BANDWIDTH_40, {149, 157}},
+    {127, beerocks::BANDWIDTH_40, {153, 161}},
+    {128, beerocks::BANDWIDTH_80, {42, 58, 106, 122, 138, 155}},
+	{129, beerocks::BANDWIDTH_160, {50, 114}},
```
ghost:
```suggestion
    {129, beerocks::BANDWIDTH_160, {50, 114}},
```

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
ghost: Looks like tabs vs spaces, and github uses a different tab size than Mario's editor is configured for.

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
ghost: So that the loop executes up until and INCLUDING `channel == UINT8_MAX`, I guess (`for (uint8_t channel = 0; channel <= UINT8_MAX; ++channel)` would loop forever)

arnout: Ah of course... If we make channel an unsigned instead of a uint8_t, that problem goes away.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
mariomaz: If I put the fix in a separate commit, then that commit will not pass existing unit tests. 
I think it's better to fix both the code and the unit tests in the same commit because they are tightly coupled. As a bonus, you can revert changes more easily if required. 

I can move the fix to a separate commit if there's a better reason to do it that I cannot think of. This is something to be decided and consistently applied because if we start writing unit tests, then it's going to happen very often.

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
mariomaz: Sorry, my bad. You are right: I recently changed my editor and I didn't realize it was configured to use tabs instead of spaces by default. 
Maybe clang-format should take care of this and replace tabs with 4 spaces. 

Fixed

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
mariomaz: Exactly!
I'm going to add a comment to clarify it

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
mariomaz: Some static analyzers include a rule by which `switch` statements should not have too many `case` clauses: "When switch statements have large sets of case clauses, it is usually an attempt to map two sets of data. A real map structure would be more readable and maintainable, and should be used instead."

For the sake of curiosity, why do you prefer a switch rather than a lookup table?

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
mariomaz: Done

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
mariomaz: Fixed

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
mariomaz: Yes it is. Fixed

rmelotte (approved):

common/beerocks/bcl/source/son/son_wireless_utils.cpp
arnout: Ah, I forgot that the current test is incorrect as well...

common/beerocks/bcl/unit_tests/wireless_utils_test.cpp
arnout: Because I think a switch is more readable and maintainable :-)

Many lint tools and even gcc with -Wswitch enabled (which is part of -Wall) will flag when an enum value is missing in a switch. AFAIK, the same thing doesn't exist in a look-up table.

In terms of readability, I think a switch is pretty readable. It would be nicer if there was a switch expression though... But I guess indeed a look-up table is more readable, at least if it's easily representable as a look-up table (which is the case here).

ghost (approved):

PR #1367, opened 2020-05-30
[RFC] framework: remove examples

arnout:
The framework contained two examples: a transport example with a
skeleton agent and controller, and a zmq example that didn't have any
code yet.

Since we now have an actual agent and controller, the transport example
is not really useful any more.

Since we're not planning on staying with zmq, the zmq example will never
be added.

Remove the examples directory, as well as its reference in
CMakeLists.txt.

Remove the BUILD_EXAMPLES CMake option, as well as its reference in
README.md.


RanRegev (approved):
I understand that this code is not maintained and irrelevant anymore.
My only concern is this: if we ever want to write an example code - is the deleted code a good base to start with?
If the answer is yes, we need to keep a reference somehow to the fact it ever existed.
(On the other hand - it is not hard, any point on master before this `hotfix` would be enough)

morantr (approved):

PR #1368, opened 2020-05-31
Hotfix/fix missing ip addr on monitor rssi

morantr:
Add an ability to pull the IP address of a client from the arp table
on montiror_rssi if it is missing.

### DIFF
--- a/agent/src/beerocks/monitor/monitor_rssi.cpp
+++ b/agent/src/beerocks/monitor/monitor_rssi.cpp
@@ -315,6 +315,21 @@ void monitor_rssi::process()
             bool is_4addr_client    = (sta_bridge_4addr_mac != network_utils::ZERO_MAC_STRING);
             std::string arp_dst_mac = is_4addr_client ? sta_bridge_4addr_mac : sta_mac;
 
+            auto sta_ipv4 = sta_node->get_ipv4();
+            if (sta_ipv4.empty() || sta_ipv4 == network_utils::ZERO_IP_STRING) {
+                LOG(DEBUG) << "Sta " << sta_mac << " IP is missing, looking at the ARP Table";
+                auto arp_table = network_utils::get_arp_table();
+                if (arp_table) {
+                    auto arp_entry_it = arp_table->find(sta_mac);
+                    if (arp_entry_it != arp_table->end()) {
+                        sta_ipv4 = arp_entry_it->second;
+                        LOG(DEBUG) << "Found IP on ARP Table, setting Sta " << sta_mac << " IP to "
+                                   << sta_ipv4;
+                        sta_node->set_ipv4(arp_entry_it->second);
+                    }
+                }
+            }
+
             LOG(DEBUG) << "state: SEND_ARP -> "
                        << (sta_node->get_arp_burst() ? "WAIT_FIRST_REPLY" : "WAIT_REPLY")
                        << ", arp_iface = " << arp_iface << ", arp_iface_ipv4 = " << arp_iface_ipv4
--- a/cmake/Findnl-route-3.cmake
+++ b/cmake/Findnl-route-3.cmake
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL_ROUTE_3_LIBRARY "libnl-route-3.so")
+find_path(NL_ROUTE_3_INCLUDE_DIRS netlink/route/route.h PATH_SUFFIXES libnl3)
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(nl-route-3 DEFAULT_MSG
+    NL_ROUTE_3_LIBRARY
+    NL_ROUTE_3_INCLUDE_DIRS
+)
+
+if (nl-route-3_FOUND)
+    add_library(nl-route-3 SHARED IMPORTED)
+
+    # Includes
+    set_target_properties(nl-route-3 PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${NL_ROUTE_3_INCLUDE_DIRS}/"
+    )
+
+    # Library
+    set_target_properties(nl-route-3 PROPERTIES
+        IMPORTED_LINK_INTERFACE_LANGUAGES "C"
+        IMPORTED_LOCATION "${NL_ROUTE_3_LIBRARY}"
+    )
+
+endif()
--- a/common/beerocks/bcl/CMakeLists.txt
+++ b/common/beerocks/bcl/CMakeLists.txt
@@ -4,6 +4,8 @@ project(bcl VERSION ${prplmesh_VERSION})
 set(MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
 
 find_package(PythonInterp REQUIRED)
+find_package(nl-3 REQUIRED)
+find_package(nl-route-3 REQUIRED)
 
 message(STATUS ${CMAKE_CURRENT_BINARY_DIR})
 
@@ -23,7 +25,7 @@ target_include_directories(${PROJECT_NAME}
         $<INSTALL_INTERFACE:include>
 )
 
-target_link_libraries(${PROJECT_NAME} PUBLIC tlvf elpp mapfcommon)
+target_link_libraries(${PROJECT_NAME} PUBLIC nl-3 nl-route-3 tlvf elpp mapfcommon)
 
 install(TARGETS ${PROJECT_NAME} EXPORT bclConfig
     ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
@@ -58,7 +60,7 @@ if (BUILD_TESTS)
         PUBLIC
             $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
     )
-    target_link_libraries(${TEST_PROJECT_NAME} tlvf elpp mapfcommon)
+    target_link_libraries(${TEST_PROJECT_NAME} nl-3 nl-route-3 tlvf elpp mapfcommon)
     target_link_libraries(${TEST_PROJECT_NAME} gtest_main)
     install(TARGETS ${TEST_PROJECT_NAME} DESTINATION bin/tests)
     add_test(NAME ${TEST_PROJECT_NAME} COMMAND $<TARGET_FILE:${TEST_PROJECT_NAME}>)
--- a/common/beerocks/bcl/include/bcl/network/network_utils.h
+++ b/common/beerocks/bcl/include/bcl/network/network_utils.h
@@ -94,6 +94,19 @@ public:
     static int get_iface_info(network_utils::iface_info &info, const std::string &iface_name);
     static bool get_raw_iface_info(const std::string &iface_name, raw_iface_info &info);
 
+    /**
+     * @brief Get the arp table as unordered map object of pairs <MAC, IP> or <IP, MAC>.
+     * The key of the map can be either a MAC address or an IP address, depends on the 'mac_as_key'
+     * argument value.
+     * If 'mac_as_key' is 'true' then the key is a MAC address, otherwise the key is an IP address.
+     * 
+     * @param[in] mac_as_key Decide whether the key of the returned unordered map object 
+     * is a MAC or an IP.
+     * @return std::shared_ptr<std::unordered_map<std::string, std::string>> 
+     */
+    static std::shared_ptr<std::unordered_map<std::string, std::string>>
+    get_arp_table(bool mac_as_key = true);
+
     //temp
     static std::string get_mac_from_arp_table(const std::string &ipv4);
 
--- a/common/beerocks/bcl/source/network/network_utils.cpp
+++ b/common/beerocks/bcl/source/network/network_utils.cpp
@@ -25,6 +25,7 @@
 #include <net/if.h>
 #include <netinet/in.h>
 #include <netinet/ip_icmp.h>
+#include <netlink/route/neighbour.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -39,6 +40,7 @@ using namespace beerocks::net;
 
 #define NL_BUFSIZE 8192
 #define MAC_ADDR_CHAR_SIZE 17
+#define IPV4_ADDR_CHAR_SIZE 15
 
 struct route_info {
     struct in_addr dstAddr;
@@ -361,6 +363,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
         return (2);
     }
     return (0);
+}
+
+std::shared_ptr<std::unordered_map<std::string, std::string>>
+network_utils::get_arp_table(bool mac_as_key)
+{
+    auto arp_table_ptr = std::make_shared<std::unordered_map<std::string, std::string>>();
+    auto &arp_table    = *arp_table_ptr.get();
+
+    // Allocate a new netlink socket
+    std::unique_ptr<nl_sock, decltype(&nl_socket_free)> nl_socket(nl_socket_alloc(),
+                                                                  &nl_socket_free);
+    if (!nl_socket) {
+        LOG(ERROR) << "Failed allocating netlink socket!";
+        return nullptr;
+    }
+
+    int err = nl_connect(nl_socket.get(), NETLINK_ROUTE);
+    if (err != 0) {
+        LOG(ERROR) << "Failed connecting the netlink socket: " << nl_geterror(err);
+        return nullptr;
+    }
+
+    // ARP table cache
+    struct nl_cache *neighbor_table_cache = nullptr;
+
+    // Allocate the cache and fill it with data
+    err = rtnl_neigh_alloc_cache(nl_socket.get(), &neighbor_table_cache);
+    if (err != 0) {
+        LOG(ERROR) << "Failed probing ARP table: " << nl_geterror(err);
+        return nullptr;
+    }
+
+    auto entries_count = nl_cache_nitems(neighbor_table_cache);
+    if (entries_count <= 0) {
+        return arp_table_ptr;
+    }
+
+    auto neighbor_entry = reinterpret_cast<rtnl_neigh *>(nl_cache_get_first(neighbor_table_cache));
+    if (!neighbor_entry) {
+        LOG(ERROR) << "neighbor_entry is null";
+        nl_cache_free(neighbor_table_cache);
+        return nullptr;
+    }
+
+    char buffer_mac[MAC_ADDR_CHAR_SIZE + 1]; // +1 for null terminator
+    char buffer_ip[IPV4_ADDR_CHAR_SIZE + 1]; // +1 for null terminator
+
+    for (int i = 0; i < entries_count; i++) {
+
+        if (neighbor_entry == nullptr) {
+            break;
+        }
+
+        // Skip NOARP state and non IPv4 records
+        if (rtnl_neigh_get_state(neighbor_entry) == NUD_NOARP ||
+            rtnl_neigh_get_family(neighbor_entry) != AF_INET) {
+            // Advance to the next record
+            neighbor_entry = reinterpret_cast<rtnl_neigh *>(
+                nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)));
+            continue;
+        }
+
+        auto nl_addr_mac = rtnl_neigh_get_lladdr(neighbor_entry);
+        if (!nl_addr_mac) {
+            // Advance to the next record
+            neighbor_entry = reinterpret_cast<rtnl_neigh *>(
+                nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)));
+            continue;
+        }
+
+        auto nl_addr_ip = rtnl_neigh_get_dst(neighbor_entry);
+        if (!nl_addr_ip) {
+            // Advance to the next record
+            neighbor_entry = reinterpret_cast<rtnl_neigh *>(
+                nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)));
+            continue;
+        }
+
+        char *mac = nl_addr2str(nl_addr_mac, buffer_mac, sizeof(buffer_mac));
+        char *ip  = nl_addr2str(nl_addr_ip, buffer_ip, sizeof(buffer_ip));
+
+        if (mac_as_key) {
+            arp_table[mac] = ip;
+        } else {
+            arp_table[ip] = mac;
+        }
+
+        // Advance to the next record
+        neighbor_entry = reinterpret_cast<rtnl_neigh *>(
+            nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)));
+    }
+
+    // Free the cache
+    nl_cache_free(neighbor_table_cache);
+
+    return arp_table_ptr;
 }
 
 std::string network_utils::get_mac_from_arp_table(const std::string &ipv4)
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -14,7 +14,7 @@ RUN apt-get update \
     && apt-get -y install build-essential cmake cppcheck valgrind \
     # Libraries
     libjson-c-dev python-yaml libzmq3-dev libssl-dev \
-    libncurses-dev libreadline-dev libnl-3-dev libnl-genl-3-dev \
+    libncurses-dev libreadline-dev libnl-3-dev libnl-route-3-dev libnl-genl-3-dev \
     # Install Helper tools
     bash-completion locales ninja-build pkg-config shellcheck \
     flake8 clang-format vim \
--- a/tools/docker/all/Dockerfile
+++ b/tools/docker/all/Dockerfile
@@ -25,6 +25,8 @@ RUN apt-get update && apt-get install -y \
 	libncurses-dev \
 	libssl-dev \
 	libjson-c-dev \
+	libnl-3-dev \
+	libnl-route-3-dev \
 	libnl-genl-3-dev \
 	libzmq3-dev \
 	python \
--- a/tools/docker/builder/ubuntu/bionic/Dockerfile
+++ b/tools/docker/builder/ubuntu/bionic/Dockerfile
@@ -12,6 +12,8 @@ RUN apt-get update && apt-get install -y \
     git \
     libjson-c-dev \
     libncurses-dev \
+    libnl-3-dev \
+    libnl-route-3-dev \
     libnl-genl-3-dev \
     libreadline-dev \
     libssl-dev \
--- a/tools/docker/builder/ubuntu/xenial/Dockerfile
+++ b/tools/docker/builder/ubuntu/xenial/Dockerfile
@@ -11,6 +11,8 @@ RUN apt-get update && apt-get install -y \
     git \
     libjson-c-dev \
     libncurses-dev \
+    libnl-3-dev \
+    libnl-route-3-dev \
     libnl-genl-3-dev \
     libreadline-dev \
     libssl-dev \
--- a/tools/docker/runner/Dockerfile
+++ b/tools/docker/runner/Dockerfile
@@ -16,6 +16,8 @@ RUN apt-get update && apt-get install -y \
 	libncurses-dev \
 	libssl-dev \
 	libjson-c-dev \
+	libnl-3-dev \
+	libnl-route-3-dev \
 	libnl-genl-3-dev \
 	libzmq3-dev \
 	python \


cmake/Findnl-route-3.cmake
```diff
@@ -0,0 +1,31 @@
...
+find_path(NL3_INCLUDE_DIRS netlink/route/route.h PATH_SUFFIXES libnl3)
+
+include(FindPackageHandleStandardArgs)
+FIND_PACKAGE_HANDLE_STANDARD_ARGS(nl-route-3 DEFAULT_MSG
+    NL3_ROUTE_LIBRARY
+    NL3_INCLUDE_DIRS
+)
+
+if (nl-route-3_FOUND)
+    add_library(nl-route-3 UNKNOWN IMPORTED)
```
RanRegev: [UNKNOWN ](https://cmake.org/cmake/help/latest/command/add_library.html#imported-libraries) means that we might link against static library or shared library.
I think we should be deterministic and set it explicitly.
```suggestion
    add_library(nl-route-3 SHARED IMPORTED)
```

common/beerocks/bcl/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(bcl VERSION ${prplmesh_VERSION})
 set(MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
 
 find_package(PythonInterp REQUIRED)
+find_package(nl-3 REQUIRED)
```
RanRegev, line 7: Adding a required version is normally a good thing to do.
```suggestion
find_package(nl-3 3.2.25 REQUIRED)
```
I think it would work well with [the unversioned find_library](https://github.com/prplfoundation/prplMesh/pull/1368/commits/bab7496200354bd101ef5d0002965631d846f126#diff-3bf015b06ea8efb6ff553fcf349e0352R8) as it would probably find also the symlinks. But we need to verify this behavior.
[cmake find_library](https://cmake.org/cmake/help/latest/command/find_library.html?highlight=find_library)

common/beerocks/bcl/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(bcl VERSION ${prplmesh_VERSION})
...
+find_package(nl-route-3 REQUIRED)
```
RanRegev, line 8:
```suggestion
find_package(nl-route-3 3.2.25 REQUIRED)
```

common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -94,6 +94,8 @@ class network_utils {
     static int get_iface_info(network_utils::iface_info &info, const std::string &iface_name);
     static bool get_raw_iface_info(const std::string &iface_name, raw_iface_info &info);
 
+    static std::unordered_map<std::string, std::string> &get_arp_table(bool mac_as_key = true);
```
RanRegev: comment is missing - map from what `std::string` to what `std::string`? what is the key, what is the value?
yes, it is an arp table, but is is `mac to ip` or `ip to mac`?
and what does it mean if I call it with `mac_as_key = false`?
And why are they represented in strings and not in their "binary" format (sMacAddr, and ?? for ip)?
Please at least add a comment to explain it.

common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -94,6 +94,8 @@ class network_utils {
...
+
     //temp
```
RanRegev, line 110: either remove the `temp` or explain why it is temp and what should replace it and why.
In short - this `//temp` does not provide any useful information to the reader
(I know it was here before the change, but it might be done "while you are in the neighborhood", maybe as a separate commit)

kantera800 (approved):
This PR hides the real problem , that DHCP notification didn't update client's IP #1371 or was not received at all  #1372
On the other hand if it works i think we should merge it anyway.
So approving.

RanRegev (changes requested):

common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -94,6 +94,9 @@ class network_utils {
     static int get_iface_info(network_utils::iface_info &info, const std::string &iface_name);
     static bool get_raw_iface_info(const std::string &iface_name, raw_iface_info &info);
 
+    static std::shared_ptr<std::unordered_map<std::string, std::string>>
```
RanRegev, line 107: 1. The interface is not clear.
Please add a comment what is the key and what is the data. (mac to ip or ip to mac).
2. Also it is better not to use string as key,value pairs, but rather use sMacAdds and whatever ip representation there is.

common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -94,6 +94,9 @@ class network_utils {
...
+    get_arp_table(bool mac_as_key = true);
```
RanRegev, line 108: The bool mac_as_key is non natural, counter intuitive and would cause bugs.
It is best to have two functions:
`get_arp_table_mac_key()`
`get_arp_table_ip_key()`
Or anything that is clear as this. You may _implement_ it using one function, but this function would be private, not part of the interface.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
     return (0);
 }
 
+std::shared_ptr<std::unordered_map<std::string, std::string>>
+network_utils::get_arp_table(bool mac_as_key)
+{
+    auto arp_table_ptr = std::make_shared<std::unordered_map<std::string, std::string>>();
+    auto arp_table     = *arp_table_ptr.get();
```
RanRegev: Here you create a copy of the just created unordered_map.
It seems to be wrong as you assign to it in [436](https://github.com/prplfoundation/prplMesh/pull/1368/commits/02696b3df302a2b283868c4286ceb764c75cc591#diff-97a22c96f0d7fa3e45d0d435e03adc8dR436) and [438](https://github.com/prplfoundation/prplMesh/pull/1368/commits/02696b3df302a2b283868c4286ceb764c75cc591#diff-97a22c96f0d7fa3e45d0d435e03adc8dR438) but never assign back to the pointer.
I suggest either use the pointer only or:
```suggestion
    auto& arp_table     = *arp_table_ptr.get();
```
For more details, paste the following code into https://cppinsights.io:
```cpp
#include <memory>
#include <unordered_map>
#include <string>

int main()
{
    auto arp_table_ptr = std::make_shared<std::unordered_map<std::string, std::string>>();
    auto arp_table     = *arp_table_ptr.get();
  
    auto int_ptr = std::make_shared<int>();
    auto int_    = *int_ptr.get();
}
```

common/beerocks/bcl/CMakeLists.txt
morantr: Done.

common/beerocks/bcl/CMakeLists.txt
morantr: Done.

common/beerocks/bcl/include/bcl/network/network_utils.h
morantr: Added a Doxygen to the function on the h file.
I have decided to use string since I wanted to have the flexibility to replace between the MAC and IP on the map object.

common/beerocks/bcl/include/bcl/network/network_utils.h
morantr: I prefer to not change it in this PR.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+
+    // Allocate a new netlink socket
+    auto nl_socket = nl_socket_alloc();
```
RanRegev: It is best to declare a unique_ptr with custom `deleter()`, then later you don't have to worry about freeing the memory manually -
```suggestion
    std::unique_ptr<std::remove_pointer<decltype(nl_socket_alloc())>::type, std::function<void(int*)>> 
        int_ptr(nl_socket_alloc(), nl_socket_free);
```
take a look here:
https://godbolt.org/z/DYBhRf

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+    err = rtnl_neigh_alloc_cache(nl_socket, &neighbor_table_cache);
+    if (err != 0) {
+        LOG(ERROR) << "Failed probing ARP table: " << nl_geterror(err);
+        nl_socket_free(nl_socket);
+        return nullptr;
+    }
+
+    nl_socket_free(nl_socket);
+
+    auto enteries_count = nl_cache_nitems(neighbor_table_cache);
```
RanRegev: the same comments as for `nl_socket`

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+        if (nl_addr_mac == nullptr) {
+            continue;
+        }
+
+        auto nl_addr_ip = rtnl_neigh_get_dst(neighbor_entry);
+        if (nl_addr_ip == nullptr) {
+            continue;
+        }
+
+        char *mac = nl_addr2str(nl_addr_mac, buffer_mac, sizeof(buffer_mac));
```
RanRegev, line 444: Isn't it possible to initialize `std::string` with the return value?
```suggestion
        std::string mac = nl_addr2str(nl_addr_mac, buffer_mac, sizeof(buffer_mac));
```

RanRegev (changes requested):
Because the functionality seems to be broken (filling a local copy instead of the pointer) I would like to re-review.

common/beerocks/bcl/include/bcl/network/network_utils.h
morantr: I added the Doxygen in the fixup. NVM I squashed them.
I prefer to keep the string as key-value.

common/beerocks/bcl/CMakeLists.txt
vitalybu: I don't we should specify a specific version, as different platforms may have different versions.

common/beerocks/bcl/include/bcl/network/network_utils.h
morantr: I don't think it is that important and I prefer to keep it as it for now.
If more people will share your opinion, I might change it.

common/beerocks/bcl/CMakeLists.txt
```diff
@@ -17,13 +19,16 @@ add_library(${PROJECT_NAME} ${bcl_sources})
 set_target_properties(${PROJECT_NAME} PROPERTIES VERSION ${prplmesh_VERSION} SOVERSION ${prplmesh_VERSION_MAJOR})
 
 # Includes
+include_directories(${PLATFORM_INCLUDE_DIR}/libnl3)
+include_directories(${PLATFORM_INCLUDE_DIR}/libnl-route3)
```
vitalybu: Are you sure this is required?
The include directories should be automatically added by the `target_link_libraries` function.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
 
+std::shared_ptr<std::unordered_map<std::string, std::string>>
+network_utils::get_arp_table(bool mac_as_key)
+{
+    auto arp_table_ptr = std::make_shared<std::unordered_map<std::string, std::string>>();
+    auto arp_table     = *arp_table_ptr.get();
+
+    // Allocate a new netlink socket
+    auto nl_socket = nl_socket_alloc();
+    if (nl_socket == nullptr) {
```
vitalybu:
```suggestion
    if (!nl_socket) {
```

common/beerocks/bcl/source/network/network_utils.cpp
morantr: Yes, I forgot the & sign when changed the return type.
Anyway fixed.

common/beerocks/bcl/CMakeLists.txt
RanRegev: minimum required version ensures that on these platforms the problem would be easily traced.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+        LOG(ERROR) << "Failed probing ARP table: " << nl_geterror(err);
+        nl_socket_free(nl_socket);
+        return nullptr;
+    }
+
+    nl_socket_free(nl_socket);
+
+    auto enteries_count = nl_cache_nitems(neighbor_table_cache);
+
+    constexpr const char *error_message = "neighbor_entry is null";
```
vitalybu: What? Why?

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+    auto neighbor_entry = reinterpret_cast<rtnl_neigh *>(nl_cache_get_first(neighbor_table_cache));
+    if (!neighbor_entry) {
+        LOG(ERROR) << error_message;
+        nl_cache_free(neighbor_table_cache);
+        return nullptr;
+    }
+
+    char buffer_mac[MAC_ADDR_CHAR_SIZE + 1]; // +1 for null terminator
+    char buffer_ip[IPV4_ADDR_CHAR_SIZE + 1]; // +1 for null terminator
```
vitalybu, line 411:
```suggestion
    char buffer_mac[MAC_ADDR_CHAR_SIZE + 1] = { 0 }; // +1 for null terminator
    char buffer_ip[IPV4_ADDR_CHAR_SIZE + 1] = { 0 }; // +1 for null terminator
```

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,95 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+
+    for (int i = 0; i < enteries_count; i++) {
```
vitalybu: You don't need a for-loop to iterate over the cache.
`nl_cache_get_next` returns NULL at the end of the list.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: I prefer keeping it as is. It looks better and more readable.
There are some things you can do on CPP that can give you the creeps just looking at them.
This is one of them :)

common/beerocks/bcl/source/network/network_utils.cpp
morantr: What is the benefit of that? I anyway doing construction of the string when inserting it to the map. So doing it will just cause extra string object construction which we don't need.

common/beerocks/bcl/CMakeLists.txt
vitalybu: Still, it's unnecessary. We're using netlink 3 libraries here, which is enough for our needs.
We don't require a specific version.

common/beerocks/bcl/CMakeLists.txt
morantr: I tried to remove it, and it compiles. So apparently it is not needed.
Removing it.

common/beerocks/bcl/CMakeLists.txt
morantr: So to remove it?

common/beerocks/bcl/source/network/network_utils.cpp
RanRegev: `std::remove_pointer<decltype(nl_socket_alloc())>::type` is used just because I was too lazy to look for the return type of `nl_socket_alloc()` - if you know the type, just write it.
`std::function<void(int*)>` is required as this is the interface of `nl_socket_free` and decltype of it doesn't work (I don't know why yet).
Though it looks frightening this is much safer and better code. It removes 90% of the errors happening when working with pointers.
I would really like us all to start writing safer and better code.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: We usually require it only for smart pointer because of a Klocwork error.
But ok, I'll change it.

common/beerocks/bcl/source/network/network_utils.cpp
RanRegev: > What is the benefit of that?

It is probably safer as you don't deal with limited buffer size.
It is also removes the need to declare the local buffer (mac and ip) with that +1 size added to the two `#define`s that with `std::string` are not needed at all. The +1 is a source for endless bugs that are simply omitted by using std::string instead of raw pointers.
All in all - uplifting form the bare pointers is safer, cleaner and better.

> doing construction of the string when inserting it to the map. So doing it will just cause extra string object construction which we don't need

Simply use: `arp_table[mac] = std::move(ip)` - no copy at all (and don't use `ip` afterwords).

common/beerocks/bcl/source/network/network_utils.cpp
morantr: I did it because I didn't want to create the same string in two locations.
If what you wonder about is the `constexpr const` - I wanted that the string itself be a const.
This is equivalent to `char const * const`.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: The is no benefit in doing it. The NL fills the buffer overrides what's on it.
I don't do it on every iteration, so there is no need to do it at the beginning.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: It is a good idea, that way I remove the entry check on the loop body, so I have only one place to print the error, and I could get rid of the constexpr error.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: The NL function puts a null terminator after it finishes writing the data into the buffer. It has nothing to do with the function return value.

common/beerocks/bcl/CMakeLists.txt
vitalybu: Yes

common/beerocks/bcl/source/network/network_utils.cpp
vitalybu: But does it add a NULL terminator?

morantr: Yes. Otherwise, I would have got garbage at the end of a shorter IP address.

RanRegev: I would initialize with zeros anyhow, and the reason is as always: for future changes.
Maybe today the function fills it with NULL at the end, but later we might use another function that does not.
working with initialized variables is (almost) always better.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
 }
 
+std::shared_ptr<std::unordered_map<std::string, std::string>>
+network_utils::get_arp_table(bool mac_as_key)
+{
+    auto arp_table_ptr = std::make_shared<std::unordered_map<std::string, std::string>>();
+    auto &arp_table    = *arp_table_ptr.get();
+
+    // Allocate a new netlink socket
+    std::unique_ptr<nl_sock, decltype(&nl_socket_free)> nl_socket(nl_socket_alloc(),
```
RanRegev, line 375: Excellent!

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+    }
+
+    int err = nl_connect(nl_socket.get(), NETLINK_ROUTE);
+    if (err != 0) {
+        LOG(ERROR) << "Failed connecting the netlink socket: " << nl_geterror(err);
+        return nullptr;
+    }
+
+    // ARP table cache
+    struct nl_cache *neighbor_table_cache = nullptr;
```
RanRegev, line 389: This, exactly as `nl_socket` should be a `std::unique_ptr`.
```suggestion
```

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+
+    // Allocate the cache and fill it with data
+    err = rtnl_neigh_alloc_cache(nl_socket.get(), &neighbor_table_cache);
```
RanRegev, line 392:
```suggestion
    struct nl_cache* tmp = nullptr;
    err = rtnl_neigh_alloc_cache(nl_socket.get(), &tmp);
```
[this example explains](https://stackoverflow.com/questions/12403750/initializing-a-stdunique-ptr-by-passing-the-address-of-the-pointer)

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+    if (err != 0) {
+        LOG(ERROR) << "Failed probing ARP table: " << nl_geterror(err);
+        return nullptr;
+    }
+
```
RanRegev, line 397:
```suggestion
std::unique_ptr<struct nl_cache,decltype(&nl_cache_free)> neighbor_table_cache(tmp);
```

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+    auto enteries_count = nl_cache_nitems(neighbor_table_cache);
```
RanRegev:
```suggestion
    auto enteries_count = nl_cache_nitems(neighbor_table_cache.get());
```

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+    if (enteries_count <= 0) {
+        return arp_table_ptr;
+    }
+
+    auto neighbor_entry = reinterpret_cast<rtnl_neigh *>(nl_cache_get_first(neighbor_table_cache));
```
RanRegev, line 403:
```suggestion
    auto neighbor_entry = reinterpret_cast<rtnl_neigh *>(nl_cache_get_first(neighbor_table_cache.get()));
```

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+    if (!neighbor_entry) {
+        LOG(ERROR) << "neighbor_entry is null";
+        nl_cache_free(neighbor_table_cache);
```
RanRegev, line 406:
```suggestion
```

RanRegev (approved):

common/beerocks/bcl/include/bcl/network/network_utils.h
arnout: It would actually be a lot better to have a class to represent an ARP table, with methods to look up by MAC and by IP.

However, there's a trivial resolution to Ran's comment: we don't need lookup by IP, and I don't expect we will ever need it because prplmesh doesn't work with IP addresses at all. So simply remove the boolean argument.

agent/src/beerocks/monitor/monitor_rssi.cpp
```diff
@@ -315,6 +315,21 @@ void monitor_rssi::process()
             bool is_4addr_client    = (sta_bridge_4addr_mac != network_utils::ZERO_MAC_STRING);
             std::string arp_dst_mac = is_4addr_client ? sta_bridge_4addr_mac : sta_mac;
 
+            auto sta_ipv4 = sta_node->get_ipv4();
+            if (sta_ipv4.empty() || sta_ipv4 == network_utils::ZERO_IP_STRING) {
+                LOG(DEBUG) << "Sta " << sta_mac << " IP is missing, looking at the ARP Table";
+                auto arp_table = network_utils::get_arp_table();
```
arnout, line 321: (nitpick) since this is the only use of get_arp_table, I think it would make more sense to have a function `std::string network_utils::get_ipv4_from_arp_table(const sMacAddr& mac)` that has the functionality below.

(And actually, it really should return a sIPv4Addr, but we don't have that struct (yet)).

agent/src/beerocks/monitor/monitor_rssi.cpp
```diff
@@ -315,6 +315,21 @@ void monitor_rssi::process()
...
+                    auto arp_entry_it = arp_table->find(sta_mac);
+                    if (arp_entry_it != arp_table->end()) {
+                        sta_ipv4 = arp_entry_it->second;
+                        LOG(DEBUG) << "Found IP on ARP Table, setting Sta " << sta_mac << " IP to "
+                                   << sta_ipv4;
+                        sta_node->set_ipv4(arp_entry_it->second);
+                    }
+                }
+            }
+
```
arnout, line 332: Shouldn't we bail out if IP address is still invalid?

cmake/Findnl-route-3.cmake
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_ROUTE_LIBRARY "libnl-route-3.so")
+find_path(NL3_INCLUDE_DIRS netlink/route/route.h PATH_SUFFIXES libnl3)
```
arnout: I know that this is copied from genl, but this really should be
```suggestion
find_path(NL_ROUTE_3_INCLUDE_DIRS netlink/route/route.h PATH_SUFFIXES libnl3)
```

cmake/Findnl-route-3.cmake
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+#
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+#
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+
+find_library(NL3_ROUTE_LIBRARY "libnl-route-3.so")
```
arnout: I know this is copied from genl, but this really should be
```suggestion
find_library(NL_ROUTE_3_LIBRARY "libnl-route-3.so")
```

common/beerocks/bcl/include/bcl/network/network_utils.h
```diff
@@ -94,6 +94,19 @@ class network_utils {
...
+     * The key of the map can be either a MAC address or an IP address, depends on the 'mac_as_key'
+     * argument value.
+     * If 'mac_as_key' is 'true' then the key is a MAC address, otherwise the key is an IP address.
+     * 
+     * @param[in] mac_as_key Decide whether the key of the returned unordered map object 
+     * is a MAC or an IP.
+     * @return std::shared_ptr<std::unordered_map<std::string, std::string>> 
+     */
+    static std::shared_ptr<std::unordered_map<std::string, std::string>>
+    get_arp_table(bool mac_as_key = true);
```
arnout, line 108: Please *NO* **NEVER** use std::string for a MAC address. Don't. Just don't. Stop it. Go away. This _has to stop_. I'm so tired of it.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
     return (0);
 }
 
+std::shared_ptr<std::unordered_map<std::string, std::string>>
+network_utils::get_arp_table(bool mac_as_key)
+{
+    auto arp_table_ptr = std::make_shared<std::unordered_map<std::string, std::string>>();
+    auto &arp_table    = *arp_table_ptr.get();
```
arnout, line 372: I saw other functions that simply passed the table as an output argument, to avoid all the shared pointer stuff. So that would be:
```
bool network_utils::get_arp_table(std::unordered_map<sMacAddr, std::string>& arp_table);
```

common/beerocks/bcl/source/network/network_utils.cpp
arnout: (nitpick) entries, not enteries

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+        if (rtnl_neigh_get_state(neighbor_entry) == NUD_NOARP ||
+            rtnl_neigh_get_family(neighbor_entry) != AF_INET) {
+            // Advance to the next record
+            neighbor_entry = reinterpret_cast<rtnl_neigh *>(
+                nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)));
+            continue;
+        }
+
+        auto nl_addr_mac = rtnl_neigh_get_lladdr(neighbor_entry);
+        if (nl_addr_mac == nullptr) {
```
arnout:
```suggestion
        if (!nl_addr_mac) {
```

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+        if (neighbor_entry == nullptr) {
+            break;
+        }
+
+        // Skip NOARP state and non IPv4 records
+        if (rtnl_neigh_get_state(neighbor_entry) == NUD_NOARP ||
+            rtnl_neigh_get_family(neighbor_entry) != AF_INET) {
+            // Advance to the next record
+            neighbor_entry = reinterpret_cast<rtnl_neigh *>(
+                nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)));
```
arnout, line 424: Instead of the loop on entries_count, wouldn't it be simpler and safer to iterate directly over the entries?
```
    for (auto neighbor_entry = reinterpret_cast<rtnl_neigh *>(nl_cache_get_first(neighbor_table_cache));
         neighbor_entry;
         neighbor_entry = reinterpret_cast<rtnl_neigh *>(nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)))) {
```

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -363,6 +365,102 @@ static int parseRoutes(struct nlmsghdr *nlHdr, std::shared_ptr<route_info> rtInf
...
+        auto nl_addr_mac = rtnl_neigh_get_lladdr(neighbor_entry);
+        if (nl_addr_mac == nullptr) {
+            // Advance to the next record
+            neighbor_entry = reinterpret_cast<rtnl_neigh *>(
+                nl_cache_get_next(reinterpret_cast<nl_object *>(neighbor_entry)));
+            continue;
+        }
+
+        auto nl_addr_ip = rtnl_neigh_get_dst(neighbor_entry);
+        if (nl_addr_ip == nullptr) {
```
arnout:
```suggestion
        if (!nl_addr_ip) {
```

arnout (approved):
I come a bit late to the party so feel free to ignore my feedback if you want.

agent/src/beerocks/monitor/monitor_rssi.cpp
arnout: Thinking a bit more about this... Shouldn't we also handle the case where the client changes IP address? Maybe when the ARP fails (there is no response) we should clear the IP address so it is re-retrieved.

Of course, as Alex mentioned, if the DHCP monitor is working, it shouldn't even be needed.

common/beerocks/bcl/include/bcl/network/network_utils.h
morantr: 🤣🤣🤣
Yeah yeah I know, but I did it to make a dynamic key to the map. Anyway, on the single place we use it now, we actually need it as a string.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: I thought about doing it, but @RanRegev told/convinced me not to, and I listen to him.
Democracy?

common/beerocks/bcl/source/network/network_utils.cpp
morantr: Actually I did it in the review cycle, and found out that nl_cache_get_next() never returns nullptr and I was stuck on an endless loop. Go figure the ones who wrote this library. 🤷‍♂️

cmake/Findnl-route-3.cmake
morantr: done.

cmake/Findnl-route-3.cmake
morantr: done.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: If it could be done without declaring a "tmp" pointer I would do it.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: Actually that is not a nitpick.
Fixed.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: done.

common/beerocks/bcl/source/network/network_utils.cpp
morantr: done.

common/beerocks/bcl/source/network/network_utils.cpp
RanRegev: I know that we do it _a lot_, but I prefer not to use `in-out` parameters.
I don't have strong justification for this, found one in the core guidelines -
[That can in itself be a problem and a source of errors](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-inout)

PR #1370, opened 2020-05-31
Fine tune dissociation and validity interval timers

kantera800:
Validity interval was incorrectly set in ms units instead
of beacon transmission time interval (1BTT = ~100ms)
Set non imminent validity interval correctly in BTT to reach 200ms (2 BTTs)

For imminent steering requests dissociation timer parameter indicates 
the time client has left to steer until he is disconnected by the ap.

Testing show that previous value of 200ms is not enough for
some clients to respond to BSS requests.
As a result client might not steer and AP may consider this client  as
non-responsive to 11v messages.

Increase dissociation timer to 2000ms .
Since we also want to keep the candidate list valid for that period  of time
increase the validity interval accordingly.


common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -90,9 +90,10 @@ enum eGlobals {
     PHY_RATE_100KB_MAX                      = 8666,
     PHY_RATE_100KB_MIN                      = 72,
     PHY_RATE_100KB_INVALID                  = 0,
-    BSS_STEER_DISASSOC_TIMER_MS             = 200, // ~200ms
-    BSS_STEER_IMMINENT_VALID_INT            = 100,
-    BSS_STEER_VALID_INT                     = 50,
+    BEACON_TX_TIME_MS                       = 100,
+    BSS_STEER_DISASSOC_TIMER_MS             = 2000,
+    BSS_STEER_IMMINENT_VALID_INT            = (BSS_STEER_DISASSOC_TIMER_MS / BEACON_TX_TIME_MS),
```
arnout: This makes the validity time equal to the disassociation timer, right? However, I'd expect the validity time to be a bit higher, no?

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -90,9 +90,10 @@ enum eGlobals {
...
+    BSS_STEER_VALID_INT                     = 2, // 200ms
```
arnout: That sounds *awfully* short as a validity time... I would actually expect it to be higher in case we don't disassociate - although really, why would the validity time be different in either case?

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -919,21 +919,23 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
-        " neighbor=" + bssid + ",0," + std::to_string(oper_class) + "," + std::to_string(chan) +
-        ",0,255";
+        + " dialog_token=" + "0" + " pref=" + "1" + " abridged=" + "1";
 
     // Divide disassoc_timer by 100, because the hostapd expects it to be in beacon interval
     // which is 100ms.
     if (disassoc_timer) {
         cmd += std::string() + " disassoc_imminent=" + "1" +
-               " disassoc_timer=" + std::to_string(disassoc_timer / 100);
+               " disassoc_timer=" + std::to_string(disassoc_timer / beerocks::BEACON_TX_TIME_MS);
```
arnout: It would actually be better to change the bwl interface to express this in BTT instead of ms...

arnout (approved):
Excellent commit messages. Just maybe in the last commit, the comment "This is not related to this PR" is not appropriate for a commit message - once the PR is merged, the concept of "this PR" is gone, so it is not relevant to be kept in the history. You should put that kind of comment in the PR description instead.

common/beerocks/bcl/include/bcl/beerocks_defines.h
kantera800: IMO it should be even a bit less in order to give the client a chance to disconnect on its own (if  some reason the client decides to wait all the validity period before he responds).
personally i never seen a client reply to BSS after more than 1 sec.

common/beerocks/bcl/include/bcl/beerocks_defines.h
kantera800: validity interval of 200ms was not enough for imminent steering, most of the times we would just disconnect the client before it even answers.
but , for non imminent steering requests client usually responds right away (maybe he doesn't have to perform a neighbor scan ) .
Also , for imminent bss requests cases i wanted it to be depended on the dissociation time in case someone changes it in the future and forget to update the validity interval as well.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
kantera800: sure no problem , that is a lot of files to change though :)

common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -90,9 +90,10 @@ enum eGlobals {
     PHY_RATE_100KB_MAX                      = 8666,
     PHY_RATE_100KB_MIN                      = 72,
     PHY_RATE_100KB_INVALID                  = 0,
-    BSS_STEER_DISASSOC_TIMER_MS             = 200, // ~200ms
-    BSS_STEER_IMMINENT_VALID_INT            = 100,
-    BSS_STEER_VALID_INT                     = 50,
+    BEACON_TX_TIME_MS                       = 100,
```
morantr: I think the correct term is "beacon interval".

common/beerocks/bcl/include/bcl/beerocks_defines.h
morantr: I think it will be a good idea to rename it and add the unit to the name:
BSS_STEER_VALID_BEACON_INTERVAL

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -919,21 +919,23 @@ bool ap_wlan_hal_dwpal::sta_bss_steer(const std::string &mac, const std::string
...
     }
 
     if (valid_int) {
         cmd += " valid_int=" + std::to_string(valid_int);
     }
 
+    // Target BSSID
+    cmd += std::string() + " neighbor=" + bssid + ",0," + std::to_string(oper_class) + "," +
+           std::to_string(chan) + ",0,255";
+
```
morantr, line 939: What is the reason for moving it?
Apply also to bwl_nl.

agent/src/beerocks/monitor/monitor_stats.cpp
```diff
@@ -280,7 +280,7 @@ void monitor_stats::process()
     //send response
     if (!requests_list.empty()) {
         auto request = requests_list.front();
-        LOG(TRACE) << "Serving reqeust: mid = " << request.message_id << " MAC = " << request.mac;
+        // LOG(TRACE) << "Serving reqeust: mid = " << request.message_id << " MAC = " << request.mac;
```
morantr, line 283: You are my hero ❤️ 
BTW, do you think we should add a reference to the issue number?

morantr (approved):

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
kantera800: I explained in commit message, hostapd expects it as last parameter

agent/src/beerocks/monitor/monitor_stats.cpp
kantera800: There are lots of bugs in the code with open issues on then. I think that adding issue number is more appropriate for a w/a

common/beerocks/bcl/include/bcl/beerocks_defines.h
kantera800: actually its beacon transmit time  (BTT)
so if TX is not clear i'll change it to BEACON_TRANSMIT_TIME_MS

common/beerocks/bcl/include/bcl/beerocks_defines.h
kantera800: changed to BSS_STEER_VALID_INT_BTT

PR #1373, opened 2020-06-01
[RFC] Move stuff from framework to common

arnout:
This PR is an RFC about moving all the code from the "framework" directory into the "common" directory.
The distinction between framework and common is pretty artificial, and probably a result of historical accident.
In particular, we'd eventually want to merge bcl with mapfcommon because there is some annoying overlap between the two.

This PR is incomplete: easylogging++ and mapfcommon itself are not moved.
That is because there are currently open PRs that affect these directories, and I didn't want to create unnecessary conflict.

Quite a few additional changes sneak in into this PR.

- For transport and bpl, there was an additional intermediate directory that I deemed redundant.
  I removed it.
- For transport, there is a case to be made to put in at top level instead of in common.
  Indeed, it's a separate executable that stands at a level similar to agent and controller.
  However, I didn't do that because I feel the top-level is already pretty crowded.
- For bpl, I didn't put it in the beerocks directory.
  The grand plan is to move everything out of the beerocks directory.
  I again consider that additional directory level to be pointless.
- ieee1905 moved from mapf to beerocks namespace.
  The grand plan is to move everything to beerocks namespace, since the distinction between beerocks and mapf is IMO useless.
  bpl was already in beerocks namespace.
  tlvf has its own namespace.
- There is now common/tlvf and common/beerocks/tlvf.
  The plan is to merge the two in a single tlvf that has both the EasyMesh and the Vendor Specific TLVs.


rmelotte (approved):
There is only some more edits to be done to `cppcheck_existing_issues.txt`.
Other than that, LGTM

mariomaz:
Top-level directory is too crowded but the problem is its contents are quite heterogeneous. I'd create a `source` directory at the top-level and move all source code inside. Also, having a different subdirectory inside `source` for each artifact looks like a good idea.

mariomaz (approved):
Top-level directory is too crowded but for me, the problem is its contents are quite heterogeneous. I'd create a source directory at the top-level and move all source code inside. Also, having a different sub-directory inside source for each artifact looks like a good idea.

RanRegev:
Maybe we should do some planning based on the lessons learnt, including documentation, for the desired goal.
Something in the form of:
## prplMesh Library Structure 
### common
* details
### bpl
* details
### etc

## prplMesh Source Structure 
### src/
* library and applications code only.
* does not contain tests and build instructions.
* does not contain unit test code
### tests/
#### unit tests/
#### flow tests/
#### etc/

I searched the documentation and didn't find anything that is focused on the structure.

arnout:
As requested by @RanRegev, and taking the input from @mariomaz in account, here's the proposed layout after moving everything:

- LICENSES/
- ci/
  - ...
- cmake/
- documentation/
  - ...
- source/
  - agent/
    - agent/
    - backhaul_manager/
    - fronthaul_manager/
    - platform_manager/
  - bcl/
  - bml/
  - bpl/
  - broker/
  - btl/
  - bwl/
  - cli/
  - common/   (this is mapf_common, not bcl)
  - external/
  - ieee1905_transport/
  - scripts/
  - tlvf/    (both the beerocks and the standard messages are in here)
- tests/
  - ...
- tools/
  - ...

And e.g. bwl would be structured like this:

- dummy/
- dwpal/
- include/bwl/
- nl80211/
- src/  (this is the common part, so what is currently in common and shared)
- unit_tests/

arnout:
Also, regarding the additional changes sneaked in here. Based on discussion in prplmesh-democracy, I came to the conclusion:
- additional directory for transport and bpl will be removed;
- if any namespace changes are made, it will not be to beerocks namespace, but to "map" (or maybe "prplmesh"? I don't like "map" very much because a map has a meaning already so it's confusing.)

RanRegev:
@arnout suggestion looks good with the following comments:
* Today, `./build/install/include` is where we place all header files to be included by the various applications internally but also if someone wants to develop something based on them (e.g. tests). These include files suppose to be the complete interface for using prplMesh API. Although prplMesh is not designed to be a library, still there _is_ an include part already in it. The concept exists.
My suggestion is to have everything that goes today into ./build/install/include to be under `include` that sits parallel to `source`. The pros are that it helps keep dependencies clear, it makes life easier for developers to know what is allowed to be included and it removes the need to "collect" header files when installing (this, and only this, `include` directory is installed).
* All libraries that share common parts should have libraries that share a common prefix (or suffix). e.g: `bwl_common`, `agent_common` (does not exist yet). And therefore share a common source directory: e.g `source/bwl/common/`, `agent/common`.
* `source/agent/agent/` seems redundant. I prefer: `/source/agent/{backhaul,fronthaul,platform}`
* I think `unit_test` directory should be under `tests/unit_test`. It keeps the source code directory clean and it also reveals dependencies: unit-tests should work with the public interface of prplMesh (the `include` directory that represents the DUT) and not with internal data. There might be exceptions for this but normally it is the case.
e.g.: [nl80211_client_test.cpp](https://github.com/prplfoundation/prplMesh/blob/master/common/beerocks/bwl/unit_tests/nl80211_client_test.cpp) includes [nl80211_client.h](https://github.com/prplfoundation/prplMesh/blob/master/common/beerocks/bwl/include/bwl/nl80211_client.h) which is NOT part of the public interface of prplMesh. However, `nl80211_client.h` itself depends _only_ on the public interface. It means that it should probably be moved to `include`.
All in all, I think that placing `unit_test` _outside_ the source code directory is better.
* I think `AutoGenerated` should be generated to the same place it would be stored if hand crafted. It keeps the logical structure. The fact that this file was generated by automatic tool and not human is not important IMO. Naturally, however, auto generated code does not tend to mix with human written code. So the suggestion is to remove `AutoGenerated` directory completely.
* I think that `external`, sometimes called `3rd-party`, should be placed externally to our sources, in parallel to `sources`.

Based on the comments above and on @arnout work, this is my suggestion for the directory tree:
- LICENSES/
- ci/
  - ...
- cmake/
- documentation/
  - ...
- **include/**
  - bwl/ (looking at the current tree I am puzzled: there is `./build/install/lib/libbwl.so` but there is no `bwl` include)
  - bcl/
  - **tlvf/ (the generated header files)**
  - ... 
- source/
  - agent/
    - **backhaul**/
    - **fronthaul**/
    - **platform**/
    - **common/ (common to all agents applications)**
  - bcl/
  - bml/
     - **common (currently there is no such, but if needed for rdkb and bml, this would be the place)**
  - bpl/
  - broker/
  - btl/
  - bwl/
  - cli/
  - **common/   (prplMesh common)**
  - ieee1905_transport/
  - scripts/
  - tlvf/    **(no AutoGenerated directory, just the sources, as if hand-written)**
   - ieee_1905_1/
   - ...
- **external/**
- tests/
   - **unit_tests/ (moved here, separate from the sources)**
  - ...
- tools/
  - ...

And e.g. bwl would be structured like this:
- dummy/
- dwpal/
- nl80211/
- **common/  (common name for all directories that have common code. this one produces `libbwl_common.so`)**

morantr:
> As requested by @RanRegev, and taking the input from @mariomaz in account, here's the proposed layout after moving everything:
> 
> * LICENSES/
> * ci/
>   
>   * ...
> * cmake/
> * documentation/
>   
>   * ...
> * source/
>   
>   * agent/
>     
>     * agent/
>     * backhaul_manager/
>     * fronthaul_manager/
>     * platform_manager/
>   * bcl/
>   * bml/
>   * bpl/
>   * broker/
>   * btl/
>   * bwl/
>   * cli/
>   * common/   (this is mapf_common, not bcl)
>   * external/
>   * ieee1905_transport/
>   * scripts/
>   * tlvf/    (both the beerocks and the standard messages are in here)
> * tests/
>   
>   * ...
> * tools/
>   
>   * ...
> 
> And e.g. bwl would be structured like this:
> 
> * dummy/
> * dwpal/
> * include/bwl/
> * nl80211/
> * src/  (this is the common part, so what is currently in common and shared)
> * unit_tests/

few comments:
1. The platform manager and backhaul manager shall be on the sub-agent directory because they are running on its context. The fronthaul will remain on a separate folder.
2. the BWL should be on the Top Agent folder since only the Agent uses it.
3. I don't see any controller folder, so it should be, and it shall include bml/cli, because they are controlling the controller.
4. I think that it will be a good idea to unify all code related to the transport on the framework (transport, mapf common/broker) under the btl.
5. All other things can be on the Common folder.
6. I don't know why it is like this, but it is really annoying that the header file is not on the same place where the cpp file is.
So I would change it as well so they will always be on the same place.

mariomaz:
> I think unit_test directory should be under tests/unit_test. It keeps the source code directory clean and it also reveals dependencies: unit-tests should work with the public interface of prplMesh (the include directory that represents the DUT) and not with internal data. There might be exceptions for this but normally it is the case.

Unit tests must often include private headers of a component. 

In some other projects I've seen auto-generated code placed in a folder `src_gen`, at the same level as `src`.

> I think that external, sometimes called 3rd-party, should be placed externally to our sources, in parallel to sources.

I agree.

And e.g. bwl would be structured like this:

src/
- dummy/
- dwpal/
- nl80211/
- common/ (code here might be placed directly one level above, so remove common)
- shared/ (I'm not particularly proud of this name, code in here is shared between the dwpal and nl80211 flavors of the BWL, and maybe it deserves a library in its own, out of bwl)

include/bwl/
unit_tests/

arnout:
I disagree with most of the remaining suggestions, which indicates that we've converged into bikeshedding territory :-)

I created a JIRA task [PPM-219](https://jira.prplfoundation.org/projects/PPM/issues/PPM-219) to cover it - we'll move future discussion, if any, there.

Here are my replies to some of the suggestions:

@RanRegev wrote:

>     * Today, `./build/install/include` is where we place all header files to be included by the various applications internally but also if someone wants to develop something based on them (e.g. tests). These include files suppose to be the complete interface for using prplMesh API. Although prplMesh is not designed to be a library, still there _is_ an include part already in it. The concept exists.

Actually, build/install/include is not used while building any part of prplMesh itself.

>       My suggestion is to have everything that goes today into ./build/install/include to be under `include` that sits parallel to `source`. The pros are that it helps keep dependencies clear, it makes life easier for developers to know what is allowed to be included and it removes the need to "collect" header files when installing (this, and only this, `include` directory is installed).

I don't really like that, because it moves the headers too far apart from the sources IMHO. I dont agree that it helps to keep dependencies clear - IMO it actually doesn't! With this approach, there will be a global `-I${ROOTDIR}/include` so you don't need any explicit dependency at compile time. You only need the dependency at link time.

Anyway - this proposal is somewhat independent of my suggested reorganisation. I only move the libraries around, without changing much inside each library. So I've removed the part about internal structure of each library from [PPM-219](https://jira.prplfoundation.org/projects/PPM/issues/PPM-219).

>     * All libraries that share common parts should have libraries that share a common prefix (or suffix). e.g: `bwl_common`, `agent_common` (does not exist yet). And therefore share a common source directory: e.g `source/bwl/common/`, `agent/common`.

Again, I think this goes beyond the fairly simple reorganisation that I'm suggesting, and I don't see much advantage.

>     * `source/agent/agent/` seems redundant. I prefer: `/source/agent/{backhaul,fronthaul,platform}`

You still need to distinguish the "main" part from the rest. So I called it "main" instead of agent.

>     * I think `unit_test` directory should be under `tests/unit_test`. It keeps the source code directory clean and it also reveals dependencies: unit-tests should work with the public interface of prplMesh (the `include` directory that represents the DUT) and not with internal data. There might be exceptions for this but normally it is the case.

Unit tests *will* often use internal headers, and that's fine (that's what makes them unit tests rather than module tests).

But again, reorganisation of the contents of each module is orthogonal to this proposal.

>     * I think `AutoGenerated` should be generated to the same place it would be stored if hand crafted. It keeps the logical structure. The fact that this file was generated by automatic tool and not human is not important IMO. Naturally, however, auto generated code does not tend to mix with human written code. So the suggestion is to remove `AutoGenerated` directory completely.

Well, I think the autogenerated shouldn't even be in git, so should be in the build directory instead of the source directory :-)

>     * I think that `external`, sometimes called `3rd-party`, should be placed externally to our sources, in parallel to `sources`.

That's a good proposal, I'm taking it.

@morantr wrote:
>     1. The platform manager and backhaul manager shall be on the sub-agent directory because they are running on its context. The fronthaul will remain on a separate folder.

So I'm leaving this for further discussion - nothing changes about the internal structure of agent in [PPM-219](https://jira.prplfoundation.org/projects/PPM/issues/PPM-219) other than removing the src/beerocks part.

>     2. the BWL should be on the Top Agent folder since only the Agent uses it.

I disagree with this. I prefer to have the different pieces, insofar as they are really independent pieces, as directories directly under source/. So I would tend to move e.g. fronthaul directly under source/ as well - although there, it really is a lot more tightly integrated with the other parts of the agent.


>     3. I don't see any controller folder, so it should be, and it shall include bml/cli, because they are controlling the controller.

 Yes forgot that one, I added it now.

 For bml, same reasoning as for bwl: it's a well-defined component so better put it at the top level.


>     4. I think that it will be a good idea to unify all code related to the transport on the framework (transport, mapf common/broker) under the btl.

"transport" is a bit overloaded here... There is ieee1905_transport which is clearly a separate component (it's the thing that listens on interfaces). The btl and broker indeed do go hand in hand, but I think it should still be two components: the btl part that is also used by all the other component, and the broker itself that is a stand-alone executable. Except that AFAIU @vitalybu intends to integrate the broker with the ieee1905_transport, which muddies the picture a bit...

>     6. I don't know why it is like this, but it is really annoying that the header file is not on the same place where the cpp file is.
>        So I would change it as well so they will always be on the same place.

The idea, I guess, is to separate the public interface from the internal interfaces. What is in include/ can be used by other components, while what is in src/ can't be used by other components. I do think it's good that the two are separated at the directory level.

@mariomaz wrote:

> In some other projects I've seen auto-generated code placed in a folder `src_gen`, at the same level as `src`.

I like that! But not for this PR/task :-)


>     * shared/ (I'm not particularly proud of this name, code in here is shared between the dwpal and nl80211 flavors of the BWL, and maybe it deserves a library in its own, out of bwl)

:+1:  to that! Although, it will only ever be used by bwl, so not really worth it maybe... So maybe just give it a more descriptive name than "shared". Actually, "nl80211" would be a good name for it, and the "nl80211" flavor of BWL is misnamed, it should have been "hostap".

arnout:
Discussion moved to  PPM-219 and the actual code changes are going to be completely different, so closing this PR.

PR #1380, opened 2020-06-02
remove steer on vaps list dependency

kantera800:
Steer on vaps list is a feature that allows user to disable/enable steering on specific VAPs.
prplmesh UCI database parameter get_hostap_vap_with_ssid holds vap names that are allowed to steer.
The problem is that each platform has different vap names but some use the same UCI database. 
for example
RAX40 has 2 vaps WLAN0.1 and WLAN2.0  (UCI default)
Axe6000 has 2 vaps WLAN0.1 and **WLAN2.2**

Same UCI  configuration applies to both since they are openWRT that use the same device name ( /tmp/sysinfo/board_name).

As a result on AXE6000 , optimal path aborted every time we are trying to steer to/from WLAN2.2

Since this feature is needed only for RDKB platform the solution is to remove it from all platforms except RDKB and modify the steering/optimal path task code  to allow all vaps to steer in case the list is missing or empty.

### DIFF
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -1558,6 +1558,32 @@ std::string db::get_hostap_ssid(const std::string &mac)
     return std::string();
 }
 
+bool db::is_vap_on_steer_list(const std::string &bssid)
+{
+    if (config.load_steer_on_vaps.empty()) {
+        return true;
+    }
+
+    auto vap_name = get_hostap_iface_name(bssid);
+    if (vap_name == "INVALID") {
+        LOG(ERROR) << "vap name is invalid for bssid " << bssid;
+        return false;
+    }
+
+    auto vap_id = get_hostap_vap_id(bssid);
+    if (vap_id == IFACE_ID_INVALID) {
+        LOG(ERROR) << "vap id is invalid for bssid " << bssid;
+        return false;
+    }
+
+    vap_name         = utils::get_iface_string_from_iface_vap_ids(vap_name, vap_id);
+    auto &steer_vaps = config.load_steer_on_vaps;
+    if (steer_vaps.find(vap_name) == std::string::npos) {
+        return false;
+    }
+    return true;
+}
+
 std::string db::get_hostap_vap_with_ssid(const std::string &mac, const std::string &ssid)
 {
     auto n = get_node(mac);
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -1595,18 +1595,15 @@ std::string db::get_hostap_vap_with_ssid(const std::string &mac, const std::stri
         return std::string();
     }
 
-    auto vap_name = get_hostap_iface_name(mac);
-    vap_name.append(".");
+    auto it = std::find_if(
+        n->hostap->vaps_info.begin(), n->hostap->vaps_info.end(),
+        [&](const std::pair<int8_t, sVapElement> &vap) { return vap.second.ssid == ssid; });
 
-    for (auto const &it : n->hostap->vaps_info) {
-        auto tmp_vap_name = vap_name;
-        tmp_vap_name.append(std::to_string(get_hostap_vap_id(it.second.mac)));
-        if ((it.second.ssid == ssid) &&
-            (config.load_steer_on_vaps.find(tmp_vap_name) != std::string::npos)) {
-            return it.second.mac;
-        }
+    if (it == n->hostap->vaps_info.end()) {
+        // no vap with same ssid is found
+        return std::string();
     }
-    return std::string();
+    return it->second.mac;
 }
 
 std::string db::get_hostap_vap_mac(const std::string &mac, const int vap_id)
--- a/controller/src/beerocks/master/db/db.h
+++ b/controller/src/beerocks/master/db/db.h
@@ -346,6 +346,13 @@ public:
                  bool backhual);
 
     std::string get_hostap_ssid(const std::string &mac);
+    /**
+     * @brief checks if vap name is on the steer list.
+     * 
+     * @param[in] bssid vap mac address.
+     * @return true if vap name is on the steer list.
+     */
+    bool is_vap_on_steer_list(const std::string &bssid);
     std::string get_hostap_vap_with_ssid(const std::string &mac, const std::string &ssid);
     std::string get_hostap_vap_mac(const std::string &mac, const int vap_id);
     std::string get_node_parent_radio(const std::string &mac);
--- a/controller/src/beerocks/master/tasks/optimal_path_task.cpp
+++ b/controller/src/beerocks/master/tasks/optimal_path_task.cpp
@@ -627,20 +627,20 @@ void optimal_path_task::work()
                                << " could not find a better path for sta " << sta_mac << std::endl);
             finish();
         } else {
-            chosen_hostap = database.get_hostap_vap_with_ssid(chosen_hostap, current_hostap_ssid);
+            chosen_bssid = database.get_hostap_vap_with_ssid(chosen_hostap, current_hostap_ssid);
             if (!database.settings_client_optimal_path_roaming_prefer_signal_strength()) {
                 LOG_CLI(DEBUG, "optimal_path_task:" << std::endl
                                                     << "    best hostap for " << sta_mac << " is "
-                                                    << chosen_hostap << " with weighted_phy_rate="
+                                                    << chosen_bssid << " with weighted_phy_rate="
                                                     << (best_weighted_phy_rate / (1024.0 * 1024.0))
                                                     << " [Mbps]" << std::endl
                                                     << "    --> steering " << sta_mac << " to "
-                                                    << chosen_hostap << std::endl);
+                                                    << chosen_bssid << std::endl);
             } else {
                 LOG_CLI(DEBUG,
                         "optimal_path_task:" << std::endl
                                              << "    best hostap (signal strength metric) for "
-                                             << sta_mac << " is " << chosen_hostap
+                                             << sta_mac << " is " << chosen_bssid
                                              << " with dl_rssi=" << (best_dl_rssi) << " [dBm]");
             }
 
@@ -1143,7 +1143,7 @@ void optimal_path_task::work()
             }
         }
 
-        TASK_LOG(DEBUG) << "end of for loop";
+        TASK_LOG(DEBUG) << "end of hostap candidate list";
 
         if (all_hostaps_below_cutoff && current_hostap_is_5ghz) {
             best_weighted_phy_rate = best_weighted_phy_rate_below_cutoff;
@@ -1167,32 +1167,29 @@ void optimal_path_task::work()
             }
         }
 
-        auto chosen_hostap_vap =
-            database.get_hostap_vap_with_ssid(chosen_hostap, current_hostap_ssid);
+        chosen_bssid = database.get_hostap_vap_with_ssid(chosen_hostap, current_hostap_ssid);
 
-        if (chosen_hostap.empty() || (chosen_hostap == current_hostap) ||
-            chosen_hostap_vap.empty()) {
+        if (chosen_hostap.empty() || (chosen_hostap == current_hostap) || chosen_bssid.empty()) {
             LOG_CLI(DEBUG, "optimal_path_task:" << std::endl
                                                 << "   could not find a better path for sta "
                                                 << sta_mac << std::endl);
             finish();
         } else {
-            chosen_hostap = chosen_hostap_vap;
             if (!database.settings_client_optimal_path_roaming_prefer_signal_strength()) {
                 LOG_CLI(DEBUG, "optimal_path_task:" << std::endl
                                                     << "    best hostap for " << sta_mac << " is "
-                                                    << chosen_hostap << " with weighted_phy_rate="
+                                                    << chosen_bssid << " with weighted_phy_rate="
                                                     << (best_weighted_phy_rate / (1024.0 * 1024.0))
                                                     << " [Mbps]" << std::endl
                                                     << "    --> steering " << sta_mac << " to "
-                                                    << chosen_hostap << std::endl);
+                                                    << chosen_bssid << std::endl);
             } else {
                 LOG_CLI(DEBUG, "optimal_path_task:"
                                    << std::endl
                                    << "    best hostap (signal strength metric) for " << sta_mac
-                                   << " is " << chosen_hostap << " with ul_rssi=" << (best_ul_rssi)
+                                   << " is " << chosen_bssid << " with ul_rssi=" << (best_ul_rssi)
                                    << " [dBm]" << std::endl
-                                   << "    --> steering " << sta_mac << " to " << chosen_hostap
+                                   << "    --> steering " << sta_mac << " to " << chosen_bssid
                                    << std::endl);
             }
             state = SEND_STEER_ACTION;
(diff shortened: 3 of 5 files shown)


controller/src/beerocks/master/db/db.cpp
```diff
@@ -1558,6 +1558,20 @@ std::string db::get_hostap_ssid(const std::string &mac)
     return std::string();
 }
 
+bool db::is_vap_on_steer_list(const std::string &bssid)
+{
+    auto current_vap_name = get_hostap_iface_name(bssid);
+    current_vap_name.append(".");
+    current_vap_name.append(std::to_string(get_hostap_vap_id(bssid)));
```
morantr: You have for that `utils::get_iface_string_from_iface_vap_ids(const std::string &iface, int8_t vap_id)`
So it could be replaced with:
```cpp
auto current_vap_name = utils::get_iface_string_from_iface_vap_ids(get_hostap_iface_name(bssid), 
                                                                   get_hostap_vap_id(bssid));
```

(nitpick)

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1558,6 +1558,20 @@ std::string db::get_hostap_ssid(const std::string &mac)
     return std::string();
 }
 
+bool db::is_vap_on_steer_list(const std::string &bssid)
```
morantr, line 1561: I don't like naming things with "is_" prefix. Since the function returns a `bool` the "is_" prefix is self implied and therefore doesn't need to be added. But that is just my opinion. 
(nitpick)

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1558,6 +1558,20 @@ std::string db::get_hostap_ssid(const std::string &mac)
...
+    auto current_vap_name = get_hostap_iface_name(bssid);
+    current_vap_name.append(".");
+    current_vap_name.append(std::to_string(get_hostap_vap_id(bssid)));
+    if (!config.load_steer_on_vaps.empty()) {
+        auto steer_vaps = config.load_steer_on_vaps;
+        if (steer_vaps.find(current_vap_name) == std::string::npos) {
+            return false;
+        }
+    }
+    return true;
```
morantr: copy `config.load_steer_on_vaps` doesn't not give any benefit, a reference is better here.
second, is it possible to write it with less indentation so it will be more readable:
```cpp
if (config.load_steer_on_vaps.empty()) {
   return true;
}
auto &steer_vaps = config.load_steer_on_vaps;
if (steer_vaps.find(current_vap_name) == std::string::npos) {
   return false;
}
return true;

```

controller/src/beerocks/master/db/db.h
```diff
@@ -346,6 +346,13 @@ class db {
                  bool backhual);
 
     std::string get_hostap_ssid(const std::string &mac);
+    /**
+     * @brief checks if vap name is on the steer list
+     * 
+     * @param bssid vap mac address
```
morantr: We usually add the "@param" direction
"@param[in]" in this case.
Also, dots at the end of the lines are always a good idea in Doxygen documentation. 😏

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -84,15 +84,14 @@ void optimal_path_task::work()
...
-        current_vap_name.append(std::to_string(database.get_node_vap_id(sta_mac)));
-        auto steer_vaps = database.config.load_steer_on_vaps;
-        if (steer_vaps.find(current_vap_name) == std::string::npos) {
-            TASK_LOG(INFO) << "current hostap vap:" << current_vap_name << " mac:" << sta_mac
-                           << " is not in steer list:" << steer_vaps << " aborting optimal task"
-                           << std::endl;
+        // Steering allowed on all vaps unless load_steer_on_vaps list is defined
+        // on the platform , in that case verify that vap is on that list
+        if (!database.config.load_steer_on_vaps.empty() &&
+            !database.is_vap_on_steer_list(current_hostap_vap)) {
```
morantr: Since `load_steer_on_vaps.empty()` is already being checked on `is_vap_on_steer_list()` it is redundant to add it to the condition.
This is enough:
```cpp
if (!database.is_vap_on_steer_list(current_hostap_vap))
{
   // code
}
```

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -84,15 +84,14 @@ void optimal_path_task::work()
...
+            TASK_LOG(WARNING) << "client " << sta_mac << " is connected to vap "
+                              << current_hostap_vap << " that is currently not in steer list: "
+                              << database.config.load_steer_on_vaps << " aborting optimal task"
+                              << std::endl;
```
morantr: Why did you add an `endl`? Just curious.

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -198,16 +197,29 @@ void optimal_path_task::work()
             }
         }
 
-        // Check if hostapd has suitable ssid
+        // Check if hostap has suitable ssid
         auto it = potential_11k_aps.begin();
         while (it != potential_11k_aps.end()) {
-            if (database.get_hostap_vap_with_ssid(it->first, current_hostap_ssid).empty()) {
+
+            const std::string candidate_bssid =
```
morantr: `const` here is redundant.

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -198,16 +197,29 @@ void optimal_path_task::work()
...
                 it = potential_11k_aps.erase(it);
-            } else {
-                ++it;
+                continue;
+            }
+
+            // Steering allowed on all vaps unless load_steer_on_vaps list is defined
+            // on the platform , in that case verify that vap is on that list
+            if (!database.config.load_steer_on_vaps.empty() &&
+                !database.is_vap_on_steer_list(candidate_bssid)) {
```
morantr: Here too,  load_steer_on_vaps.empty() is redundant.

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -198,16 +197,29 @@ void optimal_path_task::work()
...
+                TASK_LOG(INFO) << "Remove candidate " << it->first << " , vap " << candidate_bssid
+                               << " is not in steer list: " << database.config.load_steer_on_vaps;
+                it = potential_11k_aps.erase(it);
```
morantr, line 217: Suggestion:
Maybe instead of removing it from the list, we will not add them in the first place?
This comment apply also to the cross candidates.

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -732,16 +744,29 @@ void optimal_path_task::work()
...
                           << ". Hostap doesn't have current_hostap_ssid " << current_hostap_ssid;
                 it = hostaps.erase(it);
-            } else {
-                ++it;
+                continue;
+            }
+
+            // Steering allowed on all vaps unless load_steer_on_vaps list is defined
+            // on the platform , in that case verify that vap is on that list
+            if (!database.config.load_steer_on_vaps.empty() &&
```
morantr: here too, `!database.config.load_steer_on_vaps.empty()` is redundant.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -144,20 +144,20 @@ int cfg_get_load_steer_on_vaps(int num_of_interfaces,
...
                 load_steer_on_vaps_str.append(std::string(hostap_iface_steer_vaps));
+                MAPF_DBG("adding interface " << hostap_iface_steer_vaps
+                                             << " to the steer on vaps list");
             }
         }
     }
+
     if (load_steer_on_vaps_str.empty()) {
-        MAPF_DBG("steer vaps list is empty");
+        MAPF_DBG("steer on vaps list not configured on this platform");
```
morantr: The "on this platform" is not valuable information, so I would remove it.

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -144,20 +144,20 @@ int cfg_get_load_steer_on_vaps(int num_of_interfaces,
...
-            MAPF_ERR("failed to get wifi interface steer vaps for agent" << index);
-        } else {
+        if (cfg_get_hostap_iface_steer_vaps(index, hostap_iface_steer_vaps) == RETURN_OK) {
             if (std::string(hostap_iface_steer_vaps).length() > 0) {
                 if (!load_steer_on_vaps_str.empty()) {
                     load_steer_on_vaps_str.append(",");
                 }
-                // for linux implementation the wlan?.0 vaps are used for band steering
                 load_steer_on_vaps_str.append(std::string(hostap_iface_steer_vaps));
+                MAPF_DBG("adding interface " << hostap_iface_steer_vaps
```
morantr, line 153: interface**s**

morantr (approved):

controller/src/beerocks/master/tasks/optimal_path_task.cpp
kantera800: Do you how many places I wanted to refractor the optimal task and held myself not to. You are probably right though... Next time 😎

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
kantera800: Its a loop adding interface at a time

morantr: The 's' at the end of 'hostap_iface_steer_vaps' has confused me...

controller/src/beerocks/master/tasks/optimal_path_task.cpp
kantera800: good point. done

controller/src/beerocks/master/tasks/optimal_path_task.cpp
kantera800: copy paste. removed

controller/src/beerocks/master/db/db.cpp
adam1985d: I actually prefer the "is_" prefix
this helps to differ between APIs that return success/fail boolean to APIs that return actual value of a configuration etc.
and it makes the code much more readable in the calling location

there are many APIs that it is hard to understand if the API performs something or only checks if something is configured and this is why the "is_" is very useful

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1558,6 +1558,21 @@ std::string db::get_hostap_ssid(const std::string &mac)
     return std::string();
 }
 
+bool db::is_vap_on_steer_list(const std::string &bssid)
+{
+    if (config.load_steer_on_vaps.empty()) {
+        return true;
+    }
+    auto current_vap_name = get_hostap_iface_name(bssid);
```
adam1985d: please add check if the returned string is "INVALID" (see API possible failures and return values)

controller/src/beerocks/master/db/db.cpp
adam1985d: I agree with @morantr 
better use an existing API for single-point-f-change later on if required
I don't think it is a nitpick

I do think that the `get_hostap_iface_name()` and `get_hostap_vap_id()` should be validated before calling the `get_iface_string_from_iface_vap_ids` though.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1584,18 +1584,15 @@ std::string db::get_hostap_vap_with_ssid(const std::string &mac, const std::stri
...
 
-    for (auto const &it : n->hostap->vaps_info) {
-        auto tmp_vap_name = vap_name;
-        tmp_vap_name.append(std::to_string(get_hostap_vap_id(it.second.mac)));
-        if ((it.second.ssid == ssid) &&
-            (config.load_steer_on_vaps.find(tmp_vap_name) != std::string::npos)) {
-            return it.second.mac;
-        }
+    if (it == n->hostap->vaps_info.end()) {
+        // no vap with same ssid is found
```
adam1985d, line 1603: how often is this operation performed?
wouldn't it be cleared to add a debug log-print the "no vap with same ssid is found" instead of the comment? (unless this is called often and can flood the logs)

controller/src/beerocks/master/tasks/optimal_path_task.cpp
```diff
@@ -732,16 +741,28 @@ void optimal_path_task::work()
...
+
+            // Steering allowed on all vaps unless load_steer_on_vaps list is defined
+            // on the platform , in that case verify that vap is on that list
+            if (!database.is_vap_on_steer_list(candidate_bssid)) {
+                TASK_LOG(INFO) << "Remove candidate " << *it << " , vap " << candidate_bssid
+                               << " is not in steer list: " << database.config.load_steer_on_vaps;
+                it = hostaps.erase(it);
             }
+            ++it;
         }
```
adam1985d, line 766: probably need to check here if `hostaps` set is empty and `finish()`

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
```diff
@@ -144,20 +144,20 @@ int cfg_get_load_steer_on_vaps(int num_of_interfaces,
...
+                                             << " to the steer on vaps list");
             }
         }
     }
+
     if (load_steer_on_vaps_str.empty()) {
-        MAPF_DBG("steer vaps list is empty");
+        MAPF_DBG("steer on vaps list is not configured");
```
adam1985d, line 160: it can also be configured as empty
`hostap_iface_steer_vaps=""`
nitpick

adam1985d (changes requested):

controller/src/beerocks/master/db/db.cpp
kantera800: so keeping it

controller/src/beerocks/master/db/db.cpp
kantera800: it will be called for each bssid in the mesh network every time optimal path task runs. 
we have `Remove candidate` print instead

controller/src/beerocks/master/tasks/optimal_path_task.cpp
kantera800: already done later in the spaghetti code

framework/platform/bpl/uci/cfg/bpl_cfg.cpp
kantera800: same thing for us

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1558,6 +1558,30 @@ std::string db::get_hostap_ssid(const std::string &mac)
...
+bool db::is_vap_on_steer_list(const std::string &bssid)
+{
+    if (config.load_steer_on_vaps.empty()) {
+        return true;
+    }
+
+    auto vap_name = get_hostap_iface_name(bssid);
+    if (vap_name == "INVALID") {
+        LOG(ERROR) << "vap name is invalid for bssid " << bssid;
+    }
```
adam1985d, line 1571: return false;

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1558,6 +1558,30 @@ std::string db::get_hostap_ssid(const std::string &mac)
...
+
+    auto vap_id = get_hostap_vap_id(bssid);
+    if (vap_id == IFACE_ID_INVALID) {
+        LOG(ERROR) << "vap id is invalid for bssid " << bssid;
+    }
```
adam1985d, line 1577: return false;

adam1985d (approved):

PR #1383, opened 2020-06-02
[BUG] agent: Channel selection request should yield a single channel selection response

### LINKED ISSUE #367: [TASK] Pass flow 4.5.2 test in testbed
# Description

Perform all steps necessary to pass the test associated with flow 4.5.2 in the certification testbed.

# Exit criteria

### LINKED ISSUE #368: 4.5.2 Channel Selection Request Message

### LINKED ISSUE #1364: [BUG] agent: Channel selection request should yield a single channel selection response
The channel selection request is forwarded to all slaves, and each slave handles it individually. As a result, each slave sends a response with a channel selection response TLV with that radio's MAC address. Since all of the responses have the same MID, only one of them actually gets sent on the network, the rest is filtered out by the transport's duplicate packet elimination.

This sometimes (often, actually) makes test 4.5.2 fail, because the packet that gets dropped is the packet that the UCC is looking for.

There are two possible solutions:
- Move the channel selection handling from son_slave to backhaul_manager. This is the step in the right direction for #1082 unified agent, but it's more work - the backhaul manager currently doesn't have any information about the current channels so that has to be added as well.
- In the son_slave, check if there is any TLV in the request targeted to that slave and ignore it if not. This is more of a workaround - it works for certification because there there is always only one radio in the channel selection request.

VladyslavTupikin:
Update channel selection according to the UML diagram

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -74,6 +74,7 @@
 #include <tlvf/wfa_map/tlvAssociatedStaLinkMetrics.h>
 #include <tlvf/wfa_map/tlvAssociatedStaTrafficStats.h>
 #include <tlvf/wfa_map/tlvBeaconMetricsQuery.h>
+#include <tlvf/wfa_map/tlvChannelPreference.h>
 #include <tlvf/wfa_map/tlvClientCapabilityReport.h>
 #include <tlvf/wfa_map/tlvClientInfo.h>
 #include <tlvf/wfa_map/tlvErrorCode.h>
@@ -83,6 +84,7 @@
 #include <tlvf/wfa_map/tlvStaMacAddressType.h>
 #include <tlvf/wfa_map/tlvSteeringPolicy.h>
 #include <tlvf/wfa_map/tlvSupportedService.h>
+#include <tlvf/wfa_map/tlvTransmitPowerLimit.h>
 
 // BPL Error Codes
 #include <bpl/bpl_cfg.h>
@@ -2216,6 +2218,9 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     case ieee1905_1::eMessageType::BEACON_METRICS_QUERY_MESSAGE: {
         return handle_1905_beacon_metrics_query(cmdu_rx, src_mac, forward_to);
     }
+    case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE: {
+        return handle_channel_selection_request(cmdu_rx, src_mac);
+    }
     default: {
         // TODO add a warning once all vendor specific flows are replaced with EasyMesh
         // flows, since we won't expect a 1905 message not handled in this function
@@ -2230,6 +2235,9 @@ bool backhaul_manager::handle_slave_1905_1_message(ieee1905_1::CmduMessageRx &cm
     switch (cmdu_rx.getMessageType()) {
     case ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE: {
         return handle_slave_ap_metrics_response(cmdu_rx, src_mac);
+    }
+    case ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE: {
+        return handle_slave_channel_selection_response(cmdu_rx, src_mac);
     }
     default: {
         LOG(DEBUG) << "Unexpected 1905 message " << int(cmdu_rx.getMessageType());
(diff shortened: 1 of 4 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4319,6 +4328,154 @@ bool backhaul_manager::add_link_metrics(const sMacAddr &reporter_al_mac,
...
+            //        << "Failed forwarding CHANNEL_SELECTION_REQUEST_MESSAGE message to son_slave";
+            //    ret = false;
+            //    continue;
+            //}
+
+            m_channel_selection_request.push_back(mac);
+            ret = true;
+            break;
+        }
+    }
```
arnout: All this is not needed, because we expect a response from *all* slaves. So what we need to do is simply to populate m_channel_selection_request with all slaves. And it has to be forwarded to the slaves, but for that we can simply return false.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4319,6 +4328,154 @@ bool backhaul_manager::add_link_metrics(const sMacAddr &reporter_al_mac,
...
+        return false;
+    }
+
+    //LOG(DEBUG) << "VT channel_selection_response->radio_uid(): "
+    //           << tlvf::mac_to_string(channel_selection_response->radio_uid());
+    //LOG(DEBUG) << "VT channel_selection_response->response_code(): "
+    //           << int(channel_selection_response->response_code());
+
+    m_channel_selection_response.push_back({channel_selection_response->radio_uid(),
+                                            uint8_t(channel_selection_response->response_code())});
```
arnout: (nitpick) Thanks to @mariomaz, I discovered that we can use emplace_back instead of push_back. It avoids an additional copy.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -552,6 +556,14 @@ class backhaul_manager : public btl::transport_socket_thread {
     std::vector<sApMetricsQuery> m_ap_metric_query;
     std::vector<sApMetricsResponse> m_ap_metric_response;
 
+    struct sChannelSelectionResponse {
+        sMacAddr radio_mac;
+        uint8_t response_code;
+    };
+
+    std::vector<sMacAddr> m_channel_selection_request;
+    std::vector<sChannelSelectionResponse> m_channel_selection_response;
```
arnout: We should also store the MID, and check it when collecting responses.
Therefore, I think it would be nicer like this:
```suggestion
   struct sExpectedChannelSelection {
       uint16_t mid;
       std::vector<sMacAddr> requested;
       std::vector<sChannelSelectionResponse> responses;
   };
   sExpectedChannelSelection m_expected_channel_selection;
```

arnout (dismissed):
Although the changes aren't that big, I'd like to re-review.

Also, still two things are missing:
- Final commit should have the magic message to run the 4.5.2 test.
- test_flows should be updated: there are a few TODOs in the channel selection that are now fixed.

In fact, the test flows stuff is how you should test if the issue is really fixed. So it's even better to first do the test_flows changes and then modify the code so the test passes.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: emphase_back does not work for the marked case.
It will be work if i add 3 useless strings for creating and filling structure elements, then structure object i will be able to put inside the vector via amphase_back

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
VladyslavTupikin: Fixed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4319,6 +4328,85 @@ bool backhaul_manager::add_link_metrics(const sMacAddr &reporter_al_mac,
...
 }
+
+bool backhaul_manager::handle_channel_selection_request(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+
+    LOG(DEBUG) << "Forwarding CHANNEL_SELECTION_REQUEST to son_slave, mid=" << std::hex << int(mid);
+    m_expected_channel_selection.mid = mid;
+    for (auto &socket : slaves_sockets) {
```
mariomaz, line 4376: If you are overwriting MID value, then I think you should first clear the list

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4319,6 +4328,85 @@ bool backhaul_manager::add_link_metrics(const sMacAddr &reporter_al_mac,
...
+        m_expected_channel_selection.requests.emplace_back(socket->radio_mac);
+    }
+    // Forward to all slaves
```
mariomaz: This is quite dangerous since we are handling this message by doing nothing and having faith that the default error handling will do the job for us. IIRC such default handling is going to change in a near future and this code will stop working.
Anyway, I think the comment should better explain what's going on here so maintainers can fix it if needed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4319,6 +4328,85 @@ bool backhaul_manager::add_link_metrics(const sMacAddr &reporter_al_mac,
...
+    // We received all responses - prepare and send response message to the controller
+    auto cmdu_header =
+        cmdu_tx.create(mid, ieee1905_1::eMessageType::CHANNEL_SELECTION_RESPONSE_MESSAGE);
+
+    if (!cmdu_header) {
+        LOG(ERROR) << "Failed building IEEE1905 CHANNEL_SELECTION_RESPONSE_MESSAGE";
+        return false;
+    }
+
+    for (auto &response : m_expected_channel_selection.responses) {
```
mariomaz:
```suggestion
    for (const auto &response : m_expected_channel_selection.responses) {
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: And also clear the list of responses

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -552,6 +556,18 @@ class backhaul_manager : public btl::transport_socket_thread {
     std::vector<sApMetricsQuery> m_ap_metric_query;
     std::vector<sApMetricsResponse> m_ap_metric_response;
 
+    struct sChannelSelectionResponse {
+        sMacAddr radio_mac;
+        uint8_t response_code;
```
mariomaz:
```suggestion
         wfa_map::tlvChannelSelectionResponse::eResponseCode response_code;
```

With this change you save two type-casts in .cpp file

mariomaz (changes requested):
Not approving because there are some changes to do

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
VladyslavTupikin: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Added advanced comment message in that place.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: In the `handle_slave_channel_selection_response` method `requests` vector erases, then if `requests` vector empty, response message will be prepared for the controller. Before sending the response to the controller,  `responses` vector erases too.

mariomaz: Sorry, I didn't explain myself properly: I meant you have to consider the case when agent receives a CHANNEL_SELECTION_REQUEST **before** it has completely processed a previous one.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4319,6 +4327,89 @@ bool backhaul_manager::add_link_metrics(const sMacAddr &reporter_al_mac,
...
     return true;
 }
+
+bool backhaul_manager::handle_channel_selection_request(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+
+    LOG(DEBUG) << "Forwarding CHANNEL_SELECTION_REQUEST to son_slave, mid=" << std::hex << int(mid);
+    m_expected_channel_selection.mid = mid;
```
arnout, line 4373: As @mariomaz said, there should be something like this:

```suggestion
    // Clear previous request, if any
    m_expected_channel_selection.requests.clear();
    m_expected_channel_selection.responses.clear();
    
    m_expected_channel_selection.mid = mid;
```

VladyslavTupikin: added proposed fix into the code

mariomaz (approved):

rmelotte (approved):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: > This is quite dangerous since we are handling this message by doing nothing and having faith that the default error handling will do the job for us. IIRC such default handling is going to change in a near future and this code will stop working.

Yes it will change, but in the best way possible: there will be no slaves any more, so no forwarding needs to be done at all.

Issue #1328, opened 2020-05-21
[TASK] Refactor AP metrics response collection


arnout:
## Description

The AP metrics query/response flow requires sending the query to several radios, collecting the responses from the radios, and finally sending out all the responses in a single message.

#1231 implemented that, but it has a few shortcomings:
- MID is not stored;
- response is not properly checked against expected responses;
- if several BSSIDs are queried, there is one query per BSSID -> redundant messages;
- corner cases could be checked better;
- instead of collecting responses in structures, it might be better to construct a cmdu_tx directly;
- #1388 introduced a special case for responses that are generated directly by the monitor, which means that a request with that "magic MID" will not work correctly.

mariomaz:
Provided solution must consider also the cases in which AP Metrics Response message must be sent not as a response to a previous AP Metrics Query message, but when an internal event occurs; like when periodic reporting metrics interval has elapsed or when channel utilization crosses configured threshold value.

VladyslavTupikin:
Plan for refactoring is:

1. handle_ap_metrics_query should just fill up expected_bssid vector and forward the query.
   -- save mid
   -- clear expected_bssid
   -- clear response (just do new message creation)

2. For certification 4.7.5 create 1 big ap metric query and send it to all slaves.
   -- save mid
   -- clear expected_bssid
   -- clear response (just do new creation)
   -- //UINT16_MAX should be changed on 0, remove comment above 

3. For handle_slave_ap_metrics_response
   -- check received mid with saved:
        if (mid != saved_mid) {
            if (mid == 0) 
               forward it to the controller;
               return;
            else 
               drop;
        }
   -- create tlvs according to the information from cmdu_rx
   -- remove all useless structures

Commits:
1st create stuff from UML and modify handle_ap_metrics_query
2nd Modifying send_slave_ap_metric_query_message, also add condition if (mid != saved_mid)
3rd Change structures to the cmdu_tx in handle_slave_ap_metrics_response
4th Remove all structures

VladyslavTupikin:
![image](https://user-images.githubusercontent.com/61505283/84124014-4f687d00-aa43-11ea-9afa-3729e6ba887f.png)

arnout:
@mariomaz @RanRegev Please review the design posted by @VladyslavTupikin

mariomaz:
> For certification 4.7.5 create 1 big ap metric query and send it to all slaves.

Please clarify what "create 1 big ap metric query" means.

[...]

PR #1388, opened 2020-06-03
agent: backhaul_manager: fix test 4.7.5

### LINKED ISSUE #394: 4.7.4,5,6 Per-AP Metrics Response

mariomaz:
One of the changes required to pass 4.7.6 broke test 4.7.5. The cause
of the problem is a design flaw, not an implementation bug.

Agent must send AP Metrics Response messages in three different
situations: when asked to do so with an AP Metrics Query message,
periodically when a reporting interval has been set and when channel
utilization for one radio crosses a threshold value.

Periodic reporting is handled in backhaul manager, which periodically
forwards an emulated AP Metrics Query message to all monitor threads,
**with MID set to 0**. The backhaul manager keeps track of AP Metrics
Response messages received from all monitor threads and, when all
responses have been received, they are assembled in a new AP Metrics
Response message and sent to the controller.

On the other hand, channel utilization measurement based reporting is
handled in monitor threads. When channel utilization crosses a threshold
value, monitor sends an AP Metrics Response to backhaul manager **also**
with MID set 0.

The backhaul manager has no way to distinguish if AP Metrics Response
messages received from monitor thread with MID set to 0 correspond to a
periodic metrics reporting or to a channel utilization change. Thus it
doesn't know if it has to wait for further responses from the rest of
monitor threads to assemble a full response or forward such response as
is right away.

To differentiate received AP Metrics Response messages we will use a
"special" MID value. Note that this patch is also flaw as real AP Metrics
Query messages might also have this not-so-special MID value. This is
just a quick and dirty fix to pass 4.7.5 and 4.7.6 for M1

To be fixed as part of #1328

MAP-4.7.5_ETH_FH5GH:netgear-rax40
MAP-4.7.5_ETH_FH5GL:netgear-rax40
MAP-4.7.5_ETH_FH24G:netgear-rax40

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -1029,9 +1029,15 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
                         }
                     }
                 }
-                // We must generate a new MID for the response, so set MID to 0 here.
+                // We must generate a new MID for the periodic AP Metrics Response messages that
+                // do not correspond to an AP Metrics Query message.
+                // We cannot set MID to 0 here because we must also differentiate periodic
+                // AP Metrics Response messages and messages received from monitor thread
+                // due to channel utilization crossed configured threshold value.
+                // As a temporary solution, set MID to UINT16_MAX here.
+                // TODO: to be fixed as part of #1328
                 if (!bssid_list.empty()) {
-                    send_slave_ap_metric_query_message(0, bssid_list);
+                    send_slave_ap_metric_query_message(UINT16_MAX, bssid_list);
                 } else {
                     LOG(DEBUG) << "Skipping AP_METRICS_QUERY for slave, empty BSSID list";
                 }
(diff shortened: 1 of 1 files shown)


ghost (approved):

VladyslavTupikin (approved):

mariomaz:
Be aware that AP Metrics Query message (any message indeed) can also have a MID set to 0 when the 16 bit integer wraps around. Thus 0 should not be used to mean that an AP Metrics Query message has not been received from controller.

rmelotte (approved):

arnout:
> Be aware that AP Metrics Query message (any message indeed) can also have a MID set to 0 when the 16 bit integer wraps around. Thus 0 should not be used to mean that an AP Metrics Query message has not been received from controller.

Note that this is a generic issue: the transport will in fact treat MID 0 as "generate new MID". -> #1391.

Issue #1369, opened 2020-05-31
[BUG] start monitor fails if client was associated before prplmesh start 


kantera800:
**Describe the bug** -
Client can be associated before prplmesh is up , in this case association task should start monitoring all already associated clients.
the problem is that start monitoring fails and thus stuck in a loop .

## platform 
UGW

## Steps to reproduce

1. Build
2. associate client 
3. start prplmesh controller/agent
4. See in the logs this and that association task is stuck trying to start monitoring the client .

DEBUG 10:07:29:783 <2002066752> association_handling_task.cpp[405] --> task association_handling id 8: response for start monitoring from 00:e0:92:00:01:47 for sta 86:15:e5:2b:a7:fe timed out! attempts=2
DEBUG 10:07:29:784 <2002066752> association_handling_task.cpp[80] --> task association_handling id 8: START_RSSI_MONITORING
DEBUG 10:07:29:785 <2002066752> association_handling_task.cpp[96] --> START_MONITORING_REQUEST hostap_mac=00:e0:92:00:01:47 sta_mac 86:15:e5:2b:a7:fe
DEBUG 10:07:29:796 <2002066752> association_handling_task.cpp[265] --> task association_handling id 8: **received ACTION_CONTROL_CLIENT_START_MONITORING_RESPONSE, success=false**
WARNING 10:07:30:300 <2002066752> statistics_polling_task.cpp[93] --> task statistics polling task id 1: timeout slave 00:e0:92:00:01:46
DEBUG 10:07:32:827 <2002066752> association_handling_task.cpp[405] --> task association_handling id 8: response for start monitoring from 00:e0:92:00:01:47 for sta 86:15:e5:2b:a7:fe timed out! attempts=4
DEBUG 10:07:32:827 <2002066752> association_handling_task.cpp[80] --> task association_handling id 8: START_RSSI_MONITORING
DEBUG 10:07:32:828 <2002066752> association_handling_task.cpp[96] --> START_MONITORING_REQUEST hostap_mac=00:e0:92:00:01:47 sta_mac 86:15:e5:2b:a7:fe
DEBUG 10:07:32:835 <2002066752> ire_network_optimization_task.cpp[68] --> task hostapd_tx_on_ack - ire_network_optimization id 9: done optimizing network
DEBUG 10:07:32:836 <2002066752> task_pool.cpp[80] --> erasing task hostapd_tx_on_ack - ire_network_optimization, id 9
[...]

PR #1390, opened 2020-06-03
Bugfix/existing clients association

itayx:
Currently there is an issue described here #1369
After looking into the client association sequence, I noticed that pre-existing clients were not added to the monitor's DB which causes the association sequence to fail during the `CLIENT_START_MONITORING` handling

To fix this issue, the `generate_connected_clients_events` that exists in the `ap_wlan_hal` needs to be copied to the `mon_wlan_hal` and implemented using the DWPAL.

### DIFF
--- a/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
+++ b/common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
@@ -167,6 +167,12 @@ bool mon_wlan_hal_dummy::channel_scan_dump_results()
     return false;
 }
 
+bool mon_wlan_hal_dummy::generate_connected_clients_events()
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED";
+    return false;
+}
+
 bool mon_wlan_hal_dummy::process_dummy_data(parsed_obj_map_t &parsed_obj)
 {
     char *tmp_str;
--- a/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
@@ -470,6 +470,73 @@ static bool get_scan_results_from_nl_msg(sChannelScanResults &results, struct nl
     return true;
 }
 
+static std::shared_ptr<char> generate_client_assoc_event(const std::string &event, int vap_id)
+{
+    auto msg_buff = ALLOC_SMART_BUFFER(sizeof(sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION));
+    auto msg = reinterpret_cast<sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION *>(msg_buff.get());
+
+    if (!msg) {
+        LOG(FATAL) << "Memory allocation failed";
+        return nullptr;
+    }
+
+    // Initialize the message
+    memset(msg_buff.get(), 0, sizeof(sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION));
+
+    char client_mac[MAC_ADDR_SIZE] = {0};
+    // These parameters are not needed, but the dwpal_string_to_struct_parse fails without them.
+    int supported_rates[16]  = {0};
+    int HT_MCS[16]           = {0};
+    int16_t VHT_MCS[1]       = {0};
+    char ht_cap[8]           = {0};
+    char ht_mcs[64]          = {0};
+    char vht_cap[16]         = {0};
+    char vht_mcs[24]         = {0};
+    int8_t max_tx_power      = 0;
+    size_t numOfValidArgs[7] = {0};
+
+    FieldsToParse fieldsToParse[] = {
+        {(void *)client_mac, &numOfValidArgs[0], DWPAL_STR_PARAM, NULL, sizeof(client_mac)},
+        {(void *)supported_rates, &numOfValidArgs[1], DWPAL_INT_HEX_ARRAY_PARAM,
+         "supported_rates=", sizeof(supported_rates)},
+        {(void *)ht_cap, &numOfValidArgs[2], DWPAL_STR_PARAM, "ht_caps_info=", sizeof(ht_cap)},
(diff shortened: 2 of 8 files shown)


common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -167,6 +167,12 @@ bool mon_wlan_hal_dummy::channel_scan_dump_results()
     return false;
 }
 
+bool mon_wlan_hal_dummy::generate_connected_clients_events()
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED";
+    return false;
```
kantera800, line 173: this is not an error you should return true

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -470,6 +470,74 @@ static bool get_scan_results_from_nl_msg(sChannelScanResults &results, struct nl
...
+
+    char client_mac[MAC_ADDR_SIZE] = {0};
+    int supported_rates[16]        = {0}; // example: supported_rates=8c 12 98 24 b0 48 60 6c
+    int HT_MCS[16]                 = {0};
+    int16_t VHT_MCS[1]             = {0};
+    char ht_cap[8]                 = {0};
+    char ht_mcs[64]                = {0};
+    char vht_cap[16]               = {0};
+    char vht_mcs[24]               = {0};
+    int8_t max_tx_power            = 0;
```
morantr: All of these are not required so it can be removed and also from the fieldsToParse.
just make sure the parser still work after you remove it.
If not, open an issue about it add a comment on the code about it with the issue number.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
```diff
@@ -167,6 +167,12 @@ bool mon_wlan_hal_dummy::channel_scan_dump_results()
     return false;
 }
 
+bool mon_wlan_hal_dummy::generate_connected_clients_events()
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED";
```
morantr, line 172: I don't think it will never be implemented on dummy mode, IMO you could remove it.

morantr (approved):

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -470,6 +470,74 @@ static bool get_scan_results_from_nl_msg(sChannelScanResults &results, struct nl
     return true;
 }
 
+static std::shared_ptr<char> generate_client_assoc_event(const std::string &event, int vap_id,
+                                                         bool radio_5G)
+{
+    // TODO: Change to HAL objects
```
kantera800: don't copy  TODOs
 if you want to do it then open an issue in github and mention that in the code.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -930,6 +998,72 @@ bool mon_wlan_hal_dwpal::channel_scan_dump_results()
...
+
+            if (!msg_buff)
+                break;
+
+            // update client mac
+            auto msg =
+                reinterpret_cast<sACTION_MONITOR_CLIENT_ASSOCIATED_NOTIFICATION *>(msg_buff.get());
+            client_mac = tlvf::mac_to_string(msg->mac);
+
+            event_queue_push(Event::STA_Connected, msg_buff); // send message to the AP manager
```
kantera800:
```suggestion
            event_queue_push(Event::STA_Connected, msg_buff); // send message to the monitor thread
```

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -470,6 +470,74 @@ static bool get_scan_results_from_nl_msg(sChannelScanResults &results, struct nl
     return true;
 }
 
+static std::shared_ptr<char> generate_client_assoc_event(const std::string &event, int vap_id,
+                                                         bool radio_5G)
```
kantera800: `radio_5G` is not used please remove it from the signature

kantera800 (approved):
some minor comments so approving

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: Tried parsing without the additional fieldsToParse.
The `dwpal_string_to_struct_parse` function fails without them.

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
itayx: Removed

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
itayx: Changed

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: Removed the TODO

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: Changed

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
itayx: Removed

common/beerocks/bwl/dummy/mon_wlan_hal_dummy.cpp
itayx: The function is not implemented therefore it does not do the functionality it should.
Why shouldn't it return false?
Also other non-implemented functions within the `mon_wlan_hal_dummy` return false.

PR #1316, opened 2020-05-19
deploy_ipk.sh: adding force dependencies on rdkb platform

LiorAmram:
The deploy on rdkb platform fails because of unsatisfied dependencies.
Since those dependencies do exist on the board, I decided to force them
in the meanwhile until we figure out this issue.

Signed-off-by: Lior Amram <lior.amram@intel.com>

PR #1392, opened 2020-06-04
easylogging: log uint8_t as int

arnout:
In many places, we need to log values of type uint8_t. Unfortunately,
uint8_t is a char type, which means it gets logged as an ASCII
character. To overcome this, we use the + operator or we cast to int.
Still, it's easy to forget to do that kind of conversion.

To fix all this in one fell swoop, add an operator<< overload for
uint8_t to easylogging++. It converts to int using the + operator.

This PR was triggered by https://github.com/prplfoundation/prplMesh/pull/1316#issuecomment-637491112

### DIFF
--- a/framework/external/easylogging/easylogging++.h
+++ b/framework/external/easylogging/easylogging++.h
@@ -2897,6 +2897,9 @@ return *this;\
   inline MessageBuilder& operator<<(const std::string& msg) {
     return operator<<(msg.c_str());
   }
+  inline MessageBuilder& operator<<(uint8_t c) {
+    return operator<<(+c);
+  }
   ELPP_SIMPLE_LOG(char)
   ELPP_SIMPLE_LOG(bool)
   ELPP_SIMPLE_LOG(signed short)


rmelotte (approved):

mariomaz (approved):

framework/external/easylogging/easylogging++.h
```diff
@@ -2897,6 +2897,9 @@ return *this;\
   inline MessageBuilder& operator<<(const std::string& msg) {
     return operator<<(msg.c_str());
   }
+  inline MessageBuilder& operator<<(uint8_t c) {
+    return operator<<(+c);
+  }
```
morantr, line 2902: We should add the same operator for signed char (int8_t).

morantr (approved):

VladyslavTupikin (approved):

framework/external/easylogging/easylogging++.h
arnout: That's a bit more risky, because it would make `LOG(DEBUG) << foo << '\n';` print "10" instead of a newline... And it's not so easy to find occurences where a char is really meant as a character.

PR #1401, opened 2020-06-05
agent: implement dev_set_config,backhaul,<radio>

arnout:
Implement the CAPI command for configuring an agent with wireless backhaul.
Also include test_flows for it.

### DIFF
--- a/ci/certification/netgear-rax40-bhwifi.yml
+++ b/ci/certification/netgear-rax40-bhwifi.yml
@@ -1,0 +1,933 @@
+.certification:netgear-rax40-bhwifi:
+  stage: certification-netgear-rax40-bhwifi
+  extends: .certification:netgear-rax40
+
+MAP-4.2.2_BH24G:netgear-rax40:
+  extends: .certification:netgear-rax40-bhwifi
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-4.2.2_BH24G:netgear-rax40.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-4.2.2_BH5GL:netgear-rax40:
+  extends: .certification:netgear-rax40-bhwifi
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-4.2.2_BH5GL:netgear-rax40.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-4.2.3_BHWIFI_FH24G:netgear-rax40:
+  extends: .certification:netgear-rax40-bhwifi
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-4.2.3_BHWIFI_FH24G:netgear-rax40.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-4.2.3_BHWIFI_FH5GL:netgear-rax40:
+  extends: .certification:netgear-rax40-bhwifi
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
+    # Otherwise, make it a manual job
+    - if: '$CI_COMMIT_DESCRIPTION =~ /.*MAP-4.2.3_BHWIFI_FH5GL:netgear-rax40.*/ && $CI_COMMIT_REF_NAME != "master"'
+      when: on_success
+    - when: manual
+      allow_failure: true
+
+MAP-4.2.3_BHWIFI_FH5GH:netgear-rax40:
+  extends: .certification:netgear-rax40-bhwifi
+  rules:
+    # If the last commit description contains the test name and we're not on master, run it.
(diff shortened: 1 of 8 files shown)


tests/test_flows.py
```diff
@@ -133,6 +130,18 @@ def check_cmdu_type_single(
             return None
         return cmdus[0]
 
+    def check_no_cmdu_type(
+        self, msg: str, msg_type: int, eth_src: str, eth_dst: str = None
+    ) -> [sniffer.Packet]:
+        '''Like check_cmdu_type, but check that *no* machting CMDU is found.'''
```
ghost, line 136: matching*

Why would you use this function over 
```
x = self.get_cmdu_capture_type(msg, msg_type, eth_src, eth_dst)
self.assertEqual(len(x), 0, "Got CMDU where none expected")
```
?
And what's the point of returning the result? IIRC `self.fail()` will raise an AssertionError, skipping the rest of the code in the function

tests/test_flows.py
```diff
@@ -280,6 +289,18 @@ def run_tests(self, tests):
...
+        '''Check behaviour of dev_reset_default CAPI command.'''
+        agent = env.agents[0]
+        agent.cmd_reply("dev_reset_default,devrole,agent,program,map,type,DUT")
+        env.checkpoint()
+        time.sleep(2)
+        self.check_no_cmdu_type("autoconfig search while in reset", 0x0007, agent.mac)
+        env.checkpoint()
+        agent.cmd_reply("dev_set_config,backhaul,eth")
+        time.sleep(1)
+        self.check_cmdu_type("autoconfig search", 0x0007, agent.mac)
```
ghost, line 302: Shouldn't you also test
```
agent.cmd_reply("dev_set_config,backhaul,radio")
```
at this point? I though that was the point of the PR ...

arnout:
It's not nearly sufficient to progress in the certification test, but it's something at least, and IMO good enough to be merged. It's possible that some of the code will need to be modified again, but so be it.

ghost:
flake8 failed in CI:
```
79 tests/test_flows.py:317:1: W191 indentation contains tabs
80 tests/test_flows.py:317:1: E101 indentation contains mixed spaces and tabs
81 tests/test_flows.py:317:3: E116 unexpected indentation (comment)
82 tests/test_flows.py:318:1: E101 indentation contains mixed spaces and tabs
```

tests/test_flows.py
arnout: It's (currently) not using the unittest module, so there's no assertEqual etc. fail is a custom function in TestFlows.

In many cases, it's convenient to not abort right away on failure - that's why in googletest we usually use `EXPECT_*` rather than `ASSERT_*`. In particular, I often have a few failing checks while developing and then I'd prefer the test to continue rather than abort. Thus, I'm personally not very keen on moving to unittest.

tests/test_flows.py
arnout: There's a separate commit that adds that test. I wanted to have a baseline first. I could have done that in a separate PR.

BTW, the syntax is dev_set_config,backhaul,<radio_uid>.

(I suspect that you reviewed this while still in draft).

tests/test_flows.py
```diff
@@ -301,6 +301,31 @@ def test_dev_reset_default(self):
         time.sleep(1)
         self.check_cmdu_type("autoconfig search", 0x0007, agent.mac)
 
+    def test_capi_wireless_onboarding(self):
+        '''Check configuration of wireless backhaul.'''
+        agent = env.agents[0]
+
+        # Step 1: reset
+        agent.cmd_reply("dev_reset_default,devrole,agent,program,map,type,DUT")
```
VladyslavTupikin, line 309: Skipped whitespaces before ','

VladyslavTupikin (approved):

common/beerocks/bwl/dummy/sta_wlan_hal_dummy.cpp
```diff
@@ -17,7 +17,8 @@ namespace bwl {
 namespace dummy {
 
 sta_wlan_hal_dummy::sta_wlan_hal_dummy(const std::string &iface_name, hal_event_cb_t callback)
-    : base_wlan_hal(), base_wlan_hal_dummy(bwl::HALType::Station, iface_name, callback, {})
+    : base_wlan_hal(bwl::HALType::Station, iface_name, IfaceType::Intel, callback, {}),
```
rmelotte, line 20: If base_wlan_hal cannot be safely default-constructed, maybe we could delete the default constructor?
It would require every derived class to call it explicitly so it's more work, but it would avoid the same issue from happening again in the future (I didn't go through the full inheritance tree, so I might have missed something).

rmelotte (approved):

ci/certification/netgear-rax40-bhwifi.yml
```diff
@@ -0,0 +1,934 @@
+.certification:netgear-rax40-bhwifi:
+  stage: certification-netgear-rax40-bhwifi
+  extends: .certification:netgear-rax40
+
+MAP-4.2.2_BH24G:netgear-rax40-bhwifi:
```
arnout: Actually, this is not good: it should be `MAP-4.2.2_BH24G:netgear-rax40` so without the -bhwifi, otherwise magic comment doesn't work as expected (you'd have to set MAP-4.2.2_BH24G:netgear-rax40-bhwifi instead of MAP-4.2.2_BH24G:netgear-rax40).

rmelotte: Yes indeed...

common/beerocks/bwl/dummy/sta_wlan_hal_dummy.cpp
arnout: We actually want to remove that entire inheritance hierarchy...

tests/test_flows.py
arnout: If you mean `dev_reset_default, devrole, ...` - no, CAPI doesn't allow that.

PR #1403, opened 2020-06-05
Update AUTHORS.md file

vitalybu:
Segev is the original creator and the first architect of prplMesh.

Signed-off-by: Vitaly Bukhovsky <vitaly@devalore.com>


vitalii-komisarenko (approved):

arnout (approved):
Thank you, I didn't know that!

If anybody else is missing, please add them.

BTW, we could also add (Intel Corporation) behind the Intel people, which would allow us to remove the separate entry for Intel.

Issue #866, opened 2020-02-21
[TASK] Automatic switching between wired and wireless


ghost:
## Description

As part of wireless backhaul support, we need to be able to detect when the backhaul link being used goes down, and switch to a different available backhaul link immediately. 

If wired backhaul link goes up while using a wireless backhaul link, we should switch to the wired backhaul link. Some consideration of link quality could also be useful: wired is not necessarily faster than wireless. Frequent switching of backhaul link for performance reasons should be avoided as the switching itself might take its time to complete.

## Exit criteria

- [ ] Dynamic backhaul switching has been implemented, and works intelligently
- [ ] The test(s) defined in #863 regarding dynamic backhaul switching pass
- [ ] There is a mechanism present to prevent overly frequent backhaul link switching (and this is tested)

## Detailed description

See #861 
To be fleshed out by the implementer
Note that this does not include explicit switching of backhaul link based on messages from the controller; that will be a separate task

vitalybu:
Today there's already some existing logic in the backhaul manager for switching between wired and wireless backhaul. It's not efficient (polling the interface state), and we actually got a bug report for high CPU usage caused by it.

Since at this point the wireless backhaul is not full supported by prplMesh anyway, a decision was made to disable this mechanism, until the implementation of this task.

When this task will be implemented, we should consider creating an event-based interface monitor (based on netlink events) to replace the current polling-based implementation.

arnout:
The existing mechanism for dynamic switching will always use the wireless backhaul if it is up. That's fine. However, @mariomaz needs to be able to select wireless backhaul when wired *is* connected. In addition, we need this in certification to do the wireless backhaul tests.

The problem then is that we have to make sure the wired backhaul is not used at all, or we're creating a loop between the two backhauls. The easiest way to do that IMO is to remove the wired backhaul from the bridge when unused. We could also power down the interface, but that's less universal (it usually works for direct ethernet ports, but not necessarily for switch ports, or it may require a non-standard mechanism for powering down).

mariomaz:
Transition from wired to wireless might be decided by controller after comparing link metrics. This is a special case of backhaul optimization and could be implemented by sending a Backhaul Steering Request message.

Also, depending on link metrics, controller might decide to switch back to wired backhaul, but that's only possible if wired link is up and running.

tomereli:
DR available in https://github.com/prplfoundation/prplMesh/pull/1423

PR #1408, opened 2020-06-08
[BHWIFI] Delay onboarding until dev_start_wps is done

arnout:
In wireless backhaul, the bSTA is not supposed to start onboardin until dev_start_wps is done.
Actually, it can start earlier, but it shouldn't go over the wired backaul, so this is more of a workaround until #866 is done.


arnout:
This work will be taken over by Tomer.

PR #1409, opened 2020-06-08
bugfix / beerocks crashes after receiving DFS_NOP_FINISHED

itayx:
We noticed that the `set_supported_channel_radar_affected` could receive an empty channel list,
and when that happens, the function reaches a **segmentation fault**

To fix this, a validation of the channel list's size is needed.
In case an empty channel list is sent, print an appropriate error message & return false


kantera800 (approved):

kantera800:
PR should have a title

VladyslavTupikin (approved):

PR #1411, opened 2020-06-09
Bugfix / update dwpal_bw_to_beerocks_bw to fit eWiFiBandwidth

itayx:
After eWiFiBandwidth was updated, the dwpal_bw_to_beerocks_bw static function was not updated, which caused an error during the handling of the DFS_NOP_FINISHED event

### DIFF
--- a/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
@@ -98,32 +98,25 @@ static ap_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
 
 static uint8_t dwpal_bw_to_beerocks_bw(const uint8_t chan_width)
 {
+    // clang-format off
+    std::map<uint8_t, beerocks::eWiFiBandwidth> bandwidths {
+        // prplMesh does not distinguish between 'CHAN_WIDTH_20_NOHT' and 'CHAN_WIDTH_20'
+        { 0 /*CHAN_WIDTH_20_NOHT*/, beerocks::BANDWIDTH_20 },
+        { 1 /*CHAN_WIDTH_20     */, beerocks::BANDWIDTH_20 },
+        { 2 /*CHAN_WIDTH_40     */, beerocks::BANDWIDTH_40 },
+        { 3 /*CHAN_WIDTH_80     */, beerocks::BANDWIDTH_80 },
+        { 4 /*CHAN_WIDTH_80P80  */, beerocks::BANDWIDTH_80_80 },
+        { 5 /*CHAN_WIDTH_160    */, beerocks::BANDWIDTH_160 },
+    };
+    // clang-format on
 
-    // 0 => CHAN_WIDTH_20_NOHT
-    // 1 => CHAN_WIDTH_20
-    // 2 => CHAN_WIDTH_40
-    // 3 => CHAN_WIDTH_80
-    // 4 => CHAN_WIDTH_80P80
-    // 5 => CHAN_WIDTH_160
-    // 6 => CHAN_WIDTH_UNKNOWN
-    uint8_t bw = chan_width;
-
-    if (chan_width > 6) {
-        LOG(ERROR) << "Invalid bandwidth value: " << int(chan_width);
-        return 0;
+    auto it = bandwidths.find(chan_width);
+    if (bandwidths.end() == it) {
+        LOG(ERROR) << "Invalid bandwidth value: " << chan_width;
+        return beerocks::BANDWIDTH_UNKNOWN;
     }
 
-    // Convert to beerocks value (we have only one value for 20Mhz bandwidth)
-    if (chan_width > 0) {
-        bw -= 1;
-    }
-
-    // Treat 80P80 as 160
-    if (chan_width >= 3) {
-        bw = (uint8_t)BANDWIDTH_160;
-    }
-
-    return bw;
+    return it->second;
 }
 
 static void get_ht_mcs_capabilities(int *HT_MCS, std::string &ht_cap_str,


common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -98,32 +98,28 @@ static ap_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
...
-    // Convert to beerocks value (we have only one value for 20Mhz bandwidth)
-    if (chan_width > 0) {
-        bw -= 1;
-    }
-
-    // Treat 80P80 as 160
-    if (chan_width >= 3) {
-        bw = (uint8_t)BANDWIDTH_160;
+    if (chan_width == 0) /*CHAN_WIDTH_20_NOHT*/ {
+        //Currently prplMesh does not support No HT
```
kantera800: I'm not sure about **doesn't support** , IMO we just handle both 20 && 20_NOHT the same.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1895,6 +1895,10 @@ bool db::set_supported_channel_radar_affected(const std::string &mac, std::vecto
     }
     auto channels_count = channels.size();
     LOG(DEBUG) << " channels_count = " << int(channels_count);
+    if (channels_count < 1) {
+        LOG(ERROR) << "the given channel list must contain at least one value";
+        return false;
+    }
```
kantera800: you already committed this change . please remove the commit

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -2442,6 +2438,8 @@ bool ap_wlan_hal_dwpal::process_dwpal_event(char *buffer, int bufLen, const std:
 
         msg->params.channel   = son::wireless_utils::freq_to_channel(msg->params.frequency);
         msg->params.bandwidth = dwpal_bw_to_beerocks_bw(chan_width);
+        LOG(DEBUG) << "channel = " << int(msg->params.channel)
+                   << " bw = " << int(msg->params.bandwidth);
```
kantera800: We have this info printed on STA_Connected event .
please remove this print.

kantera800 (changes requested):
Just small changes.
Also I think @mariomaz  should take a look at this.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -1895,6 +1895,10 @@ bool db::set_supported_channel_radar_affected(const std::string &mac, std::vecto
...
     auto it =
         find_if(std::begin(n->hostap->supported_channels), std::end(n->hostap->supported_channels),
                 [&](beerocks::message::sWifiChannel supported_channel) {
```
morantr: It crashes because of the "find" on line 1905 that iterate over `n->hostap->supported_channels` and compare it to `*channels.begin()` which is the given channel list on the function.
I don't understand why we do this. Why do we compare it only to the given list first element?
I would expect to do loop within a loop to and compare all the channels on list A with all the channels on list B, that way the loop would not have been performed on zero size scenario.
Yes, your solution will fix the crash, but it seems to be a bandaid to bad and undocumented code.

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
morantr: I agree. We support it, we just do not distinguish between it and regular 20Mhz bandwidth.
So I would change it to:
`// prplMesh does not distinguish between 'CHAN_WIDTH_20_NOHT' and 'CHAN_WIDTH_20'.`

morantr (approved):

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: Changed

controller/src/beerocks/master/db/db.cpp
itayx: Didn't rebase the PR, after a rebase this is fixed

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: Removed

controller/src/beerocks/master/db/db.cpp
itayx: Moved this to a new issue #1414 because:
1) That code is already merged, and shouldn't be a part of this PR
2) This PR needs to stay focused on fixing the crush issue

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -98,32 +98,27 @@ static ap_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
...
+        1 => CHAN_WIDTH_20       | 1 => BANDWIDTH_20
+        2 => CHAN_WIDTH_40       | 2 => BANDWIDTH_40
+        3 => CHAN_WIDTH_80       | 3 => BANDWIDTH_80
+        4 => CHAN_WIDTH_80P80    | 4 => BANDWIDTH_80_80
+        5 => CHAN_WIDTH_160      | 5 => BANDWIDTH_160
+        6 => CHAN_WIDTH_UNKNOWN  | 0 => BANDWIDTH_UNKNOWN
+    */
+
+    if (chan_width == 6) /*CHAN_WIDTH_UNKNOWN*/ {
         LOG(ERROR) << "Invalid bandwidth value: " << int(chan_width);
```
mariomaz: There is no need to cast uint8_t to int since #1392

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -98,32 +98,27 @@ static ap_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
...
-        return 0;
+        return int(eWiFiBandwidth::BANDWIDTH_UNKNOWN);
```
mariomaz: Should cast to uint8_t

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -98,32 +98,27 @@ static ap_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
...
-    if (chan_width > 0) {
-        bw -= 1;
-    }
-
-    // Treat 80P80 as 160
-    if (chan_width >= 3) {
-        bw = (uint8_t)BANDWIDTH_160;
+    if (chan_width == 0) /*CHAN_WIDTH_20_NOHT*/ {
+        // prplMesh does not distinguish between 'CHAN_WIDTH_20_NOHT' and 'CHAN_WIDTH_20'.
+        return int(eWiFiBandwidth::BANDWIDTH_20);
```
mariomaz: Should cast to uint8_t

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -98,32 +98,27 @@ static ap_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
...
     }
 
-    return bw;
+    return chan_width;
```
mariomaz: This is not a valid eWiFiBandwith if chan_with > 6

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
```diff
@@ -98,32 +98,27 @@ static ap_wlan_hal::Event dwpal_to_bwl_event(const std::string &opcode)
...
 }
```
mariomaz, line 120: Let me suggest an implementation like this one:
```c
static uint8_t dwpal_bw_to_beerocks_bw(const uint8_t chan_width)
{
    // clang-format off
    std::map<uint8_t, beerocks::eWiFiBandwidth> bandwidths {
        // prplMesh does not distinguish between 'CHAN_WIDTH_20_NOHT' and 'CHAN_WIDTH_20'
        { 0, beerocks::BANDWIDTH_20 },
        { 1, beerocks::BANDWIDTH_20 },
        { 2, beerocks::BANDWIDTH_40 },
        { 3, beerocks::BANDWIDTH_80 },
        { 4, beerocks::BANDWIDTH_80_80 },
        { 5, beerocks::BANDWIDTH_160 },
    };
    // clang-format on

    auto it = bandwidths.find(chan_width);
    if (bandwidths.end() == it) {
        LOG(ERROR) << "Invalid bandwidth value: " << chan_width;
        return beerocks::BANDWIDTH_UNKNOWN;
    }

    return it->second;
}
```

mariomaz (changes requested):
Function returns an invalid eWiFiBandwith if chan_with parameter is greater than 6

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: Didn't know that, makes life much easier

common/beerocks/bwl/dwpal/ap_wlan_hal_dwpal.cpp
itayx: I agree
Updated

mariomaz (approved):
LGTM

kantera800 (approved):

PR #1412, opened 2020-06-09
[agent]: Refactor handlers for AP_METRICS_RESPONSE_MES…

### LINKED ISSUE #1328: [TASK] Refactor AP metrics response collection

VladyslavTupikin:
Create new structure sExpectedApMetricsResponse for
saving mid, bssid vector, cmdutx message. Update
current implementation of handle_ap_metrics_query:
clear bssid vector, cmdutx message, save mid, change
sending message to the son_slave on simple forwarding.

Add new structure and member of structure, update
handle_ap_metrics_query method.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2604,43 +2604,71 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
     }
 
     auto bssid_tlv = ap_metrics_tlv->bssid();
-    auto mac       = std::find_if(
-        m_ap_metric_query.begin(), m_ap_metric_query.end(),
-        [&bssid_tlv](sApMetricsQuery const &query) { return query.bssid == bssid_tlv; });
+    auto mac       = std::find_if(m_expected_ap_metrics_response.expected_bssids.begin(),
+                            m_expected_ap_metrics_response.expected_bssids.end(),
+                            [&bssid_tlv](sMacAddr const &query) { return query == bssid_tlv; });
 
-    if (mac == m_ap_metric_query.end()) {
+    if (mac == m_expected_ap_metrics_response.expected_bssids.end()) {
         LOG(ERROR) << "Failed search in ap_metric_query for bssid: " << bssid_tlv
                    << " from mid=" << std::hex << int(mid);
         return false;
     }
 
-    sApMetrics metric;
-    // Copy data to the response vector
-    metric.bssid               = ap_metrics_tlv->bssid();
-    metric.channel_utilization = ap_metrics_tlv->channel_utilization();
-    metric.number_of_stas_currently_associated =
-        ap_metrics_tlv->number_of_stas_currently_associated();
-    metric.estimated_service_parameters = ap_metrics_tlv->estimated_service_parameters();
-    auto info                           = ap_metrics_tlv->estimated_service_info_field();
-    for (size_t i = 0; i < ap_metrics_tlv->estimated_service_info_field_length(); i++) {
-        metric.estimated_service_info_field.push_back(info[i]);
-    }
-    std::vector<sStaTrafficStats> traffic_stats_response;
+    m_expected_ap_metrics_response.response.create(
+        mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
 
+    // Prepare AP Metrics TLV
+    auto ap_metrics_response_tlv =
+        m_expected_ap_metrics_response.response.addClass<wfa_map::tlvApMetrics>();
+    if (!ap_metrics_response_tlv) {
+        LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetrics>";
+        return false;
+    }
+    ap_metrics_response_tlv->bssid()               = ap_metrics_tlv->bssid();
+    ap_metrics_response_tlv->channel_utilization() = ap_metrics_tlv->channel_utilization();
+    ap_metrics_response_tlv->number_of_stas_currently_associated() =
+        ap_metrics_tlv->number_of_stas_currently_associated();
+    ap_metrics_response_tlv->estimated_service_parameters() =
+        ap_metrics_tlv->estimated_service_parameters();
+
+    if (!ap_metrics_response_tlv->alloc_estimated_service_info_field(
+            ap_metrics_tlv->estimated_service_info_field_length())) {
+        LOG(ERROR) << "Couldn't allocate "
+                      "ap_metrics_response_tlv->alloc_estimated_service_info_field";
+        return false;
+    }
+    auto rx_info = ap_metrics_tlv->estimated_service_info_field();
+    auto tx_info = ap_metrics_response_tlv->estimated_service_info_field();
+    for (size_t i = 0; i < ap_metrics_tlv->estimated_service_info_field_length(); i++) {
+        tx_info[i] = rx_info[i];
+    }
+
+    // Prepare STA Traffic Stats TLV's
     for (auto &sta_traffic : cmdu_rx.getClassList<wfa_map::tlvAssociatedStaTrafficStats>()) {
         if (!sta_traffic) {
             LOG(ERROR) << "Failed to get class list for tlvAssociatedStaTrafficStats";
             continue;
         }
 
-        traffic_stats_response.push_back(
-            {sta_traffic->sta_mac(), sta_traffic->byte_sent(), sta_traffic->byte_recived(),
-             sta_traffic->packets_sent(), sta_traffic->packets_recived(),
-             sta_traffic->tx_packets_error(), sta_traffic->rx_packets_error(),
-             sta_traffic->retransmission_count()});
+        auto sta_traffic_response_tlv = m_expected_ap_metrics_response.response
+                                            .addClass<wfa_map::tlvAssociatedStaTrafficStats>();
+
+        if (!sta_traffic_response_tlv) {
+            LOG(ERROR) << "Failed addClass<wfa_map::tlvAssociatedStaTrafficStats>";
+            continue;
+        }
+
+        sta_traffic_response_tlv->sta_mac()              = sta_traffic->sta_mac();
+        sta_traffic_response_tlv->byte_sent()            = sta_traffic->byte_sent();
+        sta_traffic_response_tlv->byte_recived()         = sta_traffic->byte_recived();
+        sta_traffic_response_tlv->packets_sent()         = sta_traffic->packets_sent();
+        sta_traffic_response_tlv->packets_recived()      = sta_traffic->packets_recived();
+        sta_traffic_response_tlv->tx_packets_error()     = sta_traffic->tx_packets_error();
+        sta_traffic_response_tlv->rx_packets_error()     = sta_traffic->rx_packets_error();
+        sta_traffic_response_tlv->retransmission_count() = sta_traffic->retransmission_count();
     }
 
-    std::vector<sStaLinkMetrics> link_metrics_response;
+    // Prepare STA Link Metrics TLV's
     for (auto &sta_link_metric : cmdu_rx.getClassList<wfa_map::tlvAssociatedStaLinkMetrics>()) {
         if (!sta_link_metric) {
             LOG(ERROR) << "Failed getClassList<wfa_map::tlvAssociatedStaLinkMetrics>";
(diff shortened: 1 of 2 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2552,14 +2559,23 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
...
     }
 
-    if (!send_slave_ap_metric_query_message(mid, bssid)) {
-        LOG(ERROR) << "Failed to forward AP_METRICS_RESPONSE to the son_slave_thread";
-        return false;
+    LOG(DEBUG) << "Forwarding AP_METRICS_QUERY_MESSAGE to all slaves, mid=" << std::hex << int(mid);
+    uint16_t length = message_com::get_uds_header(cmdu_rx)->length;
+    cmdu_rx.swap(); // swap back before forwarding
+    for (const auto &soc : slaves_sockets) {
+        if (!message_com::forward_cmdu_to_uds(soc->slave, cmdu_rx, length)) {
```
mariomaz, line 2569: If CMDU fails to be forwarded to one slave, backhaul manager will never get the expected response from that slave.
It's pointless forwarding the CMDU to the rest of the slaves, therefore loop should break.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2552,14 +2559,23 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
...
+            LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to slave: "
+                       << tlvf::mac_to_string(soc->radio_mac);
```
mariomaz, line 2571:
```suggestion
                       << soc->radio_mac;
```
There's no need to convert MAC to string

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -519,6 +520,15 @@ class backhaul_manager : public btl::transport_socket_thread {
                           const sLinkNeighbor &link_neighbor, const sLinkMetrics &link_metrics,
                           ieee1905_1::eLinkMetricsType link_metrics_type);
 
+    struct sExpectedApMetricsResponse {
```
mariomaz, line 523:
```suggestion
    struct sApMetricsResponse {
```
I think we should reconsider the name for this struct, though I'm not happy with this name either. Suggestions are welcome.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -519,6 +520,15 @@ class backhaul_manager : public btl::transport_socket_thread {
...
+        uint16_t mid;
+        std::vector<sMacAddr> expected_bssids;
+        uint8_t response_buffer[message::MESSAGE_BUFFER_LENGTH];
+        ieee1905_1::CmduMessageTx response = {response_buffer, sizeof(response_buffer)};
+    };
+
+    sExpectedApMetricsResponse m_expected_ap_metrics_response;
```
mariomaz, line 530:
```suggestion
    sApMetricsResponse m_ap_metrics_response_under_construction;
```
The same goes for the variable.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1235,51 +1230,52 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+        i++;
+        m_ap_metric_query.push_back({nullptr, mac}); // Old implementation will be removed
+    }
+
+    auto ret = false;
+    for (auto socket : slaves_sockets) {
+        if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
+            LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
+            ret = false;
+            continue;
```
mariomaz, line 1270: No need for `continue`

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1235,51 +1230,52 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+    for (const auto &mac : bssid_list) {
+        auto list         = query->bssid_list(i);
+        std::get<0>(list) = true;
+        std::get<1>(list) = mac;
+        i++;
+        m_ap_metric_query.push_back({nullptr, mac}); // Old implementation will be removed
+    }
+
+    auto ret = false;
+    for (auto socket : slaves_sockets) {
```
mariomaz, line 1266:
```suggestion
    for (const auto &socket : slaves_sockets) {
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2585,33 +2581,21 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
...
-     * to controller.
-     * This might happen when channel utilization value has crossed configured threshold or when
-     * periodic metrics reporting interval has elapsed.
-     */
-    if (0 == mid) {
-        uint16_t length = message_com::get_uds_header(cmdu_rx)->length;
-        cmdu_rx.swap(); //swap back before forwarding
-        return send_cmdu_to_bus(cmdu_rx, controller_bridge_mac, bridge_info.mac, length);
+    if (mid != m_expected_ap_metrics_response.mid) {
+        LOG(ERROR) << "Received AP_METRICS_RESPONSE_MESSAGE with bad mid=" << std::hex << int(mid);
```
mariomaz, line 2584: Please log expected mid and actual mid

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2585,33 +2581,21 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
...
-     * Note that this design is flaw as a real query might also have this special mid value. This
-     * is just a quick and dirty fix to pass 4.7.5 and 4.7.6 for M1
-     * TODO: to be fixed as part of #1328
-     */
-    if (UINT16_MAX == mid) {
-        mid = 0;
+    if (mid == 0) {
+        LOG(DEBUG) << "Forward AP_METRICS_RESPONSE_MESSAGE to controller, mid=" << std::hex
+                   << int(mid);
+        return send_cmdu_to_bus(cmdu_rx, controller_bridge_mac, bridge_info.mac,
```
mariomaz, line 2591: Now if mid is equal to 0, it can be either because channel utilization threshold is crossed or because of we are in the middle of a periodic reporting. At this point you don't know it. 
However processing of the response should be different in each case: forward as is vs. save response and wait till all slaves have replied.

mariomaz (changes requested):
I don't fully agree with adopted design. Also some of the tests are failing. I'll want to re-review when they pass.

arnout:
I haven't looked at the code, but the current implementation clearly doesn't do the right thing. E.g. [the 4.7.5 tests fail](https://gitlab.com/prpl-foundation/prplMesh/-/jobs/590867621/artifacts/browse/logs/FAIL/MAP-4.7.5_ETH_FH5GH/) because there are still 2 responses being sent (one of which is empty), as can be seen in the [pcap file](https://gitlab.com/prpl-foundation/prplMesh/-/jobs/590867621/artifacts/raw/logs/FAIL/MAP-4.7.5_ETH_FH5GH/pcap/_MAP-4.7.5-5GH-STEP5-11). There should be only one response.

test_flows should verify this. We currently don't have any test flow for 4.7.5 or 4.7.6, so that really should be added.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2552,14 +2559,23 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
             LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
             return false;
         }
-        bssid.push_back(std::get<1>(bssid_tuple));
+        auto mac = std::get<1>(bssid_tuple);
```
arnout: (nitpick) please use bssid for a bssid.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1235,51 +1230,51 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
-            i++;
+    if (!query->alloc_bssid_list(bssid_list.size())) {
+        LOG(ERROR) << "Failed allocate memory for bssid_list, size: " << bssid_list.size();
+        return false;
+    }
+
+    int i = 0;
+    for (const auto &mac : bssid_list) {
+        auto list         = query->bssid_list(i);
+        std::get<0>(list) = true;
```
arnout, line 1260: This is wrong - you have to check the first element of the tuple, not set it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1235,51 +1230,51 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
-                }
 
-                auto query = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
-                if (!query) {
-                    LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetricQuery>";
-                    return false;
-                }
+    // Save mid and clear previous query
+    m_expected_ap_metrics_response.mid = mid;
+    m_expected_ap_metrics_response.expected_bssids.clear();
```
arnout, line 1236: It is cleared here, but never set... This should be
```suggestion
    m_expected_ap_metrics_response.expected_bssids = bssid_list;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1235,51 +1230,51 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+        auto list         = query->bssid_list(i);
+        std::get<0>(list) = true;
+        std::get<1>(list) = mac;
+        i++;
+    }
+
+    auto ret = false;
+    for (auto socket : slaves_sockets) {
+        if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
+            LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
```
arnout, line 1268: Nitpick: it's not forwarding, just sending.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: It's not a "bad" MID.

More importantly: if it's MID 0, it should be forwarded as is, because then it's a channel utlilization treshold crossing (4.7.6). And in fact, I would always forward it if the MID is not the expected MID - it gives us duplicates if we had two parallel events, but as I mentioned dozens of times: that's fine.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2569,33 +2579,21 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
...
-     * TODO: to be fixed as part of #1328
-     */
-    if (UINT16_MAX == mid) {
-        mid = 0;
+    if (mid == 0) {
+        LOG(DEBUG) << "Forward AP_METRICS_RESPONSE_MESSAGE to controller, mid=" << std::hex
+                   << int(mid);
+        return send_cmdu_to_bus(cmdu_rx, controller_bridge_mac, bridge_info.mac,
+                                cmdu_rx.getMessageLength());
     }
```
arnout, line 2593: This entire condition should be removed.

RanRegev:
#1467 replaces this PR.

PR #1413, opened 2020-06-10
Feature/prplwrt build with a script

rmelotte:
Openwrt: move the build to a separate script, and add an option to build an "openwrt-builder" image only.

Moving the build part to a script makes it more readable, and easier
to maintain.

It also makes it easier to repeat the same steps in a container
interactively, to debug a failing build for example.

As a bonus, it makes it possible to run shellcheck and formatting
tools on it.

Some developers like to build OpenWrt interactively, to configure
different build options for example, or to be able to debug failing
builds more easily than what can be done when building a docker image.

With these changes, it's now possible to build the "openwrt-builder"
image:

tools/docker/builder/openwrt/build.sh --docker-target-stage
openwrt-builder --target-device netgear-rax40

The image can then be used to compile OpenWrt interactively.

### DIFF
--- a/tools/docker/builder/openwrt/Dockerfile
+++ b/tools/docker/builder/openwrt/Dockerfile
@@ -29,9 +29,9 @@ RUN apt-get update \
 RUN useradd -ms /bin/bash openwrt
 
 ####
-## Prebuilt OpenWrt and prplMesh dependencies, but not the prplMesh ipk itself
+## OpenWrt tree and scripts to build it.
 ####
-FROM openwrt-prerequisites as prplmesh-builder
+FROM openwrt-prerequisites as openwrt-builder
 
 USER openwrt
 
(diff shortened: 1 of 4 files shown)


tools/docker/builder/openwrt/Dockerfile
```diff
@@ -33,29 +40,38 @@ USER openwrt
 #
 # OpenWrt repository to use. Can also be a prplWrt repository:
 ARG OPENWRT_REPOSITORY
+ENV OPENWRT_REPOSITORY=$OPENWRT_REPOSITORY
```
tomereli, line 43: Will this set an environment variable available in the container?

tomereli (approved):

tools/docker/builder/openwrt/Dockerfile
rmelotte: Yes exactly, it makes them also available at run time.

arnout (approved):
Instead of adding the --docker-target-stage option (which is a bit annoying as an open-ended option, because it really can only be openwrt-prerequisites, openwrt-builder or prplmesh-builder), why don't we just tag all three of the, and change the --tag option to set only the part behind the colon?

That is however something that can be done in a separate commit/PR.

rmelotte:
@arnout Good idea. Since merging this is not urgent, I'll leave it open for now and update it.

rmelotte:
I have some more changes to make that will need to be reviewed anyway, so in the end I will merge this one first and open a new PR.

PR #1418, opened 2020-06-11
Socket refactorings

### LINKED ISSUE #546: Proper implementation of Socket
There are several things wrong with the Socket and beerocks_socket implementations:

- In general, readBytes() should be called in a loop to fetch the entire message;
- If a message is bigger than the buffer size, the socket becomes unusable (no recv() is ever done on it);

Probably more as well. It needs review.

vitalii-komisarenko:
https://github.com/prplfoundation/prplMesh/issues/546

This PR contains two changes:
1) added flag to prevent socket from returning partial data;
2) added method to discard very large messages.

### DIFF
--- a/common/beerocks/bcl/include/bcl/network/socket.h
+++ b/common/beerocks/bcl/include/bcl/network/socket.h
@@ -52,7 +52,7 @@ public:
 
     ssize_t getBytesReady();
     ssize_t readBytes(uint8_t *buf, size_t buf_size, bool blocking, size_t buf_len = 0,
-                      bool isPeek = false);
+                      bool isPeek = false, bool waitAll = false);
     ssize_t writeBytes(const uint8_t *buf, size_t buf_len, int port = 0,
                        struct sockaddr_in addr_in = {});
     ssize_t writeString(std::string msg)
--- a/common/beerocks/bcl/include/bcl/network/socket.h
+++ b/common/beerocks/bcl/include/bcl/network/socket.h
@@ -53,6 +53,12 @@ public:
     ssize_t getBytesReady();
     ssize_t readBytes(uint8_t *buf, size_t buf_size, bool blocking, size_t buf_len = 0,
                       bool isPeek = false, bool waitAll = false);
+    /**
+     * @brief read and discard specified number of bytes
+     *
+     * @return error code if an error occured or number of discarded bytes otherwise
+     */
+    ssize_t discardBytes(size_t bytes_to_discard);
     ssize_t writeBytes(const uint8_t *buf, size_t buf_len, int port = 0,
                        struct sockaddr_in addr_in = {});
     ssize_t writeString(std::string msg)
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -151,7 +151,7 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
 
     // Check if UDS Header exists
     available_bytes = sd->readBytes(rx_buffer, sizeof(rx_buffer), true, sizeof(message::sUdsHeader),
-                                    true); // PEEK UDS Header, blocking
+                                    true, true); // PEEK UDS Header, blocking, peek the whole header
 
     if (available_bytes < sizeof(message::sUdsHeader)) {
         THREAD_LOG(ERROR) << "available bytes = " << available_bytes
@@ -164,8 +164,8 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     size_t message_size             = uds_header->length + sizeof(message::sUdsHeader);
 
     // Try to read all message
-    available_bytes =
-        sd->readBytes(rx_buffer, sizeof(rx_buffer), true, message_size); // blocking read
+    available_bytes = sd->readBytes(rx_buffer, sizeof(rx_buffer), true, message_size, false,
+                                    true); // blocking read, whole message
 
     if (available_bytes != message_size) {
         THREAD_LOG(ERROR) << "available bytes = " << available_bytes
--- a/common/beerocks/bcl/source/beerocks_socket_thread.cpp
+++ b/common/beerocks/bcl/source/beerocks_socket_thread.cpp
@@ -170,6 +170,13 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     if (available_bytes != message_size) {
         THREAD_LOG(ERROR) << "available bytes = " << available_bytes
                           << " message_size = " << message_size;
+        // discard remaining bytes
+        ssize_t bytes_to_discard = message_size - available_bytes;
+        ssize_t bytes_discarded  = sd->discardBytes(bytes_to_discard);
+        if (bytes_discarded != bytes_to_discard) {
+            THREAD_LOG(ERROR) << "discardBytes expected to return " << bytes_to_discard
+                              << " but returned " << bytes_discarded;
+        }
         return false;
     }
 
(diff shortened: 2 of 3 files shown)


common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -151,7 +151,7 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
 
     // Check if UDS Header exists
     available_bytes = sd->readBytes(rx_buffer, sizeof(rx_buffer), true, sizeof(message::sUdsHeader),
-                                    true); // PEEK UDS Header, blocking
+                                    true, true); // PEEK UDS Header, blocking, peek the whole header
```
arnout, line 154: We don't need this here, because it's a Unix Socket which is message-oriented, so recv() will always return the full message if the buffer is large enough.

I think the only place where we need it is in UCC listener, which is TCP.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -170,6 +170,13 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     if (available_bytes != message_size) {
         THREAD_LOG(ERROR) << "available bytes = " << available_bytes
                           << " message_size = " << message_size;
+        // discard remaining bytes
+        ssize_t bytes_to_discard = message_size - available_bytes;
+        ssize_t bytes_discarded  = sd->discardBytes(bytes_to_discard);
+        if (bytes_discarded != bytes_to_discard) {
+            THREAD_LOG(ERROR) << "discardBytes expected to return " << bytes_to_discard
+                              << " but returned " << bytes_discarded;
+        }
```
arnout, line 179: Again, this is not needed: on UDS, if a message doesn't fit in the buffer, the rest of the message will be discarded.

morantr:
I now see that I revied it on another PR of yours. So I'll just comment it here as well.
https://github.com/prplfoundation/prplMesh/pull/1450#discussion_r442125528

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -53,6 +53,12 @@ class Socket {
     ssize_t getBytesReady();
     ssize_t readBytes(uint8_t *buf, size_t buf_size, bool blocking, size_t buf_len = 0,
                       bool isPeek = false, bool waitAll = false);
+    /**
+     * @brief read and discard specified number of bytes
+     *
+     * @return error code if an error occured or number of discarded bytes otherwise
```
mariomaz, line 59: Add @param bytes_to_discard

common/beerocks/bcl/source/network/socket.cpp
```diff
@@ -212,6 +212,29 @@ ssize_t Socket::readBytes(uint8_t *buf, size_t buf_size, bool blocking, size_t b
...
+    constexpr size_t buf_size = 256;
+    uint8_t buf[buf_size];
+
+    size_t already_discarded = 0;
+
+    while (already_discarded < bytes_to_discard) {
+        size_t remaining_bytes = bytes_to_discard - already_discarded;
+        size_t request_size    = remaining_bytes > buf_size ? buf_size : remaining_bytes;
+
+        ssize_t len = readBytes(buf, request_size, true, request_size, false, true);
```
mariomaz, line 226: IIUC, since parameter `waitAll` is true, this method will block until `request_size` bytes are read. A malicious or flawed peer could make the program block forever.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
mariomaz, line 154: If header is not complete, program would block here forever

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -164,8 +164,8 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     size_t message_size             = uds_header->length + sizeof(message::sUdsHeader);
 
     // Try to read all message
-    available_bytes =
-        sd->readBytes(rx_buffer, sizeof(rx_buffer), true, message_size); // blocking read
+    available_bytes = sd->readBytes(rx_buffer, sizeof(rx_buffer), true, message_size, false,
```
mariomaz, line 167: If payload is not complete, program would block here forever

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -170,6 +170,13 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     if (available_bytes != message_size) {
         THREAD_LOG(ERROR) << "available bytes = " << available_bytes
                           << " message_size = " << message_size;
+        // discard remaining bytes
+        ssize_t bytes_to_discard = message_size - available_bytes;
+        ssize_t bytes_discarded  = sd->discardBytes(bytes_to_discard);
```
mariomaz, line 175: If we didn't receive all of the expected bytes, most probably the call to `discardBytes()` will fail or block

common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -170,6 +170,13 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     if (available_bytes != message_size) {
```
mariomaz, line 170: Sorry, but I don't understand how these changes match the commit description.

If this condition is true is because readBytes() could not read the whole message. According to `recv` man page and the MSG_WAITALL flag description, this can happen if a signal is caught or an error or disconnect occurs, not because message is very large.

A *very large message* is a message bigger than the buffer size so I was expecting something like `if (message_size > sizeof(rx_buffer))` somewhere in this method.

mariomaz (changes requested):
I think very large messages are not properly handled yet.
I think MSG_WAITALL cannot be part of the solution because it could block the program forever.

Method `handle_cmdu_message_uds()' should try to read as many bytes as possible into reception buffer without blocking. If full message has been received, then it should return true. Otherwise it should store the number of bytes received so far in a member variable and return false. Next call to this method would continue filling reception buffer from the offset position where it left. Measuring elapsed time since last call to the method is necessary if you want to discard rotten data and start back again.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
vitalii-komisarenko: `    explicit Socket(SOCKET s, long readTimeout = 1000);`
`    Socket(SOCKET s, const std::string &peer_ip, int port_port, long readTimeout = 1000);`
`    Socket(const std::string &uds_path = std::string(), long readTimeout = 1000);`

All `Socket` contructors set 1s read timeout by default. Probably it applies here as well.

common/beerocks/bcl/source/beerocks_socket_thread.cpp
vitalii-komisarenko: Timeout should apply here as well.
At least it is what https://stackoverflow.com/questions/45864828/msg-waitall-combined-with-so-rcvtimeo says

common/beerocks/bcl/source/beerocks_socket_thread.cpp
vitalii-komisarenko: If we don't receive the expected bytes the timeout will occur.
If we receive the missing bytes after timeout they will be considered a new message. That is indeed ugly... for TCP sockets.
UDP and UNIX sockets keep message boundaries, so maybe the `discardBytes` should be adjusted not to span accross several messages...

common/beerocks/bcl/source/beerocks_socket_thread.cpp
mariomaz: This method should not block at all.
It should read all available bytes or up to a complete message, whatever happens first. If there are more bytes available because more than one message has been received, then caller should invoke this method again.

PR #1420, opened 2020-06-11
Send per-interface Topology Discovery

### LINKED ISSUE #1160: [TASK] Per-interface Topology Discovery
## Description

Currently, we sent a single Topology Discovery message through the bridge.

However, according to IEEE1905.1, the message should include a MAC Address TLV which contains the address of the interface on which you send the message. Thus, a different message should be sent on each interface.

Therefore, the agent should send a separate Discovery message on each interface.

To support this use case, the cmdu_tx topic on the ieee1905_transport should get an additional parameter to specify the interface on which it should be sent. If empty, it will be sent on all interfaces (as currently).

mariomaz:
See #1160


VladyslavTupikin (approved):

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -525,9 +525,15 @@ std::vector<std::string> network_utils::linux_get_iface_list_from_bridge(const s
...
 
-int network_utils::linux_get_iface_index(const std::string &iface)
+uint32_t network_utils::linux_get_iface_index(const std::string &iface_name)
 {
-    return if_nametoindex(iface.c_str());
+    uint32_t iface_index = if_nametoindex(iface_name.c_str());
+    if (0 == iface_index) {
+        LOG(ERROR) << "Failed to read the index of interface " << iface_name << ": "
+                   << strerror(errno);
+    }
```
vitalybu: Since there's no error handling except logging, you could do this:
```suggestion
    LOG_IF(!iface_index, ERROR) << "Failed to read the index of interface " << iface_name << ": " 
                                << strerror(errno);
```

common/beerocks/btl/btl.cpp
```diff
@@ -41,16 +41,23 @@ void transport_socket_thread::set_select_timeout(unsigned msec) { poll_timeout_m
 bool transport_socket_thread::send_cmdu_to_bus(ieee1905_1::CmduMessageTx &cmdu_tx,
                                                const std::string &dst_mac,
                                                const std::string &src_mac)
+{
+    return send_cmdu_to_bus(cmdu_tx, 0, dst_mac, src_mac);
+}
+
+bool transport_socket_thread::send_cmdu_to_bus(ieee1905_1::CmduMessageTx &cmdu_tx,
+                                               uint32_t iface_index, const std::string &dst_mac,
```
vitalybu: Why not add the `iface_index` as the last parameter with a default value of 0?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1119,27 +1119,69 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
...
+    auto ifaces = network_utils::linux_get_iface_list_from_bridge(m_sConfig.bridge_iface);
+    for (const auto &iface_name : ifaces) {
+
+        if (!network_utils::linux_iface_is_up_and_running(iface_name)) {
+            continue;
+        }
+
+        uint32_t iface_index = network_utils::linux_get_iface_index(iface_name);
+        if (0 == iface_index) {
+            return false;
```
vitalybu: Why is this a failure?
Not being able to get an interface index for its' name is indeed an issue, but what if some other interface will work?
Maybe just logging this as a WARNING and continuing to the next interface?

common/beerocks/btl/btl.cpp
mariomaz: Because there is another method in the same class with the same name and parameters but with an extra parameter `uint16_t length` at the end.
Before my changes there were these two methods in `transport_socket_thread`:
```c
public:
bool send_cmdu_to_bus(ieee1905_1::CmduMessage &cmdu, const std::string &dst_mac, const std::string &src_mac);
protected:
bool send_cmdu_to_bus(ieee1905_1::CmduMessage &cmdu, const std::string &dst_mac, const std::string &src_mac, uint16_t length);
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Ok, I see your point and I'll adopt your suggestion.
What do you think about the rest of conditions that are also considered as a failure in this method? Shall we also log and continue?

common/beerocks/bcl/source/network/network_utils.cpp
mariomaz: Ok

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
vitalybu: As long as we'll have at least one valid interface that we can send the discovery message on, I think yes.

framework/tlvf/yaml/tlvf/ieee_1905_1/eTlvType.yaml
```diff
@@ -6,7 +6,7 @@ eTlvType:
   _type: enum_class
   _enum_storage: uint8_t
   TLV_END_OF_MESSAGE: 0x00
-  TLV_AL_MAC_ADDRESS_TYPE: 0x01
+  TLV_AL_MAC_ADDRESS: 0x01
   TLV_MAC_ADDRESS: 0x02
```
arnout, line 10: I'm not sure I agree with this change. The standard clearly says "AL MAC address type TLV" and "MAC address type TLV" (and, for example, "Device bridging capability TLV"). So if anything, TLV_MAC_ADDRESS should be changed to TLV_MAC_ADDRESS_TYPE.

That said, the standard is probably just stupid in this case :-)

common/beerocks/btl/include/btl/btl.h
```diff
@@ -38,6 +38,18 @@ class transport_socket_thread : public socket_thread {
     bool send_cmdu_to_bus(ieee1905_1::CmduMessageTx &cmdu, const std::string &dst_mac,
                           const std::string &src_mac);
 
+    /**
+     * @brief Sends CDMU to transport for dispatching.
+     *
+     * @param cmdu Control Message Data Unit to send.
+     * @param iface_index Network interface index (set to 0 to send on all interfaces).
```
arnout: We agreed at some point that going forward, we would use interface names everywhere, rather than interface indexes. I can't immediately find a reference to the discussion, but IIRC the reasoning was that:
- in practice, interfaces are not renamed while prplmesh is running, so interface name is as stable as interface index;
- conversely, interface index is likely to change, e.g. when an AP is reconfigured all its VAP interfaces are brought down and up again, so they will have the same name but a different ifindex;
- in addition, the name is easier for debugging.

Thus, I'd prefer an interface name rather than an index here.

This also nicely resolved @vitalybu's suggestion to give it a default value (of empty string then, rather than 0).

Obviously, at this time it means you have to do the ifname to ifindex conversion in transport_socket_thread::bus_send(), but:
- that's easy to do;
- it makes it easier in the future to change the message itself to carry an ifname rather than an ifindex.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1119,27 +1119,69 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
...
+        }
+
+        sMacAddr iface_mac;
+        if (!get_iface_mac(iface_name, iface_mac)) {
+            return false;
+        }
```
arnout: (nitpick) I think it would be more logical to move this down into the send_1905_topology_discover_message(iface) function. In general, I don't think it makes sense for functions to have parameters between which a one-to-one correspondence is expected.

arnout (approved):
Pretty nice! This PR was a dream to review...

Ideally we should have a test_flows to validate this. It's a bit tricky though, because it requires several backhaul interfaces to be emulated in docker. I'm not sure we can easily do that. Maybe, however, it's possible to attach a sniffer to a dummy interface as well, so in that case it's simply a matter of running an additional dumpcap inside the container.

I'm starting to think that @fvbogaert is right and we should be using mininet instead of containers for testing...

common/beerocks/btl/btl.cpp
```diff
@@ -41,16 +41,23 @@ void transport_socket_thread::set_select_timeout(unsigned msec) { poll_timeout_m
...
         THREAD_LOG(ERROR) << "finalize failed";
         return false;
     }
 
-    return send_cmdu_to_bus(cmdu_tx, dst_mac, src_mac, cmdu_tx.getMessageLength());
+    return send_cmdu_to_bus(cmdu_tx, iface_index, dst_mac, src_mac, cmdu_tx.getMessageLength());
 }
 
-bool transport_socket_thread::send_cmdu_to_bus(ieee1905_1::CmduMessage &cmdu,
+bool transport_socket_thread::send_cmdu_to_bus(ieee1905_1::CmduMessage &cmdu, uint32_t iface_index,
```
morantr: I don't understand it.
The message is sent to the bus with src mac which is the bridge/al_mac.
doesn't ifindex means index of radio interface? Because we send the message from the "bridge" interface. Does it mean face index of the bridge maybe? It is not clear to me.
Anyway, I don't think that it is a good idea to add it as an extra argument. For whatever reason we need it, couldn't it be generated internally on the function?

common/beerocks/btl/include/btl/btl.h
mariomaz: Done.
@arnout: Following this same criteria, shall I store interface name instead of interface index in topology database?
#1440 is already merged so before proceeding, I'd create a hotfix PR for the change

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp

common/beerocks/btl/btl.cpp
mariomaz: According to the 1905.1 standard, we need to transmit some CMDUs (i.e. Topology Discovery message) once on each and every of the available 1905.1 interfaces. Thus why we have modified this method to include which interface to use. In a following commit we call this method in a loop for each bridge interface.

Please see description of the issue #1160 for further details

common/beerocks/btl/btl.cpp
mariomaz: As requested in this [comment](https://github.com/prplfoundation/prplMesh/pull/1420#discussion_r441604127), iface_index has been replaced by iface_name and added as the last parameter with a default value of ""

rmelotte:
@mergifyio rebase

common/beerocks/btl/btl.cpp
arnout: I think @morantr's question was: shouldn't the source address be the interface address rather than the AL-MAC? The answer is: the IEEE1905.1 standard does allow the interface address to be used as the source address, but prefers the AL-MAC address.

PR #1423, opened 2020-06-11
RFC: dynamic backhaul switching HLD

tomereli:


tomereli:
DO NOT MERGE - I should have opened this as draft.

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+# Introduction
+
+Backhaul Switching is the process where the Repeater / Extender switches from one backhaul link to another.
+It can be triggered by any of the following:
+1.	Controller backhaul steering request as part of backhaul optimization process ([1] section 12)
+2.	Wired backhaul link up/down events – wired link down will trigger backhaul switch to the wireless link, whereas wired link up will trigger a backhaul switch to the wired link.
+Future enhancement for selecting the better link (based on phy rate) is out of scope of this document.
+3.	UCC commands – for certification, dev_reset_default should reset any previous onboarding state. This means disabling the backhaul, whichever it is. Later, when a dev_set_config is received, switching to the requested backhaul type will take place.
+
+## Current state
```
arnout, line 11: (nitpick, I guess...)
After you have implemented backhaul switching, the section "Current state" will no longer reflect the current state. Therefore, I think it's better for HLD documentation to consist of two commits: the first one that describes the current state (in the section High Level Design, I suppose) and a second one that changes it to the desired state.
That said, for text documents this can probably be a bit awkward. It does work very well for figures like the state diagram.

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+
+Currently, there is no way to dynamically switch from wired to wireless backhaul and vice versa.
+There is a disabled polling-based mechanism which switches to the wired backhaul if it is up, but it is disabled since it consumed a lot of CPU. 
+Specifically, for certification, only wired backhaul onboarding is supported – e.g. only wired dev_set_config is supported now.
```
arnout, line 15:
```suggestion
Specifically, for certification, only wired backhaul onboarding is supported – i.e. only wired dev_set_config is supported now.
```
"e.g." means "for example".

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
```diff
@@ -0,0 +1,84 @@
...
+WAIT_ENABLE --> ENABLED
+ENABLED --> MASTER_DISCOVERY : wired
+ENABLED --> WIRELESS_FSM : wireless
+state WIRELESS_FSM {
+    [*] --> INIT_HAL
+    INIT_HAL --> WPA_ATTACH
+    WPA_ATTACH --> CONNECT
+    CONNECT --> MASTER_DISCOVERY : connection success
+}
+MASTER_DISCOVERY --> OPERATIONAL
```
arnout, line 14: Note that with unified agent, the master discovery state will (probably) be removed from the backhaul manager.

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+ The Wireless FSM state is out of scope for this design, since backhaul switching takes place in OPERATIONAL state alone - once the FSM is at the OPERATIONAL state, backhaul switching can be triggered in the backhaul manager on demand, and will result with FSM restart, which in turn will result with selecting a new backhaul.
+
+---
+
+## Requirements
+
+* Generic implementation in the backhaul manager which will provide a single entrypoint to perform backhaul switching from multiple clients
+* Supported clients - UCC, Controller, netlink events
+* Prevent loops - use only one backhaul at a time
+* Repeater static
```
arnout, line 37: What does this mean?

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+* Repeater static 
+
+## Assumptions
+
+* Wireless backhaul support for certification exists, including `dev_set_config,name,DUT,program,map,backhaul,RUID`
+* Repeater static configuration
+  * hostapds are configured with start_disabled=1 and static channel at boot
+    * the effect is that the AP will NOT hold a channel and will not put anything on-the-air
+  * Supplicants are configured for WPS_PBC - `wpa_cli wps_pbc multi_ap=1
+` in onboarding mode
```
arnout, line 46:
```suggestion
  * Supplicants are configured for WPS_PBC - `wpa_cli wps_pbc multi_ap=1` in onboarding mode
```

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+  * Supplicants are configured with credentials in operational mode
```
arnout, line 47:
```suggestion
  * OR supplicants are configured with credentials in operational mode
```
I think that's what you mean?

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+
+# High Level Design
+
+Dynamic backhaul switching will be implemented as an API in the backhaul manager - `backhaul_manager_thread::switch_backhaul(const std::string iface_name)` which will move the backhaul manager FSM to the RESTART state, which also causes both radio agents restart.
```
arnout, line 51: I believe (@morantr please confirm) that after unified agent, the backhaul's RESTART state will no longer cause the ap_managers to restart. Which is a good thing IMO. But I don't think it's worth mentioning here either way.

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+* Wireless STA interfaces - `prplmesh.radio1.sta_iface='wlanX'`
+* Wired interface - `prplmesh.config.wired_backhaul='ethX'`
+
+---
+
+Currently, the wired interface is stored in config file - it is preferred to move it to UCI as a preparative commit to the dynamic backhaul switching feature.
+
+---
+
+In order to support automatic backhaul selection at init, another UCI configuration will be added to indicate the preferred backhaul type: `prplmesh.config.preferred_backhaul='wired/24g/5g/auto'`.
```
arnout, line 67: Probably (nitpick) but: why not specify the actual interface as the preferred backhaul (with the addition of 'auto' of course)? I can't really image a scenario where e.g. there are two 5g radios, both with a backhaul STA, and you would prefer 5g but not a specific radio.

Also, the possibility of expressing a preference (other than auto) is something I'm not sure we need for M3.

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+
+A replication of the preferred backhaul type `m_preferred_backhaul` will be stored as a member variable in the backhaul manager thread which will be used for the dynamic backhaul switching, and can be overriden by clients requesting a backhaul switch.
+Note that in order to do a persistent switch, the client needs to also update the persistent UCI DB.
+
+---
+
+Note that `prplmesh.config.preferred_backhaul=auto` is out of scope for this design. It is expected that the client selecting the backhaul switch will set the specific backhaul band, which is the case with `dev_set_config` command and the controller Backhaul Steering Request.
```
arnout, line 74: Ah, right, in that case, the preferred backhaul is needed of course... But I think specifying an interface is better.

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+
+The init sequence described here is assumed to already exist in the platform, and is out of scope of this document.
+
+## Dynamic backhaul switch
+
+On backhaul switch request, the backhaul manager thread will update `m_preferred_backhaul` to the requested backhaul, then move to the `RESTART` state.
+In the `RESTART` state, all potential backhauls are first disabled by either one of the following methods:
+1. Interface disable vi netlink command
+2. Removing the interface from the bridge
+3. Setting up an ebtable rule
```
arnout, line 92: As part of this design, we're going to have to *choose* one of these...

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+
+If wired backhaul is up, the above will be skipped.
+
+The init sequence described here is assumed to already exist in the platform, and is out of scope of this document.
+
+## Dynamic backhaul switch
+
+On backhaul switch request, the backhaul manager thread will update `m_preferred_backhaul` to the requested backhaul, then move to the `RESTART` state.
+In the `RESTART` state, all potential backhauls are first disabled by either one of the following methods:
+1. Interface disable vi netlink command
```
arnout, line 90:
```suggestion
1. Interface disable via netlink or wpa_ctrl command
```
(netlink for wired, wpa_ctrl for wireless)

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
```diff
@@ -0,0 +1,84 @@
...
+box "controller" #LightSlateGray
+participant controller
+end box
+
+box "prplmesh agent" #LightBlue
+participant backhaul_manager
+participant radio_agent
+end box
+
+box "platoform"
```
arnout, line 31:
```suggestion
box "platform"
```

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
```diff
@@ -0,0 +1,84 @@
...
+backhaul_manager -> backhaul_manager : restart
+note left
+m_preferred_backhaul = backhaul
+end note
+backhaul_manager -> supplicant : tear down
+note right
+disable all supplicants
+disable wired interface
+end note
+backhaul_manager -> radio_agent: disconnect
```
arnout, line 58: As I wrote: this will no longer happen in unified agent, I think.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
```diff
@@ -0,0 +1,84 @@
...
+deactivate backhaul_manager
+activate radio_agent
+radio_agent -> radio_agent : restart
+radio_agent -> hostapd : tear down
```
arnout, line 62: That's not good... When we switch backhaul, we *must not* tear down the APs! In fact, in a steering scenario, ideally we should only remove the old backhaul when the new one is fully up and usable.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
```diff
@@ -0,0 +1,84 @@
...
+deactivate radio_agent
+== init ==
+...
+backhaul_manager -> controller : autoconfig search
```
arnout, line 66: Autoconfig search can only start after the backhaul has been established, so this has to go after the enable_interface.

arnout (changes requested):
I think the design has a major flaw in the context of backhaul steering: we should **not** reset everything while switching backhaul. Thus, I think we should decouple the wireless FSM from the overall backhaul manager FSM. I think we should have a wireless FSM for each sta_iface. I also think we should keep everything at least in attached state - there is really no reason to disconnect from wpa_supplicant/hostapd, we just have to put them to sleep if unused. So I think the wireless FSM needs an additional DISABLED state between ATTACH and CONNECT, and that backhaul switching goes through DISABLED -> CONNECT -> OPERATIONAL (for the new one) and OPERATIONAL -> DISABLED (for the old one).

Since this is a pretty big change, I'll ask for a re-review.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
VladyslavTupikin: As I know a unified agent will be implemented in M3. This PR describes features for M2.

arnout: Vlad is right - for this HLD, we don't need and shouldn't take into account unified agent yet.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
tomereli: Switching from one backhaul band to another requires changing the fronthaul channel - which means we need to either manually steer the clients before we do the switch or tear down. I think that for M1/M2 tear down is fine, which is why I chose it. I can update the design to say that in M3 we will do the steering for stations before we steer the backhaul.

tomereli: As an example, consider a dual-band repeater connected with 5G backhaul.
The 2.4G far AP is on channel 6, while the repeater fronthaul 2.4G is on channel 1.
The controller sends a backhaul steering request to the far 2.4G AP, so the clients connected to the repeater's 2.4G front AP needs to switch channel.
The bSTA however needs to do scan, and since there is no background scan, it has to tear down the front AP. This is how I understood this, please correct me if I'm wrong.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
tomereli: This is incorrect, we can start the autoconfig search without an active backhaul, and when the backhaul will be established, the CMDUs will arrive at the controller.
This is IMO a better design since it doesn't require ordering between the WPS and the onboarding operations - in fact, the UCC first sends dev_set_config, and only after does it trigger WPS - so your suggestion will not work in certification.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
tomereli: @morantr please confirm

documentation/flows/agent/dynamic_backhaul_switching.md
tomereli: Mistake, will be removed

documentation/flows/agent/dynamic_backhaul_switching.md
tomereli: You are right, but this should be updated in the implementation PR - otherwise it will be too confusing for the implementor (which will not be me IMO)

documentation/flows/agent/dynamic_backhaul_switching.md
morantr: It depends on what we decide that will happen.
Personally, I think that if the backhaul is disconnected (because of steering request or any other reason) and there is no link backward, there is no reason to allow clients to remain connected since there they will not get service for who knows how much time.
Also, when we need to establish a wireless link, we need to scan the air first, which means that the radio channel will have to change, so it is better if the AP will be down at that moment.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
morantr: https://github.com/prplfoundation/prplMesh/pull/1423#discussion_r443345749

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
+
+# Introduction
+
+Backhaul Switching is the process where the Repeater / Extender switches from one backhaul link to another.
+It can be triggered by any of the following:
+1.	Controller backhaul steering request as part of backhaul optimization process ([1] section 12)
+2.	Wired backhaul link up/down events – wired link down will trigger backhaul switch to the wireless link, whereas wired link up will trigger a backhaul switch to the wired link.
```
morantr, line 7: whereas wired link up will trigger a backhaul switch to the wired link. ->
whereas wired link up will trigger **a wireless backhaul link to** switch to the wired link.

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+* Supported clients - UCC, Controller, netlink events
+* Prevent loops - use only one backhaul at a time
+* Repeater static 
+
+## Assumptions
+
+* Wireless backhaul support for certification exists, including `dev_set_config,name,DUT,program,map,backhaul,RUID`
+* Repeater static configuration
+  * hostapds are configured with start_disabled=1 and static channel at boot
+    * the effect is that the AP will NOT hold a channel and will not put anything on-the-air
```
morantr, line 44: will not put anything on-the-air -> will not transmit

documentation/flows/agent/dynamic_backhaul_switching.md
```diff
@@ -0,0 +1,113 @@
...
+
+---
+
+Note that `prplmesh.config.preferred_backhaul=auto` is out of scope for this design. It is expected that the client selecting the backhaul switch will set the specific backhaul band, which is the case with `dev_set_config` command and the controller Backhaul Steering Request.
+
+---
+
+## Init sequence
+
+On initialization, prplMesh agent will attach to hostapd and all supplicants, then establish backhaul connection via WPS (first boot) or previousely stored UCI configuration on the preferred backhaul based on `prplmesh.config.wireless_backhaul` uci setting.
```
morantr, line 80: Actually, the attachment to the supplicant should occur only if on backhaul enable, we to decide to start a wireless link.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
morantr: I think that "master_discovery" (although it probably should be "controller_dicovery" in HLD document) should not be included on the backhaul state machine. The backhaul manager is (and will be) responsible only for backhaul link establishment, and therefore, the state machine should end after link establishment.
So even if it is planned for M3, I still think it should no be included.

morantr (approved):

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
arnout: OK, for M2 doing teardown is fine.

If possible, it would be nice to make sure that we only tear down the radio on which the backhaul is active. But nice-to-have for M2.

Note that in a wireless-backhaul-only scenario, you would typically configure all APs on the same channel for exactly this reason. Let's say the gateway has a bAP on channel 36. Then repeater1 that is connected to it must have its bSTA on channel 36 as well. If it operates a bAP on the same radio, it will be on channel 36 too. Therefore, repeater2 that connects to repeater1 must have its bSTA on channel 36 as well. And so on. In other words, in practice a channel switch is only required if there's at least one wired link in the network.

For the scan, same thing: if it stays on the same channel, it can do the scan while operating the AP. Well, if the driver supports that, of course...

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
arnout: Let me rephrase: autoconfig search can only reach the controller once backhaul is established :-)

So to be fully correct, it would have to be:
```
backhaul_manager -> backhaul_manager : go to autoconfig search state
... (set up backhaul)
backhaul_manager -> controller : autoconfig search
```
But OK this complicates things, and it's better explained in text anyway.

documentation/flows/agent/dynamic_backhaul_switching.md
arnout: > Personally, I think that if the backhaul is disconnected (because of steering request or any other reason) and there is no link backward, there is no reason to allow clients to remain connected since there they will not get service for who knows how much time.

If we do it well, the time that it remains disconnected can be small. Also, if we bring down the AP, most clients will try to connect to another AP, which is usually not what we want.

> Also, when we need to establish a wireless link, we need to scan the air first, which means that the radio channel will have to change, so it is better if the AP will be down at that moment.

Do we actually need to scan? Typically the APs don't change very much, so we can cache scan results for a very long time. Also, I think in many situations all the bAPs will be on the same channel (https://github.com/prplfoundation/prplMesh/pull/1423#discussion_r444854037) so scan/probe can easily be done while operating the AP. Also, at least some drivers do support off-channel scanning while operating an AP (at least I think so). And don't forget the situation where there's a dedicated radio for backhaul - then backhaul steering without interruption is definitely possible. Also if we switch radios. And finally, because the steering request contains the target channel and we have the BSSID, it's not even necessary to scan, we can directly send a probe request to that bssid (though I don't think this is supported by wpa_supplicant at the moment).

documentation/flows/agent/dynamic_backhaul_switching.md
tomereli: Correct.

documentation/flows/agent/dynamic_backhaul_switching.md
tomereli: Eventually, we don't need either - this is done as part of the supplicant platform initialization which uses the `-b <bridge>` option.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
tomereli: Agreed.

documentation/images/plantuml/agent/state_machines/backhaul_manager_fsm.puml
tomereli: eventually, it is as you said in the first place (probably since you implemented it :) ):
```
backhaul_manager -> backhaul_manager : set up backhaul
ucc -> backhaul_manager : start_wps_pbc
... (establish backhaul connection vi wps_pbc)
backhaul_manager -> controller : autoconfig search
```

PR #1419, opened 2020-06-11
[documentation]  Add UML diaram for Backhaul STA steering

VladyslavTupikin:
According to the test plan and source code
create the UML diagram which describes
data flow and methods for Backhaul STA steering.

PR #1424, opened 2020-06-11
[agent]: Turn on components for Backhaul STA Steering

### LINKED ISSUE #408: [TASK] Backhaul STA steering
## Description

The agent should perform steering of the backhaul STA interface based on the controller's backhaul STA steering request.

## Exit criteria

- BWL implementation of backhaul STA steering on nl80211.
- BWL implementation of backhaul STA steering on dummy.
- Agent does backhaul steering according to policy message.
- Test flow for the above.

VladyslavTupikin:
According to the UML diagram from PR #1419 need to add Bachaul STA Steering TLV's to the builds, update method parseNextTlv in CmduMessageRx - add cases with new TLV's, add stubs in the HAL for associate method.

### DIFF
--- a/framework/tlvf/AutoGenerated/include/tlvf/wfa_map/tlvBackhaulSteeringRequest.h
+++ b/framework/tlvf/AutoGenerated/include/tlvf/wfa_map/tlvBackhaulSteeringRequest.h
@@ -1,0 +1,58 @@
+///////////////////////////////////////
+// AUTO GENERATED FILE - DO NOT EDIT //
+///////////////////////////////////////
+
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2016-2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _TLVF_WFA_MAP_TLVBACKHAULSTEERINGREQUEST_H_
+#define _TLVF_WFA_MAP_TLVBACKHAULSTEERINGREQUEST_H_
+
+#include <cstddef>
+#include <stdint.h>
+#include <tlvf/swap.h>
+#include <string.h>
+#include <memory>
+#include <tlvf/BaseClass.h>
+#include <tlvf/ClassList.h>
+#include "tlvf/wfa_map/eTlvTypeMap.h"
+#include "tlvf/common/sMacAddr.h"
+
+namespace wfa_map {
+
+
+class tlvBackhaulSteeringRequest : public BaseClass
+{
+    public:
+        tlvBackhaulSteeringRequest(uint8_t* buff, size_t buff_len, bool parse = false);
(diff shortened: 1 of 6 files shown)


VladyslavTupikin:
PR should be merged after https://github.com/prplfoundation/prplMesh/pull/1419

VladyslavTupikin:
PR #1419 merged. This PR can be reviewed.

common/beerocks/bwl/include/bwl/sta_wlan_hal.h
```diff
@@ -48,6 +48,8 @@ class sta_wlan_hal : public virtual base_wlan_hal {
                          bool mem_only_psk, const std::string &bssid, uint8_t channel,
                          bool hidden_ssid) = 0;
 
+    virtual bool associate(const std::string &bssid, uint8_t channel) = 0;
```
arnout: Please don't use strings for mac addresses in new functions.
```suggestion
    virtual bool associate(const sMacAddr &bssid, uint8_t channel) = 0;
```
Yes, it requires an additional string-to-mac conversion in the caller, and then a mac-to-string conversion to send it to wpa_supplicant, but this is the only way we can ever hope to stamp out the evil use of strings for mac addresses.

common/beerocks/bwl/dummy/sta_wlan_hal_dummy.cpp
```diff
@@ -41,6 +41,8 @@ bool sta_wlan_hal_dummy::connect(const std::string &ssid, const std::string &pas
     return true;
 }
 
+bool sta_wlan_hal_dummy::associate(const std::string &bssid, uint8_t channel) { return true; }
```
arnout: I think it's useful to already add a log message in each of the implementations, even if it's currently unimplemented.

arnout (approved):
I actually meant to also add the logic that handles the steering request and makes the bwl call, but in fact it makes sense to merge the current state.

common/beerocks/bwl/include/bwl/sta_wlan_hal.h
VladyslavTupikin: fixed

common/beerocks/bwl/dummy/sta_wlan_hal_dummy.cpp
VladyslavTupikin: added debug prints

abelog (approved):
LGTM

common/beerocks/bwl/include/bwl/sta_wlan_hal.h
```diff
@@ -48,6 +48,8 @@ class sta_wlan_hal : public virtual base_wlan_hal {
                          bool mem_only_psk, const std::string &bssid, uint8_t channel,
                          bool hidden_ssid) = 0;
 
+    virtual bool associate(const sMacAddr &bssid, uint8_t channel) = 0;
```
arnout: I just noticed now: the roam() function, three lines below, is in fact exactly what we need, so this commit is not needed at all.

I've removed the Ready for Merge label so you can remove this commit.

VladyslavTupikin: removed useless commit

PR #1429, opened 2020-06-11
PPM-83 unified agent infrastructure

morantr:
This PR features new database and tasks infrastructure to the future unified agent.
#1082 - Phase 2

Discussion:
The featured database on this PR is planned to be used among all the main agent threads (agent, backhaul, platform).
The database needs to be thread-safe, therefore I have written an automatic lock and release mechanism to the database. It works like this:
```cpp
// When getting the db instance the db is automatically being locked. The release occurs when 'db' is destructed.
auto db = AgentDB::get();
db->foo = 42;
```
A problem pops out if someone will try to get a reference to a database member:
```cpp
auto &my_foo = AgentDB::get()->foo;
my_foo = 5;
```
After getting the reference the lock is released since the instance of the db is destroyed, and changing the value is unsafe.
There are 3 options that we could choose.

This is the table that sum it up. (Detailed explanation is below the table).

  |   | Convinient | Preformance | Code Size | Readbility | Adding/Editing   Code | Memory   Consumption | Safty | Implementation   Time
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Multhread   Shared-DB | Direct Access | Best | Best- | Best | Best | Best | Best | Data Validation   Issue     Hard to detect on review | Low
Multhread   Shared-DB | Setters   Getters | Medium | Medium+ | Medium | Medium | Best- | Best | Data   Validation Issue     Easier to detect on review | Medium
Single Thread DB |   | Best | Best- | Worst | Worst | Worst | Worst | Safe | High

Please vote:
❤️ - Option 1
🚀 - Option 2
👀 - Option 3

1. I have added to the class 'how to use documentation' explanation about it, asking from the user not to use it that way.
It is important to understand that this is a general C/C++ problem. This language allows a developer to do unsafe memory operation, so since we are careful to not do it generally, then we need to be careful here as well since it is the same thing.

   ***Convenient: BEST*** 
      - On a single call, it is possible to get by copy/reference and set members directly.
      <br />

   ***Performance: BEST-***
      - Since accessing db members it is required to allocate db subclass on the stack which locks the mutex and unlock it on destruction. The lock and unlock could be used once for several member access, and not for each one.
      - Since getting members by reference is possible.
      - Since the database is used by a few threads, then sending data between threads is no longer needed.
      <br />

   ***Code size: BEST***
      - The db contains only members and therefore it is slim.
      - Since it is shared between threads, there is no need to save a copy of memory on each thread and the tlvf code will decrease too.
      <br />

   ***Readability: BEST***
      - Since the db contain only members, its class is not gigantic and therefore more readable.
      - Since it is shared between threads on a single location, no need to track messages to see the whole flow and lifetime and copies of some configuration.
      <br />

   ***Adding/Editing code: BEST***
      - Since adding/changing only in one place - db class.
      <br />

   ***Memory consumption: BEST***
      - Since it is shared between threads, there is no need to save a copy of memory on each thread.
      <br />

   ***Safety: C/C++ Safty - Harder to detect on code review***
      - The mechanism has a general C/C++ safety issue which allows touching shared memory without being protected by a mutex if using the db API incorrectly. It is important to understand that it cannot cause to a crash, only to invalid data set/get.
      - Since it not using setters and getter the wrong use of the DB can be anywhere on the code, and potentially it will be harder to notice it on code review.
<br />

2. Using getters and setters with restriction of getting values only by copy. The problem still remains with setters and getters since the getter could be written with returning by reference, but the places where the mistake could be done is only on the getters/setters function on the db, instead of all over the code.
The disadvantage of it is that it is much more convenient to no use setters/getters and just use directly the member is needed, also it will make the database class more than 3 times longer.
Each getter will have to strict the return value to be a copy.

   ***Convenient: MEDIUM*** 
      - Using getters and setters is less convenient than accessing class members directly.
      <br />

   ***Performance: MEDIUM+***
      - Since accessing db members it is required to allocate db subclass on the stack which locks the mutex and unlock it on destruction. The lock and unlock could be used once for several member access, and not for each one.
      - Since getting members by reference is not possible.
      - Since the database is used by a few threads, then sending data between threads is no longer needed.
      <br />

   ***Code size: MEDIUM***
      - The db contains members and getter and setter for each one which increases code size.
      - Since it is shared between threads, there is no need to save a copy of memory on each thread.
      <br />

   ***Readability: MEDIUM***
      - Since the db contains setter and getter for each member it will be gigantic (like the controller DB) and therefore less readable.
      - Since it is shared between threads on a single location, no need to track messages to see the whole flow and lifetime and copies of some configuration.
      <br />

   ***Adding/Editing code: BEST-***
      - Since adding/changing only in one place - db class, but need to add setter and getter.
      <br />

   ***Memory consumption: BEST***
      - Since it is shared between threads, there is no need to save a copy of memory on each thread.
      <br />

   ***Safety: C/C++ Safty - Easier to detect on code review***
      - The mechanism has a general C/C++ safety issue which allows touching shared memory without being protected by a mutex if using the db API incorrectly. It is important to understand that it cannot cause a crash, only to invalid data set/get.
      - Since using setters and getter the wrong use of the DB can only on the DB class, so it will be written once, and so there is a higher chance to detect it on code review.
      <br />

3. Not to use shared memory and creating a thread-local copy of memory which is sent by messages.

   ***Convenient: BEST*** 
      - Since each data has a local copy of the access to it easiest.
      <br />

   ***Performance: BEST-*** 
      - Since each configuration has a local copy the performance accessing it is the best.
      - Since each configuration needs to be sent on a CMDU.
      <br />

   ***Code size: WORST***
      - Since the code contains a copy of memory on each thread.
      - Since there is a code containing the data on messages, code that copy the data to a message to send it, and code that handles the message and saves it on a local copy.
      <br />

   ***Readability: WORST***
      - Since there is a code containing the data on messages, code that copy the data to a message to send it, and code that handles the message and saves it on a local copy.
      <br />

   ***Adding/Editing code: WORST***
      - Add/Change the code on each thread the copy of data exists, and also on tlvf.
      <br />

   ***Memory consumption: WORST***
      - Since it there is a local copy on each thread.
      <br />

   ***Safety: Safe***
      - No safety issue.
      <br />

### DIFF
--- a/agent/src/beerocks/slave/agent_db.h
+++ b/agent/src/beerocks/slave/agent_db.h
@@ -1,0 +1,79 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2019-2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include <iostream>
+#include <memory>
+#include <mutex>
+
+namespace beerocks {
+
+/**
+ * @brief AgentDB is a class used by all main Agent threads on the Agent process to store the Agent
+ * common data.
+ * 
+ * The class is implemented using singleton design pattern.
+ * It is thread safe and being locked and released automatically.
+ * 
+ * How to use:
+ * @code
+ * auto db = AgentDB::get();            // Get DB and automatically lock it. The lock will be
+ *                                      // Released when 'db' will be destructed.
+ * 
+ * db->foo = 42;                        // Change database member 'foo' to value 42.
+ * db->bar = "prpl"                     // Set another database member.
+ * auto &my_foo = db->foo;              // Get a refernce to a member.
+ * 
+ * AgentDB::get()->foo = 44;            // Set database member directly.
+ * 
+ * 
+ * Unsafe operation which should never be done:
+ * auto my_foo = AgentDB::get()->foo;   // Unsafe! Don't do it! The database might be unlocked,
+ *                                      // if used on ranged loop. So better not use it.
+ * 
+ * auto &foo = AgentDB::get()->foo;     // Unsafe! Don't do it! The database will be unlocked,
+ *                                      // but the reference will remain.
+ * 
+ * std::string &get_foo() {             // Unsafe! Don't do it! Returning refernce to the database
+ *     auto db = AgentDB::get();        // on a wrapper function is unsafe because the database will
+ *     return db->foo;                  // be unlocked when the function ends, and the caller will
+ * }                                    // hold a refernce to it.                                 // hold a refernce to it.
+ * @endcode
+ */
+class AgentDB {
+public:
+    class SafeDB {
+    public:
+        SafeDB(AgentDB &db) : m_db(db) { m_db.db_lock(); }
+        ~SafeDB() { m_db.db_unlock(); }
+        AgentDB *operator->() { return &m_db; }
+
+    private:
+        AgentDB &m_db;
+    };
+    static SafeDB get()
+    {
+        // Guaranteed to be destroyed.
+        // Instantiated on first use.
+        static AgentDB instance;
+        return SafeDB(instance);
+    }
+    AgentDB(const AgentDB &) = delete;
+    void operator=(const AgentDB &) = delete;
+
+private:
+    // Private constructor so that no objects can be created.
+    AgentDB() = default;
+    std::recursive_mutex m_db_mutex;
+    void db_lock() { m_db_mutex.lock(); }
+    void db_unlock() { m_db_mutex.unlock(); }
+
+public:
+    /* Put here database members used by the Agent modules */
+};
+
+} // namespace beerocks


kantera800:
I prefer option 2, 
but only if there is an additional task to remove all parameter passing via CMDUs and use 
shared memory set/get dedicated APIs.

morantr:
> I prefer option 2,
> but only if there is an additional task to remove all parameter passing via CMDUs and use
> shared memory set/get dedicated APIs.

There is a cleanup phase on the Epic.
Just be aware that option 2 will have the highest development time.

VladyslavTupikin:
I prefer option 1 because it gives a lot of advantages. Safety issues can be solved by adding a validation method (methods) for the each data structure.

agent/src/beerocks/slave/task_pool.h
```diff
@@ -0,0 +1,39 @@
...
+public:
+    task_pool() {}
+    ~task_pool() {}
+
+    bool add_task(std::shared_ptr<task> new_task);
+    bool is_task_running(int id);
+    void kill_task(int id);
+    void push_event(int task_id, int event_type, void *obj = nullptr);
+    void response_received(std::string mac,
+                           std::shared_ptr<beerocks::beerocks_header> beerocks_header);
```
arnout: The commit message says "since the bcl depends on btlvf", but I guess the problem is actually that btlvf depends on bcl and this one depends on btlvf, so it can't be put in bcl.

However, I think this dependency on btlvf should be removed from the task implementation. Indeed, I expect that in the unified aent, some of the responses received will not be beerocks vendor specific messages, but rather standard wfa_map messages. So this dependency on btlvf should be removed, eventually.

That means refactoring the task code, and such refactoring is extremely annoying if we have two copies of the same code. Therefore, I would propose to instead, as a transitory measure, move the task to btlvf. I know it doesn't belong there at all, but you can add a big TODO saying so.

On the other hand...
I think that before we start using this task infrastructure in the agent, we should invest in refactoring it in other ways than only the response_received (to make it easier to work with, mockable and testable, safer, ...). So if it is shared with the controller, refactoring requires changing the controller code as well, which is risky since we have no tests for the controller tasks... So in that sense, there's something to be said for "abandon the controller, focus on getting the agent right and just copy it so we can change whatever we like".

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+#include <mutex>
+
+using namespace beerocks_message;
+using namespace son;
+using namespace beerocks;
+
+/**
+ * @brief AgentDB is a class used by all main Agent threads on the Agent process to store the Agent
+ * common data.
+ * The class is implemented using singleton design pattern.
```
arnout, line 19: @RanRegev and @mariomaz will probably shoot this down right away :-)

I'm more inclined to accept it since it's supposed to be passive data, so no need for mocking it or replacing it with anything else.

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+ * It is thread safe and being locked and released automatically.
+ * 
+ * How to use:
+ * 
+ * auto db = AgentDB::get();            // Get DB and automatically lock it. The lock will be
+ *                                      // Release when 'db' will be destructed.
+ * 
+ * db->foo = 42;                        // Change database member 'foo' to value 42
+ * db->bar = std::string("prpl")        // Set another database member
```
arnout: (nitpick) std::string should be redundant here.

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+        // Instantiated on first use.
+        static AgentDB instance;
+        return SafeDB(instance);
+    }
+    AgentDB(AgentDB const &) = delete;
+    void operator=(AgentDB const &) = delete;
+
+private:
+    // Private constructor so that no objects can be created.
+    AgentDB() {}
```
arnout:
```suggestion
    AgentDB() = default;
```
(is a bit more explicit)

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+class AgentDB {
+public:
+    class SafeDB {
+    public:
+        SafeDB(AgentDB &db) : m_db(db) { m_db.db_lock(); }
+        ~SafeDB() { m_db.db_unlock(); }
+        AgentDB *operator->() { return &m_db; }
+
+    private:
+        AgentDB &m_db;
```
arnout, line 56: (nitpick) I like an explicit `SafeDB() = delete;` (redundant because there is a reference member).

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+ * auto my_foo = AgentDB::get()->foo;   // Get a copy of database member.
+ * 
+ * auto &foo = AgentDB::get()->foo;     // Unsafe! Don't do it! The database will be unlocked,
+ *                                      // but the reference will remain.
+ */
+class AgentDB {
+public:
+    class SafeDB {
+    public:
+        SafeDB(AgentDB &db) : m_db(db) { m_db.db_lock(); }
```
arnout, line 51: I think we could use a move constructor as well, to avoid unnecessary unlock-lock sequences.

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+ *                                      // Release when 'db' will be destructed.
+ * 
+ * db->foo = 42;                        // Change database member 'foo' to value 42
+ * db->bar = std::string("prpl")        // Set another database member
+ * 
+ * AgentDB::get()->foo = 44;            // Set database member directly.
+ * auto my_foo = AgentDB::get()->foo;   // Get a copy of database member.
+ * 
+ * auto &foo = AgentDB::get()->foo;     // Unsafe! Don't do it! The database will be unlocked,
+ *                                      // but the reference will remain.
```
arnout: This is getting me worried a little...

A typical use case would be something like this:
```
for (foo : AgentDB::get()->foo_list) ...
```
This doesn't work as expected, because the lock is only held during the evaluation of the loop expression, not during the body of the loop. You *have* to write it like this:
```
auto db = AgentDB::get();
for (foo : db->foo_list) ...
```

Therefore, I think it's better to disallow any other use than `auto db = AgentDB::get();`. That makes it *much* easier to check that the code is correct.

(Note: lifetimes of temporaries are extremely difficult to reason about, so my analysis may be incorrect. But, since they are extremely difficult to reason about, there's no way to be sure if there are no situations that the lifetime is too short. So better safe than sorry...)

Therefore, I would write:
```suggestion
 * AgentDB::get()->foo = 44;            // Set database member directly.
 * auto my_foo = AgentDB::get()->foo;   // Unsafe! Don't do it! 
 *                                      // This specific example is OK, but it's hard to know when the database will be unlocked.
 * auto &foo = AgentDB::get()->foo;     // Unsafe! Don't do it! The database will be unlocked,
 *                                      // but the reference will remain.
```

Another tricky situation is where you get a nested reference from the database - that's actually the same as a reference, but less obvious. For example, let's say the database has a member `std::shared_ptr<TopologyTask> topology_task`. Then `auto task = db->topology_task;` will get you a reference (implied through the smart pointer). So we always have to be careful about when the db is going to be unlocked.

Bottom line: the SafeDB is a useful tool, but it's not sufficient by far to be safe.

We could make it more reliable by saying that the database must not contain any references to begin with. So that would already exclude task_pool, which has references to tasks... It means that references should be managed through identifiers, and every time you want to dereference something, you have to go through the database. But even that is not a silver bullet, because when you start passing those identifiers around, you can end up with unlocking the database in-between and consistency is no longer guaranteed (e.g. you have some function that takes a station identifier out of the database and returns it; if the caller hasn't locked the database already, the database is unlocked when the function returns, and by the time the caller locks the database again to look up the reference, it is already gone).

Bottom line (really, this time): I prefer not to have to deal with multithreaded code :-)

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+
+public:
+    /* Put here database members used by the Agent modules */
+
+    /* Tasks */
+    task_pool tasks;
+
+    struct sTasksId {
+        int topology_task = -1;
+    } tasks_id;
```
arnout: So, this would be an example of using identifiers rather than references. Like this, it's pretty useless, you would need something to get the task object based on the task ID as well.

Since it's not used yet, I would just remove the tasks and tasks_id for the moment - it makes much more sense to add them once you start using them.

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+#include "task_pool.h"
+
+#include <iostream>
+#include <memory>
+#include <mutex>
+
+using namespace beerocks_message;
+using namespace son;
```
arnout: (nitpick) personally, I don't like `using namespace` statements much. Especially in a header file...

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,76 @@
...
+using namespace beerocks;
```
arnout:
```suggestion
namespace beerocks {
```
*Everything* should be in the beerocks namespace.

arnout (commented):
Big feedback in my review. I'm not going to block it, but I don't feel quite ready to approve it either. That said, if you want to go forward with the current implementation, I can approve it. The only changes I really want are:
- remove the members of the database;
- put it in beerocks namespace.

Ideally there should also be a unit test, at least of the database. However, I'm not sure how I would write a multithreaded unit test... You probably should make it mockable and mock away the mutex to make it testable (that way you can emulate and check various race conditions). A unit test would also be a nice way to document the limitations of the automatic locking.

agent/src/beerocks/slave/agent_db.h
RanRegev: There are few problems with singletons  (yes, I prefer to shoot it) -
* mocking, there is no way to replace it.
* it creates dependencies in such a way that are both hidden and unflexible:
```cpp
void ClassMine::foo(){
// hidden dependency, in the header of ClassMine it is not mentioned at all.
auto& db = AgendDB::instance();
}
```
Better approach, if for example, ClassMine requires a Database, _state it clearly_ -
```cpp
class ClassMine{
public:
ClassMine(AgentDB&);
};
```
if only `foo()` needs the database:
```cpp
ClassMine::foo(AgentDB&) { /*...*/ }
```
I am not _against_ a single instance of the database in the program.
It is OK to have a single object, but let the owner the responsibility of making it one. don't let the user decide it. in this case `foo()` is the user, not the owner of the db.

agent/src/beerocks/slave/task_pool.h
morantr: Actually I have decided to remove the task infrastructure from the DB before you commented it, because I realized it does not work well with the needs of the UA.
So sorry you put so much effort into this comment, I removed the task from the DB.

agent/src/beerocks/slave/agent_db.h
morantr: fixed.

agent/src/beerocks/slave/agent_db.h
morantr: ok. I removed the others for now. I think I added it because of the task infra which I removed, so...

agent/src/beerocks/slave/agent_db.h
morantr: removed

agent/src/beerocks/slave/agent_db.h
morantr: ok

agent/src/beerocks/slave/agent_db.h
morantr: I'm don't understand why would we ever use std::move() on it? We can pass it as a reference to a sub-function and when it finishes the object is destructed.
Can you please give an example?

agent/src/beerocks/slave/agent_db.h
morantr: I added/changed the "how to use".
I think we should give it a shot.

agent/src/beerocks/slave/agent_db.h
arnout: It would be to allow inclusion in a super-structure. However, it's not a use case that is needed at the moment, and if we need it, we can still add the move constructor. So my bad.

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,78 @@
...
+ * How to use:
+ * 
+ * auto db = AgentDB::get();            // Get DB and automatically lock it. The lock will be
+ *                                      // Released when 'db' will be destructed.
+ * 
+ * db->foo = 42;                        // Change database member 'foo' to value 42.
+ * db->bar = "prpl"                     // Set another database member.
+ * auto &my_foo = db->foo;              // Get a refernce to a member.
+ * 
+ * AgentDB::get()->foo = 44;            // Set database member directly.
```
arnout, line 31:
```suggestion
 * @code
 * auto db = AgentDB::get();            // Get DB and automatically lock it. The lock will be
 *                                      // Released when 'db' will be destructed.
 * 
 * db->foo = 42;                        // Change database member 'foo' to value 42.
 * db->bar = "prpl"                     // Set another database member.
 * auto &my_foo = db->foo;              // Get a refernce to a member.
 * 
 * AgentDB::get()->foo = 44;            // Set database member directly.
 * @endcode
```
(same with the other examples).

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,78 @@
...
+#include <iostream>
+#include <memory>
+#include <mutex>
+
+namespace beerocks {
+
+/**
+ * @brief AgentDB is a class used by all main Agent threads on the Agent process to store the Agent
+ * common data.
+ * The class is implemented using singleton design pattern.
```
arnout:
```suggestion
 * @brief AgentDB is a class used by all main Agent threads on the Agent process to store the Agent
 * common data.
 *
 * The class is implemented using singleton design pattern.
```
(otherwise the entire paragraph will be part of the `@brief`.

agent/src/beerocks/slave/agent_db.h
```diff
@@ -0,0 +1,78 @@
...
+    };
+    static SafeDB get()
+    {
+        // Guaranteed to be destroyed.
+        // Instantiated on first use.
+        static AgentDB instance;
+        return SafeDB(instance);
+    }
+    AgentDB(AgentDB const &) = delete;
+    void operator=(AgentDB const &) = delete;
```
arnout: It doesn't make a difference, but we usually write
```suggestion
    AgentDB(const AgentDB &) = delete;
    void operator=(const AgentDB &) = delete;
```

agent/src/beerocks/slave/agent_db.h
arnout: The thing is, in this case it is rather difficult to make it anything else than a singleton and still have the "safety" aspect. You'd need to add another flyweight class, e.g. UnsafeDB, that acts as a unique reference and that can be constructed through a factory method.

So I propose to keep it as singleton and add that reference class when there is a real need for it.

arnout (approved):

agent/src/beerocks/slave/agent_db.h
mariomaz: Singleton is an anti-pattern, you should avoid it

kantera800 (approved):

PR #1437, opened 2020-06-15
deploy_firmware.py: try to kill the prplMesh init script first

### LINKED ISSUE #1436: [BUG] CI: firmware upgrade does not work if the current firmware doesn't have working wireless interfaces
See https://gitlab.com/prpl-foundation/prplMesh/-/jobs/594448066

When we deploy a firmware (even on a side branch) that doesn't have working wireless interfaces, the device is stuck with the same firmware unless we take action manually.

The reason is that the /etc/init.d/S99prplmesh script waits for the wireless interfaces forever and prevents the reboot from happening.

rmelotte:
When the rax40 has a firmware with non-working wireless interfaces,
the S99prplmesh script prevents it from rebooting.

Try to kill it first, before doing a reboot.

Fixes https://github.com/prplfoundation/prplMesh/issues/1436.

### DIFF
--- a/tools/deploy_firmware.py
+++ b/tools/deploy_firmware.py
@@ -174,8 +174,12 @@ class NetgearRax40(PrplwrtDevice):
             # make the shell prompt appear:
             self.set_prompt(shell)
             shell.expect(self.serial_prompt)
+            # kill any instance of the init script, if the current
+            # firmware doesn't have working wireless interfaces it
+            # will prevent it from rebooting:
+            shell.sendline("pgrep -f 'S99prplmesh boot' | xargs kill")
             # reboot:
-            shell.sendline("reboot")
+            shell.sendline("reboot -f")
             shell.expect(["Hit any key to stop autoboot:",
                           pexpect.EOF, pexpect.TIMEOUT], timeout=120)
             # stop autoboot:


tools/deploy_firmware.py
```diff
@@ -174,6 +174,10 @@ def upgrade_uboot(self):
             # make the shell prompt appear:
             self.set_prompt(shell)
             shell.expect(self.serial_prompt)
+            # kill any instance of the init script, if the current
+            # firmware doesn't have working wireless interfaces it
+            # will prevent it from rebooting:
+            shell.sendline("pgrep -f 'S99prplmesh boot' | xargs kill")
             # reboot:
             shell.sendline("reboot")
```
arnout: It would be better to also do `reboot -f` here - which would have avoided the problem to begin with. There's really not much reason to go through the init scripts.

arnout (approved):

tools/deploy_firmware.py
tomereli: Didn't know about the -f option

tomereli (approved):

tools/deploy_firmware.py
rmelotte: I didn't know busybox's reboot had that. 
Thanks for the suggestion! 
I'll do that.

PR #1439, opened 2020-06-15
Port ApConfigRenew test of test_flow.py to boardfarm

abelog:
## Description
This PR brings port of `ApConfigRenew` test of `test_flows.py` to boardfarm and include it to the general test suite.

## Changes

- Keep MAC of radios in dummy docker and real devices as part of relevant classes. 
- Keep `ucc_socket` which provides methods to communicate with devices over IEEE1905.1 for dummy docker devices and real devices as part of relevant classes.

## Notes

Log of SSID setting was changed comparing to the original `ApConfigRenew` of `test_flows.py` as dummy docker implementation and RAX40 board has different logs about incoming configuration event.
`time.sleep(30)` is required to pass test on RAX40. This should be replaced by wait of unique log in future.

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
@@ -1,0 +1,67 @@
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+import time
+
+from .prplmesh_base_test import PrplMeshBaseTest
+from capi import tlv
+
+
+class ApConfigRenew(PrplMeshBaseTest):
+    """Check initial configuration on device."""
+
+    def runTest(self):
+        # Locate agents and controller
+        for dev in self.dev:
+            if dev.controller_entity:
+                controller = dev.controller_entity
+            if dev.agent_entity:
+                agent = dev.agent_entity
+
+        dev.wired_sniffer.start(self.__class__.__name__ + "-" + dev.name)
+        # Regression test: MAC address should be case insensitive
+        mac_repeater1_upper = agent.mac.upper()
+        controller.ucc_socket.cmd_reply("DEV_RESET_DEFAULT")
+        controller.ucc_socket.cmd_reply(
+            "DEV_SET_CONFIG,"
+            "bss_info1,{} 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,"
+            "bss_info2,{} 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0"
+            .format(mac_repeater1_upper, agent.mac))
+        controller.ucc_socket.dev_send_1905(agent.mac, 0x000A,
+                                            tlv(0x01, 0x0006, "{" + controller.mac + "}"),
+                                            tlv(0x0F, 0x0001, "{0x00}"),
+                                            tlv(0x10, 0x0001, "{0x00}"))
+
+        time.sleep(30)
+        self.check_log(agent.radios[0],
+                       r"ssid: Multi-AP-24G-1 .*"
+                       r"fronthaul: true backhaul: false",
+                       timeout=60)
+        self.check_log(agent.radios[0],
+                       r"ssid: Multi-AP-24G-2 .*"
+                       r"fronthaul: false backhaul: true",
+                       timeout=60)
+        self.check_log(agent.radios[1],
+                       r"tear down radio",
+                       timeout=60)
+        bssid1 = agent.ucc_socket.dev_get_parameter('macaddr',
+                                                    ruid='0x' +
+                                                    agent.radios[0].mac.replace(':', ''),
+                                                    ssid='Multi-AP-24G-1')
+        if not bssid1:
+            self.fail("repeater1 didn't configure Multi-AP-24G-1")
+        # simulate wps onboarding to the backhaul vap
+        agent.ucc_socket.start_wps_registration("24G")
+        self.check_log(agent.radios[0], r"Start WPS PBC", timeout=60)
+
+    @classmethod
+    def teardown_class(cls):
+        """Teardown method, optional for boardfarm tests."""
+        test = cls.test_obj
+        for dev in test.dev:
+            if dev.agent_entity:
+                print("Sniffer - stop")
+                dev.wired_sniffer.stop()
--- a/tests/environment.py
+++ b/tests/environment.py
@@ -458,7 +458,7 @@ class ALEntityPrplWrt(ALEntity):
         # Multiply timeout by 100, as test sets it in float.
         return _device_wait_for_log(self.device,
                                     "{}/beerocks_{}.log".format(self.log_folder, program),
-                                    regex, start_line, timeout*100)
+                                    regex, start_line, timeout)
 
     def prprlmesh_status_check(self):
         return self.device.prprlmesh_status_check()
@@ -483,7 +483,7 @@ class RadioHostapd(Radio):
         ''' Poll the Radio's logfile until it match regular expression '''
         # Multiply timeout by 100, as test sets it in float.
         return _device_wait_for_log(self.agent.device, "{}/beerocks_agent_{}.log".format(
-            self.log_folder, self.iface_name), regex, timeout*100, start_line)
+            self.log_folder, self.iface_name), regex, timeout, start_line)
 
 
 class VirtualAPHostapd(VirtualAP):
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/testsuites.cfg
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/testsuites.cfg
@@ -5,3 +5,4 @@
 
 [test_flows]
 InitialApConfig
+ApConfigRenew


tests/environment.py
```diff
@@ -458,7 +459,7 @@ def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
         # Multiply timeout by 100, as test sets it in float.
```
VladyslavTupikin, line 458: Multiplication was removed but commentary still there.

tests/environment.py
```diff
@@ -471,19 +472,19 @@ def __init__(self, agent: ALEntityPrplWrt, iface_name: str):
...
+        super().__init__(agent, self.mac)
 
-        VirtualAPHostapd(self, mac)
-        VirtualAPHostapd(self, mac)
+        VirtualAPHostapd(self, self.mac)
+        VirtualAPHostapd(self, self.mac)
 
     def wait_for_log(self, regex: str, start_line: int, timeout: float):
         ''' Poll the Radio's logfile until it match regular expression '''
         # Multiply timeout by 100, as test sets it in float.
```
VladyslavTupikin, line 484: Multiplication was removed but the commentary still there...

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
```diff
@@ -0,0 +1,66 @@
...
+        controller.ucc_socket.dev_send_1905(agent.mac, 0x000A,
+                                            tlv(0x01, 0x0006, "{" + controller.mac + "}"),
+                                            tlv(0x0F, 0x0001, "{0x00}"),
+                                            tlv(0x10, 0x0001, "{0x00}"))
+
+        time.sleep(30)
+        self.check_log(agent.radios[0],
+                       r"ssid: Multi-AP-24G-1 .*"
+                       r"fronthaul: true backhaul: false",
+                       timeout=60)
```
VladyslavTupikin, line 42: I think we need to add commentary with an explanation of what 60 does mean, like time in seconds.

VladyslavTupikin (approved):
Approve, marked changes small and easy to fix.

tests/environment.py
```diff
@@ -279,10 +280,10 @@ def __init__(self, name: str, device: None = None, is_controller: bool = False):
             device_ip = re.search(r'inet (?P<ip>[0-9.]+)',
                                   device_ip_output.decode('utf-8')).group('ip')
 
-        ucc_socket = UCCSocket(device_ip, ucc_port)
-        mac = ucc_socket.dev_get_parameter('ALid')
+        self.ucc_socket = UCCSocket(device_ip, ucc_port)
+        self.mac = self.ucc_socket.dev_get_parameter('ALid')
 
-        super().__init__(mac, ucc_socket, installdir, is_controller)
+        super().__init__(self.mac, self.ucc_socket, installdir, is_controller)
```
arnout: This makes no sense... `super().__init__(...)` starts with the following code:
```
        self.mac = mac
        self.ucc_socket = ucc_socket
```

So why would you need to set them already in the child class constructor?

It doesn't hurt of course, but I don't like it when something is assigned twice, it makes it hard to understand where it is coming from.

tests/environment.py
```diff
@@ -241,6 +241,7 @@ def _device_wait_for_log(device: None, log_path: str, regex: str,
         matched_line = int(device.match.group('line_number')) + start_line
         return (True, matched_line, match)
     else:
+        err("Can't find {} in log of ".format(regex, device.name))
```
arnout: This change is unrelated -> separate commit.

tests/environment.py
```diff
@@ -458,7 +459,7 @@ def wait_for_log(self, regex: str, start_line: int, timeout: float) -> bool:
         # Multiply timeout by 100, as test sets it in float.
         return _device_wait_for_log(self.device,
                                     "{}/beerocks_{}.log".format(self.log_folder, program),
-                                    regex, start_line, timeout*100)
+                                    regex, start_line, timeout)
```
arnout, line 461: So, now suddenly I'm right and it's actually in seconds rather than centiseconds? :-)

This should be a separate commit.

tests/environment.py
```diff
@@ -471,19 +472,19 @@ def __init__(self, agent: ALEntityPrplWrt, iface_name: str):
         self.iface_name = iface_name
         self.agent = agent
         ip_raw = self.agent.command("ip link list dev {}".format(self.iface_name))
-        mac = re.search(r"link/ether (([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})",
-                        ip_raw).group(1)
+        self.mac = re.search(r"link/ether (([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})",
+                             ip_raw).group(1)
         self.log_folder = agent.log_folder
-        super().__init__(agent, mac)
+        super().__init__(agent, self.mac)
```
arnout: Same here, super().__init__() already sets self.mac.

tests/environment.py
```diff
@@ -471,19 +472,19 @@ def __init__(self, agent: ALEntityPrplWrt, iface_name: str):
...
 
-        VirtualAPHostapd(self, mac)
-        VirtualAPHostapd(self, mac)
+        VirtualAPHostapd(self, self.mac)
+        VirtualAPHostapd(self, self.mac)
```
arnout: Oh, I hadn't noticed before: this is not good! The bssids of the VAPs should be unique. Also, you should use the *actual* BSSIDs of the *actual* VAPs that are created on the device. You will need that to be able to associate stations successfully.

Anyway, fixing that is for another PR.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
```diff
@@ -0,0 +1,66 @@
...
+    def runTest(self):
+        # Locate agents and controller
+        for dev in self.dev:
+            if dev.controller_entity:
+                controller = dev.controller_entity
+            if dev.agent_entity:
+                agent = dev.agent_entity
+
+        dev.wired_sniffer.start(self.__class__.__name__ + "-" + dev.name)
+        mac_repeater1_upper = agent.mac.upper()
```
arnout, line 26: Please preserve the comments:
```suggestion
        # Regression test: MAC address should be case insensitive
        mac_repeater1_upper = agent.mac.upper()
```

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
```diff
@@ -0,0 +1,66 @@
...
+            "DEV_SET_CONFIG,"
+            "bss_info1,{} 8x Multi-AP-24G-1 0x0020 0x0008 maprocks1 0 1,"
+            "bss_info2,{} 8x Multi-AP-24G-2 0x0020 0x0008 maprocks2 1 0"
+            .format(mac_repeater1_upper, agent.mac))
+        controller.ucc_socket.dev_send_1905(agent.mac, 0x000A,
+                                            tlv(0x01, 0x0006, "{" + controller.mac + "}"),
+                                            tlv(0x0F, 0x0001, "{0x00}"),
+                                            tlv(0x10, 0x0001, "{0x00}"))
+
+        time.sleep(30)
```
arnout, line 38: Autoconfig renew should finish in a few seconds... Does it really take 30 seconds before it finishes?

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
arnout, line 42: No, not 30 seconds but *90* seconds? If the default timeout of half a seconds is not enough, we have a problem...

arnout (approved):
The first commit seems entirely redundant (at least, the parts that are documented in the commit message). I wonder what triggered it?

Except for the timeouts and some little changes, it looks OK.

tests/environment.py
abelog: Yep, i haven't noticed that parent class already assign those fields.  This is determinately redundant change.

tests/environment.py
abelog: Agreed. Removing it. I want to handle timeouts of `_device_wait_for_log` and display more informative message about what message wasn't found in log. Currently, you have to look for exact issue in crash log of boardfarm, which isn't convenient.

tests/environment.py
abelog: Yes, you are :smiley:

tests/environment.py
abelog: Separate task created: #1484

tests/environment.py
abelog: Exactly.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
abelog: Thanks for noticing.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
abelog: I wasn't doing a lot of experiments to find a minimal value. 30 seconds is enough for RAX40 to pass the test. Do you think is it worth to invest time and find minimal time required for test to pass on RAX40?

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
abelog: There is no great logic behind this 60 sec. timeout, so we might decrease it.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
arnout: Well... This is an actual *sleep*, not just a timeout, so we really loose that time. For this one test it's not a big deal, but if we do something like that in every test, our boardfarm tests are going to take *very long*.

And this really should take only a few hundred milliseconds. If it takes longer than that, we probably have a problem somewhere...

PR #1445, opened 2020-06-16
[agent] Implement Backhaul STA Steering

### LINKED ISSUE #408: [TASK] Backhaul STA steering

VladyslavTupikin:
## Description

According to UML diagram need to do an implementation for the _**Backhaul STA Steering**_:

1. Add implementation for the **_associate()_** method in __bwl/dwpal__: it should send `NETWORK_ENABLE `message to the **_wpa suplicant_**.
2. Add implementation for the **_Backhaul STA Steering Request_** message handler in the **_backhaul_manager_**: method should parse TLV and if channel from TLV correct, trigger the **_associate()_** method.
3. Add implementation for the **_Backhaul STA Steering Response_** message creator method in the **_backhaul_manager_**.
4. Add reaction in Backhaul manager for getting `ENABLED `event from the **_wpa supplicant_** in the **_backhaul_manager_**: **_backhaul_manager_** should check specific flag which was turned on in the handler, and **if it is true**, send  **_Backhaul STA Steering Response_** message to the controller.
5. Add implementation for the **_Backhaul STA Steering Response_** message in the controller code.

https://jira.prplfoundation.org/browse/PPM-168

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -4581,6 +4581,56 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
     return true;
 }
 
+bool backhaul_manager::create_backhaul_steering_response(
+    const wfa_map::tlvErrorCode::eReasonCode &error_code)
+{
+    auto cmdu_tx_header =
+        cmdu_tx.create(0, ieee1905_1::eMessageType::BACKHAUL_STEERING_RESPONSE_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "Failed to create Backhoul STA Steering Response message";
+        return false;
+    }
+
+    auto bh_steering_resp_tlv = cmdu_tx.addClass<wfa_map::tlvBackhaulSteeringResponse>();
+    if (!bh_steering_resp_tlv) {
+        LOG(ERROR) << "Couldn't addClass<wfa_map::tlvBackhaulSteeringResponse>";
+        return false;
+    }
+
+    auto active_hal = get_wireless_hal();
+    if (!active_hal) {
+        LOG(ERROR) << "Couldn't get active HAL";
+        return false;
+    }
+
+    sMacAddr sta_mac;
+    auto interface = active_hal->get_iface_name();
+    get_iface_mac(interface, sta_mac);
+
+    LOG(DEBUG) << "Interface: " << interface << "MAC: " << sta_mac;
+
+    bh_steering_resp_tlv->target_bssid()         = tlvf::mac_from_string(active_hal->get_bssid());
+    bh_steering_resp_tlv->backhaul_station_mac() = sta_mac;
+
+    if (!error_code) {
+        bh_steering_resp_tlv->result_code() =
+            wfa_map::tlvBackhaulSteeringResponse::eResultCode::SUCCESS;
+    } else {
+        bh_steering_resp_tlv->result_code() =
+            wfa_map::tlvBackhaulSteeringResponse::eResultCode::FAILURE;
+
+        auto error_code_tlv = cmdu_tx.addClass<wfa_map::tlvErrorCode>();
+        if (!bh_steering_resp_tlv) {
+            LOG(ERROR) << "Couldn't addClass<wfa_map::tlvErrorCode>";
+            return false;
+        }
+
+        error_code_tlv->reason_code() = error_code;
+    }
+
+    return true;
+}
+
 const std::string backhaul_manager::freq_to_radio_mac(eFreqType freq) const
 {
     auto it =
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -73,6 +73,8 @@
 #include <tlvf/wfa_map/tlvAssociatedClients.h>
 #include <tlvf/wfa_map/tlvAssociatedStaLinkMetrics.h>
 #include <tlvf/wfa_map/tlvAssociatedStaTrafficStats.h>
+#include <tlvf/wfa_map/tlvBackhaulSteeringRequest.h>
+#include <tlvf/wfa_map/tlvBackhaulSteeringResponse.h>
 #include <tlvf/wfa_map/tlvBeaconMetricsQuery.h>
 #include <tlvf/wfa_map/tlvChannelPreference.h>
 #include <tlvf/wfa_map/tlvClientCapabilityReport.h>
@@ -279,6 +281,7 @@ bool backhaul_manager::init()
             ieee1905_1::eMessageType::TOPOLOGY_DISCOVERY_MESSAGE,
             ieee1905_1::eMessageType::TOPOLOGY_QUERY_MESSAGE,
             ieee1905_1::eMessageType::VENDOR_SPECIFIC_MESSAGE,
+            ieee1905_1::eMessageType::BACKHAUL_STEERING_REQUEST_MESSAGE,
         })) {
         LOG(ERROR) << "Failed to init mapf_bus";
         return false;
@@ -2244,6 +2247,9 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     }
     case ieee1905_1::eMessageType::CHANNEL_SELECTION_REQUEST_MESSAGE: {
         return handle_channel_selection_request(cmdu_rx, src_mac);
+    }
+    case ieee1905_1::eMessageType::BACKHAUL_STEERING_REQUEST_MESSAGE: {
+        return handle_backhaul_steering_request(cmdu_rx, src_mac);
     }
     default: {
         // TODO add a warning once all vendor specific flows are replaced with EasyMesh
@@ -4515,6 +4521,64 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
 
     LOG(DEBUG) << "Sending CHANNEL_SELECTION_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
     return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+}
+
+bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received BACKHAUL_STA_STEERING message, mid=" << std::hex << mid;
+
+    auto bh_sta_steering_req = cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringRequest>();
+    if (!bh_sta_steering_req) {
+        LOG(WARNING) << "Failed cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringRequest>(), mid="
+                     << std::hex << mid;
+        return false;
+    }
+
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+
+    LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
+    send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+
+    /*
+        TODO: Need to do channel validation. If channel invalid, send ACK and Backhaul STA Steering
+              Response with the error code without trigerring HAL.
+    */
+
+    /*
+        TODO: BACKHAUL_STA_STEERING can be accepted in wired backhaul too.
+              Code below is incorrect in that case.
+    */
+    auto active_hal = get_wireless_hal();
+    if (!active_hal) {
+        LOG(ERROR) << "Couldn't get active HAL";
+        return false;
+    }
+
+    auto channel = bh_sta_steering_req->target_channel_number();
+    auto bssid   = bh_sta_steering_req->target_bssid();
+
+    auto associate = active_hal->roam(bssid, channel);
+    if (!associate) {
+        LOG(ERROR) << "Couldn't associate active HAL with bssid: " << bssid;
+
+        LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
+        send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+
+        /*
+            TODO: Need to add sending Backhaul STA Steering Response with the FAILURE error code.
+        */
+
+        return false;
+    }
+
+    return true;
 }
 
 const std::string backhaul_manager::freq_to_radio_mac(eFreqType freq) const
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -4558,10 +4558,27 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
     LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
     send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 
-    /*
-        TODO: Need to do channel validation. If channel invalid, send ACK and Backhaul STA Steering
-              Response with the error code without trigerring HAL.
-    */
+    auto channel    = bh_sta_steering_req->target_channel_number();
+    auto oper_class = bh_sta_steering_req->operating_class();
+
+    auto is_valid_channel = son::wireless_utils::is_channel_in_operating_class(oper_class, channel);
+
+    if (!is_valid_channel) {
+        LOG(WARNING) << "Invalid channel number";
+
+        auto response = create_backhaul_steering_response(
+            wfa_map::tlvErrorCode::eReasonCode::
+                BACKHAUL_STEERING_REQUEST_REJECTED_CANNOT_OPERATE_ON_CHANNEL_SPECIFIED);
+
+        if (!response) {
+            LOG(ERROR) << "Failed to build Backhaul STA Steering Response message.";
+            return false;
+        }
+
+        send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+
+        return false;
+    }
 
     /*
         TODO: BACKHAUL_STA_STEERING can be accepted in wired backhaul too.
@@ -4573,8 +4590,7 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
         return false;
     }
 
-    auto channel = bh_sta_steering_req->target_channel_number();
-    auto bssid   = bh_sta_steering_req->target_bssid();
+    auto bssid = bh_sta_steering_req->target_bssid();
 
     auto associate = active_hal->roam(bssid, channel);
     if (!associate) {
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -973,6 +973,18 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
         eth_link_poll_timer = std::chrono::steady_clock::now();
         m_eth_link_up       = network_utils::linux_iface_is_up_and_running(m_sConfig.wire_iface);
         FSM_MOVE_STATE(OPERATIONAL);
+
+        // This event may come as a result of enabling the backhaul, but also as a result
+        // of steering. *Only* in case it was the result of steering, we need to send a steering
+        // response.
+        if (m_backhaul_sta_steering_enable) {
+            m_backhaul_sta_steering_enable = false;
+
+            create_backhaul_steering_response(wfa_map::tlvErrorCode::eReasonCode::RESERVED);
+
+            LOG(DEBUG) << "Sending BACKHAUL_STA_STEERING_RESPONSE_MESSAGE";
+            send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+        }
         break;
     }
     // Backhaul manager is OPERATIONAL!
@@ -4571,12 +4583,21 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
         LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
         send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 
-        /*
-            TODO: Need to add sending Backhaul STA Steering Response with the FAILURE error code.
-        */
+        auto response = create_backhaul_steering_response(
+            wfa_map::tlvErrorCode::eReasonCode::
+                BACKHAUL_STEERING_REQUEST_REJECTED_TARGET_BSS_SIGNAL_NOT_SUITABLE);
+
+        if (!response) {
+            LOG(ERROR) << "Failed to build Backhaul STA Steering Response message.";
+            return false;
+        }
+
+        send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 
         return false;
     }
+
+    m_backhaul_sta_steering_enable = true;
 
     return true;
 }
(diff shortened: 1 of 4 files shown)


common/beerocks/bwl/dwpal/sta_wlan_hal_dwpal.cpp
```diff
@@ -242,7 +242,42 @@ bool sta_wlan_hal_dwpal::connect(const std::string &ssid, const std::string &pas
 
 bool sta_wlan_hal_dwpal::associate(const sMacAddr &bssid, uint8_t channel)
 {
-    LOG(DEBUG) << "sta_wlan_hal_dwpal::associate() is not implemented!";
+    auto bssid_str = tlvf::mac_to_string(bssid);
+
+    LOG(DEBUG) << " Associate iface " << get_iface_name() << " to BSSID = " << bssid_str
+               << ", Channel = " << int(channel);
```
arnout: As mentioned in #1424 this function isn't needed. It was wrong anyway because it didn't do any steering.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4458,6 +4464,49 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
     return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received BACKHAUL_STA_STEERING message, mid=" << std::hex << int(mid);
```
```suggestion
    LOG(DEBUG) << "Received BACKHAUL_STA_STEERING message, mid=" << std::hex << mid;
```
Explicit conversion to int is only needed for enums and for int8_t. It used to be needed for uint8_t as well, but that has been solved by adding an explicit operator<< overload.

And anyway, mid is uint16_t so this was never needed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4458,6 +4464,49 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
...
+
+    auto bh_sta_steering_req = cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringRequest>();
+    if (!bh_sta_steering_req) {
+        LOG(ERROR) << "Failed cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringRequest>(), mid="
+                   << std::hex << int(mid);
```
```suggestion
                   << std::hex << mid;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4458,6 +4464,49 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
...
 
+bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageRx &cmdu_rx,
+                                                        const std::string &src_mac)
+{
+    const auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received BACKHAUL_STA_STEERING message, mid=" << std::hex << int(mid);
+
+    auto bh_sta_steering_req = cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringRequest>();
+    if (!bh_sta_steering_req) {
+        LOG(ERROR) << "Failed cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringRequest>(), mid="
```
```suggestion
        LOG(WARNING) << "Failed cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringRequest>(), mid="
```
Since this message comes from outside, it is not an "error" (in the sense that something is wrong with our system). Rather , it is invalid input.

Doesn't matter much of course, but we should probably do this right.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4458,6 +4464,49 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
...
+              Response with the error code without trigerring HAL.
+    */
+
+    auto active_hal = get_wireless_hal();
+    if (!active_hal) {
+        LOG(ERROR) << "Couldn't get active HAL";
+        return false;
+    }
+
+    auto associate = active_hal->associate(bssid, channel);
```
```suggestion
    auto associate = active_hal->roam(bssid, channel);
```

The proper way is probably to use the state machine instead. That said, maybe it's not needed, and it's way more efficient if we can roam directly here rather than going through the state machine.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4458,6 +4464,49 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
...
+        LOG(ERROR) << "Couldn't associate active HAL with bssid: " << tlvf::mac_to_string(bssid);
+    }
+
+    // build ACK message CMDU
+    auto cmdu_tx_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::ACK_MESSAGE);
+    if (!cmdu_tx_header) {
+        LOG(ERROR) << "cmdu creation of type ACK_MESSAGE, has failed";
+        return false;
+    }
+    LOG(DEBUG) << "sending ACK message to the originator, mid=" << std::hex << int(mid);
```
```suggestion
    LOG(DEBUG) << "sending ACK message to the originator, mid=" << std::hex << mid;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4507,6 +4506,31 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
     return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::create_backhaul_steering_response(
+    const sMacAddr &sta_mac, const sMacAddr &target_bssid,
+    const wfa_map::tlvBackhaulSteeringResponse::eResultCode &result_code)
+{
+    auto cmdu_tx_header = m_expected_backhaul_sta_steering.response.create(
+        0, ieee1905_1::eMessageType::BACKHAUL_STEERING_RESPONSE_MESSAGE);
```
arnout: Since we are not building up the message from different responses here but we're just expecting a single one, there's no need to keep a response buffer in m_expected_backhaul_sta_steering. We can simply use cmdu_tx here.
```suggestion
    auto cmdu_tx_header = cmdu_tx.create(
        0, ieee1905_1::eMessageType::BACKHAUL_STEERING_RESPONSE_MESSAGE);
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -886,6 +886,14 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
                 FSM_MOVE_STATE(MASTER_DISCOVERY);
             }
 
+            if (m_expected_backhaul_sta_steering.backhaul_sta_steering_enable) {
+                LOG(DEBUG) << "Sending BACKHAUL_STA_STEERING_RESPONSE_MESSAGE";
+
+                m_expected_backhaul_sta_steering.backhaul_sta_steering_enable = false;
+
+                send_cmdu_to_bus(m_expected_backhaul_sta_steering.response, controller_bridge_mac,
+                                 bridge_info.mac);
```
arnout: Here, we should also create the response.

The bssid is active_hal->get_bssid(). Funnily enough, the station MAC address is not so easy to get: you have to get it from active_hal->get_iface_name() using get_iface_mac().

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1794,6 +1799,25 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
     return true;
 }
 
+bool master_thread::handle_cmdu_1905_backhaul_sta_steering_response(
+    const std::string &src_mac, ieee1905_1::CmduMessageRx &cmdu_rx)
+{
+    auto mid = cmdu_rx.getMessageId();
+    LOG(DEBUG) << "Received BACKHAUL_STA_STEERING_MESSAGE from " << src_mac << ", mid=" << std::hex
+               << int(mid);
```
arnout:
```suggestion
               << mid;
```

arnout (approved):
Although some comments are pretty major, since we don't have any way to test this yet there's no point gating the merge on that. So accepting.

common/beerocks/bwl/dwpal/sta_wlan_hal_dwpal.cpp
VladyslavTupikin: This commit removed.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: accepted!

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Roger that.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: accepted

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: solved

controller/src/beerocks/master/son_master_thread.cpp
VladyslavTupikin: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4580,6 +4579,41 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
     return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::create_backhaul_steering_response(
+    const wfa_map::tlvBackhaulSteeringResponse::eResultCode &result_code)
+{
+    auto cmdu_tx_header =
+        cmdu_tx.create(0, ieee1905_1::eMessageType::BACKHAUL_STEERING_RESPONSE_MESSAGE);
```
vitalii-komisarenko, line 4625: Is it OK to use message ID 0?
Often message ID of the request and the response should be the same.
Certification test 4.9.1 does not check it. Can it change in the future?

vitalii-komisarenko (approved):
Some commit messages are too long.
"target_bssid" is menioned two times in the commit message of 7454c656c19467af759282066cc7f793f247f056.
Other than that looks OK.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4505,6 +4511,58 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
...
+    /*
+        TODO: Need to do channel validation. If channel invalid, send ACK and Backhaul STA Steering
+              Response with the error code without trigerring HAL.
+    */
+
+    auto active_hal = get_wireless_hal();
+    if (!active_hal) {
+        LOG(ERROR) << "Couldn't get active HAL";
+        return false;
+    }
```
arnout, line 4591: This can actually happen if we have wired backhaul. I'm not sure what to do in that case, but it would be good to at least add a TODO here. I think we should still send an ACK anyway. Sending ACK should be done before getting active hal.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4505,6 +4511,58 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
...
+        /*
+            TODO: Need to add sending Backhaul STA Steering Response with the FAILURE error code.
+        */
+
+        return false;
+    }
+
+    LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
+    return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+}
```
arnout: Sending ACK should already be done before doing the roam, because we have to do it even if the roam fails.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4563,6 +4562,41 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
     return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 }
 
+bool backhaul_manager::create_backhaul_steering_response(
+    const wfa_map::tlvBackhaulSteeringResponse::eResultCode &result_code)
```
arnout: I think we should have as parameter the error code rather than the result code (see below).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: Yes it has to be a new MID. The request is answered with an ACK that has the same MID as the request, while the response uses a new MID. It's a bit confusing...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4563,6 +4562,41 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
...
+    if (!active_hal) {
+        LOG(ERROR) << "Couldn't get active HAL";
+        return false;
+    }
+
+    sMacAddr sta_mac;
+    auto interface = active_hal->get_iface_name();
+    get_iface_mac(interface, sta_mac);
+
+    LOG(DEBUG) << "Interface: " << interface << "MAC: " << tlvf::mac_to_string(sta_mac);
```
```suggestion
    LOG(DEBUG) << "Interface: " << interface << "MAC: " << sta_mac;
```

vitalii-komisarenko (approved):
My review is identical to previous one: https://github.com/prplfoundation/prplMesh/pull/1445#pullrequestreview-438281483
Mostly related to commit messages.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4563,6 +4562,41 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
...
+
+    bh_steering_resp_tlv->target_bssid()         = tlvf::mac_from_string(active_hal->get_bssid());
+    bh_steering_resp_tlv->backhaul_station_mac() = sta_mac;
+    bh_steering_resp_tlv->result_code()          = result_code;
```
arnout: If result code is failure, we also have to provide an error code TLV with the error code. See the 4th paragraph of section 12.

So I think it's better to pass the error code directly, and to set the result code (and include the error TLV) depending on whether the error code is 0 or not.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
```diff
@@ -597,6 +601,8 @@ class backhaul_manager : public btl::transport_socket_thread {
 
     sExpectedChannelSelection m_expected_channel_selection;
 
+    bool backhaul_sta_steering_enable = false;
```
arnout: This change is unrelated and not needed I think.

arnout (commented):
Only reviewed the first 2 commits, but I have to take a break now :-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -969,6 +969,16 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
         eth_link_poll_timer = std::chrono::steady_clock::now();
         m_eth_link_up       = network_utils::linux_iface_is_up_and_running(m_sConfig.wire_iface);
         FSM_MOVE_STATE(OPERATIONAL);
+
+        if (backhaul_sta_steering_enable) {
```
arnout: This needs a comment IMO. Something like:

```
// This event may come as a result of enabling the backhaul, but also as a result
// of steering. *Only* in case it was the result of steering, we need to send a steering
// response.
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4551,13 +4561,21 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
         LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
         send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
 
-        /*
-            TODO: Need to add sending Backhaul STA Steering Response with the FAILURE error code.
-        */
+        auto response = create_backhaul_steering_response(
+            wfa_map::tlvBackhaulSteeringResponse::eResultCode::FAILURE);
```
arnout: As mentioned earlier, in case of failure we also need to give an error code. We have the choice of BACKHAUL_STEERING_REQUEST_REJECTED_CANNOT_OPERATE_ON_CHANNEL_SPECIFIED, BACKHAUL_STEERING_REQUEST_REJECTED_TARGET_BSS_SIGNAL_NOT_SUITABLE or BACKHAUL_STEERING_REQUEST_AUTHENTICATION_OR_ASSOCIATION_REJECTED. Actually none of these are applicable here because if the roam() call failed, it just means there's a problem with the connection to wpa_supplicant. I think BACKHAUL_STEERING_REQUEST_REJECTED_TARGET_BSS_SIGNAL_NOT_SUITABLE is the closest to be accurate.

controller/src/beerocks/master/son_master_thread.cpp
```diff
@@ -1794,6 +1799,25 @@ bool master_thread::handle_cmdu_1905_topology_response(const std::string &src_ma
...
+
+    auto tlv_backhaul_sta_steering_resp = cmdu_rx.getClass<wfa_map::tlvBackhaulSteeringResponse>();
+    if (!tlv_backhaul_sta_steering_resp) {
+        LOG(ERROR) << "Failed getClass<wfa_map::tlvBackhaulSteeringResponse>";
+        return false;
+    }
+
+    LOG(DEBUG) << "BACKHAUL_STA_STEERING_MESSAGE result_code: "
+               << int(tlv_backhaul_sta_steering_resp->result_code());
+
```
arnout, line 1817: The controller should respond with an ACK.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4554,6 +4544,33 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
...
+    auto is_valid_channel = son::wireless_utils::is_channel_in_operating_class(oper_class, channel);
+
+    if (!is_valid_channel) {
+        LOG(WARNING) << "Invalid channel number";
+
+        LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
+        send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+
+        auto response = create_backhaul_steering_response(
+            wfa_map::tlvBackhaulSteeringResponse::eResultCode::FAILURE);
```
arnout: Here the error code must be BACKHAUL_STEERING_REQUEST_REJECTED_CANNOT_OPERATE_ON_CHANNEL_SPECIFIED.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4554,6 +4544,33 @@ bool backhaul_manager::handle_backhaul_steering_request(ieee1905_1::CmduMessageR
         return false;
     }
 
+    auto is_valid_channel = son::wireless_utils::is_channel_in_operating_class(oper_class, channel);
+
+    if (!is_valid_channel) {
+        LOG(WARNING) << "Invalid channel number";
+
+        LOG(DEBUG) << "Sending ACK message to the originator, mid=" << std::hex << mid;
+        send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
```
arnout: This is independent on whether the channel is valid or not, so it can be done before is_valid_channel.

arnout (approved):
Except for including an error code TLV in the response in case of failure, none of my comments are really critical. Also there's no need for me to re-review.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Added TODO above getting HAL. Sending ACK message moved upper than getting active HAL.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: accepted!

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: You are right. Added creation for ErrorCodeTlv

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Accepted

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
VladyslavTupikin: Moved to commit with usage.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Accepted

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Accepted.

controller/src/beerocks/master/son_master_thread.cpp
VladyslavTupikin: Added sending ACK message to the agent and fixed commit message.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: Accepted

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -4517,6 +4535,139 @@ bool backhaul_manager::handle_slave_channel_selection_response(ieee1905_1::CmduM
...
+            wfa_map::tlvBackhaulSteeringResponse::eResultCode::FAILURE;
+    }
+
+    auto error_code_tlv = cmdu_tx.addClass<wfa_map::tlvErrorCode>();
+    if (!bh_steering_resp_tlv) {
+        LOG(ERROR) << "Couldn't addClass<wfa_map::tlvErrorCode>";
+        return false;
+    }
+
+    error_code_tlv->reason_code() = error_code;
```
arnout: This should only be added if error_code != 0.

VladyslavTupikin: fixed

PR #1447, opened 2020-06-17
bcl: UCC listener: prevent buffer overflow

vitalii-komisarenko:
The code inside `beerocks_ucc_listener::custom_message_handler`
does not check the value returned by `readBytes`.
It is treated as a length of a string, the assumed end of the string is then marked by '\0'.

This approach has two issues:
* if `readBytes` returns an error (negative number) memory before the read buffer is written
* if `readBytes` returns a number equal to the length of the buffer, '\0' is written right after the buffer.

This commit prevents these issues.

Signed-off-by: Vitalii Komisarenko <vitalii.komisarenko@gmail.com>

### DIFF
--- a/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
+++ b/common/beerocks/bcl/source/beerocks_ucc_listener.cpp
@@ -438,9 +438,15 @@ bool beerocks_ucc_listener::custom_message_handler(Socket *sd, uint8_t *rx_buffe
                                                    size_t rx_buffer_size)
 {
     std::string command_string;
-    auto available_bytes       = sd->readBytes(rx_buffer, rx_buffer_size, true);
+    auto available_bytes = sd->readBytes(rx_buffer, rx_buffer_size, true);
+
+    if (available_bytes <= 0) {
+        LOG(ERROR) << "Cannot read from socket";
+        return true;
+    }
+
     rx_buffer[available_bytes] = '\0';
-    command_string.assign(reinterpret_cast<char *>(rx_buffer));
+    command_string.assign(reinterpret_cast<char *>(rx_buffer), available_bytes);
 
     beerocks::string_utils::trim(command_string);
 


common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -438,14 +438,17 @@ bool beerocks_ucc_listener::custom_message_handler(Socket *sd, uint8_t *rx_buffe
...
 
-    beerocks::string_utils::trim(command_string);
-
-    handle_wfa_ca_command(command_string);
+        beerocks::string_utils::trim(command_string);
 
+        handle_wfa_ca_command(command_string);
+    } else {
+        LOG(ERROR) << "Cannot read from socket";
+    }
```
morantr:
```cpp
if (error) {
   // handle error
}

// rest of the code
```

morantr (approved):

ghost (approved):

common/beerocks/bcl/source/beerocks_ucc_listener.cpp
```diff
@@ -438,7 +438,13 @@ bool beerocks_ucc_listener::custom_message_handler(Socket *sd, uint8_t *rx_buffe
...
 {
     std::string command_string;
-    auto available_bytes       = sd->readBytes(rx_buffer, rx_buffer_size, true);
+    auto available_bytes = sd->readBytes(rx_buffer, rx_buffer_size - 1, true);
+
+    if (available_bytes <= 0) {
+        LOG(ERROR) << "Cannot read from socket";
+        return true;
+    }
+
```
arnout, line 447: Actually, if available_bytes is equal to the buffer size, it is also an error because the command will be incomplete and we're not going to read a second time and append to the buffer. But I guess handle_wfa_command will deal with that.

arnout (approved):
My comment has no bearing on this PR really, so approving.

PR #1435, opened 2020-06-14
ZMQ Replacement #1 (Event Loop)

vitalybu:
Due to licensing issues with the ZMQ project, this PR aims to to replace the existing ZMQ broker/bus implementation with a custom broker.

PR #1449, opened 2020-06-18
Yet another reactor design pattern implementation

mariomaz:
Based in `select()`, not in `epoll()`
File descriptors can correspond to any OS resource implementing the "file descriptor" interface, not only sockets (for example, it can use timers too). In fact, there's no code related to sockets and reading and writing is left out of this library.
It's not fully compliant with prplMesh coding style rules.
Includes some unit tests


ghost:
I haven't looked at your PR in detail yet, but generally, `select()` should always be avoided.

See e.g. https://beesbuzz.biz/code/5739-The-problem-with-select-vs-poll for why

mariomaz:
Sorry, ignore all commits except the last one

mariomaz:
> select() should always be avoided.

I know. I gave what I had. If you like the design, we can easily replace select() with epoll()

common/beerocks/btl/include/btl/btl.h
```diff
@@ -50,12 +62,22 @@ class transport_socket_thread : public socket_thread {
...
+     * @param length Message length.
+     * @return True on success and false otherwise.
+     */
+    bool send_cmdu_to_bus(ieee1905_1::CmduMessage &cmdu, uint32_t iface_index,
+                          const std::string &dst_mac, const std::string &src_mac, uint16_t length);
 
 private:
     bool bus_init();
-    bool bus_send(ieee1905_1::CmduMessage &cmdu, const std::string &dst_mac,
+    bool bus_send(ieee1905_1::CmduMessage &cmdu, uint32_t iface_index, const std::string &dst_mac,
```
morantr, line 80: I don't understand it.
The message is sent to the bus with src mac which is the bridge/al_mac.
doesn't ifindex means index of radio interface? Because we send the message from the "bridge" interface. Does it mean face index of the bridge maybe? It is not clear to me.
Anyway, I don't think that it is a good idea to add it as an extra argument. For whatever reason we need it, couldn't it be generated internally on the function?

mariomaz: Please ignore all commits except the last one
I made a mistake and created the PR before committing my code. Sorry for that
Your comment belongs to #1420, just in case you want to review it

source/reactor/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
+project(reactor VERSION ${prplmesh_VERSION})
+
+# Set the base path for the current module
+set(MODULE_PATH ${CMAKE_CURRENT_LIST_DIR})
+
+# Sources
+file(GLOB_RECURSE reactor_sources ${MODULE_PATH}/src/*.c*)
```
arnout, line 7: I don't like wildcards for specifying the source, because it means that when a file is added or removed, your Makefile/build.ninja will not be updated automatically.

(nitpick)

source/reactor/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+
+# Support more permissive c++ language
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
```
arnout, line 10: Which specific feature needs that?

source/reactor/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+
+# Build the library
+add_library(${PROJECT_NAME} ${reactor_sources})
+set_target_properties(${PROJECT_NAME} PROPERTIES LINK_FLAGS "-Wl,-z,defs")
+set_target_properties(${PROJECT_NAME} PROPERTIES VERSION ${prplmesh_VERSION} SOVERSION ${prplmesh_VERSION_MAJOR})
+target_include_directories(${PROJECT_NAME}
+    PUBLIC
+        $<BUILD_INTERFACE:${MODULE_PATH}/inc>
```
arnout, line 18: Currently we use `include` everywhere. Is there any reason to switch to `inc`?

(We also use `src` rather than `source` everywhere except for bcl, but that's a more defendable change.)

source/reactor/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+    if (COVERAGE)
+        set_target_properties(${TEST_PROJECT_NAME} PROPERTIES COMPILE_FLAGS "--coverage -fPIC -O0")
+        set_target_properties(${TEST_PROJECT_NAME} PROPERTIES LINK_FLAGS "--coverage")
+    endif()
+    target_include_directories(${TEST_PROJECT_NAME}
+        PRIVATE
+            ${PLATFORM_INCLUDE_DIR}
+        PUBLIC
+            $<BUILD_INTERFACE:${MODULE_PATH}/inc>
+            ${MODULE_PATH}/src
```
arnout, line 48: This shouldn't be needed. For include files in the source directory itself, simply use `"..."`.

source/reactor/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+    )
+    target_link_libraries(${TEST_PROJECT_NAME} reactor)
```
arnout, line 50: In other unit tests, we link the library sources directly instead of linking with the library. This is good because it allows us to configure it for coverage only for testing.

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
+#ifndef Reactor_EventHandlers_H
+#define Reactor_EventHandlers_H
+
+#include <functional>
+
+namespace u {
```
arnout, line 6: Although this stuff is obviously not beerocks-related, I think it should still be in the beerocks namespace.

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
...
+namespace reactor {
+class EventLoop;
+
+/**
+ * @brief Event handler function definition.
+ *
+ * Parameters to the event handler function are:
+ * - The file descriptor of the OS resource where the event was originated at. Event handler can
+ * perform, for example, accept, read, write and close operations on the file descriptor.
```
arnout, line 15: I was thinking - wouldn't it be useful to have an explicit class for file descriptors, rather than a generic int? It would make things much clearer IMO. And we could even consider giving it a move constructor and doing close() in the destructor...

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
...
+
+namespace u {
+namespace reactor {
+class EventLoop;
+
+/**
+ * @brief Event handler function definition.
+ *
+ * Parameters to the event handler function are:
+ * - The file descriptor of the OS resource where the event was originated at. Event handler can
```
arnout, line 14: Since this is an actual function, you can document it with
```suggestion
 * @param[in] fd The file descriptor of the OS resource where the event was originated at.
```
(of course, it requires naming the parameters in the function signature).

source/reactor/inc/reactor/EventLoop.h
```diff
@@ -0,0 +1,73 @@
...
+ */
+class EventLoop {
+public:
+    virtual ~EventLoop() = default;
+
+    /**
+     * @brief Registers a set of event handlers for the given file descriptor.
+     *
+     * Event handler for the event that occurred will be called back when the file descriptor is
+     * ready for a read/write O operation, when an error occurs or when given timeout expires.
```
arnout, line 31:
```suggestion
     * ready for a read/write operation, when an error occurs or when given timeout expires.

```

source/reactor/inc/reactor/EventLoop.h
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @param fd File descriptor. Handle that identifies a resource managed within an OS:
+     * network connections, open files, timers, synchronization objects, etc.
+     * @param timeout Time to wait in milliseconds (-1 to wait indefinitely) for a read/write
+     * event to occur.
```
arnout, line 36: Is this part of the reactor pattern?

I don't think it's very useful. There is rarely a need for timing out on a read or write of a specific file descriptor. What you do need is a generic timeout, i.e. a callback that is called so many seconds after now, or at a specific time. The only case I can think of where you can use this kind of timeout if in a request-response interaction. But for such a pattern, we would like the handler to be one-shot as well. And it makes no sense to combine it with a write handler either. So I think it would be better to leave that for another day.

Implementing timers is a bit more complicated, so I think it's better to start without them and with a single timeout that is set globally. That is already sufficient to replace the Poller that we have now. Timers can be added later.

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
...
+    EventHandler onRead = nullptr;
+
+    /**
+     * Hook method that is called back by the event loop to handle write events.
+     * Write events are dispatched when the file descriptor is ready for a write operation (a write
+     * operation will not block).
+     * @param fd File descriptor the event was originated at.
+     * @param loop Event loop where the event was caught on.
+     */
+    EventHandler onWrite = nullptr;
```
arnout, line 45: I don't think it's useful to have this. We're not using it at the moment because we rely on kernel buffering for all writes.

On the other hand, I guess it doesn't hurt to have it.

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
...
+ * perform, for example, accept, read, write and close operations on the file descriptor.
+ * - The event loop where it was caught. Event handler can install new handlers, remove existing
+ * handlers and even ask for loop termination.
+ */
+using EventHandler = std::function<void(int, EventLoop *)>;
+
+/**
+ * Set of event handler functions, one function to handle each possible event happened.
+ * Handlers are grouped into a struct to facilitate passing them as a single parameter to the install and
+ * remove handlers functions of the event loop.
```
arnout, line 24: Since I think there shouldn't be a write or timeout handler, I don't think there's much added value to collecting them in a struct. Even with all four of them I don't think there's much added value - it just motivates people to write
```
EventHandlers eh;
eh.onRead = [&](int, EventLoop *){...};
eh.onError = ...
```
which is exactly the pattern that @RanRegev wants to stomp out.

source/reactor/inc/reactor/EventLoop.h
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @brief Runs message loop.
+     *
+     * Method returns when loop is explicitly terminated, when an error occurs or when there is
+     * no event to wait for (i.e.: no handler installed).
+     *
+     * @return True on success and false otherwise.
+     */
+    virtual bool run(void) = 0;
```
arnout, line 59: I think there should be a run_once instead, and *maybe* also a run() that just calls run_once in a loop.

source/reactor/inc/reactor/EventLoopImpl.h
```diff
@@ -0,0 +1,52 @@
...
+namespace reactor {
+class EventReaction;
+
+class EventLoopImpl : public EventLoop {
+public:
+
+    /**
+     * @see EventLoop::registerHandlers
+     */
+    bool registerHandlers(int fd, long timeout, const EventHandlers &handlers) override;
```
arnout, line 19: Did we ever converge on whether or not there should be an explicit virtual for override?

source/reactor/inc/reactor/EventLoopImpl.h
```diff
@@ -0,0 +1,52 @@
...
+     */
+    void terminate(void) override;
+
+protected:
+    /**
+     * Map of registered event reactions.
+     * Map key is the file descriptor and the value is the event reaction, a structure made of a
+     * file descriptor, a timeout for read/write events and the set of event handlers.
+     */
+    std::map<int, std::shared_ptr<EventReaction>> m_eventReactions;
```
arnout, line 42: (nitpick) std::map is not the go-to class for maps; std::unordered_map is. Actually, CppGuidelines even says to use std::vector if the map is going to be smaller than a few KB, because vectors are simply super-efficient. But honestly I think the convenience of using `m_eventReactions.find(fd)` rather than a find_if construct is worth it.

Stll, unordered_map should be preferred over map. Efficiency will actually probably be about the same because the container is very small, but I think it's semantically more appropriate, unless we really want it to be *sorted*.

source/reactor/src/EventLoopImpl.cpp
```diff
@@ -0,0 +1,144 @@
...
+        // descriptor in any of the three sets, plus 1.
+        int max_fd = 0;
+
+        // The timeout argument for select() is computed as the minimum time to wait (-1 if not
+        // set yet) of all file descriptors
+        long min_timeout = -1;
+
+        // Get current time (time at this loop iteration)
+        struct timespec current;
+        clock_gettime(CLOCK_MONOTONIC, &current);
```
arnout, line 49: As mentioned, I think we should remove the timeout handling and add it as proper timers later. Also, I think that should be tested, which will require reworking the implementation of it so the clock and select calls don't need to be mocked.

source/reactor/inc/reactor/EventLoop.h
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @param fd File descriptor. Handle that identifies a resource managed within an OS:
+     * network connections, open files, timers, synchronization objects, etc.
+     * @param timeout Time to wait in milliseconds (-1 to wait indefinitely) for a read/write
+     * event to occur.
+     * @param handlers Set of event handlers: class with the methods to be called back when an
+     * event occurs.
+     * @return True on success and false otherwise.
+     */
+    virtual bool registerHandlers(int fd, long timeout, const EventHandlers &handlers) = 0;
```
arnout, line 41: If we have timeouts, I think we should use an expressive type for them. `std::chrono::milliseconds` seems extremely appropriate.

source/reactor/src/EventLoopImpl.cpp
```diff
@@ -0,0 +1,144 @@
...
+        // timeout events.
+        // The select() call indicates which handles can have operations invoked on them
+        // synchronously without blocking the application process.
+        // select() returns 0 if timeout, 1 if input/output available or exception, -1 if error.
+        int rc = select(max_fd + 1, &read_fds, &write_fds, &exception_fds,
+                        (-1 == min_timeout) ? NULL : &timeout);
+
+        // Get current time to compute elapsed time for each handler and determine if its timeout
+        // has elapsed for.
+        clock_gettime(CLOCK_MONOTONIC, &current);
```
arnout, line 102: Please don't reuse the same variable name, it is confusing...

source/reactor/src/EventLoopImpl.cpp
```diff
@@ -0,0 +1,144 @@
...
+{
+    if (m_eventReactions.find(fd) == m_eventReactions.end()) {
+        m_eventReactions[fd] = std::make_shared<EventReaction>(fd, timeout, handlers);
+        return true;
+    }
+
+    return false;
+}
+
+bool EventLoopImpl::removeHandlers(int fd) { return m_eventReactions.erase(fd) > 0; }
```
arnout, line 21: Honest question: why are these implemented here and not in the base class? There's no example here of how the mock will be used, but I think you'd typically use it to inject event, not to add expectations about handlers that are added to it. So it makes sense to keep track of the handlers in the mock as well.

source/reactor/src/TimeTools.h
```diff
@@ -0,0 +1,31 @@
...
+class TimeTools {
+public:
+    /**
+     * Gets elapsed time from start till end time (in milliseconds).
+     * @param start Start time.
+     * @param end End time.
+     * @return Elapsed time in milliseconds.
+     */
+    static long getElapsedTime(const struct timespec *start, const struct timespec *end);
+};
```
arnout, line 27: I think it's better to use std::duration instead.

source/reactor/unit_tests/EventReaderMock.h
```diff
@@ -0,0 +1,29 @@
...
+    {
+        onRead    = [&](int fd, EventLoop *loop) { handleRead(fd, loop); };
+        onError   = [&](int fd, EventLoop *loop) { handleError(fd, loop); };
+        onTimeout = [&](int fd, EventLoop *loop) { handleTimeout(fd, loop); };
+    }
+
+    MOCK_METHOD(void, handleRead, (int fd, EventLoop *loop));
+    MOCK_METHOD(void, handleError, (int fd, EventLoop *loop));
+    MOCK_METHOD(void, handleTimeout, (int fd, EventLoop *loop));
+};
```
arnout, line 25: Just for my understanding...

If we didn't have the EventHandlers structure, we'd need something like this I think, right?
```
class EventReaderMock {
public:
    MOCK_METHOD(void, operator(), (int fd, EventLoop *loop));
};
```
and we'd be able to use that directly as the onRead handler?

Actually, can't we use the same mock class for all types of handlers, since they have the same type anyway?

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * @brief Event handler function definition.
+ *
+ * Parameters to the event handler function are:
+ * - The file descriptor of the OS resource where the event was originated at. Event handler can
+ * perform, for example, accept, read, write and close operations on the file descriptor.
+ * - The event loop where it was caught. Event handler can install new handlers, remove existing
+ * handlers and even ask for loop termination.
+ */
+using EventHandler = std::function<void(int, EventLoop *)>;
```
arnout, line 19: Why is EventLoop a pointer rather than a reference?

source/reactor/unit_tests/EventHandlersTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+    StrictMock<EventWriterMock> writer;
+
+    int sv[2];
+    int rc = socketpair(AF_UNIX, SOCK_STREAM, 0, sv);
+    ASSERT_NE(-1, rc);
+
+    {
+        InSequence sequence;
+
+        EXPECT_CALL(writer, handleWrite(sv[0], &loop)).WillOnce(Invoke([](int fd, EventLoop *loop) {
```
arnout, line 33: Does EXPECT_CALL also imply that we don't expect any of the other methods to be called? If not, shouldn't we add that?

source/reactor/unit_tests/EventHandlersTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+constexpr long test_timeout = 100;
+
+namespace {
+
+TEST(EventHandlersTest, read_write_ShouldSucceed)
+{
+    EventLoopImpl loop;
+    StrictMock<EventReaderMock> reader;
+    StrictMock<EventWriterMock> writer;
```
arnout, line 24: Is there any reason to have separate mock classes for reader and writer? We could just define all handlers in a single calls and do
```suggestion
    StrictMock<EventHandlerMock> reader;
    StrictMock<EventHandlerMock> writer;
```
Or am I missing something?

source/reactor/unit_tests/EventHandlersTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+TEST(EventHandlersTest, timeout_ShouldSucceed)
+{
+    EventLoopImpl loop;
+    StrictMock<EventReaderMock> reader;
+    StrictMock<EventWriterMock> writer;
+
+    int sv[2];
+    int rc = socketpair(AF_UNIX, SOCK_STREAM, 0, sv);
+    ASSERT_NE(-1, rc);
```
arnout, line 62: Would be useful to factorize this in a setup function, I think.

source/reactor/unit_tests/EventLoopImplTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    ASSERT_TRUE(ok);
+}
+
+TEST(EventLoopImplTest, terminate_ShouldSucceed)
+{
+    EventLoopImpl loop;
+    EventHandlers handlers;
+
+    handlers.onWrite = [&](int fd, EventLoop *loop) { loop->terminate(); };
```
arnout, line 76: If you're not actually using any variable, IMO it's better to remove the reference:
```suggestion
    handlers.onWrite = [](int fd, EventLoop *loop) { loop->terminate(); };
```
To me, it feels like "const" in the sense that it tells the reader that whatever the function is doing is not going to leak into my local variables.

source/reactor/unit_tests/EventLoopImplTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    loop.registerHandlers(1, 0, handlers);
```
arnout, line 78: Ugh, that feels like a hack, assuming that stdout is writable... I would actually prefer for all tests to have a setup that creates a socketpair, so we have some dedicated file descriptors.

source/reactor/unit_tests/EventLoopImplTest.cpp
```diff
@@ -0,0 +1,109 @@
...
+    bool ok = loop.run();
+    loop.removeHandlers(1);
+
+    ASSERT_TRUE(ok);
+}
+
+TEST(EventLoopImplTest, timer_ShouldSucceed)
```
arnout, line 85: So... This just tests that timerfd works... It doesn't actually test anything from EventLoopImpl, right? I mean, there is simply no way to break EventLoopImpl that makes this one fail while the other tests still work...

In other words, I don't think this test is useful.

source/reactor/unit_tests/EventHandlersTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+        EXPECT_CALL(reader, handleTimeout(sv[1], &loop))
+            .WillOnce(Invoke([](int fd, EventLoop *loop) {
+                loop->removeHandlers(fd);
+                close(fd);
+            }));
+    };
+
+    loop.registerHandlers(sv[0], test_timeout, writer);
+    loop.registerHandlers(sv[1], test_timeout, reader);
```
arnout, line 82: I'm missing a test that verifies that the timeout is not hit if the event happens first.

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
+#ifndef Reactor_EventHandlers_H
```
arnout, line 1: Missing copyright header (in all files)

source/reactor/unit_tests/Main.cpp
```diff
@@ -0,0 +1,17 @@
...
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+int main(int argc, char *argv[])
+{
+    ::testing::InitGoogleTest(&argc, argv);
+    ::testing::InitGoogleMock(&argc, argv);
+    int rc = RUN_ALL_TESTS();
+
+    return rc;
```
arnout, line 16: Why don't we need this for the other unit tests we already have?

source/reactor/src/EventReaction.cpp
```diff
@@ -0,0 +1,55 @@
+#include "EventReaction.h"
+#include "TimeTools.h"
+
+namespace u {
+namespace reactor {
+EventReaction::EventReaction(int fd, long timeout, const EventHandlers &handlers)
+{
+    m_fd       = fd;
+    m_timeout  = timeout;
+    m_handlers = handlers;
```
arnout, line 10: I forgot... Did we agree that this was the way to do it, or rather the C++ way of using initializers?

source/reactor/src/EventReaction.h
```diff
@@ -0,0 +1,86 @@
...
+     * @param timeout Timeout value (in milliseconds).
+     * @param handlers Set of event handlers.
+     */
+    EventReaction(int fd, long timeout, const EventHandlers &handlers);
+
+    /**
+     * Returns file descriptor.
+     * @return File descriptor.
+     */
+    int getFd(void) const;
```
arnout, line 26: Coding style:
```suggestion
    int fd(void) const;
```
(I don't think we say in the coding style to remove the get part, but we should :-)

Same comment in other places as well.

Also, I prefer to have these simple functions to be defined inline.

arnout (approved):
Although I have 33 comments :-) I think this is an excellent contribution.

It should be easy to copy and paste the add, remove and run functions from https://github.com/prplfoundation/prplMesh/pull/1435/commits/a28e6ee7df5aa961fcbc415b61698ea2a1059c8f and use epoll instead. And the unit test will prove that it works!

There are a few comments that really have to be fixed, like the copyright headers and the coding style. Then there is one major feedback on the design, where the timeout handling is not appropriate IMO. That said, I'm OK with merging the timeout handling as is and refactor it later, or maybe it will even turn out to be useful. Therefore, I'm approving already.

source/reactor/CMakeLists.txt
RanRegev: I don't like the `GLOB` option. It seems that even CMake itself doesn't like it:
[Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.](https://cmake.org/cmake/help/latest/command/file.html)

source/reactor/inc/reactor/EventHandlers.h
RanRegev: What does `u` mean? 
I prefer meaningful names.

source/reactor/inc/reactor/EventLoop.h
RanRegev: I think that the only thing that matters is that `indefinitely` **would not** be an option.
Justification: like any other resource in the system, the time must be restricted explicitly.
Having a global timeout is good enough and simplifies development.

source/reactor/inc/reactor/EventLoop.h
RanRegev: Actually you stole my comment... But past experience shows that the following approach is even better:
```suggestion
    template<class Duration>
    virtual bool registerHandlers(int fd, Duration timeout, const EventHandlers &handlers) = 0;
```
So the caller can have the timeout in any `std::chrono` she likes.
(C++20 gives us _restriction_ (predicate) on the Duration type, but C++11 does not)

source/reactor/src/EventReaction.cpp
RanRegev: We should use [member initialization](https://arne-mertz.de/2015/08/new-c-features-default-initializers-for-member-variables/).

source/reactor/inc/reactor/EventHandlers.h
```diff
@@ -0,0 +1,68 @@
...
+using EventHandler = std::function<void(int, EventLoop *)>;
+
+/**
+ * Set of event handler functions, one function to handle each possible event happened.
+ * Handlers are grouped into a struct to facilitate passing them as a single parameter to the install and
+ * remove handlers functions of the event loop.
+ * Event handlers are optional and if not set for a given event, that event will be silently
+ * ignored.
+ */
+struct EventHandlers {
```
RanRegev, line 28: For me Event-Handlers sounds like _many_ handlers for the _same_ event.
For me it is natural to call this: Event**s**-Handler as it is a single handler for few events.

source/reactor/inc/reactor/EventHandlers.h
RanRegev, line 19: Both the int (that should be a `class FD` as @arnout correctly commented) and EventLoop are the _context_ of the function call, the data to be processed. 
Personally I prefer having the context grouped together and not spread as function parameters list. It does lots of good things and gives flexibility.
My OO approach would be therefore to write this class in this way:
```cpp
struct EventContext{
int fd;
EventLoop& originator;
// potentially add more context in the future
};
class EventsHandler{
void on_read(const &EventContext) = 0;
void on_write(const &EventContext) = 0;
// etc.
};
```

If we give up the idea of _replacing_ handlers on the fly, then my template based approach would be:
```cpp
template<class ConcreteEventsHandler> 
class EventLoop{
/* everything */
ConcreteEventsHandler _installedHandler_
};
```

source/reactor/inc/reactor/EventHandlers.h
RanRegev: Yes. I prefer having a context for the function when it is called, some state it can query, some other structure it can grab information from. Not expecting everything to be passed with the parameters.
This can be achieved for example if `on_read()`, `on_write()`, etc. are part of an object that can be initialized with the required data.

RanRegev (approved):
As @arnout said, there are comments but we probably can continue with this implementation and fix, improve or change later.
The separation approach is _good_. It gives us the ability to do develop small units instead of giant monoliths.

vitalybu:
@arnout @RanRegev this PR was created just for reference.
For proposed implementation have a look at: https://github.com/prplfoundation/prplMesh/pull/1435

source/reactor/src/EventReaction.cpp
mariomaz: I also prefer member initialization but I think @morantr won this [match](https://www.strawpoll.me/20160132/r)

source/reactor/CMakeLists.txt
mariomaz: Some header files required to write unit tests are not public but defined in the source folder.

source/reactor/inc/reactor/EventHandlers.h
mariomaz: TIL.
However, I'm not really sure it's worth the effort and I find it much more complex, not clearer

source/reactor/inc/reactor/EventLoop.h
mariomaz: Original [article ](http://www.dre.vanderbilt.edu/~schmidt/PDF/reactor-siemens.pdf)describing reactor pattern includes timeout handlers.

Its usage is optional. It facilitates timeout handling if required.

source/reactor/inc/reactor/EventHandlers.h
mariomaz: Some unit tests rely on it. If we didn't have `on_write`, we should start a thread to send bytes a few milliseconds after calling the run() method of the event loop. That's slow, brittle and ugly.

source/reactor/inc/reactor/EventLoopImpl.h
mariomaz: Before C++11, the usage of virtual keyword was recommended, but not required, to tell the users of the API that such method was virtual without them having to go to base class and check it. Since C++11, override does the same job and also allows the compiler to complain if the code is inadvertently ill-formed. 

So do not use virtual but do use override

source/reactor/src/EventLoopImpl.cpp
mariomaz: Base class here is an interface. By definition, "An interface describes the behavior or capabilities of a C++ class without committing to a particular implementation of that class".

Since the m_eventReactions member is used by a particular implementation of the EventLoop interface, it must be defined in that implementation, not in the interface. 

Also I cannot think of a test where the mock would ever need to *store* installed handlers. I'm sure there will be tests with expectations on which handlers are installed.

source/reactor/unit_tests/EventReaderMock.h
mariomaz: There are two types of EventHandlers in the tests: one used for reading and one for writing. And we need this differentiation: the reader doesn't define the handleWrite function because otherwise EventLoop would fire an event on each loop iteration.

source/reactor/unit_tests/EventHandlersTest.cpp
mariomaz: We don't expect any other methods to be called and in this case, if they were called, it would be an error. This behavior around unexpected calls is "configurable": Google Mock is automatically checking this for us because the mock has been declared as `StrictMock<>`
Please read [The Nice, the Strict, and the Naggy](https://github.com/google/googletest/blob/master/googlemock/docs/cook_book.md#the-nice-the-strict-and-the-naggy-nicestrictnaggy)

source/reactor/unit_tests/EventHandlersTest.cpp
mariomaz: EventLoop behaves differently depending on which handler functions have been defined. For example, if the `onWrite` function has not been defined, then the file descriptor is not added to the `fd_set` handed to the `select()` and no write event is fired.

source/reactor/unit_tests/EventLoopImplTest.cpp
mariomaz: You are right. This test was added with the sole purpose of demonstrating we can use EventLoop to set up timers. I realize now that I should have added a comment explaining it.

source/reactor/unit_tests/Main.cpp
mariomaz: You can use Google Test libraries in different ways: either you link against gtest_main or against gtest. In the first option there is a default `main()` function included. You use the second option when default `main()` doesn't suit your need. This is the case here because we need to init Google Mock as well.

Please see [Writing the main() Function](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#writing-the-main-function)

source/reactor/inc/reactor/EventHandlers.h
mariomaz: Actually they are a set of handlers, one for each different event that might get fired (read, write, error and timeout).

source/reactor/inc/reactor/EventLoop.h
arnout: > Original [article ](http://www.dre.vanderbilt.edu/~schmidt/PDF/reactor-siemens.pdf)describing reactor pattern includes timeout handlers.

It's not because it was originally described like that that it's a good idea :-)

If it is something like CppCoreGuidelines or Boost, which are the result of years of iterative improvements from a large community, I'd be less inclined to go against the "rules". But a single paper from 1995 is nothing more than good inspiration.

None of the event loop approaches I know (Qt, Glib, libev, libevent, twisted, and I'm probably forgetting half a dozen) has this kind of per-file-descriptor timeout. Because it simply makes little sense. It is only useful for one case: in a request-response scenario over a non-multiplexed channel. This is a bit of a corner case, and would be much better served with a specific pattern for that scenario (which, BTW, would be easily extendible to multiplexed channels).

> Its usage is optional.

Sure, but it adds complexity to the implementation, and complexity == bugs.

I have a very strong dislike for dead code, even if it's unit tested.

> It facilitates timeout handling if required.

Again, in one specific use case, and it doesn't serve that case very well.

mariomaz: I've been thinking a lot about setting specific timeouts for each file descriptor since you wrote your first comment and the more I think, the less convinced I am about its suitability. 
I got finally convinced against its use after reviewing #1435. If I had to write this API it again, I would not include timeouts.
A `Timer` wrapper class around `timerfd` would perfectly do the job.

arnout:
This can be closed since #1435 is merged I think?

PR #1450, opened 2020-06-18
Task/socket mockable

vitalii-komisarenko:


common/beerocks/bcl/source/beerocks_socket_thread.cpp
```diff
@@ -170,6 +170,13 @@ bool socket_thread::handle_cmdu_message_uds(Socket *sd)
     if (available_bytes != message_size) {
         THREAD_LOG(ERROR) << "available bytes = " << available_bytes
                           << " message_size = " << message_size;
+        // discard remaining bytes
+        ssize_t bytes_to_discard = message_size - available_bytes;
+        ssize_t bytes_discarded  = sd->discardBytes(bytes_to_discard);
```
morantr, line 175: I don't understand why would we want to do that?
I would expect that if the message is too big, then we read what we can, and let the message handler that receives the data to know that there are more data bytes that will arrive. The handler function should handle it.
Besides that, it is probably will be a better idea to fail the building of a message if it is too big and print an error message, instead of throwing away data on the receiver side.

Anyway, If data is being thrown away, you should add an ERROR print about it.

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -43,28 +43,79 @@ typedef unsigned __int64 ssize_t;
...
+    std::string m_peer_mac;
+    bool m_accepted_socket  = false;
+    bool m_external_handler = false;
+    bool m_is_server        = false;
+
+private:
+    static int m_ref;
+};
+
+class DefaultSocket : public Socket {
```
morantr, line 109: I didn't understand the commit message and the rationale of what you did here.
For a start, I don't understand the use of the word "mock" here.
AFAIK the meaning of it is "tease or laugh at in a scornful or contemptuous manner."
Can you please provide an explanation with an example to why or what it is helpful to create "DefaultSocket" class which is derived from "Socket" class. Why not to use "Socket" directly?

vitalii-komisarenko: The word "mock" is used here in a sence of "a fake object used to test the functionality of other object". Similar to "stub", but probably not the same.
This word is used in the name of GoogleMock framework. The basic idea is to create an abstract class and have two implementation derived from it: the real one and the fake "mocked" one. 
When testing a class that is using sockets we don't want it to fail due to socket issues because we want to test sockets separately. In this case socket user (e.g. beerocks_socket_thread) would recieve a mock to run a unit test.

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -43,28 +43,83 @@ typedef unsigned __int64 ssize_t;
 
 class Socket {
 public:
-    explicit Socket(SOCKET s, long readTimeout = 1000);
-    Socket(SOCKET s, const std::string &peer_ip, int port_port, long readTimeout = 1000);
-    Socket(const std::string &uds_path = std::string(), long readTimeout = 1000);
+    // explicit Socket(SOCKET s, long readTimeout = 1000);
+    // Socket(SOCKET s, const std::string &peer_ip, int port_port, long readTimeout = 1000);
+    // Socket(const std::string &uds_path = std::string(), long readTimeout = 1000);
```
arnout, line 48: Don't leave stuff commented out.

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -43,28 +43,83 @@ typedef unsigned __int64 ssize_t;
...
+                               struct sockaddr_in addr_in = {}) = 0;
+    virtual ssize_t writeString(std::string msg)
     {
         return writeBytes((const uint8_t *)msg.c_str(), (size_t)msg.length());
     }
+    virtual bool setWriteTimeout(long msec) = 0;
+    virtual bool setReadTimeout(long msec)  = 0;
+    virtual void closeSocket()              = 0;
+    virtual bool isOpen()                   = 0;
+    virtual SOCKET getSocketFd()            = 0;
```
arnout, line 73: I personally don't believe that it's necessary to make *all* functions pure virtual to make a mockable class - but @mariomaz please correct me if I'm wrong. I think the parts which are simply "struct-like" can stay. Even the setters - you could claim that it's relevant to define an expectation for a setter, but I don't think that is correct - the expectation should be on whatever other mocked functionality is using that information.

On the other hand, this does require you to think about testing at the time you define the interface, why just making everything pure virtual is more straightforward. And it is definitely better if you don't have to think about everything at once. So maybe I'm wrong in this.

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -43,28 +43,83 @@ typedef unsigned __int64 ssize_t;
 
 class Socket {
 public:
-    explicit Socket(SOCKET s, long readTimeout = 1000);
-    Socket(SOCKET s, const std::string &peer_ip, int port_port, long readTimeout = 1000);
-    Socket(const std::string &uds_path = std::string(), long readTimeout = 1000);
```
arnout, line 48: Since you were able to remove these two constructors without adding a corresponding factory method, I think you should simply remove them in a preparatory commit.

Well actually the last one is still used by the SocketClient and SocketServer classes, but I believe the constructor that takes a peer_ip is unused.

agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
```diff
@@ -209,7 +209,7 @@ void ap_manager_thread::ap_manager_fsm()
         // External events
         int ext_events_fd = ap_wlan_hal->get_ext_events_fd();
         if (ext_events_fd > 0) {
-            ap_hal_ext_events = new Socket(ext_events_fd);
+            ap_hal_ext_events = Socket::socketFactory(ext_events_fd);
```
arnout, line 212: (nitpick) It would be better to have a separate commit for creating the factor method, and a different one for adding the DefaultSocket class.

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -43,28 +43,83 @@ typedef unsigned __int64 ssize_t;
...
+    // explicit Socket(SOCKET s, long readTimeout = 1000);
+    // Socket(SOCKET s, const std::string &peer_ip, int port_port, long readTimeout = 1000);
+    // Socket(const std::string &uds_path = std::string(), long readTimeout = 1000);
+    static Socket *socketFactory(SOCKET s, long readTimeout = 1000);
     virtual ~Socket();
```
arnout, line 50: Does this work?
```suggestion
    virtual ~Socket() = default;
```
(We only want to declare it virtual, we don't want to provide an implementation.)

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -43,28 +43,83 @@ typedef unsigned __int64 ssize_t;
...
+    bool m_accepted_socket  = false;
+    bool m_external_handler = false;
+    bool m_is_server        = false;
+
+private:
+    static int m_ref;
+*/
+};
+
+class DefaultSocket : public Socket {
```
arnout, line 109: I think "DefaultSocket" is not a great name. I like "SocketImpl" similar to what @mariomaz uses in #1449 is more appropriate.
(nitpick)

common/beerocks/bcl/include/bcl/network/socket.h
```diff
@@ -96,19 +151,22 @@ class Socket {
...
+    SOCKET getSocket() const { return m_socket; }
+    void setAcceptedSocket(bool value) { m_accepted_socket = value; }
+    void setUdsPath(const std::string &value) { m_uds_path = value; }
 
 private:
     static int m_ref;
 };
 
-class SocketClient : public Socket {
+class SocketClient : public DefaultSocket {
```
arnout, line 162: This one and SocketServer will eventually have to be made mockable as well... which will bring us in the joys of multiple inheritance :-(

Anyway, better to focus on the easy cases first, so it's good to do it like this for now.

common/beerocks/bcl/source/network/socket.cpp
```diff
@@ -30,9 +30,14 @@ typedef int socklen_t;
 
 #include <easylogging++.h>
 
-int Socket::m_ref = 0;
+int DefaultSocket::m_ref = 0;
```
arnout, line 33: This is only used in the windows code, which could be removed.

arnout (commented):
This is a nice PoC, but I think it will still be hard to actually use it to create a unit test anywhere, because things are still so intermingled in most places. I think a lot more will be needed before we get anywhere.

Therefore, I think it's better to try and identify one place where we want to add a unit test, and drive the refactoring from that.

Also, there is #1449 which doesn't directly conflict with this but there's still some correlation between the two attempts.

common/beerocks/bcl/include/bcl/network/socket.h
mariomaz: Same answer as [here](https://github.com/prplfoundation/prplMesh/pull/1449#discussion_r443612127) 

Interfaces are contracts and IMO they should not contain anything about a particular implementation.

PR #1452, opened 2020-06-18
Dev/replace dwpal parser

morantr:
Draft for replacing dwpal parser with a general key-value parser.

### DIFF
--- a/common/beerocks/bwl/common/key_value_parser.cpp
+++ b/common/beerocks/bwl/common/key_value_parser.cpp
@@ -1,0 +1,224 @@
+
+#include <bwl/key_value_parser.h>
+
+#include <bcl/beerocks_state_machine.h>
+#include <bcl/beerocks_string_utils.h>
+#include <bcl/network/network_utils.h>
+
+#include <easylogging++.h>
+#include <sstream>
+
+namespace bwl {
+constexpr char EVENT_KEYLESS_PARAM_OPCODE[] = "_opcode";
+constexpr char EVENT_KEYLESS_PARAM_MAC[]    = "_mac";
+constexpr char EVENT_KEYLESS_PARAM_IFACE[]  = "_iface";
+
+//////////////////////////////////////////////////////////////////////////////
+////////////////////////// Local Module Functions ////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+/**
+ * @brief Finds first occurrence of two delimiters with string between them, on given string 'str':
+ * <'delim_near'><any_characters except delimiters><'delim_far'>
+ * 
+ * @param[in] str String to find the first occurrence of two delimiters with string between them.  
+ * @param[in] pos Postion of the first characters on 'str' to be considered in the search.
+ * @param[in] delim_near Near delimiter.
+ * @param[in] delim_far Near delimiter.
+ * @return std::string::size_type Postion of the first character that matches. If no matches is 
+ * found, returns string::npos.
+ */
+static std::string::size_type find_first_of_delimiter_pair(const std::string &str,
+                                                           std::string::size_type pos,
+                                                           char delim_near, char delim_far)
+{
+    // First search the end delimiter.
+    auto idx = str.find_first_of(delim_far, pos);
+    if (idx == std::string::npos) {
+        return idx;
+    }
+    // Search back for the start delimiter from there.
+    idx = str.rfind(delim_near, idx);
+    return idx;
+}
+
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Implementation ///////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+void KeyValueParser::parse_line(std::istringstream &iss_in, std::list<char> delimiter_list,
+                                parsed_line_t &parsed_line)
+{
+    if (delimiter_list.empty()) {
+        return;
+    }
+
+    std::string str_storage, key;
+    bool kv = true; // '1'=key, '0'=val
+    while (std::getline(iss_in, str_storage, delimiter_list.front())) {
+        // If reached to 1 delimiter resolution, the data can be parsed into key and value
+        if (delimiter_list.size() == 1) {
+            if (kv) {
+                // Save key
+                key = str_storage;
+            } else {
+                // Save Value
+                parsed_line[key] = str_storage;
+            }
+            kv = !kv;
+
+        } else {
+            auto delimiter_list_out(delimiter_list);
+            // Delete first delimiter
+            delimiter_list_out.erase(delimiter_list_out.begin());
+            std::istringstream iss_out(str_storage);
+            parse_line(iss_out, delimiter_list_out, parsed_line);
+        }
+    }
+}
+
+void KeyValueParser::parse_multiline(std::istringstream &iss_in, std::list<char> delimiter_list,
+                                     parsed_multiline_t &parsed_multiline)
+{
+    if (delimiter_list.empty()) {
+        return;
+    }
+
+    std::string str_storage;
+    while (std::getline(iss_in, str_storage, delimiter_list.front())) {
+        auto delimiter_list_out(delimiter_list);
+
+        // Delete first delimiter
+        delimiter_list_out.erase(delimiter_list_out.begin());
+        std::istringstream iss_out(str_storage);
+        parsed_line_t parsed_line;
+        parse_line(iss_out, delimiter_list_out, parsed_line);
+        parsed_multiline.push_back(parsed_line);
+    }
+}
+
+void KeyValueParser::parse_event(std::string event_str, parsed_line_t &parsed_line)
+{
+    // Eliminate event log level from the begining of the event string : "<3>".
+    size_t idx_start = event_buffer_start_process_idx(event_str);
+
+    // An event string might start with keyless values such as Even Name.
+    // Find where the keyless parameters ends by finding where is the first parameter with key,
+    // assuming it has a preceding " <SomeKeyString>='".
+    auto idx = find_first_of_delimiter_pair(event_str, idx_start, ' ', '=');
+
+    // Put null terminator at the end of our key=val, for string stream construction.
+    if (idx != std::string::npos) {
+        event_str[idx] = '\0';
+    }
+
+    // Parse keyless params part and insert to the 'parsed_line' map object.
+    parse_event_keyless_params(event_str, idx_start, parsed_line);
+
+    // Fill the map with the rest of event data.
+    while (idx != std::string::npos) {
+
+        idx_start = ++idx;
+
+        // Find first '=' to skip on it
+        idx = event_str.find_first_of('=', idx_start);
+
+        // Find the next pair of delimiters index.
+        idx = find_first_of_delimiter_pair(event_str, ++idx, ' ', '=');
+
+        if (idx != std::string::npos) {
+            // Put null terminator at the end of our key=val, for ss_in construction.
+            event_str[idx] = '\0';
+        }
+
+        // Parse key=value
+        std::istringstream iss_in(event_str.c_str() + idx_start);
+        parse_line(iss_in, {'='}, parsed_line);
+    }
+}
+
+size_t KeyValueParser::event_buffer_start_process_idx(const std::string &event_str)
+{
+    // Eliminate event log level from the begining of the event string : "<3>"
+    auto idx_start = event_str.find_first_of(">");
+    if (idx_start != std::string::npos) {
+        idx_start++;
+    } else {
+        LOG(WARNING) << "Empty event! event_string: " << event_str;
+    }
+    return idx_start;
+}
+
+void KeyValueParser::parse_event_keyless_params(const std::string &event_str, size_t idx_start,
+                                                parsed_line_t &parsed_line)
+{
+    // Insert to map known prams without key
+    std::istringstream iss(event_str.c_str() + idx_start);
+    std::string str_storage;
+    bool opcode = true;
+    while (std::getline(iss, str_storage, ' ')) {
+        if (opcode) {
+            // assume that the second param is data or event name
+            parsed_line[EVENT_KEYLESS_PARAM_OPCODE] = str_storage;
+            opcode                                  = false;
+        } else if (beerocks::net::network_utils::is_valid_mac(str_storage)) {
+            parsed_line[EVENT_KEYLESS_PARAM_MAC] = str_storage;
+        } else if (!strncmp(str_storage.c_str(), "wlan", 4)) {
+            parsed_line[EVENT_KEYLESS_PARAM_IFACE] = str_storage;
+        }
+    }
+}
+
+bool KeyValueParser::read_param(const std::string &key, parsed_line_t &obj, int64_t &value,
+                                bool ignore_unknown)
+{
+    auto val_iter = obj.find(key);
+    if (val_iter == obj.end()) {
+        LOG(ERROR) << "param :" << key << " does not exist";
+        return false;
+    }
+
+    static const std::string unknown_string = "UNKNOWN";
+
+    if (ignore_unknown && !unknown_string.compare(val_iter->second)) {
+        value = 0;
+        return true;
+    }
+
+    value = beerocks::string_utils::stoi(val_iter->second);
+
+    return true;
+}
+
+bool KeyValueParser::read_param(const std::string &key, parsed_line_t &obj, const char **value)
+{
+    auto val_iter = obj.find(key);
+    if (val_iter == obj.end()) {
+        LOG(ERROR) << "param :" << key << " does not exist";
+        return false;
+    }
+
+    *value = reinterpret_cast<const char *>((val_iter->second).c_str());
+    return true;
+}
+
+void KeyValueParser::parsed_obj_debug(const parsed_line_t &obj)
+{
+    LOG(DEBUG) << std::endl << "parsed_obj_debug: ";
+    for (const auto &kv_element : obj) {
+        LOG(DEBUG) << "key: " << kv_element.first << ", value: " << kv_element.second;
+    }
+}
+
+void KeyValueParser::parsed_obj_debug(const parsed_multiline_t &obj)
+{
+    LOG(DEBUG) << std::endl << "parsed_obj_debug: ";
+    int element_num = 0;
+    for (const auto &list_element : obj) {
+        LOG(DEBUG) << "vector element: " << element_num;
+        parsed_obj_debug(list_element);
+        element_num++;
+    }
+}
+
+} // namespace bwl
--- a/common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
@@ -442,7 +442,7 @@ bool base_wlan_hal_dwpal::set(const std::string &param, const std::string &value
 {
     const std::string cmd = "SET " + param + " " + value;
     if (!dwpal_send_cmd(cmd, vap_id)) {
-        LOG(ERROR) << "FAiled setting param " << param;
+        LOG(ERROR) << "Failed setting param " << param;
         return false;
     }
 
@@ -453,6 +453,37 @@ bool base_wlan_hal_dwpal::ping()
 {
     if (!dwpal_send_cmd("PING")) {
         return false;
+    }
+
+    return true;
+}
+
+bool base_wlan_hal_dwpal::dwpal_send_cmd(const std::string &cmd, parsed_line_t &reply, int vap_id)
+{
+    if (!dwpal_send_cmd(cmd, vap_id)) {
+        return false;
+    }
+
+    if (m_wpa_ctrl_buffer[0] != '\0') {
+        // if reply is not empty
+        std::istringstream iss_in(m_wpa_ctrl_buffer);
+        parse_line(iss_in, {'\n', '='}, reply);
+    }
+
+    return true;
+}
+
+bool base_wlan_hal_dwpal::dwpal_send_cmd(const std::string &cmd, parsed_multiline_t &reply,
+                                         int vap_id)
+{
+    if (!dwpal_send_cmd(cmd, vap_id)) {
+        return false;
+    }
+
+    if (m_wpa_ctrl_buffer[0] != '\0') {
+        // if reply is not empty
+        std::istringstream iss_in(m_wpa_ctrl_buffer);
+        parse_multiline(iss_in, {'\n', ' ', '='}, reply);
     }
 
     return true;
--- a/common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
+++ b/common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
@@ -12,6 +12,7 @@
 #include "base_wlan_hal_dwpal_types.h"
 
 #include <bwl/base_wlan_hal.h>
+#include <bwl/key_value_parser.h>
 #include <bwl/nl80211_client.h>
 
 #include <bcl/beerocks_state_machine.h>
@@ -35,7 +36,8 @@ enum class dwpal_fsm_event { Attach, Detach };
  * Read more about virtual inheritance: https://en.wikipedia.org/wiki/Virtual_inheritance
  */
 class base_wlan_hal_dwpal : public virtual base_wlan_hal,
-                            protected beerocks::beerocks_fsm<dwpal_fsm_state, dwpal_fsm_event> {
+                            protected beerocks::beerocks_fsm<dwpal_fsm_state, dwpal_fsm_event>,
+                            public KeyValueParser {
 
     // Public methods:
 public:
@@ -77,8 +79,16 @@ protected:
     bool set(const std::string &param, const std::string &value,
              int vap_id = beerocks::IFACE_RADIO_ID);
 
+    bool dwpal_send_cmd(const std::string &cmd, parsed_line_t &reply,
+                        int vap_id = beerocks::IFACE_RADIO_ID);
+
+    bool dwpal_send_cmd(const std::string &cmd, parsed_multiline_t &reply,
+                        int vap_id = beerocks::IFACE_RADIO_ID);
+
+    // for external process
     bool dwpal_send_cmd(const std::string &cmd, char **reply,
-                        int vap_id = beerocks::IFACE_RADIO_ID); // for external process
+                        int vap_id = beerocks::IFACE_RADIO_ID);
+
     bool dwpal_send_cmd(const std::string &cmd, int vap_id = beerocks::IFACE_RADIO_ID);
     bool attach_ctrl_interface(int vap_id);
 
--- a/common/beerocks/bwl/include/bwl/key_value_parser.h
+++ b/common/beerocks/bwl/include/bwl/key_value_parser.h
@@ -1,0 +1,66 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * Copyright (c) 2016-2019 Intel Corporation
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _BWL_KEY_VALUE_PARSER_H_
+#define _BWL_KEY_VALUE_PARSER_H_
+
+#include <list>
+#include <string>
+#include <unordered_map>
+
+namespace bwl {
+
+/**
+ * @brief On KeyValueParser there are two types of object that can be parsed:
+ * parsed_line_t - Shall be used on key-value data with unique keys for each element, seperated by
+ * a space or a new line (Default seperator, can be modified on use). For example:
+ * "key1=value1 key2=value2 key3=value3 ..."
+ * --Or--
+ * "key1=value1
+ *  key2=value2
+ *  key3=value3
+ *  ..."
+ * 
+ * parsed_multiline_t - Shall be used in case there are 2 or more "parsed_line_t" which each "line"
+ * could have the same keys as the previous one. For example: 
+ * "key1=Line1value1 key2=Line1value2 key3=Line1value3 ...
+ *  key1=Line2value1 key2=Line2value2 key3=Line2value3 ...
+ *  key1=Line3value1 key2=Line3value2 key3=Line3value3 ...
+ *  ...""
+ */
+
+typedef std::unordered_map<std::string, std::string> parsed_line_t;
+typedef std::list<parsed_line_t> parsed_multiline_t;
+
+class KeyValueParser {
+protected:
+    static void parse_line(std::istringstream &iss, std::list<char> delimiter_list,
+                           parsed_line_t &parsed_line);
+
+    static void parse_multiline(std::istringstream &iss, std::list<char> delimiter_list,
+                                parsed_multiline_t &parsed_multiline);
+
+    static bool read_param(const std::string &key, parsed_line_t &obj, int64_t &value,
+                           bool ignore_unknown = false);
+
+    static bool read_param(const std::string &key, parsed_line_t &obj, const char **value);
+
+    static void parsed_obj_debug(const parsed_line_t &obj);
+    static void parsed_obj_debug(const parsed_multiline_t &obj);
+
+    virtual void parse_event(std::string event_str, parsed_line_t &parsed_line);
+
+    virtual size_t event_buffer_start_process_idx(const std::string &event_str);
+
+    virtual void parse_event_keyless_params(const std::string &event_str, size_t idx_start,
+                                            parsed_line_t &parsed_line);
+};
+
+} // namespace bwl
+
+#endif // _BWL_KEY_VALUE_PARSER_H_


common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include <bwl/key_value_parser.h>
+
+#include <bcl/beerocks_state_machine.h>
+#include <bcl/beerocks_string_utils.h>
+#include <bcl/network/network_utils.h>
+
+#include <easylogging++.h>
+#include <sstream>
+
+using namespace bwl;
```
arnout: Just put the entire file in bwl namespace rather than using it.

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+namespace bwl {
+constexpr char EVENT_KEYLESS_PARAM_OPCODE[] = "_opcode";
```
arnout, line 12: Coding style:
```suggestion
constexpr char event_keyless_param_opcode[] = "_opcode";
```

However:
- probably this should be a std::string instead of char[];
- I would make this stuff parameters of the constructor rather than hardcoding.

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+ * <'delim_near'><any_characters except delimiters><'delim_far'>
+ * 
+ * @param[in] str String to find the first occurrence of two delimiters with string between them.  
+ * @param[in] pos Postion of the first characters on 'str' to be considered in the search.
+ * @param[in] delim_near Near delimiter.
+ * @param[in] delim_far Near delimiter.
+ * @return std::string::size_type Postion of the first character that matches. If no matches is 
+ * found, returns string::npos.
+ */
+static std::string::size_type find_first_of_delimiter_pair(std::string &str,
```
arnout:
```suggestion
static std::string::size_type find_first_of_delimiter_pair(const std::string &str,
```

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+/**
+ * @brief Finds first occurrence of two delimiters with string between them, on given string 'str':
+ * <'delim_near'><any_characters except delimiters><'delim_far'>
+ * 
+ * @param[in] str String to find the first occurrence of two delimiters with string between them.  
+ * @param[in] pos Postion of the first characters on 'str' to be considered in the search.
+ * @param[in] delim_near Near delimiter.
+ * @param[in] delim_far Near delimiter.
+ * @return std::string::size_type Postion of the first character that matches. If no matches is 
+ * found, returns string::npos.
```
arnout, line 29: I don't particularly like this interface - I think it would be better to just return a std::string with the stuff you're looking for. Yes, it's marginally less efficient to copy the strings, but I think that's a case of premature optimisation. The natural thing would be to use a string_view, but that's C++17 :-(

Anyway, I'm OK with merging as is. It's still progress.

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+ */
+static std::string::size_type find_first_of_delimiter_pair(std::string &str,
+                                                           std::string::size_type pos,
+                                                           char delim_near, char delim_far)
+{
+    // Finds first occurrence on string 'str' of
+    // <'delim_near'><any_characters except delimiters><'delim_far'>
+    auto idx = str.find_first_of(delim_far, pos);
```
arnout: This just repeats the doxygen comment so redundant.

It would be more useful to say something like "First search the end delimiter, and search back for the start delimiter from there. This makes sure that start delimiter doesn't occur between start and end."

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+    }
+    idx = str.rfind(delim_near, idx);
+    return idx;
+}
+
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Implementation ///////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+void KeyValueParser::parse_line(std::stringstream &ss_in, std::list<char> delimiter_list,
```
arnout: This could probably be
```suggestion
void KeyValueParser::parse_line(std::istream &in, std::list<char> delimiter_list,
```
there's no need to specialize to stringstream.

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+                                parsed_line_t &parsed_line)
+{
+    if (delimiter_list.empty()) {
+        return;
+    }
+
+    std::string str_storage, key;
+    bool kv = true; // '1'=key, '0'=val
+    while (std::getline(ss_in, str_storage, delimiter_list.front())) {
```
arnout: I find this recursive parse_line implementation with a delimiter list really hard to understand, for no discernable purpose... It would make a lot more sense to me to have something like this:
```
lines = str_split(reply, "\n");
for (line : lines) {
    auto idx = argument.find('=');
    if (idx != std::string::npos) {
        // store key/value
    } else {
       // store key only
    }
}
```

And to reuse it in a multiline, just move out the body that parses a single key-value into a helper function.

It could be even better with std::istream_iterator and algorithms, to avoid the copying into vectors. But: premature optimisation.

That said, it's a big change, and the current code works, so I can accept the current code.

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+    while (std::getline(ss, str_storage, ' ')) {
+        if (opcode) {
+            // assume that the second param is data or event name
+            parsed_line[EVENT_KEYLESS_PARAM_OPCODE] = str_storage;
+            opcode                                  = false;
+        } else if (beerocks::net::network_utils::is_valid_mac(str_storage)) {
+            parsed_line[EVENT_KEYLESS_PARAM_MAC] = str_storage;
+        } else if (!strncmp(str_storage.c_str(), "wlan", 4)) {
+            parsed_line[EVENT_KEYLESS_PARAM_IFACE] = str_storage;
+        }
```
arnout, line 168: I've always hated this abomination as well...

It would make a whole lot more sense to me if parsed_line would have both a vector of keyless parameters, and a map of key-value parameters. And I don't think it makes any sense at all to duck-type the MAC address: either you expect a MAC address in that position and you parse it as such, or you expect something else and then you *should* parse it as a MAC address. Who knows, it might e.g. be an SSID that happens to be formatted as a MAC address...

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+        LOG(WARNING) << "Empty event! event_string: " << event_str;
+    }
+    return idx_start;
+}
+
+void KeyValueParser::parse_event_keyless_params(const std::string &event_str, size_t idx_start,
+                                                parsed_line_t &parsed_line)
+{
+    // Insert to map known prams without key
+    std::stringstream ss(event_str.c_str() + idx_start);
```
arnout: This is a bit icky to say the least...

We could actually very easily parse the entire event (including start thing) with a relatively simple istream construct:

```
std::string dummy;
std::getline(in, dummy, ">");
// We could check here that dummy starts with < and is an integer
```

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+            parsed_line[EVENT_KEYLESS_PARAM_MAC] = str_storage;
+        } else if (!strncmp(str_storage.c_str(), "wlan", 4)) {
+            parsed_line[EVENT_KEYLESS_PARAM_IFACE] = str_storage;
+        }
+    }
+}
+
+bool KeyValueParser::read_param_int(const std::string &key, parsed_line_t &obj, int64_t &value,
+                                    bool ignore_unknown)
+{
```
arnout: This would be better as
```
bool parsed_line_t::get_param(const std::string &key, int64_t &value, bool ignore_unknown);
```
(when we have this object-oriented stuff, we could just as well use it :-P

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,225 @@
...
+    auto val_iter = obj.find(key);
+    if (val_iter == obj.end()) {
+        LOG(ERROR) << "param :" << key << " is not exist";
```
arnout:
```suggestion
        LOG(ERROR) << "param :" << key << " does not exist";
```

common/beerocks/bwl/include/bwl/key_value_parser.h
```diff
@@ -0,0 +1,66 @@
...
+ * parsed_multiline_t - Shall be used in case there are 2 or more "parsed_line_t" which each "line"
+ * could have the same keys as the previous one. For example: 
+ * "key1=Line1value1 key2=Line1value2 key3=Line1value3 ...
+ *  key1=Line2value1 key2=Line2value2 key3=Line2value3 ...
+ *  key1=Line3value1 key2=Line3value2 key3=Line3value3 ...
+ *  ...""
+ */
+
+typedef std::unordered_map<std::string, std::string> parsed_line_t;
+typedef std::list<parsed_line_t> parsed_multiline_t;
```
arnout, line 38: IMO both of them should be a full class.

And multiline should be a vector, not a list.

common/beerocks/bwl/include/bwl/key_value_parser.h
```diff
@@ -0,0 +1,66 @@
...
+
+class KeyValueParser {
+protected:
```
arnout, line 41: There is no reason to make this protected. See my comment in next commit.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
```diff
@@ -35,7 +36,8 @@ enum class dwpal_fsm_event { Attach, Detach };
  * Read more about virtual inheritance: https://en.wikipedia.org/wiki/Virtual_inheritance
  */
 class base_wlan_hal_dwpal : public virtual base_wlan_hal,
-                            protected beerocks::beerocks_fsm<dwpal_fsm_state, dwpal_fsm_event> {
+                            protected beerocks::beerocks_fsm<dwpal_fsm_state, dwpal_fsm_event>,
+                            public KeyValueParser {
```
arnout, line 40: There is no reason to derive from KeyValueParser, is there?

In fact, there's no reason for KeyValueParser to even be a class. It would make a lot more sense for parsed_line_t and parsed_multiline_t to be classes with a static parse() member and private constructor, and to hide all the internal stuff as static functions in the cpp file.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
```diff
@@ -458,6 +458,37 @@ bool base_wlan_hal_dwpal::ping()
     return true;
 }
 
+bool base_wlan_hal_dwpal::dwpal_send_cmd(const std::string &cmd, parsed_line_t &reply, int vap_id)
+{
+    if (!dwpal_send_cmd(cmd, vap_id)) {
+        return false;
+    }
+
+    if (strncmp(m_wpa_ctrl_buffer, "", 1)) {
```
arnout:
```suggestion
    if (m_wpa_ctrl_buffer[0] != '\0') {
```
(at least, for me that's easier to read)

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -677,94 +677,93 @@ bool mon_wlan_hal_dwpal::update_vap_stats(const std::string &vap_iface_name, SVa
...
-    calc_curr_traffic(BytesReceived, sta_stats.rx_bytes_cnt, sta_stats.rx_bytes);
-    calc_curr_traffic(PacketsSent, sta_stats.tx_packets_cnt, sta_stats.tx_packets);
-    calc_curr_traffic(PacketsReceived, sta_stats.rx_packets_cnt, sta_stats.rx_packets);
+    // Last Downlink (TX) Rate
+    if (!read_param_int("LastDataDownlinkRate", reply, tmp_int)) {
+        LOG(ERROR) << "Failed reading LastDataDownlinkRate parameter!";
+        return false;
+    }
+
+    sta_stats.tx_phy_rate_100kb = (uint16_t)(tmp_int / double(100));
```
arnout: It originally used integer division rather than floating-point division. Why change it? At least it should be in a separate commit that explains why.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -677,94 +677,93 @@ bool mon_wlan_hal_dwpal::update_vap_stats(const std::string &vap_iface_name, SVa
...
+        return false;
+    }
+
+    calc_curr_traffic(tmp_int, sta_stats.rx_bytes_cnt, sta_stats.rx_bytes);
+
+    // TX Packets
+    if (!read_param_int("PacketsSent", reply, tmp_int)) {
+        LOG(ERROR) << "Failed reading PacketsSent parameter!";
+        return false;
+        return false;
```
arnout:
```suggestion
        return false;
```

arnout (approved):
As usual I have a shitload of suggestions about how to improve it. But even so, I think this is already infinitely better than the original, so I'm all for merging it and improving afterwards.

common/beerocks/bwl/include/bwl/key_value_parser.h
morantr: No, it shouldn't be a vector. It could get quite big, and reallocations are definitely something we would like to avoid here.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: Why do you hate 'using namespace' so much?

common/beerocks/bwl/common/key_value_parser.cpp
morantr: - It is not possible to define `std::string` as `constexpr`.
- I used capital letters since constexpr is like a Macro (CPP way).

common/beerocks/bwl/common/key_value_parser.cpp
morantr: `str_split` is parsing the data by copying it to a vector which does reallocations, and then the vector is copied to the return value. Very wasteful function. I don't want to use it.
Second, it is recursive because it needs to handle different delimiters list which can be the non default one.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: done.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: done.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: Hmm yes, we can get a little better performance if we use `istringstream` (input only string stream).
`istream` is something we cannot use because it can't be constructed from c-string.
Anyway, replaced to `istringstream`.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: I preferred to use `str::string find_first_of()` because the difference between this and what you suggested is a construction  of an additional string (and also copying the string into it).

common/beerocks/bwl/common/key_value_parser.cpp
morantr: I'm sorry that you don't like this "abomination", but this solution is designed to be a single function that handles whatever output of hostapd bring to us.
Believe me, I would be happy if all of hostapd output was ordered on "key=value.." and I wouldn't have to do it. 

If you think you have a better implementation, feel free to change it. Just make sure it supports all of the hostapd output events.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: What? `parsed_line_t` is an `unordered_map`. C++ doesn't support extension functions.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: ok

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
morantr: Everything (well, most of the things) is possible. I choose to do it that way. It looks simple and convenient to use for me.
I did it as a class since I assumed that other bwl types could use it, so part of the functions could be overridden.

Again, if you have a different vision of how it should look like, feel free to change it.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.cpp
morantr: sure.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
morantr: I copied it from the old implementation of bwl-wav.
I'll change it.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
morantr: fixed.

morantr:
Removed the replacement of "get_sta_measuremet" to kv parser, and move it to a different PR.

common/beerocks/bwl/common/key_value_parser.cpp
arnout: > Why do you hate 'using namespace' so much?

This particular one is confusing: everything we are *defining* in this file *is* actually in the bwl namespace. I think it's a lot clearer to have
```
namespace foo {
void xxx() {} // This defines foo::xxx
}
```
than
```
using namespace foo;

void xxx() {} // This defines ::xxx() unless there was a preceding declaration of foo::xxx()
```

Second, in general I don't like `using namespace` because it basically bypasses the namespace. Namespaces are meant to split things up in modules so you can clearly see to which module each thing belongs. `using namespace` throws them back into a single heap, so then why did you put them into two separate namespaces to begin with?

I don't consider it that important, but I *will* consistently comment about it :-)

morantr: Since it does not matter to be that much. I'll change it to how you want. 🙂

common/beerocks/bwl/common/key_value_parser.cpp
yaniv-yitzhak: There are more keyless parameters other than MAC and interfaces, such as "freq'" in GET_FAILSAFE_CHAN event.
Look in ap/mon_wlan_hal_dwpal.cpp for all events with:
        {(void *)freq, &numOfValidArgs[0], DWPAL_STR_PARAM, **NULL**, sizeof(freq)},

NULL means keyless param, so it's not generic enough, need to think on another solution, you can look in dwpal parser to see how to identify them.

In addition, interface names can be other than "wlanX", in some platforms it can be "wlX" etc., so again, it's not generic enough.

common/beerocks/bwl/include/bwl/key_value_parser.h
vitalybu, line 38: Can't you replace those two with a single `std::unordered_multimap`?

common/beerocks/bwl/common/key_value_parser.cpp
```diff
@@ -0,0 +1,224 @@
...
+    std::string str_storage;
+    bool opcode = true;
+    while (std::getline(iss, str_storage, ' ')) {
+        if (opcode) {
+            // assume that the second param is data or event name
+            parsed_line[EVENT_KEYLESS_PARAM_OPCODE] = str_storage;
+            opcode                                  = false;
+        } else if (beerocks::net::network_utils::is_valid_mac(str_storage)) {
+            parsed_line[EVENT_KEYLESS_PARAM_MAC] = str_storage;
+        } else if (!strncmp(str_storage.c_str(), "wlan", 4)) {
```
vitalybu, line 166: Why is "wlan" hard-coded?
If this parser supposed to be generic, it should probably allow controlling the interface prefix..

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
vitalybu: I somewhat agree with @arnout here, there's already too much inheritance going on here.
It's fine to keep the parser as a class, but it should be a member of the BWL class.

vitalybu (approved):

kantera800 (commented):
Did a partial review 
IMO  its not generic enough.
We could have a problem  parsing  string parameter  with spaces for example, 
 so delimiter types should not be hardcoded.
Also @yaniv-yitzhak  has a point  [here ](https://github.com/prplfoundation/prplMesh/pull/1452/commits/2c345606200f2881e04c6d3f6a57ab5706bbd95d#r448793907)
I see that there is already 2 approves so adding don't merge tag for now just to make sure we don't merge it by mistake before clarification that.

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
morantr: Maybe, but the dummy also using it, and it will need to override some of the functions. That is why it has to be with inheritance.
Other bwls will might need to override too.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: GET_FAILSAFE_CHAN is on the ap_dwapl and not the mon, but anyway it is not an "event" and therefore will not be parsed as an event on the code you gave the comments on.
In that case, there is nothing that can be done, except parsing it manually, and that what I did on the bwl_wav.
"multiap_common/browse/beerocks/bwl/wav/ap_wlan_hal_wav.cpp:
```cpp
bool ap_wlan_hal_wav::failsafe_channel_get(int &chan, int &bw)
{
    LOG(TRACE) << __func__;

    // Build command string
    std::string cmd = "GET_FAILSAFE_CHAN";
    char *reply;

    // Send command
    if (!wpa_ctrl_send_msg(cmd, &reply)) {
        LOG(ERROR) << "failsafe_channel_get() failed!";
        return false;
    }

    // Custom reply parsing
    if (!strncmp(reply, "UNSPECIFIED", 11)) {
        chan = -1;
        bw   = -1;
    } else if (!strncmp(reply, "ACS", 3)) {
        chan = bw = 0;

    } else {
        int freq;
        sscanf(reply, "%d %*s %*s bandwidth=%d", &freq, &bw);
        chan = beerocks::utils::wifi_freq_to_channel(freq);
    }

    return true;
}
```

This solution is not and can't be 100% proof. It was deisgned to handle the majority of cases, and the tiny part it could not handle has to be handled manually.
If it was 100% successful, it probably looked like the dwpal parser, which is something we definitely don't want.

common/beerocks/bwl/include/bwl/key_value_parser.h
morantr: No. Some of the hostapd dumps are repeating the same keys (ACS report for example).
So I have to create a list of  `std::unordered_multimap`.

common/beerocks/bwl/common/key_value_parser.cpp
morantr: Hmm, basically it is "wlan" because it was designed originally to our platform. But sure, I change it to use custom iface prefix.

morantr:
> Did a partial review
> IMO its not generic enough.
> We could have a problem parsing string parameter with spaces for example,
> so delimiter types should not be hardcoded.
> Also @yaniv-yitzhak has a point [here ](https://github.com/prplfoundation/prplMesh/pull/1452/commits/2c345606200f2881e04c6d3f6a57ab5706bbd95d#r448793907)
> I see that there is already 2 approves so adding don't merge tag for now just to make sure we don't merge it by mistake before clarification that.

The delimiters are not hardcoded, they are default and could be changed by giving a different list to the parsing function.
Second, it handles perfectly values with spaces in them. You probably mean to "key1=val1 key2=%d %d %d %d key3=val3"/
In that case, trying to get key2s value as string output, you will get "%d %d %d %d" and that is as something that can be parsed manually.

Again, if you guys expect that it would do everything like returning a vector of ints, it is just not what it designed for. Otherwise, it would be the dawpal parser.

common/beerocks/bwl/common/key_value_parser.cpp
yaniv-yitzhak: Let's say I'm a new developer, adding new event\cpmmand reply or anything to parse, how should I know if the parser fits my needs?
Probably I will look at a different event as an example, and see the parser being used, so I will use it as well right?

We can't rely on that developers should know how the parser works (as also now, not all of us know how dwpal parser works), and it's not a good approach sometimes to use the parser and sometimes not, it should be unified.

And if the dwpal parser can be generic enough, what's so wrong with it? 
As I see it now, if the C++ parser will not be unified for ALL parsing and will not be generic enough, it's not better than the dwpal parser!

yaniv-yitzhak (changes requested):
Let's say I'm a new developer, adding new event\command reply or anything to parse, how should I know if the parser fits my needs?
Probably I will look at a different event as an example, and see the parser being used, so I will use it as well right?

We can't rely on that developers should know how the parser works (also now, not all of us know how dwpal parser works), and it's not a good approach sometimes to use the parser and sometimes not, it should be unified.

And if the dwpal parser can be generic enough, what's so wrong with it?
As I see it now, if the C++ parser will not be unified for ALL parsing and will not be generic enough, it's not better than the dwpal parser!

common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h
vitalybu: Why? It's a helper class, why would you need to override it's functionality?

common/beerocks/bwl/include/bwl/key_value_parser.h
vitalybu: Isn't the idea of `std::unordered_multimap` is to be able to multiple elements for the same key?

morantr:
> Let's say I'm a new developer, adding new event\command reply or anything to parse, how should I know if the parser fits my needs?
> Probably I will look at a different event as an example, and see the parser being used, so I will use it as well right?
> 
> We can't rely on that developers should know how the parser works (also now, not all of us know how dwpal parser works), and it's not a good approach sometimes to use the parser and sometimes not, it should be unified.
> 
> And if the dwpal parser can be generic enough, what's so wrong with it?
> As I see it now, if the C++ parser will not be unified for ALL parsing and will not be generic enough, it's not better than the dwpal parser!

You have to know how the output of hostapd looks like in order to parse something, whether we will use dwpal parser or the kv parser, correct? 
So if you are a new developer and you expect to read something from the reply that it does not have a key, you will see that it cannot be done, and you will parse it manually.

So why it is better to use the kv-parser?
1. It has only effective ~150 lines of code compare to ~1000 on dwpal. That makes it easy to debug and maintain.
2. It is more generic, and has only two kinds of outputs: int and string, compare to dwpal that has:
```cpp
        DWPAL_STR_PARAM = 0,
	DWPAL_STR_ARRAY_PARAM,  /* Note: the output param for this type MUST be an array of strings with a length of 
        HOSTAPD_TO_DWPAL_VALUE_STRING_LENGTH, i.e. "char non_pref_chan[32][HOSTAPD_TO_DWPAL_VALUE_STRING_LENGTH];" */
	DWPAL_CHAR_PARAM,
	DWPAL_UNSIGNED_CHAR_PARAM,
	DWPAL_SHORT_INT_PARAM,
	DWPAL_INT_PARAM,
	DWPAL_UNSIGNED_INT_PARAM,
	DWPAL_LONG_LONG_INT_PARAM,
	DWPAL_UNSIGNED_LONG_LONG_INT_PARAM,
	DWPAL_INT_ARRAY_PARAM,
	DWPAL_INT_HEX_PARAM,
	DWPAL_INT_HEX_ARRAY_PARAM,
	DWPAL_BOOL_PARAM
```
3. It use only safe cpp core code which is safer, and does not allocate memory with c function, compare to dwpal that god knows 
if it frees all the memory it is allocating there. So less chances you get crashes memory leaks and etc.

4. The parsing interface is much cleaner and readable.
kv-parser:
```cpp
 // Last Downlink (TX) Rate
    if (!read_param_int("LastDataDownlinkRate", reply, tmp_int)) {
        LOG(ERROR) << "Failed reading LastDataDownlinkRate parameter!";
        return false;
    }

    sta_stats.tx_phy_rate_100kb = (tmp_int / 100);

    // Last Uplink (RX) Rate
    if (!read_param_int("LastDataUplinkRate", reply, tmp_int)) {
        LOG(ERROR) << "Failed reading LastDataUplinkRate parameter!";
        return false;
    }

    sta_stats.rx_phy_rate_100kb = (tmp_int / 100);

    // TX Bytes
    if (!read_param_int("BytesSent", reply, tmp_int)) {
        LOG(ERROR) << "Failed reading BytesSent parameter!";
        return false;
    }

    calc_curr_traffic(tmp_int, sta_stats.tx_bytes_cnt, sta_stats.tx_bytes);

    // RX Bytes
    if (!read_param_int("BytesReceived", reply, tmp_int)) {
        LOG(ERROR) << "Failed reading BytesReceived parameter!";
        return false;
    }

    calc_curr_traffic(tmp_int, sta_stats.rx_bytes_cnt, sta_stats.rx_bytes);

    // TX Packets
    if (!read_param_int("PacketsSent", reply, tmp_int)) {
        LOG(ERROR) << "Failed reading PacketsSent parameter!";
        return false;
    }

    calc_curr_traffic(tmp_int, sta_stats.rx_packets_cnt, sta_stats.rx_packets);

    // Retranmission Count
    if (!read_param_int("RetransCount", reply, tmp_int, true)) {
        LOG(ERROR) << "Failed reading RetransCount parameter!";
        return false;
    }
```
dwpal parser:
```cpp
    size_t numOfValidArgs[10] = {0}, replyLen = strnlen(reply, HOSTAPD_TO_DWPAL_MSG_LENGTH);
    uint64_t BytesSent = 0, BytesReceived = 0, PacketsSent = 0, PacketsReceived = 0,
             LastDataDownlinkRate = 0, LastDataUplinkRate = 0, Active = 0;
    char ShortTermRSSIAverage[32] = {'\0'};
    char SNR[32]                  = {'\0'};
        {(void *)&BytesSent, &numOfValidArgs[0], DWPAL_LONG_LONG_INT_PARAM, "BytesSent=", 0},
        {(void *)&BytesReceived, &numOfValidArgs[1], DWPAL_LONG_LONG_INT_PARAM,
         "BytesReceived=", 0},
        {(void *)&PacketsSent, &numOfValidArgs[2], DWPAL_LONG_LONG_INT_PARAM, "PacketsSent=", 0},
        {(void *)&PacketsReceived, &numOfValidArgs[3], DWPAL_LONG_LONG_INT_PARAM,
         "PacketsReceived=", 0},
        {(void *)&sta_stats.retrans_count, &numOfValidArgs[4], DWPAL_INT_PARAM, "RetransCount=", 0},
        {(void *)ShortTermRSSIAverage, &numOfValidArgs[5], DWPAL_STR_PARAM,
         "ShortTermRSSIAverage=", sizeof(ShortTermRSSIAverage)},
        {(void *)SNR, &numOfValidArgs[6], DWPAL_STR_PARAM, "SNR=", sizeof(SNR)},
        {(void *)&Active, &numOfValidArgs[7], DWPAL_LONG_LONG_INT_PARAM, "Active=", 0},
        {(void *)&LastDataDownlinkRate, &numOfValidArgs[8], DWPAL_LONG_LONG_INT_PARAM,
         "LastDataDownlinkRate=", 0},
        {(void *)&LastDataUplinkRate, &numOfValidArgs[9], DWPAL_LONG_LONG_INT_PARAM,
         "LastDataUplinkRate=", 0},

        /* Must be at the end */

    if (dwpal_string_to_struct_parse(reply, replyLen, fieldsToParse, sizeof(SStaStats)) ==
        DWPAL_FAILURE) {
        LOG(ERROR) << "DWPAL parse error ==> Abort";

    /* TEMP: Traces... */
    // LOG(DEBUG) << "GET_STA_MEASUREMENTS reply= \n" << reply;
    // LOG(DEBUG) << "numOfValidArgs[0]= " << numOfValidArgs[0] << " BytesSent= " << BytesSent;
    // LOG(DEBUG) << "numOfValidArgs[1]= " << numOfValidArgs[1] << " BytesReceived= " << BytesReceived;
    // LOG(DEBUG) << "numOfValidArgs[2]= " << numOfValidArgs[2] << " PacketsSent= " << PacketsSent;
    // LOG(DEBUG) << "numOfValidArgs[3]= " << numOfValidArgs[3] << " PacketsReceived= " << PacketsReceived;
    // LOG(DEBUG) << "numOfValidArgs[4]= " << numOfValidArgs[4] << " RetransCount= " << sta_stats.retrans_count;
    // LOG(DEBUG) << "numOfValidArgs[5]= " << numOfValidArgs[5] << " ShortTermRSSIAverage= " << ShortTermRSSIAverage;
    // LOG(DEBUG) << "numOfValidArgs[6]= " << numOfValidArgs[6] << " SNR= " << SNR;
    // LOG(DEBUG) << "numOfValidArgs[0]= " << numOfValidArgs[7] << " Active= " << Active;
    // LOG(DEBUG) << "numOfValidArgs[7]= " << numOfValidArgs[8] << " LastDataDownlinkRate= " << LastDataDownlinkRate;
    // LOG(DEBUG) << "numOfValidArgs[8]= " << numOfValidArgs[9] << " LastDataUplinkRate= " << LastDataUplinkRate;
    /* End of TEMP: Traces... */

    for (uint8_t i = 0; i < (sizeof(numOfValidArgs) / sizeof(size_t)); i++) {
        if (numOfValidArgs[i] == 0) {
            LOG(ERROR) << "Failed reading parsed parameter " << (int)i << " ==> Abort";
            return false;
        }
    }
````
5. The fact prplMesh holds the paring code, is a great advantage. If something will break on the parser, everyone on the community could fix it, because it is accessible, compare to dwpal code which is closed and only the Intel team can debug it.

Last thing, If the dwpal "knows" how to parse everything it is not making it generic, it makes it specific, and that is really a bad thing.

File: common/beerocks/bwl/include/bwl/key_value_parser.h (hunk 12 above)
Comment (morantr) [reply common/beerocks/bwl/include/bwl/key_value_parser.h:38]: Yes, but all the values on a single list entry are related only to values on the entry and not another entry.
If it was multimap we could not know who belongs to who.
For example:
```cpp
ch=36 bw=80
ch=38 bw=20
````
`bw=80` belongs only to channel 36. If it was multimap you couldn't know that.

File: common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h (hunk 14 above)
Comment (morantr) [reply common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h:40]: Because of differences in event output of different bwls. It is not "pure helper".

Comment (vitalybu) [reply common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h:40]: Didn't understand your explanation. BWL has a common interface, what event output differences?

Comment (morantr) [reply common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h:40]: Specifically, dummy mode parser adds an additional keyless param "EVENT" or "DATA".

File: common/beerocks/bwl/include/bwl/key_value_parser.h (hunk 12 above)
Comment (arnout) [reply common/beerocks/bwl/include/bwl/key_value_parser.h:38]: I agree with Moran on this one.

File: common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h (hunk 14 above)
Comment (arnout) [reply common/beerocks/bwl/dwpal/base_wlan_hal_dwpal.h:40]: Although I still feel the inheritance is not a good idea here, it's something that can be changed later when refactoring this for use in other BWLs.

**Comment by arnout** [comment]
I'm absolutely in favour of merging this as is.

I think it can still be simplified, i.e. I think it still does *too much* (that is in fact only needed for parsing the weird output of Intel's hostapd, i.e. that won't be needed for any other BWL). On the other hand, I think that it's possible to add generic code to handle the special cases as mentioned in e.g. https://github.com/prplfoundation/prplMesh/pull/1452/commits/2c345606200f2881e04c6d3f6a57ab5706bbd95d#r448909648 - but that should be done in a compositional way, i.e. the basic kv-parser is something that can be used by anything, and the more complicated one builds on top of it. Actually the multiline parser already follows that concept to some extent.

That said, I think we should move gradually to such an ideal world, and this is good progress. Especially since (together with #1491) it solves https://jira.prplfoundation.org/browse/PPM-11 .

File: common/beerocks/bwl/include/bwl/key_value_parser.h [hunk 19]
```diff
@@ -0,0 +1,66 @@
...
+    static void parse_line(std::istringstream &iss, std::list<char> delimiter_list,
+                           parsed_line_t &parsed_line);
+
+    static void parse_multiline(std::istringstream &iss, std::list<char> delimiter_list,
+                                parsed_multiline_t &parsed_multiline);
+
+    static bool read_param_int(const std::string &key, parsed_line_t &obj, int64_t &value,
+                               bool ignore_unknown = false);
```
arnout: As observed by @vitalybu in https://github.com/prplfoundation/prplMesh/pull/1491/commits/d8f21c02da9d711eb21ece722aa4889b5ff12e3c#r448829613 there's no need to specify the _int in the function name.
```suggestion
    static bool read_param(const std::string &key, parsed_line_t &obj, int64_t &value,
                               bool ignore_unknown = false);
```

common/beerocks/bwl/include/bwl/key_value_parser.h
```diff
@@ -0,0 +1,66 @@
...
+
+    static bool read_param_str(const std::string &key, parsed_line_t &obj, const char **value);
```
arnout: Why would you ever want a char* here?
```suggestion
    static bool read_param_str(const std::string &key, parsed_line_t &obj, const std::string&);
```

arnout (approved):
Two more comments, still approving :-)

common/beerocks/bwl/include/bwl/key_value_parser.h
arnout: D'oh
```suggestion
    static bool read_param(const std::string &key, parsed_line_t &obj, std::string& value);
```

Note that this does mean an extra copy of the string. But if you simply move the definition of the function to the header file, it will be inlined (it's small enough for that) so the copy will disappear.

kantera800 (approved):

common/beerocks/bwl/include/bwl/key_value_parser.h
morantr: Yes, the `char **` is for not create an extra copy of the string, and instead get the pointer to the value on the container.
I don't see any real benefit to do your suggestion, and it doesn't matter how you twist it, mine has better performance, so I will stick with it. 🙂

common/beerocks/bwl/common/key_value_parser.cpp
morantr: I've decided not to do it as part of this PR since the whole code has hard-coded "wlan" strings here and there (`get_ids_from_iface_string()` for example).
This could be refactored on the future PR together with all other places that use hard-coded "wlan" string.

common/beerocks/bwl/include/bwl/key_value_parser.h
morantr: done.

PR #1453, opened 2020-06-18
tests: boardfarm: port Client association link metrics test

abelog:
## Description

Port test_client_association_link_metrics from test_flows.py to
boardfarm.

## Changes

- Port `test_client_association_link_metrics` from `test_flows.py`
- Create new device to represent dummy station (`PrplMeshStationDummy`)
- Use new device in ported test
- Change naming scheme in boardfarm configuration file, so we can address devices directly by name
- Split dummy device docker image name and name of device.
In order to address devices by name, name set in configuration file should be a member of boardfarm build-in [enum](https://github.com/mattsm/boardfarm/blob/100521fde1fb67536682cafecc2f91a6e2e8a6f8/boardfarm/lib/DeviceManager.py#L25).  But name of docker images will be same, so different user's won't be able to run test at same time. To avoid it, split name of device and name of docker image spawned by the test.

## Notes

Those changes breaking all other tests, due to introduction of new device. As boardfarm isn't used at the moment it's not critical. Fix of other tests are introduced in #1486 PR, to ease review of this PR.
Part of: https://jira.prplfoundation.org/browse/PPM-12

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
@@ -5,10 +5,10 @@
 
 import os
 import time
-
 import boardfarm
-from environment import ALEntityDocker, _get_bridge_interface
+
 from .prplmesh_base import PrplMeshBase
+from environment import ALEntityDocker, _get_bridge_interface
 from sniffer import Sniffer
 
 rootdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../..'))
@@ -33,7 +33,8 @@ class PrplMeshDocker(PrplMeshBase):
         # Getting unic ID to distinguish devices and network they belong to
         self.unique_id = os.getenv("SUDO_USER", os.getenv("USER", ""))
 
-        self.name = "-".join((config.get("name", "prplmesh_docker"), self.unique_id))
+        self.name = config.get("name", "prplmesh_docker")
+        self.docker_name = "-".join((config.get("name", "prplmesh_docker"), self.unique_id))
         self.role = config.get("role", "agent")
         self.cleanup_cmd = config.get("cleanup_cmd", None)
         self.conn_cmd = config.get("conn_cmd", None)
@@ -42,7 +43,7 @@ class PrplMeshDocker(PrplMeshBase):
                                          "prplMesh-net-{}".format(self.unique_id))
 
         docker_cmd = os.path.join(rootdir, "tools", "docker", "run.sh")
-        docker_args = ["--verbose", "--detach", "--force", "--name", self.name,
+        docker_args = ["--verbose", "--detach", "--force", "--name", self.docker_name,
                        "--network", self.docker_network, "--expose", "8002"]
 
         if self.role == "controller":
@@ -51,14 +52,16 @@ class PrplMeshDocker(PrplMeshBase):
             self._run_shell_cmd(docker_cmd, docker_args)
 
             time.sleep(self.delay)
-            self.controller_entity = ALEntityDocker(self.name, device=self, is_controller=True)
+            self.controller_entity = ALEntityDocker(self.docker_name,
+                                                    device=self, is_controller=True)
         else:
             # Spawn dockerized agent
             docker_args.append("start-agent")
             self._run_shell_cmd(docker_cmd, docker_args)
 
             time.sleep(self.delay)
-            self.agent_entity = ALEntityDocker(self.name, device=self, is_controller=False)
+            self.agent_entity = ALEntityDocker(self.docker_name,
+                                               device=self, is_controller=False)
 
         self.wired_sniffer = Sniffer(_get_bridge_interface(self.docker_network),
                                      boardfarm.config.output_dir)
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
@@ -1,18 +1,24 @@
 {
     "prplmesh_docker": {
-        "name": "dockerized_device",
+        "name": "controller",
         "board_type": "prplmesh_docker",
         "role": "controller",
         "conn_cmd": "",
         "devices": [
             {
-                "name": "repeater1",
+                "name": "lan",
                 "type": "prplmesh_docker",
                 "conn_cmd": ""
             },
             {
-                "name": "repeater2",
+                "name": "lan2",
                 "type": "prplmesh_docker",
+                "conn_cmd": ""
+            },
+            {
+                "name": "wifi",
+                "type": "STA_dummy",
+                "mac": "51:a1:10:20:00:01",
                 "conn_cmd": ""
             }
         ]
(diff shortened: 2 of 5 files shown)


tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
```diff
@@ -0,0 +1,64 @@
...
+    an Associated STA Link Metrics Query message with an Associated STA Link Metrics
+    Response message containing an Associated STA Link Metrics TLV for the associated STA.'''
+
+    def runTest(self):
+        # Locate test participants in array
+        for dev in self.dev:
+            if dev.controller_entity:
+                controller = dev.controller_entity
+            if dev.agent_entity:
+                agent = dev.agent_entity
```
arnout: It's really time we refactor this setup and teardown work.

I'll convert this comment to an issue and assign it to @aciliketcap

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
```diff
@@ -0,0 +1,64 @@
...
+
+        # Regression check
+        # Don't connect not existing STAtion
+        dev.wired_sniffer.start(self.__class__.__name__ + "-" + dev.name)
+        sta_mac = "11:11:33:44:55:66"
+        debug("Send link metrics query for unconnected STA")
+        controller.ucc_socket.dev_send_1905(agent.mac, 0x800D,
```
arnout, line 39: I think the ALEntity has a helper method so you don't need the explicit ucc_socket:
```suggestion
        controller.dev_send_1905(agent.mac, 0x800D,
```

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
```diff
@@ -0,0 +1,64 @@
...
+                                            tlv(0x95, 0x0006, '{sta_mac}'.format(sta_mac=sta_mac)))
+        self.check_log(agent,
+                       "client with mac address {sta_mac} not found".format(sta_mac=sta_mac),
+                       timeout=30)
```
arnout, line 43: Please explain in the commit message why you need such a huge timeout.

I guess it's because of the overhead of doing a shell command on the target - although I can't image that would take 30 seconds... Anyway, if that's the case, then the _check_log function that does the shell command should simply add a constant offset to the timeout.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
```diff
@@ -33,7 +33,8 @@ def __init__(self, *args, **kwargs):
         # Getting unic ID to distinguish devices and network they belong to
         self.unique_id = os.getenv("SUDO_USER", os.getenv("USER", ""))
 
-        self.name = "-".join((config.get("name", "prplmesh_docker"), self.unique_id))
+        self.name = config.get("name", "prplmesh_docker")
```
arnout, line 36: Is self.name something that is used by boardfarm?

If not, I think it would be better to call this something else, and to keep using self.name for the thing with unique_id appended. Otherwise it's confusing that PrplMeshDocker.name of "prplmesh_docker", but the corresponding ALEntity.name is "prplmesh_docker-arnout".

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
```diff
@@ -1,19 +1,25 @@
...
     "prplmesh_docker": {
-        "name": "dockerized_device",
+        "name": "controller",
         "board_type": "prplmesh_docker",
         "role": "controller",
         "conn_cmd": "",
         "devices": [
             {
-                "name": "repeater1",
+                "name": "lan",
```
arnout, line 9: Why rename this? And as usual:
- should have been a separate commit;
- should be explained why in the commit message.

(You don't need to split it in a separate commit now, but please do explain in the commit message.)

If the rename is not needed, I'd much prefer it.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
```diff
@@ -19,16 +20,20 @@ class ClientAssociationLinkMetrics(PrplMeshBaseTest):
...
+        # Locate test participants
+        controller = self.dev.DUT.controller_entity
+        agent = self.dev.lan.agent_entity
+        sta = self.dev.wifi
+
+        # This test doesn't work for real HW.
+        # Skip it for prplWRT device.
+        # TODO: Remove as soon, as test works for prplWRT device
+        if self.dev.DUT is PrplMeshPrplWRT:
+            self.skipTest("This test isn't ready for prplWRT devices")
```
arnout, line 32: Could we instead do something like
```
if not self.dev.has_wifi:
    self.skipTest(...)
```

Or maybe:

```
try:
    sta = self.dev.wifi
except AttributeError:
    self.skipTest(...)
```

Note that the latter could be factored in a setup function. Let's say that PrplMeshBaseTest has a setup method that looks a bit like this:
```
def setUp(self):
    try:
        self.configure()
    except AttributeError, e:
        self.skipTest(... use e to find which device was missing ...)
```

Alternatively, we could add a get_device function that raises the skipTest if needed.

tests/boardfarm_plugins/boardfarm_prplmesh/testsuites.cfg
```diff
@@ -6,3 +6,4 @@
 [test_flows]
 InitialApConfig
 ApConfigRenew
+ClientAssociationLinkMetrics
```
arnout, line 9: This should have been in the first commit.

arnout (approved):
I really like how you approached this PR: first commit that is basically copy and paste of the test_flows code, and then a follow-up commit that "fixes" it to work with the boardfarm device. Good example for @aciliketcap.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
abelog: Yep, it is. `self.name` should be one of boardfarm build-in [enum](https://github.com/mattsm/boardfarm/blob/100521fde1fb67536682cafecc2f91a6e2e8a6f8/boardfarm/lib/DeviceManager.py#L25). If it doesn't match - we can't address devices directly by name in test.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
abelog: Actually, part of setup is refactored in #1486

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
abelog: 30 second were chosen blindly.

tests/boardfarm_plugins/boardfarm_prplmesh/prplmesh_config.json
abelog: Without rename, we can't address devices by name.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_docker.py
arnout: In that case, we should probably (eventually) rename ALEntity.name.

tests/boardfarm_plugins/boardfarm_prplmesh/testsuites.cfg
aciliketcap: I guess he added it last because test is only working once last commit is in place.

aciliketcap (approved):

aciliketcap (approved):

PR #1454, opened 2020-06-18
use SIGTERM instead of SIGKILL in killall_program()

kantera800:
Currently `prplmesh_utills.sh stop `sends SIGKILL to all processes.
Since SIGKILL cannot be handled monitor/agent/controller 
processes are not shutdown properly.
As a result old wpa ctrl socket are not detached and files
are not deleted.

change killall_program() to send SIGTERM instead on default.

### DIFF
--- a/common/beerocks/scripts/prplmesh_utils.sh.in
+++ b/common/beerocks/scripts/prplmesh_utils.sh.in
@@ -27,7 +27,7 @@ run() {
 
 killall_program() {
     PROGRAM_NAME=$1
-    KILL_SIG=${2:-KILL}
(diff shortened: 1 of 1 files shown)


VladyslavTupikin (approved):

vitalybu (approved):

arnout (approved):
It would have been nice to note in the commit message that the SIGKILL was introduced by 8fbbed86020703e8e7da0be5f4073954ed0e197d but it was unintentional - before that, no signal was specified so it defaulted to SIGTERM.

arnout:
@mergifyio rebase

PR #1440, opened 2020-06-15
Feature/store neighbor interface

mariomaz:
See #1425

PR #1456, opened 2020-06-19
Use interface name instead of index in topology database

mariomaz:
As explained in this [comment](https://github.com/prplfoundation/prplMesh/pull/1420#discussion_r441604127), it is better to use interface name instead of interface index. 

This PR is a hotfix for #1440 where we started storing interface index in topology database.


mariomaz:
Please review the last two commits only

mariomaz:
Created a new PR to avoid mistakes with extra commits

Issue #1308, opened 2020-05-17
Now that we have packet captures integrated into test_flows.py, we should use it whenever it's possible.


RanRegev:
Now that we have packet captures integrated into test_flows.py, we should use it whenever it's possible.
Here you should be able to check that the "ACK" message is sent by using `check_cmdu_type`

_Originally posted by @rmelotte in https://github.com/prplfoundation/prplMesh/pull/1273_

PR #1457, opened 2020-06-19
Verify using sniffed packets instead of logs in link metric query and client capability query tests

aciliketcap:


aciliketcap:
This PR is part of #1308, it utilizes sniffer in two more tests.

aciliketcap:
Do we expect tests to fail at first issue (like most unit tests) or should they run until they can report each and every issue they encountered? Is it OK if they throw unhandled exceptions or should they mark themselves as failed and then log and swallow any expected exceptions?

I wrote the tests reporting as much issues as they can and never raising exceptions (that we expect).

VladyslavTupikin:
Hi. @aciliketcap  Can you please prepare commits according to the instruction in CONTRIBUTING.md?

tests/test_flows.py
```diff
@@ -814,53 +834,131 @@ def test_client_capability_query(self):
...
+                            "00:00:00:00:00:00:00:00:00:00:00:00:00:00:18:e6:"\
+                            "e1:09:00:bf:0c:b0:79:d1:33:fa:ff:0c:03:fa:ff:0c:"\
+                            "03:c7:01:10"
+
+        # first check capability query fails on unconnected station
         debug("Send client capability query for unconnected STA")
-        env.controller.dev_send_1905(env.agents[0].mac, 0x8009,
-                                     tlv(0x90, 0x000C,
-                                         '{} {}'.format(env.agents[0].radios[0].mac, sta1.mac)))
+        mid1 = env.controller.dev_send_1905(env.agents[0].mac, 0x8009,
```
abelog: *Nitpick*: i really don't like such senseless variables.
```suggestion
        mid_cap_query = env.controller.dev_send_1905(env.agents[0].mac, 0x8009,
```
Will have more sense and easier to read, IMHO.

tests/test_flows.py
```diff
@@ -814,53 +834,131 @@ def test_client_capability_query(self):
...
-        # check that both radio agents received it, in the future we'll add a check to verify which
-        # radio the query was intended for.
-        self.check_log(env.agents[0], r"CLIENT_CAPABILITY_QUERY_MESSAGE")
 
-        debug("Confirming client capability report message has been received on controller")
-        self.check_log(env.controller, r"Received CLIENT_CAPABILITY_REPORT_MESSAGE")
-        self.check_log(env.controller,
-                       r"Result Code= FAILURE, client MAC= {}, BSSID= {}"
-                       .format(sta1.mac, env.agents[0].radios[0].mac))
+        query1 = self.check_cmdu_type_single("client capability query", 0x8009, env.controller.mac,
```
abelog: Same nitpick about sensless naming of variable.
```suggestion
        client_cap_query = self.check_cmdu_type_single("client capability query", 0x8009, env.controller.mac,
```

tests/test_flows.py
```diff
@@ -814,53 +834,131 @@ def test_client_capability_query(self):
...
+                try:
+                    if query2_tlv.client_info_mac_addr != sta2.mac:
+                        self.fail("Wrong mac address in query")
+                except AttributeError:
+                    self.fail("Query does not contain mac address")
+                try:
+                    if query2_tlv.client_info_bssid != env.agents[0].radios[0].mac:
+                        self.fail("Wrong bssid in query")
+                except AttributeError:
+                    self.fail("Query does not contain bssid")
```
abelog: This block repeats frequently, should it be a separate method? It will significantly improve readability.

tests/test_flows.py
```diff
@@ -814,53 +834,131 @@ def test_client_capability_query(self):
...
+                    self.fail("Report does not contain bssid")
+
+            report2_cap_report_tlv = self.check_cmdu_has_tlvs(report2, 0x91)[0]
+            if report2_cap_report_tlv:
+                try:
+                    if int(report2_cap_report_tlv.client_capability_result, 0) != 0x00:
+                        self.fail("Capability report result is not successful")
+                    if report2_cap_report_tlv.client_capability_frame != association_frame:
+                        self.fail("Capability report does not contain correct frame")
+                        debug(f"Frame received\n{report2_cap_report_tlv.client_capability_frame}")
```
abelog: *Nitpick*: mainly, we tend to use `format` for string formatting.

abelog (changes requested):
I still think that we should avoid repeat of code blocks. So not approving yet.
*Nitpicks* are optional to fix.

tests/test_flows.py
aciliketcap: I kept it as mid since the parameter name in check_cmdu_type method is mid and since it is not really used in test code.

tests/test_flows.py
aciliketcap: yeah, I guess I can make this more meaningful, like query_for_error instead of query1 and query_for_success instead of query2.

tests/test_flows.py
aciliketcap: All of them are removed since I'm using assert statements now.

tests/test_flows.py
aciliketcap: Oh, if most of the code is written that way I can change it too. But did you consider using format strings instead of .format() ? Many people (for example devs in my  current company :)) like format strings more.

tests/test_flows.py
```diff
@@ -695,17 +695,37 @@ def test_ap_capability_query(self):
...
+            except (AttributeError, ValueError):
+                self.fail("Requested metrics are not recognized")
+
+        resp = self.check_cmdu_type_single("link metric response", 0x0006,
+                                           env.agents[0].mac, env.controller.mac,
+                                           mid)
+        if resp:
+            debug("Checking response contains both Tx and Rx metrics")
+            self.check_cmdu_has_tlvs(resp, 0x09)
+            self.check_cmdu_has_tlvs(resp, 0x0a)
```
arnout: We can check a little more about these TLVs. E.g. at least the controller should be a neighbour, so it should be mentioned in both metrics.

tests/test_flows.py
arnout: I think "query" would be OK, but as soon as you need more than one, "query1" and "query2" are meaningless.

In this specific case, however, I think it would be better to split the test into two separate tests. The two parts of the test are pretty much independent of each other.

aciliketcap: I changed the variable names in that test. This time the lines are too long but I think it is OK for test code, what we are testing should be clear in variable names.

Btw I didn't used client_cap_query name since it is evident in the test name and check_cmdu_type error string parameter. I instead changed the name to reflect that we are testing for a query which returns a report with error and then a query with a successful report.

tests/test_flows.py
```diff
@@ -834,53 +834,131 @@ def test_client_capability_query(self):
...
-73 74 75 76 77 78 7c 7d 7e 7f 80 82 3b 16 0c 01
-02 03 04 05 0c 16 17 18 19 1a 1b 1c 1d 1e 1f 20
-21 80 81 82 46 05 70 00 00 00 00 46 05 71 50 50
-00 04 7f 0a 04 00 0a 82 21 40 00 40 80 00 dd 07
-00 50 f2 02 00 01 00 2d 1a 2d 11 03 ff ff 00 00
-00 00 00 00 00 00 00 00 00 00 00 00 00 00 18 e6
-e1 09 00 bf 0c b0 79 d1 33 fa ff 0c 03 fa ff 0c
-03 c7 01 10 """
-
+        association_frame = "00:0e:4d:75:6c:74:69:2d:41:50:2d:32:34:47:2d:31:"\
```
arnout: Since you're changing this anyway, it would make sense to also move it down to the place where you actually need it (i.e. where the station associates).

aciliketcap:
The CI flow gave flake8 error but it does not appear when I run flake8 with 'flake8 --config .flake8 tests/' ?

tests/test_flows.py
```diff
@@ -834,53 +834,131 @@ def test_client_capability_query(self):
...
+                try:
+                    if query1_tlv.client_info_bssid != env.agents[0].radios[0].mac:
+                        self.fail("Wrong bssid in query")
+                except AttributeError:
+                    self.fail("Query does not contain bssid")
+
+        report1 = self.check_cmdu_type_single("client capability report", 0x800a,
+                                              env.agents[0].mac, env.controller.mac, mid1)
+        if report1:
+            report1_client_info_tlv = self.check_cmdu_has_tlvs(report1, 0x90)[0]
```
arnout: I don't like that you take `[0]` here - especially since the order of the TLVs is not defined. So either there is just one, and you should use check_cmdu_has_tlv_single, or there are several and you should iterate over them to find the one you want to check.

In this case, since we only requested a single station, there must be only a single client info tlv, so _single is appropriate.

Same issue occurs in a few other places.

tests/test_flows.py
arnout: Unfortunately, Intel is still using build machines with Python 3.4.

tests/test_flows.py
```diff
@@ -278,6 +279,10 @@ def run_tests(self, tests):
             self.check_error = 0
             try:
                 getattr(self, test_full)()
+            except AssertionError as ae:
+                self.fail(f"{ae}")
+            except Exception as e:
+                self.fail(f"Test has thrown an exception\n{traceback.format_exc()}")
```
arnout: I don't think we should print the full traceback. Also, I think it's worth repeating the test name.
```suggestion
                self.fail("Test {} has thrown an exception\n{}".format(test, e))
```

(Incidentally, this will also fix the flake8 error because `e` is now used)

With that, there's also no real need to handle the AssertionError differently.

BTW, it would be nice to show in the commit message some examples of what the output will be when an exception is handled. Something like:
```
For example, when a non-existing TLV field is accessed, the following failure will be printed:
...
```
For reviewers, that's much easier to help decide if it is done properly.

tests/test_flows.py
```diff
@@ -707,30 +707,28 @@ def test_link_metric_query(self):
         query = self.check_cmdu_type_single("link metric query", 0x0005,
                                             env.controller.mac, env.agents[0].mac,
                                             mid)
+        assert query
 
-        if query:
-            query_tlv = self.check_cmdu_has_tlv_single(query, 0x08)
+        query_tlv = self.check_cmdu_has_tlv_single(query, 0x08)
+        assert query_tlv
```
arnout: I would actually move the assert inside the check_ function. Yes, there are some cases where we actually could continue the test even if the CMDU is missing, but that's not the common case. In the few cases where we do want to continue, we can still add a try block.

And actually, the assertion isn't even needed, because it's implied when you start using query_tlv below.

tests/test_flows.py
```diff
@@ -707,30 +707,28 @@ def test_link_metric_query(self):
...
-                    self.fail("Query type is not 'All neighbors'")
-            except (AttributeError, ValueError):
-                self.fail("Query type is not recognized")
-            try:
-                if int(query_tlv.link_metrics_requested) != 0x02:
-                    self.fail("Metrics for both Tx and Rx is not requested")
-            except (AttributeError, ValueError):
-                self.fail("Requested metrics are not recognized")
+        debug("Checking query type and queried metrics are correct")
+        assert query_tlv.link_metric_query_type, "Query type is not recognized"
```
arnout: This is not needed. We can pretty much assume that it's there, otherwise the tlv wouldn't exist to begin with. And if it doesn't, the following statement will simply raise an AttributeError.

tests/test_flows.py
```diff
@@ -707,30 +707,28 @@ def test_link_metric_query(self):
...
+        assert int(query_tlv.link_metric_query_type) == 0x00, \
+            "Query type is not 'All neighbors'"
```
arnout: I don't think we should use asserts everywhere. It's better if we can continue the test, so assert isn't always good. And in this case I actually think the original code is easier to read.

tests/test_flows.py
```diff
@@ -852,31 +852,33 @@ def test_client_capability_query(self):
 
         # first check capability query fails on unconnected station
         debug("Send client capability query for unconnected STA")
-        mid1 = env.controller.dev_send_1905(env.agents[0].mac, 0x8009,
-                                            tlv(0x90, 0x000C, '{} {}'.format(
-                                                env.agents[0].radios[0].mac, sta1.mac)))
+        mid_for_error = env.controller.dev_send_1905(env.agents[0].mac, 0x8009,
```
arnout: As noted elsewhere, this is not needed if we simply split the test into two tests.

Also, changes that are a result of review should typically be done as fixups. So in this case, do an interactive rebase and edit the commit that introduces mid1 and fix it there.

arnout (dismissed):
Pretty good PR for your first contribution!

Myself, I would have put the commit that adds the exception handlers first, so the "use sniffed data" commits could immediately use the new approach. However, I kind of like this style as well, because it clearly shows why such a refactoring is needed - it's a bit of the [OneTwoThreeRefactor](http://c2.com/cgi/wiki?OneTwoThreeRefactor) approach.

Our review workflow is that I already approve now, even though there are still plenty of review comments to address. By approving already, you can simply make the requested changes and merge right away. And if there is something you're not sure of or disagree with, we discuss it either in the comments or in Slack before merging.

tests/test_flows.py
aciliketcap: Ah, I totally missed that this was causing the flake8 error.

aciliketcap: I expect tests to throw AssertionError when they fail and decide to quit. However I consider other exceptions as unexpected run-time exceptions and therefore I conclude that the test code has failed, not the DUT. I think we definitely need the stack trace (in debug / verbose mode) so that we can have a chance to figure out what went wrong without having to debug the test.

abelog:
Removing requested changes mark as branch is rebuild by author.

abelog (dismissed):

tests/test_flows.py
aciliketcap: Then I'll definitely remove all format strings.

tests/test_flows.py
aciliketcap: I will be checking both links to controller and the other agent using what we know in env.

tests/test_flows.py
aciliketcap: I do that to prevent the AttributeError, test should only fail with AssertionError I think, otherwise we can't discern if the test failed or the test code failed.

tests/test_flows.py
```diff
@@ -278,6 +279,18 @@ def run_tests(self, tests):
...
             try:
                 getattr(self, test_full)()
+            except AssertionError as ae:
+                # do not add empty message if test has already been marked as failed
+                # and AssertError does not contain a message
+                if self.check_error:
+                    if not str(ae):
+                        pass
+                else:
+                    self.fail("{}".format(ae))
```
arnout: The True branch of the condition does nothing at all, so
```suggestion
                if not self.check_error:
                    self.fail("{}".format(ae))
```

However, I don't see why you would depend on check_error. If the assertion came after a completely unrelated failure, you still want to mark the failure. So I'd just do:

```suggestion
                if str(ae):
                    self.fail("{}".format(ae))
```

aciliketcap: I still wanted test to fail if self.fail() is not called before (inside test method) and there is no assertion failure message. I ended up printing a generic "Assertion failed" message in that case.

tests/test_flows.py
arnout: It doesn't actually prevent the attribute error - the assert will never even trigger if the attribute doesn't exist. You'd need
```
assert hasattr(query_tlv, "link_metric_query_type")
```

However, I think it is much more convenient if we can just access attributes without explicit asserts. In this case, we could actually implement a `__getattr__` function on TlvStruct that asserts.

tests/test_flows.py
```diff
@@ -266,6 +269,13 @@ def check_topology_notification(self, eth_src: str, neighbors: list,
 
         return True
 
+    def safe_check_obj_attribute(self, obj, attrib_name, expected_val, fail_str):
```
abelog: *Nitpick*: please document arguments types. Makes usage of this method easier for others.

tests/test_flows.py
```diff
@@ -114,20 +116,19 @@ def check_cmdu_type(
...
 
     def check_cmdu_type_single(
         self, msg: str, msg_type: int, eth_src: str, eth_dst: str = None, mid: int = None
-    ) -> Union[sniffer.Packet, None]:
+    ) -> sniffer.Packet:
         '''Like check_cmdu_type, but also check that only a single CMDU is found.'''
         cmdus = self.check_cmdu_type(msg, msg_type, eth_src, eth_dst, mid)
         if not cmdus:
-            return None  # Failure already reported by check_cmdu
+            assert False  # Failure already reported by check_cmdu
```
abelog, line 126: *Nitpick*: We tend to have commentaries above the string, not inline.

tests/test_flows.py
```diff
@@ -810,58 +846,83 @@ def test_combined_infra_metrics(self):
         vap1.disassociate(sta1)
         vap2.disassociate(sta2)
 
-    def test_client_capability_query(self):
-        sta1 = env.Station.create()
-        sta2 = env.Station.create()
+    def base_test_client_capability_query(self, sta):
```
abelog: *Nitpick*: please document arguments types. Makes usage of this method easier for others.

tests/test_flows.py
```diff
@@ -810,58 +846,83 @@ def test_combined_infra_metrics(self):
...
+        report = self.base_test_client_capability_query(sta)
 
-        debug("Confirming client capability report message has been received on controller")
-        self.check_log(env.controller, r"Received CLIENT_CAPABILITY_REPORT_MESSAGE")
-        self.check_log(env.controller,
-                       r"Result Code= FAILURE, client MAC= {}, BSSID= {}"
-                       .format(sta1.mac, env.agents[0].radios[0].mac))
+        cap_report_tlv = self.check_cmdu_has_tlv_single(report, 0x91)
+        self.safe_check_obj_attribute(cap_report_tlv, 'client_capability_result', '0x00000001',
+                                      "Capability report does not contain error")
```
abelog: To make message more informative.
```suggestion
                                      "Capability query was successful for disconnected STA")
```

tests/test_flows.py
```diff
@@ -810,58 +846,83 @@ def test_combined_infra_metrics(self):
...
+        try:
+            report = self.base_test_client_capability_query(sta)
 
-        env.agents[0].radios[0].vaps[0].disassociate(sta2)
+            cap_report_tlv = self.check_cmdu_has_tlvs(report, 0x91)[0]
+            self.safe_check_obj_attribute(cap_report_tlv, 'client_capability_result', '0x00000000',
+                                          "Capability report result is not successful")
+            try:
+                if cap_report_tlv.client_capability_frame != expected_association_frame:
+                    self.fail("Capability report does not contain correct frame")
```
abelog: Just small suggestion.
```suggestion
                    self.fail("Capability report does not contain expected frame")
```

tests/test_flows.py
```diff
@@ -810,58 +846,83 @@ def test_combined_infra_metrics(self):
...
+                    debug(f"Frame received\n{cap_report_tlv.client_capability_frame}")
+                    debug(f"Frame expected\n{expected_association_frame}")
```
abelog, line 916: Seems, those debug messages should be before `self.fail` call.

abelog (approved):
Nitpicks only, nothing crucial. In general - LGTM.

tests/test_flows.py
```diff
@@ -80,6 +80,7 @@ def check_cmdu(
         result = env.wired_sniffer.get_cmdu_capture(match_function)
         if not result:
             self.fail("No CMDU {} found".format(msg))
+            assert False
```
arnout: (nitpick)
```suggestion
        assert result, "No CMDU {} found".format(msg)
```

You already had something like this below, so I guess it was just an oversight :-) But there are still other cases that could be simplified like this.

tests/test_flows.py
arnout: No, we have both actually.

tests/test_flows.py
```diff
@@ -174,24 +175,25 @@ def check_cmdu_has_tlvs(
...
         '''Like check_cmdu_has_tlvs, but also check that only one TLV of that type is found.'''
         tlvs = self.check_cmdu_has_tlvs(packet, tlv_type)
         if not tlvs:
-            return None
+            assert False
         if len(tlvs) > 1:
             self.fail("More than one ({}) TLVs of type 0x{:02x} found".format(len(tlvs), tlv_type))
             debug("  {}".format(packet))
-            return None
+            assert False
```
arnout, line 189: Hm, it would be nice if we had something like `ifdebug` that allows us to do:
```
assert len(tlvs) == 1, "More than one ({}) TLVs of type 0x{:02x} found{}".format(
    len(tlvs), tlv_type, ifdebug("\n  " + str(packet)))
```

Not for this PR, probably.

tests/test_flows.py
```diff
@@ -174,24 +175,25 @@ def check_cmdu_has_tlvs(
...
         return tlvs[0]
 
     def check_cmdu_has_tlvs_exact(
         self, packet: Union[sniffer.Packet, None], tlvs: [sniffer.Tlv]
-    ) -> None:
+    ) -> bool:
```
arnout: If you change this in asserts as well, there's no need to correct the return type.

tests/test_flows.py
```diff
@@ -716,17 +716,32 @@ def test_ap_capability_query(self):
...
     def test_link_metric_query(self):
-        env.controller.dev_send_1905(env.agents[0].mac, 0x0005,
-                                     tlv(0x08, 0x0002, "0x00 0x02"))
+        mid = env.controller.dev_send_1905(env.agents[0].mac, 0x0005,
+                                           tlv(0x08, 0x0002, "0x00 0x02"))
         time.sleep(1)
 
-        debug("Confirming link metric query has been received on agent")
-        self.check_log(env.agents[0], "Received LINK_METRIC_QUERY_MESSAGE")
+        query = self.check_cmdu_type_single("link metric query", 0x0005,
```
arnout, line 718: Commit message is not correct... It says "use assert statements" but you're actually using the sniffer instead of check_log.

tests/test_flows.py
```diff
@@ -846,58 +846,83 @@ def test_combined_infra_metrics(self):
...
+        # then check capability query is successful with connected station
+        try:
+            report = self.base_test_client_capability_query(sta)
 
-        env.agents[0].radios[0].vaps[0].disassociate(sta2)
+            cap_report_tlv = self.check_cmdu_has_tlvs(report, 0x91)[0]
+            self.safe_check_obj_attribute(cap_report_tlv, 'client_capability_result', '0x00000000',
+                                          "Capability report result is not successful")
+            try:
+                if cap_report_tlv.client_capability_frame != expected_association_frame:
```
arnout, line 913: Doesn't safe_check_obj_attribute work here? Or is it just for the debug info that you have to special case it?

arnout (approved):
*Excellent* commit messages! And in general, this is a pretty good pull request.

In the future though I think it's better to merge things a little sooner and improve in follow-up pull requests. Review feedback doesn't necessarily mean that you have to make those changes right away.

tests/test_flows.py
aciliketcap: Yeah, but I though when you are reading you'd first want to be informed that the test has failed and then how it failed. If self.fail() made the test fail then of course I'd have to do it the other way around.

tests/test_flows.py
aciliketcap: We can prepare a string before hand (empty string if not debug) and then print it. Doing this in a separate function as you suggested is cleaner of course.

tests/test_flows.py
aciliketcap: I did it. This method is not used btw :shrug:

tests/test_flows.py
aciliketcap: You are absolutely right

tests/test_flows.py
aciliketcap: Yes, that's because I wanted to stop and print debug info if attribute exists but it is wrong. Maybe I can return False when safe_check_attribute fails and write it like;
if safe_check_obj_attrib():
  if cap_report.frame != expected_frame:
    debug_stuff()

PR #1458, opened 2020-06-19
Use neighbor's interface info in Topology Response

### LINKED ISSUE #1428: [TASK] Use neighbor's interface info in Topology Response
Current implementation does not build Topology Response message properly: it is considering that there is a single interface, the bridge interface, behind which all neighbors are visible. and it should however report all the local interfaces together with the neighbors that are connected to those interfaces..

This task depends on #1425, where topology database will be extended to also include the interface that links to each neighbor.

mariomaz:
See #1428

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -2927,40 +2927,64 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
         }
     }
 
-    // The tlv1905NeighborDevice we construct here is not entirely correct, but close enough.
-    // We should create a tlv1905NeighborDevice for every local interface on which a 1905 neighbor
-    // is visible, and then add a list of neighbors visible on that interface. However, since we
-    // don't keep track of the incoming interface of the received topology discovery messages, we
-    // can't do that. Instead, we model it as a single interface, the bridge interface, behind which
-    // all neighbors are visible, with a bridge between them. This is accurate enough to allow us to
-    // work in practice.
-    auto tlv1905NeighborDevice = cmdu_tx.addClass<ieee1905_1::tlv1905NeighborDevice>();
-    if (!tlv1905NeighborDevice) {
-        LOG(ERROR) << "addClass ieee1905_1::tlv1905NeighborDevice failed, mid=" << std::hex
-                   << (int)mid;
-        return false;
-    }
-    tlv1905NeighborDevice->mac_local_iface() = tlvf::mac_from_string(bridge_info.mac);
+    /**
+     * Add a 1905.1 neighbor device TLV for each local interface for which this management entity
+     * has inferred the presence of a 1905.1 neighbor device. Include each discovered neighbor
+     * device in its corresponding 1905.1 neighbor device TLV.
+     *
+     * First, group known 1905 neighbor devices by the local interface that links to them. Create
+     * a map which key is the name of the local interface and the value is the list of neighbor
+     * devices inferred from that interface.
+     */
+    std::unordered_map<std::string, std::vector<sNeighborDevice>> neighbor_devices_by_local_iface;
+    for (const auto &entry : m_1905_neighbor_devices) {
+        const auto &neighbor_device = entry.second;
+        auto &neighbor_devices      = neighbor_devices_by_local_iface[neighbor_device.if_name];
 
-    if (!tlv1905NeighborDevice->alloc_mac_al_1905_device(m_1905_neighbor_devices.size())) {
-        LOG(ERROR) << "alloc_mac_al_1905_device() has failed";
-        return true;
+        neighbor_devices.emplace_back(neighbor_device);
     }
 
-    size_t index = 0;
-    for (const auto &neighbor_device : m_1905_neighbor_devices) {
-        const auto &neighbor_al_mac = neighbor_device.first;
+    /**
+     * Second, create a tlv1905NeighborDevice for every local interface on which a 1905 neighbor
+     * is visible, and then add the list of neighbors visible on that interface
+     */
+    for (const auto &entry : neighbor_devices_by_local_iface) {
+        const auto &iface_name       = entry.first;
+        const auto &neighbor_devices = entry.second;
 
-        auto mac_al_1905_device_tuple = tlv1905NeighborDevice->mac_al_1905_device(index);
-        if (!std::get<0>(mac_al_1905_device_tuple)) {
-            LOG(ERROR) << "getting mac_al_1905_device element has failed";
+        sMacAddr iface_mac;
+        if (!get_iface_mac(iface_name, iface_mac)) {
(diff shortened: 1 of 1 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2927,40 +2927,64 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+     */
+    std::unordered_map<std::string, std::vector<sNeighborDevice>> neighbor_devices_by_local_iface;
+    for (const auto &entry : m_1905_neighbor_devices) {
+        const auto &neighbor_device = entry.second;
+        auto &neighbor_devices      = neighbor_devices_by_local_iface[neighbor_device.if_name];
 
-    if (!tlv1905NeighborDevice->alloc_mac_al_1905_device(m_1905_neighbor_devices.size())) {
-        LOG(ERROR) << "alloc_mac_al_1905_device() has failed";
-        return true;
+        neighbor_devices.emplace_back(neighbor_device);
```
arnout, line 2944: @RanRegev How would you write this without raw loop?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2927,40 +2927,64 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+     * has inferred the presence of a 1905.1 neighbor device. Include each discovered neighbor
+     * device in its corresponding 1905.1 neighbor device TLV.
+     *
+     * First, group known 1905 neighbor devices by the local interface that links to them. Create
+     * a map which key is the name of the local interface and the value is the list of neighbor
+     * devices inferred from that interface.
+     */
+    std::unordered_map<std::string, std::vector<sNeighborDevice>> neighbor_devices_by_local_iface;
+    for (const auto &entry : m_1905_neighbor_devices) {
+        const auto &neighbor_device = entry.second;
```
arnout, line 2941: This is the reason I don't like using map too much. You often have several cases where you don't actually need the key, and thus you add unnecessary complexity to the code.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2927,40 +2927,64 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
     }
 
-    size_t index = 0;
-    for (const auto &neighbor_device : m_1905_neighbor_devices) {
-        const auto &neighbor_al_mac = neighbor_device.first;
+    /**
+     * Second, create a tlv1905NeighborDevice for every local interface on which a 1905 neighbor
+     * is visible, and then add the list of neighbors visible on that interface
+     */
+    for (const auto &entry : neighbor_devices_by_local_iface) {
```
arnout, line 2951: (nitpick) It would make more sense to me to just iterate over all interfaces and then for each interface, iterate over the neighbours to collect the ones connected to that interface.

Note that we already collect the local interfaces and sent that information to the transport, and that is the way the transport knows which interfaces to listen on. See backhaul_fsm_main, state MASTER_DISCOVERY. Only, that list of interfaces is not stored at the moment.

I also wonder... Wouldn't it be better to keep an actual data model, rather than constructing the information we need ad hoc? If we had a data model, we could very easily do what I propose: query the list of local interfaces, then for each interface query which neighbours are visible from it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2927,40 +2927,64 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+        auto iface_name              = entry.first;
```
```suggestion
        const auto &iface_name              = entry.first;
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2927,40 +2927,64 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+        const auto &neighbor_devices = entry.second;
 
-        auto mac_al_1905_device_tuple = tlv1905NeighborDevice->mac_al_1905_device(index);
-        if (!std::get<0>(mac_al_1905_device_tuple)) {
-            LOG(ERROR) << "getting mac_al_1905_device element has failed";
+        sMacAddr iface_mac;
+        if (!get_iface_mac(iface_name, iface_mac)) {
+            return false;
```
arnout, line 2957: With a proper data model, we would store the mac address as part of the model rather than requesting it every time.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2927,40 +2927,64 @@ bool backhaul_manager::handle_1905_topology_query(ieee1905_1::CmduMessageRx &cmd
...
+        }
+
+        auto tlv1905NeighborDevice = cmdu_tx.addClass<ieee1905_1::tlv1905NeighborDevice>();
+        if (!tlv1905NeighborDevice) {
+            LOG(ERROR) << "addClass ieee1905_1::tlv1905NeighborDevice failed, mid=" << std::hex
+                       << (int)mid;
```
```suggestion
                       << mid;
```

arnout (approved):
Typo in commit message: din't -> don't or didn't.

I think it would be even better with a full-fledged data model, but let's go for this one already.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: Well... I admit... I tried and failed.
I feel uncomfortable with building "local structures" and then `transforming` between them whenever there is a need. This is opening a can of possible bugs and creates confusion, uncertainty and repeats.
In this case both `m_1905_neighbor_devices` and `neighbor_devices_by_local_iface` are local structures in the sense that they are pure, native, std structures.
A more robust approach, IMO, would be to have:
```cpp
class NeighborDevices { 
/* construct, fill, query */
// and also -
temaplte<class outIt> get_all_nieghbor_for_ifname(const std::string &if_name, OutIt out);
};
```
Or anything alike.
The idea is this -
Instead of letting the developer of the algorithm mass around with pure structures **when implementing the logical flow** - hide these searches and transformation in a separate place, not in the main flow, not where the logic is. Transformation, searches, validity checks, they all must be moved elsewhere, leaving the flow clean, expressive and accurate.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: I agree it would be better to keep a data model instead of fetching required information each time we need it. But that's maybe for another PR #1462

VladyslavTupikin (approved):

PR #1459, opened 2020-06-19
Hotfix/use iface name instead of index in topology db

mariomaz:
As explained in this comment, it is better to use interface name instead of interface index.

This PR is a hotfix for #1440 where we started storing interface index in topology database.

### DIFF
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -3329,6 +3329,13 @@ bool backhaul_manager::handle_1905_topology_discovery(const std::string &src_mac
         return false;
     }
 
+    uint32_t if_index   = message_com::get_uds_header(cmdu_rx)->if_index;
+    std::string if_name = network_utils::linux_get_iface_name(if_index);
+    if (if_name.empty()) {
+        LOG(ERROR) << "Failed getting interface name for index: " << if_index;
+        return false;
+    }
+
     auto new_device =
         m_1905_neighbor_devices.find(tlvAlMac->mac()) == m_1905_neighbor_devices.end();
 
@@ -3336,7 +3343,7 @@ bool backhaul_manager::handle_1905_topology_discovery(const std::string &src_mac
     sNeighborDevice neighbor_device;
     neighbor_device.al_mac    = tlvAlMac->mac();
     neighbor_device.mac       = tlvMac->mac();
-    neighbor_device.if_index  = message_com::get_uds_header(cmdu_rx)->if_index;
+    neighbor_device.if_name   = if_name;
     neighbor_device.timestamp = std::chrono::steady_clock::now();
 
     m_1905_neighbor_devices[tlvAlMac->mac()] = neighbor_device;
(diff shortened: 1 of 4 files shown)


arnout:
Since #1440 was already merged, I changed the target of this PR to master.

arnout:
@mergifyio rebase

arnout:
Oh my bad... I hadn't noticed that this is actually on top of #1420 which hasn't been merged yet... Now I had to do a manual rebase to restore it...

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -533,6 +533,18 @@ uint32_t network_utils::linux_get_iface_index(const std::string &iface_name)
     return iface_index;
 }
 
+std::string network_utils::linux_get_iface_name(uint32_t iface_index)
+{
+    char iface_name[IF_NAMESIZE] = {0};
+    if (!if_indextoname(iface_index, iface_name)) {
```
arnout, line 539: There are currently 6 open-coded calls to if_indextoname, it would be good to replace those as well. In a separate PR, of course.

common/beerocks/bcl/source/network/network_utils.cpp
```diff
@@ -533,6 +533,18 @@ uint32_t network_utils::linux_get_iface_index(const std::string &iface_name)
...
+        LOG(ERROR) << "Failed to read the name of interface with index " << iface_index << ": "
```
arnout, line 540: (nitpick) Although at first sight it makes sense to log an error here, the problem is that when the error occurs, you have absolutely no context from the log - except maybe a hint from the preceding message, but that would typically only be if DEBUG is enabled.
Therefore, I think in general it is better to not do any logging in the utility functions, but instead do it in the caller. Well, to be very precise: only do logging of information that is not propagated to the caller (e.g. if there's an errno to report).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -3329,14 +3329,20 @@ bool backhaul_manager::handle_1905_topology_discovery(const std::string &src_mac
         return false;
     }
 
+    uint32_t if_index   = message_com::get_uds_header(cmdu_rx)->if_index;
+    std::string if_name = network_utils::linux_get_iface_name(if_index);
+    if (if_name.empty()) {
+        return false;
```
arnout, line 3336: So log the error here, where you can explicitly say it's in the topology discovery handling.

arnout (approved):

common/beerocks/bcl/source/network/network_utils.cpp
mariomaz: Done: #1461

common/beerocks/bcl/source/network/network_utils.cpp
mariomaz: This is one of the reasons I prefer using exceptions.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Done

VladyslavTupikin (approved):

PR #1460, opened 2020-06-21
Persistent Database: BPL stubs

itayx:
To support the persistent DB feature, BPL APIs that support the
reading/writing functionality are needed

### DIFF
--- a/framework/platform/bpl/include/bpl/bpl_db.h
+++ b/framework/platform/bpl/include/bpl/bpl_db.h
@@ -1,0 +1,96 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _BPL_DB_H_
+#define _BPL_DB_H_
+
+#include "bpl.h"
+#include "bpl_err.h"
+#include <stdint.h>
+#include <string>
+#include <unordered_map>
+
+namespace beerocks {
+namespace bpl {
+
+/****************************************************************************/
+/******************************* Definitions ********************************/
+/****************************************************************************/
+
+/****************************************************************************/
+/******************************* Structures *********************************/
+/****************************************************************************/
+
+/****************************************************************************/
+/******************************** Functions *********************************/
+/****************************************************************************/
+
+/**
+ * @brief find if entry exists
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @return true if entry exists in db, false otherwise
+ */
+bool db_has_entry(const std::string &entry_type, const std::string &entry_name);
+
+/**
+ * @brief Add new entry with initial value.
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @param[in] params initial params of the entry.
+ * @return true on success, false otherwise
+ */
+bool db_add_entry(const std::string &entry_type, const std::string &entry_name,
+                  const std::unordered_map<std::string, std::string> &params);
+/**
+ * @brief Set values in entry, append attributes and override existing.
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @param[in] params unordered_map containing the params the requested entry, to be set.
+ * @return true on success, false otherwise.
+ */
+bool db_set_entry(const std::string &entry_type, const std::string &entry_name,
+                  const std::unordered_map<std::string, std::string> &params);
+/**
+ * @brief Get values of entry, return all avaliable values.
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @param [in/out] params unordered_map containing the params of the requested entry.
+ * @return true on success, false otherwise.
+ */
+bool db_get_entry(const std::string &entry_type, const std::string &entry_name,
+                  std::unordered_map<std::string, std::string> &params);
+
+/**
+ * @brief Get all entries by type as a nested map <Entry-name, <Attrebute-name, value>>.
+ * 
+ * @param[in] entry_type type of the requested entries.
+ * @param[in/out] nested_params nested map of params (with entry's name as key).
+ * @return true on success, false otherwise.
+ */
+bool db_get_entries_by_type(
+    const std::string &entry_type,
+    std::unordered_map<std::string, std::unordered_map<std::string, std::string>> &nested_params);
+
+/**
+ * @brief Remove entry by name.
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @return true on success, false otherwise.
+ */
+bool db_remove_entry(const std::string &entry_type, const std::string &entry_name);
+
+} // namespace bpl
+} // namespace beerocks
+
+#endif /* _BPL_DB_H_ */
--- a/framework/platform/bpl/linux/bpl_db.cpp
+++ b/framework/platform/bpl/linux/bpl_db.cpp
@@ -1,0 +1,62 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "bpl/bpl_db.h"
+
+#include <mapf/common/logger.h>
+
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Implementation ///////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+namespace beerocks {
+namespace bpl {
+
+bool db_has_entry(const std::string &entry_type, const std::string &entry_name)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return false;
+}
+
+bool db_add_entry(const std::string &entry_type, const std::string &entry_name,
+                  const std::unordered_map<std::string, std::string> &params)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return true;
+}
+
+bool db_set_entry(const std::string &entry_type, const std::string &entry_name,
+                  const std::unordered_map<std::string, std::string> &params)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return true;
+}
+
+bool db_get_entry(const std::string &entry_type, const std::string &entry_name,
+                  std::unordered_map<std::string, std::string> &params)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return true;
+}
+
+bool db_get_entries_by_type(
+    const std::string &entry_type,
+    std::unordered_map<std::string, std::unordered_map<std::string, std::string>> &nested_params)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return true;
+}
+
+bool db_remove_entry(const std::string &entry_type, const std::string &entry_name)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return true;
+}
+
+} // namespace bpl
+} // namespace beerocks
--- a/framework/platform/bpl/uci/db/bpl_db.cpp
+++ b/framework/platform/bpl/uci/db/bpl_db.cpp
@@ -1,0 +1,62 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "bpl/bpl_db.h"
+
+#include <mapf/common/logger.h>
+
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Implementation ///////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+namespace beerocks {
+namespace bpl {
+
+bool db_has_entry(const std::string &entry_type, const std::string &entry_name)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return false;
+}
+
+bool db_add_entry(const std::string &entry_type, const std::string &entry_name,
+                  const std::unordered_map<std::string, std::string> &params)
+{
+    LOG(TRACE) << " - NOT IMPLEMENTED!";
+    return true;
+}
+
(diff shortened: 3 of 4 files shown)


framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,90 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2016-2020 the prplMesh contributors (see AUTHORS.md)
```
kantera800:
```suggestion
 * SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
```

framework/platform/bpl/uci/db/bpl_db.cpp
```diff
@@ -0,0 +1,57 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2019-2020 the prplMesh contributors (see AUTHORS.md)
```
kantera800: same

kantera800 (changes requested):
you should have a bit more description in the commit message for these stubs

framework/platform/bpl/include/bpl/bpl_db.h
itayx: Fixed

framework/platform/bpl/uci/db/bpl_db.cpp
itayx: Fixed

kantera800 (approved):
approved. 
just change to `return true` in all stubs (we don't want to fail the dummy tests).

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,97 @@
...
+bool db_remove_param(const std::string &path, const std::string &object);
+
+/**
+ * @brief Get all sub-entries of a parent path.
+ * 
+ * Get all sub-entries of a given parent path in a JSON styled string object
+ * {entry1: val1,entry2: val2,...}
+ * 
+ * @param path name of the requested section.
+ * @param values [out] JSON styled string object.
```
adam1985d: please remove any indication to json in the comments and implementation
1. if we eventually use JSON - the BPL is unaware of it
2. the returned value in this example is a map and not a string

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,97 @@
...
+ * @param path name of the requested section.
+ * @return true on success, false otherwise.
+ */
+bool db_remove_param(const std::string &path, const std::string &object);
+
+/**
+ * @brief Get all sub-entries of a parent path.
+ * 
+ * Get all sub-entries of a given parent path in a JSON styled string object
+ * {entry1: val1,entry2: val2,...}
```
adam1985d: see my other comment regarding JSON

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * @brief Add parameter to the presistent DB
+ * 
+ * @param path name of the requested section.
+ * @param object full name of the parameter.
+ * @param val Initial value of the parameter.
+ * @return true on success, false otherwise.
+ */
+bool db_add_param(const std::string &path, const std::string &object, const std::string &val);
```
adam1985d: it's ok you didn't waste time on it now - but before this PR is merged, each API must have its own separate documentation

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * @param path name of the requested section.
+ * @return true on success, false otherwise.
+ */
+bool db_clear_sub_entries(const std::string &path);
+
+} // namespace bpl
+} // namespace beerocks
+
+#endif /* _BPL_CFG_H_ */
```
adam1985d: replace `_BPL_CFG_H_` with `_BPL_DB_H_`

framework/platform/bpl/linux/bpl_db.cpp
```diff
@@ -0,0 +1,75 @@
...
+//////////////////////////////////////////////////////////////////////////////
+/////////////////////////////// Implementation ///////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+namespace beerocks {
+namespace bpl {
+bool db_add_param(const std::string &path, const std::string &object, const std::string &val)
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED!";
+    return false;
```
adam1985d: # for all stubs #
please return true,
I know this hides the point that this function is called but isn't implemented but we are going to add the implementation now so "true" should be returned.

framework/platform/bpl/uci/db/bpl_db.cpp
```diff
@@ -0,0 +1,75 @@
...
+    return false;
+}
+
+bool db_set_param(const std::string &path, const std::string &object, const std::string &val)
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED!";
+    return false;
+}
+
+bool db_set_param(const std::string &path, const std::string &object, const int &val)
```
adam1985d: for simple data types `set` operation no need for `const ref&` - a copy is ok

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,97 @@
...
+ * @return true on success, false otherwise.
+ */
+bool db_add_param(const std::string &path, const std::string &object, const std::string &val);
+bool db_add_param(const std::string &path, const std::string &object, const int &val);
+/**
+ * @brief Set a parameter given a specific path.
+ * 
+ * @param path name of the requested section.
+ * @param object full name of the parameter.
+ * @param val [in] value of the given parameter.
```
morantr: For const reference need to write the oxygen like this;
`@param[in] param_name` and not `@param param_name [in]`

Also, it is required to add the direction for all  **non-const** references, so here it is not required.
Anyway, if you decide to do it also to **const** reference arguments, it is probably best if you do it on all of them in the file.

This comment apply to all the changes you've done in this commit.

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,97 @@
...
+/****************************************************************************/
+/******************************* Structures *********************************/
+/****************************************************************************/
+
+/****************************************************************************/
+/******************************** Functions *********************************/
+/****************************************************************************/
+
+/**
+ * @brief Add parameter to the presistent DB
```
morantr: Missing dot at sentence end.

adam1985d (changes requested):
the PR is mostly OK but please address each of the comments I provided and let me re-review

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,97 @@
...
+ * 
+ * @param path name of the requested section.
+ * @param object full name of the parameter.
+ * @param val Initial value of the parameter.
+ * @return true on success, false otherwise.
+ */
+bool db_add_param(const std::string &path, const std::string &object, const std::string &val);
+bool db_add_param(const std::string &path, const std::string &object, const int &val);
```
morantr: There is no point passing `int` to a function as a const reference.  We pass parameters as const reference only if it's size is or might be greater than sizeof(int), which in our case is 4 bytes.
Apply to other functions in the file as well.

morantr (approved):

framework/platform/bpl/include/bpl/bpl_db.h
itayx: Updated the API in question & doxygen

framework/platform/bpl/include/bpl/bpl_db.h
itayx: Reworked the API in question, now it receives an unordered_map<string, string>

framework/platform/bpl/include/bpl/bpl_db.h
itayx: changed

framework/platform/bpl/linux/bpl_db.cpp
itayx: Updated

framework/platform/bpl/uci/db/bpl_db.cpp
itayx: Changed to use an unordered_map<string, string>, therefore using a reference is better since we're working with a container and do not want to copy the whole thing

framework/platform/bpl/include/bpl/bpl_db.h
itayx: Removed all mentions to JSON

framework/platform/bpl/include/bpl/bpl_db.h
itayx: Done

framework/platform/bpl/include/bpl/bpl_db.h
itayx: This is no longer relevant, but I understood the point, thanks.

framework/platform/bpl/include/bpl/bpl_db.h
itayx: updated the oxygen according to the rest of the BPL apis

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,87 @@
...
+/****************************************************************************/
+/******************************** Functions *********************************/
+/****************************************************************************/
+
+/**
+ * @brief Add new entry with initial value.
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @param[in] values initial values of the entry.
```
mariomaz: Please document what are the keys and values for the map.
If keys are the entry name then the function should accept a vector and not a map

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,87 @@
...
+ * @return true on success, false otherwise
+ */
+bool db_add_entry(const std::string &entry_type, const std::string &entry_name,
+                  const std::unordered_map<std::string, std::string> &values);
+/**
+ * @brief Set values in entry, append attributes and override existing.
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @param[in] values unordered_map containing the values the requested entry, to be set.
```
mariomaz: values should be a vector, not a map.

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,87 @@
...
+ * @return true on success, false otherwise.
+ */
+bool db_set_entry_params(const std::string &entry_type, const std::string &entry_name,
```
mariomaz: The names of the API functions shall be consistent. I mean if `db_add_entry` is used, then this function shall be `db_set_entry`

framework/platform/bpl/include/bpl/bpl_db.h
```diff
@@ -0,0 +1,87 @@
...
+                         const std::unordered_map<std::string, std::string> &values);
+/**
+ * @brief Get values of entry, return all avaliable values.
+ * 
+ * @param[in] entry_type type of the requested entry.
+ * @param[in] entry_name name of the requested entry.
+ * @param [in/out] values unordered_map containing the values of the requested entry.
+ * @return true on success, false otherwise.
+ */
+bool db_get_entry_params(const std::string &entry_type, const std::string &entry_name,
```
mariomaz: values should be a vector, not a map.

mariomaz: The names of the API functions shall be consistent. I mean if `db_add_entry` is used, then this function shall be `db_get_entry`

framework/platform/bpl/include/bpl/bpl_db.h
mariomaz: What is the `entry_type` used for? If the `entry_name` is unique then no other value is required for disambiguation to look up the entry.
This comment applies to the rest of functions below too.

framework/platform/bpl/linux/bpl_db.cpp
```diff
@@ -0,0 +1,56 @@
...
+/////////////////////////////// Implementation ///////////////////////////////
+//////////////////////////////////////////////////////////////////////////////
+
+namespace beerocks {
+namespace bpl {
+
+bool db_add_entry(const std::string &entry_type, const std::string &entry_name,
+                  const std::unordered_map<std::string, std::string> &values)
+{
+    LOG(TRACE) << __func__ << " - NOT IMPLEMENTED!";
```
mariomaz: I think `__func__` is not needed because `LOG` is already writing the function name

mariomaz (approved):
Choosing sensible names for API functions can drastically smooth out the learning curve for new developers, helping them intuitively know what to look for and where to find it.

Consistency is key when naming functions and parameters. No matter how well implemented the API is, your API will always feel clumsy if names are inconsistent. .

Last but not least, choose the most appropriate parameter types in each occasion. 

My comments are very easy to fix in case you want to do it, so approving.

framework/platform/bpl/linux/bpl_db.cpp
itayx: removed

itayx:
@mariomaz I have a few notes regarding your comments:
- The `entry_name` is only partly unique and the only restriction is that entries of the same type are forced to use unique name.
- Each entry will contain parameters within in a pair `<name,value>`.
And each entry will probably contain multiple parameters which are differentiated by their parameter name
Therefore to use the DB an unordered_map is needed, though I understand your point to the argument naming
So I'm changing the "values" argument to "param" so it will be more sensible

framework/platform/bpl/include/bpl/bpl_db.h
itayx: see my comment https://github.com/prplfoundation/prplMesh/pull/1460#issuecomment-651028711

framework/platform/bpl/include/bpl/bpl_db.h
itayx: see my comment https://github.com/prplfoundation/prplMesh/pull/1460#issuecomment-651028711

itayx: see my comment https://github.com/prplfoundation/prplMesh/pull/1460#issuecomment-651028711

framework/platform/bpl/include/bpl/bpl_db.h
itayx: see my comment https://github.com/prplfoundation/prplMesh/pull/1460#issuecomment-651028711

framework/platform/bpl/include/bpl/bpl_db.h
itayx: see my comment https://github.com/prplfoundation/prplMesh/pull/1460#issuecomment-651028711

framework/platform/bpl/include/bpl/bpl_db.h
itayx: see my comment https://github.com/prplfoundation/prplMesh/pull/1460#issuecomment-651028711

adam1985d (approved):

PR #1465, opened 2020-06-22
hotfix! Change bssid type for roam() method

VladyslavTupikin:
All bssid which passes as method parameters
should be with type sMacAddr instead string.

roam() method will be used in the Backhaul
STA Steering impelemntation. That activity
triggered to do hotfix for the bssid type.

Also fixed roam() call in the backhaul manager:
tranformed give argument from string to the
sMacAddr structure.


arnout (approved):
Except for the build failure, looks good to me.

VladyslavTupikin:
> Except for the build failure, looks good to me.

Fixed.

mariomaz (approved):
LGTM

Issue #1381, opened 2020-06-02
[TASK] Integrate boardfarm in gitlab-ci


arnout:
The script `tests/run_bf.sh` runs boardfarm. This should also be done from CI.

- [x] Create a Dockerfile with boardfarm
- [x] Build this in CI
- [x] Run boardfarm in CI
- [ ] Run boardfarm on prpWrt device (#1072) in CI

For the latter, the docker in CI has to be able to bridge to the external network. It needs to be investigated how to do that with a gitlab-runner.

PR #1467, opened 2020-06-22
refactor ap_metrics

RanRegev:
refactored ap metrics query and response by collecting all responses together before sending to the controller

### DIFF
--- a/agent/src/beerocks/slave/CMakeLists.txt
+++ b/agent/src/beerocks/slave/CMakeLists.txt
@@ -11,6 +11,7 @@ file(GLOB beerocks_agent_sources
     ${MODULE_PATH}/link_metrics/*.c*
     ${MODULE_PATH}/platform_manager/*.c*
     ${MODULE_PATH}/gate/*.c*
+    ${MODULE_PATH}/backhaul_manager/internal/*.c*
     ${MODULE_PATH}/*.c*
 )
 
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -29,19 +29,17 @@
  */
 
 #include "backhaul_manager_thread.h"
-
 #include "../link_metrics/ieee802_11_link_metrics_collector.h"
 #include "../link_metrics/ieee802_3_link_metrics_collector.h"
 #include "../tlvf_utils.h"
-
+#include "internal/expected_ap_metrics_response.h"
 #include <bcl/beerocks_utils.h>
 #include <bcl/son/son_wireless_utils.h>
-#include <easylogging++.h>
-
 #include <beerocks/tlvf/beerocks_message.h>
 #include <beerocks/tlvf/beerocks_message_backhaul.h>
 #include <beerocks/tlvf/beerocks_message_control.h>
 #include <beerocks/tlvf/beerocks_message_platform.h>
+#include <easylogging++.h>
 
 /*
  * TODO:
@@ -1023,21 +1021,16 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
                 for (const auto &socket : slaves_sockets) {
                     if (socket) {
                         for (int i = 0; i < beerocks::IFACE_TOTAL_VAPS; ++i) {
-                            if (socket->vaps_list.vaps[i].mac != network_utils::ZERO_MAC) {
+                            if ((socket->vaps_list.vaps[i].mac != network_utils::ZERO_MAC) &&
+                                (socket->vaps_list.vaps[i].ssid[0] != '\0')) {
                                 bssid_list.push_back(socket->vaps_list.vaps[i].mac);
                             }
                         }
                     }
                 }
-                // We must generate a new MID for the periodic AP Metrics Response messages that
-                // do not correspond to an AP Metrics Query message.
-                // We cannot set MID to 0 here because we must also differentiate periodic
-                // AP Metrics Response messages and messages received from monitor thread
-                // due to channel utilization crossed configured threshold value.
-                // As a temporary solution, set MID to UINT16_MAX here.
-                // TODO: to be fixed as part of #1328
+
                 if (!bssid_list.empty()) {
-                    send_slave_ap_metric_query_message(UINT16_MAX, bssid_list);
+                    send_slave_ap_metric_query_message(0, bssid_list);
                 } else {
                     LOG(DEBUG) << "Skipping AP_METRICS_QUERY for slave, empty BSSID list";
                 }
@@ -1258,49 +1251,81 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
 bool backhaul_manager::send_slave_ap_metric_query_message(uint16_t mid,
                                                           const std::vector<sMacAddr> &bssid_list)
 {
-    bool ret = false;
+    // save mid and clear previous query
+    m_expected_ap_metrics_response.reset_to_new_mid(mid);
+
+    // compare two mac addresses using string compare on their octets
+    auto mac_comp = [](const sMacAddr &addr1, const sMacAddr &addr2) -> bool {
+        return memcmp((char *)addr1.oct, (char *)addr2.oct, sizeof(sMacAddr)) < 0;
+    };
+
+    // sort the given bssid list
+    std::vector<sMacAddr> given_bssid_list = bssid_list;
+    std::sort(given_bssid_list.begin(), given_bssid_list.end(), mac_comp);
+
+    // for each socket - send only relevant bssids
+    auto ret = true;
     for (auto socket : slaves_sockets) {
-        for (const auto &mac : bssid_list) {
-            int i = 0;
-            if (mac == socket->vaps_list.vaps[i].mac) {
-                LOG(DEBUG) << "Forwarding AP_METRICS_QUERY_MESSAGE message to son_slave, bssid: "
-                           << std::hex << tlvf::mac_to_string(mac);
 
-                auto forward =
-                    cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
-                if (!forward) {
-                    LOG(ERROR) << "Failed to create AP_METRICS_QUERY_MESSAGE";
-                    return false;
-                }
+        // create the message to forward to current slaves
+        auto forward = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_QUERY_MESSAGE);
+        if (!forward) {
+            LOG(ERROR) << "Failed to create AP_METRICS_QUERY_MESSAGE";
+            return false;
+        }
 
-                auto query = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
-                if (!query) {
-                    LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetricQuery>";
-                    return false;
-                }
+        // copy current vaps bssid
+        std::vector<sMacAddr> current_socket_bssids;
+        std::transform(socket->vaps_list.vaps, socket->vaps_list.vaps + beerocks::IFACE_TOTAL_VAPS,
+                       std::back_inserter(current_socket_bssids),
+                       [](const beerocks_message::sVapInfo &vap_info) { return vap_info.mac; });
 
-                if (!query->alloc_bssid_list(1)) {
-                    LOG(ERROR) << "Failed allocate memory for bssid_list";
-                    return false;
-                }
+        // sort current vaps bssid
+        std::sort(current_socket_bssids.begin(), current_socket_bssids.end(), mac_comp);
 
-                auto list         = query->bssid_list(0);
-                std::get<0>(list) = true;
-                std::get<1>(list) = mac;
+        // fill a list of the common bssids: given list with current vaps
+        std::vector<sMacAddr> required_bssids;
+        std::set_intersection(given_bssid_list.begin(), given_bssid_list.end(),
+                              current_socket_bssids.begin(), current_socket_bssids.end(),
+                              std::back_inserter(required_bssids), mac_comp);
 
-                if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
-                    LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
-                    ret = false;
-                    continue;
-                } else {
-                    ret = true;
-                    // Fill a query vector
-                    m_ap_metric_query.push_back({socket->slave, mac});
-                }
-            }
+        if (required_bssids.empty()) {
+            // skip if there is nothing to do
+            LOG(DEBUG) << "empty required bssid list, skipping";
+            continue;
+        }
+
+        // add metric query tlv
+        auto query = cmdu_tx.addClass<wfa_map::tlvApMetricQuery>();
+        if (!query) {
+            LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetricQuery>";
+            return false;
+        }
+
+        // alocate memory
+        if (!query->alloc_bssid_list(required_bssids.size())) {
+            LOG(ERROR) << "Failed allocate memory for bssid_list, size: " << bssid_list.size();
+            return false;
+        }
+
+        int i = 0;
+        for (const auto &mac : required_bssids) {
+            auto list         = query->bssid_list(i);
+            std::get<0>(list) = true;
+            std::get<1>(list) = mac;
             i++;
         }
+
+        // save the required list
+        m_expected_ap_metrics_response.add_expected_bssid(required_bssids.begin(),
+                                                          required_bssids.end());
+
+        if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
+            LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
+            ret = false;
+        }
     }
+
     return ret;
 }
 
@@ -2554,8 +2579,12 @@ bool backhaul_manager::handle_ap_capability_query(ieee1905_1::CmduMessageRx &cmd
 bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_rx,
                                                const std::string &src_mac)
 {
-    std::vector<sMacAddr> bssid;
-    const auto mid           = cmdu_rx.getMessageId();
+
+    const auto mid = cmdu_rx.getMessageId();
+
+    // save mid and erase data for previous query and response
+    m_expected_ap_metrics_response.reset_to_new_mid(mid);
+
     auto ap_metric_query_tlv = cmdu_rx.getClass<wfa_map::tlvApMetricQuery>();
     if (!ap_metric_query_tlv) {
         LOG(ERROR) << "AP Metrics Query CMDU mid=" << mid << " does not have AP Metric Query TLV";
@@ -2567,14 +2596,27 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
             LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
             return false;
         }
-        bssid.push_back(std::get<1>(bssid_tuple));
+
+        m_expected_ap_metrics_response.add_expected_bssid(std::get<1>(bssid_tuple));
+
         LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
                    << "  bssid " << std::get<1>(bssid_tuple);
     }
 
-    if (!send_slave_ap_metric_query_message(mid, bssid)) {
-        LOG(ERROR) << "Failed to forward AP_METRICS_RESPONSE to the son_slave_thread";
+    LOG(DEBUG) << "Forwarding AP_METRICS_QUERY_MESSAGE to all slaves, mid=" << std::hex << int(mid);
+    auto header = message_com::get_uds_header(cmdu_rx);
+    if (!header) {
+        LOG(ERROR) << "Failed to extract header from rx message";
         return false;
+    }
+
+    uint16_t length = header->length;
+    cmdu_rx.swap(); // swap back before forwarding
+    for (const auto &soc : slaves_sockets) {
+        if (!message_com::forward_cmdu_to_uds(soc->slave, cmdu_rx, length)) {
+            LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to slave: "
+                       << tlvf::mac_to_string(soc->radio_mac);
+        }
     }
 
     return true;
@@ -2584,32 +2626,22 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
                                                         const std::string &src_mac)
 {
     auto mid = cmdu_rx.getMessageId();
-    LOG(DEBUG) << "Received AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
 
-    /**
-     * If AP Metrics Response message does not correspond to a previously received and forwarded
-     * AP Metrics Query message (which we know because message id is not set), then forward message
-     * to controller.
-     * This might happen when channel utilization value has crossed configured threshold or when
-     * periodic metrics reporting interval has elapsed.
-     */
     if (0 == mid) {
-        uint16_t length = message_com::get_uds_header(cmdu_rx)->length;
-        cmdu_rx.swap(); //swap back before forwarding
-        return send_cmdu_to_bus(cmdu_rx, controller_bridge_mac, bridge_info.mac, length);
+        LOG(DEBUG) << "AP_METRICS_RESPONSE_MESSAGE. mid is zero (0), forwarding to controller "
+                      "unconditionaly.";
+        cmdu_rx.swap();
+        return send_cmdu_to_bus(cmdu_rx, controller_bridge_mac, bridge_info.mac,
+                                cmdu_rx.getMessageLength());
     }
 
-    /**
-     * When periodic metrics reporting interval has elapsed, we emulate that we have received an
-     * AP Metrics Query message from controller. To differentiate real queries from emulated ones,
-     * we use a "special" mid value.
-     * Note that this design is flaw as a real query might also have this special mid value. This
-     * is just a quick and dirty fix to pass 4.7.5 and 4.7.6 for M1
-     * TODO: to be fixed as part of #1328
-     */
-    if (UINT16_MAX == mid) {
-        mid = 0;
+    if (m_expected_ap_metrics_response.get_mid() != mid) {
+        LOG(ERROR) << "Received AP_METRICS_RESPONSE_MESSAGE with bad mid: existing:" << std::hex
+                   << m_expected_ap_metrics_response.get_mid() << " received: " << int(mid);
+        return false;
     }
+
+    LOG(DEBUG) << "Received AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
 
     auto ap_metrics_tlv = cmdu_rx.getClass<wfa_map::tlvApMetrics>();
     if (!ap_metrics_tlv) {
@@ -2618,44 +2650,64 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
         return false;
     }
 
-    auto bssid_tlv = ap_metrics_tlv->bssid();
-    auto mac       = std::find_if(
-        m_ap_metric_query.begin(), m_ap_metric_query.end(),
-        [&bssid_tlv](sApMetricsQuery const &query) { return query.bssid == bssid_tlv; });
-
-    if (mac == m_ap_metric_query.end()) {
-        LOG(ERROR) << "Failed search in ap_metric_query for bssid: " << bssid_tlv
+    if (!m_expected_ap_metrics_response.find_expected_bssid(ap_metrics_tlv->bssid())) {
+        LOG(ERROR) << "Failed search in ap_metric_query for bssid: " << ap_metrics_tlv->bssid()
                    << " from mid=" << std::hex << int(mid);
         return false;
     }
 
-    sApMetrics metric;
-    // Copy data to the response vector
-    metric.bssid               = ap_metrics_tlv->bssid();
-    metric.channel_utilization = ap_metrics_tlv->channel_utilization();
-    metric.number_of_stas_currently_associated =
-        ap_metrics_tlv->number_of_stas_currently_associated();
-    metric.estimated_service_parameters = ap_metrics_tlv->estimated_service_parameters();
-    auto info                           = ap_metrics_tlv->estimated_service_info_field();
-    for (size_t i = 0; i < ap_metrics_tlv->estimated_service_info_field_length(); i++) {
-        metric.estimated_service_info_field.push_back(info[i]);
+    // prepare ap metrics tlv
+    auto &ap_metrics_tx_message  = m_expected_ap_metrics_response.create_tx_message();
+    auto ap_metrics_response_tlv = ap_metrics_tx_message.addClass<wfa_map::tlvApMetrics>();
+    if (!ap_metrics_response_tlv) {
+        LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetrics>";
+        return false;
     }
-    std::vector<sStaTrafficStats> traffic_stats_response;
+    ap_metrics_response_tlv->bssid()               = ap_metrics_tlv->bssid();
+    ap_metrics_response_tlv->channel_utilization() = ap_metrics_tlv->channel_utilization();
+    ap_metrics_response_tlv->number_of_stas_currently_associated() =
+        ap_metrics_tlv->number_of_stas_currently_associated();
+    ap_metrics_response_tlv->estimated_service_parameters() =
+        ap_metrics_tlv->estimated_service_parameters();
 
+    if (!ap_metrics_response_tlv->alloc_estimated_service_info_field(
+            ap_metrics_tlv->estimated_service_info_field_length())) {
+        LOG(ERROR) << "Couldn't allocate "
+                      "ap_metrics_response_tlv->alloc_estimated_service_info_field";
+        return false;
+    }
+    auto rx_info = ap_metrics_tlv->estimated_service_info_field();
+    auto tx_info = ap_metrics_response_tlv->estimated_service_info_field();
+    for (size_t i = 0; i < ap_metrics_tlv->estimated_service_info_field_length(); i++) {
+        tx_info[i] = rx_info[i];
+    }
+
+    // Prepare STA Traffic Stats TLV's
     for (auto &sta_traffic : cmdu_rx.getClassList<wfa_map::tlvAssociatedStaTrafficStats>()) {
         if (!sta_traffic) {
             LOG(ERROR) << "Failed to get class list for tlvAssociatedStaTrafficStats";
             continue;
         }
 
-        traffic_stats_response.push_back(
-            {sta_traffic->sta_mac(), sta_traffic->byte_sent(), sta_traffic->byte_recived(),
-             sta_traffic->packets_sent(), sta_traffic->packets_recived(),
-             sta_traffic->tx_packets_error(), sta_traffic->rx_packets_error(),
-             sta_traffic->retransmission_count()});
+        auto sta_traffic_response_tlv =
+            ap_metrics_tx_message.addClass<wfa_map::tlvAssociatedStaTrafficStats>();
+
+        if (!sta_traffic_response_tlv) {
+            LOG(ERROR) << "Failed addClass<wfa_map::tlvAssociatedStaTrafficStats>";
+            continue;
+        }
+
+        sta_traffic_response_tlv->sta_mac()              = sta_traffic->sta_mac();
+        sta_traffic_response_tlv->byte_sent()            = sta_traffic->byte_sent();
+        sta_traffic_response_tlv->byte_recived()         = sta_traffic->byte_recived();
+        sta_traffic_response_tlv->packets_sent()         = sta_traffic->packets_sent();
+        sta_traffic_response_tlv->packets_recived()      = sta_traffic->packets_recived();
+        sta_traffic_response_tlv->tx_packets_error()     = sta_traffic->tx_packets_error();
+        sta_traffic_response_tlv->rx_packets_error()     = sta_traffic->rx_packets_error();
+        sta_traffic_response_tlv->retransmission_count() = sta_traffic->retransmission_count();
     }
 
-    std::vector<sStaLinkMetrics> link_metrics_response;
+    // Prepare STA Link Metrics TLV's
     for (auto &sta_link_metric : cmdu_rx.getClassList<wfa_map::tlvAssociatedStaLinkMetrics>()) {
         if (!sta_link_metric) {
             LOG(ERROR) << "Failed getClassList<wfa_map::tlvAssociatedStaLinkMetrics>";
@@ -2666,98 +2718,35 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
             continue;
         }
         auto response_list = sta_link_metric->bssid_info_list(0);
-        link_metrics_response.push_back({sta_link_metric->sta_mac(), std::get<1>(response_list)});
+
+        auto sta_link_metric_response_tlv =
+            ap_metrics_tx_message.addClass<wfa_map::tlvAssociatedStaLinkMetrics>();
+
+        if (!sta_link_metric_response_tlv) {
+            LOG(ERROR) << "Failed addClass<wfa_map::tlvAssociatedStaLinkMetrics>";
+            continue;
+        }
+
+        sta_link_metric_response_tlv->sta_mac() = sta_link_metric->sta_mac();
+        if (!sta_link_metric_response_tlv->alloc_bssid_info_list(1)) {
+            LOG(ERROR) << "Failed alloc_bssid_info_list";
+            continue;
+        }
+        auto &sta_link_metric_response =
+            std::get<1>(sta_link_metric_response_tlv->bssid_info_list(0));
+        sta_link_metric_response = std::get<1>(response_list);
     }
 
-    // Fill a response vector
-    m_ap_metric_response.push_back({metric, traffic_stats_response, link_metrics_response});
+    // remove an entry from the processed query
+    m_expected_ap_metrics_response.remove_expected_bssid(ap_metrics_tlv->bssid());
 
-    // Remove an entry from the processed query
-    m_ap_metric_query.erase(
-        std::remove_if(m_ap_metric_query.begin(), m_ap_metric_query.end(),
-                       [&](sApMetricsQuery const &query) { return mac->bssid == query.bssid; }),
-        m_ap_metric_query.end());
-
-    if (!m_ap_metric_query.empty()) {
+    // waiting for all responses before sending the response
+    if (!m_expected_ap_metrics_response.is_expected_bssid_empty()) {
         return true;
     }
 
-    // We received all responses - prepare and send response message to the controller
-    auto cmdu_header = cmdu_tx.create(mid, ieee1905_1::eMessageType::AP_METRICS_RESPONSE_MESSAGE);
-
-    if (!cmdu_header) {
-        LOG(ERROR) << "Failed building IEEE1905 AP_METRICS_RESPONSE_MESSAGE";
-        return false;
-    }
-
-    // Prepare tlvApMetrics for each processed query
-    for (const auto &response : m_ap_metric_response) {
-        auto ap_metrics_response_tlv = cmdu_tx.addClass<wfa_map::tlvApMetrics>();
-        if (!ap_metrics_response_tlv) {
-            LOG(ERROR) << "Failed addClass<wfa_map::tlvApMetrics>";
-            return false;
-        }
-
-        ap_metrics_response_tlv->bssid()               = response.metric.bssid;
-        ap_metrics_response_tlv->channel_utilization() = response.metric.channel_utilization;
-        ap_metrics_response_tlv->number_of_stas_currently_associated() =
-            response.metric.number_of_stas_currently_associated;
-        ap_metrics_response_tlv->estimated_service_parameters() =
-            response.metric.estimated_service_parameters;
-        if (!ap_metrics_response_tlv->alloc_estimated_service_info_field(
-                response.metric.estimated_service_info_field.size())) {
-            LOG(ERROR) << "Couldn't allocate "
-                          "ap_metrics_response_tlv->alloc_estimated_service_info_field";
-            return false;
-        }
-        std::copy_n(response.metric.estimated_service_info_field.begin(),
-                    response.metric.estimated_service_info_field.size(),
-                    ap_metrics_response_tlv->estimated_service_info_field());
-
-        for (auto &stat : response.sta_traffic_stats) {
-            auto sta_traffic_response_tlv =
-                cmdu_tx.addClass<wfa_map::tlvAssociatedStaTrafficStats>();
-
-            if (!sta_traffic_response_tlv) {
-                LOG(ERROR) << "Failed addClass<wfa_map::tlvAssociatedStaTrafficStats>";
-                continue;
-            }
-
-            sta_traffic_response_tlv->sta_mac()              = stat.sta_mac;
-            sta_traffic_response_tlv->byte_sent()            = stat.byte_sent;
-            sta_traffic_response_tlv->byte_recived()         = stat.byte_recived;
-            sta_traffic_response_tlv->packets_sent()         = stat.packets_sent;
-            sta_traffic_response_tlv->packets_recived()      = stat.packets_recived;
-            sta_traffic_response_tlv->tx_packets_error()     = stat.tx_packets_error;
-            sta_traffic_response_tlv->rx_packets_error()     = stat.rx_packets_error;
-            sta_traffic_response_tlv->retransmission_count() = stat.retransmission_count;
-        }
-
-        for (auto &link_metric : response.sta_link_metrics) {
-            auto sta_link_metric_response_tlv =
-                cmdu_tx.addClass<wfa_map::tlvAssociatedStaLinkMetrics>();
-
-            if (!sta_link_metric_response_tlv) {
-                LOG(ERROR) << "Failed addClass<wfa_map::tlvAssociatedStaLinkMetrics>";
-                continue;
-            }
-
-            sta_link_metric_response_tlv->sta_mac() = link_metric.sta_mac;
-            if (!sta_link_metric_response_tlv->alloc_bssid_info_list(1)) {
-                LOG(ERROR) << "Failed alloc_bssid_info_list";
-                continue;
-            }
-            auto &sta_link_metric_response =
-                std::get<1>(sta_link_metric_response_tlv->bssid_info_list(0));
-            sta_link_metric_response = link_metric.bssid_info;
-        }
-    }
-
-    // Clear the m_ap_metric_response vector after preparing response to the controller
-    m_ap_metric_response.clear();
-
     LOG(DEBUG) << "Sending AP_METRICS_RESPONSE_MESSAGE, mid=" << std::hex << int(mid);
-    return send_cmdu_to_bus(cmdu_tx, controller_bridge_mac, bridge_info.mac);
+    return send_cmdu_to_bus(ap_metrics_tx_message, controller_bridge_mac, bridge_info.mac);
 }
 
 /**
--- a/agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.cpp
@@ -1,0 +1,46 @@
+
+#include "expected_ap_metrics_response.h"
+
+namespace beerocks {
+
+// mid
+void ExpectedApMetricsResponse::reset_to_new_mid(uint16_t mid)
+{
+    m_mid = mid;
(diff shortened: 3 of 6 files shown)


agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,51 +1261,101 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+    // compare two mac addresses using thier hash values
+    auto mac_comp = [](const sMacAddr &addr1, const sMacAddr &addr2) -> bool {
+        return std::hash<sMacAddr>()(addr1) < std::hash<sMacAddr>()(addr2);
+    };
+
+    // sort the given bssid list
+    std::vector<sMacAddr> given_bssid_list = bssid_list;
+    std::sort(given_bssid_list.begin(), given_bssid_list.end(), mac_comp);
+
+    // debug
```
vitalybu: Not sure if this PR should be reviewed or not, as there's a lot of DEBUG code in here...

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,51 +1261,101 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
-                    m_ap_metric_query.push_back({socket->slave, mac});
-                }
-            }
+        // fill a list of the common bssids: given list with current vaps
+        std::vector<sMacAddr> required_bssids;
+        std::set_intersection(given_bssid_list.begin(), given_bssid_list.end(),
+                              current_socket_bssids.begin(), current_socket_bssids.end(),
+                              std::back_inserter(required_bssids), mac_comp);
+
+        // debug
```
VladyslavTupikin: Please, remove debug code.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2585,14 +2632,22 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
             LOG(ERROR) << "Failed to get bssid " << bssid_idx << " from AP_METRICS_QUERY";
             return false;
         }
-        bssid.push_back(std::get<1>(bssid_tuple));
+
+        // #1421 implementation
```
VladyslavTupikin: Do we need this commentary?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2602,78 +2657,88 @@ bool backhaul_manager::handle_slave_ap_metrics_response(ieee1905_1::CmduMessageR
...
                    << " from mid=" << std::hex << int(mid);
         return false;
     }
 
-    sApMetrics metric;
-    // Copy data to the response vector
-    metric.bssid               = ap_metrics_tlv->bssid();
-    metric.channel_utilization = ap_metrics_tlv->channel_utilization();
-    metric.number_of_stas_currently_associated =
+    // prepare ap metrics tlv
```
VladyslavTupikin, line 2659: I think better to create the separate method for creation ap metrics response with al necessary TLV's

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.cpp
```diff
@@ -0,0 +1,42 @@
+
+#include "expected_ap_metrics_response.h"
+
+// mid
+void ExpectedApMetricsResponse::reset_to_new_mid(uint16_t mid)
```
VladyslavTupikin, line 7: It will be good to add Doxygen header for each method.

VladyslavTupikin (dismissed):
combined_infra_metrics starting
 Starting tcpdump, output file test_combined_infra_metrics.pcap
 Capturing on 'br-a8d5d8787c22'
 ...
 Checking for CMDU ACK (0x8000) from 0e:0c:da:fb:f8:be
 Checking for CMDU AP metrics response (0x800c) from 0e:0c:da:fb:f8:be
 Check AP Metrics Response message has AP Metrics TLV
 Check AP metrics response has STA Link Metrics
 **FAIL: STA metrics with wrong MAC 51:a1:10:20:00:01 instead of 51:a1:10:20:00:09**
 ...
 Terminating tcpdump
 combined_infra_metrics failed

Please, take a look on the results of test_flows.py in CI.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: yes, my bad.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1025,21 +1023,16 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
...
                         }
                     }
                 }
-                // We must generate a new MID for the periodic AP Metrics Response messages that
-                // do not correspond to an AP Metrics Query message.
-                // We cannot set MID to 0 here because we must also differentiate periodic
-                // AP Metrics Response messages and messages received from monitor thread
-                // due to channel utilization crossed configured threshold value.
-                // As a temporary solution, set MID to UINT16_MAX here.
-                // TODO: to be fixed as part of #1328
```
vitalybu, line 1038: What's different now that we can use 0 as MID?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
 bool backhaul_manager::send_slave_ap_metric_query_message(uint16_t mid,
                                                           const std::vector<sMacAddr> &bssid_list)
 {
-    bool ret = false;
+    // save mid and clear previous query
+    m_expected_ap_metrics_response.reset_to_new_mid(mid);
+
+    // compare two mac addresses using thier hash values
+    auto mac_comp = [](const sMacAddr &addr1, const sMacAddr &addr2) -> bool {
+        return std::hash<sMacAddr>()(addr1) < std::hash<sMacAddr>()(addr2);
```
vitalybu: Hashing the entire struct just for the purpose of comparison is highly inefficient.
Not to mention that `std::hash` is entirely implementation specific, and there are no guaranties about the result value, so I'm not sure what you achieve here with the `<` operator.

If you need this only for the purpose of sorting MAC addresses, I think using `strcmp()` on `sMacAddr.oct` is the fastest and the safest way to achieve this.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+    };
+
+    // sort the given bssid list
+    std::vector<sMacAddr> given_bssid_list = bssid_list;
+    std::sort(given_bssid_list.begin(), given_bssid_list.end(), mac_comp);
```
vitalybu, line 1264: Any specific reason you need this sorted?
`std::set_intersection` constructs a **sorted** range consisting of elements that are found in both sorted ranges.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
             i++;
         }
+
+        // save the required list
+        m_expected_ap_metrics_response.add_expected_bssid(required_bssids.begin(),
+                                                          required_bssids.end());
+
+        if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
+            LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
+            ret = false;
```
vitalybu, line 1325: If the next iteration of the loop will succeed sending the message, the value or `ret` will be changed to `true`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2585,14 +2616,22 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
...
+
         LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
                    << "  bssid " << std::get<1>(bssid_tuple);
     }
 
-    if (!send_slave_ap_metric_query_message(mid, bssid)) {
-        LOG(ERROR) << "Failed to forward AP_METRICS_RESPONSE to the son_slave_thread";
-        return false;
+    LOG(DEBUG) << "Forwarding AP_METRICS_QUERY_MESSAGE to all slaves, mid=" << std::hex << int(mid);
+    uint16_t length = message_com::get_uds_header(cmdu_rx)->length;
```
vitalybu: `message_com::get_uds_header` can return `nullptr`

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,54 @@
...
+    // bssid related operations
+public:
+    template <class FirstIt, class LastIt> void add_expected_bssid(FirstIt first, LastIt last);
+    void add_expected_bssid(const sMacAddr &bssid);
+    void remove_expected_bssid(const sMacAddr &);
+    bool find_expected_bssid(const sMacAddr &) const;
+    bool is_expected_bssid_empty() const;
+
+private:
+    uint16_t _mid = 0;
```
vitalybu: [Code style](https://github.com/prplfoundation/prplMesh/blob/master/CODINGSTYLE.md#classes)
```suggestion
    uint16_t m_mid = 0;
```

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,54 @@
...
+    std::unordered_set<sMacAddr> _expected_bssid_list;
```
vitalybu:
```suggestion
    std::unordered_set<sMacAddr> m_expected_bssid_list;
```

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,54 @@
...
+     * that is set as was given bt the parameter (mid default value is zero (0))
+     * @param mid - the new mid value
+     * @return radio mac of the found slave if found, otherwise empty string
+     */
+    void reset_to_new_mid(uint16_t mid = 0);
+    uint16_t get_mid() const;
+    ieee1905_1::CmduMessageTx &create_tx_message();
+
+    // bssid related operations
+public:
```
vitalybu: Why another `public` section?

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,54 @@
...
+    void add_expected_bssid(const sMacAddr &bssid);
+    void remove_expected_bssid(const sMacAddr &);
+    bool find_expected_bssid(const sMacAddr &) const;
+    bool is_expected_bssid_empty() const;
+
+private:
+    uint16_t _mid = 0;
+    std::unordered_set<sMacAddr> _expected_bssid_list;
+
+private:
```
vitalybu, line 82: Why another `private` section?

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,54 @@
...
+    uint8_t _response_buffer[beerocks::message::MESSAGE_BUFFER_LENGTH] = {0};
```
vitalybu:
```suggestion
    uint8_t m_response_buffer[beerocks::message::MESSAGE_BUFFER_LENGTH] = {0};
```

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,54 @@
...
+    ieee1905_1::CmduMessageTx _response = {_response_buffer, sizeof(_response_buffer)};
```
vitalybu:
```suggestion
    ieee1905_1::CmduMessageTx m_response = {m_response_buffer, sizeof(m_response_buffer)};
```

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,54 @@
...
+
+#ifndef _EXPECTED_AP_METRICS_RESPONSE_H
+#define _EXPECTED_AP_METRICS_RESPONSE_H
+
+#include <bcl/beerocks_defines.h>
+#include <tlvf/CmduMessageTx.h>
+#include <tlvf/common/sMacAddr.h>
+#include <tlvf/tlvftypes.h>
+#include <unordered_set>
+
```
vitalybu, line 17: Namespace?

vitalybu (dismissed):

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: I am not sure I fully understand all considerations. However, there is a discussion in #1381 that might shed some light on the issue.
What I can tell is that we use `mid=0` as the mid for the repeated, internally initiated, metrics requests (as oppose to specifically requested by the controller and as oppose to those that are initiated by the monitor because of changes in the stations).
The monitor generated vs. controller generated vs repeatedly generated metrics queries is not hermetically closed - there was a long oral discussion of how to ensure that the responses that come from the different initiators are handled correctly - no bullet proof solution was found. However, current situation seems to be good enough for practical usage.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: > Hashing the entire struct just for the purpose of comparison is highly inefficient

I kindly refuse to discuss any performance issues without measuring.

I changed the implementation to `strcmp()`  as requested.
I think this code should be moved to where sMacAddr is declared (in another PR).
Although "naturally" mac addresses are not sort-able, i.e. there is no greater or less relationships between mac addresses, for any practical usage when grouping mac addresses together, ordering is required and enables algorithm usage.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: fixed

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: this code is about to be refactored soon, fixed anyway.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
RanRegev: true, done.

morantr:
I would really happy if no refactoring will be done on the backhaul manager until I'll finish moving all flows to tasks as part of the unified agent epic. It is making task more difficult.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
RanRegev: Removed.
However -
Like always: **maintainability**.
Having the class header divided into sections makes it _much_ more readable when maintaining the code, when using the code. It is an additional information for the reader about the intentions of the writer, the way she sees the internal structure of the class. 
In this example, the `public` separator also has a comment above it "bssid related operations" which means that it is easier to rational behind the intent of the writer.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
RanRegev: removed. ditto.

RanRegev:
> I would really happy if no refactoring will be done on the backhaul manager until I'll finish moving all flows to tasks as part of the unified agent epic. It is making task more difficult.

@arnout ? @vitalybu ?

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
RanRegev: beerocks

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: > `std::set_intersection` constructs a sorted range consisting of elements that are found in both **sorted** ranges

I might be missing something but as I understand this sentence `std::set_intersection` expects both inputs to be sorted.

vitalybu: You're right. That's what happens when reviewing code at night ;)

vitalybu:
> I would really happy if no refactoring will be done on the backhaul manager until I'll finish moving all flows to tasks as part of the unified agent epic. It is making task more difficult.

@morantr when do you plan to complete you changes?

morantr:
> > I would really happy if no refactoring will be done on the backhaul manager until I'll finish moving all flows to tasks as part of the unified agent epic. It is making task more difficult.
> 
> @morantr when do you plan to complete you changes?

It will not be very quick, it will take some time. It is a little bit hard to estimate it right now, but I already started moving the topology flow to a different task, and I planned to do the same on the link metrics right afterward.

arnout:
> > > I would really happy if no refactoring will be done on the backhaul manager until I'll finish moving all flows to tasks as part of the unified agent epic. It is making task more difficult.
> > 
> > 
> > @morantr when do you plan to complete you changes?
> 
> It will not be very quick, it will take some time. It is a little bit hard to estimate it right now, but I already started moving the topology flow to a different task, and I planned to do the same on the link metrics right afterward.

I understand the feeling, but we can't stop the world while waiting for that to be finished...

But point taken, we'll try to focus on different parts of the code.

morantr:
> > > > I would really happy if no refactoring will be done on the backhaul manager until I'll finish moving all flows to tasks as part of the unified agent epic. It is making task more difficult.
> > > 
> > > 
> > > @morantr when do you plan to complete you changes?
> > 
> > 
> > It will not be very quick, it will take some time. It is a little bit hard to estimate it right now, but I already started moving the topology flow to a different task, and I planned to do the same on the link metrics right afterward.
> 
> I understand the feeling, but we can't stop the world while waiting for that to be finished...
> 
> But point taken, we'll try to focus on different parts of the code.

@arnout, I don't ask to "stop the world", only refactoring tasks.
If it is a bug fix or important feature that is needed for certification or something like that, go ahead, but refactoring tasks are by definition something that could be delayed for the right moment.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
VladyslavTupikin: 0 means the response message will be generated with the new MID

mariomaz: Discussion took place at #1328. Design flaw was temporarily patched at #1388

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
 bool backhaul_manager::send_slave_ap_metric_query_message(uint16_t mid,
                                                           const std::vector<sMacAddr> &bssid_list)
 {
-    bool ret = false;
+    // save mid and clear previous query
+    m_expected_ap_metrics_response.reset_to_new_mid(mid);
+
+    // compare two mac addresses using string compare on their octets
+    auto mac_comp = [](const sMacAddr &addr1, const sMacAddr &addr2) -> bool {
+        return strncmp((char *)addr1.oct, (char *)addr2.oct, 6) < 0;
```
mariomaz: `strncmp` will stop comparing characters if a terminating null-character is reached. 
Function `memcmp` does not stop comparing after finding a null character.

Also use `sizeof(sMacAddr)` instead of magic number 6

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
 bool backhaul_manager::send_slave_ap_metric_query_message(uint16_t mid,
                                                           const std::vector<sMacAddr> &bssid_list)
 {
-    bool ret = false;
+    // save mid and clear previous query
+    m_expected_ap_metrics_response.reset_to_new_mid(mid);
+
+    // compare two mac addresses using string compare on their octets
+    auto mac_comp = [](const sMacAddr &addr1, const sMacAddr &addr2) -> bool {
```
mariomaz, line 1258: Sorting MAC addresses looks like a very handy and recurrent operation. Please consider defining `operator<` for sMacAddr in tlvftypes.h

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+        // alocate memory
+        if (!query->alloc_bssid_list(required_bssids.size())) {
+            LOG(ERROR) << "Failed allocate memory for bssid_list, size: " << bssid_list.size();
+            return false;
+        }
+
+        int i = 0;
+        for (const auto &mac : required_bssids) {
+            auto list         = query->bssid_list(i);
+            std::get<0>(list) = true;
```
mariomaz, line 1314: This assignment is pointless. First element in the tuple is used only when reading, to know if the second element is valid.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
             i++;
         }
+
+        // save the required list
+        m_expected_ap_metrics_response.add_expected_bssid(required_bssids.begin(),
+                                                          required_bssids.end());
+
+        if (!message_com::send_cmdu(socket->slave, cmdu_tx)) {
+            LOG(ERROR) << "Failed forwarding AP_METRICS_QUERY_MESSAGE message to son_slave";
+            ret &= false;
```
mariomaz:
```suggestion
            ret = false;
```

mariomaz: If CMDU fails to be forwarded to one slave, backhaul manager will never get the expected response from that slave.
It's pointless forwarding the CMDU to the rest of the slaves, therefore loop should break.
See my previous [comment ](https://github.com/prplfoundation/prplMesh/pull/1412#discussion_r439243938)to this very same issue.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+        } else {
+            ret &= true;
+        }
```
mariomaz:
```suggestion
        }
```

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1268,49 +1261,83 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
     }
+
     return ret;
```
mariomaz, line 1329: If this method returns false, CMDU is going to be forwarded to all the slaves as per `handle_cmdu()`.
I think this should be avoided by returning true always.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2585,95 +2616,118 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
...
         }
-        bssid.push_back(std::get<1>(bssid_tuple));
+
+        m_expected_ap_metrics_response.add_expected_bssid(std::get<1>(bssid_tuple));
+
         LOG(DEBUG) << "Received AP_METRICS_QUERY_MESSAGE, mid=" << std::hex << int(mid)
                    << "  bssid " << std::get<1>(bssid_tuple);
     }
 
-    if (!send_slave_ap_metric_query_message(mid, bssid)) {
```
mariomaz, line 2575: Method `send_slave_ap_metric_query_message()` was created to forward AP Metrics Query message to a subset of the slaves only, filtering out those that will ignore it. 

Here message is being forwarded to **all** slaves, equally.

mariomaz (changes requested):
Not finished yet.
cursor = handle_slave_ap_metrics_response()

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2585,95 +2616,118 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
...
-     * AP Metrics Query message (which we know because message id is not set), then forward message
-     * to controller.
-     * This might happen when channel utilization value has crossed configured threshold or when
-     * periodic metrics reporting interval has elapsed.
-     */
-    if (0 == mid) {
-        uint16_t length = message_com::get_uds_header(cmdu_rx)->length;
-        cmdu_rx.swap(); //swap back before forwarding
-        return send_cmdu_to_bus(cmdu_rx, controller_bridge_mac, bridge_info.mac, length);
+    if (m_expected_ap_metrics_response.get_mid() != mid) {
```
mariomaz: Consider the following scenario:
- Controller sends AP Metrics Query message
- backhaul manager forwards message to slaves
- All slaves process message and send response to backhaul manager
- backhaul manager gathers all responses and sends full AP Metrics Response message to controller
- One monitor/slave sends AP Metrics Response message with mid = 0 because channel utilization crossed threshold.

Since this equal-to-0 mid does not match the mid of previously processed AP Metrics Query message, this last response will wrongly be discarded.

Also see @arnout's [comment](https://github.com/prplfoundation/prplMesh/pull/1412#discussion_r440750953)

mariomaz (changes requested):
Even though tests are passing, I think there is at least one scenario not properly considered.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: done

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: @arnout ?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: done.

RanRegev: I don't know. Technically you are right. However, we might change the response implementation in the future, to send back "whatever was returned". In this case stopping on first fail would be wrong.
Moreover, this loop creates an asynchronous process of requests. Stopping on failure may result in different runs just because the order of the the sockets is different.
All in all, I am not sure that stopping on error is correct here.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: it looked logical to me to have `&= false` and `&=true` to indicate that there is some loop here and that we are "collecting" the status one by one.
But setting the return to `true` and changing it to `false` in case of an error states the same thing.
thanks.
:-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: `handle_ap_metrics_query` is going to be unite with `send_slave_ap_metric_query_message`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Be aware also that you are mixing up logical AND operator and bitwise AND operator

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: Sorry, but I don't follow you. Can you please elaborate on this?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
RanRegev: We had a long discussion regarding how to handle AP Metrics Responses correctly for all cases: specific request, repeated requests and spontaneous AP metrics response - taking into account that mid=0 is legitimate for all cases.
There was no `correct and simple` solution.
Options were either simple or correct, but not both.
Anyhow I moved the `forward unconditionally on mid=0` to be the first check.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: So my point of view is that it is a bad idea to define a comparison operator on things that are not ordered. mac1 < mac2 has no meaning at all. It would be the same as defining operator+ on MAC addresses. The only thing that defining such an operator does is to motivate people to use std::map or std::set, which are not appropriate containers in most cases.

The only reason why a comparison operation is defined here, is because the standard library developers are too lazy :-) to implement an intersection algorithm that works on unordered containers (probably because that would be highly inefficient). And of course, for this particular case, there is a good reason to use an ordered container, because it makes the intersection operation efficient.

However, if both Ran and Mario have no qualms about defining an operator that has only implementation semantics and no real-world semantics, I'm not going to stop it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: You shall not unconditionally forward responses from slaves to controller if mid is equal to 0.
If you do so, responses to the query in case of periodic metrics reporting are not gathered together in a single response.
The proof is that tests are now failing.

Please check it, but I think that restoring the code as it was when I wrote my comment and calling `m_expected_ap_metrics_response.reset_to_new_mid(0);` if `m_expected_ap_metrics_response.is_expected_bssid_empty()`, right before sending AP Metrics Response message will do the job. Well, at least if the mid in AP Metrics Query message was not also 0, but this will happen once in a blue moon only.

> There was no correct and simple solution.

My suggestion is yet another another patch to the list of patches used to implement this flow. The problem we are trying to solve is or resembles a lot to a well known problem called *stack ripping*. I've been thinking on how to correctly implement this flow by addressing the root cause of the issue and I think that I've come up with a solution: using one new sequence byte in the UDS header called `rid` (for "request ID") would allow us to correctly match queries and responses. The trick is that, unlike the mid value, the rid would not be allowed to be 0, as this value would be reserved for responses without a query (for the channel utilization use case). 

Maybe it's too late for this PR, but I had to write this just in case it can be used to implement a similar flow in the future.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,97 @@
...
+class ExpectedApMetricsResponse {
+public:
+    /**
+     * @brief resets this object with a new mid
+     * @details resets all members of this object to its empty initial state except for the mid 
+     * that is set as was given bt the parameter (mid default value is zero (0))
+     * @param mid the new mid value
+     * @return radio mac of the found slave if found, otherwise empty string
+     */
```
morantr, line 28: Please fix Doxygen documentation on all this file:
Missing a capital letter at sentences beginning.
Missing dots at the sentences end.
After `@brief` there should be an empty line. 
If a function does not return anything, don't add `@return`.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,97 @@
...
+     * @brief creates and returns a AP-RESPONSE-METRICS Tx message 
+     * @return a fresh new CmduMessage with the correct type and mid
+     */
+    ieee1905_1::CmduMessageTx &create_tx_message();
+
+    // bssid related operations
+
+    /**
+     * @brief stores expected bssids
+     * @details adds the bssids in the range [first,last) to the already existing internal list
```
morantr, line 47: Unequal parenthesis: `[first,last)`
Edit: I see now that on a lot of places on documentation website do it as you did, not sure why. So fill free to ignore this comment.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,97 @@
...
+     * @param first iterator to the source begin
+     * @param last iterator to the source last
+     * @return - nothing
+     */
+    template <class FirstIt, class LastIt> void add_expected_bssid(FirstIt first, LastIt last);
```
morantr, line 52: Why do you use template here? There is only one use if this function, so please replace it to non-template function.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,97 @@
...
+     * @return nothing
+     */
+    void add_expected_bssid(const sMacAddr &bssid);
+
+    /**
+     * @brief removes the given bssid from the internal list
+     * @param bssid the bssid to remove
+     * @return - nothing
+     */
+    void remove_expected_bssid(const sMacAddr &);
```
morantr, line 67: Please add argument name:
`const sMacAddr &bssid`
Also on `find_expected_bssid()`.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,97 @@
...
+     * @brief checks if the internal list is empty
+     * @return bool empty / not empty
+     */
+    bool is_expected_bssid_empty() const;
+
+private:
+    uint16_t m_mid = 0;
+    std::unordered_set<sMacAddr> m_expected_bssid_list;
+    uint8_t m_response_buffer[beerocks::message::MESSAGE_BUFFER_LENGTH] = {0};
+    ieee1905_1::CmduMessageTx m_response = {m_response_buffer, sizeof(m_response_buffer)};
```
morantr, line 86: Why would you define it? It is a waste of memory. If you create the instance on the backhaul manager, you can pass the cmdu_tx of the backhaul manager as a reference and use it.
Please remove it, and declare a reference of cmdu_tx instead, and initialize it on the class constructor.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+void ExpectedApMetricsResponse::remove_expected_bssid(const sMacAddr &bssid)
+{
+    m_expected_bssid_list.erase(bssid);
+}
+
+bool ExpectedApMetricsResponse::is_expected_bssid_empty() const
+{
+    return m_expected_bssid_list.empty();
+}
```
morantr, line 31: Most of all the class functions have one line only. It is totally ok to add one line implementation to the header file instead of the CPP file. We do it a lot. (nitpick).

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
```diff
@@ -0,0 +1,97 @@
...
+
+#include <bcl/beerocks_defines.h>
+#include <tlvf/CmduMessageTx.h>
+#include <tlvf/common/sMacAddr.h>
+#include <tlvf/tlvftypes.h>
+#include <unordered_set>
+
+namespace beerocks {
+
+class ExpectedApMetricsResponse {
```
morantr, line 20: After fully reviewing this class. I gotta say I don't understand why it is even needed.
You always talk about how much it is bad to add complexity level redundantly.
This class is just a wrapper to unodered_set, and uint16_t (mid). And all of the class functions are just wrappers to the unordered_set function. **No logic is implemented here**.
There is no good reason why this class should exist. It can be replaced by two members on the backhaul manager thread (the unordered_set, and the mid), and use them directly.
Please remove this class and add the required data members to the backhaul manager.

agent/src/beerocks/slave/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ file(GLOB beerocks_agent_sources
     ${MODULE_PATH}/link_metrics/*.c*
```
morantr, line 11: Regarding this commit:
1. Commit message "agent: backhaul_manager: Refactor handler for AP_METRICS_RESPONSE_MES…
…SAGE" is too long.
2. There Jira reference is missing.
3. I would totally split it into two commits. One for the new class, and a second one for using it.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1023,21 +1021,16 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
                 for (const auto &socket : slaves_sockets) {
                     if (socket) {
                         for (int i = 0; i < beerocks::IFACE_TOTAL_VAPS; ++i) {
-                            if (socket->vaps_list.vaps[i].mac != network_utils::ZERO_MAC) {
+                            if ((socket->vaps_list.vaps[i].mac != network_utils::ZERO_MAC) &&
```
morantr, line 1024: I know you didn't do it, but it will be greater if you refactor it to:
```cpp
if (!socket) {
   continue;
}
// rest of the code
```

(nitpick)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1023,21 +1021,16 @@ bool backhaul_manager::backhaul_fsm_main(bool &skip_select)
...
-                // do not correspond to an AP Metrics Query message.
-                // We cannot set MID to 0 here because we must also differentiate periodic
-                // AP Metrics Response messages and messages received from monitor thread
-                // due to channel utilization crossed configured threshold value.
-                // As a temporary solution, set MID to UINT16_MAX here.
-                // TODO: to be fixed as part of #1328
+
                 if (!bssid_list.empty()) {
-                    send_slave_ap_metric_query_message(UINT16_MAX, bssid_list);
+                    send_slave_ap_metric_query_message(0, bssid_list);
```
morantr, line 1033: It is probably best if the mid will be optional value with `0` as default.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: If it is going to be defined somewhere, it needs to be defined on "tlvtypes.h" on the framework.
And on a separate preparative commit of course.
Also, please don't use memcmp(), replace it with `addr1 == addr2`.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1258,49 +1251,81 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+    m_expected_ap_metrics_response.reset_to_new_mid(mid);
+
+    // compare two mac addresses using string compare on their octets
+    auto mac_comp = [](const sMacAddr &addr1, const sMacAddr &addr2) -> bool {
+        return memcmp((char *)addr1.oct, (char *)addr2.oct, sizeof(sMacAddr)) < 0;
+    };
+
+    // sort the given bssid list
+    std::vector<sMacAddr> given_bssid_list = bssid_list;
+    std::sort(given_bssid_list.begin(), given_bssid_list.end(), mac_comp);
```
morantr, line 1264: If you need it sorted, change the type `bssid_list` to `std::set` on `backhaul_fsm_main()`.
It is promised to be sorted, and benefit the fact that no reallocation will occur on every `push_back`.
And also you don't need to copy the input argument to a local argument.
(although it was totally ok to define it as a non-const reference and you wouldn't have to do it anyway).

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1258,49 +1251,81 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
+        std::vector<sMacAddr> current_socket_bssids;
+        std::transform(socket->vaps_list.vaps, socket->vaps_list.vaps + beerocks::IFACE_TOTAL_VAPS,
+                       std::back_inserter(current_socket_bssids),
+                       [](const beerocks_message::sVapInfo &vap_info) { return vap_info.mac; });
 
-                auto list         = query->bssid_list(0);
-                std::get<0>(list) = true;
-                std::get<1>(list) = mac;
+        // sort current vaps bssid
+        std::sort(current_socket_bssids.begin(), current_socket_bssids.end(), mac_comp);
```
morantr, line 1284: Also here, make it `std::set`.
And I don't think that `std::transform` is a great choice. `std::fill` could fit more.
One more thing, when you iterate over the vaps, you need to check that vap is valid. Some elements on the array are empty.

```cpp
std::set<sMacAddr> current_socket_bssids;
for (cont auto & bssid : socket->vaps_list.vaps) {
    if (bssid.mac == network_utils::ZERO_MAC) {
        continue;
    }
    current_socket_bssids.insert(bssid.mac);
}

File: agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp [hunk 38]
```diff
@@ -1258,49 +1251,81 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
...
-                    ret = false;
-                    continue;
-                } else {
-                    ret = true;
-                    // Fill a query vector
-                    m_ap_metric_query.push_back({socket->slave, mac});
-                }
-            }
+        // fill a list of the common bssids: given list with current vaps
+        std::vector<sMacAddr> required_bssids;
```
morantr, line 1287: `std::vector` is good for use if you need push elements only once or few times.
Since every call of this function you will push new things, it is not the best container to choose from.
Please change it to `std::deque` is it just like a vector but without nasty reallocation.

morantr (changes requested):
Please rebase to #1504, and use available data from the AgentDB.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
arnout: It's mathematical notation for a range that includes the beginning (square bracket) but doesn't include the end (round bracket).

Your comment, however, indicates that this is not knowledge that we can assume everyone has, so indeed it's better to replace it with
```suggestion
     *
     * Adds the bssids in the range @a first (inclusive) to @a last (exclusive) to the already existing internal list
```

Additional notes:
- capitalization, as @morantr already indicated;
- use `@a` to indicate an argument;
- don't use `@details` to separate from the brief, instead just leave an empty line.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
arnout: Template is good, because different containers have a different iterator type.

However, first and last must be the same type, so
```suggestion
    template <class Iterator> void add_expected_bssid(Iterator first, Iterator last);
```

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
arnout: The response message is built up for every partial response that we get from the slaves. If we would use cmdu_tx from backhaul manager for it, it would be cleared any time a different CMDU arrives on the backhaul manager.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: Begging to differ here...

std::vector is most efficient for small containers, especially if you only append at the end, not in the middle. Reallocation doesn't actually happen that often, because it will over-allocate a bit. A std::list, for instance, will usually have a lot more allocations. A deque will have slightly less allocations, but use a lot more space. Also, even though both vector and deque have O(1) access, the deque access takes twice as much time (it needs two pointer dereferences instead of just one).

So, the only good reasons to use a deque are:
- you did profiling and it turns out to be faster for realistic use cases;
- the semantics we need are more like a deque than a general container.

I think neither is true here.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
morantr: That is not true. Arriving messages are getting into cmdu_rx which has different buffer than cmdu_tx.
Anyway, it doesn't matter, the whole class is redundant and can be inlined on the backhaul manager code which in that case would use the cmdu_tx of the backhaul.

arnout: It's used here: https://github.com/prplfoundation/prplMesh/pull/1467/files#diff-80c48bbadac679bad34f8ee37e30f772R2661

There are several responses coming from several slaves. Each of them goes into handle_slave_ap_metrics_response, which adds a TLV to the existing CMDU. Only if all of them have been received (line 2743) the CMDU is actually sent.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: I kind of getting the feeling that `std::vector` is kind of the default container everyone uses without giving much thought if it the right container for the job.
It is good for use if you push or insert elements once or very few times. This is not the case here.

Here is a link to a flow chart that explains how to choose a container:
https://stackoverflow.com/a/471461/6039092
I think we should adopt it, and add it to our coding style document.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2567,95 +2596,118 @@ bool backhaul_manager::handle_ap_metrics_query(ieee1905_1::CmduMessageRx &cmdu_r
...
         return false;
     }
 
-    sApMetrics metric;
-    // Copy data to the response vector
-    metric.bssid               = ap_metrics_tlv->bssid();
-    metric.channel_utilization = ap_metrics_tlv->channel_utilization();
-    metric.number_of_stas_currently_associated =
+    // prepare ap metrics tlv
+    auto &ap_metrics_tx_message  = m_expected_ap_metrics_response.create_tx_message();
```
arnout, line 2660: This is wrong: create_tx_message will delete all previously received metrics. The create itself should be done when the expected ap metrics response is initialized.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
morantr: But here it is redundant because we don't have different containers. We have a single container.

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
morantr: I think that your new comment here https://github.com/prplfoundation/prplMesh/pull/1467#discussion_r453668934 got me to understand why I thought this whole class was useless.
What you wrote here is not what was implemented on the PR. I saw that each time we create a new message.

So this whole thing is a tradeoff between adding all the information to some container, and when all the messages from son_slave are done copy the information to tlvs, to adding a tlv from each newly received data on side cmdu_tx, and send it when it finishes???

I'm not sure which option I dislike the most. 😑
@arnout does SharedDB could help somehow to avoid these two options?

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
RanRegev: [no definition at the moment](https://jira.prplfoundation.org/browse/PPM-44)

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
RanRegev: [ok, but let's make it explicit and not on comments basis](https://jira.prplfoundation.org/browse/PPM-44)

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
morantr: I'm not sure we need to add something that is already on Doxygen documentation to our documentation.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
arnout: I agree that we should adopt something.

However, I'm kind of a fan of the flow chart from [CppGuideline SL.con.2](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#slcon2-prefer-using-stl-vector-by-default-unless-you-have-a-reason-to-use-a-different-container): Use std::array if size is constant, std::vector if not. The implicit corollary is: only use one of the other containers if profiling proves it's more efficient. I like that as a programmer, I don't have to think about it but I can just use std::vector. Simpler is better IMO. But I'm not against using the flow chart either.

Note that for this particular case, the flow chart will point to std::vector (order is not important, no key lookup, no persistent positions, size doesn't vary wildly) (although I'm not entirely sure what "size varies wildly" means).

agent/src/beerocks/slave/backhaul_manager/internal/expected_ap_metrics_response.h
arnout: Since sharedDB is not actually shared with the monitor, I don't see how it would help.

The original code indeed had adding the information to a container instead of adding it to a CMDU. I thought the latter would be easier to understand - but apparently not :-)

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: "Size doesn't vary wildly" means that you push or insert elements once or a few times.
In this case, the vector will be filled each time handling the message, so its size varies wildly. 
Also, a case when you push elements one by one is a case of "size varies wildly" (Unless you know how much you are going to push and reserve the size first).

I don't agree with what is said on "CppGuideline SL.con.2". It is too general and could miss bigtime.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
mariomaz: We need memcmp() because we are implementing operator<

morantr: The `memcmp` is in the internal lambda function. I could compare it however you want. Not see how it is related to operator<.

mariomaz: Please note that the result of `memcmp()` is compared against 0 with operator<, not with operator==

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -1258,49 +1251,81 @@ bool backhaul_manager::send_autoconfig_search_message(std::shared_ptr<sRadioInfo
 bool backhaul_manager::send_slave_ap_metric_query_message(uint16_t mid,
                                                           const std::vector<sMacAddr> &bssid_list)
 {
-    bool ret = false;
+    // save mid and clear previous query
+    m_expected_ap_metrics_response.reset_to_new_mid(mid);
+
+    // compare two mac addresses using string compare on their octets
+    auto mac_comp = [](const sMacAddr &addr1, const sMacAddr &addr2) -> bool {
+        return memcmp((char *)addr1.oct, (char *)addr2.oct, sizeof(sMacAddr)) < 0;
```
mariomaz, line 1259: BTW, casting to (char*) is not required.

PR #1469, opened 2020-06-22
[agent]  Add validate_channel method

VladyslavTupikin:
For correct implementation Backhaul STA
Steering need to add new method in the
wireless_utils for the channel validation.

### DIFF
--- a/common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
+++ b/common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
@@ -170,6 +170,7 @@ public:
     static uint8_t get_operating_class_by_channel(const beerocks::message::sWifiChannel &channel);
     static std::list<sChannelPreference>
     get_channel_preferences(const beerocks::message::sWifiChannel supported_channels[]);
+    static bool channel_validate(const uint8_t &channel);
 
 private:
     enum eAntennaFactor {
--- a/common/beerocks/bcl/source/son/son_wireless_utils.cpp
+++ b/common/beerocks/bcl/source/son/son_wireless_utils.cpp
@@ -869,3 +869,15 @@ std::list<uint8_t> wireless_utils::string_to_wsc_oper_class(const std::string &o
         return {};
     }
 }
+
+bool wireless_utils::channel_validate(const uint8_t &channel)
+{
+    for (auto oper_class : operating_classes_list) {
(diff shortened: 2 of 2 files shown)


common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
```diff
@@ -170,6 +170,7 @@ class wireless_utils {
     static uint8_t get_operating_class_by_channel(const beerocks::message::sWifiChannel &channel);
     static std::list<sChannelPreference>
     get_channel_preferences(const beerocks::message::sWifiChannel supported_channels[]);
+    static bool channel_validate(const uint8_t &channel);
```
vitalybu, line 173: Please add a doxygen comment for this method

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -869,3 +869,15 @@ std::list<uint8_t> wireless_utils::string_to_wsc_oper_class(const std::string &o
...
+bool wireless_utils::channel_validate(const uint8_t &channel)
+{
+    for (auto oper_class : operating_classes_list) {
+        auto ch = oper_class.second.channels.find(channel);
+
+        if (ch != oper_class.second.channels.end()) {
+            return true;
+        }
+    }
+    return false;
```
vitalybu, line 882: That's not a very efficient implementation for this operation.
If we don't care about the operating classes, maybe we need to have an `std::set` with all the "valid" channels and search it instead of iterating over all the operating classes.

Also should use a const reference where possible:
```suggestion
    for (const auto &op_class : operating_classes_list) {
        if (op_class.second.channels.find(channel) != op_class.second.channels.end()) {
            return true;
        }
    }
    return false;
```

arnout: > That's not a very efficient implementation for this operation.
> If we don't care about the operating classes, maybe we need to have an std::set with all the "valid" channels and search it instead of iterating over all the operating classes.

IMO, we shouldn't bother with efficiency too much.

There are a dozen different ways that the operating class table is accessed. If we would keep a copy of the table for each access pattern, it would be very hard to maintain all those copies. For how many microseconds gain? We're talking about a tiny table with just about a 100 entries here...

PR #1471, opened 2020-06-23
[agent] Add validate_channel method

VladyslavTupikin:
For correct implementation Backhaul STA
Steering need to add new method in the
wireless_utils for the channel validation.

Add implementation for the method.

### DIFF
--- a/common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
+++ b/common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
@@ -171,6 +171,15 @@ public:
     static std::list<sChannelPreference>
     get_channel_preferences(const beerocks::message::sWifiChannel supported_channels[]);
 
+    /**
+    * @brief Match channel number in the given operating class.
+    *
+    * @param operating_class operating class
+    * @param channel channel number
+    * @return True if channel matches to operating class
+    */
+    static bool is_channel_in_operating_class(uint8_t operating_class, uint8_t channel);
+
 private:
     enum eAntennaFactor {
         ANT_FACTOR_1X1 = 0,
--- a/common/beerocks/bcl/source/son/son_wireless_utils.cpp
+++ b/common/beerocks/bcl/source/son/son_wireless_utils.cpp
@@ -869,3 +869,10 @@ std::list<uint8_t> wireless_utils::string_to_wsc_oper_class(const std::string &o
         return {};
     }
 }
+
+bool wireless_utils::is_channel_in_operating_class(uint8_t operating_class, uint8_t channel)
+{
+    auto channel_set = operating_class_to_channel_set(operating_class);
+
+    return (channel_set.find(channel) != channel_set.end());
+}


common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -869,3 +869,20 @@ std::list<uint8_t> wireless_utils::string_to_wsc_oper_class(const std::string &o
...
+ * @return True if channel matches to operating class
+ */
+bool wireless_utils::channel_validate(uint8_t operating_class, uint8_t channel)
+{
+    auto channel_set = operating_class_to_channel_set(operating_class);
+
+    if (channel_set.find(channel) != channel_set.end()) {
+        return true;
+    }
+    return false;
```
rmelotte:
```suggestion
    return (channel_set.find(channel) != channel_set.end())
```

rmelotte (approved):
Shouldn't we also check if the channel is operable (e.g. if vht capability is required)?

Or is left to the controller to only send steering requests for operable channels?
I couldn't find this information easily in the specification.

common/beerocks/bcl/source/son/son_wireless_utils.cpp
VladyslavTupikin: accepted!

common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
```diff
@@ -170,6 +170,7 @@ class wireless_utils {
     static uint8_t get_operating_class_by_channel(const beerocks::message::sWifiChannel &channel);
     static std::list<sChannelPreference>
     get_channel_preferences(const beerocks::message::sWifiChannel supported_channels[]);
+    static bool channel_validate(uint8_t operating_class, uint8_t channel);
```
adam1985d: I think the naming should better indicate what the method is meant to do:
1. `channel_validate` can mean so many things - a `validate_channel_in_operating_class` would better explain itself
2. nitpick: it returns bool - so I prefer to have prefix `is_` to clarify we don't do an actual operation here - just perform a check. if you choose to accept this nitpick the naming would result in `is_channel_in_operating_class` - which I think is the best option in terms of self-explainable

common/beerocks/bcl/source/son/son_wireless_utils.cpp
```diff
@@ -869,3 +869,17 @@ std::list<uint8_t> wireless_utils::string_to_wsc_oper_class(const std::string &o
         return {};
     }
 }
+
+/**
+ * @brief convert operating class to bandwidth based on Table 4-E in the ieee 802.11 specification
```
adam1985d: probably a copy-paste typo? the method brief description is incorrect

adam1985d (changes requested):

common/beerocks/bcl/include/bcl/son/son_wireless_utils.h
VladyslavTupikin: accepted!

common/beerocks/bcl/source/son/son_wireless_utils.cpp
VladyslavTupikin: This is not a related issue.

VladyslavTupikin: Accepted! Moved Doxygen to the header file, according to the poll results in democracy conversation.

adam1985d (approved):

PR #1476, opened 2020-06-24
Feature/track easymesh cert

### LINKED ISSUE #1184: [TASK] run_easymesh_cert.sh - add option to checkout to easymesh_cert repo sha1
## Description

Currently, the `run_easymesh_cert.sh` relies on the easymesh_cert repository which contains the `run_test_file.sh` script which is used to run certification. The `run_easymesh_cert.sh` is simply a wrapper script which in addition to running certification tests using the `run_test_file.sh` script, first downloads the latest prplmesh.ipk from owncloud, and after the tests complete uploads the results to owncloud.

Since we are using two separate repositories, this can easily break, for example, if the easymesh_cert repo run_test_file.sh script is modified and we are testing an older prplmesh.

For this reason, the `run_easymesh_cert.sh` script should have an option to checkout the easymesh_cert repo to a specific sha1, and use a default which is defined in the script which we will change when needed.

## Exit criteria

Add 2 new options to `run_easymesh_cert.sh` script (and change the current one used only for setting the easymesh-cert path):
```
--easymesh-cert-path <path> -  path to easymesh_cert repo (default ../easymesh_cert)
--easymesh-cert-fixed - when set, cleans and checkouts to a specific easymesh_cert SHA1 (default "${EASYMESH_CERT_REVISION}")
--easymesh-cert-revision <revision> - used in conjunction with --easymesh-cert-fixed to override the default revision
```

rmelotte:
Tracking the easymesh_cert repository in the prplMesh repository will
allow us to test new changes to the UCC test files and the sigma agents, but more
importantly it will make the CI pipelines easier to repeat at two
different points in time.

- add a script to reset a git repository to a specific version
- add the current commit hash of the easymesh_cert repo
- use the script both in Gitlab CI, and in run_easymesh_cert.sh

Fixes https://github.com/prplfoundation/prplMesh/issues/1184


ci/git-clean-reset.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh -e
+
+# Reset the repository pointed by $1 to the revision given by $2
+
+cd "$1" || {
```
RanRegev, line 5: Please also verify the existence and validity of $2.
I think `git reflog` may have some validation abilities for $2.

ci/easymesh_cert_version
```diff
@@ -0,0 +1 @@
+5d6bdb9168c6c2dd0b4c10fe25303ce5e7f28cd7
```
RanRegev: Is it possible to a comment line with a link to this sha in github?
And / or a comment that explains _why_ this sha was chosen.

.gitlab-ci.yml
```diff
@@ -300,6 +300,7 @@ run-certification-tests:
           echo "Deploying to $DEVICE_UNDER_TEST"
           tools/deploy_ipk.sh --certification-mode $DEVICE_UNDER_TEST build/$DEVICE_UNDER_TEST/prplmesh.ipk
       fi
+    - ci/git-clean-reset.sh /easymesh_cert "$(<"ci/easymesh_cert_version")"
```
RanRegev, line 303: regarding my previous comment on adding a comment to `easymesh_cert_version` - it will require some "hulahoop" here, something along those lines (please verify) -
```suggestion
    - ci/git-clean-reset.sh /easymesh_cert "$(<"grep -m1 ^[^#] ci/easymesh_cert_version")"
```

ci/certification/generic.yml
```diff
@@ -4,6 +4,7 @@
     GIT_CLONE_PATH: "/builds/prpl-foundation/prplMesh/"
   script:
       - echo $CI_COMMIT_DESCRIPTION
+      - ci/git-clean-reset.sh /easymesh_cert "$(<"ci/easymesh_cert_version")"
```
RanRegev, line 7: ditto

ci/run_easymesh_cert.sh
```diff
@@ -24,6 +24,7 @@ usage() {
     echo "      -v|--verbose - set verbosity (ucc logs also redirected to stdout)"
```
RanRegev, line 24: nitpick - sort the options alphabetically.

RanRegev (approved):
minor comments to enhance readability.

ci/git-clean-reset.sh
tomereli: I think adding defaults is a good idea - we can assume that easymesh_cert is in the same level as prplmesh, and the default revision can just be the master branch.
```suggestion
path=${1-../../easymesh_cert}
revision=${2-master}
cd "$1" || {
```

ci/easymesh_cert_version
tomereli: > Is it possible to a comment line with a link to this sha in github?
> And / or a comment that explains _why_ this sha was chosen.

This SHA1 will change all the time. It points to the easymesh_cert revision which prplMesh was tested with. Every time we will update the easymesh cert, we will update this SHA1. Comment is completely redundant here, not everything deserves a comment - in this case the file name is enough IMO.

.gitlab-ci.yml
tomereli: Which is exactly why adding a redundant comment is not a good idea.

ci/run_easymesh_cert.sh
```diff
@@ -24,6 +24,7 @@ usage() {
...
     echo "      -o|--log-folder - path to put the logs to (make sure it does not contain previous logs)."
     echo "      -e|--easymesh-cert - path to easymesh_cert repository (default ../easymesh_cert)"
+    echo "      --reset-easymesh-cert - reset the easymesh_cert repository to the version specified in this file."
```
tomereli, line 27: I think it will be useful to allow the user to select the revision using another option - `--easymesh-cert-revision`, and if it is not selected use the default in this file.

tomereli (approved):

ci/easymesh_cert_version
RanRegev: I wonder where I'll have a chance to follow at least some of the decisions that led to these rapid changes.
I thought that a comment in this specific file might help, but maybe with the commit message that changes it would be good enough too. 
The point is that I am looking for _easy access_: where this sha is taken from (https://???), why was it chosen (we needed feature XYZ to solve issue #ccc, etc.) 
And again, I care less about the specific position these comments are written, and more about the fact that they _will_ be written. 
E.g In this very commit or PR I don't see anything that explains the why, or that links to something that might answer the selection of this specific sha.

ci/git-clean-reset.sh
rmelotte: @RanRegev  
If $2 doesn't exist, git will report the reference doesn't exist, and the script will exit, so I don't think any additional checks are necessary.

@tomereli 
This script is about resetting any git repository, it has nothing specific to the easymesh_cert repo. The default value you mention is already in `run_easymesh-cert.sh`

ci/easymesh_cert_version
rmelotte: @Ran you're right when we update the version we should ideally mention what has changed in a nutshell (why we need to update it).
I added both the link to the target repo and the reason why I chose this hash in the commit message.

.gitlab-ci.yml
rmelotte: Having only the commit hash in this file makes it easy to use, so I prefer to keep it like that.
I agree we should add any comments that could be useful (like why the commit hash has to be updated) to the commit message.

ci/run_easymesh_cert.sh
rmelotte: I have the impression that when you want to try things out locally, you'd prefer to use uncommitted changes, so not using the option at all. 
We can still add the option later if it turns out someone need it.

rmelotte:
It has to be rebased after https://github.com/prplfoundation/prplMesh/pull/1478 is merged.

PR #1483, opened 2020-06-24
Feature/issue 1381 boardfarm in gitlab ci

### LINKED ISSUE #1381: [TASK] Integrate boardfarm in gitlab-ci

odkq:
https://jira.prplfoundation.org/browse/PPM-30


dctest.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+function dc {
```
arnout: `dc` is already an existing tool (decimal calculator, I think), so better call it something unique. "compose" for example.

dctest.sh
```diff
@@ -0,0 +1,14 @@
...
+    ROOT_DIR=$(dirname $(pwd)) docker-compose -f tools/docker/boardfarm-ci/docker-compose.yml $@
+}
+if [ $# -eq 0 ]; then
+    dc up boardfarm
+else
+    if [ $1 == "-h" ]; then
```
arnout: I believe shellcheck will complain here...
```suggestion
    if [ "$1" == "-h" ]; then
```

tools/docker/boardfarm-ci/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
+version: '3'
+services:
+    controller:
+        image: registry.gitlab.com/prpl-foundation/prplmesh/prplmesh-runner
```
arnout: We'll need an explicit tag here as well. Is there a way for docker-compose to get variables that get expanded? Like ARG in Dockerfile?

tools/docker/boardfarm-ci/docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+        environment:
+            - USER=$USER
+            - INSTALL_DIR=$ROOT_DIR/build/install
+            - ROOT_DIR=$ROOT_DIR
+        expose:
+            - "5000"
+            - "8002"
+        volumes:
+            - "$ROOT_DIR:$ROOT_DIR"
+        entrypoint: ["/root/entrypoint.sh", "/usr/bin/start-controller-agent"]
```
arnout: OMG, this is a lot more readable than the stupid run script we have now... Nice!

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_compose.py
```diff
@@ -0,0 +1,100 @@
...
+
+import boardfarm
+from environment import ALEntityDocker, _get_bridge_interface
+from .prplmesh_base import PrplMeshBase
+from sniffer import Sniffer
+
+rootdir = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../../..'))
+
+
+class PrplMeshCompose(PrplMeshBase):
```
arnout, line 17: I don't think we should define a new class. Instead, change the PrplMeshDocker class to use compose instead of naked docker.

arnout (commented):
Not a full review - just a few loose comments. Please also see my comments in https://github.com/prplfoundation/prplMesh/pull/1492#pullrequestreview-439317396 where they're relevant for this one.

I don't really like this docker compose approach to launch several containers. Indeed, it forces us to duplicate the "knowledge" of which containers to use between the compose.yml file and the boardfarm config file. However, I do look how the compose yml file describes how to launch the container much better than how the shell script does it. So I'm all for using that rather than shell scripts to launch our containers. Wouldn't it be possible to make two compose.yml files, one for boardfarm itself and one for prplmesh, and do a compose boardfarm which launches boardfarm which launches two times compose prplmesh which luanches agent resp. controller?

rmelotte:
@odkq can you please:
- fix the shellcheck and flake8 issues that are reported on Gitlab: https://gitlab.com/prpl-foundation/prplMesh/-/pipelines/173659597
- make sure your "Signed-off-by" is there for every commit

Then we will have to cleanup the history (at least the easy parts), merge it, then do additional cleanup like removing the old boardfarm test, and integrating https://github.com/prplfoundation/prplMesh/pull/1556.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_compose.py
```diff
@@ -0,0 +1,112 @@
...
+        self.conn_cmd = config.get("conn_cmd", None)
+        self.delay = config.get("delay", 7)
+        self.docker_network = "boardfarm-ci_default"
+
+        if self.role == "controller":
+            self._docker_compose(["-d", "--name", self.name, "controller"],
+                                 "run", "start-controller")
+            time.sleep(self.delay)
+            self.controller_entity = \
+                ALEntityDocker(self.name, is_controller=True, compose=True)
```
rmelotte, line 51:
```suggestion
                ALEntityDocker(self.name, device=self, is_controller=True, compose=True)
```

rmelotte:
Superseded by https://github.com/prplfoundation/prplMesh/pull/1587

PR #1485, opened 2020-06-25
Feature/add agent task infrastructure

morantr:
Add tasks infrastructure to the Agent. as part of the unified Agent #1082 - Phase 3.

### DIFF
--- a/agent/src/beerocks/slave/CMakeLists.txt
+++ b/agent/src/beerocks/slave/CMakeLists.txt
@@ -10,6 +10,7 @@ file(GLOB beerocks_agent_sources
     # [TASK] Move link metric related classes to BPL #910
     ${MODULE_PATH}/link_metrics/*.c*
     ${MODULE_PATH}/platform_manager/*.c*
+    ${MODULE_PATH}/tasks/*.c*
     ${MODULE_PATH}/gate/*.c*
     ${MODULE_PATH}/*.c*
 )
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
@@ -536,6 +536,9 @@ void backhaul_manager::after_select(bool timeout)
         }
         send_cmdu_to_bus(cmdu_tx, network_utils::MULTICAST_1905_MAC_ADDR, bridge_info.mac);
     }
+
+    // Run Tasks
+    m_task_pool.run_tasks();
 }
 
 bool backhaul_manager::finalize_slaves_connect_state(bool fConnected,
@@ -2170,9 +2173,13 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<sRadioInfo>
         break;
     }
     default: {
-        LOG(ERROR) << "Unhandled message received from master: "
-                   << int(beerocks_header->action_op());
-        return false;
+        bool handled = m_task_pool.handle_cmdu(cmdu_rx, sMacAddr(), beerocks_header);
+        if (!handled) {
+            LOG(ERROR) << "Unhandled message received from the Controller: "
+                       << int(beerocks_header->action_op());
+            return false;
+        }
+        return true;
     }
     }
 
@@ -2241,10 +2248,11 @@ bool backhaul_manager::handle_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
     default: {
         // TODO add a warning once all vendor specific flows are replaced with EasyMesh
         // flows, since we won't expect a 1905 message not handled in this function
-        return false;
+        return m_task_pool.handle_cmdu(cmdu_rx, tlvf::mac_from_string(src_mac));
     }
     }
 }
+
 bool backhaul_manager::handle_slave_1905_1_message(ieee1905_1::CmduMessageRx &cmdu_rx,
                                                    const std::string &src_mac)
 {
@@ -2256,8 +2264,12 @@ bool backhaul_manager::handle_slave_1905_1_message(ieee1905_1::CmduMessageRx &cm
         return handle_slave_channel_selection_response(cmdu_rx, src_mac);
     }
     default: {
-        LOG(DEBUG) << "Unexpected 1905 message " << int(cmdu_rx.getMessageType());
-        return false;
+        bool handled = m_task_pool.handle_cmdu(cmdu_rx, tlvf::mac_from_string(src_mac));
+        if (!handled) {
+            LOG(DEBUG) << "Unexpected 1905 message " << int(cmdu_rx.getMessageType());
+            return false;
+        }
+        return true;
     }
     }
 }
--- a/agent/src/beerocks/slave/tasks/task.h
+++ b/agent/src/beerocks/slave/tasks/task.h
@@ -1,0 +1,74 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2016-2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _TASK_H_
+#define _TASK_H_
+
+#include <beerocks/tlvf/beerocks_header.h>
+#include <tlvf/CmduMessageRx.h>
+
+namespace beerocks {
+
+/**
+* @brief Enum which defines all tasks type.
+* 
+* Each task must have an entry on the enum eType, so it will be possible to send the task an
+* event. 
+*/
+enum eTaskType : uint8_t {
+    TOPOLOGY,
+};
+
+class Task {
+public:
+    Task() = delete;
+    explicit Task(eTaskType task_type) : m_task_type(task_type) {}
+    virtual ~Task() {}
+
+    /**
+     * @brief Retrun the task type.
+     * 
+     * Called by TaskPool send_event().
+     * 
+     * @return eType Task type.
+     */
+    virtual eTaskType get_task_type() { return m_task_type; }
+
+    /**
+     * @brief Work function. Shall do the task routine. Called by TaskPool run_task() function.
+     */
+    virtual void work(){};
+
+    /**
+     * @brief Handle events, which are being send from containing thread.
+     * 
+     * @param event_enum_value Event enum value which shall be defined on the task.
+     */
+    virtual void handle_event(uint8_t event_enum_value){};
+
+    /**
+     * @brief Handle CMDU message.
+     * 
+     * @param cmdu_rx CMDU object conataining the message.
+     * @param src_mac AL Mac of the sender.
+     * @param beerocks_header Beerocks Message header (Only on VS message).
+     * @return true if the message has been handled, otherwise false.
+     */
+    virtual bool handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, const sMacAddr &src_mac,
+                             std::shared_ptr<beerocks_header> beerocks_header)
+    {
+        return false;
+    }
+
+private:
+    eTaskType m_task_type;
+};
+
+} // namespace beerocks
+
+#endif // _TASK_H_
--- a/agent/src/beerocks/slave/tasks/task_pool.cpp
+++ b/agent/src/beerocks/slave/tasks/task_pool.cpp
@@ -1,0 +1,53 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2016-2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "task_pool.h"
+
+using namespace beerocks;
+
+bool TaskPool::add_task(const std::shared_ptr<Task> &new_task)
+{
+    if (!new_task) {
+        LOG(ERROR) << "new task pointer is nullptr";
+        return false;
+    }
+    m_task_pool[new_task->get_task_type()] = new_task;
+    return true;
+}
+
+void TaskPool::run_tasks()
+{
+    for (auto &task_element : m_task_pool) {
+        auto &task = task_element.second;
+        task->work();
+    }
+}
+
+void TaskPool::send_event(eTaskType task_type, uint8_t event)
+{
+    auto task_it = m_task_pool.find(task_type);
+    if (task_it == m_task_pool.end()) {
+        LOG(ERROR) << "task of type " << int(task_type) << " does not exist in the task_pool";
+        return;
+    }
+
+    auto &task = task_it->second;
+    task->handle_event(event);
+}
+
+bool TaskPool::handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, sMacAddr src_mac,
+                           std::shared_ptr<beerocks_header> beerocks_header)
+{
+    for (auto &task_element : m_task_pool) {
+        auto &task = task_element.second;
+        if (task->handle_cmdu(cmdu_rx, src_mac, beerocks_header)) {
+            return true;
+        }
+    }
+    return false;
+}
--- a/agent/src/beerocks/slave/tasks/task_pool.h
+++ b/agent/src/beerocks/slave/tasks/task_pool.h
@@ -1,0 +1,66 @@
+/* SPDX-License-Identifier: BSD-2-Clause-Patent
+ *
+ * SPDX-FileCopyrightText: 2016-2020 the prplMesh contributors (see AUTHORS.md)
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#ifndef _TASK_POOL_H_
+#define _TASK_POOL_H_
+
+#include "task.h"
+#include <unordered_map>
+
+namespace beerocks {
+
+/**
+ * @brief The TaskPool is single instance tasks container which responsible to run each
+ * task and allow passing messages and events from the Agent to the task.
+ * 
+ */
+class TaskPool {
+public:
+    /**
+     * @brief Add new task to the task pool.
+     * 
+     * @param new_task Shared pointer to the task.
+     * @return true on success, otherwise false.
+     */
+    bool add_task(const std::shared_ptr<Task> &new_task);
+
+    /**
+     * @brief Run all tasks on the pool, by calling each task work() function.
+     */
+    void run_tasks();
+
+    /**
+     * @brief Send an 'event' defined on a specific task 'task_type'. 
+     * 
+     * @param task_type Task type, defined on Task base class.
+     * @param event Event type, defined on the task itself.
+     */
+    void send_event(eTaskType task_type, uint8_t event);
+
+    /**
+     * @brief Iterate over all tasks on the pool and pass them the message on 'cmdu_rx'.
+     * 
+     * @param cmdu_rx CMDU object containing the message.
+     * @param src_mac MAC address of the message sender.
+     * @param beerocks_header Beerocks header (Only on VS message).
+     * @return true if the message has been handled, otherwise false.
+     */
+    bool handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, sMacAddr src_mac,
+                     std::shared_ptr<beerocks_header> beerocks_header = nullptr);
+
+private:
+    // Decalaring unordered_map with key which is an enum, does not compiles on older gcc version.
+    // It was considered a defect in the standard, and was fixed in C++14, and also fixed in the
+    // version of libstdc++ shipping with gcc as of 6.1.
+    // To make unordered_map work with an enum as key, std::hash<int> function was added as third
+    // template argument.
+    std::unordered_map<eTaskType, std::shared_ptr<Task>, std::hash<int>> m_task_pool;
+};
+
+} // namespace beerocks
+#endif // _TASK_POOL_H_
--- a/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
+++ b/agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.h
@@ -9,6 +9,7 @@
 #ifndef _BACKHAUL_MANAGER_THREAD_H
 #define _BACKHAUL_MANAGER_THREAD_H
 
+#include "../tasks/task_pool.h"
 #include "wan_monitor.h"
 
 #include <bcl/beerocks_backport.h>
@@ -272,6 +273,8 @@ private:
     std::string bssid_bridge_mac;
 
     std::unique_ptr<beerocks::agent_ucc_listener> m_agent_ucc_listener;
+
+    TaskPool m_task_pool;
 
     /**
      * AP Metrics Reporting configuration and status information type.


agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,47 @@
...
+ * See LICENSE file for more details.
+ */
+
+#include "task_pool.h"
+
+using namespace beerocks;
+
+void TaskPool::add_task(std::shared_ptr<Task> new_task)
+{
+    m_task_pool[new_task->get_task_type()] = new_task;
```
adam1985d, line 19: what if the new_task shared pointer is null?
you dereference it without checking

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,47 @@
...
+        auto &task = task_element.second;
+        task->work();
+    }
+}
+
+void TaskPool::send_event(Task::eType task_type, uint8_t event)
+{
+    auto task_it = m_task_pool.find(task_type);
+    if (task_it == m_task_pool.end()) {
+        LOG(ERROR) << "task of type " << int(task_type) << " is not exist on the task_pool";
```
adam1985d: you don't return after reaching the end of task-pool and then dereference the iterator
either `return;` here or wrap the following functionality with successful case

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,47 @@
...
+void TaskPool::add_task(std::shared_ptr<Task> new_task)
+{
+    m_task_pool[new_task->get_task_type()] = new_task;
+}
+
+void TaskPool::run_tasks()
+{
+    for (auto &task_element : m_task_pool) {
+        auto &task = task_element.second;
+        task->work();
```
adam1985d, line 27: dereference of shared pointer without checking it first

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+void TaskPool::send_event(Task::eType task_type, uint8_t event)
+{
+    auto task_it = m_task_pool.find(task_type);
+    if (task_it == m_task_pool.end()) {
+        LOG(ERROR) << "task of type " << int(task_type) << " is not exist on the task_pool";
+    }
+
+    auto &task = task_it->second;
+    task->handle_event(event);
```
adam1985d, line 40: dereference of shared pointer without checking it first

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,47 @@
...
+}
+
+bool TaskPool::handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, sMacAddr src_mac,
+                           std::shared_ptr<beerocks_header> beerocks_header)
+{
+    for (auto &task_element : m_task_pool) {
+        auto &task = task_element.second;
+        if (task->handle_cmdu(cmdu_rx, src_mac, beerocks_header)) {
```
adam1985d, line 48: dereference of shared pointer without checking it first

agent/src/beerocks/slave/tasks/task_pool.h
```diff
@@ -0,0 +1,60 @@
...
+     */
+    void run_tasks();
+
+    /**
+     * @brief Send an 'event' defined on a specific task 'task_type'. 
+     * 
+     * @param task_type Task type, defined on Task base class.
+     * @param event Event type, defined on the task itself.
+     */
+    void send_event(Task::eType task_type, uint8_t event);
```
adam1985d: what happens if you have 2 tasks of the same type? task type isn't enough

agent/src/beerocks/slave/tasks/task_pool.h
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * @brief Iterate over all tasks on the pool and passing them the message on 'cmdu_rx'.
```
adam1985d: nitpick: replace `passing` with `pass`

agent/src/beerocks/slave/tasks/task_pool.h
```diff
@@ -0,0 +1,60 @@
...
+ * 
+ */
+class TaskPool {
+public:
+    /**
+     * @brief Add new task to the task pool.
+     * 
+     * @param new_task Shared pointer to the task.
+     */
+    void add_task(std::shared_ptr<Task> new_task);
```
adam1985d: if you have an `add_task` method - you should have a `remove_task` method as well

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2187,9 +2187,13 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<sRadioInfo>
         break;
     }
     default: {
-        LOG(ERROR) << "Unhandled message received from master: "
-                   << int(beerocks_header->action_op());
-        return false;
+        bool handled = m_task_pool.handle_cmdu(cmdu_rx, sMacAddr(), beerocks_header);
```
adam1985d, line 2176: nitpick:
I think instead of handling each type separately (1905/backhaul/etc..) there should be one call from the `backhaul_manager::handle_cmdu` main method (the one that calls the different types of handling.
I understand this required refactoring the main method and you didn't want to get into this for this PR - but consider this change in the coming PRs

adam1985d (changes requested):
I really liked how this PR is limited to the task_pool only and doesn't introduce the actual use-cases
makes it much easier to review.
Kudos!

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: fixed.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: Actually that could not be dereferenced to nullptr since it is a shared_ptr. As long as the class holding the pointer, it is promised that the pointer is valid.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: fixed.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: Same answer:
https://github.com/prplfoundation/prplMesh/pull/1485#discussion_r446616830

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: Same answer:
https://github.com/prplfoundation/prplMesh/pull/1485#discussion_r446616830

agent/src/beerocks/slave/tasks/task_pool.h
morantr: Currently, the task pool is designed for a single instance task which is implementing some Agent responsibility.
Meaning, that tasks should not be removed, at least of now POV.
If we will need such functionality, it will be added in the future.

agent/src/beerocks/slave/tasks/task_pool.h
morantr: Not possible. This task pool support a single instance of a specific task.

agent/src/beerocks/slave/tasks/task_pool.h
morantr: changed.

itayx (approved):
Looks good to me

agent/src/beerocks/slave/tasks/task_pool.h
adam1985d: I don't think this is good.
the DCS, for example, has 2 task instances - one per radio.
we are currently limited to one-scan-at-a-time - and there isn't a requirement for simultaneous scans on multiple radios at the same time - but if that will change - the task class should not be hard limiting us like that.

an index with default value 0 is preferable, or some other key to allow multiple tasks of the same type.
(side-note - the current controller task_pool support task id for that - but you can suggest something else of course)
@kantera800

agent/src/beerocks/slave/tasks/task_pool.h
adam1985d: @arnout would appreciate your thoughts on this

agent/src/beerocks/slave/tasks/task_pool.h
morantr: You can manage DCS for different radios under the same task.
This is the new approach now. The unified agent will manage all fronthauls so the topology task and so on.

adam1985d: ok

adam1985d (approved):
good small and reviewable PR
need to address the 2 topics left open in future PRs:
1. moving the call to the tasks' handle_cmdu() one lever up (to the main function handle_cmdu)
2. should task_pool have a remove_task Method

agent/src/beerocks/slave/tasks/task_pool.cpp
RanRegev: `add_task` should ask for a reference to `Task` instead of a pointer.
If you can't handle nullptr, **do not allow it in your interface**.

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @brief Enum which defines all tasks type.
+     * 
+     * Each task must have an entry on the enum eType, so it will be possible to send the task an
+     * event. 
+     */
+    enum eType : uint8_t {
+        TOPOLOGY,
+    };
```
RanRegev:
```suggestion
```
It is a bad idea to start moving enums all over. It is even worse as _users_ of the Task class need to modify it (this is a bad sign for a problematic design).
See below about the alternative.

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include <beerocks/tlvf/beerocks_header.h>
+#include <tlvf/CmduMessageRx.h>
+
+namespace beerocks {
+
+class Task {
+public:
+    Task() {}
+    virtual ~Task() {}
```
RanRegev, line 31: make it pure.
```suggestion
    virtual ~Task() = 0;
```

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+    };
+
+    /**
+     * @brief Retrun the task type.
+     * 
+     * Called by TaskPool send_event().
+     * 
+     * @return eType Task type.
+     */
+    virtual eType get_task_type() = 0;
```
RanRegev: No need to deal with enums, over complicated.
```suggestion
```

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @brief Work function. Shall do the task routine. Called by TaskPool run_task() function.
+     */
+    virtual void work(){};
```
RanRegev, line 45: Do we expect any derived `Task` to have _correct_  empty `work()` implementation?
If yes, then fine, the default implementation is correctly set as empty.
If not, and it is illegal to have an empty `work` - please state it clearly -
```suggestion
    virtual void work() = 0;
```

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @brief Handle events, which are being send from containing thread.
+     * 
+     * @param event_enum_value Event enum value which shall be defined on the task.
```
RanRegev, line 50: If we expect enum, it is not stated in the interface.
If we expect uint8_t then we shouldn't mention enums. It is up to the user to decide what this uint8_t is representing.

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+     */
+    virtual void handle_event(uint8_t event_enum_value){};
```
RanRegev, line 52: 1. At his point I really feel that I would like to have a documentation explaining the idea of `Task` and what it should do and how it is designed.
2. It seems strange that the default implementation does nothing with the event.
```suggestion
    virtual void handle_event(uint8_t event_enum_value) = 0;
```

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @brief Handle CMDU message.
+     * 
+     * @param cmdu_rx CMDU object conataining the message.
+     * @param src_mac AL Mac of the sender.
+     * @param beerocks_header Beerocks Message header (Only on VS message).
+     * @return true if the message has been handled, otherwise false.
+     */
+    virtual bool handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, const sMacAddr &src_mac,
```
RanRegev, line 62: I think it is better to pack all required variables for handling a cmdu into a struct that can be passed along. We keep attaching the required variables separately, and I think we should state clearly what is required to andle a cmdu-message.
I might be wrong here, but I have the feeling that it would be better to encapsulates the needs together:
```cpp
class CmduHandleContext{
    /* setting and validation of the internal relationships between the various
    fields that are required to handle cmduRx.
    e.g. is it legal to have beerocks_header == nullptr?
    e.g. can we work with ZERO_MAC as the src addr? does it imply somethimg?
    etc.
     */
private:
    ieee1905_1::CmduMessageRx &m_cmdu_rx;
    const sMacAddr &m_src_mac;
    std::shared_ptr<beerocks_header> m_beerocks_header;

Comment (RanRegev) [inline agent/src/beerocks/slave/tasks/task.h:62]: Also, if by default, we fail to handle cmdu, why should we catch it only in run-time? 
Let's force the implementer to add a code here. In this way we prevent the problem instead of solving it later.
```suggestion
    virtual bool handle_cmdu(CmduHandleContext cmdu_context) = 0;
```

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * This code is subject to the terms of the BSD+Patent license.
+ * See LICENSE file for more details.
+ */
+
+#include "task_pool.h"
+
+using namespace beerocks;
+
+bool TaskPool::add_task(std::shared_ptr<Task> new_task)
```
RanRegev:
```suggestion
bool TaskPool::add_task(Task& new_task)
```

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    if (!new_task) {
+        LOG(ERROR) << "new task pointer is nullptr";
+        return false;
+    }
```
RanRegev, line 18: Gratuitous.
Ask for a reference (i.e. a real object) and remove this check. 
Also:
1. prevents the issue instead of solving it later (by seeing in the log that there is a problem).
2. faster at run time (a bit, but when apply to all places we ask for a pointer and check it before doing something... maybe we would see an overall improvement).
```suggestion
```

agent/src/beerocks/slave/tasks/task_pool.h
```diff
@@ -0,0 +1,66 @@
...
+    bool handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, sMacAddr src_mac,
+                     std::shared_ptr<beerocks_header> beerocks_header = nullptr);
+
+private:
+    // Decalaring unordered_map with key which is an enum, does not compiles on older gcc version.
+    // It was considered a defect in the standard, and was fixed in C++14, and also fixed in the
+    // version of libstdc++ shipping with gcc as of 6.1.
+    // To make unordered_map work with an enum as key, std::hash<int> function was added as third
+    // template argument.
+    std::unordered_map<Task::eType, std::shared_ptr<Task>, std::hash<int>> m_task_pool;
```
RanRegev: No need in Task::eType, see other comments on the issue.
```suggestion
    std::vector<Task*> m_task_pool;
```

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,53 @@
...
+    m_task_pool[new_task->get_task_type()] = new_task;
```
RanRegev, line 19:
```suggestion
    m_task_pool.push_back(&new_task);
```

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+void TaskPool::run_tasks()
+{
+    for (auto &task_element : m_task_pool) {
+        auto &task = task_element.second;
+        task->work();
+    }
+}
+
+void TaskPool::send_event(Task::eType task_type, uint8_t event)
```
RanRegev: Here I really think we are over-engineering a simple thing.
Please follow these lines -
* In this function we are _searching_ for previously, (probably, no interface is given to do it at the moment) installed etype that is associated with a `Task` (not-null) pointer.
* When found we are simply calling `handle_event()` on it.
* when not found, we are producing an error.
From the **_caller_** point of view, the one that calls `handle_event(event)` - the `task_type` is **known**  at coding time, otherwise she wouldn't be able to call this function (this is the first parameter to `send_event`).
```cpp
// somewhere in the code:
task_pool.register_eType(beerocks::eType::TOPOLOGY, topology_task); // this function does not exists in this PR, but it should be.
```
and later in the code:
```cpp
task_pool.send_event(beerocks::eType::TOPOLOGY, event);
```
This creates an unnecessary layer of complexity to the code.
Instead of registering and then calling the pool with enum, we could simply ask the task to handle the event directly:
```cpp
topology_task.handle_event(event);
```
The point is that we KNOW the task that is going to handle the event - we pass its enum to the pool.
So instead of knowing by this additional layer, we could simply keep the correct `Task` object and call it directly.
We really have all the required information, so `TaskPool` is just another layer of complexity:
```cpp
Class MyCode{
// either as members -
Task topology_task;
Task abc_task;
// or references - 
Task& e_task;
};
```

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    auto &task = task_it->second;
+    task->handle_event(event);
+}
+
+bool TaskPool::handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, sMacAddr src_mac,
+                           std::shared_ptr<beerocks_header> beerocks_header)
+{
+    for (auto &task_element : m_task_pool) {
```
RanRegev, line 46: OK: Here we are looping over _all tasks_ and letting each one handle the CMDU, so a list and a pool is required.

agent/src/beerocks/slave/tasks/task_pool.h
RanRegev: how this list is populated?
I am missing an interface to do so.

agent/src/beerocks/slave/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ file(GLOB beerocks_agent_sources
     # [TASK] Move link metric related classes to BPL #910
     ${MODULE_PATH}/link_metrics/*.c*
     ${MODULE_PATH}/platform_manager/*.c*
+    ${MODULE_PATH}/tasks/*.c*
```
RanRegev, line 13: I really prefer that we switch for explicit naming of the files we want to be part of the build.
But OK for now to keep the same method and change it all together later.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2255,8 +2260,12 @@ bool backhaul_manager::handle_slave_1905_1_message(ieee1905_1::CmduMessageRx &cm
         return handle_slave_channel_selection_response(cmdu_rx, src_mac);
     }
     default: {
-        LOG(DEBUG) << "Unexpected 1905 message " << int(cmdu_rx.getMessageType());
-        return false;
+        bool handled = m_task_pool.handle_cmdu(cmdu_rx, tlvf::mac_from_string(src_mac));
+        if (!handled) {
+            LOG(DEBUG) << "Unexpected 1905 message " << int(cmdu_rx.getMessageType());
```
RanRegev, line 2269: nitpick
```suggestion
            LOG(DEBUG) << "Unhandled 1905 message " << int(cmdu_rx.getMessageType());
```

RanRegev (changes requested):
I would like to have at least a discussion regarding my comments.

agent/src/beerocks/slave/tasks/task.h
morantr: I didn't want to make it pure. If needed it can be overridden on the derived class.

agent/src/beerocks/slave/tasks/task.h
morantr: It is possible that a task would not have a work function. Same as handling CMDUs or events.
For example, a task that only responsible for handling CMDUs.

agent/src/beerocks/slave/tasks/task.h
morantr: It is stated:
"@param event_enum_value **Event enum value which shall be defined on the task**."

Wasn't it clear enough?

agent/src/beerocks/slave/tasks/task.h
morantr: This is because not all tasks will necessarily handle events.
It is optional.

agent/src/beerocks/slave/tasks/task.h
morantr: For what? So I would have to create a struct, copy to it all data members before passing it to the handler?
I don't agree with it.

morantr: The boolean returned value is for indicating that the message has been handled or not. Not for indicating that the handler function has failed or not.
That why it returns false by default, so if the Task derived class will not implement it, the task pool will continue iterating on other tasks, on TaskPool:handle_cmdu().

I have stated it on the Doxygen documentation.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: Sorry, but that is a terrible idea because it forces the caller to keep the task object. Not something we want here.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: No, it doesn't fit our needs. See my comment above.
If we declare the task object in some scope, that will force us to save it somewhere.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: Can you explain why do you think it is a better approach?
The unordered map allows fast lookup, when receiving events.
What is the benefit of vector here? iterating on all tasks is linear on both containers.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: I respect your opinion, but I don't agree with it. I didn't want to hold the task object on the caller. I wanted to create a managing layer that will handle it.
Also when handling cmdu for instance, on your suggestion I would have to call each tasks "handle_cmdu()" separately.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: But on the event handler, I use "find()".
Iterating on map and list/vector are both linear. So it doesn't make much difference.

agent/src/beerocks/slave/tasks/task_pool.h
morantr: I don't understand the question.
Can you please explain.

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: Actually Unexpected fit here more IMO.

agent/src/beerocks/slave/tasks/task_pool.cpp
RanRegev: > Also when handling cmdu for instance, on your suggestion I would have to call each tasks "handle_cmdu()" separately.

No. For CMDU it works fine. For `handle_event` it is over engineered.

agent/src/beerocks/slave/tasks/task.h
vitalybu: > make it pure.

Why? What's the benefit here with a virtual destructor?

agent/src/beerocks/slave/tasks/task.h
vitalybu: It's OK to have an enum for the task types, but it shouldn't be part of the base class.
I think it should be an `enum class` in a separate file, and the constructor should accept a `type` argument.
This will also prevent the need for implementing `get_task_type()` in every derived class.

agent/src/beerocks/slave/tasks/task.h
vitalybu: So how do you handle errors? Or the `TaskPool` doesn't care about errors?

agent/src/beerocks/slave/tasks/task.h
```diff
@@ -0,0 +1,70 @@
...
+                             std::shared_ptr<beerocks_header> beerocks_header)
```
vitalybu, line 63: If the `beerocks_header` argument is optional, I suggest providing a `nullptr` default value

agent/src/beerocks/slave/tasks/task.h
vitalybu: This class is not really a base class (as there's no logic) nor an interface (as it's not pure-virtual).
I think that making it an interface and forcing the user to implement all the methods is not a bad idea.

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,53 @@
...
+        auto &task = task_element.second;
+        task->work();
+    }
+}
+
+void TaskPool::send_event(Task::eType task_type, uint8_t event)
+{
+    auto task_it = m_task_pool.find(task_type);
+    if (task_it == m_task_pool.end()) {
+        LOG(ERROR) << "task of type " << int(task_type) << " is not exist on the task_pool";
```
vitalybu:
```suggestion
        LOG(ERROR) << "task of type " << int(task_type) << " does not exist in the task_pool";
```

agent/src/beerocks/slave/tasks/task_pool.cpp
```diff
@@ -0,0 +1,53 @@
...
+    auto &task = task_it->second;
+    task->handle_event(event);
+}
+
+bool TaskPool::handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, sMacAddr src_mac,
+                           std::shared_ptr<beerocks_header> beerocks_header)
+{
+    for (auto &task_element : m_task_pool) {
+        auto &task = task_element.second;
+        if (task->handle_cmdu(cmdu_rx, src_mac, beerocks_header)) {
```
vitalybu, line 48: I think that in most cases, a CMDU will be relevant for 1 task.
Maybe we should allow the tasks to "register" for specific message types?
Then the pool will be able to pass this CMDU only to the tasks expecting to receive it...

agent/src/beerocks/slave/tasks/task_pool.h
```diff
@@ -0,0 +1,66 @@
...
+ */
+class TaskPool {
+public:
+    /**
+     * @brief Add new task to the task pool.
+     * 
+     * @param new_task Shared pointer to the task.
+     * @return true on success, otherwise false.
+     */
+    bool add_task(std::shared_ptr<Task> new_task);
```
vitalybu: I think it's safe to make this a const reference, no need to create additional copy as another one will be created when storing the task in the map:
```suggestion
    bool add_task(const std::shared_ptr<Task> &new_task);
```

agent/src/beerocks/slave/tasks/task_pool.h
vitalybu: > No need in Task::eType, see other comments on the issue.

I don't understand your comment. You wrote "No need in Task::eType", but replaced the unordered_map with a vector.
Lookup times are much better in an unordered_map..

vitalybu: > how this list is populated?
> I am missing an interface to do so.

https://github.com/prplfoundation/prplMesh/pull/1485/commits/21cc249484f6994b73601a270fd0e3e9331118e7#diff-2adb7ad23cc08452466296d700b98e6dR19

agent/src/beerocks/slave/CMakeLists.txt
vitalybu: Why? What do you have against GLOBs?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
```diff
@@ -2169,9 +2169,13 @@ bool backhaul_manager::handle_slave_backhaul_message(std::shared_ptr<sRadioInfo>
         break;
     }
     default: {
-        LOG(ERROR) << "Unhandled message received from master: "
-                   << int(beerocks_header->action_op());
-        return false;
+        bool handled = m_task_pool.handle_cmdu(cmdu_rx, sMacAddr(), beerocks_header);
+        if (!handled) {
+            LOG(ERROR) << "Unhandled message received from master: "
```
```suggestion
            LOG(ERROR) << "Unhandled message received from controller: "
```

agent/src/beerocks/slave/tasks/task.h
morantr: What is to handle? We print an error and stop the handling on the task. We have nothing to do about it on the outside.
Also, if you follow the return value of the handlers functions across the code you will see that except for printing error we do nothing everywhere.

agent/src/beerocks/slave/tasks/task.h
RanRegev: > Why? What's the benefit here with a virtual destructor?

When you leave a default implementation you encourage users to forget freeing resources they acquired.
By forcing them to implement a destructor at least you give them the opportunity to remember.

agent/src/beerocks/slave/tasks/task.h
RanRegev: nitpick, m2c -
Comments are not the interface, they are used as a complementary for the code. They should not contradict it or add embarrassment to the reader.
In this specific case: there is the word `enum` both in the comment and in the variable name. But the _type_ of the variable is `unit8_t`.
So now, as a user I am confused: should I declare an enum? what this enum represent? etc.
The fact that `unit8_t` can be converted to to an enum is not important.
For example I can use this `uint8_t` value as an index to a array, or use it as the code of a CMDU message.
The intent of the function is not clear, it is not aligned with the interface.
It might look like of over-engineering, or overlooked issue - after all, what is the issue here, the comment? - when occurred once it is not an issue, but when a project have more and more of these, they make it harder to maintain, and of course bugs are introduced.

agent/src/beerocks/slave/tasks/task.h
RanRegev: For many reasons:
>   setting and validation of the internal relationships between the various
>   fields that are required to handle cmduRx.
>    e.g. is it legal to have beerocks_header == nullptr?
>    e.g. can we work with ZERO_MAC as the src addr? does it imply somethimg?
>    etc.

The alternative is to leave these validations and integrity of the data to the writer of the function. In this case either:
* it wouldn't be done.
* it would be done wrongly
* it would be re-written (copy/pasted) along the code.

RanRegev: > The boolean returned value is for indicating that the message has been handled or not. Not for indicating that the handler function has failed or not.

This is counter intuitive, exactly as it is today, and also as @vitalybu said, does not leave space for error-handling.
It also prevents more than one client to handle a message. When needed, patches are starting to pop like: "ok we'll return false here because we want the other task to be called as well".
The `return true means was handled` idiom should be replaced with something else, we failed with it.
One Option:
```suggestion
    // ret.first - handled/not handled
    // ret.second - success/fail
    // (true,true) - handled succefully. 
    // (true,false) - handled but error occurred 
    // (false,don't care) - not handled.
    virtual std::pair<bool,bool> handle_cmdu(ieee1905_1::CmduMessageRx &cmdu_rx, const sMacAddr &src_mac,
```

Another option, as @vitalybu suggested, some kind of registration to the CMDU that are with interest by the task.
In this way, the return value may be a simple `bool` indicating success/fail.

agent/src/beerocks/slave/tasks/task_pool.cpp
RanRegev: There are many ways to express ownership in C++. Ownership is _important_ in code structure as it sets lifetimes, responsibilities and overall structure. Declaring `std::shared_ptr` is our "easy" way to forget about ownership and lifetime of objects. It is like that we are having a garbage collector to keep the memory and lifetime in a good shape.
So the question raised based on your comment is: who is responsible to keep the Task object?
If it is someone else, outside the `TaskPool`, then fine, just request a reference, someone else will have the real object.
If it is the `TaskPool` that keeps the task objects (sounds reasonable to me), then state it clearly:
```suggestion
bool TaskPool::add_task(Task&& new_task);
```
Now the ownership is _moved_ to the TaskPool.
If you could also handle a nullptr (which you can't, but for the example only) -
```suggestion
bool TaskPool::add_task(std::unique_ptr<Task> new_task)
```
Also I am not against _copying_ the task internally into the `TaskPool` -
```suggestion
bool TaskPool::add_task(Task new_task) {
m_task_pool.push_back(std::move(new_task));
}
```

agent/src/beerocks/slave/tasks/task_pool.cpp
RanRegev: See my comment above.
You don't want to leave the ownership decision left alone. you really want to _decide_ who is the owner.
Only in rare cases, usually when working with multi-threading and you **really** don't know who will need the object last, you would correctly select `std::shared_ptr` - because then it is really shared between more than one, and all have real ownership of it.
In other cases it is best to decide who holds the object, who is responsible for its lifetime. All other entities in the system just declare their needs of the object (a reference, a pointer, a copy or even requesting an ownership by `&&` or `unique_ptr`)

agent/src/beerocks/slave/tasks/task_pool.h
RanRegev: My claim is that there is no need to wrap the call to `task.handle_event` with `task_pool.handle_event`.
We gain nothing, and we add more complexity where it is not needed.
The entire enum is redundant. We don't need a map, we don't need to search at all.
We keep the vector for `cmdu` to loop over _all_ tasks.
`send_event`, as declared in this interface just sends the event to the single task that its enum was requested. therefore, the caller already knows who to send the event to, so there is no need to go via the task pool:
It should be as simple as: `task_a.send_event(0x45);`  which is much simpler, faster and cleaner then:
`task_pool.send_event( eTaskA, 0x45);`

agent/src/beerocks/slave/tasks/task_pool.cpp
RanRegev: No. See my other comment.
There is no need to search at all. 
The vector is for iterating on all, not for lookup.

agent/src/beerocks/slave/CMakeLists.txt
RanRegev: I have a lot against GLOB, but my opinion is not important.
The developers of CMake [recommend](https://cmake.org/cmake/help/latest/command/file.html#filesystem) not to use GLOB.

>  Note: We do not recommend using GLOB to collect a list of source files from your source tree.

They give one reason that might be irrelevant to us, but there are more reasons.

agent/src/beerocks/slave/tasks/task_pool.cpp
RanRegev: This is sound reasonable. But also might over complicate the mechanism.
In the way it is written now, each task first needs to somehow check if this CMDU belongs to it or not.
I am not sure it is that bad.
The difference between the two suggestions is this:
Checking is done _once_ at the TaskPool before sending, but we add a registration complexity.
Checking is done _many_ times, in each Task, but we save the registration complexity.

RanRegev (approved):
There are lots of issues that are of general issues about the way of writing C++ code, interfaces, intentions, and so on that I mentioned.
Overall I feel that changes are needed, but they are also can be refactored later.

agent/src/beerocks/slave/tasks/task.h
morantr: I thought about defining it in a separate class or file, but then I thought it would be better if the user will have to write `Task::eType::TASK_NAME`. It put the task name in the context of the Task class which I think is more correct.
Also, IMO it is ok to define the type on Task class because it defines what kind of tasks could be derived from it, just like pure virtual function force the derived class to implement them.

As for the `get_task_type()`. Actually that what I did it first. But when I used it looked a little bit weird:
```cpp
    auto topology_task = std::make_shared<TopologyTask>(*this, cmdu_tx);
    
    // Adding the enum to the constructor seems redundant because I already defined task of type "TopologyTask", so
    // it look weird to state it again as a constructor argument.
    m_task_pool.add_task(Task::eType::TOPOLOGY, topology_task);

    // This looks better IMO. The topology task "know" which task type is it. No need to tell it.
    m_task_pool.add_task(topology_task);
```

I don't think there is a right or wrong here.

agent/src/beerocks/slave/tasks/task.h
morantr: I don't think there will be resources to free. We don't use `new` so we have to use `delete` on the destructor. Almost every resource we use it automatically being freed.

agent/src/beerocks/slave/tasks/task.h
morantr: I try to explain it as much as I could to make it clear.
I wanted that each task will define its own event, which will be passed through the task_pool.
On the task pool, I must give it a uint8_t.
If you have a suggestion on how to do it more clear, please enlighten me.
Keeping the task object on the caller class without using the task pool is not an option.

agent/src/beerocks/slave/tasks/task.h
morantr: It is legal to have beerocks_header=nullptr.
And it is legal that src_mac=ZERO_MAC.

agent/src/beerocks/slave/tasks/task.h
vitalybu: That's not what I meant. The task type should be an argument to the constructor of the base Task class, so that when you are implementing a new task, you'll pass it's type to the base class:
```c++
class TopologyTask : public Task
{
    TopologyTask() : Task(eTaskType::Topology) {}
    ...
};
```

`add_task` should only get the task instance as it'll be able to read the type by calling `get_task_type()`.

And no, it's not OK to modify the base class every time you add a new task.
It should be an interface (as there's no implementation here right now).

agent/src/beerocks/slave/tasks/task_pool.h
vitalybu: Task pool is a container of tasks. For example, it allows sending an event to a specific task without having a reference to this task. How do you achieve this without a container?

agent/src/beerocks/slave/tasks/task.h
morantr: I defined it with a default argument on the task pool. So if it is not filled it will pass the default argument of the task pool anyway.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: done.

agent/src/beerocks/slave/tasks/task_pool.cpp
morantr: It will add more code and complexity, which is unnecessary, with no actual valuable benefit.

agent/src/beerocks/slave/tasks/task_pool.h
morantr: But then the reference count will not increase, and the memory will be freed at the end of the caller scope.
And the task will hold invalid memory.
Am I missing something?

agent/src/beerocks/slave/backhaul_manager/backhaul_manager_thread.cpp
morantr: fixed.

agent/src/beerocks/slave/tasks/task.h
RanRegev: It is not OK to know in the base class the type of the derived class. period.
It breaks the essence of interface.
Once you know the type of the derived there is no more meaning to have a base class.
Adding an enum to define the type of the concrete instance is to know its type.
Instead, we should work with the type itself, and mass around with numbers-that-represent-the-concrete-type (enum).
enums to indicate types are just another layer of complexity.

agent/src/beerocks/slave/tasks/task.h
RanRegev: > I don't think there will be resources to free

That is the the beauty of interfaces, you don't need to think - You simply force.

agent/src/beerocks/slave/tasks/task.h
RanRegev: > Keeping the task object on the caller class without using the task pool is not an option.

This is the best option.
Other than that:
```suggestion
     * @param message uint8_t that represent anything in the context of the targeted Task
```

agent/src/beerocks/slave/tasks/task.h
RanRegev: Once you have a structure (class) that represents CMDU to be handled you _have a place to enforce, to validate, to check_ the relation between these parameters.
Even if now you don't see it, I am quite sure that there _are_ relationships that we would like to enforce, now, or in the future. Spreading these 3 parameters in each function call just makes it much harder to enforce them now or in the future.
And don't worry about copy - this structure has only references and pointers to existing objects. it is cheep - i.e. you will not pay more than you pay now when you pass the variables separately.

agent/src/beerocks/slave/tasks/task.h
vitalybu: Saying that it's not ok is a bit of an overstatement. It's perfectly fine.
At some point you have to figure out the type of the class you're working with. 
Polymorphism is not only about implementing interfaces but also about extending the interface with additional functionality which is not necessarily available in the interface you're implementing.
It can be achieved in several ways, and storing the type in the base class is one of them.
I guess you could also use `dynamic_cast` but in some (most?) implementations it relies on RTTI, which can be disabled in embedded systems.

agent/src/beerocks/slave/tasks/task_pool.h
vitalybu: The reference count will increase once the pointer is stored in the map.
When you pass it by value, the reference count increases twice:
1. When the pointer is copied to the function argument.
2. When the pointer is stored in the map.

Once the function ends, the reference count decreases by 1.
By passing a reference, you prevent the extra increase-decrease.

agent/src/beerocks/slave/tasks/task.h
RanRegev: I think it is better to force.

agent/src/beerocks/slave/tasks/task_pool.h
RanRegev: > it allows sending an event to a specific task without having a reference to this task

Q: How?
A: We are keeping a handle (enum) to the task instead of the task itself.

We know at the point of calling the task we want to send a message to.
By sending it to a third party (TaskPool) to call it for us we are doing overwork.
It is much simpler to send the message directly to the task that we want, without the loop via the task-pool. In this way we gain:
* much simpler code - no enum at all.
* expressiveness - we state what we do directly: `task_x.handle(message)`
* run time - no need to search at all.

agent/src/beerocks/slave/tasks/task.h
morantr: I moved the enum type outside of the class, but still on the same file, and set the infrastructure to what Vitali has suggested.

agent/src/beerocks/slave/tasks/task_pool.h
morantr: done :-)

agent/src/beerocks/slave/tasks/task_pool.h
arnout: I like the "if we need it, it will be added" kind of approach.

It does make sense to think about it now - but adding a remove_task method is clearly something that can be done later.

agent/src/beerocks/slave/tasks/task_pool.cpp
arnout: > `add_task` should ask for a reference to `Task` instead of a pointer.
> If you can't handle nullptr, **do not allow it in your interface**.

Although I agree with the principle, the problem is that we need to express two things at the same time:
- if it is optional (i.e., can be nullptr);
- lifetime/ownership of the object.

Here, the shared_ptr is used to indicate co-ownership between the task pool and the caller - both of them can simultaneously keep the task alive. If we pass it as reference, this co-ownership is not possible.

I don't know if it is really *needed*, but it's definitely convenient to have the option.

Is there any other way to express this shared ownership? Or a pointer that is not allowed to be nullptr? Should we define a shared_ref type that is a shared_ptr but can't be nullptr?

agent/src/beerocks/slave/tasks/task_pool.h
arnout: I think that the background is that with the vendor-specific messages, it is possible to include the task ID in the CMDU itself, so that way you can find out which task to send it to. I think... @morantr @vitalybu ?

kantera800 (approved):
I agree with some of the comments especially the ones that prefer simple stupid methods that don't leave less room for human error.
But, since there is no prplmesh coding style violations and no bugs i think that every developer should have the ability to implement "his own way".
So approving.

PR #1486, opened 2020-06-26
boardfarm: use names of devices, instead of searching through devices

abelog:
## Description

In the `initial_ap_config`, `ap_config_renew` tests we are locating test participants (agents, controller and etc) by checking "agent_entity" field in the device class.
This works **only** in case if **all** devices defined in boardfarm configuration are of same class (or have this field). Adding some device, which doesn't have such field will cause tests to crash, as we are trying to reach non-existing field of class. 

In #1453, we are changing boardfarm configuration file to use names which are part of [build-in boardfarm enum](https://github.com/mattsm/boardfarm/blob/100521fde1fb67536682cafecc2f91a6e2e8a6f8/boardfarm/lib/DeviceManager.py#L25), as consequence we can address devices by name. So, we can now safely address devices by name and avoid crash described above.

## Changes

- Replace search through array of device available to test by assign of devices based name
- Related to such rename refactoring

## Notes

This PR should be merged **after** #1453.

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
@@ -14,14 +14,11 @@ class ApConfigRenew(PrplMeshBaseTest):
     """Check initial configuration on device."""
 
     def runTest(self):
-        # Locate agents and controller
-        for dev in self.dev:
-            if dev.controller_entity:
-                controller = dev.controller_entity
-            if dev.agent_entity:
-                agent = dev.agent_entity
+        # Locate test participants
+        agent = self.dev.DUT.agent_entity
+        controller = self.dev.wan.controller_entity
 
-        dev.wired_sniffer.start(self.__class__.__name__ + "-" + dev.name)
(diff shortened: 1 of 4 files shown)


abelog:
I have noticed issue with this PR: we should keep same definition of DUT (boardfarm threats first device defined in setup as DUT) across setups. So i have to recreate this branch and include fix for that.
So, marking this PR as draft.

odkq (approved):

aciliketcap (approved):

tests/boardfarm_plugins/boardfarm_prplmesh/tests/ap_config_renew.py
```diff
@@ -35,7 +32,7 @@ def runTest(self):
                                             tlv(0x0F, 0x0001, "{0x00}"),
                                             tlv(0x10, 0x0001, "{0x00}"))
 
-        time.sleep(30)
+        time.sleep(5)
```
aciliketcap, line 35: (nitpick) I think this should be in fixed in another PR but opening a PR just for this is too much.

arnout (approved):

PR #1490, opened 2020-06-28
Decouple BCL from mapfcommon

vitalybu:
BCL depend on mapf::common for a single function [copy_string()](https://github.com/prplfoundation/prplMesh/blob/master/framework/common/utils.cpp#L29).

As a small step towards https://github.com/prplfoundation/prplMesh/pull/1373, break this dependency by copying the implementation of this function to the `beerocks::string_utils` class.

This change is separate from the https://github.com/prplfoundation/prplMesh/pull/1373 PR, as it may be required for the ZMQ replacement.

### DIFF
--- a/controller/src/beerocks/master/tasks/association_handling_task.cpp
+++ b/controller/src/beerocks/master/tasks/association_handling_task.cpp
@@ -155,7 +155,7 @@ void association_handling_task::work()
             parent_mac); // the bssid which will be reported. for all bssid, use wildcard "ff:ff:ff:ff:ff:ff"
         //measurement_request.params.use_optional_ssid = true;
         measurement_request->params().expected_reports_count = 1;
-        //mapf::utils::copy_string(measurement_request.params.ssid, database.get_hostap_vap_ssid(parent_mac).c_str(), sizeof(measurement_request.params.ssid));
+        //string_utils::copy_string(measurement_request.params.ssid, database.get_hostap_vap_ssid(parent_mac).c_str(), sizeof(measurement_request.params.ssid));
         add_pending_mac(radio_mac, beerocks_message::ACTION_CONTROL_CLIENT_BEACON_11K_RESPONSE);
         TASK_LOG(DEBUG) << "requested beacon measurement request from sta: " << sta_mac
                         << " on hostap: " << parent_mac;
--- a/agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
+++ b/agent/src/beerocks/fronthaul_manager/ap_manager/ap_manager_thread.cpp
@@ -8,6 +8,7 @@
 
 #include "ap_manager_thread.h"
 
+#include <bcl/beerocks_string_utils.h>
 #include <bcl/beerocks_utils.h>
 #include <bcl/network/network_utils.h>
 #include <bcl/son/son_wireless_utils.h>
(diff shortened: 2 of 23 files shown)


morantr (commented):
It is kind of revert this commit:
https://github.com/prplfoundation/prplMesh/commit/a9ed7f9fec88270bf885f70ccd27d02f735e910b

copy_string() was originally placed on bcl, and the above commit has moved it.
So for what moving it back to bcl is good for?

vitalybu:
> It is kind of revert this commit:
> [a9ed7f9](https://github.com/prplfoundation/prplMesh/commit/a9ed7f9fec88270bf885f70ccd27d02f735e910b)
> 
> copy_string() was originally placed on bcl, and the above commit has moved it.
> So for what moving it back to bcl is good for?

mapfcommon duplicates code from BCL (e.g. `Socket`), and this PR (https://github.com/prplfoundation/prplMesh/pull/1373) aims to merge mapfcommon with BCL anyway.
As part of the effort to remove ZMQ, we need to create a client broker class for communicating with the broker server (which will probably be part of the ieee1905 transport). This client will be implemented in mapfcommon, but will depend on some BCL classes, so the dependency should be the other way around (`mapfcommon->BCL` and not `BCL->mapfcommon`).

morantr (approved):
Ok, thanks for the explanation. Approving, but I think it you should squash all the commits where you changed the use of the function from map to bcl. Clearly you did find-replace, I don't think you need to a commit for the same change just on different lib or module.

controller/src/beerocks/master/tasks/association_handling_task.cpp
```diff
@@ -155,7 +155,7 @@ void association_handling_task::work()
             parent_mac); // the bssid which will be reported. for all bssid, use wildcard "ff:ff:ff:ff:ff:ff"
         //measurement_request.params.use_optional_ssid = true;
         measurement_request->params().expected_reports_count = 1;
-        //mapf::utils::copy_string(measurement_request.params.ssid, database.get_hostap_vap_ssid(parent_mac).c_str(), sizeof(measurement_request.params.ssid));
+        //string_utils::copy_string(measurement_request.params.ssid, database.get_hostap_vap_ssid(parent_mac).c_str(), sizeof(measurement_request.params.ssid));
```
RanRegev, line 158: Please either remove the commented code or explain why do you keep it.

RanRegev (approved):
Now we have _two_ copy/pasted implementations of the same function.
1. when do you expect `mapf::utils::copy_string` will be **deleted**?
2. maybe it is the time to change all `framework/{platform,transport}` to use the new `string_utils::copy_string` from `common`?

controller/src/beerocks/master/tasks/association_handling_task.cpp
vitalybu: I didn't comment out this line, simply updated the commented out call.
Maybe it's a disabled feature? Maybe the task doesn't run without this comment? In any case, it's not part of this PR.

vitalybu:
> Now we have _two_ copy/pasted implementations of the same function.
> 
> 1. when do you expect `mapf::utils::copy_string` will be **deleted**?
> 2. maybe it is the time to change all `framework/{platform,transport}` to use the new `string_utils::copy_string` from `common`?

I believe that the `mapf` version will be deleted as part of https://github.com/prplfoundation/prplMesh/pull/1373

PR #1491, opened 2020-06-28
PPM-11 replace get sta measurement to kv parser

morantr:
replace get sta measurement to kv parser.
Moved out from #1452 to here as a new PR.

### DIFF
--- a/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
+++ b/common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
@@ -677,94 +677,88 @@ bool mon_wlan_hal_dwpal::update_vap_stats(const std::string &vap_iface_name, SVa
 bool mon_wlan_hal_dwpal::update_stations_stats(const std::string &vap_iface_name,
                                                const std::string &sta_mac, SStaStats &sta_stats)
 {
-    char *reply = nullptr;
+    const char *tmp_str;
+    int64_t tmp_int;
+    parsed_line_t reply;
 
     std::string cmd = "GET_STA_MEASUREMENTS " + vap_iface_name + " " + sta_mac;
 
-    if (!dwpal_send_cmd(cmd, &reply)) {
+    if (!dwpal_send_cmd(cmd, reply)) {
         LOG(ERROR) << __func__ << " failed";
         return false;
     }
 
-    size_t numOfValidArgs[10] = {0}, replyLen = strnlen(reply, HOSTAPD_TO_DWPAL_MSG_LENGTH);
-    uint64_t BytesSent = 0, BytesReceived = 0, PacketsSent = 0, PacketsReceived = 0,
-             LastDataDownlinkRate = 0, LastDataUplinkRate = 0, Active = 0;
-    char ShortTermRSSIAverage[32] = {'\0'};
-    char SNR[32]                  = {'\0'};
-    FieldsToParse fieldsToParse[] = {
-        {(void *)&BytesSent, &numOfValidArgs[0], DWPAL_LONG_LONG_INT_PARAM, "BytesSent=", 0},
-        {(void *)&BytesReceived, &numOfValidArgs[1], DWPAL_LONG_LONG_INT_PARAM,
-         "BytesReceived=", 0},
-        {(void *)&PacketsSent, &numOfValidArgs[2], DWPAL_LONG_LONG_INT_PARAM, "PacketsSent=", 0},
-        {(void *)&PacketsReceived, &numOfValidArgs[3], DWPAL_LONG_LONG_INT_PARAM,
-         "PacketsReceived=", 0},
-        {(void *)&sta_stats.retrans_count, &numOfValidArgs[4], DWPAL_INT_PARAM, "RetransCount=", 0},
-        {(void *)ShortTermRSSIAverage, &numOfValidArgs[5], DWPAL_STR_PARAM,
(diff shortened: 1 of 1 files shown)


common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -677,94 +677,92 @@ bool mon_wlan_hal_dwpal::update_vap_stats(const std::string &vap_iface_name, SVa
...
-         "LastDataUplinkRate=", 0},
-
-        /* Must be at the end */
-        {NULL, NULL, DWPAL_NUM_OF_PARSING_TYPES, NULL, 0}};
-
-    if (dwpal_string_to_struct_parse(reply, replyLen, fieldsToParse, sizeof(SStaStats)) ==
-        DWPAL_FAILURE) {
-        LOG(ERROR) << "DWPAL parse error ==> Abort";
+    // RSSI
+    if (!read_param_str("ShortTermRSSIAverage", reply, &tmp_str)) {
```
vitalybu: Why not have a single `read_param` (or `parse_param`) that accepts different variable types (string, int etc.)?

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -677,94 +677,92 @@ bool mon_wlan_hal_dwpal::update_vap_stats(const std::string &vap_iface_name, SVa
...
         }
     }
 
-    //save average SNR in watt
-    // SNR values are space-separated:
-    for (const auto &snr : beerocks::string_utils::str_split(SNR, ' ')) {
-        float s_float = float(beerocks::string_utils::stoi(std::string(snr)));
-        if (s_float > beerocks::SNR_MIN) {
+    // SNR
+    if (!read_param_str("SNR", reply, &tmp_str)) {
```
arnout: (nitpick, two times)
- don't use tmp_str, but give it a proper name and use it only once;
- declare just before using it.
```suggestion
    std::string SNR;
    if (!read_param("SNR", reply, SNR)) {
```

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
```diff
@@ -677,94 +677,92 @@ bool mon_wlan_hal_dwpal::update_vap_stats(const std::string &vap_iface_name, SVa
...
 
-    // TODO: Update RSSI externally!
-    sta_stats.tx_phy_rate_100kb = (LastDataDownlinkRate / 100);
-    sta_stats.rx_phy_rate_100kb = (LastDataUplinkRate / 100);
-    calc_curr_traffic(BytesSent, sta_stats.tx_bytes_cnt, sta_stats.tx_bytes);
-    calc_curr_traffic(BytesReceived, sta_stats.rx_bytes_cnt, sta_stats.rx_bytes);
-    calc_curr_traffic(PacketsSent, sta_stats.tx_packets_cnt, sta_stats.tx_packets);
-    calc_curr_traffic(PacketsReceived, sta_stats.rx_packets_cnt, sta_stats.rx_packets);
+    // Last Downlink (TX) Rate
+    if (!read_param_int("LastDataDownlinkRate", reply, tmp_int)) {
```
arnout:
```suggestion
    int64_t LastDataDownlinkRate;
    if (!read_param_int("LastDataDownlinkRate", reply, LastDataDownlinkRate)) {
```

arnout (approved):
Why is this still draft?

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
morantr: Why do you against declaring temporary helper variables? It seems redundant to me to declare a new variable to each parameter we want to read. Also, it makes the code size longer, and unnecessarily increases the stack size (lesser performance).
I don't see how it even contributes to the readability.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
morantr: Since it is duplicate/related to comment https://github.com/prplfoundation/prplMesh/pull/1491#discussion_r449626524.
Resolving.

kantera800 (approved):
i would prefer that you didn't use  `tmp_int `  `tmp_str`   
but, besides that its fine (and it works) so approving.

common/beerocks/bwl/dwpal/mon_wlan_hal_dwpal.cpp
arnout: It's more of a personal preference and it's not mentioned in our coding style so feel free to do however you like. My reasoning:
- (bad) it makes the vertical size of the code slightly longer;
- (good) it gives meaning to the variable - not really relevant here because the "SNR" is really close to where it is used, but there is some code below where there's more distance and you do manipulations on a tmp_something that has no clear meaning;
- (good) it makes it trivial to know which value the variable has when looking at part of the code.
- (good) it allows you to reorder the code as you like - currently, because they use the same variable name, you can't keep the calc_curr_traffic calls for PacketsSent and PacketsReceived etc. together because it's always tmp_int;
- (neutral) it makes it slightly easier for the compiler to optimize the code - not really relevant for us because gcc converts everything to full SSA, i.e. every assignment already creates a new variable;
- (good) in the light of the above: it can make debugging easier because there's a bigger chance the debug info will be able to point to that specific variable.

PR #1492, opened 2020-06-29
Feature/ci run boardfarm

rmelotte:
Run boardfarm using docker-only in CI, until the docker-compose integration can be done.


tools/docker/boardfarm-ci/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable"
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+    docker-ce \
+    docker-ce-cli \
+    containerd.io \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN curl -L "https://github.com/docker/compose/releases/download/1.26.0/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+RUN chmod 755 /usr/local/bin/docker-compose
```
arnout, line 37: For my info: why isn't all this simply taken from Debian/Ubuntu repos? Are we using such advanced features?

tools/docker/boardfarm-ci/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    gnupg-agent \
+    libsnmp-dev \
+    software-properties-common \
+    wireshark-common \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY requirements.txt /app/requirements.txt
+
+WORKDIR app
+RUN pip3 install -r requirements.txt
```
arnout, line 21: Why do we need this? Before, we just took the requirements.txt from boardfarm. I.e.
```

RUN git clone https://github.com/mattsm/boardfarm.git \
    && cd boardfarm \
    && git checkout 100521fde1fb67536682cafecc2f91a6e2e8a6f8 \
    && pip3 install -r requirements.txt
    && python3 setup.py install 
```

Yes, it does mean that all the pip stuff has to be re-downloaded whenever we change the boardfarm hash, but that's going to be rare anyway.

tests/run_bf.sh
```diff
@@ -15,4 +15,4 @@ else
 fi
 export PYTHONPATH
 
-exec python3 "${bf_dir}"/bft -c "${bf_plugins_dir}"/boardfarm_prplmesh/prplmesh_config.json -n prplmesh_docker -x test_flows
+exec bft -c "${bf_plugins_dir}"/boardfarm_prplmesh/prplmesh_config.json -n prplmesh_docker -x test_flows
```
arnout, line 18: When Oleksii originally integrated boardfarm, I asked *not* do to this exactly because it requires installing boardfarm, which I wanted to avoid :-)

But I guess it's OK now that we have it in docker.

.gitlab-ci.yml
```diff
@@ -215,6 +215,20 @@ run-tests:
...
+  script:
+    - export USER=gitlab-runner BFT_DEBUG=y
+    - tests/run_bf.sh
+  artifacts:
+    paths:
+      - logs
+  needs:
+    - job: build-in-docker
+  tags:
+    - boardfarm
```
arnout: The commit message says something completely different...

> It runs in a shell runner

No, it doesn't have the shell tag and it specifies an image.

> It also requires docker to be available on the runner, so add the
"docker-build" tag.

No, it doesn't have the docker-build tag.

.gitlab-ci.yml
```diff
@@ -215,6 +215,20 @@ run-tests:
     - job: image-build-tests-runner
     - job: image-build-runner
 
+boardfarm-tests:
+  stage: test
+  image: $CI_REGISTRY_IMAGE/prplmesh-boardfarm-ci:$CI_PIPELINE_ID
+  script:
+    - export USER=gitlab-runner BFT_DEBUG=y
```
arnout: These exports should be explained in a comment or in the commit message.

tests/run_bf.sh
odkq: This change is actually bogus; I added a script run_bf_compose.sh that still uses the boardfarm in it's local copy, as above

tools/docker/boardfarm-ci/Dockerfile
odkq: Boardfarm itself is not installed in this setup, it is used from the rootdir/boardfarm dir directly to allow for any changes made locally to be used.

As we do not expect to change boardfarm itself, this will be changed to installing as before, cloning and the rest.

tools/docker/boardfarm-ci/Dockerfile
rmelotte: Good question, I took the dockerfile from https://github.com/prplfoundation/prplMesh/pull/1483 and tried to make only the changes I needed.
Maybe @odkq needs a more recent version than the base image provides?

.gitlab-ci.yml
rmelotte: I changed it back and forth when I noticed the shell runner had the issue with the user permissions.
It's a docker executor now, I edited the commit message.

.gitlab-ci.yml

tools/docker/boardfarm-ci/Dockerfile
odkq: We could, debian stable is 1.21 and I don't think we are going to use later development, but as documentation is not tagged (always the latest) with this we have the mechanism in place for future upgrades (just change the 1.26.0 in the url and test)

tools/docker/boardfarm-ci/Dockerfile
rmelotte: @arnout good point, I hadn't noticed boardfarm itself also included the requirements, I though it was added by us.

@odkq with these changes, boardfarm _is_ installed in the image itself. Is it something you want to do differently?

rmelotte:
Some of the commits belongs to @odkq , so I still need to add a SOB line for them.

arnout (approved):
A few of my comments haven't been addressed yet, but it's anyway stuff that can be done later, so approving already.

.gitlab-ci.yml
```diff
@@ -212,6 +215,20 @@ run-tests:
...
 
+boardfarm-tests:
+  stage: test
+  image: $CI_REGISTRY_IMAGE/prplmesh-boardfarm-ci:$CI_PIPELINE_ID
+  script:
+    - export USER=''
+    - tests/run_bf.sh
+  artifacts:
+    paths:
+      - logs
```
abelog, line 227: Boardfarm stores it's logs in `results` folder, we should store it also. It have all logs received from consoles of devices during tests and some miscellaneous info.
```suggestion
      - logs
      - results
```

.gitlab-ci.yml
```diff
@@ -212,6 +215,20 @@ run-tests:
     - job: image-build-tests-runner
     - job: image-build-runner
 
+boardfarm-tests:
+  stage: test
+  image: $CI_REGISTRY_IMAGE/prplmesh-boardfarm-ci:$CI_PIPELINE_ID
+  script:
+    - export USER=''
```
abelog, line 223: Why `Null`-fy `$USER` variable?. Internally, in `PrplMeshDocker` it's used to generate `docker network` ID.

abelog (approved):
In general, LGTM.
Small questions left, but nothing big.

aciliketcap (approved):

.gitlab-ci.yml
rmelotte: Currently prplMesh is built as root, and also has to be run as root with and empty USER value (see https://jira.prplfoundation.org/projects/PPM/issues/PPM-208 and https://jira.prplfoundation.org/projects/PPM/issues/PPM-209).
Because of that we also need to run boardfarm with the same (empty) user.

.gitlab-ci.yml
rmelotte: Indeed, thanks for the suggestion I applied it.

PR #1493, opened 2020-06-29
tests: sniffer: Fix problem with multiple tlvs of same type

aciliketcap:
Sniffer only returns one of the tlvs if there are multiple tlvs of the same type in a ieee1905 packet. This is caused by tshark creating an invalid JSON with duplicate keys, using the same key for all the tlvs of same type.

Now subsequent tlvs will have a counter after the key name.

### DIFF
--- a/tests/test_flows.py
+++ b/tests/test_flows.py
@@ -11,6 +11,7 @@ import os
 import sys
 import time
 import traceback
+from collections import Counter
 from typing import Callable, Union, Any, NoReturn
 
 import connmap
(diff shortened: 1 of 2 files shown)


tests/test_flows.py
```diff
@@ -278,6 +279,17 @@ def run_tests(self, tests):
             self.check_error = 0
             try:
                 getattr(self, test_full)()
+            except AssertionError as ae:
+                # do not add empty message if test has already been marked as failed
+                # and AssertionError does not contain a message
+                if str(ae):
+                    self.fail("{}".format(ae))
+                elif not self.check_error:
+                    self.fail("Assertion failed")
```
rmelotte: Personally I would prefer to have a trace in all cases. 
This way if someone forgets to comply to "please add messages to assert statements", we would have the trace anyway. (nitpick)

tests/test_flows.py
```diff
@@ -735,13 +736,44 @@ def test_link_metric_query(self):
...
+
+                try:
+                    # a tshark (v3.2.4) bug causes "neighbor_mac_addr" field to
+                    # show up as "responder_mac_addr_2"
+                    if link.responder_mac_addr_2 in expected_neighbors:
+                        verified_neighbors += [link.responder_mac_addr_2]
+                    else:
+                        unexpected_neighbors += [link.responder_mac_addr_2]
+                except AttributeError as e:
+                    self.fail(e)
```
rmelotte, line 757: Shouldn't we also try `neighbor_mac_addr` if `responder_mac_addr_2` doesn't exist (i.e. if we use a different tshark version)?

rmelotte (approved):
Excellent commit description for "tests: wrap exceptions thrown by individual tests"!
Thanks!

tests/test_flows.py
```diff
@@ -266,6 +270,13 @@ def check_topology_notification(self, eth_src: str, neighbors: list,
 
         return True
 
+    def safe_check_obj_attribute(self, obj, attrib_name, expected_val, fail_str):
```
abelog: *Nitpick*: please, document return type and arguments type

tests/test_flows.py
```diff
@@ -695,17 +717,63 @@ def test_ap_capability_query(self):
...
+                                           mid)
+
+        debug("Checking response contains expected links to/from agent and nothing else")
+        # neighbor macs are based on the setup in "launch_environment_docker" method
+        expected_tx_link_neighbors = [env.controller.mac, env.agents[1].mac]
+        expected_rx_link_neighbors = [env.controller.mac, env.agents[1].mac]
+        actual_tx_links = self.check_cmdu_has_tlvs(resp, 0x09)
+        actual_rx_links = self.check_cmdu_has_tlvs(resp, 0x0a)
+
+        def verify_links(links, expected_neighbors, link_type):
```
abelog, line 742: *Nitpick*: please, document return type and arguments type

tests/test_flows.py
```diff
@@ -695,17 +717,63 @@ def test_ap_capability_query(self):
...
+                        verified_neighbors += [link.responder_mac_addr_2]
+                    else:
+                        unexpected_neighbors += [link.responder_mac_addr_2]
+                except AttributeError as e:
+                    self.fail(e)
+
+            # we expect each neighbor to appear only once
+            for neighbor, count in Counter(verified_neighbors).items():
+                if count != 1:
+                    self.fail("{} link to neighbor {} appears {} times"
```
abelog: Worth to add expected amount of appears in `self.fail` message.
```suggestion
                    self.fail("{} link to neighbor {} appears {} times. We expect it to appear only once."
```

tests/test_flows.py
```diff
@@ -810,58 +878,83 @@ def test_combined_infra_metrics(self):
...
 
-        env.agents[0].radios[0].vaps[0].disassociate(sta2)
+            cap_report_tlv = self.check_cmdu_has_tlvs(report, 0x91)[0]
+            self.safe_check_obj_attribute(cap_report_tlv, 'client_capability_result', '0x00000000',
+                                          "Capability report result is not successful")
+            try:
+                if cap_report_tlv.client_capability_frame != expected_association_frame:
+                    self.fail("Capability report does not contain correct frame")
+                    debug(f"Frame received\n{cap_report_tlv.client_capability_frame}")
+                    debug(f"Frame expected\n{expected_association_frame}")
```
abelog: Should `debug` messages be before `self.fail`?

abelog (approved):
LGTM. Only nitpicks.

tests/test_flows.py
aciliketcap: I added log to only the case where generic "Assertion failed" is printed. Other messages can be easily searched and found in the test code. If we print stack traces everywhere than all the failed tests will have stack traces, I am not sure everybody would agree on that.

tests/test_flows.py
aciliketcap: I did it in other branch so resolving this one.

tests/test_flows.py
aciliketcap: I read the wireshark code and I suspect all Wireshark versions have the same problem. Also the other variables are always filled with zeros. I'll double check with EasyMesh spec but I decided not to use them until I did so.

tests/test_flows.py
aciliketcap: It is a method defined inside the test just to prevent code duplication so I am keeping it as is since it feels cleaner to me.

tests/test_flows.py
aciliketcap: I wanted to but I was not sure how to say it :+1:

tests/test_flows.py
aciliketcap: I wanted them to appear after the fail message. But I would move them above if the test had to exit on fail.

PR #1494, opened 2020-06-29
PPM-5 add smart steering persistent data to db

adam1985d:
This PR introduces client parameters to enable smart-steering capabilities and persistency of those parameters across prplmesh restarts and system reboots.

The BPL provides the interface to the persistent DB and the 

The added configuration is saved to the persistent DB via BPL and enables a per-client steering configuration persistent across prplmesh-restarts/reboots.

This PR includes:
Add the client parameters to the Node class (runtime controller DB).
Add setters/getters APIs to the client parameters (get is from runtime DB, set can be to runtime DB or also to persistent DB (write-through))
Add APIs to clear/update client persistent information.
Add API to get all clients from persistent DB

### DIFF
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -2419,6 +2419,85 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
 }
 
 //
+// Client Persistent Data
+//
+bool db::is_client_in_persistent_db(const sMacAddr &mac) { return false; }
+
+bool db::add_client_to_persistent_db(const sMacAddr &mac,
+                                     std::unordered_map<std::string, std::string> params)
+{
+    return true;
+}
+
+std::chrono::steady_clock::time_point db::get_client_parameters_last_edit(const sMacAddr &mac)
+{
+    return std::chrono::steady_clock::time_point::min();
+}
+
+bool db::set_client_time_life_delay(const sMacAddr &mac,
+                                    const std::chrono::seconds &time_life_delay_sec,
+                                    bool save_to_persistent_db)
+{
+    return true;
+}
+
+std::chrono::seconds db::get_client_time_life_delay(const sMacAddr &mac)
+{
+    return std::chrono::seconds::zero();
+}
+
+bool db::set_client_stay_on_initial_radio(const sMacAddr &mac, bool stay_on_initial_radio,
+                                          bool save_to_persistent_db)
+{
+    return true;
+}
+
+ePersistentParamBool db::get_client_stay_on_initial_radio(const sMacAddr &mac)
+{
+    return ePersistentParamBool::NOT_CONFIGURED;
+}
+
+bool db::set_client_initial_radio(const sMacAddr &mac, const sMacAddr &initial_radio_mac,
+                                  bool save_to_persistent_db)
+{
+    return true;
+}
+
+sMacAddr db::get_client_initial_radio(const sMacAddr &mac) { return network_utils::ZERO_MAC; }
+
+bool db::set_client_stay_on_selected_band(const sMacAddr &mac, bool stay_on_selected_band,
+                                          bool save_to_persistent_db)
+{
+    return true;
+}
+
+ePersistentParamBool db::get_client_stay_on_selected_band(const sMacAddr &mac)
+{
+    return ePersistentParamBool::NOT_CONFIGURED;
+}
+
+bool db::set_client_selected_bands(const sMacAddr &mac, beerocks::eFreqType selected_bands,
+                                   bool save_to_persistent_db)
+{
+    return true;
+}
+
+beerocks::eFreqType db::get_client_selected_bands(const sMacAddr &mac)
+{
+    return beerocks::eFreqType::FREQ_UNKNOWN;
+}
+
+bool db::clear_client_persistent_db(const sMacAddr &mac) { return true; }
+
+bool db::update_client_persistent_db(const sMacAddr &mac) { return true; }
+
+std::unordered_map<std::string, std::unordered_map<std::string, std::string>>
+db::load_persistent_db_clients()
+{
+    return {};
+}
+
+//
 // CLI
 //
 void db::add_cli_socket(Socket *sd)
@@ -3491,6 +3570,21 @@ std::shared_ptr<node> db::get_node(sMacAddr al_mac, sMacAddr ruid)
         key = tlvf::mac_to_string(al_mac) + tlvf::mac_to_string(ruid);
 
     return get_node(key);
+}
+
+std::shared_ptr<node> db::get_node_verify_type(const sMacAddr &mac, beerocks::eType type)
+{
+    auto node = get_node(mac);
+    if (!node) {
+        LOG(ERROR) << "node not found for mac " << mac;
+        return nullptr;
+    } else if (node->get_type() != type) {
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " type(" << node->get_type()
+                   << ") != requested-type(" << type << ")";
+        return nullptr;
+    }
+
+    return node;
 }
 
 std::shared_ptr<node::radio> db::get_hostap_by_mac(const sMacAddr &mac)
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -20,6 +20,13 @@ using namespace beerocks_message;
 using namespace son;
 using namespace net;
 
+static const std::string timestamp_str            = "timestamp";
+static const std::string timelife_delay_str       = "timelife";
+static const std::string initial_radio_enable_str = "initial_radio_enable";
+static const std::string initial_radio_str        = "initial_radio";
+static const std::string selected_band_enable_str = "selected_band_enable";
+static const std::string selected_bands_str       = "selected_bands";
+
 // static
 std::string db::type_to_string(beerocks::eType type)
 {
@@ -2475,11 +2482,38 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
 //
 // Client Persistent Data
 //
-bool db::is_client_in_persistent_db(const sMacAddr &mac) { return false; }
+bool db::is_client_in_persistent_db(const sMacAddr &mac)
+{
+    auto client_db_entry = client_db_entry_from_mac(mac);
+
+    return bpl::db_has_entry(type_to_string(beerocks::eType::TYPE_CLIENT), client_db_entry);
+}
 
 bool db::add_client_to_persistent_db(const sMacAddr &mac,
                                      std::unordered_map<std::string, std::string> params)
 {
+    auto db_entry = client_db_entry_from_mac(mac);
+
+    if (!bpl::db_has_entry(std::string("type"), db_entry)) {
+        // if entry already exists in DB
+        if (!bpl::db_remove_entry(std::string("type"), db_entry)) {
+            LOG(ERROR) << "failed to remove client entry " << db_entry
+                       << "from persistent db (for re-adding)";
+            return false;
+        }
+    } else if (!bpl::db_has_entry(std::string(), db_entry)) {
+        // if entry exists in db but with different type
+        LOG(ERROR) << "client entry cannot be added to persistent db, " << db_entry
+                   << " already exists but with different type";
+        return false;
+    }
+    // add entry to the persistent db
+    if (!bpl::db_add_entry(type_to_string(beerocks::eType::TYPE_CLIENT), db_entry, params)) {
+        LOG(ERROR) << "failed to add client entry " << db_entry << " to persistent db";
+        return false;
+    }
+    LOG(DEBUG) << "added client entry " << db_entry << " to persistent db";
+
     return true;
 }
 
@@ -2509,6 +2543,16 @@ bool db::set_client_time_life_delay(const sMacAddr &mac,
     auto timestamp = std::chrono::steady_clock::now();
     if (save_to_persistent_db) {
         LOG(DEBUG) << "configuring persistent-db, timelife = " << time_life_delay_sec.count();
+
+        std::unordered_map<std::string, std::string> values_map;
+        values_map.at(timestamp_str)      = timestamp_to_string_seconds(timestamp);
+        values_map.at(timelife_delay_str) = std::to_string(time_life_delay_sec.count());
+
+        // update the persistent db
+        if (!update_client_entry_in_persistent_db(mac, values_map)) {
+            LOG(ERROR) << "failed to update client entry in persistent-db to for " << mac;
+            return false;
+        }
     }
 
     node->client_time_life_delay_sec  = time_life_delay_sec;
@@ -2542,6 +2586,16 @@ bool db::set_client_stay_on_initial_radio(const sMacAddr &mac, bool stay_on_init
     auto timestamp = std::chrono::steady_clock::now();
     if (save_to_persistent_db) {
         LOG(DEBUG) << "configuring persistent-db, initial_radio_enable = " << stay_on_initial_radio;
+
+        std::unordered_map<std::string, std::string> values_map;
+        values_map.at(timestamp_str)            = timestamp_to_string_seconds(timestamp);
+        values_map.at(initial_radio_enable_str) = std::to_string(stay_on_initial_radio);
+
+        // update the persistent db
+        if (!update_client_entry_in_persistent_db(mac, values_map)) {
+            LOG(ERROR) << "failed to update client entry in persistent-db to for " << mac;
+            return false;
+        }
     }
 
     node->client_stay_on_initial_radio =
@@ -2576,6 +2630,16 @@ bool db::set_client_initial_radio(const sMacAddr &mac, const sMacAddr &initial_r
     auto timestamp = std::chrono::steady_clock::now();
     if (save_to_persistent_db) {
         LOG(DEBUG) << "configuring persistent-db, initial_radio = " << initial_radio_mac;
+
+        std::unordered_map<std::string, std::string> values_map;
+        values_map.at(timestamp_str)     = timestamp_to_string_seconds(timestamp);
+        values_map.at(initial_radio_str) = tlvf::mac_to_string(initial_radio_mac);
+
+        // update the persistent db
+        if (!update_client_entry_in_persistent_db(mac, values_map)) {
+            LOG(ERROR) << "failed to update client entry in persistent-db to for " << mac;
+            return false;
+        }
     }
 
     node->client_initial_radio        = initial_radio_mac;
@@ -2609,6 +2673,16 @@ bool db::set_client_stay_on_selected_band(const sMacAddr &mac, bool stay_on_sele
     auto timestamp = std::chrono::steady_clock::now();
     if (save_to_persistent_db) {
         LOG(DEBUG) << "configuring persistent-db, selected_band_enable = " << stay_on_selected_band;
+
+        std::unordered_map<std::string, std::string> values_map;
+        values_map.at(timestamp_str)            = timestamp_to_string_seconds(timestamp);
+        values_map.at(selected_band_enable_str) = std::to_string(stay_on_selected_band);
+
+        // update the persistent db
+        if (!update_client_entry_in_persistent_db(mac, values_map)) {
+            LOG(ERROR) << "failed to update client entry in persistent-db to for " << mac;
+            return false;
+        }
     }
 
     node->client_stay_on_selected_band =
@@ -2643,6 +2717,16 @@ bool db::set_client_selected_bands(const sMacAddr &mac, beerocks::eFreqType sele
     auto timestamp = std::chrono::steady_clock::now();
     if (save_to_persistent_db) {
         LOG(DEBUG) << ", configuring persistent-db, selected_bands = " << int(selected_bands);
+
+        std::unordered_map<std::string, std::string> values_map;
+        values_map.at(timestamp_str)      = timestamp_to_string_seconds(timestamp);
+        values_map.at(selected_bands_str) = std::to_string(selected_bands);
+
+        // update the persistent db
+        if (!update_client_entry_in_persistent_db(mac, values_map)) {
+            LOG(ERROR) << "failed to update client entry in persistent-db to for " << mac;
+            return false;
+        }
     }
 
     node->client_selected_bands       = selected_bands;
@@ -2680,16 +2764,101 @@ bool db::clear_client_persistent_db(const sMacAddr &mac)
     node->client_selected_bands        = beerocks::eFreqType::FREQ_UNKNOWN;
 
     LOG(DEBUG) << "removing client " << mac << " from persistent db";
+    auto db_entry        = client_db_entry_from_mac(mac);
+    auto type_client_str = type_to_string(beerocks::eType::TYPE_CLIENT);
+    if (!bpl::db_has_entry(type_client_str, db_entry)) {
+        LOG(DEBUG) << "client entry does not exist in persistent-db for " << db_entry;
+        return true;
+    }
+
+    if (!bpl::db_remove_entry(type_client_str, db_entry)) {
+        LOG(ERROR) << "failed to remove client entry " << db_entry;
+        return false;
+    }
 
     return true;
 }
 
-bool db::update_client_persistent_db(const sMacAddr &mac) { return true; }
+bool db::update_client_persistent_db(const sMacAddr &mac)
+{
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return false;
+    }
+
+    // any persistent parameter update also sets the last-edit timestamp
+    // if it is with default value - no other persistent configuration was performed
+    if (node->client_parameters_last_edit == std::chrono::steady_clock::time_point::min()) {
+        LOG(DEBUG) << "persistent client parameters are empty for " << mac
+                   << ", no need to update persistent-db";
+        return true;
+    }
+
+    std::unordered_map<std::string, std::string> values_map;
+
+    //fill values map of client persistent params
+    values_map.at(timestamp_str) = timestamp_to_string_seconds(node->client_parameters_last_edit);
+
+    if (node->client_time_life_delay_sec != std::chrono::seconds::zero()) {
+        LOG(DEBUG) << "setting client time-life-delay in persistent-db to for " << mac << " to "
+                   << node->client_time_life_delay_sec.count();
+        values_map.at(timelife_delay_str) =
+            std::to_string(node->client_time_life_delay_sec.count());
+    }
+
+    if (node->client_stay_on_initial_radio != ePersistentParamBool::NOT_CONFIGURED) {
+        auto enable = (node->client_stay_on_initial_radio == ePersistentParamBool::ENABLE);
+        LOG(DEBUG) << "setting client stay-on-initial-radio in persistent-db to for " << mac
+                   << " to " << enable;
+        values_map.at(initial_radio_enable_str) = std::to_string(enable);
+    }
+
+    if (node->client_initial_radio != network_utils::ZERO_MAC) {
+        LOG(DEBUG) << "setting client initial-radio in persistent-db to for " << mac << " to "
+                   << node->client_initial_radio;
+        values_map.at(initial_radio_str) = tlvf::mac_to_string(node->client_initial_radio);
+    }
+
+    if (node->client_stay_on_selected_band != ePersistentParamBool::NOT_CONFIGURED) {
+        auto enable = (node->client_stay_on_selected_band == ePersistentParamBool::ENABLE);
+        LOG(DEBUG) << "setting client stay-on-selected-band in persistent-db to for " << mac
+                   << " to " << enable;
+        values_map.at(selected_band_enable_str) = std::to_string(enable);
+    }
+
+    if (node->client_selected_bands != beerocks::eFreqType::FREQ_UNKNOWN) {
+        LOG(DEBUG) << "setting client selected-bands in persistent-db to for " << mac << " to "
+                   << node->client_selected_bands;
+        values_map.at(selected_bands_str) = std::to_string(node->client_selected_bands);
+    }
+
+    // update the persistent db
+    if (!update_client_entry_in_persistent_db(mac, values_map)) {
+        LOG(ERROR) << "failed to update client entry in persistent-db to for " << mac;
+        return false;
+    }
+
+    LOG(DEBUG) << "client successfully updated in persistent-db for " << mac;
+
+    return true;
+}
 
 std::unordered_map<std::string, std::unordered_map<std::string, std::string>>
 db::load_persistent_db_clients()
 {
-    return {};
+    std::unordered_map<std::string, std::unordered_map<std::string, std::string>> clients;
+    if (!bpl::db_get_entries_by_type(type_to_string(beerocks::eType::TYPE_CLIENT), clients)) {
+        LOG(ERROR) << "failed to get all clients from persistent DB";
+        return {};
+    }
+
+    if (clients.empty()) {
+        LOG(DEBUG) << "persistent DB doesn't exist (or empty) or doesn't contain clients";
+        return {};
+    }
+
+    return clients;
 }
 
 //
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -2431,63 +2431,204 @@ bool db::add_client_to_persistent_db(const sMacAddr &mac,
 
 std::chrono::steady_clock::time_point db::get_client_parameters_last_edit(const sMacAddr &mac)
 {
-    return std::chrono::steady_clock::time_point::min();
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return std::chrono::steady_clock::time_point::min();
+    }
+
+    return node->client_parameters_last_edit;
 }
 
 bool db::set_client_time_life_delay(const sMacAddr &mac,
                                     const std::chrono::seconds &time_life_delay_sec,
                                     bool save_to_persistent_db)
 {
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return false;
+    }
+
+    LOG(DEBUG) << "time_life_delay_sec = " << time_life_delay_sec.count();
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
+        LOG(DEBUG) << "configuring persistent-db, timelife = " << time_life_delay_sec.count();
+    }
+
+    node->client_time_life_delay_sec  = time_life_delay_sec;
+    node->client_parameters_last_edit = timestamp;
+
     return true;
 }
 
 std::chrono::seconds db::get_client_time_life_delay(const sMacAddr &mac)
 {
-    return std::chrono::seconds::zero();
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return std::chrono::seconds::zero();
+    }
+
+    return node->client_time_life_delay_sec;
 }
 
 bool db::set_client_stay_on_initial_radio(const sMacAddr &mac, bool stay_on_initial_radio,
                                           bool save_to_persistent_db)
 {
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return false;
+    }
+
+    LOG(DEBUG) << "stay_on_initial_radio = " << stay_on_initial_radio;
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
+        LOG(DEBUG) << "configuring persistent-db, initial_radio_enable = " << stay_on_initial_radio;
+    }
+
+    node->client_stay_on_initial_radio =
+        (stay_on_initial_radio) ? ePersistentParamBool::ENABLE : ePersistentParamBool::DISABLE;
+    node->client_parameters_last_edit = timestamp;
+
     return true;
 }
 
 ePersistentParamBool db::get_client_stay_on_initial_radio(const sMacAddr &mac)
 {
-    return ePersistentParamBool::NOT_CONFIGURED;
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return ePersistentParamBool::NOT_CONFIGURED;
+    }
+
+    return node->client_stay_on_initial_radio;
 }
 
 bool db::set_client_initial_radio(const sMacAddr &mac, const sMacAddr &initial_radio_mac,
                                   bool save_to_persistent_db)
 {
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return false;
+    }
+
+    LOG(DEBUG) << "initial_radio = " << initial_radio_mac;
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
+        LOG(DEBUG) << "configuring persistent-db, initial_radio = " << initial_radio_mac;
+    }
+
+    node->client_initial_radio        = initial_radio_mac;
+    node->client_parameters_last_edit = timestamp;
+
     return true;
 }
 
-sMacAddr db::get_client_initial_radio(const sMacAddr &mac) { return network_utils::ZERO_MAC; }
+sMacAddr db::get_client_initial_radio(const sMacAddr &mac)
+{
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return network_utils::ZERO_MAC;
+    }
+
+    return node->client_initial_radio;
+}
 
 bool db::set_client_stay_on_selected_band(const sMacAddr &mac, bool stay_on_selected_band,
                                           bool save_to_persistent_db)
 {
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return false;
+    }
+
+    LOG(DEBUG) << "stay_on_selected_band = " << stay_on_selected_band;
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
+        LOG(DEBUG) << "configuring persistent-db, selected_band_enable = " << stay_on_selected_band;
+    }
+
+    node->client_stay_on_selected_band =
+        (stay_on_selected_band) ? ePersistentParamBool::ENABLE : ePersistentParamBool::DISABLE;
+    node->client_parameters_last_edit = timestamp;
+
     return true;
 }
 
 ePersistentParamBool db::get_client_stay_on_selected_band(const sMacAddr &mac)
 {
-    return ePersistentParamBool::NOT_CONFIGURED;
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return ePersistentParamBool::NOT_CONFIGURED;
+    }
+
+    return node->client_stay_on_selected_band;
 }
 
 bool db::set_client_selected_bands(const sMacAddr &mac, beerocks::eFreqType selected_bands,
                                    bool save_to_persistent_db)
 {
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return false;
+    }
+
+    LOG(DEBUG) << "selected_band = " << int(selected_bands);
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
+        LOG(DEBUG) << ", configuring persistent-db, selected_bands = " << int(selected_bands);
+    }
+
+    node->client_selected_bands       = selected_bands;
+    node->client_parameters_last_edit = timestamp;
+
     return true;
 }
 
 beerocks::eFreqType db::get_client_selected_bands(const sMacAddr &mac)
 {
-    return beerocks::eFreqType::FREQ_UNKNOWN;
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return beerocks::eFreqType::FREQ_UNKNOWN;
+    }
+
+    return node->client_selected_bands;
 }
 
-bool db::clear_client_persistent_db(const sMacAddr &mac) { return true; }
+bool db::clear_client_persistent_db(const sMacAddr &mac)
+{
+    auto node = get_node_verify_type(mac, beerocks::TYPE_CLIENT);
+    if (!node) {
+        LOG(ERROR) << "client node not found for mac " << mac;
+        return false;
+    }
+
+    LOG(DEBUG) << "setting client " << mac << " runtime info to default values";
+
+    node->client_parameters_last_edit  = std::chrono::steady_clock::time_point::min();
+    node->client_time_life_delay_sec   = std::chrono::seconds::zero();
+    node->client_stay_on_initial_radio = ePersistentParamBool::NOT_CONFIGURED;
+    node->client_initial_radio         = network_utils::ZERO_MAC;
+    node->client_stay_on_selected_band = ePersistentParamBool::NOT_CONFIGURED;
+    node->client_selected_bands        = beerocks::eFreqType::FREQ_UNKNOWN;
+
+    LOG(DEBUG) << "removing client " << mac << " from persistent db";
+
+    return true;
+}
 
 bool db::update_client_persistent_db(const sMacAddr &mac) { return true; }
 
--- a/controller/src/beerocks/master/db/db.cpp
+++ b/controller/src/beerocks/master/db/db.cpp
@@ -10,6 +10,7 @@
 
 #include <bcl/beerocks_utils.h>
 #include <bcl/son/son_wireless_utils.h>
+#include <bpl/bpl_db.h>
 #include <easylogging++.h>
 
 #include <algorithm>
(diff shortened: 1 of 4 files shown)


common/beerocks/bcl/include/bcl/beerocks_defines.h
```diff
@@ -470,6 +470,8 @@ enum class eChannelScanOperationCode : uint8_t {
     INVALID_PARAMS_CHANNELPOOL
 };
 
+enum class ePersistentParamBool : int8_t { DISABLE = 0, ENABLE = 1, NOT_CONFIGURED = -1 };
```
vitalybu: Please add a doxygen comment explaining this type.
I don't understand its' purpose form the name. Bool suggests a TRUE/FALSE or ENABLED/DISABLED values, but you actually have 3 values.

```suggestion
enum class ePersistentParamBool : int8_t { NOT_CONFIGURED = -1, DISABLED = 0, ENABLED = 1 };
```

controller/src/beerocks/master/db/node.cpp
```diff
@@ -121,6 +121,25 @@ std::ostream &operator<<(std::ostream &os, const node &n)
                << "   Failed5ghzSteerAttemps: " << int(n.failed_5ghz_steer_attemps) << std::endl
                << "   Failed24ghzSteerAttemps: " << int(n.failed_24ghz_steer_attemps) << std::endl;
         }
+
+        // persistent db
+        if (node_type == beerocks::TYPE_CLIENT) {
+            os << "Persistent configuration and data:" << std::endl
+               << "   ClientPersistentDataLastEdit: "
+               << float(std::chrono::duration_cast<std::chrono::duration<double>>(
```
vitalybu: Why it this a double casted to float? In what units is the duration?

controller/src/beerocks/master/db/node.cpp
```diff
@@ -121,6 +121,25 @@ std::ostream &operator<<(std::ostream &os, const node &n)
...
+                            tCurrTime - n.client_persistent_data_last_edit)
+                            .count())
+               << std::endl
+               << "   ClientTimeLifeDelay: "
+               << int(n.client_time_life_delay_sec.count() / (60 * 60 * 24)) << " days" << std::endl
```
vitalybu:
```suggestion
               << int(std::chrono::duration_cast<std::chrono::days>(n.client_time_life_delay_sec).count()) << " days" << std::endl
```

controller/src/beerocks/master/db/node.h
```diff
@@ -267,6 +267,33 @@ class node {
...
+     * persistent configurations - start
+     * Client persistent configuration aging is refreshed on persistent configurations set
+     * persistent configuration of aged clients removed from the persistent-db and cleared in the runtime-db
+     */
+
+    // indicates when client config was last updated, minimal value is used as invalid value
+    std::chrono::steady_clock::time_point client_persistent_data_last_edit =
+        std::chrono::steady_clock::time_point::min();
+    // optional - if configured the client has its own configured timelife delay
+    std::chrono::seconds client_time_life_delay_sec = std::chrono::seconds(0);
```
vitalybu:
```suggestion
    std::chrono::seconds client_time_life_delay_sec = std::chrono::seconds::zero();
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2418,80 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+{
+    return true;
+}
+
+beerocks::ePersistentParamBool db::get_client_stay_on_initial_radio(const sMacAddr &mac)
+{
+    return beerocks::ePersistentParamBool::NOT_CONFIGURED;
+}
+
+bool db::set_client_initial_radio(const sMacAddr &mac, const sMacAddr &initial_radio_mac,
```
vitalybu, line 2619: Why do we need `set_client_stay_on_initial_radio` and `set_client_initial_radio`?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2418,80 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+// Client Persistent Data
+//
+std::chrono::steady_clock::time_point db::get_client_persistent_data_last_edit(const sMacAddr &mac)
+{
+    return std::chrono::steady_clock::time_point::min();
+}
+
+bool db::set_client_time_life_delay(const sMacAddr &mac,
+                                    const std::chrono::seconds &time_life_delay_sec,
+                                    bool set_persistent_db)
```
vitalybu: What is `set_persistent_db` for?

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * @brief Get the client's last persistent data edit time.
+     * 
+     * @param mac MAC address of client.
+     * 
+     * @return client persistent data last edit time, or time_point::min() if not-configured or failure.
+     */
+    std::chrono::steady_clock::time_point get_client_persistent_data_last_edit(const sMacAddr &mac);
+
+    /**
+     * @brief Set the client's time-life delay.
```
vitalybu, line 669: What is "time-life"?

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * @brief Get the client's time-life delay.
+     * 
+     * @param mac MAC address of client.
+     * 
+     * @return client time-life delay, value of 0 means not-configured.
+     */
+    std::chrono::seconds get_client_time_life_delay(const sMacAddr &mac);
+
+    /**
+     * @brief Set the client's stay-on-initial-radio.
```
vitalybu, line 689: What is "stay-on-initial-radio"?
I think you need to add a short explanation in the comment about these features..

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * 
+     * @param mac MAC address of client.
+     * @param stay_on_initial_radio enable client stay on the radio it initially connected to.
+     * @param set_persistent_db set-persistent-db flag - default is true (write-through).
+     * 
+     * @return true on success.
+     * @return false on failure.
+     */
+    bool set_client_stay_on_initial_radio(const sMacAddr &mac, bool stay_on_initial_radio,
```
vitalybu, line 696:
```suggestion
    bool set_force_client_to_initial_radio(const sMacAddr &mac, bool force_to_initial_radio,
```

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * @brief Set the client's initial-radio.
+     * 
+     * @param mac MAC address of client.
+     * @param initial_radio_mac the mac of radio that the client has initially connected to.
+     * @param set_persistent_db set-persistent-db flag - default is true (write-through).
+     * 
+     * @return true on success.
+     * @return false on failure.
+     */
+    bool set_client_initial_radio(const sMacAddr &mac, const sMacAddr &initial_radio_mac,
```
vitalybu, line 715: Can you really set the radio before the client connected?
Or is the purpose of this method is to update the MAC of the radio the client connected to?

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * @return enable client stay on the radio it initially connected to.
+     */
+    beerocks::ePersistentParamBool get_client_stay_on_initial_radio(const sMacAddr &mac);
+
+    /**
+     * @brief Set the client's initial-radio.
+     * 
+     * @param mac MAC address of client.
+     * @param initial_radio_mac the mac of radio that the client has initially connected to.
+     * @param set_persistent_db set-persistent-db flag - default is true (write-through).
```
vitalybu: You can't describe a variable with the variable's name but changing underscores to hypens :)

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+    bool update_client_persistent_db(const sMacAddr &mac);
+
+    /**
+     * @brief Get all clients from persistent db
+     * 
+     * @return unordered map of clients, for each client unordered map of params as key-value
+     * @return an empty map of clients is returned if the persistent db is empty
+     */
+    std::unordered_map<std::string, std::unordered_map<std::string, std::string>>
+    get_all_persistent_db_clients();
```
vitalybu:
```suggestion
    load_persistent_db();
```
?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2423,75 +2423,252 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    } else if (n->get_type() != beerocks::TYPE_CLIENT) {
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid client!";
+        return false;
+    }
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (set_persistent_db) {
+        LOG(DEBUG) << __FUNCTION__ << ", configuring persistent-db, stay_on_initial_radio = "
+                   << string_utils::bool_str(stay_on_initial_radio);
```
vitalybu:
```suggestion
                   << std::boolalpha << stay_on_initial_radio;
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2423,75 +2423,252 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+    auto initial_radio_str = tlvf::mac_to_string(initial_radio_mac);
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (set_persistent_db) {
+        LOG(DEBUG) << __FUNCTION__
+                   << ", configuring persistent-db, initial_radio = " << initial_radio_str;
+    }
+
+    LOG(DEBUG) << __FUNCTION__
+               << ", stay_on_initial_radio = " << tlvf::mac_to_string(initial_radio_mac);
```
vitalybu:
```suggestion
               << ", stay_on_initial_radio = " << initial_radio_str;
```

common/beerocks/bcl/include/bcl/beerocks_defines.h
morantr: It is not a bool, it is tri-state configuration.
So I would call it eTriStateConf or eConfEnabled.

morantr: I don't think it should be defined here. It is used only by the controller, its database to be accurate. 
I think it should be defined there, and not on beerocks_defines.h which almost all files on the project includes.

adam1985d: added Doxygen
reordered the values as suggested

controller/src/beerocks/master/db/node.cpp
adam1985d: `std::chrono::days` isn't available - but will use the `std::chrono::hours` and `/24` only

controller/src/beerocks/master/db/node.h
adam1985d: replaced all locations of `std::chrono::seconds(0)` to use `std::chrono::seconds::zero()`

controller/src/beerocks/master/db/db.cpp
adam1985d: the client can configure a client to stay on its initial radio even before the client was connected for the first time, so the information should be able to be configured separately.
we might also want to save in the future the `client_initial_radio` even if the `stay_on_initial_radio` is false

controller/src/beerocks/master/db/db.h
adam1985d: this API is an interface for the controller/tasks.
the `set_stay_on_initial_radio` on the other hand can be triggered by the son_management as a response to bml-request

controller/src/beerocks/master/db/db.h
adam1985d: true.
changed to `* @param set_persistent_db if set to true, update the persistent-db (write-through), default is true.`

controller/src/beerocks/master/db/db.h
adam1985d: `load_persistent_db_clients`
?

controller/src/beerocks/master/db/db.cpp
adam1985d: wow,
we have so many places in code using the `string_utils::bool_str`
👍

controller/src/beerocks/master/db/db.cpp
adam1985d: fixed

controller/src/beerocks/master/db/node.cpp
adam1985d: I changed the way I print this information - please re-review it and let me know what you think

controller/src/beerocks/master/db/node.cpp
```diff
@@ -121,6 +121,29 @@ std::ostream &operator<<(std::ostream &os, const node &n)
                << "   Failed5ghzSteerAttemps: " << int(n.failed_5ghz_steer_attemps) << std::endl
                << "   Failed24ghzSteerAttemps: " << int(n.failed_24ghz_steer_attemps) << std::endl;
         }
+
+        // persistent db
+        if (node_type == beerocks::TYPE_CLIENT) {
+            std::chrono::hours client_persistent_data_last_edit_hours =
+                std::chrono::duration_cast<std::chrono::hours>(tCurrTime -
+                                                               n.client_persistent_data_last_edit);
+            std::chrono::hours client_time_life_delay_hours =
```
morantr: Since this integers represent the time on hours, I would call the `count()` directly from here, and also use `auto` decleration:
```cpp
auto client_persistent_data_last_edit_hours = std::chrono::duration_cast<std::chrono::hours>(tCurrTime -n.client_persistent_data_last_edit).count();

auto client_time_life_delay_hours =  std::chrono::duration_cast<std::chrono::hours>(n.client_time_life_delay_sec).count();

```

controller/src/beerocks/master/db/node.cpp
```diff
@@ -121,6 +121,29 @@ std::ostream &operator<<(std::ostream &os, const node &n)
...
+                std::chrono::duration_cast<std::chrono::hours>(n.client_time_life_delay_sec);
+            os << "Persistent configuration and data:" << std::endl
+               << "   ClientPersistentDataLastEdit: "
+               << int(client_persistent_data_last_edit_hours.count() / 24) << " days, "
+               << int(client_persistent_data_last_edit_hours.count() % 24) << " hours" << std::endl
+               << "   ClientTimeLifeDelay: " << int(client_time_life_delay_hours.count() / 24)
+               << " days, " << int(client_time_life_delay_hours.count() % 24) << " hours"
```
morantr: The `count()` function should return printable return value, so no need to cast it to int.

controller/src/beerocks/master/db/node.cpp
```diff
@@ -121,6 +121,29 @@ std::ostream &operator<<(std::ostream &os, const node &n)
...
+               << std::endl
+               << "   ClientStayOnInitialRadio: " << int(n.client_stay_on_initial_radio)
+               << std::endl
+               << "   ClientInitialRadio: " << n.client_initial_radio << std::endl
+               << "   ClientStayOnSelectedBand: " << int(n.client_stay_on_selected_band)
+               << std::endl
+               << "   ClientSelectedBands: " << int(n.client_selected_bands) << std::endl;
```
morantr: Suggestion:
Add an operator overload for `<<` so when using it on stream, it will print "Enabled", "Disabled" or "Not Configured".
(nitpick)

controller/src/beerocks/master/db/node.h
```diff
@@ -267,6 +267,33 @@ class node {
...
+    beerocks::ePersistentParamBool client_stay_on_selected_band =
+        beerocks::ePersistentParamBool::NOT_CONFIGURED;
+    // contains the selected bands for the client
+    // if the client_stay_on_selected_band is enabled, the client_selected_bands defines which bands
+    // the client should be steered to (if set to FREQ_UNKNOWN - value is not configured)
+    beerocks::eFreqType client_selected_bands = beerocks::eFreqType::FREQ_UNKNOWN;
+
+    /*
+     * persistent configurations - end
+     */
```
morantr: It would look better if each comment would start with a capital letter and end with a dot.
Also, a space between declaration to next declaration comment would look nicer IMO:
```cpp
Foo foo;

// Bar comment.
Bar bar;
```
(nitpick)

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
     const std::list<sChannelScanResults> &get_channel_scan_results(const sMacAddr &mac,
                                                                    bool single_scan);
 
+    //
+    // Client Persistent Data
+    //
+    /**
+     * @brief Get the client's last persistent data edit time.
+     * 
+     * @param mac MAC address of client.
```
morantr: "... of **a** client".
This needs to be fixed on several places on this file.

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * 
+     * @return client persistent data last edit time, or time_point::min() if not-configured or failure.
+     */
+    std::chrono::steady_clock::time_point get_client_persistent_data_last_edit(const sMacAddr &mac);
+
+    /**
+     * @brief Set the client's time-life delay.
+     * 
+     * @param mac MAC address of client.
+     * @param time_life_delay_sec client-specific aging time.
```
morantr: Please make sure to start the argument description with a capital letter.

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * @param set_persistent_db if set to true, update the persistent-db (write-through), default is true.
+     * 
+     * @return true on success.
+     * @return false on failure.
```
morantr: @return true on success, otherwise false.

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     */
+    std::chrono::steady_clock::time_point get_client_persistent_data_last_edit(const sMacAddr &mac);
+
+    /**
+     * @brief Set the client's time-life delay.
+     * 
+     * @param mac MAC address of client.
+     * @param time_life_delay_sec client-specific aging time.
+     * @param set_persistent_db if set to true, update the persistent-db (write-through), default is true.
+     *
```
morantr: No need to add a space between @param and @return. Only after the @brief.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2423,73 +2423,257 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+    LOG(DEBUG) << __FUNCTION__ << ", time_life_delay_sec = " << time_life_delay_sec.count();
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (set_persistent_db) {
+        LOG(DEBUG) << __FUNCTION__ << ", configuring persistent-db, time_life_delay_sec = "
+                   << time_life_delay_sec.count();
+    }
+
+    n->client_time_life_delay_sec       = time_life_delay_sec;
+    n->client_persistent_data_last_edit = timestamp;
```
simonbaren: AFAIU - the naming here is misleading `client_persistent_data_last_edit` - in the case that `set_persistent_db` is set to false you are not actually updating the persistent DB, you only update the controller runtime DB.

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+     * @return enable client stay on the selected band/bands.
+     */
+    beerocks::ePersistentParamBool get_client_stay_on_selected_band(const sMacAddr &mac);
+
+    /**
+     * @brief Set the client's selected-bands.
+     * 
+     * @param mac MAC address of client.
+     * @param selected_bands: client selected band/bands. FREQ_UNKNOWN is considered as "not-configured"
+     * @param set_persistent_db: set-persistent-db flag - default is true (write-through).
```
morantr: You should not add colons between the argument name to the description.
Second, does set_persistent_db means "Save configuration on persistent database"?
Because it is much more understandable (at least for me) to phrase it like that.

controller/src/beerocks/master/db/db.h
```diff
@@ -589,6 +589,158 @@ class db {
...
+    bool update_client_persistent_db(const sMacAddr &mac);
+
+    /**
+     * @brief Get all clients from persistent db
+     * 
+     * @return unordered map of clients, for each client unordered map of params as key-value
+     * @return an empty map of clients is returned if the persistent db is empty
+     */
+    std::unordered_map<std::string, std::unordered_map<std::string, std::string>>
+    get_all_persistent_db_clients();
```
morantr: Please add dots at the end of the sentences.
Second, maybe you could define a type for this return value with Doxygen documentation and then use it here. This could make the return value explanation much more simple.
Also, when using MAC address please use sMacAddr and not std::string.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2423,73 +2423,257 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+    return n->client_persistent_data_last_edit;
 }
 
 bool db::set_client_time_life_delay(const sMacAddr &mac,
                                     const std::chrono::seconds &time_life_delay_sec,
                                     bool set_persistent_db)
 {
+    auto n = get_node(mac);
+    if (!n) {
+        LOG(ERROR) << "node not found.... ";
```
morantr: What are the dots at the end of the sentence are good for? 
Also, I think that printing the given node is valuable information.

Apply this to several places on this file.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2423,73 +2423,257 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    } else if (n->get_type() != beerocks::TYPE_CLIENT) {
+        LOG(ERROR) << __FUNCTION__ << "node " << tlvf::mac_to_string(mac)
```
morantr: No need to ` tlvf::mac_to_string(mac)`, it is done automatically.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2423,73 +2423,257 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+                   << " is not a valid client!";
+        return false;
+    }
+
+    LOG(DEBUG) << __FUNCTION__ << ", time_life_delay_sec = " << time_life_delay_sec.count();
```
morantr: If time_life_delay_sec is in a unit of seconds, Why do you use `count()` on it? It doesn't make sense.
I would expect that it will be really with a second unit.

controller/src/beerocks/master/db/db.cpp
morantr: I also don't understand. I thought that if `set_persistent_db` is set, we should save something on the persistent db.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2423,73 +2423,257 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    } else if (n->get_type() != beerocks::TYPE_CLIENT) {
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid client!";
+        return false;
+    }
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (set_persistent_db) {
+        LOG(DEBUG) << __FUNCTION__
+                   << ", configuring persistent-db, stay_on_initial_radio = " << std::boolalpha
```
morantr: boolalpha is configured by default on beerocks logging. No need to add it.

morantr (approved):
Nothing critical, so approving.

controller/src/beerocks/master/db/db.cpp
adam1985d: @simonbaren 
client persistent data can be set to the runtime db with multiple calls and then a call to `update_client_persistent_db()` will sync the runtime configuration to the persistent DB. the timestamp is handling the time of last edit (of configuration) and not the last time of actual synchronization to the persistent DB.
in any case - as part of the orderly shutdown of the controller - the runtime configuration should be saved to the persistent db - but again - with the timestamp indicating the last time of client configuration edit and not when it was synched to the persistent DB!
and naming suggestion is welcomed

@morantr - this is correct, but I will quote the draft PR description:
"Draft PR - missing the calls to BPL APIs will be added after #1460 is merged"
(at time of the draft PR creation BPL interface was not available yet)

common/beerocks/bcl/include/bcl/beerocks_defines.h
adam1985d: moved it to the `node.h`

controller/src/beerocks/master/db/node.cpp
adam1985d: fixed

controller/src/beerocks/master/db/node.cpp
adam1985d: removed the cast

controller/src/beerocks/master/db/node.cpp
adam1985d: done

controller/src/beerocks/master/db/db.cpp
adam1985d: fixed

controller/src/beerocks/master/db/db.cpp
adam1985d: removed it

controller/src/beerocks/master/db/db.cpp
adam1985d: reomved it

controller/src/beerocks/master/db/db.cpp
adam1985d: as we dicsussed offline, leaving it as is since the main usage is in the std::chrono:seconds() type

controller/src/beerocks/master/db/db.h
adam1985d: done

controller/src/beerocks/master/db/db.h
adam1985d: done

controller/src/beerocks/master/db/db.h
adam1985d: ok - removed extra spacing

controller/src/beerocks/master/db/db.h
adam1985d: replaced all occurrences

controller/src/beerocks/master/db/db.h
adam1985d: done

controller/src/beerocks/master/db/db.h
adam1985d: fixed all occurrences

controller/src/beerocks/master/db/node.h
adam1985d: fixed - if that makes it cleared for other devs I'm happy to adopt

controller/src/beerocks/master/db/node.h
```diff
@@ -35,6 +35,14 @@ typedef struct {
     bool backhaul_vap;
 } sVapElement;
 
+/**
+ * @brief Extended boolean parameter to support "not configured" value for persistent configuration.
+ * For persistent data, it is important to differ between configured enable/disable to uncofigured value.
+ */
+enum class ePersistentParamBool : int8_t { NOT_CONFIGURED = -1, DISABLE = 0, ENABLE = 1 };
```
vitalybu, line 42:
```suggestion
enum class ePersistency : int8_t { NOT_CONFIGURED = -1, DISABLE = 0, ENABLE = 1 };
```

controller/src/beerocks/master/db/db.h
vitalybu: Why do you want to make this method client-specific?
Tomorrow we can additional persistent values, why not have a single method for loading all the information from the DB?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -4018,3 +4019,22 @@ void db::set_prplmesh(const sMacAddr &mac)
     }
     get_node(mac)->is_prplmesh = true;
 }
+
+bool db::update_client_entry_in_persistent_db(
+    const sMacAddr &mac, std::unordered_map<std::string, std::string> values_map)
+{
+    auto db_entry = client_db_entry_from_mac(mac);
+
+    if (!bpl::db_has_entry(type_to_string(beerocks::eType::TYPE_CLIENT), db_entry)) {
```
vitalybu: If you convert the type to string more than once, it's probably better to use a variable.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2475,11 +2475,38 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+
+    return bpl::db_has_entry(type_to_string(beerocks::eType::TYPE_CLIENT), client_db_entry);
+}
 
 bool db::add_client_to_persistent_db(const sMacAddr &mac,
                                      std::unordered_map<std::string, std::string> params)
 {
+    auto db_entry = client_db_entry_from_mac(mac);
+
+    if (!bpl::db_has_entry(std::string("type"), db_entry)) {
```
vitalybu, line 2497: Why not simply:
```suggestion
    if (!bpl::db_has_entry("type", db_entry)) {
```

vitalybu: Also, what is the API of `bpl::db_has_entry()`?
What does it mean to `bpl::db_has_entry("type", ...)`?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2516,6 +2543,16 @@ bool db::set_client_time_life_delay(const sMacAddr &mac,
     if (save_to_persistent_db) {
         LOG(DEBUG) << __FUNCTION__
                    << ", configuring persistent-db, timelife = " << time_life_delay_sec.count();
+
+        std::unordered_map<std::string, std::string> values_map;
+        values_map.at("timestamp") = timestamp_to_string_seconds(timestamp);
```
vitalybu: Why do you need a function to convert the timestamp to seconds?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2721,16 +2798,102 @@ bool db::clear_client_persistent_db(const sMacAddr &mac)
...
+    if (n->client_parameters_last_edit == std::chrono::steady_clock::time_point::min()) {
+        LOG(DEBUG) << __FUNCTION__ << "persistent client parameters are empty for " << mac
+                   << ", no need to update persistent-db";
+        return true;
+    }
+
+    std::unordered_map<std::string, std::string> values_map;
+
+    //fill values map of client persistent params
+    values_map.at("timestamp") = timestamp_to_string_seconds(n->client_parameters_last_edit);
```
vitalybu: All static strings (e.g. "timestamp", "timelife" etc.) should be constexpr.

controller/src/beerocks/master/db/node.h
adam1985d: the enum doesn't only provide an indication of whether the value is persistent or not - it also provides its value.
I am open to naming suggestions but think `ePersistency` is misleading.

controller/src/beerocks/master/db/db.cpp
adam1985d: fixed on all occurrences

controller/src/beerocks/master/db/db.cpp
adam1985d: an entry (in UCI correlates to section) has a type and a name - all clients have a section-type `client` and looking them up in DB is by the client mac which is used as its name.
similar to the `db::has_node()` the `bpl::db_has_entry()` returns true if entry exist in DB.
to prevent a situation where a mac of the same client was used to describe something that **is not** of section-type `client` we can also specify expected entry type. (if we do not provide it - no validation of entry-type will be performed)

```
/**
 * @brief find if entry exists
 * 
 * @param[in] entry_type type of the requested entry.
 * @param[in] entry_name name of the requested entry.
 * @return true if entry exists in db, false otherwise
 */
bool db_has_entry(const std::string &entry_type, const std::string &entry_name);
```

controller/src/beerocks/master/db/db.cpp
adam1985d: I am using this conversion in many places in the code so I thought it would be beneficial to have:
1. meaningful function to understand the operation performed
2. single-point-of-change in case we wish to change how we save the timestamp in the future

controller/src/beerocks/master/db/db.cpp
adam1985d: the map expects `std::string` while in c++11 it is not supported as `constexpr`
I have changed to `static const std::string` at the beginning of the `db.cpp` file

reference to the mentioned above:
https://stackoverflow.com/questions/27123306/is-it-possible-to-use-stdstring-in-a-constexpr

P.S. - I didn't use the `constexpr char constString[] = "constString";` example they suggested as the this would cause a runtime `std::string` creation from `char[]` on every use.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -19,6 +20,66 @@ using namespace beerocks_message;
...
+static const std::string timelife_delay_str       = "timelife";
+static const std::string initial_radio_enable_str = "initial_radio_enable";
+static const std::string initial_radio_str        = "initial_radio";
+static const std::string selected_band_enable_str = "selected_band_enable";
+static const std::string selected_bands_str       = "selected_bands";
+
+// static
+std::string db::type_to_string(beerocks::eType type)
+{
+    if (type == beerocks::eType::TYPE_GW) {
```
VladyslavTupikin: I think the switch-case will be better than if-else here.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+    }
+
+    return n->client_parameters_last_edit;
+}
+
+bool db::set_client_time_life_delay(const sMacAddr &mac,
+                                    const std::chrono::seconds &time_life_delay_sec,
+                                    bool save_to_persistent_db)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+
+    n->client_time_life_delay_sec  = time_life_delay_sec;
+    n->client_parameters_last_edit = timestamp;
+
+    return true;
+}
+
+std::chrono::seconds db::get_client_time_life_delay(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return std::chrono::seconds::zero();
+    }
+
+    return n->client_time_life_delay_sec;
+}
+
+bool db::set_client_stay_on_initial_radio(const sMacAddr &mac, bool stay_on_initial_radio,
+                                          bool save_to_persistent_db)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    } else if (n->get_type() != beerocks::TYPE_CLIENT) {
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid client!";
+        return false;
+    }
+
+    LOG(DEBUG) << __FUNCTION__ << ", stay_on_initial_radio = " << stay_on_initial_radio;
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
```
VladyslavTupikin, line 2587: This code is copy-paste from the method `set_client_time_life_delay`. I think you should create a new method for updating the database for avoid copy-paste issues.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+    n->client_stay_on_initial_radio =
+        (stay_on_initial_radio) ? ePersistentParamBool::ENABLE : ePersistentParamBool::DISABLE;
+    n->client_parameters_last_edit = timestamp;
+
+    return true;
+}
+
+ePersistentParamBool db::get_client_stay_on_initial_radio(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return ePersistentParamBool::NOT_CONFIGURED;
+    }
+
+    return n->client_stay_on_initial_radio;
+}
+
+bool db::set_client_initial_radio(const sMacAddr &mac, const sMacAddr &initial_radio_mac,
+                                  bool save_to_persistent_db)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    } else if (n->get_type() != beerocks::TYPE_CLIENT) {
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid client!";
+        return false;
+    }
+
+    LOG(DEBUG) << __FUNCTION__ << ", initial_radio = " << initial_radio_mac;
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
```
VladyslavTupikin, line 2631: Code copy-paste. See the commentary above.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+
+    n->client_initial_radio        = initial_radio_mac;
+    n->client_parameters_last_edit = timestamp;
+
+    return true;
+}
+
+sMacAddr db::get_client_initial_radio(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return network_utils::ZERO_MAC;
+    }
+
+    return n->client_initial_radio;
+}
+
+bool db::set_client_stay_on_selected_band(const sMacAddr &mac, bool stay_on_selected_band,
+                                          bool save_to_persistent_db)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    } else if (n->get_type() != beerocks::TYPE_CLIENT) {
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid client!";
+        return false;
+    }
+
+    LOG(DEBUG) << __FUNCTION__ << ", stay_on_selected_band = " << stay_on_selected_band;
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
```
VladyslavTupikin, line 2674: Copy-paste code.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+    n->client_stay_on_selected_band =
+        (stay_on_selected_band) ? ePersistentParamBool::ENABLE : ePersistentParamBool::DISABLE;
+    n->client_parameters_last_edit = timestamp;
+
+    return true;
+}
+
+ePersistentParamBool db::get_client_stay_on_selected_band(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return ePersistentParamBool::NOT_CONFIGURED;
+    }
+
+    return n->client_stay_on_selected_band;
+}
+
+bool db::set_client_selected_bands(const sMacAddr &mac, beerocks::eFreqType selected_bands,
+                                   bool save_to_persistent_db)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    } else if (n->get_type() != beerocks::TYPE_CLIENT) {
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid client!";
+        return false;
+    }
+
+    LOG(DEBUG) << __FUNCTION__ << ", selected_band = " << int(selected_bands);
+
+    auto timestamp = std::chrono::steady_clock::now();
+    if (save_to_persistent_db) {
```
VladyslavTupikin, line 2718: Copy-paste code.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+
+    n->client_selected_bands       = selected_bands;
+    n->client_parameters_last_edit = timestamp;
+
+    return true;
+}
+
+beerocks::eFreqType db::get_client_selected_bands(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        LOG(ERROR) << __FUNCTION__ << "node " << mac << " is not a valid client!";
+        return beerocks::eFreqType::FREQ_UNKNOWN;
+    }
+
+    return n->client_selected_bands;
+}
+
+bool db::clear_client_persistent_db(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        LOG(ERROR) << "failed to remove client entry " << db_entry;
+        return false;
+    }
+
+    return true;
+}
+
+bool db::update_client_persistent_db(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

controller/src/beerocks/master/db/node.cpp
```diff
@@ -29,6 +29,18 @@ node::node(beerocks::eType type_, const std::string &mac_)
...
+        os << "Disabled";
+    } else if (value == ePersistentParamBool::ENABLE) {
+        os << "Enabled";
+    } else {
+        os << "Not-Configured";
+    }
+    return os;
+}
+
 std::ostream &operator<<(std::ostream &os, const node &n)
```
VladyslavTupikin, line 44:
```suggestion
std::ostream &operator<<(std::ostream &os, const node &node)
```
Looks weird but in the code, I prefer to see node instead of magic n.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -19,6 +20,66 @@ using namespace beerocks_message;
...
+    } else if (type == beerocks::eType::TYPE_UNDEFINED) {
+        return std::string();
+    }
+    return std::string();
+}
+
+std::string db::client_db_entry_from_mac(const sMacAddr &mac)
+{
+    std::string db_entry = tlvf::mac_to_string(mac);
+    std::replace(db_entry.begin(), db_entry.end(), ':', '_');
```
VladyslavTupikin, line 56: Why do we need to replace ':' character?

controller/src/beerocks/master/db/db.cpp
```diff
@@ -2418,6 +2479,431 @@ const std::list<sChannelScanResults> &db::get_channel_scan_results(const sMacAdd
...
+        return false;
+    }
+    LOG(DEBUG) << "added client entry " << db_entry << " to persistent db";
+
+    return true;
+}
+
+std::chrono::steady_clock::time_point db::get_client_parameters_last_edit(const sMacAddr &mac)
+{
+    auto n = get_node(mac);
```
VladyslavTupikin:
```suggestion
    auto node = get_node(mac);
```

VladyslavTupikin (changes requested):

controller/src/beerocks/master/db/db.cpp
adam1985d: both options are valid.
I actually used to use switch-case for such implementations but received comments exactly the opposite to your comment ("please use if-else") - I guess it is a matter of taste

controller/src/beerocks/master/db/db.cpp
adam1985d: uci doesn't support `:` in the section name

controller/src/beerocks/master/db/node.cpp
adam1985d: on this specific location I am going to leave as is - but on the new introduced APIs (in following commits) will change from `n` to `node`

controller/src/beerocks/master/db/db.h
adam1985d: added a description to the parameter (in node) and elaborate a bit more in the function description

controller/src/beerocks/master/db/db.h
adam1985d: added a description to the parameter (in node) and elaborate a bit more in the function description

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: the next commit introduces the core-functionality which is inside this condition and moving it to external function more than it is in the next commit is not really beneficial IMO

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: see my response above

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: see my comment above

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: see my comments above

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced all new-code occurrences of `n` to `node`

controller/src/beerocks/master/db/node.cpp
VladyslavTupikin: agree

controller/src/beerocks/master/db/db.cpp
VladyslavTupikin: I disagree with you. The switch-case mechanism was written especially for situations as we can see here and use if-else for replacing standard switch-case looks weird.

controller/src/beerocks/master/db/db.cpp
```diff
@@ -19,6 +20,66 @@ using namespace beerocks_message;
...
+static const std::string initial_radio_enable_str = "initial_radio_enable";
+static const std::string initial_radio_str        = "initial_radio";
+static const std::string selected_band_enable_str = "selected_band_enable";
+static const std::string selected_bands_str       = "selected_bands";
+
+// static
+std::string db::type_to_string(beerocks::eType type)
+{
+    if (type == beerocks::eType::TYPE_GW) {
+        return std::string("gateway");
```
MuteSpirit: Why not use
```
return "gateway";
```
and not explicitly create `std::string` object?

controller/src/beerocks/master/db/db.cpp
MuteSpirit: I vote to `switch-case` only because the shorter code

adam1985d: replaced with switch-case that return `"*"` instead of `std::string("*")`

controller/src/beerocks/master/db/db.cpp
adam1985d: replaced with switch-case that return `"*"` instead of `std::string("*")`

VladyslavTupikin (approved):

PR #1498, opened 2020-06-30
boardfarm: add "prplWRT_STA" device

abelog:
## Description

This PR introduces HW representation of station and includes in `netgear-rax40-1` boardfarm configuration.
Changes the way, how `VirtualAPHostad`s which represent real AP are spawned in Radio.
Adds methods required to connect and verify that station is associated with AP.

## Changes

- Add `prplWRT_STA` device
- Add `get_ssid` and `get_psk` methods to the `VirtualAPHostapd`
- Fix `_device_wait_for_log` usage
- Add `wifi_connect` and related methods to the `prplWRT_STA` to associate station and verify connection to the AP

## Notes

`prplWRT_STA` class was tested using RPI3 as host. Commands and utilities used in that class are generic (`ip`, `wpa_supplicant` and etc), so it should work on any other platform (e.g: OpenWRT board, Debian-based host with USB WiFi adapter).
In order to `prplWRT_STA` class to work, user must be able to run `wpa_supplicant` without sudo. `sudo_sendline` method which should deal with `sudo` of boardfarm seems to be broken and doesn't work.

Part of: https://jira.prplfoundation.org/browse/PPM-12
Also, resolved https://jira.prplfoundation.org/browse/PPM-1 on the go.

### DIFF
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
@@ -5,6 +5,8 @@
 # See LICENSE file for more details.
 ###############################################################
 
+import pexpect
+
 from boardfarm.devices.debian_wifi import DebianWifi
 from boardfarm.devices import connection_decider
 from environment import VirtualAPHostapd
@@ -35,8 +37,10 @@ class PrplMeshStation(DebianWifi):
         self.driver_name = config.get("driver", "nl80211,wext")
         self.mac = self.get_mac()
 
-        self.wifi_disconnect()
         # kill all wpa_supplicant relevant to active interface
+        self.wifi_disconnect()
+        # Turn on and off wlan iface just in case
+        self.disable_and_enable_wifi()
 
     def wifi_connect(self, vap: VirtualAPHostapd) -> bool:
         """Connect to the Access Point. Return True if successful."""
@@ -55,10 +59,31 @@ class PrplMeshStation(DebianWifi):
         self.expect(self.prompt)
         # Start wpa_supplicant with created configuration
         # Typical coommand on RPI: wpa_supplicant -B -c/tmp/temp.conf -iwlan0 -Dnl80211,wext
-        self.sudo_sendline("wpa_supplicant -B -D{} -i {} -c {}".format(
+        self.sudo_sendline("wpa_supplicant -B -D{} -i{} -c{}".format(
             self.driver_name, self.iface_wifi, config_file_path))
         self.expect("Successfully initialized wpa_supplicant")
         return bool(self.match)
+
+    def wifi_connect_check(self, vap: VirtualAPHostapd) -> bool:
+        """Connect to a SSID and verify WIFI connectivity"""
+        for _ in range(5):
+            self.wifi_connect(vap)
+            self.expect(pexpect.TIMEOUT, timeout=10)
+            verify_connect = self.wifi_connectivity_verify()
+            if verify_connect:
+                break
+            else:
+                self.wifi_disconnect()
+        return verify_connect
+
+    def wifi_connectivity_verify(self):
+        """Verify that wifi is connected. Return bool"""
+        self.sendline("iw %s link" % self.iface_wifi)
+        matched = self.expect(["Connected", "Not connected", pexpect.TIMEOUT])
+        if matched == 0:
+            return True
+        else:
+            return False
 
     def get_mac(self) -> str:
         """Get MAC of STA iface"""
@@ -67,5 +92,5 @@ class PrplMeshStation(DebianWifi):
         # wdev 0x1
         # addr 96:4e:c9:cc:7a:2c
         # type managed
-        self.expect("addr (?P<mac>..:..:..:..:..:..)\r\n\ttype")
+        self.expect("addr (?P<mac>..:..:..:..:..:..)\r\n\t(type|ssid)")
         return self.match.group('mac')
--- a/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
+++ b/tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
@@ -1,0 +1,71 @@
+###############################################################
+# SPDX-License-Identifier: BSD-2-Clause-Patent
+# SPDX-FileCopyrightText: 2020 the prplMesh contributors (see AUTHORS.md)
+# This code is subject to the terms of the BSD+Patent license.
+# See LICENSE file for more details.
+###############################################################
+
+from boardfarm.devices.debian_wifi import DebianWifi
+from boardfarm.devices import connection_decider
+from environment import VirtualAPHostapd
+
+
+class PrplMeshStation(DebianWifi):
+    """Client of prplMesh enabled Access Point."""
+
+    linesep = "\r"
+    model = "prplWRT_STA"
+    prompt = ['.*:~', '/.*#']
+
+    def __init__(self, *args, **kwargs):
+        """Init station and wlan iface."""
+        self.args = args
+        self.kwargs = kwargs
+
+        config = kwargs.get("config", kwargs)
+        self.connection_type = config.get("connection_type", None)
+        self.connection = connection_decider.connection(device=self,
+                                                        conn_type=self.connection_type,
+                                                        **kwargs)
+        self.connection.connect()
+        self.consoles = [self]
+        super(PrplMeshStation, self).__init__(*args, **kwargs)
+        self.iface_dut = self.iface_wifi = self.kwargs.get(
+            'iface', 'wlan0')
+        self.driver_name = config.get("driver", "nl80211,wext")
+        self.mac = self.get_mac()
+
+        self.wifi_disconnect()
+        # kill all wpa_supplicant relevant to active interface
+
+    def wifi_connect(self, vap: VirtualAPHostapd) -> bool:
+        """Connect to the Access Point. Return True if successful."""
+        config_file_name = "boardfarm_tmp.conf"
+        config_file_path = "/tmp/{}".format(config_file_name)
+
+        # Create network configuration for SSID
+        bssid = "bssid={}".format(vap.bssid)
+        ssid = "ssid=\"{}\"".format(vap.get_ssid())
+        key = "psk=\"{}\"".format(vap.get_psk())
+        network_config = "network={{\n{}\n{}\n{}\n}}".format(bssid, ssid, key)
+        # Clean up previous configuration
+        self.sendline("rm -f \"{}\"".format(config_file_path))
+        self.expect(self.prompt)
+        self.sendline("echo \'{}\' > \"{}\"".format(network_config, config_file_path))
+        self.expect(self.prompt)
+        # Start wpa_supplicant with created configuration
+        # Typical coommand on RPI: wpa_supplicant -B -c/tmp/temp.conf -iwlan0 -Dnl80211,wext
+        self.sudo_sendline("wpa_supplicant -B -D{} -i {} -c {}".format(
+            self.driver_name, self.iface_wifi, config_file_path))
+        self.expect("Successfully initialized wpa_supplicant")
+        return bool(self.match)
+
+    def get_mac(self) -> str:
+        """Get MAC of STA iface"""
+        self.sendline("iw {} info".format(self.iface_dut))
+        # We are looking for MAC definition of STA
+        # wdev 0x1
+        # addr 96:4e:c9:cc:7a:2c
+        # type managed
+        self.expect("addr (?P<mac>..:..:..:..:..:..)\r\n\ttype")
+        return self.match.group('mac')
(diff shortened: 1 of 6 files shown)


abelog:
@mergify refresh

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
```diff
@@ -0,0 +1,71 @@
...
+        ssid = "ssid=\"{}\"".format(vap.get_ssid())
+        key = "psk=\"{}\"".format(vap.get_psk())
+        network_config = "network={{\n{}\n{}\n{}\n}}".format(bssid, ssid, key)
+        # Clean up previous configuration
+        self.sendline("rm -f \"{}\"".format(config_file_path))
+        self.expect(self.prompt)
+        self.sendline("echo \'{}\' > \"{}\"".format(network_config, config_file_path))
+        self.expect(self.prompt)
+        # Start wpa_supplicant with created configuration
+        # Typical coommand on RPI: wpa_supplicant -B -c/tmp/temp.conf -iwlan0 -Dnl80211,wext
```
tomereli, line 61: What about WPS? this assumes you already know the network, meaning after WPS or manual configuration. But in practice, and especially in certification, WPS onboarding always happens so there is no network configuration involved - everything happens as part of WPS.

About the RPI network configuration - does it have a bridge? If so, I suggest to use the `-b` option of the wpa_supplicant (provided it is available in upstream hostapd) so it will manage the STA interface bridge operations.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
```diff
@@ -55,17 +59,38 @@ def wifi_connect(self, vap: VirtualAPHostapd) -> bool:
...
+            verify_connect = self.wifi_connectivity_verify()
+            if verify_connect:
+                break
+            else:
+                self.wifi_disconnect()
+        return verify_connect
+
+    def wifi_connectivity_verify(self):
+        """Verify that wifi is connected. Return bool"""
+        self.sendline("iw %s link" % self.iface_wifi)
```
tomereli, line 81: I think it is better to use `wpa_cli -iwlan0 stat` for that.

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
```diff
@@ -39,12 +31,11 @@ def runTest(self):
...
                                             tlv(0x95, 0x0006, '{sta_mac}'.format(sta_mac=sta_mac)))
         self.check_log(agent,
-                       "client with mac address {sta_mac} not found".format(sta_mac=sta_mac),
-                       timeout=30)
+                       "client with mac address {sta_mac} not found".format(sta_mac=sta_mac))
         time.sleep(1)
 
         debug('sta: {}'.format(sta.mac))
-        sta.wifi_connect(agent.radios[0].vaps[0])
+        sta.wifi_connect_check(agent.radios[0].vaps[0])
```
tomereli, line 38: Perhaps a better name should be `sta.connected()->bool`

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
```diff
@@ -54,8 +45,7 @@ def runTest(self):
         time.sleep(1)
         self.check_log(agent,
                        "Send AssociatedStaLinkMetrics to controller, mid = {}".format(mid),
-                       timeout=30)
-        sta.wifi_disconnect(agent.radios[0].vaps[0])
+                       timeout=20)
```
tomereli, line 48: This change is not described in the commit message

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
```diff
@@ -71,3 +61,4 @@ def teardown_class(cls):
             # If AttributeError was raised - we are dealing with dummy devices.
             # We don't have to additionaly send Ctrl+C for dummy devices.
             pass
+        test.dev.wifi.disable_wifi()
```
tomereli, line 64: Here as well, a better naming IMO is `test.dev.wifi.disable()`

tomereli (approved):

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
arnout: Actually, in certification the client is always associated with explicit SSID, password and BSSID. WPS is only used for associating the CTT Agent.

As to the bridge: we definitely don't want it to be in the bridge, because then we create a loop (we already have a backhaul connection from the RPi to the DUT).

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
```diff
@@ -0,0 +1,71 @@
...
+        self.iface_dut = self.iface_wifi = self.kwargs.get(
+            'iface', 'wlan0')
+        self.driver_name = config.get("driver", "nl80211,wext")
+        self.mac = self.get_mac()
+
+        self.wifi_disconnect()
+        # kill all wpa_supplicant relevant to active interface
+
+    def wifi_connect(self, vap: VirtualAPHostapd) -> bool:
+        """Connect to the Access Point. Return True if successful."""
```
arnout, line 46: IIUC, this function already exists in DebianWifi, but you override it to support WPA2, right? It would have been good to mention this in the commit message or in a comment (nitpick).

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
```diff
@@ -0,0 +1,71 @@
...
+        # Start wpa_supplicant with created configuration
+        # Typical coommand on RPI: wpa_supplicant -B -c/tmp/temp.conf -iwlan0 -Dnl80211,wext
+        self.sudo_sendline("wpa_supplicant -B -D{} -i {} -c {}".format(
+            self.driver_name, self.iface_wifi, config_file_path))
+        self.expect("Successfully initialized wpa_supplicant")
+        return bool(self.match)
+
+    def get_mac(self) -> str:
+        """Get MAC of STA iface"""
+        self.sendline("iw {} info".format(self.iface_dut))
```
arnout, line 90: iw output is not considered stable - its help output says so explicitly.

It would be better to use wpa_cli instead.

(nitpick - we can fix this when it becomes a problem)

tests/environment.py
```diff
@@ -492,6 +492,27 @@ class VirtualAPHostapd(VirtualAP):
     def __init__(self, radio: RadioHostapd, bssid: str):
         super().__init__(radio, bssid)
 
+    def get_ssid(self) -> str:
+        """Get current SSID of attached radio. Return string."""
+        device = self.radio.agent.device
+        device.sendline("iw {}.0 info".format(self.radio.iface_name))
```
arnout: Commit message says that you get it from the log file...

The `.0` is incorrect - it's only correct for the first VAP. When creating the VirtualAPHostapd, instead of creating two of them, we should look for the VAPs that actually exist and use them.

A better way would be to parse the hostapd.conf file. Even better is to use hostapd_cli, but I don't think you can get the credentials that way.

All of this is not needed for this PR.

tests/environment.py
```diff
@@ -476,26 +476,44 @@ def __init__(self, agent: ALEntityPrplWrt, iface_name: str):
...
         super().__init__(agent, mac)
 
-        VirtualAPHostapd(self, mac)
-        VirtualAPHostapd(self, mac)
+        # Find out amount of VAPs avalaible on device.
+        # If 0 - spawn one VAP to represent AP.
+        self.agent.device.sendline("ip link list | grep -c \"{}\\.\"".format(self.iface_name))
+        # Look for number of 1 or 2 digits surrounded by CRLF.
+        self.agent.device.expect("\r\n(?P<vaps>[0-9]{1,2})\r\n")
+        vap_amount = int(self.agent.device.match.group('vaps'))
```
arnout, line 484: Instead of counting and then generating it ourselves, it would be better to use the regular expression to get the list of actual VAP names, and initialize the VirtualAPHostapd objects with that.

And even better would be to use the hostapd_cli STATUS command to get them.

tests/environment.py
```diff
@@ -476,26 +476,44 @@ def __init__(self, agent: ALEntityPrplWrt, iface_name: str):
...
         # Multiply timeout by 100, as test sets it in float.
         return _device_wait_for_log(self.agent.device, "{}/beerocks_agent_{}.log".format(
             self.log_folder, self.iface_name), regex, timeout, start_line)
 
+    def get_mac(self, iface: str) -> str:
+        """Return mac of specified iface"""
+        device = self.agent.device
+        device.sendline("ip link show {}".format(iface))
+        device.expect("link/ether (?P<mac>([0-9a-fA-F]{2}:){5}[0-9a-fA-F]{2})")
+        return device.match.group('mac')
```
arnout, line 503: (nitpick) since this function is not specific to a radio, it may be better to move it to the PrplMeshPrplWRT class. Or even PrplMeshBase, because it applies to docker as well.

tests/environment.py
```diff
@@ -476,26 +476,44 @@ def __init__(self, agent: ALEntityPrplWrt, iface_name: str):
...
 
     def __init__(self, radio: RadioHostapd, bssid: str):
         super().__init__(radio, bssid)
+        self.iface = self.get_iface(self.bssid)
 
     def get_ssid(self) -> str:
         """Get current SSID of attached radio. Return string."""
         device = self.radio.agent.device
-        device.sendline("iw {}.0 info".format(self.radio.iface_name))
+        device.sendline("iw {} info".format(self.iface))
```
arnout, line 516: Ah, much better :-)

tests/environment.py
```diff
@@ -476,26 +476,44 @@ def __init__(self, agent: ALEntityPrplWrt, iface_name: str):
...
+        self.agent.device.sendline("ip link list | grep -c \"{}\\.\"".format(self.iface_name))
+        # Look for number of 1 or 2 digits surrounded by CRLF.
+        self.agent.device.expect("\r\n(?P<vaps>[0-9]{1,2})\r\n")
+        vap_amount = int(self.agent.device.match.group('vaps'))
+        if vap_amount == 0:
+            VirtualAPHostapd(self, mac)
+        else:
+            for vap_number in range(0, vap_amount):
+                vap_mac = self.get_mac("{}.{}".format(self.iface_name, vap_number))
+                VirtualAPHostapd(self, vap_mac)
```
arnout, line 490: It's a bit weird to first call get_mac to get the bssid from the interface name, and then call get_iface to get the interface name for that bssid again... Just use the interface name as an argument and get the corresponding bssid in the VirtualAPHostapd contructor.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
```diff
@@ -55,17 +59,38 @@ def wifi_connect(self, vap: VirtualAPHostapd) -> bool:
...
+        return verify_connect
+
+    def wifi_connectivity_verify(self):
+        """Verify that wifi is connected. Return bool"""
+        self.sendline("iw %s link" % self.iface_wifi)
+        matched = self.expect(["Connected", "Not connected", pexpect.TIMEOUT])
+        if matched == 0:
+            return True
+        else:
+            return False
```
arnout, line 86:
```suggestion
        return matched == 0
```
(nitpick)

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
arnout: No, because it's a combination of connecting, checking if it succeeded, and retrying.

arnout (approved):
I have a bunch of comments, as usual, but I think we should probably fix them later and not as part of this PR.

tests/environment.py
abelog: > Commit message says that you get it from the log file...

PSK for AP is received from the log, but not the SSID. I should recheck commit message if it states otherwise.

tests/environment.py
abelog: Main intention was to keep same interface with Dummy device. Is it worth to create a note in Confluence with notes about refactoring and separate task to refactor according to that note?

tests/boardfarm_plugins/boardfarm_prplmesh/tests/client_association_link_metrics.py
abelog: This is build-in boardfarm method, so we may create a wrapper.

tests/boardfarm_plugins/boardfarm_prplmesh/devices/prplmesh_station.py
abelog: You are right. We are overriding it to count in `bssid` setting (build-in method just don't care about `bssid` when we may have few boards with same SSID nearby) and to add WPA2 support.

abelog:
As there is no crucial commentaries to fix (nitpicks mainly), merging it as it is.

